PR #2074, opened 2018-01-16
Issue #2943: Add check to user page callback to prevent redirects.

quicksketch:
Fixes https://github.com/backdrop/backdrop-issues/issues/2943.

### DIFF
--- a/core/modules/user/user.module
+++ b/core/modules/user/user.module
@@ -3024,9 +3024,10 @@ function user_modules_uninstalled($modules) {
  * proper destination after a user has been properly logged in.
  */
 function user_login_destination() {
+  global $user;
   $destination = backdrop_get_destination();
   if ($destination['destination'] == 'user/login') {
-    $destination['destination'] = 'user';
+    $destination['destination'] = 'user/' . $user->uid;
   }
   return $destination;
 }
--- a/core/modules/user/user.pages.inc
+++ b/core/modules/user/user.pages.inc
@@ -557,6 +557,15 @@ function user_cancel_confirm($account, $timestamp = 0, $hashed_pass = '') {
  */
 function user_page() {
   global $user;
+
+  // To help users upgrading from versions of Backdrop prior to 1.9.0 where this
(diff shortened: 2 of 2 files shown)


core/modules/user/user.module
```diff
@@ -3024,9 +3024,10 @@ function user_modules_uninstalled($modules) {
  * proper destination after a user has been properly logged in.
  */
 function user_login_destination() {
+  global $user;
   $destination = backdrop_get_destination();
   if ($destination['destination'] == 'user/login') {
-    $destination['destination'] = 'user';
+    $destination['destination'] = 'user/' . $user->uid;
```
opi, line 3030: Maybe check for `$user->uid` first ?

core/modules/user/user.pages.inc
```diff
@@ -557,6 +557,14 @@ function user_cancel_confirm($account, $timestamp = 0, $hashed_pass = '') {
  */
 function user_page() {
   global $user;
+
+  // To help users upgrading from previous versions of Backdrop where this
```
opi: - s/previous/older than 1.9.0/

- Maybe mention the https://github.com/backdrop/backdrop-issues/issues/260 issue number

opi (commented):
Untested yet, but LGTM

core/modules/user/user.module
jenlampton: >Third-party authentication modules may use this function to determine the proper destination after a user has been properly logged in. 

After you've been "properly logged in", it's safe to assume that `$user->uid` already exists.

jenlampton:
I've tested this PR on two of my sites during the update to 1.9.0 and in both cases it solved the problem I faced last time. I think this is a nice addition for user/developer experience, and is well worth having a `menu_rebuid` happen in a strange place.

core/modules/user/user.module
opi: Fair enough

quicksketch:
Merged into 1.x and 1.9.x.

PR #2076, opened 2018-01-19
Issue #2922: [UX] Views: Add a '[hidden]' label to fields when exclud…

opi:
…ed from display


quicksketch:
Awesome! Thanks @opi! Merged into 1.x and 1.9.x.

PR #2077, opened 2018-01-19
2776 drupal 2451449 users contact setting reverts to default when saving

opi:
[Issue #2451449 by drumm: User's contact setting reverts to default when saving](http://drupal.org/node/2451449)
Commit [ef9608dc1d](http://drupalcode.org/project/drupal.git/commit/ef9608dc1d8b7ef0eaa7b76b07b64a16f73e3bbe)

### DIFF
--- a/core/modules/contact/contact.module
+++ b/core/modules/contact/contact.module
@@ -269,7 +269,12 @@ function contact_form_user_profile_form_alter(&$form, &$form_state) {
  */
 function contact_user_presave($account) {
   $config = config('contact.settings');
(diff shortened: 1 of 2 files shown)


opi:
2 commits here as one is not a Drupal's crossport, but a simple fix that became evident to me when looking at this part of the code.
This commit could be removed and discussed in another issue if you think so

quicksketch:
Looks great, thanks @opi! Merged into 1.9.x and 1.x!

PR #2079, opened 2018-01-22
Issue #2949: Provide chained term tokens for taxonomy_term_reference fields

juanolalla:
Implements a solution for issue https://github.com/backdrop/backdrop-issues/issues/2949


jenlampton:
Replaced by https://github.com/backdrop/backdrop/pull/2874

PR #2081, opened 2018-01-25
Issue #2955: Add a visbility condition for taxonomy term depth.

### LINKED ISSUE #2955: [UX] Added a "Taxonomy term: Depth" Visibility condition for blocks and layouts.

Just as I often have different layouts for nodes, by node-type, I often need different layouts for categories based on term-depth. For example, my top-level terms need landing pages, but my deeper terms could use their own layout (or even the default layout).

However, without a Visibility condition to select only the top-level terms, I can't do that. Here's one that matches the term depth condition from Panels. 

Backdrop now includes a "Taxonomy term: Depth" Visibility condition for blocks and layouts.  This condition will allow a different block or layout to be selected, depending on the depth of the term you are viewing. For example: a top-level term can have its own landing page layout, and child terms could use the default layout, instead.

---

Advocate: @stpaultim

jenlampton:
Fixes https://github.com/backdrop/backdrop-issues/issues/2955

### DIFF
--- a/core/modules/taxonomy/plugins/access/taxonomy_term_depth_access.inc
+++ b/core/modules/taxonomy/plugins/access/taxonomy_term_depth_access.inc
@@ -1,0 +1,97 @@
+<?php
+/**
+ * @file
+ * Plugin to provide access control based upon taxonomy term depth.
+ */
+class TaxonomyTermDepthLayoutAccess extends LayoutAccess {
+  /**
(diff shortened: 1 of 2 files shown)


jenlampton:
PR updated with explanation and example.

jenlampton:
Replaced by https://github.com/backdrop/backdrop/pull/4062

PR #2068, opened 2018-01-15
Issue #2562: Make image paths support subdirectory installations.

quicksketch:
Fixes https://github.com/backdrop/backdrop-issues/issues/2562.

PR #2084, opened 2018-01-30
Issue #2963 Include managed files in autocomplete for links.

Graham-72:
This is my initial attempt to add files to the links autocomplete function.

### DIFF
--- a/core/modules/filter/filter.pages.inc
+++ b/core/modules/filter/filter.pages.inc
@@ -111,15 +111,11 @@ function _filter_link_autocomplete($string) {
       ->range(0, 10 - $match_count)
       ->execute();
     foreach ($results as $result) {
-      $path = $result->uri;
-      // The stored uri may be prefixed with 'public://'
-      // If so, this must be replaced to obtain a local path
-      // modified to allow for installation in a subdirectory.
-      // To do: provide for a uri prefixed with 'private://'
-      if(substr($path,0,9) == 'public://'){
-        $public_path = config_get('system.core','file_public_path');
-        $path = parse_url($GLOBALS['base_url'], PHP_URL_PATH) . '/' . $public_path . '/' . substr($path,9);
-      }
+      // Try to make a local path if possible for better portability.
+      $absolute_path = parse_url($GLOBALS['base_url'], PHP_URL_PATH) . '/';
+      $url = file_create_url($result->uri);
+      $path = str_replace($GLOBALS['base_url'] . '/', $absolute_path, $url);
+      
       $matches[$path] = check_plain($result->filename) . ' (' . t('File') . ')';
       $match_count++;
     }
--- a/core/modules/filter/filter.pages.inc
+++ b/core/modules/filter/filter.pages.inc
@@ -100,6 +100,25 @@ function _filter_link_autocomplete($string) {
       $matches['/' . $path] = check_plain($result->title) . ' (' . t('Content') . ')';
       $match_count++;
     }
+    
+    				// Get file matches.
+    $results = db_select('file_managed')
+      ->fields('file_managed', array('fid', 'filename', 'uri', 'filemime', 'status'))
+      ->condition('filename', db_like($string) . '%', 'LIKE')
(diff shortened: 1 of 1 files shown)


core/modules/filter/filter.pages.inc
```diff
@@ -100,6 +100,25 @@ function _filter_link_autocomplete($string) {
...
+      ->fields('file_managed', array('fid', 'filename', 'uri', 'filemime', 'status'))
+      ->condition('filename', db_like($string) . '%', 'LIKE')
+      ->condition('status', 1)
+						->condition('filemime', 'image%', 'NOT LIKE')
+      ->orderBy('fid', 'DESC')
+      ->range(0, 10)
+      ->execute();
+    foreach ($results as $result) {
+      $path = $result->uri;
+      if(substr($path,0,9) == 'public://'){
```
opi: Maybe add a comment to explain this part. Right now, I don't understand it.

core/modules/filter/filter.pages.inc
```diff
@@ -100,6 +100,25 @@ function _filter_link_autocomplete($string) {
...
     }
+    
+    				// Get file matches.
+    $results = db_select('file_managed')
+      ->fields('file_managed', array('fid', 'filename', 'uri', 'filemime', 'status'))
+      ->condition('filename', db_like($string) . '%', 'LIKE')
+      ->condition('status', 1)
+						->condition('filemime', 'image%', 'NOT LIKE')
+      ->orderBy('fid', 'DESC')
+      ->range(0, 10)
```
opi: `range` should handle `$max_count` as for users, see https://github.com/backdrop/backdrop/pull/2084/files#diff-8d9f5bbb77a85ec42c32392f64402cc4R128

opi (commented):
untested, only visual review. Nice improvment though, thx !

core/modules/filter/filter.pages.inc
Graham-72: Thanks @opi I hadn't spotted this.

core/modules/filter/filter.pages.inc
Graham-72: @opi I have added some comments that I hope will help.

core/modules/filter/filter.pages.inc
```diff
@@ -100,6 +100,25 @@ function _filter_link_autocomplete($string) {
...
+      ->condition('status', 1)
+      ->condition('filemime', 'image%', 'NOT LIKE')
+      ->orderBy('fid', 'DESC')
+      ->range(0, 10 - $match_count)
+      ->execute();
+    foreach ($results as $result) {
+      // Try to make a local path if possible for better portability.
+      $absolute_path = parse_url($GLOBALS['base_url'], PHP_URL_PATH) . '/';
+      $url = file_create_url($result->uri);
+      $path = str_replace($GLOBALS['base_url'] . '/', $absolute_path, $url);
```
opi, line 117: I don't really understand what you try to achieve in the last 3 lines. Also, the usage of `base_path()` can make the code cleaner.  But why the output of `file_create_url()` is not good as is ?

Graham-72: This is to do with the problem of Backdrop installed in subdirectories identified and solved elsewhere in this file thanks to #2068 which was for images but also applies to files uploaded via WYSIWYG.

opi: this is why `base_path()` exists, isn't it ?

Graham-72: Well, base_path() certainly works for my installations in subdirectories, and I am not sure why issue #2562 has been fixed this other way. I have copied that solution thinking there must be configurations where it is needed rather than base_path().

jenlampton: it could be that the other solution was imperfect. I definitely think `base_path` is the way to go here. Can we try it with base_path() instead? I hate to see those globals used this way. We can file a follow-up to https://github.com/backdrop/backdrop/pull/2068 to fix that issue.

PR #2089, opened 2018-02-01
Part fix for issue #2965 allowing for subdirectory.

Graham-72:
This adds the subdirectory name into the value of the image_path setting being stored in config for the layout.


quicksketch:
Thanks @Graham-72! This seems to work and the code looks good. Merged into 1.x and 1.9.x. Thanks!

PR #2091, opened 2018-02-02
Issue 2956: Make custom blocks translateable

bjoern-st:
This solves https://github.com/backdrop/backdrop-issues/issues/2956


bjoern-st:
@olafgrabienski @herbdool 
I made the changes and updated the PR. Now all the tests are passed successfully.
http://2091.backdrop.backdrop.qa.backdropcms.org/

core/modules/block/block.admin.inc
```diff
@@ -62,14 +68,32 @@ function block_admin_list() {
...
+    elseif (isset($form['langcode'])) {
+      $langcode = $form['langcode']['#value'];
+    }
+    // Set or load the block.
+    if (is_array($delta)) {
+      $custom_block = $delta;
+    }
+    else {
+      $custom_block = block_custom_block_load($delta, $langcode);
+    }
```
quicksketch: Is this if/else statement necessary? I can't see where it's called with $delta being the block. Generally it's good to avoid a parameter taking on multiple meanings. If this is necessary, the docblock needs to be updated. If it's not necessary it should be removed.

core/modules/block/block.admin.inc
```diff
@@ -78,6 +102,13 @@ function block_admin_configure($form, &$form_state, $delta = NULL) {
     backdrop_set_title(t('Add custom block'));
   }
 
+  // Set title for translate block form.
+  if (isset($form['langcode'])) {
+    $language_list = language_list();
+    $language = $language_list[$form['langcode']['#value']]->name;
```
quicksketch: Simplify using `$language = language_load($form['langcode']['#value'])`, though you'll need to use `$language->name` in the `backdrop_set_title()` call just below.

core/modules/block/block.admin.inc
```diff
@@ -95,8 +126,16 @@ function block_admin_configure($form, &$form_state, $delta = NULL) {
...
   $options = backdrop_parse_url($path);
+  // Check if we have a clean path without language prefix.
+  if (module_exists('language')) {
+    $path = $options['path'];
+    $language_list = language_list();
+    $result = language_url_split_prefix($path, $language_list);
+    if ($result[0] !== FALSE) {
+      $options['path'] = $result[1];
+    }
+  }
```
quicksketch: I'm not sure what's going on here. Could you explain what this does?

core/modules/block/block.admin.inc
```diff
@@ -208,3 +247,140 @@ function block_custom_block_delete_submit($form, &$form_state) {
...
+ * Form constructor for the block translation list form.
+ *
+ * @param $form
+ *   The form.
+ * @param $form_state
+ *   The form_state
+ * @param $delta
+ *   Unique ID of the block within the context of $module.
+ * @ingroup forms
+ */
```
quicksketch: Typical docblocks for forms include:

```
 * @param array $form
 *   An associative array containing the structure of a portion of the form.
 * @param array $form_state
 *   A keyed array containing the current state of the form.
```

Adding string to `@param string $delta` would also be good.

core/modules/block/block.module
```diff
@@ -149,10 +174,26 @@ function block_block_view($delta = '') {
  *   - body: Block contents.
  *   - format: Filter ID of the filter format for the body.
```
quicksketch, line 175: The new keys `default_language` and `translations` should be documented here

quicksketch: I'd also recommend we introduce `langcode` as a property here so we can tell which language is currently loaded into the block's body and title.

core/modules/block/block.module
```diff
@@ -219,6 +261,24 @@ function block_custom_block_form($edit = array(), $stand_alone = TRUE) {
...
+  // Add language select dropdown if edit info is empty.
+  // This combination should only be set if adding a new block.
+  if (isset($edit['default_language']) && empty($edit['info'])) {
+    $options = array(LANGUAGE_NONE => t('- None -'));
+    $language_list = language_list();
+    foreach ($language_list as $value) {
+      $options[$value->langcode] = $value->name;
+    }
+
+    $form['default_language'] = array(
```
quicksketch: Renaming this to `default_langcode` would help identify this value as being a string, rather than a language object.

core/modules/block/block.module
```diff
@@ -237,6 +297,11 @@ function block_custom_block_form($edit = array(), $stand_alone = TRUE) {
       '#default_value' => TRUE,
       '#disabled' => TRUE,
     );
+
+    // Add a notice for translation.
+    if (module_exists('language')) {
+      $form['reusable']['#description'] .= t('NOTE: You have module "language" enabled. If you want to make this block translatable, this block must be marked as reusable.');
```
quicksketch: I would suggest simplifying this down to "If you want to make this block translatable, this block must be marked as reusable."

core/modules/block/block.module
```diff
@@ -267,11 +337,36 @@ function block_custom_block_save(array $edit, $delta = NULL) {
...
       'body' => array('value' => '', 'format' => ''),
+      'default_language' => LANGUAGE_NONE,
     );
   }
-  foreach ($block as $key => $value) {
-    if (isset($edit[$key])) {
-      $block[$key] = $edit[$key];
+  // If langcode == NULL we come from add or edit form, so
+  // store the default values.
+  // Same if the langcode is the deafault language.
```
quicksketch: `deafault` => `default`

core/modules/layout/includes/block.text.inc
```diff
@@ -79,6 +79,10 @@ class BlockText extends Block {
         '#description' => t('If enabled, this block will be made reusable across different layouts and be listed in on the !block_link page.', array('!block_link' => l(t('Custom blocks'), 'admin/structure/block'))),
         '#weight' => 1,
       );
+      // Add a notice for translation.
+      if (module_exists('language')) {
+        $form['convert']['reusable']['#description'] .= '<br />' . t('To make this block translatable, it must be made reusable as well.');
```
quicksketch: Let's use the same translation string as we use in the normal form (whatever that becomes).

core/modules/block/block.module
```diff
@@ -8,11 +8,16 @@
  * Implements hook_permission().
  */
 function block_permission() {
-  return array(
+  $permissions = array(
     'administer blocks' => array(
       'title' => t('Administer custom blocks'),
     ),
   );
+  if (module_exists('translation')) {
```
quicksketch: I don't think Translation module is needed here, this looks like it should just require `'language'` like everywhere else in this PR.

quicksketch (changes requested):
I have some implementation suggestions but the functionality looks good!

Some key points:
- Let's use `default_langcode` instead of `default_language`. Langcode makes it clear that it's a string and not a language object.
- `block_block_info()` needs to be updated to use the new translated "info" value. Also we should trim the unnecessary properties from the returned values there, and just return "info" and "description" as would be expected.

core/modules/block/block.admin.inc
bjoern-st: No - this one is not needed any more. It is an relict of an former approach. 
I'll remove this.

core/modules/block/block.admin.inc
bjoern-st: While testing I sometimes had the problem, that $path here was something like en/foo/bar .. with a language prefix.
This ended up in a resulting url like http://www.example.com/en/en/foo/bar - so I had to remove the language prefix from the $path to receive a working path value

core/modules/block/block.admin.inc
bjoern-st: I have changed that.

core/modules/block/block.admin.inc
bjoern-st: Actually this was on my list, but ups ..

core/modules/block/block.module
bjoern-st: Changed to language

PR #2096, opened 2018-02-08
Drupal 7.56 cross-ports

quicksketch:
Fixes backdrop/backdrop-issues#2776

Replaces https://github.com/backdrop/backdrop/pull/2090.


quicksketch:
We can use https://github.com/backdrop/backdrop/pull/2090 after all. Tests are now running over there.

PR #2072, opened 2018-01-16
2879 expand dropdown

jenlampton:
Fixes https://github.com/backdrop/backdrop-issues/issues/2879
Replaces https://github.com/backdrop/backdrop/pull/2029

PR #2103, opened 2018-02-20
Issue #2072: Extend link dialogue to include a file upload option.

Graham-72:
This is ~~by way of a 'proof of concept'~~ now a working proposal and includes changes that
- use code based on that used for uploading and linking to images for selecting, uploading and linking to files within WYSIWYG.
- provide autocomplete for managed files previously uploaded.
- extend the WYSIWYG (CKEditor) admin to configure file uploads similarly to image uploads.
- add a user permission for allowing file uploads.
- adds script to enable the text of the link to be edited.

Updated 22 March 2018

### DIFF
--- a/core/modules/ckeditor/ckeditor.admin.inc
+++ b/core/modules/ckeditor/ckeditor.admin.inc
@@ -39,6 +39,17 @@ function ckeditor_settings_form(&$form, $form_state, $format) {
     '#parents' => array('editor_settings', 'image_upload'),
   );
 
+  $elements['plugins']['file'] = filter_editor_file_upload_settings_form($format);
+  $elements['plugins']['file'] += array(
+    '#type' => 'fieldset',
+    '#title' => t('File uploading'),
+    '#attributes' => array(
+      'class' => array('ckeditor-plugin-backdropimage'),
+      'data-ckeditor-feature-dependency' => 'BackdropImage',
+    ),
+    '#parents' => array('editor_settings', 'file_upload'),
+  );
+
   $elements['plugins']['style'] = array(
     '#type' => 'fieldset',
     '#title' => t('Style list'),
--- a/core/modules/ckeditor/ckeditor.admin.inc
+++ b/core/modules/ckeditor/ckeditor.admin.inc
@@ -44,8 +44,8 @@ function ckeditor_settings_form(&$form, $form_state, $format) {
     '#type' => 'fieldset',
     '#title' => t('File uploading'),
     '#attributes' => array(
-      'class' => array('ckeditor-plugin-backdropimage'),
-      'data-ckeditor-feature-dependency' => 'BackdropImage',
+      'class' => array('ckeditor-plugin-backdroplink'),
+      'data-ckeditor-feature-dependency' => 'BackdropLink',
     ),
     '#parents' => array('editor_settings', 'file_upload'),
   );
--- a/core/modules/ckeditor/js/plugins/backdroplink/plugin.js
+++ b/core/modules/ckeditor/js/plugins/backdroplink/plugin.js
@@ -70,18 +70,21 @@ CKEDITOR.plugins.add('backdroplink', {
       exec: function (editor) {
         var backdropImageUtils = CKEDITOR.plugins.backdropimage;
         var focusedImageWidget = backdropImageUtils && backdropImageUtils.getFocusedWidget(editor);
+        // Get currently selected text.
         var selectedText = editor.getSelection().getSelectedText();
+        // Get a currently selected link as CKEDITOR.dom.element
         var linkElement = getSelectedLink(editor);
         if (linkElement && (selectedText == "")) {
-          selectedText = linkElement.$.firstChild.data;
+          // Get the text if cursor is somewhere in an existing link.
+          selectedText = linkElement.$.text;
         }
 
-        // Set existing values based on selected element.
+        // Set existing attribute values based on selected element.
         var existingValues = {};
         if (linkElement && linkElement.$) {
           existingValues = parseAttributes(editor, linkElement);
 
-          // Override the displayed link text.
+          // Update the displayed link text
           existingValues.linktext = selectedText;
         }
         // Or, if an image widget is focused, we're editing a link wrapping
@@ -97,12 +100,10 @@ CKEDITOR.plugins.add('backdroplink', {
         // Prepare a save callback to be used upon saving the dialog.
         var saveCallback = function (returnValues) {
           editor.fire('saveSnapshot');
-
           // Ignore a disabled target attribute.
           if (returnValues.attributes.target === 0) {
             delete returnValues.attributes.target;
           }
-
           // If an image widget is focused, we're not editing an independent
           // link, but we're wrapping an image widget in a link.
           if (focusedImageWidget) {
@@ -110,56 +111,48 @@ CKEDITOR.plugins.add('backdroplink', {
             editor.fire('saveSnapshot');
             return;
           }
+          // If not an image, replace text of link with new text.
           else {
-            // If not an image, replace text of link with new text.
+            // Get the current selection.
             var selection = editor.getSelection();
+            // Get the text of the current selection.
             var oldtext = selection.getSelectedText();
+            // And the replacement text.
             var newtext = returnValues.attributes.linktext;
+            // Get the range of the selection.
             var range = selection.getRanges(1)[0];
 
-            // If no text is selected within a link
-            // replace the link with the new text.
-            if (newtext && (oldtext == "")){
-              // Enlarge the range to include the whole link.
-              range.enlarge(CKEDITOR.ENLARGE_BLOCK_CONTENTS);
-
-              // Create a new node to insert at range.
-              var text = new CKEDITOR.dom.text( newtext );
-
-              // Replace current link with new text.
-              range.deleteContents();
-              range.collapse(true);
-              range.insertNode(text);
-              range.selectNodeContents(text);
-
-              // Make new text into a link.
-              var style = new CKEDITOR.style({element: 'a', attributes: returnValues.attributes});
-              style.type = CKEDITOR.STYLE_INLINE;
-              style.applyToRange(range);
-            }
-            // If new text is different, replace the item specified by range.
-            else if (newtext && (newtext != oldtext)) {
-              // Create a new node to insert at range.
-              var text = new CKEDITOR.dom.text( newtext );
-              // Delete current contents at range.
-              range.deleteContents();
-              range.insertNode(text);
-              range.selectNodeContents(text);
+            // If the selection is a link, replace the text
+            var element = selection.getStartElement();
+            this.element = element;
+            if(element.$.nodeName === "A") {
+              this.element.setText(newtext);
             }
           }
 
           // Create a new link element if needed.
           if (!linkElement && returnValues.attributes.href) {
-
             // Use link URL as text with a collapsed cursor.
             if (range.collapsed) {
-              // Shorten mailto URLs to just the email address.
-              var text = new CKEDITOR.dom.text(returnValues.attributes.href.replace(/^mailto:/, ''), editor.document);
+
+              if(newtext != ""){
+                var text = new CKEDITOR.dom.text( newtext );
+              }
+              else {
+                // Use href as link text
+                // Shorten mailto URLs to just the email address.
+                var text = new CKEDITOR.dom.text(returnValues.attributes.href.replace(/^mailto:/, ''), editor.document);
+              }
               range.insertNode(text);
               range.selectNodeContents(text);
             }
 
             // Create the new link by applying a style to the new text.
+            // Remove attributes that are not required.
+            if ( !(returnValues.attributes['data-file-id'] > 0)){
+              delete returnValues.attributes['data-file-id'];
+            }
+            delete returnValues.attributes.linktext;
             var style = new CKEDITOR.style({element: 'a', attributes: returnValues.attributes});
             style.type = CKEDITOR.STYLE_INLINE;
             style.applyToRange(range);
@@ -174,13 +167,17 @@ CKEDITOR.plugins.add('backdroplink', {
             range.setEndAfter(linkElement);
             range.select();
           }
-          // Update the link properties.
+          // Update the link properties and remove redundant items.
           else if (linkElement) {
             for (var attrName in returnValues.attributes) {
               if (returnValues.attributes.hasOwnProperty(attrName)) {
+                var value = returnValues.attributes[attrName];
+                // Remove attribute data-file-id if 0 or null.
+                if ((attrName == "data-file-id") && !(value > 0)){
+                  linkElement.removeAttribute(attrName);
+                }
                 // Update the property if a value is specified.
-                if (returnValues.attributes[attrName].length > 0) {
-                  var value = returnValues.attributes[attrName];
+                else if ((returnValues.attributes[attrName].length > 0) && (attrName !== "linktext")) {
                   linkElement.data('cke-saved-' + attrName, value);
                   linkElement.setAttribute(attrName, value);
                 }
@@ -309,16 +306,24 @@ CKEDITOR.plugins.add('backdroplink', {
  * @param {CKEDITOR.editor} editor
  */
 function getSelectedLink(editor) {
+  // Get the current selection
   var selection = editor.getSelection();
+  // Get the currently selected element.
   var selectedElement = selection.getSelectedElement();
+  // If a selected element exists and is 'a'
   if (selectedElement && selectedElement.is('a')) {
     return selectedElement;
   }
-
+  // Get the range of the current selection
   var range = selection.getRanges(true)[0];
-
+  // If it has a value, decreases the range to make sure that boundaries
+  // always anchor beside text nodes or the innermost element.
   if (range) {
     range.shrink(CKEDITOR.SHRINK_TEXT);
+    // Find the node which fully contains the range.
+    // Return an element path for the selection in the editor.
+    // elementPath.contains = Search the path elements that meets the specified criteria.
+    // getCommonAncestor = Find the node which fully contains the range.
     return editor.elementPath(range.getCommonAncestor()).contains('a', 1);
   }
   return null;
--- a/core/modules/ckeditor/js/plugins/backdroplink/plugin.js
+++ b/core/modules/ckeditor/js/plugins/backdroplink/plugin.js
@@ -85,7 +85,7 @@ CKEDITOR.plugins.add('backdroplink', {
           existingValues = parseAttributes(editor, linkElement);
 
           // Update the displayed link text
-          existingValues.linktext = selectedText;
+          existingValues.text = selectedText;
         }
         // Or, if an image widget is focused, we're editing a link wrapping
         // an image widget.
@@ -94,7 +94,7 @@ CKEDITOR.plugins.add('backdroplink', {
         }
         // Or, if selected element is not an existing link or an image.
         else {
-          existingValues.linktext = selectedText;
+          existingValues.text = selectedText;
         }
 
         // Prepare a save callback to be used upon saving the dialog.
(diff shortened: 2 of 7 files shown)


herbdool:
@Graham-72 I've tested and overall it looks good. Though I've got a few suggestions.

Tested:

- [x] can upload and link
- [x] can highlight text and link to an uploaded file
- [x] files show up in Manage files page with usage
- [x] permissions work -- created a regular authenticated user and they can't upload files unless given the permission
- [x] unchecking in CKeditor settings removes ability to upload files
- [x] autocomplete works

Some suggestions:
- [x] Consider making it default that authenticated users can upload inline files. I believe this is the default for inline images.
- [x] Default filepath could make it more obvious that this is for inline files. Make it "/inline-files/" to match images.
- [x] There seem to be a few unrelated commits--perhaps because of the way you merged in upstream. You'll need to clean up this PR before it can be merged.
- [ ] We might need tests written before it can be merged in. There might be ones for inline images you could use for inspiration.

Graham-72:
@herbdool thank you very much for looking at this and for your suggestions. I doubt that I have the technical know-how to take this any further myself, for example by adding tests as this is something I have never yet done. Hopefully someone else can do this.

herbdool:
@Graham-72 let's where I can help. In terms of cleaning up the commits, this might work:

```
git fetch upstream 1.x
git pull --rebase upstream 1.x
git push -f Graham-72 issue-2072rev1
```
This depends on what you've named the upstream (backdrop/backdrop) compared to your own fork.

For the authenticated users getting the permission by default, you can add another update hook to filter.install like this one:

```
/**
 * Add a separate permission for image upload access.
 */
function filter_update_1001() {
  user_role_grant_permissions(BACKDROP_AUTHENTICATED_ROLE, array('upload editor images'));

  // Enable uploading on existing text formats. Access previously was not
  // checked previously, so all formats had uploading enabled.
  $format_configs = config_get_names_with_prefix('filter.format.');

  foreach ($format_configs as $config_name) {
    $config = config($config_name);
    if ($config->get('editor')) {
      $editor_settings = (array) $config->get('editor_settings');
      $upload_settings = isset($editor_settings['image_upload']) ? $editor_settings['image_upload'] : array();
      $upload_settings['status'] = 1;
      $upload_settings += array(
        'dimensions' => array('max_width' => '', 'max_height' => ''),
        'max_size' => NULL,
        'scheme' => 'public',
        'directory' => 'inline-images',
      );
      $editor_settings['image_upload'] = $upload_settings;
      $config->set('editor_settings', $editor_settings);
      $config->save();
    }
  }
}
```

In terms of changing the default directory, that will depend on what other people think, though I do think `inline-files` will help differentiate it better from other uploads (since every file is "uploaded" from any file field as well).

I took a look for existing tests for inline images and I didn't see any. So maybe skip that for now unless Nate brings it up.

Graham-72:
@herbdool thank you for your helpful suggestions. I have managed to rebase my PR and have now added an update hook that I hope will do what is needed for existing sites.

core/modules/filter/filter.install
```diff
@@ -85,6 +85,34 @@ function filter_update_1001() {
...
+
+  // Enable uploading on existing text formats for which
+  // image uploading is already enabled.
+  $format_configs = config_get_names_with_prefix('filter.format.');
+
+  foreach ($format_configs as $config_name) {
+    $config = config($config_name);
+    if ($config->get('editor')) {
+      $editor_settings = (array) $config->get('editor_settings');
+      if (isset($editor_settings['image_upload'])){
```
herbdool: I think this needs to be 'file_upload' and not 'image_uload'.

herbdool:
@Graham-72 I think it's pretty close. I see the permission in the test site -- I haven't tested by uninstalling and re-installing but the code is mostly good. I made one comment on a small mistake.

Graham-72:
> I made one comment on a small mistake.

@herbdool I cannot find the comment you made 6 days ago

herbdool:
@Graham-72 now you should see it. I didn't realize that comments on code weren't published until I clicked to finish the code review.

Basically, on line 102 you have: `if (isset($editor_settings['image_upload'])){` but it should probably be `file_upload`.

core/modules/filter/filter.install
Graham-72: @herbdool The purpose here is to find out if any user is already already set to use 'image upload' for one or more of the formats and if so to set their 'file upload' setting as well for that format.

herbdool: I see. What's the purpose for that?

Graham-72: I was responding, slightly differently, to your suggestion
>For the authenticated users getting the permission by default, you can add another update hook to filter.install like this one:

So, any existing formats will get the file upload permission where they already have it for image upload.

herbdool:
@graham-72 you should either output a message to the person updating so they know why not all formats don't have the permission or don't be selective. I think the end user needs to know why.

Graham-72:
This latest commit https://github.com/backdrop/backdrop/pull/2103/commits/2828b8da24f0da021572f862d022bf724abda321 extends the CKEditor Backdroplink plugin script so that the link text is displayed in the UI and can be edited.

It also fixes a problem whereby URIs other than file addresses were not being handled correctly in previous commits.

Graham-72:
@herbdool 
re:
> you should either output a message to the person updating so they know why not all formats don't have the permission or don't be selective. I think the end user needs to know why.'

I have been thinking more about this and have trialled outputting a message but I am doubtful about including this because the message will appear when updating core and not just the filter module and so may be obscured by everything else that is happening.

On balance I think it would be better not to include an update to set any permissions and leave it to the site admin to realise that they must use the text editor configuration to set the 'file upload' permission where needed. So I suggest that commit https://github.com/backdrop/backdrop/pull/2103/commits/e530d9497e4b4bdc61a15f446e18440dda47a4e0 is not included.

Graham-72:
This latest commit https://github.com/backdrop/backdrop/pull/2103/commits/2828b8da24f0da021572f862d022bf724abda321 extends the CKEditor Backdroplink plugin script so that the link text is displayed in the UI and can be edited.

It also fixes a problem whereby URIs other than file addresses were not being handled correctly in previous commits.

Graham-72:
Closed and reopening in order to get new sandbox.

herbdool:
>On balance I think it would be better not to include an update to set any permissions and leave it to the site admin to realise that they must use the text editor configuration to set the 'file upload' permission where needed. So I suggest that commit e530d94 is not included.

Might be useful to get a third opinion. @jenlampton As part of this PR should it give everyone permissions to add files in the rich text editor by default (like the image embedding); or don't assign permissions to anyone by default; or be selective (e.g. just those who already have image embedding permissions)? And whether to show a message or not?

jenlampton:
Hm. Taking a step back for a moment...  why does this need a separate permission? Is there a use-case for limiting who can link to files (different from who can upload images)? I don't know, as I've never needed this feature.

Graham-72:
I think we need to provide the permission for uploading files in order that the site owner has full control of who can add things to the server.

jenlampton:
Right, but does that need to be a different permission from the one we use for image embedding? Is there a use case for someone who CAN link files, but who CANNOT embed images, or vice versa?

herbdool:
I think there is for vice versa. Someone might want to restrict just to uploading images but not other files.

Regardless, the name of the embed images permission makes it too restrictive if it's supposed to cover both. How hard is it too rename it?

jenlampton:
> How hard is it to rename it?

Not hard, we'd just need to add an update.

core/modules/ckeditor/ckeditor.admin.inc
```diff
@@ -39,6 +39,17 @@ function ckeditor_settings_form(&$form, $form_state, $format) {
     '#parents' => array('editor_settings', 'image_upload'),
   );
 
+  $elements['plugins']['file'] = filter_editor_file_upload_settings_form($format);
```
jenlampton, line 42: this looks good, let's wrap it in an access check for our new permission :)

core/modules/filter/filter.pages.inc
```diff
@@ -15,25 +15,33 @@
  */
 function filter_tips_long($format = NULL) {
   if (!empty($format)) {
-    $output = theme('filter_tips', array('tips' => _filter_tips($format->format, TRUE), 'long' => TRUE));
+    $output = theme('filter_tips', array(
```
jenlampton, line 18: nice improvement here.

core/modules/ckeditor/ckeditor.admin.inc
```diff
@@ -39,6 +39,17 @@ function ckeditor_settings_form(&$form, $form_state, $format) {
...
+  $elements['plugins']['file'] += array(
+    '#type' => 'fieldset',
+    '#title' => t('File uploading'),
+    '#attributes' => array(
+      'class' => array('ckeditor-plugin-backdropimage'),
+      'data-ckeditor-feature-dependency' => 'BackdropImage',
```
quicksketch: `backdropimage` should be `backdropfile` and `BackdropImage` should be `BackdropLink` here. That will make these options only appear when the "Link" button is dragged in. Right now the File uploading options disappear if the "Image" button is removed on the editor settings page.

core/modules/ckeditor/js/plugins/backdroplink/plugin.js
```diff
@@ -128,13 +167,17 @@ CKEDITOR.plugins.add('backdroplink', {
...
             range.select();
           }
-          // Update the link properties.
+          // Update the link properties and remove redundant items.
           else if (linkElement) {
             for (var attrName in returnValues.attributes) {
               if (returnValues.attributes.hasOwnProperty(attrName)) {
+                var value = returnValues.attributes[attrName];
+                // Remove attribute data-file-id if 0 or null.
+                if ((attrName == "data-file-id") && !(value > 0)){
```
quicksketch, line 176: `!(value > 0)` is a confusing comparison. Could this be simplified to `!value`?

core/modules/ckeditor/js/plugins/backdroplink/plugin.js
```diff
@@ -70,50 +70,89 @@ CKEDITOR.plugins.add('backdroplink', {
...
+            var oldtext = selection.getSelectedText();
+            // And the replacement text.
+            var newtext = returnValues.attributes.linktext;
+            // Get the range of the selection.
             var range = selection.getRanges(1)[0];
 
+            // If the selection is a link, replace the text
+            var element = selection.getStartElement();
+            this.element = element;
+            if(element.$.nodeName === "A") {
```
quicksketch: I think as in other places, you might be able to use `element.is('a')` here instead of using the raw DOM object from `$`.

core/modules/ckeditor/js/plugins/backdroplink/plugin.js
```diff
@@ -263,16 +306,24 @@ CKEDITOR.plugins.add('backdroplink', {
...
   var range = selection.getRanges(true)[0];
-
+  // If it has a value, decreases the range to make sure that boundaries
+  // always anchor beside text nodes or the innermost element.
   if (range) {
     range.shrink(CKEDITOR.SHRINK_TEXT);
+    // Find the node which fully contains the range.
+    // Return an element path for the selection in the editor.
+    // elementPath.contains = Search the path elements that meets the specified criteria.
+    // getCommonAncestor = Find the node which fully contains the range.
```
quicksketch, line 326: Thank you for adding all of these helpful comments.

core/modules/filter/filter.pages.inc
```diff
@@ -50,15 +58,67 @@ function filter_format_editor_link_form($form, &$form_state, $format) {
...
-    '#title' => t('URL'),
+    '#title' => t('Link value'),
     '#type' => 'textfield',
     '#element_validate' => array('_filter_format_editor_link_url_validate'),
-    '#default_value' => isset($values['href']) ? $values['href'] : FALSE,
+    '#placeholder' => ' ',
+    '#default_value' => isset($values['href']) ? $values['href'] : NULL,
     '#parents' => array('attributes', 'href'),
+    '#wrapper_attributes' => array(
+      'data-editor-image-toggle' => t('Link to existing content'),
```
quicksketch: I can see why you re-used these existing data attributes. We should rename them so they have a generic purpose instead of being namespaced under "image".

core/modules/filter/filter.admin.inc
```diff
@@ -617,3 +617,104 @@ function filter_editor_image_upload_settings_form($format) {
...
+ *
+ * @ingroup forms
+ */
+function filter_editor_file_upload_settings_form($format) {
+  // Defaults.
+  $settings = isset($format->editor_settings['file_upload']) ? $format->editor_settings['file_upload'] : array();
+  $settings += array(
+    'status' => FALSE,
+    'scheme' => file_default_scheme(),
+    'directory' => 'uploads',
```
quicksketch: I wonder if this is too generic. I'm tempted to use "inline-files" to match "inline-images", but I am ambivalent on that.

core/modules/filter/filter.admin.inc
```diff
@@ -617,3 +617,104 @@ function filter_editor_image_upload_settings_form($format) {
...
+    'max_size' => '',
+    'file_extensions' => 'txt pdf',
```
quicksketch: File upload fields default to just "txt" and make you specify all extensions manually. I see the obvious use-case of allowing pdf files by default here though. It's also a likely file type for adding links to, as opposed to file fields that might be used for anything. So this default seems fine to me.

core/modules/filter/filter.admin.inc
```diff
@@ -617,3 +617,104 @@ function filter_editor_image_upload_settings_form($format) {
...
+      '#type' => 'radios',
+      '#title' => t('File storage'),
+      '#default_value' => $settings['scheme'],
+      '#options' => $options,
+      '#states' => $show_if_file_uploads_enabled,
+      '#access' => count($options) > 1,
+    );
+  }
+  // Set data- attributes with human-readable names for all possible stream
+  // wrappers, so that drupal.ckeditor.drupalimage.admin's summary rendering
```
quicksketch, line 680: Again, probably best if we could namespace the image-related functionality to a generic one.

core/modules/filter/filter.admin.inc
```diff
@@ -617,3 +617,104 @@ function filter_editor_image_upload_settings_form($format) {
...
+    '#type' => 'textfield',
+    '#title' => t('Allowed file extensions'),
+    '#default_value' => $settings['file_extensions'],
+    '#description' => t('Separate extensions with a space or comma and do not include the leading dot.'),
+    '#element_validate' => array('_file_generic_settings_extensions'),
+    '#weight' => 1,
+    '#maxlength' => 256,
+    // By making this field required, we prevent a potential security issue
+    // that would allow files of any type to be uploaded.
+    '#required' => TRUE,
```
quicksketch, line 716: :+1:

core/modules/filter/filter.pages.inc
```diff
@@ -50,15 +58,67 @@ function filter_format_editor_link_form($form, &$form_state, $format) {
...
   );
+
+
+  // Construct strings to use in the upload validators.
+  $upload_settings = isset($format->editor_settings['file_upload']) ? $format->editor_settings['file_upload'] : array();
+
+  $max_filesize = !empty($upload_settings['max_size']) ? min(parse_size($upload_settings['max_size']), file_upload_max_size()) : file_upload_max_size();
+  $file_extensions = !empty($upload_settings['file_extensions']) ? $upload_settings['file_extensions'] : 'txt';
+  $file_directory = !empty($upload_settings['directory']) ? $upload_settings['directory'] : 'uploads';
+  $existing_file = !empty($values['data-file-id']) ? file_load($values['data-file-id']) : NULL;
```
quicksketch: All of these `!empty()` checks can be a bit hard to interpret. If you're just worried about undefined values, we can just do the following:

```
$upload_settings += array(
  'max_size' => file_upload_max_size(),
  'file_extensions' => 'txt',
  'directory' => 'uploads',
  'data-file-id' => NULL,
);
```

Then you can remove at least the first 3 ternary operators in these 4 lines.

quicksketch (changes requested):
There's only one thing that absolutely needs to be fixed here, and that's the hiding/showing of File upload options on the editor configuration form. They should hide/show based on the "Link" button, not the "Image" button. Fixes to the namespacing of the Image-related classes and attributes would be appreciated, but we can do those in a follow-up.

core/modules/ckeditor/js/plugins/backdroplink/plugin.js
```diff
@@ -70,50 +70,89 @@ CKEDITOR.plugins.add('backdroplink', {
...
+
+          // Create a new link element if needed.
+          if (!linkElement && returnValues.attributes.href) {
             // Use link URL as text with a collapsed cursor.
             if (range.collapsed) {
-              // Shorten mailto URLs to just the email address.
-              var text = new CKEDITOR.dom.text(returnValues.attributes.href.replace(/^mailto:/, ''), editor.document);
+
+              if(newtext != ""){
+                var text = new CKEDITOR.dom.text( newtext );
```
opi: minor coding standard here, remove space in parenthesis

core/modules/filter/filter.admin.inc
herbdool: Oh, I thought it was changed to inline-files at one point.

core/modules/ckeditor/ckeditor.admin.inc
Graham-72: Thanks, understood (now!).

core/modules/ckeditor/js/plugins/backdroplink/plugin.js
Graham-72: Yes, thanks.

core/modules/ckeditor/js/plugins/backdroplink/plugin.js
Graham-72: As a newbie to CKEditor scripts I really struggled to understand what was going on - and still do!

core/modules/filter/filter.admin.inc
Graham-72: Yes, agreed.

core/modules/filter/filter.pages.inc
Graham-72: I don't feel confident enough to make this change myself.

core/modules/filter/filter.pages.inc
Graham-72: Thanks, done.

core/modules/ckeditor/js/plugins/backdroplink/plugin.js
Graham-72: I need to get my head around this one, again!

core/modules/ckeditor/js/plugins/backdroplink/plugin.js
Graham-72: I will try this in a test. I'm not at all confident with javascript!

core/modules/filter/filter.pages.inc
herbdool: @Graham-72 this is a good learning process. Since your work gets reviewed anyway, it's okay to make mistakes!

Graham-72: Yes. thanks for your encouragement but I need to understand the terminology e.g. 'being namespaced under "image"'. 😃

quicksketch (approved):

quicksketch:
Merged into 1.x for 1.10. Thanks!

core/modules/filter/filter.pages.inc
jenlampton: What he meant was "We should rename them, so that people don't think they are only used for toggling an image". 

I had assumed that's what the data attribute was for, so I had removed the data attribute in https://github.com/backdrop/backdrop/pull/2638 since toggling an image there didn't make any sense.

I can tell from this conversation that this data attribute is important, but I still have no idea why. I can't see any change in behavior with/without it on the `href` field.

Graham-72: As I understand it the attribute data-editor-image-toggle is required for communicating from `filter.pages.inc` to the script `filter.js`. It gives the text to be used in the toggle action. Yes, it should have been renamed when extending the use of the toggle action to include inserting links as well as images.

jenlampton:
Ah!

core/modules/ckeditor/js/plugins/backdroplink/plugin.js
klonos: As a general (and perpetual) newbie, I appreciate the detailed documentation. Oftentimes I need it to remember what my own code does 😅

PR #2106, opened 2018-03-03
Issue #2987: Check for valid file admin config before attempting to u…

jenlampton:
Fixes https://github.com/backdrop/backdrop-issues/issues/2987


quicksketch:
Merged into 1.x and 1.9.x. Thanks!

PR #2017, opened 2017-09-21
Issue #2858: Print the update errors collected to the screen.

jenlampton:
Fixes https://github.com/backdrop/backdrop-issues/issues/2858

PR #2107, opened 2018-03-07
Issue #2858: Print the update errors collected to the screen.

opi:
Follow-up https://github.com/backdrop/backdrop/pull/2017 with some concern addressed.

### DIFF
--- a/core/update.php
+++ b/core/update.php
@@ -79,6 +79,7 @@ function update_script_selection_form($form, &$form_state) {
   $updates = update_get_update_list();
   $starting_updates = array();
   $incompatible_updates_exist = FALSE;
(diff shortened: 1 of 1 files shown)


core/update.php
```diff
@@ -89,6 +90,7 @@ function update_script_selection_form($form, &$form_state) {
         '#suffix' => '</div>',
       );
       $incompatible_updates_exist = TRUE;
+      $incompatible_updates_messages[] = $update;
```
opi, line 93: Should we increment `$incompatible_count` here too ? I'm a little bit lost with the whole logic.

jenlampton:
Merged these changes into https://github.com/backdrop/backdrop/pull/2017 manually.

PR #2116, opened 2018-03-29
#3010: Empty elements being inserted in head

laryn:


core/includes/theme.inc
```diff
@@ -2446,7 +2446,9 @@ function theme_head_tag($variables) {
   $element = $variables['element'];
   $attributes = isset($element['#attributes']) ? backdrop_attributes($element['#attributes']) : '';
   if (!isset($element['#value'])) {
-    return '<' . $element['#tag'] . $attributes . " />\n";
+    if (isset($element['#tag'])) {
```
opi: nice catch, but can be single ligned with the upper condition

opi (approved):

core/includes/theme.inc
opi: moreover, using `!empty()` instead of `isset()` can prevent empty value of `#tag`.

laryn: @opi That's what I did at first, but then it gets sent into the `else` section below and outputs other empty elements -- `<><>` instead of `< />< />`

laryn: `!empty` sounds like a good idea then.

core/includes/theme.inc
```diff
@@ -2446,7 +2446,9 @@ function theme_head_tag($variables) {
   $element = $variables['element'];
   $attributes = isset($element['#attributes']) ? backdrop_attributes($element['#attributes']) : '';
   if (!isset($element['#value'])) {
-    return '<' . $element['#tag'] . $attributes . " />\n";
+    if (!empty($element['#tag'])) {
+      return '<' . $element['#tag'] . $attributes . " />\n";
+    }
   }
   else {
     $output = '<' . $element['#tag'] . $attributes . '>';
```
opi, line 2454: #tag key is also required here ; This key is not mentionned as optional in [the documentation](https://api.backdropcms.org/api/backdrop/core%21includes%21theme.inc/function/theme_head_tag/1)

I see 2 different ways to *fix* this : 
- add a check at the beginning of the theme function to return an empty string if #tag is not defined.
- The notice is already displayed + this parameter is required in documentation + an empty html element is not that bad (even if dirty) = the issue belong to the contrib module ;)

PR #2124, opened 2018-04-10
Issue #2542 Merging custom colors into the default color array, so th…

hosef:
…at when you add a new color to the theme, it will show up in the UI.


quicksketch:
Closing/reopening to re-run tests.

quicksketch:
An array_merge() also would have worked here, but the fix is good as written. Merged into 1.x and 1.9.x. Thanks!

quicksketch:
Nevermind, you explained the reason in the issue:

> I went with a slightly more complicated change than in the original patch because simply merging the arrays would cause new items to show up at the bottom of the list and stay there forever. using a foreach, and overwriting keys puts the existing custom colors in the correct place in the list.

Thanks!

PR #2126, opened 2018-04-13
Issue #496: Create database during installation if possible.

quicksketch:
Fixes https://github.com/backdrop/backdrop-issues/issues/496.

As this deals with the installer, this is not testable via SimpleTest. Some manual review will be required.

### DIFF
--- a/core/includes/database/mysql/database.inc
+++ b/core/includes/database/mysql/database.inc
@@ -11,7 +11,6 @@
  */
 
 class DatabaseConnection_mysql extends DatabaseConnection {
-
   /**
    * Flag to indicate if the cleanup function in __destruct() should run.
    *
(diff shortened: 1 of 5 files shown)


core/includes/database/mysql/database.inc
```diff
@@ -127,6 +128,25 @@ class DatabaseConnection_mysql extends DatabaseConnection {
...
+   *
+   * @param string $database
+   *   The name of the database to create.
+   *
+   * @throws DatabaseNotFoundException
+   */
+  public function createDatabase($database) {
+    try {
+      // Create the database and set it as active.
+      $this->exec("CREATE DATABASE $database");
```
serundeputy: @quicksketch should we put a `IF NOT EXISTS` on this?
https://dev.mysql.com/doc/refman/5.7/en/create-database.html

quicksketch: I don't think adding it would do any harm. Though I don't think we'd have any situation where that should occur.

quicksketch: I've updated the PR to include that.

serundeputy:
tested w/ mariadb and mysql; looks good works good.

PR #2128, opened 2018-04-13
Issue #2973: Module page search should work on category names.

jenlampton:
Fixes https://github.com/backdrop/backdrop-issues/issues/2973

### DIFF
--- a/core/modules/system/js/modules.js
+++ b/core/modules/system/js/modules.js
@@ -10,7 +10,7 @@ Backdrop.behaviors.moduleFilterByText = {
   attach: function(context, settings) {
     var $input = $('input.table-filter-text').once('table-filter-text');
     var $form = $('#system-modules');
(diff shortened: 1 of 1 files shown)


jenlampton:
new sandbox please.

quicksketch (dismissed):
This looks inefficient. With this approach we are doing the following:

- Showing all fieldsets.
- Hiding all fieldsets except those that contain module string matches.
- Then looping through fieldsets and showing them again if the label matches.

Instead of doing a 3rd step, we should simply not hide the fieldset if its fieldset label matches in the second step.

jenlampton:
> This looks inefficient. 

I was afraid you might say that... I'll give it another shot!

quicksketch:
I pushed an updated approach that cuts down on the show/hide/show approach and now just hides the elements once based on the current matching.

quicksketch:
New sandbox.

PR #2132, opened 2018-04-14
Color module support in basis

hosef:
Issue #3028: Adds support for the Color module to Basis, cleans up the color codes so that they are consistent, and reduces the number of unique colors in the theme.


core/themes/basis/color/color.inc
```diff
@@ -0,0 +1,72 @@
...
+    'header' => t('Header background'), // And active pager link.
+    'sitename' => t('Site name background'),
+    'slogan' => t('Slogan background'),
+    'titleslogan' => t('Site name and slogan'), // And main menu links.
+    'bg' => t('Main background'), // And submenu text, pager text, and drop down arrows
+    'footer' => t('Footer background'),
+    'text' => t('Text color'),
+    'link' => t('Link color'),
+    'hovermenu' => t('Hover menu link background'),
+    'border_thingy' => t('Border thingies'), //These are all over the place.
```
wesruv: Lol, how about "Section Divider Color" ?

core/themes/basis/color/color.inc
```diff
@@ -0,0 +1,72 @@
...
+    'form_focus_border' => t('Form element borders when selected'),
```
wesruv: I'm used to seeing the word "active", form "active" state color, or something?

But if you folks think this is more clear, I think what's there is good, just thought I'd throw in 2 cents.

core/themes/basis/color/color.inc
```diff
@@ -0,0 +1,72 @@
...
+    'action_item_background' => t('Action item background'),
```
wesruv: What's this mean? I don't understand it and I know the theme well...

wesruv (commented):
Verbiage comments, I checked out the color adjustments using styleguide module and navigating the stuff that's there and I think they're fine, especially considering what we're gaining.

Thanks for this update! I got such a headache trying to implement this :/ glad it's getting in!

core/themes/basis/color/color.inc
wesruv: Section border color?

hosef: I guess that would work. the thing that threw me off was that it's also used in the vertical tabs and form element borders on edit forms.
borders

core/themes/basis/color/color.inc
hosef: The action items are the 'view', 'edit', etc. links at the top pf pages.
action_items

wesruv: I'd just call it Tab background

PR #2133, opened 2018-04-15
Issue #3034: Add button-primary class to system_settings_form() submi…

jenlampton:
Fixes https://github.com/backdrop/backdrop-issues/issues/3034


jenlampton:
not needed.

PR #2138, opened 2018-04-21
Issue #3040: Show a theme screenshot in the installer.

quicksketch:
Replaces https://github.com/backdrop/backdrop/pull/2135. Just rerolling for new sandbox.

### DIFF
--- a/core/modules/installer/installer.theme.inc
+++ b/core/modules/installer/installer.theme.inc
@@ -127,6 +127,15 @@ function template_preprocess_installer_browser_project_dialog(&$variables) {
   $variables['project_page'] = l(t('View on BackdropCMS.org'), 'https://backdropcms.org/project/' . $project['name']);
   $variables['description'] = _filter_htmlcorrector(filter_xss($project['description'], array('h2', 'p', 'a', 'ul', 'li', 'pre')));
   $variables['title'] = $project['title'];
+  $variables['image'] = FALSE;
+
+  // If the project has an image, render it.
+  if (!empty($project['image'])) {
+    $variables['image'] = theme('image', array(
+      'path' => $project['image'],
+      'alt' => $project['title'],
+    ));
+  }
 }
 
 /**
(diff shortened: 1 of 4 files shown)


core/modules/installer/installer.theme.inc
```diff
@@ -127,6 +127,15 @@ function template_preprocess_installer_browser_project_dialog(&$variables) {
   $variables['project_page'] = l(t('View on BackdropCMS.org'), 'https://backdropcms.org/project/' . $project['name']);
   $variables['description'] = _filter_htmlcorrector(filter_xss($project['description'], array('h2', 'p', 'a', 'ul', 'li', 'pre')));
   $variables['title'] = $project['title'];
+  $variables['image'] = FALSE;
+
+  // If the project has an image, render it.
+  if (!empty($project['image'])) {
+    $variables['image'] = theme('image', array(
+      'path' => $project['image'],
+      'alt' => t($project['title']),
```
opi: Not sure that project title must go through `t()`

core/modules/installer/installer.theme.inc
```diff
@@ -140,18 +149,38 @@ function template_preprocess_installer_browser_project(&$variables) {
...
+  );
+
+  // If the project has an image, render it.
+  $variables['image'] = FALSE;
+
+  if ($project['type'] == 'theme') {
+    if (!empty($project['image'])) {
+      $image = theme('image', array(
+        'path' => $project['image'],
+        'alt' => t($project['title']),
```
opi: same here (use of `t()`)

opi (commented):
Quick review on my phone

core/modules/installer/installer.theme.inc
docwilmot: More so:
>You should never use t() to translate variables, such as calling t($text);

From Backdrop API docs for `t()`

quicksketch: Right you are. Especially in a situation like this where the project title is almost certainly not going to be in the site's existing translation list (nor should it be). And project titles typically aren't translated anyway, as we treat them as proper nouns in most cases.

Since I now "own" this PR, I've made these fixes and repushed.

PR #1262, opened 2016-02-20
Issue #857: Add schedule publication of nodes functionality.

serundeputy:
This fixes https://github.com/backdrop/backdrop-issues/issues/857

PR #2139, opened 2018-04-21
Schedule3

docwilmot:
Fixes backdrop/backdrop-issues#857
Replaces #1262

### DIFF
--- a/core/modules/node/js/node.types.js
+++ b/core/modules/node/js/node.types.js
@@ -14,12 +14,20 @@ Backdrop.behaviors.contentTypes = {
     // Publishing settings.
     $context.find('#edit-workflow').backdropSetSummary(function() {
       var vals = [];
-      if (parseInt($context.find('input[name="status_default"]:checked').val())) {
-        vals.push(Backdrop.t('Published'));
-      }
-      else {
-        vals.push(Backdrop.t('Unpublished'));
-      }
+      // if (parseInt($context.find('input[name="status_default"]:checked').val())) {
+      //   vals.push(Backdrop.t('Draft'));
+      // }
+      // else if (parseInt($context.find('input[name="status_default"]:checked').val())) {
+      //   vals.push(Backdrop.t('Published'));
+      // }
+      // else {
+      //   vals.push(Backdrop.t('Schedule Publish Time'));
+      // }
+      $('div#edit-status-default', context).parent().change(function () {
+        var published_state = "";
+        published_state = $('input[type="radio"]:checked').next('label:first').text();
+        vals.push(Backdrop.checkPlain($.trim(published_state)));
+      }).trigger('change');
       if ($context.find('input[name="sticky_default"]:checked').length) {
         vals.push(Backdrop.t('Sticky'));
       }
--- a/core/modules/node/js/node.types.js
+++ b/core/modules/node/js/node.types.js
@@ -25,7 +25,7 @@ Backdrop.behaviors.contentTypes = {
       return vals.join(', ');
     });
 
-    // Uncheck the "Publish later" option if scheduling is disabled.
+    // Uncheck the "Schedule for later" option if scheduling is disabled.
     $context.find('input[name="scheduling_enabled"]').once().on('change', function() {
       var $checkedStatusDefault = $context.find('input[name="status_default"]:checked');
       if ($checkedStatusDefault.val() === '2') {
--- a/core/modules/node/js/node.types.js
+++ b/core/modules/node/js/node.types.js
@@ -14,20 +14,8 @@ Backdrop.behaviors.contentTypes = {
     // Publishing settings.
     $context.find('#edit-workflow').backdropSetSummary(function() {
       var vals = [];
-      // if (parseInt($context.find('input[name="status_default"]:checked').val())) {
-      //   vals.push(Backdrop.t('Draft'));
-      // }
-      // else if (parseInt($context.find('input[name="status_default"]:checked').val())) {
-      //   vals.push(Backdrop.t('Published'));
-      // }
-      // else {
-      //   vals.push(Backdrop.t('Schedule Publish Time'));
-      // }
-      $('div#edit-status-default', context).parent().change(function () {
-        var published_state = "";
-        published_state = $('input[type="radio"]:checked').next('label:first').text();
-        vals.push(Backdrop.checkPlain($.trim(published_state)));
-      }).trigger('change');
+      var defaultStatus = $context.find('input[name="status_default"]:checked').parent().text();
+      vals.push(Backdrop.checkPlain($.trim(defaultStatus)));
       if ($context.find('input[name="sticky_default"]:checked').length) {
         vals.push(Backdrop.t('Sticky'));
       }
@@ -37,6 +25,14 @@ Backdrop.behaviors.contentTypes = {
       return vals.join(', ');
     });
 
+    // Uncheck the "Publish later" option if scheduling is disabled.
+    $context.find('input[name="scheduling_enabled"]').once().on('change', function() {
+      var $checkedStatusDefault = $context.find('input[name="status_default"]:checked');
+      if ($checkedStatusDefault.val() === '2') {
+        $checkedStatusDefault.prop('checked', false);
+        $context.find('input[name="status_default"]:first').prop('checked', true);
+      }
+    });
 
     // Revision settings.
     $context.find('#edit-revision').backdropSetSummary(function() {
@@ -110,7 +106,7 @@ Backdrop.behaviors.contentTypes = {
 
     // Focus the input#edit-path-pattern field when clicking on the token
     // browser on the /admin/structure/types/add pages (add a content type).
-    $context.find('#edit-path .token-browser-link').click(function(){
+    $context.find('#edit-path .token-browser-link').once().on('click', function(){
       $('input#edit-path-pattern').focus();
     });
   }
--- a/core/modules/node/node.install
+++ b/core/modules/node/node.install
@@ -102,6 +102,13 @@ function node_schema() {
         'not null' => TRUE,
         'default' => 0,
       ),
+      'published_on' => array(
+        'description' => 'The UNIX UTC timestamp when to publish',
+        'type' => 'int',
+        'unsigned' => TRUE,
+        'not null' => TRUE,
+        'default' => 0,
+      ),
     ),
     'indexes' => array(
       'node_changed'        => array('changed'),
@@ -1085,6 +1092,21 @@ function node_update_1011() {
   if (db_table_exists('node_type')) {
     db_drop_table('node_type');
   }
+}
+
+/**
+ * Add column 'published_on' to the node table to allow nodes to be scheduled
+ * for publication at a future date.
+ */
+function node_update_1012() {
+  $spec = array(
+    'type' => 'int',
+    'description' => 'The UNIX UTC timestamp when to publish',
+    'unsigned' => TRUE,
+    'not null' => TRUE,
+    'default' => 0,
+  );
+  db_add_field( 'node', 'published_on', $spec);
 }
 
 /**
--- a/core/modules/node/node.install
+++ b/core/modules/node/node.install
@@ -1144,8 +1144,7 @@ function node_update_1014() {
 }
   
 /**
- * Add column 'published' to the node table to allow nodes to be scheduled
- * for publication at a future date.
+ * Add column for scheduling the publishing of content in the future.
  */
 function node_update_1015() {
   if (db_field_exists('node', 'scheduled')) {
--- a/core/modules/node/node.install
+++ b/core/modules/node/node.install
@@ -71,6 +71,12 @@ function node_schema() {
         'not null' => TRUE,
         'default' => 0,
       ),
+      'scheduled' => array(
+        'description' => 'A Unix timestamp indicating content is scheduled to be published in the future.',
+        'type' => 'int',
+        'not null' => TRUE,
+        'default' => 0,
+      ),
       'comment' => array(
         'description' => 'Whether comments are allowed on this node: 0 = no, 1 = closed (read only), 2 = open (read/write).',
         'type' => 'int',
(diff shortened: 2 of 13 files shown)


core/modules/node/js/node.types.js
```diff
@@ -14,12 +14,20 @@ Backdrop.behaviors.contentTypes = {
     // Publishing settings.
     $context.find('#edit-workflow').backdropSetSummary(function() {
       var vals = [];
-      if (parseInt($context.find('input[name="status_default"]:checked').val())) {
-        vals.push(Backdrop.t('Published'));
-      }
-      else {
-        vals.push(Backdrop.t('Unpublished'));
-      }
+      // if (parseInt($context.find('input[name="status_default"]:checked').val())) {
```
jenlampton: When we're ready to commit we should remove this instead of comment out.

core/modules/node/node.install
```diff
@@ -102,6 +102,13 @@ function node_schema() {
         'not null' => TRUE,
         'default' => 0,
       ),
+      'published_on' => array(
```
jenlampton: We should probably name this database column `published` to be more consistent with the other timestamp fields for created and updated. (In general we don't include prepositions in machine names or property names.)

core/modules/node/node.module
```diff
@@ -905,6 +917,11 @@ function node_validate($node, $form, &$form_state) {
     form_set_error('date', t('You have to specify a valid date.'));
   }
 
+  // Validate the "Publish on" field.
+  //if (!empty($node->future_publish) && strtotime($node->future_publish) < REQUEST_TIME) {
```
jenlampton: more comments here...  delete?

core/modules/node/node.pages.inc
```diff
@@ -293,10 +293,47 @@ function node_form($form, &$form_state, Node $node) {
...
+  }
+  $form['options']['publishing_states'] = array(
+    '#type' => 'radios',
+    '#options' => array(
+      1 => t('Publish now'),
+      0 => t('Save draft'),
+      2 => t('Publish later'),
+    ),
+    '#default_value' => $is_scheduled,
+    '#description' => t('Content saved as draft is only accessible to the content creator or site administrators. Content set to Publish later will be saved as draft when the content is created, but will be automatically published on the date and time selected; this requires a <a href="@cron">cron maintenance task</a>.', array('@cron' => url('admin/reports/status')))
```
jenlampton: Let's use `author` instead of `content creator` here.

core/modules/node/node.module
```diff
@@ -112,6 +112,18 @@ function node_cron() {
   db_delete('history')
     ->condition('timestamp', NODE_NEW_LIMIT, '<')
     ->execute();
+
+  $get_nodes_to_publish = db_query(
+      "SELECT * FROM {node} WHERE published_on < UNIX_TIMESTAMP(NOW())
+      AND status = '0'
+      AND published_on != 0"
+  );
+  foreach($get_nodes_to_publish as $n) {
```
jenlampton: This is a fairly expensive operation, we should put some kind of a limit on here so a cron task won't take the site down. Limiting to 20 nodes would cover the most common use-cases. 

How about we switch this `db_query()` to a `db_query_range()`?

core/modules/node/node.pages.inc
```diff
@@ -431,6 +479,21 @@ function node_form_submit_build_node($form, &$form_state) {
   form_execute_handlers('submit', $form, $form_state);
 
   $node = $form_state['node'];
+  if ($form_state['values']['publishing_states'] == 1) {
+    $node->status = 1;
+    $node->published_on = 0;
```
jenlampton: Same thing here with the property name. The others are `created` and `changed` so this one should be `published`

core/modules/node/node.pages.inc
```diff
@@ -293,10 +293,47 @@ function node_form($form, &$form_state, Node $node) {
...
+    elseif ($node->status == 1) {
+      $is_scheduled = 1;
+    }
+  }
+  $form['options']['publishing_states'] = array(
+    '#type' => 'radios',
+    '#options' => array(
+      1 => t('Publish now'),
+      0 => t('Save draft'),
+      2 => t('Publish later'),
```
jenlampton: Can we change this label to something more like `Schedule for later` so that it is clear that we need to set a date?

core/modules/node/node.pages.inc
```diff
@@ -293,10 +293,47 @@ function node_form($form, &$form_state, Node $node) {
...
+      $is_scheduled = 2;
+    }
+    elseif ($node->status == 0) {
+      $is_scheduled = 0;
+    }
+    elseif ($node->status == 1) {
+      $is_scheduled = 1;
+    }
+  }
+  $form['options']['publishing_states'] = array(
```
jenlampton: I'm not in love with the name `publishing_states` as it seems a little verbose (though, accurate). I think I'd prefer simply `state`.

And if we change this to `state` than `status_default` should also become `state_default` to match.

jenlampton (commented):
code review.

core/modules/node/node.pages.inc
jenlampton: screen shot 2018-04-21 at 3 21 08 pm

This will need to be split into different bits of help text to be matched with each option. That's a little tricky with FormAPI, let me look up the syntax.... Here we go:
```
$form['publishing_states']['0']['#description'] = t('Will only be accessible to the author or site administrators');
$form['publishing_states']['2']['#description'] =  t('Will be saved as draft now, and automatically published on the date and time selected; this requires a <a href="@cron">cron maintenance task</a>.', array('@cron' => url('admin/reports/status')));
```

We'll also need an access check on `admin/reports/status` before adding `; this requires a <a href="@cron">cron maintenance task</a>.` to the message (or before adding the link)

core/modules/node/node.pages.inc
```diff
@@ -394,6 +433,15 @@ function node_form_submit($form, &$form_state) {
     watchdog('content', '@type: updated %title.', $watchdog_args, WATCHDOG_NOTICE, $node_link);
     backdrop_set_message(t('@type %title has been updated.', $t_args));
   }
+  if ($node->published_on) {
+    $message = '';
+    if (!empty($original_status)) {
+      $message .= t('@type has been unpublished.', array('@type' => $type)) . ' ';
+    }
+    $today = date("D M j G:i:s", strtotime($node->future_publish));
```
jenlampton: screen shot 2018-04-21 at 3 32 24 pm
This date/time format is hard for me to understand. We should use one of the predefined formats so that if a site admin changes from am/pm to 24-hour site-wide (or vice versa) this time will update too. How about we use `short` like so:

```php
$today = format_date($node->future_publish, 'short');
```

core/modules/node/node.install
```diff
@@ -1136,6 +1143,21 @@ function node_update_1014() {
...
+ */
+function node_update_1015() {
+  $spec = array(
+    'type' => 'int',
+    'description' => 'The UNIX UTC timestamp when to publish',
+    'unsigned' => TRUE,
+    'not null' => TRUE,
+    'default' => 0,
+  );
+  db_add_field( 'node', 'published_on', $spec);
```
jenlampton: `published_on` => `published`

core/modules/node/node.module
```diff
@@ -112,6 +112,18 @@ function node_cron() {
   db_delete('history')
     ->condition('timestamp', NODE_NEW_LIMIT, '<')
     ->execute();
+
+  $get_nodes_to_publish = db_query(
+      "SELECT * FROM {node} WHERE published_on < UNIX_TIMESTAMP(NOW())
```
jenlampton: `published_on` => `scheduled`

core/modules/system/system.admin.inc
```diff
@@ -1516,7 +1516,7 @@ function system_cron_settings($form, &$form_state) {
     '#title' => t('Run cron every'),
     '#description' => t('More information about setting up scheduled tasks can be found by <a href="@url">reading the cron tutorial on drupal.org</a>.', array('@url' => url('http://drupal.org/cron'))),
     '#default_value' => config('system.core')->get('cron_safe_threshold'),
-    '#options' => array(0 => t('Never')) + backdrop_map_assoc(array(3600, 10800, 21600, 43200, 86400, 604800), 'format_interval'),
+    '#options' => array(0 => t('Never')) + backdrop_map_assoc(array(600, 1800, 3600, 10800, 21600, 43200, 86400, 604800), 'format_interval'),
```
jenlampton, line 1519: This is fantastic :) Can we please add options for 5m (300) and 15m (900) in here too?

core/modules/node/node.pages.inc
herbdool: We can probably leave out the cron message here, no? By default, cron runs anyway. And those who know enough to turn it off, probably don't need to be told.

BTW, for the comment automatic closer functionality I don't mention cron.

jenlampton:
@herbdool I thought about that too, but in the end I decided it would be helpful for someone who scheduled something to publish, but then didn't see it get published (or get published on time) to have an answer right here as to why not.

core/modules/node/node.install
docwilmot: This is going to confuse the heck out of everyone next to the `status` column. I'm suggesting `published_date` or `publish_date` (since its not yet happened).

core/modules/node/node.module
docwilmot: We could use the `node_mass_update()` function. Or better, Queue API.
The difficulty with limiting the numbers in either case is that if we have a cron job every 6 hours, and the first batch don't get published, the user will wonder why their `12:00am` post only published itself at `06:00am`. 
Would a direct DB update of the status column be too bad? (hooks wont run).

core/modules/node/node.module
docwilmot: I wasnt sure what to do here, should have said. We could either have a validation function that prevents `Publish on` dates in the past, or just ignore and simply publish immediately if the date is past. I'm thinking if the user filled that field he probably intended to schedule things and would be good to let him know he might have made a typo.

core/modules/node/node.pages.inc
docwilmot: I personally think we should keep `status` in case people been `form_altering`.

core/modules/node/node.pages.inc
docwilmot: Agree

core/modules/node/node.pages.inc
docwilmot: > Let's use author instead of content creator here.

this is the current expression in Backdrop; I didnt add this. (`admin/structure/types/manage/page`). But author is good.
> This will need to be split into different bits of help text to be matched with each option ... FormAPI...

I think this is a bit too busy. With clearer titles for the radios as you suggested earlier, a single simple explanation should be enough. We dont need to hand-hold that much.
> We'll also need an access check 
We can probably leave out the cron message here

I forgot cron runs by default. I think we should delete this

core/modules/node/node.pages.inc
docwilmot: True.

core/modules/node/node.module
ghost: > I'm thinking if the user filled that field he probably intended to schedule things and would be good to let him know he might have made a typo.

I agree.

core/modules/node/node.module
herbdool: I didn't put any batching in the comment auto closer and it updates nodes in a similar manner, which passed Nate's review. It's probably unlikely there will be a tonne of nodes but you never know.

core/modules/node/node.module
herbdool: Though we only want there error message if the node isn't already published. Otherwise it'll trigger whenever someone edits old nodes.

herbdool (commented):
I feel we should make this optional per content type. There will be some content types where this doesn't make sense.

core/modules/node/node.module
quicksketch: I agree, for the initial implementation no batching is necessary.

quicksketch:
The most recent commit isn't getting updates from Zen.ci here on GitHub, but looking at the results in Zen.ci, I can see they are all currently passing.

core/modules/node/node.types.inc
```diff
@@ -163,11 +163,18 @@ function node_type_form($form, &$form_state, $type = NULL) {
...
     '#title' => t('Default status'),
     '#options' => array(
-      1 => t('Published'),
-      0 => t('Unpublished'),
+      NODE_PUBLISHED => t('Publish now'),
+      NODE_NOT_PUBLISHED => t('Save draft'),
+      NODE_SCHEDULED => t('Publish later'),
+    ),
+    '#default_value' => isset($type->settings['status_default']) ? $type->settings['status_default'] : NODE_PUBLISHED,
+    '#description' => t('Content saved as draft is only accessible to the content creator or site administrators.') . ' ' . t('Content set to Schedule for later will be saved as draft when the content is created, but will be automatically published on the date and time selected.'),
```
docwilmot: "Schedule for later" isnt an option in the radio list. Need to change either the radio list or the description text.

herbdool:
I've tested. Created a node and scheduled for the future and ran cron once the date had passed. It published. I created a node further into the future and ran cron. It stayed unpublished as expected. I also created a content type where the option was disabled. That worked as expected as well.

Where did the "Promoted" and "Sticky" checkboxes go?

I'm not sure about the switch from being *states* of the node to *transitions*. Before if the node was published the checkbox "Published" was checked. Now it says "Publish now" which doesn't necessarily convey to the user that it is actually published, but will be if we leave it on that option.

laryn:
@herbdool If the page is being edited and is already published, maybe we can change that to say "Update and keep published" or just "Keep published"?

herbdool:
@laryn Or add a message saying the node is published, in addition to the radio buttons.

quicksketch:
> Where did the "Promoted" and "Sticky" checkboxes go?

They are still there on "Post" types, but "Page" does not have them enabled out of the box. They are still where they were before, below the Publish element (now radio buttons instead of a checkbox).

image

> I'm not sure about the switch from being states of the node to transitions. Before if the node was published the checkbox "Published" was checked. Now it says "Publish now" which doesn't necessarily convey to the user that it is actually published, but will be if we leave it on that option.

Agreed on this. The state transitions can also be difficult to convey. As @laryn suggested we could attempt to make the radio buttons match the transitions. But there are more than I would have expected. You'd have to have labels for every different kind of transition, e.g. new to draft, published to draft, draft to draft. You end up with state^2 different labels.

So let's go back to just labeling as states instead of transitions. I've set the labels now to "Published", "Draft" and "Schedule for later". (Perhaps that should be "Scheduled for later", but either way seems good to me).

quicksketch:
With the RTBC in the main issue, let's move on this and follow-up as needed with further minor updates. I've merged this into 1.x for 1.10. Thanks everyone here!

PR #2141, opened 2018-04-21
Issue #2475: Allow override display renderer

docwilmot:
Fixes https://github.com/backdrop/backdrop-issues/issues/2475


quicksketch:
Merged into 1.x for 1.10.

PR #2142, opened 2018-04-22
Issue #2353: Improve UX on dblog form.

serundeputy:


herbdool:
This needs some improvement on smaller screens. 
screenshot_20180422-133520

The divs should either be full width or two columns. The first select seems to be hard-coded to height so too much white space.

serundeputy:
Fair enough @herbdool I don't `dblog` on mobile ever; so I don't know what to do about it ... wanna take it from here?

PR #2143, opened 2018-04-23
Added entity_revision_is_default() function. Useful for porting modules from D7 that use or implement revisions.

moonray:


moonray:
An additional note: this function is required to port paragraphs module.

PR #2146, opened 2018-04-24
Issue #3052, Draggable contact categories.

opi:


core/modules/contact/contact.admin.inc
```diff
@@ -146,7 +160,7 @@ function contact_category_edit_form($form, &$form_state, array $category = array
       0 => t('No'),
       1 => t('Yes'),
     ),
-    '#default_value' => $selected,
+    '#default_value' => (isset($category['cid']) && $default_category == $category['cid']),
```
quicksketch, line 167: Nice shortening of the code here.

core/modules/contact/contact.admin.inc
```diff
@@ -10,68 +10,90 @@
...
+
     }
   }
 
-  if (!$rows) {
-    $rows[] = array(array(
-      'data' => t('No categories available.'),
-      'colspan' => 5,
-    ));
+  $form['actions'] = array('#type' => 'actions', '#tree' => FALSE);
```
quicksketch: Rather than setting #tree on the entire form and then setting FALSE on the actions, you could just set $form['categories']['#tree'] = TRUE.

core/modules/contact/contact.module
```diff
@@ -5,6 +5,22 @@
...
+/**
+ * Implements hook_theme().
+ */
+function contact_theme() {
+  $base = array(
+    'file' => 'contact.theme.inc',
+  );
+
+  return array(
+    'contact_category_list' => array(
```
quicksketch: Although theming the entire form works, it would be better to theme just the $form['categories'] element, so that other form elements may be placed around the form via hook_form_alter() if needed.

quicksketch (changes requested):
This works great! A few minor suggestions that could be implemented, but I'll push them up in a follow-up commit.

quicksketch (approved):

quicksketch:
Yay, well my follow-up commit also fixed the test failures, so I think we're good to go here. I can't believe Contact module was still making people use "weight" fields! Tabledrag has been around for ages I'm really surprised it wasn't utilized here. Merged into 1.x for 1.10.

opi:
thanks for your follow-up, it's cleaner now without my shameless copypasta from taxonomy module !

PR #2153, opened 2018-04-28
Issue #2956: Translate custom blocks 

bjoern-st:


core/modules/block/tests/block.tests.info
```diff
@@ -3,3 +3,14 @@ name = Block functionality
...
 group = Block
 file = block.test
+
+[BlockTranslationTestCase]
+name = Block translation functionality
+description = Add, translate and delete custom block.
+group = Block
+file = block.translation.test
+
+; Added by Backdrop CMS packaging script on 2017-09-15
```
herbdool: Need to delete everything with "Added by Backdrop..." and after.

core/modules/block/tests/block_test/block_test.info
```diff
@@ -5,3 +5,8 @@ version = BACKDROP_VERSION
 type = module
 backdrop = 1.x
 hidden = TRUE
+
+; Added by Backdrop CMS packaging script on 2017-09-15
```
herbdool: Again, delete "Added by..."

core/modules/block/block.module
```diff
@@ -8,11 +8,16 @@
...
  */
 function block_permission() {
-  return array(
+  $permissions = array(
     'administer blocks' => array(
       'title' => t('Administer custom blocks'),
     ),
   );
+  if (module_exists('locale')) {
+    $permissions['translate blocks'] = array(
```
herbdool: The two permissions in the array would read better if in the same format.

core/modules/block/block.info
```diff
@@ -5,3 +5,8 @@ version = BACKDROP_VERSION
 type = module
 backdrop = 1.x
 configure = admin/structure/block
+
+; Added by Backdrop CMS packaging script on 2017-09-15
```
herbdool, line 9: Delete "Added by..."

Also, there seem to a few files that show as having changed file permissions. We should try to avoid that.

core/modules/block/block.admin.inc
```diff
@@ -13,15 +13,13 @@ function block_admin_list() {
...
     t('Block'),
-    array('data' => t('Description'), 'class' => array('priority-low')),
     t('Operations'),
   );
 
   $rows = array();
   foreach ($custom_blocks_info as $delta => $block_info) {
     $row = array();
     $row[] = check_plain($block_info['info']);
-    $row[] = check_plain($block_info['description']);
```
herbdool: Why was this removed?

core/modules/block/block.admin.inc
```diff
@@ -13,15 +13,13 @@ function block_admin_list() {
 
   $header = array(
     t('Block'),
-    array('data' => t('Description'), 'class' => array('priority-low')),
```
herbdool: Why was this removed?

herbdool:
@bjoern-st overall this is looking pretty good! I found a couple issues that hopefully aren't too difficult to fix. 

When adding a custom block from a layout, there should be an option to choose the language once someone checks "make block reusable". It should default to None.

Currently there is no language option. There also seems to be an issue with the submit button disappearing off the page which might be a JS issue.

screenshot from 2018-04-29 10-58-11

It should look more like this from the custom blocks admin page. Though here too we should probably default to None rather than the default language.

screenshot from 2018-04-29 10-57-41

The problem with having no language option on the layout page means that the block ends up showing as one of the translation options even though it's None. This is also a problem with any block that has language set to None. This results in a nonsensical admin screen. And trying to edit the block doesn't provide an option to change the language from None to something else.

screenshot from 2018-04-29 10-53-58

Lastly, there seems to be an issue in the last screenshot when clicking "Configure block" in the breadcrumb. The URL is missing the /configure from the end so I get a white screen.

core/modules/block/block.admin.inc
bjoern-st: @herbdool 
Thanks for reviewing!
I removed that column because by now there was only the static "A reusable custom block." 
But in connection to https://github.com/backdrop/backdrop-issues/issues/2722 it makes completely sense to print that column again. Thank you for pointing this out.

I removed the packaging infos from the info files an will update the PR.

Bjoern

bjoern-st:
@herbdool 
You're completely right, that the language selection is missing on layout, if you create a new custom block. 

With the translation list of custom blocks here https://user-images.githubusercontent.com/1071483/39408016-97b8e34e-4b9d-11e8-9192-5a86f1a4e8fe.png it is a compromise ..
If you install a site in a different language you automatically have the modules "language" and "locale" enabled. So the translation facilities are triggered. 
So if you have a site like this and you update you'll automatically have the translation facilities.
So there are two choices. Give the block the neutral language  or the default language. 
The choice to make it language neutral is, that it might be easier to make a "before this update created" backdrop site multilingual. If you then go on your translation site you'll always have your original block as source and you can translate it.
The other reason was that you might have a block that you never thought to make it translatable like those, where for example just display an IFrame of a map, suddenly needs to be translatable because there suddenly appear some text, too. 
So the module right now is build like that - if no translation is present, the default language option is used .. and it thar case, it would be the language neutral option .. what I thought, it might be the best idea. 

I do not find it disturbing if the source is language neutral an you can edit it on that page and to add / edit your translations.
But of course that can be discussed.

The breadcrumb link SHOULD be covered .. I'll have to take a look at this point.

And thanks again for this review. 

Cheers Bjoern

olafgrabienski:
I've had a look at the sandbox site and was able to reproduce the issue mentioned by @herbdool with the submit button: When I try to add a custom block via the *Layout* page and check the `Make this block reusable` box, the section expands and the `Add block` button gets pushed too far below so that it's not visible anymore. I *think*, in other cases like this, where the form gets longer, you can scroll to the button but that's not possible here. (Was able to see the button again when I decreased the browser zoom to 90%.)

olafgrabienski:
Regarding the discussion about the missing language option when creating a custom block via a *Layout* page, I'm undecided about the user interface and about which option should be chosen as default. I think however that the defaults should be the same, no matter if you create a block via the *Custom blocks* page or via a *Layout* page.

herbdool:
@bjoern-st I can understand wanting to make a block with no language set. That part makes sense and is how it works with nodes. But the odd part that has never been a pattern with i18n nor with built-in Drupal/Backdrop is that Language Neutral is considered one of the translations on the translate tab. I think that will probably confuse people. The translate tab probably shouldn't be available until someone has changed the set language on a block to be something other than Neutral.

herbdool:
@bjoern-st here's my idea of how it would work more closely to node translation:
[2153b.patch.txt](https://github.com/backdrop/backdrop/files/1962512/2153b.patch.txt). The translate tab only shows up if the default language is not neutral. And the language select is visible so long as the block is neutral.

The node translation is a bit more complicated since it allows the language to change even if there are translations. Not sure if we want to go that far.

quicksketch:
New sandbox.

core/modules/block/block.module
```diff
@@ -89,7 +111,7 @@ function block_block_info() {
   foreach ($config_names as $config_name) {
     $block = config_get($config_name);
     $blocks[$block['delta']] = $block;
-    $blocks[$block['delta']]['description'] = !empty($block['description']) ? $block['description'] : t('A reusable custom block.');
+    $blocks[$block['delta']]['description'] = t('A reusable custom block.');
```
quicksketch: This is a regression that would cause block descriptions to get lost. This was a recently added feature in https://github.com/backdrop/backdrop-issues/issues/2722

core/modules/block/block.module
```diff
@@ -194,15 +237,6 @@ function block_custom_block_form($edit = array(), $stand_alone = TRUE) {
...
   );
-  $form['description'] = array(
-    '#type' => 'textfield',
-    '#title' => t('Admin description'),
-    '#default_value' => $edit['description'],
-    '#maxlength' => 128,
-    '#description' => t('Used in administrative interfaces and not shown to end-users.'),
-    '#weight' => -18,
-    '#id' => 'block-description',
-  );
```
quicksketch: More accidental reverts here that should be maintained.

core/modules/block/block.module
```diff
@@ -215,14 +249,6 @@ function block_custom_block_form($edit = array(), $stand_alone = TRUE) {
...
     '#disabled' => isset($edit['delta']),
   );
-  $form['title'] = array(
-    '#type' => 'textfield',
-    '#title' => t('Display title'),
-    '#default_value' => $edit['title'],
-    '#maxlength' => 255,
-    '#description' => t('The title of the block as shown to the user. This will affect all places where this block is used.'),
-    '#weight' => -17,
-  );
```
quicksketch: Not sure if this one is accidental or not.

core/modules/block/block.module
```diff
@@ -273,22 +321,64 @@ function block_custom_block_form($edit = array(), $stand_alone = TRUE) {
...
+    }
+  }
+
+  // if langcode != NULL we come from a translation form, so
+  // store the translation.
+  if ($translateable &&
+      ($langcode != NULL) &&
+      isset($block['default_langcode']) &&
+      ($langcode != $block['default_langcode'])) {
+    // store the translations
```
quicksketch: Minor code style things, code comments should be Sentence case and end with a period.

core/modules/block/tests/block.translation.test
```diff
@@ -0,0 +1,141 @@
+<?php
+
+/**
+ * @file
+ * Tests for block.module.
+ */
+
+class BlockTranslationTestCase extends BackdropWebTestCase {
+  protected $admin_user;
```
quicksketch, line 9: Ideally new tests added should use `protected $profile = 'testing';` to cut down on the installation time. But if that is problematic we can keep using the standard profile.

quicksketch (changes requested):
I haven't functional tested this yet. But code-review I found that this accidentally reverts https://github.com/backdrop/backdrop-issues/issues/2722. We'll definitely have to fix that before merging.

quicksketch:
I have merged in the updated mirror of this PR at https://github.com/backdrop/backdrop/pull/2163 into 1.x for 1.10. Thanks @bjoern-st  et al for the amazing work here!

PR #1528, opened 2016-09-09
Issue #1152: Add the ability to add a hero image block.

jenlampton:
Fixes https://github.com/backdrop/backdrop-issues/issues/1152#issuecomment-245833504

PR #2154, opened 2018-04-29
Issue #1528: Allow a layout to be selected on multiple paths

docwilmot:


docwilmot:
This is still early, pushing to get Zen CI's opinion. 
- Allows multiple paths via a textarea instead of a textfield
- Creates a listing for each path in the layouts list. This would need tweaking to group all paths sharing a layout.
- Creates a Menu Item for each appropriate path
- Generates a machine name for each menu item
- Creates a corresponding menu_item config, named by this machine name, as in `layout.menu_item.<machine name>.json`

Not working:
- Layout contexts are currently keyed by path position; this cannot work with multiple paths as the keys will overwrite each other. WIll need to figure out how to redesign `Layout::getContexts()`
- Upgrade path
- Tests

PR #2159, opened 2018-04-30
Issue #2948: Add support for PHP 7.2

hosef:
Applied changes from drupal.org/project/drupal/issues/2947772 and then made more changes to remove all exceptions from tests when using PHP 7.2.

I did not apply the change to update Tar.php because I did have any of the errors reported against Drupal's version of Tar.php.


quicksketch:
Merged into 1.x and 1.9.x. Thanks @hosef!

PR #2165, opened 2018-05-01
Issue #2694, Add easy-to-enable HTTPS redirect rules in .htaccess.

opi:
opiniated change included, I remove the directive that define the `[E=protossl]` env variable, unused since https://github.com/backdrop/backdrop/commit/2f9a04eb03066c24852c5564fae6482d8770ec80#diff-8052c42ab3b8aa06a3f5f788a4ddccc2

### DIFF
--- a/.htaccess
+++ b/.htaccess
@@ -51,13 +51,6 @@ DirectoryIndex index.php index.html index.htm
 <IfModule mod_rewrite.c>
   RewriteEngine on
 
-  # Set "protossl" to "s" if we were accessed via https://.  This is used later
-  # if you enable "www." stripping or enforcement, in order to ensure that
-  # you don't bounce between http and https.
-  RewriteRule ^ - [E=protossl]
-  RewriteCond %{HTTPS} on
-  RewriteRule ^ - [E=protossl:s]
-
   # Make sure Authorization HTTP header is available to PHP
   # even when running as CGI or FastCGI.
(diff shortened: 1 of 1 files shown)


.htaccess
```diff
@@ -51,13 +51,6 @@ DirectoryIndex index.php index.html index.htm
 <IfModule mod_rewrite.c>
   RewriteEngine on
 
-  # Set "protossl" to "s" if we were accessed via https://.  This is used later
-  # if you enable "www." stripping or enforcement, in order to ensure that
-  # you don't bounce between http and https.
-  RewriteRule ^ - [E=protossl]
-  RewriteCond %{HTTPS} on
```
jenlampton, line 58: Why was this removed? Isn't it still needed for the other www and non-www redirects?

opi: Nope, not needed since https://github.com/backdrop/backdrop/commit/2f9a04eb03066c24852c5564fae6482d8770ec80#diff-8052c42ab3b8aa06a3f5f788a4ddccc2, as we use plain `example.com` domain.

opi: but as I said (in this PR first comment), it's an opiniated change I can revert to focus on the https redirect rules

quicksketch:
Thanks @opi for the explanation. Looks like you're right. Merged this into 1.x for 1.10.

PR #1630, opened 2016-10-30
Issue #2313: [UX] Rephrase "An illegal choice has been detected"

klonos:
Fixes https://github.com/backdrop/backdrop-issues/issues/2313

PR #2177, opened 2018-05-05
Issue #2313: Rephrase 'An illegal choice has been detected' since no …

jenlampton:
Fixes https://github.com/backdrop/backdrop-issues/issues/2313
Replaces https://github.com/backdrop/backdrop/pull/1630

### DIFF
--- a/core/includes/form.inc
+++ b/core/includes/form.inc
@@ -1389,8 +1389,8 @@ function _form_validate(&$elements, &$form_state, $form_id = NULL) {
           $value = in_array($elements['#type'], array('checkboxes', 'tableselect')) ? array_keys($elements['#value']) : $elements['#value'];
           foreach ($value as $v) {
             if (!isset($options[$v])) {
-              form_error($elements, $t('An illegal choice has been detected. Please contact the site administrator.'));
-              watchdog('form', 'Illegal choice %choice in !name element.', array('%choice' => $v, '!name' => empty($elements['#title']) ? $elements['#parents'][0] : $elements['#title']), WATCHDOG_ERROR);
+              form_error($elements, $t('Invalid option %choice in %name element.', array('%choice' => $v, '%name' => empty($elements['#title']) ? $elements['#parents'][0] : $elements['#title'])));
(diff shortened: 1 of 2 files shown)


core/includes/form.inc
```diff
@@ -1389,8 +1389,8 @@ function _form_validate(&$elements, &$form_state, $form_id = NULL) {
           $value = in_array($elements['#type'], array('checkboxes', 'tableselect')) ? array_keys($elements['#value']) : $elements['#value'];
           foreach ($value as $v) {
             if (!isset($options[$v])) {
-              form_error($elements, $t('An illegal choice has been detected. Please contact the site administrator.'));
-              watchdog('form', 'Illegal choice %choice in !name element.', array('%choice' => $v, '!name' => empty($elements['#title']) ? $elements['#parents'][0] : $elements['#title']), WATCHDOG_ERROR);
+              form_error($elements, $t('Invalid option %choice in !name element.', array('%choice' => $v, '!name' => empty($elements['#title']) ? $elements['#parents'][0] : $elements['#title'])));
+              watchdog('form', 'Invalid option %choice in !name element.', array('%choice' => $v, '!name' => empty($elements['#title']) ? $elements['#parents'][0] : $elements['#title']), WATCHDOG_ERROR);
```
quicksketch: I hadn't noticed that we use two different ways of displaying the form element title. Here we used `!name`, but below we used `%name` with the exact same string and input value. Let's switch to `%name` here, better to fully escape the form label or parent rather than assume it's safe.

quicksketch (changes requested):
The invalid option test check is failing, that needs to be fixed before we can merge.

core/includes/form.inc
jenlampton: done.

quicksketch: Thanks! Though the same test is still failing. We still need that fixed.

jenlampton: yeah, I can't seem to figure out why it's failing.  Maybe I don't understand  the proper use of `format_string()`.

jenlampton:
Got it!  Testing now...

jenlampton:
Rebased again and pushed, still passing tests.

quicksketch (approved):

quicksketch:
Looks great. Merged into 1.x and 1.10.x.

PR #2178, opened 2018-05-06
Issue #3083: Help user of insert link dialogue by automatically adding URL protocol prefix to external links.

Graham-72:


quicksketch:
Thanks @Graham-72! Using the same functions and places in the code, I made an alternative PR at https://github.com/backdrop/backdrop/pull/2188 that accomplishes this without copying code out of link module.

PR #2193, opened 2018-05-16
Issue #3123: Hide the langcode markup if language module is not enabled.

quicksketch:
Fixes https://github.com/backdrop/backdrop-issues/issues/3123.


quicksketch:
Merged into 1.x and 1.10.x.

PR #2198, opened 2018-05-19
adapt 'Insert Image' to provide image library.

Graham-72:
The image library comprises the image files that have been previously uploaded and stored as managed image files.


docwilmot:
This sounds promising, but its not loading the library in the sandbox. Not ready yet?

Graham-72:
@docwilmot I find there is some difficulty caused by update to BD v10.0, and the view used to display the library is no longer being fetched, although the view preview is working. More news soon, hopefully.

I find it is the line `'#description' => views_embed_view('image_library'),` that is causing the problem. Maybe there is some new restriction on what can be entered as a description?

After some searching and testing I find the failure is a result of the change to jQuery UI introduced by issue #2493. The new version is not accepting the output of views_embed_view('image_library'). If I revert to the previous version used in BD 1.9.5 it works.

Further testing shows that it is not this line of code that is causing the problem. The code works in other places.

Graham-72:
Close this then reopen to create a new sandbox.

Graham-72:
The sandbox site is still not working, probably because of a Javascript problem. In my browser I am getting the following error report
```
Uncaught TypeError: Cannot read property 'form' of undefined
    at HTMLDocument.a.fn.form (jquery.ui.form.min.js?v=1.12.1:4)
    at i (jquery.js?v=1.12.4:2)
    at Object.add [as done] (jquery.js?v=1.12.4:2)
    at n.fn.init.n.fn.ready (jquery.js?v=1.12.4:2)
    at new n.fn.init (jquery.js?v=1.12.4:2)
    at n (jquery.js?v=1.12.4:2)
    at new Backdrop.ajax (ajax.js?v=1.11.x-dev:150)
    at new Backdrop.views.ajaxView (eval at <anonymous> (jquery.js?v=1.12.4:2), <anonymous>:73:26)
    at Object.eval (eval at <anonymous> (jquery.js?v=1.12.4:2), <anonymous>:14:37)
    at Function.each (jquery.js?v=1.12.4:2)
```
At one time I was getting this report on my own test server but this has now cleared.

Graham-72:
Well, my PR works on my test server if I revert back to jQuery UI version 1.10.4 as in BD 1.9.5, also revert the two other files changed in issue #2493.

Graham-72:
Reopening for new sandbox.

Graham-72:
Closing in order to reopen and add a contrib module.

Graham-72:
I have now identified two not very satisfactory answers 
(a) use it with older version of jQuery UI
or
(b) use it with contrib module 'Autologout' installed!!!

I now need to find out what it is about (b) that enables this to work. The specific problem is that the `function views_embed_view('image_library')` will not work when within `function filter_format_editor_image_form()`.

docwilmot:
@Graham-72  your Image Library View was configured to use AJAX. I turned off AJAX and disabled Auto Logout and it works fine.
But not so sure about embedding the View as a description. Couldnt it be added as its own form element?

Graham-72:
@docwilmot thank you very much for your comments. This is unfamiliar territory for me. I found an interesting comment at https://www.drupal.org/project/drupal/issues/2102349 about ajaxing within ajax, but clearly it is better not to use ajax in the view if this works.
Maybe I should put the view in its own form element; I just copied the solution by @daggerhart which worked well before the upgrade of jQuery UI. However, any suggestions for improving the coding are most welcome.

docwilmot:
I was curious and tried. This works for getting the View in its own form element. You'd need to fix the JS to make clicking an image add its url to the textfield as needed and make sure saving works.

``` 
 $max_filesize = !empty($upload_settings['max_size']) ? min(parse_size($upload_settings['max_size']), file_upload_max_size()) : file_upload_max_size();
  $existing_file = !empty($values['data-file-id']) ? file_load($values['data-file-id']) : NULL;
  $fid = $existing_file ? $existing_file->fid : NULL;
  $form['image']['fid'] = array(
    '#title' => t('Image upload'),
    '#type' => 'managed_file',
    '#upload_location' => $upload_settings['scheme'] . '://' . $upload_settings['directory'],
    '#default_value' => $fid ? $fid : NULL,
    '#upload_validators' => array(
      'file_validate_extensions' => array('gif png jpg jpeg'),
      'file_validate_size' => array($max_filesize),
      'file_validate_image_resolution' => array($max_dimensions),
    ),
    '#wrapper_attributes' => array(
      'data-editor-image-toggle' => t('Image Upload'),
    ),
    '#parents' => array('fid'),
    '#weight' => -10,
    '#access' => $element_supports_uploads && $upload_settings['status'] && user_access('upload editor images'),
  );

  // Image library, very similar to Image URL
  global $base_url;
  $form['image']['library_all'] = array(
    '#attributes' => array(
      'data-editor-image-toggle' => t('Select image'),
      'data-editor-image-library-base-url' => $base_url,
    ),
    '#title' => t('Select image'),
    '#parents' => array('image_library', 'src'),
    '#type' => 'container',
    '#weight' => -2,
  );
  $form['image']['library_all']['library_url'] = array(
    '#title' => t('Select image'),
    '#type' => 'textfield',
    '#element_validate' => array('_filter_format_editor_link_url_validate'),
    '#placeholder' => '/example/image.jpg',
    '#default_value' => isset($values['src']) ? $values['src'] : NULL,
    '#weight' => -2,
  );
  $form['image']['library_all']['library_all_library'] = array(
    '#title' => t('Select image'),
    '#markup' => views_embed_view('image_library'),
    '#type' => 'markup',
    '#weight' => -1,
  );
```

docwilmot:
But maybe having the View as a description is OK, I dont know. Maybe ask @quicksketch

quicksketch:
> @Graham-72 your Image Library View was configured to use AJAX. I turned off AJAX and disabled Auto Logout and it works fine.

I think we need AJAX enabled so that the pager and search work. Otherwise, going to the second page of thumbnails would load a new page.

>  But not so sure about embedding the View as a description. Couldnt it be added as its own form element?

Yeah, putting the entire view in `#description` definitely doesn't feel clean, plus you'd probably encounter all kinds of unusual styling problems inheriting from the description.

I'm not sure where the jQuery Forms problem is coming from. It would take some debugging.

docwilmot:
@Graham-72 I push a PR to your repo https://github.com/Graham-72/backdrop with code to open a gallery in a dialog instead. Problems:
- If AJAX is enabled on the View you have to push the "Select image" button twice (due to that same JS form error on the first click).
- (major) unfortunately the process of using the gallery seems like it detaches CKEditor from the textarea somehow, so when you click insert it doesnt actually insert. 😢

Graham-72:
I have made some changes to my PR, mainly in order to move the view out of the description field, and also to place the image display below the image settings.

It is work in progress.

docwilmot:
Did you have a look at my PR? I think opening the gallery in a separate dialog is the way to go.

Graham-72:
@docwilmot yes I have looked at and tried your PR, with limited success, and will now do more work on that or a similar approach, with the gallery in a separate dialog. You may well be right that this is the way to go.

docwilmot:
@Graham-72 actually right after I wrote the above I thought I should change my mind. Widening the dialog as youve done would solve the issues we have now, especially if we could get AJAX to work. BUT I would try and float the gallery off to the right and have the other form fields on the left. If AJAX doesnt cooperate, we could even give the gallery a scrollbar and fix the position of the other form elements, stuck to the left side. I think that would do nicely for now, dont you?

Graham-72:
> float the gallery off to the right and have the other form fields on the left

An excellent idea, I will try it.

Graham-72:
The gallery is now on the right, thanks @docwilmot for this excellent suggestion, though I haven't yet achieved this at all elegantly. 

I am much happier with it now as a UI, though I think there is need for more adjustment of the css and js. I have had to adjust the view so that it does not use ajax nor paging. A scrollbar will be a good addition. Also it would be better if the view were to be based on 150px wide versions of images to save download of originals. Other ideas most welcome.

docwilmot:
@Graham-72 I dont think pure CSS is enough to do this. I tried this, quickly and it works well; will need to fix the CSS (need to add a padding between the form and the gallery, and also this messes with the height of the dialog) but the principle is sound.

Add in `filter.theme.inc`

```
function theme_filter_format_editor_image_form($variables) {
  $form = $variables['form'];

  $output = '<div class="image-form-wrapper"><div class="image-form-left">';
  $output .= backdrop_render($form['image']);
  $output .= backdrop_render($form['alt']);
  $output .= backdrop_render($form['size']);
  $output .= backdrop_render($form['align']);
  $output .= backdrop_render($form['caption']);
  $output .= '</div><div class="image-form-right">';
  $output .= backdrop_render($form['library_view']);
  $output .= '</div></div>';

  $output .= backdrop_render_children($form);

  return $output;
}
```

Add to `filter_theme()`
```
    'filter_format_editor_image_form' => array(
      'render element' => 'form',
    ) + $base,
```

Add to `filter_format_editor_image_form()`
```
$form['#theme'] = 'filter_format_editor_image_form';
```

Add to the CSS
```
.image-form-wrapper {
  width:100%;
}
.image-form-left {
  float: left;
  width:30%;
}
.image-form-right {
  float: right;
  width:70%;
}
```

docwilmot:
Also, unrelated, you will notice that with your JS code, if you open the image dialog first time, clicking an image works to transfer its filename to the field with class `edit-image-library-src`, as expected. BUT if you close the dialog and click the icon to open it again, clicking an image in the library doesnt work anymore. That is because every time you re-open the dialog, Backdrop renames the form elements for some reason, so `edit-image-library-src` is renamed to `edit-image-library-src--2`.

In the initial PR I sent, I used the `name` of the textfield (`image_library[src]`) instead in the JS code because that remains the same.

laryn:
@Graham-72 Looking good! A few suggestions for consideration:

- How do you feel about using a square "scale and crop" style so all image previews are the same aspect ratio and we can get a nicely aligned rows and columns?
- If the width of `.library-view` is not hard-coded at 600px and we change from a grid display format, it would allow wider screens to display more images without requiring a scroll.
- Having a simple exposed filter for `filename contains` would be nice.

Graham-72:
Closing in order to open a fresh sandbox.

Graham-72:
Re-opening.

Graham-72:
~~Temporary problem with revised PR - please try later!~~
Amended PR now working in sandbox.
Library view has no paging and Ajax is off.

core/modules/filter/css/filter.css
```diff
@@ -91,3 +91,49 @@ input#edit-filters-filter-html-settings-allowed-html {
...
+  padding-left: 5px;
+  padding-right: 5px;
+  height: 170px;
+}
+
+.image-form-wrapper {
+  display: flex;
+}
+.image-form-left {
+  width:400px;
```
docwilmot: Curious, why a fixed 400px? I thought it displayed better with relative % widths, eg 30% 70%.

core/modules/filter/css/filter.css
```diff
@@ -91,3 +91,49 @@ input#edit-filters-filter-html-settings-allowed-html {
...
+}
+.image-form-right {
+  padding-left: 20px;
+  height: 400px;
+  overflow-y: scroll;
+  min-width: 50%;
```
docwilmot: Again, curious.

core/modules/filter/css/filter.css
```diff
@@ -91,3 +91,49 @@ input#edit-filters-filter-html-settings-allowed-html {
...
+.image-form-wrapper {
+  display: flex;
+}
+.image-form-left {
+  width:400px;
+}
+.image-form-right {
+  padding-left: 20px;
+  height: 400px;
+  overflow-y: scroll;
```
docwilmot: overflow `auto` would hide the scrollbar until it overflows.

core/modules/filter/css/filter.css
Graham-72: By using flex and a fixed LH column we can maximise the width and space available for images in the RH area, particularly relevant on large screens. I think maybe I will try reducing from 400 to nearer 300, whatever preserves the lavout within the LH area.

core/modules/filter/css/filter.css
Graham-72: The minimum width is of course only relevant in narrower displays. I think perhaps I could better set it to be 500px rather than 50% so that the minimum image library width is three images, albeit with horizontal scrolling.

core/modules/filter/css/filter.css
docwilmot: Actually, I think the opposite. By applying a fixed width for the left it enforces that width on smaller screens at the expense of the right. I would recommend width: 30%, max-width: 300px for the left column and 70% for right.

Graham-72: Well, I will try it next time I am using my large display. Perhaps @laryn has a preference?

laryn: I think @docwilmot has a good suggestion with the left at width 30% and max-width 300px (and I'd add flex-grow: 0). The right can be 70% (and flex-grow: 1) so it can expand as needed.

Graham-72: Thanks both of you. I will make the change.

docwilmot:
Looking great @Graham-72 . Can we completely hide the right column when viewing the image upload screen? Right now it the dialog shrinks size but we still have a horizontal scrollbar. Maybe a `hide()` somewhere during the switch. And maybe an `animate()` as well?

Graham-72:
@docwilmot thanks. This latest small change to the script removes the horizontal scrollbar by hiding the right hand column when it isn't needed. On my PC screen I still get the vertical scrollbar and I'm tempted to reduce line size and spacing slightly in the image upload/settings dialog to make things even tidier. Animation can wait for another day 😃

docwilmot:
The CSS for both right and left columns need to be redeclared individually when swithcing windows. With the PR now, when on the upload screen the left column still has width 350px as its width, so it doesnt occupy the full window. So in the JS, we'd need
```
        $(".image-form-left").css({"width":"100%"});
        $(".image-form-right").css({"display":"none"});
```
then on the switch:
```
        $(".image-form-left").css({"max-width":"whatever"});
        $(".image-form-right").css({"width":"whatever"});
```

I've written "whatever" there because that would need adjusting, since, on that note, I see youre still enforcing a fixed with rather than percentage widths.

docwilmot:
Or more ideally, toggle a class to the `.image-form-wrapper` on switch, and let the `filter.css` file handle the child widths, like `.image-form-wrapper.small .image-form-left {width: 100%}` and `.image-form-wrapper.wide .image-form-left {width: whatever}`

Graham-72:
@docwilmot thank you for your suggestions. The toggling action now introduces an extra style `image-form-left-full` to make the left-hand part of the dialog use the full 500px width when in 'image upload', which is a variation on your idea, keeping the numerical values in `filter.css`.

I have persisted with using `flex` with a fixed width for the left hand section of the dialog rather than use percentage widths because it seems to work well in tests and keeps a maximum variable width for the image library on the right. Minimum is set so that at least three columns are formed, albeit with horizontal scrolling on narrower screens.

docwilmot:
We'll have to disagree on the percentage.
If you open the dialog, click the toggles, close it, and reopen, it sticks to 500px. I suspect something in the dialog JS enforces the max-width.
Has this PR been discussed on a Thursday? Would be sad for you to do all this work and have consensus be that it isn't the accepted approach. @jenlampton @quicksketch ?

Graham-72:
> I suspect something in the dialog JS enforces the max-width.

@docwilmot thanks for spotting this. I had mistakenly left in the 'old' line of script for initialising things. Replacing it now.

jenlampton:
Close/Opening for new sandbox.

jenlampton:
Looking good!  I think we should re-work the view to rewrite existing fields instead of using global text fields to add markup, something more like this: 
```
{
    "_config_name": "views.view.image_library",
    "name": "image_library",
    "description": "",
    "tag": "default",
    "disabled": false,
    "base_table": "file_managed",
    "human_name": "Image Library",
    "core": "1.7",
    "display": {
        "default": {
            "display_title": "Master",
            "display_plugin": "default",
            "display_options": {
                "query": {
                    "type": "views_query",
                    "options": []
                },
                "access": {
                    "type": "perm"
                },
                "cache": {
                    "type": "none"
                },
                "exposed_form": {
                    "type": "basic"
                },
                "pager": {
                    "type": "none",
                    "options": {
                        "offset": "0"
                    }
                },
                "style_plugin": "default",
                "row_plugin": "fields",
                "fields": {
                    "uri": {
                        "id": "uri",
                        "table": "file_managed",
                        "field": "uri",
                        "relationship": "none",
                        "group_type": "group",
                        "ui_name": "",
                        "label": "Path",
                        "exclude": 1,
                        "alter": {
                            "alter_text": 0,
                            "text": "",
                            "make_link": 0,
                            "path": "",
                            "absolute": 0,
                            "external": 0,
                            "replace_spaces": 0,
                            "path_case": "none",
                            "trim_whitespace": 0,
                            "alt": "",
                            "rel": "",
                            "link_class": "",
                            "prefix": "",
                            "suffix": "",
                            "target": "",
                            "nl2br": 0,
                            "max_length": "",
                            "word_boundary": 1,
                            "ellipsis": 1,
                            "more_link": 0,
                            "more_link_text": "",
                            "more_link_path": "",
                            "strip_tags": 0,
                            "trim": 0,
                            "preserve_tags": "",
                            "html": 0
                        },
                        "element_type": "",
                        "element_class": "",
                        "element_label_type": "",
                        "element_label_class": "",
                        "element_label_colon": 1,
                        "element_wrapper_type": "",
                        "element_wrapper_class": "",
                        "element_default_classes": 1,
                        "empty": "",
                        "hide_empty": 0,
                        "empty_zero": 0,
                        "hide_alter_empty": 1,
                        "link_to_file": 0,
                        "file_download_path": 1
                    },
                    "filemime": {
                        "id": "filemime",
                        "table": "file_managed",
                        "field": "filemime",
                        "relationship": "none",
                        "group_type": "group",
                        "ui_name": "",
                        "label": "Mime type",
                        "exclude": 1,
                        "alter": {
                            "alter_text": 0,
                            "text": "",
                            "make_link": 0,
                            "path": "",
                            "absolute": 0,
                            "external": 0,
                            "replace_spaces": 0,
                            "path_case": "none",
                            "trim_whitespace": 0,
                            "alt": "",
                            "rel": "",
                            "link_class": "",
                            "prefix": "",
                            "suffix": "",
                            "target": "",
                            "nl2br": 0,
                            "max_length": "",
                            "word_boundary": 1,
                            "ellipsis": 1,
                            "more_link": 0,
                            "more_link_text": "",
                            "more_link_path": "",
                            "strip_tags": 0,
                            "trim": 0,
                            "preserve_tags": "",
                            "html": 0
                        },
                        "element_type": "",
                        "element_class": "",
                        "element_label_type": "",
                        "element_label_class": "",
                        "element_label_colon": 1,
                        "element_wrapper_type": "",
                        "element_wrapper_class": "",
                        "element_default_classes": 1,
                        "empty": "",
                        "hide_empty": 0,
                        "empty_zero": 0,
                        "hide_alter_empty": 1,
                        "link_to_file": 0,
                        "filemime_image": 0
                    },
                    "filesize": {
                        "id": "filesize",
                        "table": "file_managed",
                        "field": "filesize",
                        "relationship": "none",
                        "group_type": "group",
                        "ui_name": "",
                        "label": "Size",
                        "exclude": 1,
                        "alter": {
                            "alter_text": 0,
                            "text": "",
                            "make_link": 0,
                            "path": "",
                            "absolute": 0,
                            "external": 0,
                            "replace_spaces": 0,
                            "path_case": "none",
                            "trim_whitespace": 0,
                            "alt": "",
                            "rel": "",
                            "link_class": "",
                            "prefix": "",
                            "suffix": "",
                            "target": "",
                            "nl2br": 0,
                            "max_length": "",
                            "word_boundary": 1,
                            "ellipsis": 1,
                            "more_link": 0,
                            "more_link_text": "",
                            "more_link_path": "",
                            "strip_tags": 0,
                            "trim": 0,
                            "preserve_tags": "",
                            "html": 0
                        },
                        "element_type": "",
                        "element_class": "",
                        "element_label_type": "",
                        "element_label_class": "",
                        "element_label_colon": 1,
                        "element_wrapper_type": "",
                        "element_wrapper_class": "",
                        "element_default_classes": 1,
                        "empty": "",
                        "hide_empty": 0,
                        "empty_zero": 0,
                        "hide_alter_empty": 1,
                        "file_size_display": "formatted"
                    },
                    "type": {
                        "id": "type",
                        "table": "file_usage",
                        "field": "type",
                        "relationship": "none",
                        "group_type": "group",
                        "ui_name": "",
                        "label": "Entity type",
                        "exclude": 1,
                        "alter": {
                            "alter_text": 0,
                            "text": "",
                            "make_link": 0,
                            "path": "",
                            "absolute": 0,
                            "external": 0,
                            "replace_spaces": 0,
                            "path_case": "none",
                            "trim_whitespace": 0,
                            "alt": "",
                            "rel": "",
                            "link_class": "",
                            "prefix": "",
                            "suffix": "",
                            "target": "",
                            "nl2br": 0,
                            "max_length": "",
                            "word_boundary": 1,
                            "ellipsis": 1,
                            "more_link": 0,
                            "more_link_text": "",
                            "more_link_path": "",
                            "strip_tags": 0,
                            "trim": 0,
                            "preserve_tags": "",
                            "html": 0
                        },
                        "element_type": "",
                        "element_class": "",
                        "element_label_type": "",
                        "element_label_class": "",
                        "element_label_colon": 1,
                        "element_wrapper_type": "",
                        "element_wrapper_class": "",
                        "element_default_classes": 1,
                        "empty": "",
                        "hide_empty": 0,
                        "empty_zero": 0,
                        "hide_alter_empty": 1
                    },
                    "module": {
                        "id": "module",
                        "table": "file_usage",
                        "field": "module",
                        "relationship": "none",
                        "group_type": "group",
                        "ui_name": "",
                        "label": "Module",
                        "exclude": 1,
                        "alter": {
                            "alter_text": 0,
                            "text": "",
                            "make_link": 0,
                            "path": "",
                            "absolute": 0,
                            "external": 0,
                            "replace_spaces": 0,
                            "path_case": "none",
                            "trim_whitespace": 0,
                            "alt": "",
                            "rel": "",
                            "link_class": "",
                            "prefix": "",
                            "suffix": "",
                            "target": "",
                            "nl2br": 0,
                            "max_length": "",
                            "word_boundary": 1,
                            "ellipsis": 1,
                            "more_link": 0,
                            "more_link_text": "",
                            "more_link_path": "",
                            "strip_tags": 0,
                            "trim": 0,
                            "preserve_tags": "",
                            "html": 0
                        },
                        "element_type": "",
                        "element_class": "",
                        "element_label_type": "",
                        "element_label_class": "",
                        "element_label_colon": 1,
                        "element_wrapper_type": "",
                        "element_wrapper_class": "",
                        "element_default_classes": 1,
                        "empty": "",
                        "hide_empty": 0,
                        "empty_zero": 0,
                        "hide_alter_empty": 1
                    },
                    "counter": {
                        "id": "counter",
                        "table": "views",
                        "field": "counter",
                        "relationship": "none",
                        "group_type": "group",
                        "ui_name": "",
                        "label": "View result counter",
                        "exclude": 1,
                        "alter": {
                            "alter_text": 0,
                            "text": "",
                            "make_link": 0,
                            "path": "",
                            "absolute": 0,
                            "external": 0,
                            "replace_spaces": 0,
                            "path_case": "none",
                            "trim_whitespace": 0,
                            "alt": "",
                            "rel": "",
                            "link_class": "",
                            "prefix": "",
                            "suffix": "",
                            "target": "",
                            "nl2br": 0,
                            "max_length": "",
                            "word_boundary": 1,
                            "ellipsis": 1,
                            "more_link": 0,
                            "more_link_text": "",
                            "more_link_path": "",
                            "strip_tags": 0,
                            "trim": 0,
                            "preserve_tags": "",
                            "html": 0
                        },
                        "element_type": "",
                        "element_class": "",
                        "element_label_type": "",
                        "element_label_class": "",
                        "element_label_colon": 1,
                        "element_wrapper_type": "",
                        "element_wrapper_class": "",
                        "element_default_classes": 1,
                        "empty": "",
                        "hide_empty": 0,
                        "empty_zero": 0,
                        "hide_alter_empty": 1,
                        "counter_start": "1"
                    },
                    "fid": {
                        "id": "fid",
                        "table": "file_managed",
                        "field": "fid",
                        "relationship": "none",
                        "group_type": "group",
                        "ui_name": "",
                        "label": "",
                        "exclude": 0,
                        "alter": {
                            "alter_text": 1,
                            "text": "<img class=\"image-library-image\" src=\"[uri]\">",
                            "make_link": 0,
                            "path": "",
                            "absolute": 0,
                            "external": 0,
                            "replace_spaces": 0,
                            "path_case": "none",
                            "trim_whitespace": 0,
                            "alt": "",
                            "rel": "",
                            "link_class": "",
                            "prefix": "",
                            "suffix": "",
                            "target": "",
                            "nl2br": 0,
                            "max_length": "",
                            "word_boundary": 1,
                            "ellipsis": 1,
                            "more_link": 0,
                            "more_link_text": "",
                            "more_link_path": "",
                            "strip_tags": 0,
                            "trim": 0,
                            "preserve_tags": "",
                            "html": 0
                        },
                        "element_type": "0",
                        "element_class": "",
                        "element_label_type": "",
                        "element_label_class": "",
                        "element_label_colon": false,
                        "element_wrapper_type": "div",
                        "element_wrapper_class": "library_fid",
                        "element_default_classes": 0,
                        "empty": "",
                        "hide_empty": 0,
                        "empty_zero": 0,
                        "hide_alter_empty": 1,
                        "link_to_file": 0
                    },
                    "filename": {
                        "id": "filename",
                        "table": "file_managed",
                        "field": "filename",
                        "relationship": "none",
                        "group_type": "group",
                        "ui_name": "",
                        "label": "",
                        "exclude": 0,
                        "alter": {
                            "alter_text": 0,
                            "text": "",
                            "make_link": 0,
                            "path": "",
                            "absolute": 0,
                            "external": 0,
                            "replace_spaces": 0,
                            "path_case": "none",
                            "trim_whitespace": 0,
                            "alt": "",
                            "rel": "",
                            "link_class": "",
                            "prefix": "",
                            "suffix": "",
                            "target": "",
                            "nl2br": 0,
                            "max_length": "",
                            "word_boundary": 0,
                            "ellipsis": 0,
                            "more_link": 0,
                            "more_link_text": "",
                            "more_link_path": "",
                            "strip_tags": 0,
                            "trim": 0,
                            "preserve_tags": "",
                            "html": 0
                        },
                        "element_type": "p",
                        "element_class": "",
                        "element_label_type": "",
                        "element_label_class": "",
                        "element_label_colon": false,
                        "element_wrapper_type": "",
                        "element_wrapper_class": "",
                        "element_default_classes": 1,
                        "empty": "",
                        "hide_empty": 0,
                        "empty_zero": 0,
                        "hide_alter_empty": 1,
                        "link_to_file": 0
                    }
                },
                "filters": {
                    "filemime": {
                        "id": "filemime",
                        "table": "file_managed",
                        "field": "filemime",
                        "relationship": "none",
                        "group_type": "group",
                        "ui_name": "",
                        "operator": "starts",
                        "value": "image/",
                        "group": "1",
                        "exposed": false,
                        "expose": {
                            "operator_id": false,
                            "label": "",
                            "description": "",
                            "use_operator": false,
                            "operator": "",
                            "identifier": "",
                            "required": false,
                            "remember": false,
                            "multiple": false,
                            "remember_roles": {
                                "authenticated": "authenticated"
                            }
                        },
                        "is_grouped": false,
                        "group_info": {
                            "label": "",
                            "description": "",
                            "identifier": "",
                            "optional": true,
                            "widget": "select",
                            "multiple": false,
                            "remember": 0,
                            "default_group": "All",
                            "default_group_multiple": [],
                            "group_items": []
                        }
                    }
                },
                "sorts": [],
                "style_options": {
                    "grouping": [],
                    "row_class": "image-library-choose-file",
                    "default_row_class": 0,
                    "row_class_special": 0
                },
                "use_ajax": false,
                "header": {
                    "area": {
                        "id": "area",
                        "table": "views",
                        "field": "area",
                        "relationship": "none",
                        "group_type": "group",
                        "ui_name": "",
                        "label": "Library header",
                        "empty": 0,
                        "content": "<p>These images are available for selection.</p>\r\n",
                        "format": "full_html",
                        "tokenize": 0
                    }
                },
                "row_options": {
                    "default_field_elements": 0,
                    "inline": [],
                    "separator": "",
                    "hide_empty": 0
                }
            }
        }
    }
}
```

(This view is on the sandbox right now if you want to take a look at it there)

docwilmot:
This View is much better. I'd been heading towards a better View by trying to merge in File Entity module, but this will be obviously quicker.
It breaks the use of image thumbnails though, but I suspect @Graham-72 will fix that.

Graham-72:
> This View is much better

Better technically or visually? It seems to me to be the same visually. But when I tried it, it would not add the selected image. I will try and see what is happening, or not as the case may be

docwilmot:
> Better technically or visually

Technically. I still dont like the idea of re-written Views in core; should have at least a custom handler for this. 
> it would not add the selected image

The field names have changed so things will need fixing in the JS as you know. Same with the thumbnails not being generated. See my comment above.

Graham-72:
> should have at least a custom handler 

Yes, we need to create a special version of the uri field that creates the uri of the thumbnail. Then this can be used by Views in the rewrite of the File ID which currently is
 ``

I will see what I can do.

core/modules/filter/views/views_handler_field_file_uri.inc
```diff
@@ -0,0 +1,52 @@
...
+
+  function render($values) {
+    $data = $values->{$this->field_alias};
+    if (!empty($this->options['file_download_path']) && $data !== NULL && $data !== '') {
+      $data = file_create_url($data);
+      if (!empty($this->options['image_thumbnail'])) {
+      // $data contains URI of image.
+      // Find insertion position of '/files'.
+        $insert_at = strpos($data, '/files');
+      // Text to insert.
```
Graham-72: Need to replace the word 'thumbnail' in this text with the value available from config settings.

core/modules/filter/filter.admin.inc
```diff
@@ -614,6 +614,13 @@ function filter_editor_image_upload_settings_form($format) {
     '#field_suffix' => 'pixels',
     '#states' => $show_if_image_uploads_enabled,
   );
+  $form['thumbnail_style'] = array(
+    '#type' => 'textfield',
+    '#default_value' => $settings['thumbnail_style'],
+    '#title' => t('Thumbnail style name'),
+    '#description' => t("The machine name of the style used for library thumbnails. <br> This style must exist in config/media/image-styles."),
+    '#states' => $show_if_image_uploads_enabled,
+  );
```
quicksketch: Two questions here:

- Do we need this setting here at all? This seems like something that could be configured in the view.
- If we do keep this, we should make this a select list that lists available image styles, not a textfield.

core/modules/filter/css/filter.css
```diff
@@ -87,7 +105,111 @@ input#edit-filters-filter-html-settings-allowed-html {
...
+  font-size: 0.85rem;
+}
+.image-form-right .form-item {
+  margin-top: 0;
+}
 .form-item label.editor-image-toggle {
   display: inline-block !important;
   padding-bottom: 10px;
 }
+.image-library-choose-file p {
```
quicksketch: We should give this `p` a class and target its class directly rather than selecting based on the tag type.

core/modules/filter/css/filter.css
```diff
@@ -87,7 +105,111 @@ input#edit-filters-filter-html-settings-allowed-html {
...
+}
+.image-library-choose-file:hover {
+  background-color: #e5f3ff;
+  border: 1px solid #e5f3ff;
+  border-radius: 3px;
+}
+
+.image-library-image-selected {
+  background-color: #cce8ff !important;
+  border: 1px solid #99d1ff !important;
```
quicksketch: We should see if we can eliminate these `!important` calls.

core/modules/filter/filter.install
```diff
@@ -85,6 +85,20 @@ function filter_update_1001() {
...
 }
 
+/**
+ * Import the image_library view
+ */
+function filter_update_1002() {
+  $config = config('views.view.image_library');
+  if ($config->isNew()) {
+    $path = backdrop_get_path('module', 'filter') . '/config/';
+    $contents = file_get_contents($path . 'views.view.image_library.json');
```
quicksketch: We can't pull in something from the installation defaults into an update hook. In the event that the default changes, it would through off subsequent update hooks. We have to copy the entire view into the update hook. See `file_update_1000()` for an example.

core/modules/filter/filter.module
```diff
@@ -2542,3 +2545,14 @@ function _filter_image_align($text) {
...
  * @} End of "Standard filters".
  */
+
+/**
+ * Implements hook_autoload_info().
+ */
+function filter_autoload_info() {
+  return array(
+    // Views handlers.
+    'views_handler_field_file_uri' => 'views/views_handler_field_file_uri.inc',
```
quicksketch: This should be in File module, not Filter module.

core/modules/filter/filter.pages.inc
```diff
@@ -297,6 +301,19 @@ function filter_format_editor_image_form($form, &$form_state, $format) {
     '#access' => !empty($format->filters['filter_image_caption']->status),
   );
 
+  $form['library_view'] = array(
+    '#title' => t('Image library'),
+    '#type' => 'textfield',
+    '#default_value' => 'Image library',
+    '#wrapper_attributes' => array(
+      'class' => array('library-view'),
+      'data-editor-image-library-view' => views_embed_view('image_library'),
```
quicksketch: I'm glad to see a data attribute used here to trigger the behavior, but embedding an entire view within a single HTML attribute is going to result in some awkward markup. This seems like it should be `#type = 'markup'`. Perhaps something like this:

```
$form['library_view'] = array(
  '#markup' => '<div class="library-view" data-editor-library-view="true">' . views_embed_view('image_library') . '</div>';  
);
```

core/modules/filter/filter.pages.inc
```diff
@@ -210,23 +210,27 @@ function filter_format_editor_image_form($form, &$form_state, $format) {
...
+      'data-editor-image-toggle' => t('Upload an image'),
     ),
     '#parents' => array('fid'),
     '#weight' => -10,
     '#access' => $element_supports_uploads && $upload_settings['status'] && user_access('upload editor images'),
   );
 
-  // Use a "textfield" rather than "url" to allow relative paths.
-  $form['image']['src'] = array(
-    '#title' => t('Image URL'),
```
quicksketch: Replacing the `Image URL` with `Select from library` breaks embedding external images of any kind. Rather than replacing `Image URL`, it seems like this should be a 3rd option.

core/modules/filter/filter.pages.inc
```diff
@@ -321,13 +338,18 @@ function filter_format_editor_image_form_submit($form, &$form_state) {
     $form_state['values']['attributes']['data-file-id'] = NULL;
   }
   // Convert any uploaded files from the FID values to the src attribute.
-  if (!empty($form_state['values']['fid'])) {
+  if (isset($form_state['values']['fid'])) {
     $fid = $form_state['values']['fid'];
-    $file = file_load($fid);
+    if ($fid == 0) {
+      $url = '/files' . $form_state['values']['image_library']['src'];
```
quicksketch: If there is an external image, this results in a path like `/fileshttp://example.com/foo.png`, where `/files` is tacked to the front of a URL.

core/modules/filter/filter.theme.inc
```diff
@@ -198,3 +198,21 @@ function theme_filter_tips($variables) {
 
   return $output;
 }
+
+function theme_filter_format_editor_image_form($variables) {
+  $form = $variables['form'];
+  
+  $output = '<div class="image-form-wrapper"><div class="image-form-left">';
+  $output .= backdrop_render($form['library_view']);
+  $output .= '</div><div class="image-form-right">';
```
quicksketch: We shouldn't use classes like "left" or "right" because on an RTL language, these will be reversed. Instead name these based on their contents, e.g. "editor-image-library" and "editor-image-fields".

core/modules/filter/js/filter.js
```diff
@@ -140,12 +141,19 @@ Backdrop.behaviors.editorImageDialog = {
...
+    // Set the class for the left-hand part.
+    $(".image-form-right").addClass("image-form-right-full");
+
     $newToggles.click(function(e) {
       var $link = $(e.target);
       if ($link.is('.editor-image-toggle') === false) {
         return;
       }
-
+      // find the first ancestor of link
```
quicksketch: Make this sentence case and with a period.

core/modules/filter/js/filter.js
```diff
@@ -140,12 +141,19 @@ Backdrop.behaviors.editorImageDialog = {
       });
     });
 
+    // Initialise styles of Dialog.
+    // Hide the left-hand (library) part of the dialog form.
+    $(".image-form-left").css({"display":"none"});
+    $(".editor-dialog").removeClass("editor-dialog-with-library");
+    // Set the class for the left-hand part.
+    $(".image-form-right").addClass("image-form-right-full");
```
quicksketch: Avoid right/left class names again here.

core/modules/filter/views/views_handler_field_file_uri.inc
```diff
@@ -0,0 +1,52 @@
...
+      '#title' => t('Display download path instead of file storage URI'),
+      '#description' => t('This will provide the full download URL rather than the internal filestream address.'),
+      '#type' => 'checkbox',
+      '#default_value' => !empty($this->options['file_download_path']),
+    );
+
+    $form['image_thumbnail'] = array(
+      '#type' => 'checkbox',
+      '#title' => t('Adapt URI for use of thumbnail'),
+      '#description' => t('Modify the URI to reference the thumbnail image'),
```
quicksketch: An example would be helpful here. I don't understand what this means.

core/modules/filter/views/views_handler_field_file_uri.inc
```diff
@@ -0,0 +1,52 @@
...
+    $options = parent::option_definition();
+    $options['file_download_path'] = array('default' => FALSE, 'bool' => TRUE);
+    $options['image_thumbnail'] = array('default' => FALSE, 'bool' => TRUE);
+    return $options;
+  }
+
+  function options_form(&$form, &$form_state) {
+    $form['file_download_path'] = array(
+      '#title' => t('Display download path instead of file storage URI'),
+      '#description' => t('This will provide the full download URL rather than the internal filestream address.'),
```
quicksketch: I'm not sure "filestream address" is accurate here. Technically it is a "stream wrapper address", but I'm not sure I'd expose language like that to the user. Just an example might be better.

quicksketch (changes requested):

core/modules/filter/css/filter.css
Graham-72: I cannot see how otherwise to override the hover styles.

core/modules/filter/filter.pages.inc
Graham-72: I suppose perhaps because I never link to external images I forgot all about this possible use.

At an early stage in producing this PR I did try using a third option but I couldn't get the toggle action to work satisfactorily. The script does strange things. Is there a way we could allow a user to enter an external URL in the field and then recognise that it is external and then override further modification of it? Personally I like the simplicity of the two option approach.

core/modules/filter/filter.pages.inc
Graham-72: Not yet done.

core/modules/filter/filter.pages.inc
Graham-72: As I ask above, is there a way we can recognise that it is an external URL and bypass this?

core/modules/filter/css/filter.css
Graham-72: Done.

core/modules/filter/filter.admin.inc
Graham-72: I have moved it to the view and used a select list there. My main concern is that the site builder should use a style that exists in the site, either a default one such as 'thumbnail' or a custom one that has been added.

core/modules/filter/filter.install
Graham-72: OK. I have copied the entire view to an update. Thanks for the example.

core/modules/filter/filter.module
Graham-72: I have now removed this from the Filter module and modified the views handler in the File module. As this is a modification that is only relevant to image files, I am wondering whether I should make bigger changes, but I am very much a beginner with views handlers.

core/modules/filter/filter.theme.inc
Graham-72: Understood. I have changed the class names as you suggest.

core/modules/filter/js/filter.js
Graham-72: Done.

core/modules/filter/js/filter.js
Graham-72: Done.

core/modules/filter/views/views_handler_field_file_uri.inc
Graham-72: I have simplified the language and hopefully a site-builder will understand, if they are competent to use the views UI.

core/modules/filter/views/views_handler_field_file_uri.inc
Graham-72: I have modified the UI and the wording which hopefully now makes it clear that the URL is being modified to link to the selected image style files.

core/modules/filter/views/views_handler_field_file_uri.inc
Graham-72: Done.

core/modules/filter/filter.pages.inc
olafgrabienski: > is there a way we could allow a user to enter an external URL in the field and then recognise that it is external (...) ?

The insert link functionality in CKEditor allows external URLs and internal paths.

Graham-72: @olafgrabienski thank you for this reminder. We addressed this in https://github.com/backdrop/backdrop-issues/issues/3083 only 3 months ago and I had forgotten!

Graham-72:
Closing to get a fresh sandbox.

Graham-72:
Re-opening.

Graham-72:
I have problems in the sandbox with the latest revisions and I am re-checking the PR.

core/modules/filter/filter.pages.inc
Graham-72: I have now added a small piece of code to make use of some that already exists, and thereby make an exception for external links to avoid modifying the supplied URL.

core/modules/filter/filter.pages.inc
Graham-72: Now fixed.

core/modules/filter/filter.pages.inc
Graham-72: Now fixed as suggested, thank you.

core/modules/filter/css/filter.css
Graham-72: Now fixed by using a more specific style specification.

quicksketch (approved):

quicksketch:
I rebased this entire PR against the latest 1.x. I think I got the AJAX problems resolved in that latest commit.

herbdool:
It seems to be inserting the thumbnail image into the text body. Maybe I missed an option to choose a different size?

herbdool:
That's better. Images are now using the data attribute and aren't just thumbnails.

quicksketch:
New sandbox.

jenlampton:
I like it. Huge improvement!
The PR is working for me locally (after running update.php to import the view). Two things that are a little weird (and these can be changed after, if necessary) are:
1) the modal is huuuuuge on my site - which only has 2 images. Eventually most sites will have lots of images so this will be fine, but as a first-time experience it's a bit odd. Having the modal size match what'll be displayed within would be ideal - though maybe quite difficult. 
2) the order of operations is 1) select something from the right hand side, and then b) refine by entering information on the left hand side. I'd prefer these were reversed and went left-to-right the same way we read (well, most of us). Maybe we can flip it for LTR readers? Anyway - a follow up!

I gave the code a quick review too, and all looks good to me!

quicksketch:
Aaaand we're done! Merged into 1.x for 1.11.0. Woot!

Graham-72:
@quicksketch very many thanks for your very impressive work on this.

PR #2199, opened 2018-05-19
Issue #3077: Adding scheduled or unpublished notification.

serundeputy:


core/modules/node/node.theme.inc
```diff
@@ -161,6 +161,32 @@ function theme_node_admin_overview($variables) {
  * @see node.tpl.php
  */
 function template_preprocess_node(&$variables) {
+  if ($variables['elements']['#node']->status == NODE_NOT_PUBLISHED) {
```
quicksketch, line 164: I feel like this should continue to be in basis_preprocess_node(), as it contains specific markup and themes need to style it separately.

core/modules/node/node.theme.inc
```diff
@@ -161,6 +161,32 @@ function theme_node_admin_overview($variables) {
...
+    $name = node_type_get_name($variables['elements']['#node']->type);
+    if ($variables['elements']['#node']->scheduled == 0) {
+      $markup = '<div class="unpublished-indicator">' . t('This @type is unpublished.', array('@type' => $name)) . '</div>';
+    }
+    else {
+      $nice_scheduled_date = format_date(
+        $variables['elements']['#node']->scheduled,
+        'custom',
+        t('j M Y g:i A', array(), array('context' => 'php date format'))
```
quicksketch, line 173: I think just using the "medium" date format might be better here.

quicksketch (changes requested):

core/modules/node/node.theme.inc
serundeputy: I feel like the functionality should be available to you regardless of the theme you use; thus why I moved it, but I'm not passionate about it.

PR #2210, opened 2018-05-25
Issue #3150: Reduce messages shown when scheduling content.

quicksketch:
Fixes https://github.com/backdrop/backdrop-issues/issues/3150.


quicksketch:
Thanks @docwilmot for reviewing in the issue. Merged into 1.x and 1.10.x.

PR #2211, opened 2018-05-26
Issue #3135: Fix version checking for dev releases and in installer.

quicksketch:
Fixes https://github.com/backdrop/backdrop-issues/issues/3135.

- Adds support for properly checking `-dev` versions (assuming dev is always the latest).
- Adds support for comparing alphas, betas, previews, and rcs against each other.
- Makes installer.module support version checking in the built-in module enabling form.
- Adds support for checking Git checkout versions.

### DIFF
--- a/core/includes/common.inc
+++ b/core/includes/common.inc
@@ -8456,7 +8456,7 @@ function backdrop_parse_dependency($dependency) {
   $parts = explode('(', $dependency, 2);
   $value['name'] = trim($parts[0]);
   if (isset($parts[1])) {
-    $value['original_version'] = ' (' . $parts[1];
+    $value['original_version'] = '(' . $parts[1];
     foreach (explode(',', $parts[1]) as $version) {
       if (preg_match("/^\s*$p_op\s*$p_core$p_major\.$p_minor/", $version, $matches)) {
         $op = !empty($matches['operation']) ? $matches['operation'] : '=';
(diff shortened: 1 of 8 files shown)


herbdool:
screenshot from 2018-05-26 21-57-32

I tested the patch with save_draft. Project installer didn't complain. I chose to manually enable it and got the above message.

herbdool:
It seems like the Backdrop version is `1.1dev` going into `_backdrop_version_compare_convert()` and outputs `1.1dev`. I can't recall where I fetched Backdrop from but it reports `1.11.x-dev` on the modules page (as per above).

UPDATE: I only see the above when I manually enable from modules page. From the installer enable page (`admin/installer/install/enable`) it interprets the version correctly.

herbdool:
@quicksketch I think I found the issue. In system.admin.inc, line 580 should be:

`if ($incompatible_version = backdrop_check_incompatibility($v, $files[$requires]->info['version'])) {`

Currently it's stripping out the dashes whereas `_backdrop_version_compare_convert()` expects the dashes:

`if ($incompatible_version = backdrop_check_incompatibility($v, str_replace(BACKDROP_CORE_COMPATIBILITY . '-', '', $files[$requires]->info['version']))) {`

quicksketch:
Thanks @herbdool, great find. I've updated the str_replace() calls to use preg_replace() instead so we only strip off the front of the version string, rather than anywhere in the entire string. Also fixed the failing `ModuleDependencyTestCase` test.

herbdool:
Looks good now in my testing.

I didn't try the git checkout of a module.

quicksketch:

PR #2214, opened 2018-05-29
Issue #3130: Check that a load callback exists before calling.

quicksketch:
Fixes https://github.com/backdrop/backdrop-issues/issues/3130.


quicksketch:
Merged into 1.x and 1.10.x.

PR #2217, opened 2018-06-01
Issue #3155: Add 'info' type for backdrop_set_message

klonos:
https://github.com/backdrop/backdrop-issues/issues/3155

### DIFF
--- a/core/includes/theme.inc
+++ b/core/includes/theme.inc
@@ -1730,24 +1730,25 @@ function theme_status_messages($variables) {
     'status' => t('Status message'),
     'error' => t('Error message'),
     'warning' => t('Warning message'),
+    'info' => t('Info message'),
   );
 
   foreach ($message_types as $type => $messages) {
     $output .= "<div class=\"messages $type\">\n";
     if (!empty($status_heading[$type])) {
-      $output .= '<h2 class="element-invisible">' . $status_heading[$type] . "</h2>\n";
+      $output .= '  <h2 class="element-invisible">' . $status_heading[$type] . "</h2>\n";
     }
     if (count($messages) > 1) {
-      $output .= " <ul>\n";
+      $output .= "  <ul>\n";
       foreach ($messages as $message) {
-        $output .= '  <li>' . $message . "</li>\n";
+        $output .= "    <li>" . $message . "</li>\n";
       }
-      $output .= " </ul>\n";
+      $output .= "  </ul>";
     }
     else {
       $output .= reset($messages);
     }
-    $output .= "</div>\n";
+    $output .= "\n</div>\n";
   }
 
   return $output;
--- a/core/modules/system/css/messages.theme.css
+++ b/core/modules/system/css/messages.theme.css
@@ -58,50 +58,50 @@
 
 .status {
   color: #234600;
-  background-color: #E9EEBC;
+  background-color: #e9eebc;
 }
 .status a {
   color: #234600;
 }
 div.status:before {
   background-image: url(../../../misc/message-24-ok.png);
-  background-color: #CFDE56;
+  background-color: #cfde56;
 }
 
 .warning {
   color: #525252;
-  background-color: #FAF5C9;
+  background-color: #faf5c9;
 }
 .warning a {
   color: #525252;
 }
 div.warning:before {
   background-image: url(../../../misc/message-24-warning.png);
-  background-color: #FCE400;
+  background-color: #fce400;
 }
 
 .error {
-  color: #9C1D0D;
-  background-color: #F9EDEC;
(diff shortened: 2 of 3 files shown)


klonos:
The test failure seems to be random ...closing and reopening to restart the tests.

core/includes/theme.inc
```diff
@@ -1730,24 +1730,25 @@ function theme_status_messages($variables) {
...
+        $output .= "    <li>" . $message . "</li>\n";
       }
-      $output .= " </ul>\n";
+      $output .= "  </ul>";
     }
     else {
       $output .= reset($messages);
     }
-    $output .= "</div>\n";
+    $output .= "\n</div>\n";
```
opi, line 1751: Nitpicking : `\n` are always printed out at the end of every line, so maybe restore this `\n` after the `</ul>` closing tag above.

klonos: This was intentional @opi ...you see, all these `\n`s and additional spaces that's going on here is in order to print the entire message `<div>` in a pretty format when one inspects the rendered code. With multiple messages of the same type, having the leading `\n` either way, renders things like so:

```
<div class="messages info">
  <h2 class="element-invisible">Info message</h2>
  <ul>
    <li>info1</li>
    <li>info2</li>
  </ul>
</div>
```

...so no problem either way. But when you have single messages, it prints them like so:

- with the `\n` after the `</ul>`:

```
<div class="messages info">
  <h2 class="element-invisible">Info message</h2>
single info message</div>
```

- with the `\n` before the `</div>`:

```
<div class="messages info">
  <h2 class="element-invisible">Info message</h2>
single info message
</div>
```

This is because moving the `\n` after the `</ul>` puts it inside the `if` statement, which is not true for single messages of the same type. This effectively makes it so that the `\n` is not printed in these cases, which makes the closing `</div>` tag print in the same line as the actual message text, which defeats the purpose of using all these `\n`s (which was to make the rendered code look pretty).

I hope that this makes sense @opi 

PS: In general, my preference would be to remove all this "witchcraft" from specific places like this, and have the entire code that is generated by Drupal/Backdrop be prettified on the fly, at a lower-level. But I guess that if possible, someone would have done it by now 😄

opi: wow, such an explanation ! And +1 for removing this tabs and line from the code. I never read raw sources anymore, browsers inspector always prettify the html now.

core/modules/system/css/messages.theme.css
```diff
@@ -58,29 +58,54 @@
...
+.info a {
+  color: #004570;
+}
+div.info:before {
+  background-image: url(../../../misc/message-24-info.png);
+  background-color: #0074bd;
+}
+
+.messages a {
+  text-decoration: underline;
```
opi, line 108: Just realized that links has no *hover* or *focus* effect. This is not nice regarding accessibility. 
The easiest could be to remove the text-decoration:
```
.messages a:hover,
.messages a:focus {
  text-decoration: none;
}
```

opi (changes requested):

core/includes/theme.inc
```diff
@@ -1730,24 +1730,25 @@ function theme_status_messages($variables) {
...
     if (!empty($status_heading[$type])) {
-      $output .= '<h2 class="element-invisible">' . $status_heading[$type] . "</h2>\n";
+      $output .= '  <h2 class="element-invisible">' . $status_heading[$type] . "</h2>\n";
     }
     if (count($messages) > 1) {
-      $output .= " <ul>\n";
+      $output .= "  <ul>\n";
       foreach ($messages as $message) {
-        $output .= '  <li>' . $message . "</li>\n";
+        $output .= "    <li>" . $message . "</li>\n";
```
jenlampton, line 1744: Looks like there's an unintentional change here, that may result in a micro-performance regression, as `''` is a smidge faster than `""` in PHP.  It would probably be better to revert this change.

core/includes/theme.inc
```diff
@@ -1730,24 +1730,25 @@ function theme_status_messages($variables) {
...
       }
-      $output .= " </ul>\n";
+      $output .= "  </ul>";
```
jenlampton, line 1746: Looks like we lost the line-break here.

core/includes/theme.inc
jenlampton: @klonos it sounds like we could add a line break to the single message then, so that it will match what happens with the multi-line message. I'd prefer not to rearrange the theme function too much, especially when that's not the point of this issue. I'm going to take a stab at a cleanup.

jenlampton:
Replaced by https://github.com/backdrop/backdrop/pull/2291

PR #792, opened 2015-03-26
Issue #824: Add six handlers to node_autoload_info form views_handler_ar...

serundeputy:
This fixes: https://github.com/backdrop/backdrop-issues/issues/824

PR #2222, opened 2018-06-07
Issue #792: Changed the Color module to update the css files when con…

hosef:
…figuration is imported, added tests for the new functionality, and made the existing tests comply with Drupal coding standards.


quicksketch:
Nice! Thanks!

PR #2223, opened 2018-06-08
#3030: Make label blank by default for custom contextual links

laryn:


quicksketch:

PR #2232, opened 2018-06-27
Issue #1582: RSS feed icon never printed

docwilmot:
Fixes https://github.com/backdrop/backdrop-issues/issues/1582

### DIFF
--- a/core/modules/node/node.module
+++ b/core/modules/node/node.module
@@ -2026,6 +2026,9 @@ function node_block_configure($delta = '', $settings = array()) {
 
 /**
  * Returns a themed RSS icon which links to a provided feed.
+ *
+ * @param array $settings
+ *   Block configuration settings.
  */
 function node_syndicate_content($settings) {
   if (isset($settings['syndicate_settings']) && $settings['syndicate_settings'] == 'page_content') {
@@ -2035,7 +2038,7 @@ function node_syndicate_content($settings) {
   return array(
     '#theme' => 'feed_icon',
     '#url' => 'rss.xml',
-    '#title' => t('Syndicate'),
+    '#title' => t('Subscribe to promoted content'),
   );
 }
 
--- a/core/modules/node/node.module
+++ b/core/modules/node/node.module
@@ -1941,7 +1941,7 @@ function node_revision_list(Node $node) {
 function node_block_info() {
   $blocks['syndicate'] = array(
     'info' => t('Syndicate'),
-    'description' => t('An RSS icon linking to the feed for the current page (if any).'),
+    'description' => t('An RSS icon linking to available feeds.'),
   );
   $blocks['recent'] = array(
     'info' => t('Recent content'),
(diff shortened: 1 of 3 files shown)


jenlampton:
Here's a crazy idea: what if we turned rss.xml into a feed display for the promoted view, and then replaced the hard-coded call to `theme_feed_icon()` with `backdrop_get_feeds()`? Would that be the same result?

core/modules/node/node.module
```diff
@@ -2014,9 +2010,35 @@ function node_block_configure($delta = '', $settings = array()) {
       '#options' => backdrop_map_assoc(array(2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12, 13, 14, 15, 16, 17, 18, 19, 20, 25, 30)),
     );
   }
+  if ($delta == 'syndicate') {
+    $form['syndicate_settings'] = array(
+      '#type' => 'radios',
+      '#title' => t('Choose the linked feed'),
+      '#default_value' => isset($settings['syndicate_settings']) ? $settings['syndicate_settings'] : 'promoted',
+      '#options' => array(
+        'promoted' => t('Promoted nodes'),
```
jenlampton: label: "Always show RSS link for promoted content"

core/modules/node/node.module
```diff
@@ -2014,9 +2010,35 @@ function node_block_configure($delta = '', $settings = array()) {
...
+        'page_content' => t('Page content'),
```
jenlampton: label: "Show RSS link for the current page, if available."

jenlampton (commented):
I'm a little confused by these labels, how about something more like this?

docwilmot:
Yes, I should have done some explaining. The original icon only showed a list of promoted nodes, which is not as useful now as when Drupal front page was just a list by default. So thought to make it more useful by allowing it to show the current page instead, especially useful if the current page is a View.

core/modules/node/node.module
docwilmot: fixed

core/modules/node/node.module
docwilmot: done

core/modules/node/node.module
```diff
@@ -2014,9 +2010,35 @@ function node_block_configure($delta = '', $settings = array()) {
...
+ */
+function node_syndicate_content($settings) {
+  if (isset($settings['syndicate_settings']) && $settings['syndicate_settings'] == 'page_content') {
+    return backdrop_get_feeds();
+  }
+  // Default to promoted content to support legacy Syndicate blocks.
+  return array(
+    '#theme' => 'feed_icon',
+    '#url' => 'rss.xml',
+    '#title' => t('Syndicate'),
```
opi: the `#title` is used to create the feed icon title attribute, then this one generate a tooltip with "Subscribe to Syndicate", maybe this PR is a good occasion to do better, something like "Subscribe to promoted content (feed)".

jenlampton: Even just "Subscribe" would be better than "Syndicate". Would anyone miss the previous language here?

quicksketch:

PR #2235, opened 2018-07-01
Argument 3 passed to comment_new_page_count() must be an instance of …

robertgarrigos:
…Node

It fixes issue https://github.com/backdrop/backdrop-issues/issues/2395


quicksketch:

PR #2244, opened 2018-07-20
Check if we can fire phpcs in ZenCI.

serundeputy:


core/misc/zen-ci/run_phpcs.sh
```diff
@@ -0,0 +1,15 @@
+#!/bin/sh
+
+##
+# Run phpcs code standards tests.
+##
+
+#set site path
```
klonos: space, capital, period

klonos:
The PR seems to also include css hex number conversions from capital to lowercase + multi-line comments conversions to docblock-style comments (even though these do not seem to be documentation comments).

Not sure if these were added intentionally for demoing purposes, but they should not be merged as part of the same PR.

core/misc/zen-ci/run_phpcs.sh
serundeputy: addressed in latest commit. ad6a5a785aa61a73c219dfcbee119adb340a2cd8

serundeputy:
@klonos https://api.backdropcms.org/php-standards#comments
and: https://api.backdropcms.org/css-standards#format

which are things that are automatically caught by `phpcs`; thus the reason we need this automation

----

Also just to note:
> docblock-style

```php
/**
 * This is  a documentation block.
 */ 
```

```php
/*
 * This is a multiline comment.
 */
```

i.e. the latter will not be picked up by the https://github.com/backdrop-contrib/api module

klonos:
Thanks @serundeputy, I have never noticed the extra asterisk in the docblock vs plain multiline comments. Still I find using `//` more subtle and less distraction than the dockblock-like equivalent. It makes them more "inline" I guess, and less "in your face" (something that is fine for actual docblocks IMO). Is there a way to have phpcs not complain about this kind of comments?

serundeputy:
Yes; but that is what is specified in our coding standards on API.backropcms.org which I've cited above.

ghost:
I'm going to close and delete this core repo branch as:
1. It should be re-created from a forked repo
2. There are merge conflicts
3. We don't use Zen.CI anymore

PR #2246, opened 2018-07-22
Issue #2094: Use a 'contrib' directory if it exists.

quicksketch:
Fixes https://github.com/backdrop/backdrop-issues/issues/2094.


quicksketch:
Merged into 1.x and 1.10.x.

PR #2253, opened 2018-07-23
Issue #2558: Update entity.api.php with entity class info

alexfinnarn:


core/modules/entity/entity.api.php
```diff
@@ -20,6 +20,8 @@
  *   An array whose keys are entity type names and whose values identify
  *   properties of those types that the system needs to know about:
  *   - label: The human-readable name of the type.
+ *   - entity class: A class the controller will use for instantiating entities.
```
klonos: This would be easier to read/understand for non-English speakers if it said "A class *that* the controller will use...", or "A class which will be used by the controller to..."

alexfinnarn: I agree. It is probably better grammar English speakers too :)

quicksketch:
Thanks! Merged into 1.x and 1.10.x!

PR #2259, opened 2018-07-31
Fixes issue #3200: Allows limiting of a single image dimension.

Graham-72:
This may require changes to tests.


core/includes/file.inc
```diff
@@ -1807,12 +1807,22 @@ function file_validate_image_resolution(File $file, $maximum_dimensions = 0, $mi
       // Check that it is smaller than the given dimensions.
       list($width, $height) = explode('x', $maximum_dimensions);
       if ($info['width'] > $width || $info['height'] > $height) {
+        if ($width == 0) {$width = NULL;}
```
quicksketch: Single-line `if` statements like this don't match our coding standards, but you could still use the ternary operator like `$width = empty($width) ? NULL : $width`.

core/modules/filter/filter.pages.inc
```diff
@@ -257,8 +257,8 @@ function filter_format_editor_image_form($form, &$form_state, $format) {
     '#title' => t('Width'),
     '#title_display' => 'attribute',
     '#type' => 'number',
-    '#default_value' => isset($values['width']) ? $values['width'] : NULL,
-    '#min' => 1,
+    '#default_value' => isset($values['width']) ? $values['width'] : 0,
+    '#min' => 0,
```
quicksketch: The default value being `0` when the field is actually empty is not what I would expect here. If it's optional, it should just be empty.

core/modules/image/image.field.inc
```diff
@@ -86,14 +86,18 @@ function image_field_instance_settings_form($field, $instance) {
 
   // Add maximum and minimum resolution settings.
   $max_dimensions = explode('x', $settings['max_dimensions']) + array('', '');
+  if ($max_dimensions[0] === 0) {$max_dimensions[0] = NULL;}
+  if ($max_dimensions[1] === 0) {$max_dimensions[1] = NULL;}
```
quicksketch: Use ternary operator instead of single-line `if` statements.

core/modules/image/image.field.inc
```diff
@@ -86,14 +86,18 @@ function image_field_instance_settings_form($field, $instance) {
...
     '#type' => 'item',
     '#title' => t('Maximum image dimensions'),
     '#element_validate' => array('_image_field_resolution_validate'),
     '#weight' => 4.1,
     '#field_prefix' => '<div class="container-inline">',
     '#field_suffix' => '</div>',
-    '#description' => t('The maximum allowed image dimensions expressed as WIDTHxHEIGHT. Leave blank for no restriction. If a larger image is uploaded, it will be resized to reflect the given width and height. Resizing images on upload will cause the loss of <a href="http://en.wikipedia.org/wiki/Exchangeable_image_file_format">EXIF data</a> in the image.'),
+    '#description' => t('The maximum allowed image dimensions expressed as WIDTHxHEIGHT. Leave blank or 0 for no restriction.
+     <br> If a larger image is uploaded, it will be resized to reflect the given width and height. 
+     <br>Resizing images on upload will cause the loss of <a href="http://en.wikipedia.org/wiki/Exchangeable_image_file_format">EXIF data</a> in the image.'),
```
quicksketch: Including white space (new lines and tabs) in a translated string makes translation more difficult. We should either split up this string into multiple `t()` calls (one per sentence) or leave it all on one line as it had been.

I think we should probably just leave this string alone, and keep the blank field as the default value. We can still accept 0, but we don't need to mention it.

quicksketch (changes requested):

core/modules/image/image.field.inc
Graham-72: Done. Incidentally, I could not find this mentioned in our coding standards.

core/modules/filter/filter.pages.inc
Graham-72: Done.

core/includes/file.inc
Graham-72: Done.

core/modules/image/image.field.inc
Graham-72: Reverted to existing version. Thanks for your review and suggestions.

core/includes/file.inc
```diff
@@ -1807,12 +1807,22 @@ function file_validate_image_resolution(File $file, $maximum_dimensions = 0, $mi
...
           image_save($image);
           $file->filesize = $image->info['file_size'];
+          if (!$height) {
+            backdrop_set_message(t('The image was resized to fit within the maximum allowed width of %width pixels.', array('%width' => $width)));
+          }
+          else if (!$width) {
+            backdrop_set_message(t('The image was resized to fit within the maximum allowed height of %height pixels.', array('%height' => $height)));
+          }
+          else {
           backdrop_set_message(t('The image was resized to fit within the maximum allowed dimensions of %dimensions pixels.', array('%dimensions' => $maximum_dimensions)));
```
quicksketch: Missing indentation here.

core/modules/file/file.theme.inc
```diff
@@ -291,7 +291,10 @@ function theme_file_upload_help($variables) {
       $descriptions[] = t('Images must be larger than !min pixels.', array('!min' => '<strong>' . $min . '</strong>'));
     }
     elseif ($max) {
-      $descriptions[] = t('Images larger than !max pixels will be scaled down.', array('!max' => '<strong>' . $max . '</strong>'));
+      list($width, $height) = explode('x', $max);
+      if ($height === '0') {$descriptions[] = t('Images with width greater than !width pixels will be scaled down.', array('!width' => '<strong>' . $width . '</strong>'));}
+      else if ($width === '0') {$descriptions[] = t('Images with height greater than !height pixels will be scaled down.', array('!height' => '<strong>' . $height . '</strong>'));}
+      else {$descriptions[] = t('Images larger than !max pixels will be scaled down.', array('!max' => '<strong>' . $max . '</strong>'));}
```
quicksketch: These should be split onto separate lines per coding standards. We can also use `elseif` instead of `else if`.

We could make these messages slightly simpler language by saying "Images wider than ..." and "Images taller than..." instead of "Images with width greater than..."

core/modules/filter/filter.pages.inc
```diff
@@ -257,8 +257,7 @@ function filter_format_editor_image_form($form, &$form_state, $format) {
     '#title' => t('Width'),
     '#title_display' => 'attribute',
     '#type' => 'number',
-    '#default_value' => isset($values['width']) ? $values['width'] : NULL,
-    '#min' => 1,
+    '#default_value' => isset($values['width']) ? $values['width'] : '',
```
quicksketch: If the default value is "0" it should just be wiped out and displayed as an empty string. So replace `isset()` with `!empty()`

core/modules/filter/filter.pages.inc
```diff
@@ -268,8 +267,7 @@ function filter_format_editor_image_form($form, &$form_state, $format) {
     '#title' => t('Height'),
     '#title_display' => 'attribute',
     '#type' => 'number',
-    '#default_value' => isset($values['height']) ? $values['height'] : NULL,
-    '#min' => 1,
+    '#default_value' => isset($values['height']) ? $values['height'] : '',
```
quicksketch: Same here, replace `isset()` with `!empty()`.

quicksketch (changes requested):
A few more changes requested in the review. We should also update `ImageFieldValidateTestCase` to include these new scenarios.

core/includes/file.inc
Graham-72: Fixed.

core/modules/file/file.theme.inc
Graham-72: Done.

core/modules/filter/filter.pages.inc
Graham-72: Done.

core/modules/filter/filter.pages.inc
Graham-72: Done.

Graham-72:
> We should also update ImageFieldValidateTestCase to include these new scenarios.

Test case is now updated. Some other issues have been identified and fixed.

quicksketch (approved):

quicksketch:
Rebased to fix a conflict. If all tests pass, this looks good to me!

PR #2207, opened 2018-05-24
Issue #2302: Change the Name lable to Username in the user listing.

quicksketch:
Fixes https://github.com/backdrop/backdrop-issues/issues/2302

Replaces https://github.com/backdrop/backdrop/pull/1628.

PR #2262, opened 2018-08-03
Issue #2284: Default value for update threshold setting should be 'all'.

opi:


core/modules/update/update.install
```diff
@@ -231,3 +231,14 @@ function update_update_1003() {
  * @} End of "addtogroup updates-7.x-to-1.x"
  * The next series of updates should start at 2000.
  */
+
+/**
+ * Fix wrong default value for update_threshold setting.
+ */
+function update_update_2000() {
```
klonos: @opi `2000` update hook numbers are reserved for Backdrop 2.x. This should be 1004 instead and added before the `@} End of "addtogroup updates-7.x-to-1.x"` docblock. See https://github.com/backdrop/backdrop/pull/2207/files#diff-5050e96d3115d990e2ff134b2f9ac876R1184 for example.

opi: damn right ! fixed.

quicksketch:
Thanks merged into 1.10.x and 1.x!

PR #2263, opened 2018-08-04
Issue #3237: Wrong radio(s) checked when having #default_value to 0 and having at least one alpha key

klonos:
[does not include tests]


quicksketch:
Replaced by https://github.com/backdrop/backdrop/pull/2267, which adds tests.

PR #2267, opened 2018-08-06
Issue #3237:  Wrong radio(s) checked when having #default_value to 0 and having at least one alpha key

alexfinnarn:


quicksketch:
Thanks @alexfinnarn! The only thing I'd request here is that we move the tests to be with other FormAPI tests. We have `FormsTestCase` in `modules/simpletest/tests/form.test`, in which these tests seem appropriate. Likewise the test form can be added in the existing `form_test.module`, rather than making a new one.

alexfinnarn:
Sorry, I didn't know where any form tests were after quickly trying to search for them. So `modules/simpletest/tests/` contains all the tests for functions that exist in `includes/`? I would first look for `includes/tests/form.test` myself. Probably not worth changing files around, but I had trouble finding the test file. I can change where the code lives and delete those files.

PR #2269, opened 2018-08-06
Issue #3244: Prevent PHP fatals in Views UI on PHP 5.6.

### LINKED ISSUE #3244: Fatal error: [] operator not supported for strings on PHP 5.6

I'm seeing PHP fatals in my logs related to Views UI
```
PHP Fatal error:  [] operator not supported for strings in /backdrop/core/modules/views_ui/views_ui.admin.inc on line 2302
```

I've found a similar issue for Drupal 7 on PHP 5.6:
Related: https://www.drupal.org/project/views/issues/2767167#comment-12719737

Steps to reproduce:

1.  Create a new content view with aggregation enabled and save it.
1. Add a new display and don't save the view.
1. Try to add a field to the new display.
1. During step 3, after selecting the "Aggregation type" and clicking "Apply and continue", an AJAX error will be displayed.
1. Watchdog also records a bunch of PHP warnings 

The problem seems to stem from the `call_user_func_array()` not passing the $view object by reference to `views_ui_build_form_state()`. 

### Relevant version/system information (if applicable)

This only seems to be a problem with PHP 5.6; it works fine on 5.5 and 5.4.

Related: https://github.com/backdrop/backdrop-issues/issues/973

jenlampton:
Fixes https://github.com/backdrop/backdrop-issues/issues/3244

### DIFF
--- a/core/modules/views_ui/views_ui.admin.inc
+++ b/core/modules/views_ui/views_ui.admin.inc
@@ -2290,10 +2290,27 @@ function views_ui_edit_form_get_bucket($type, $view, $display) {
 
 /**
  * Regenerate the current tab for AJAX updates.
+ *
+ * @param view $view
+ *   The view object to work on.
+ * @param array $output
+ *   The output array.
+ * @param string $display_id
+ *   The ID of the display in the $view that is being viewed.
  */
 function views_ui_regenerate_tab(&$view, &$output, $display_id) {
   if (!$view->set_display('default')) {
     return;
+  }
+
+  // Make sure $output is an array.
+  if (!is_array($output)) {
+    if ($output === NULL || $output === '') {
+      $output = array();
+    }
+    else {
+      $output = array($output);
+    }
   }
 
   // Regenerate the main display area.


core/modules/views_ui/views_ui.admin.inc
```diff
@@ -2829,7 +2846,7 @@ function views_ui_ajax_form($js, $key, $view, $display_id = '') {
   array_splice($args, 0, 4);
 
   $form_state = views_ui_build_form_state($js, $key, $view, $display_id, $args);
-  // check to see if this is the top form of the stack. If it is, pop
+  // Check to see if this is the top form of the stack. If it is, pop
```
klonos: You seem to have been fixing some of the comments in this file, but this one seems to have some weird wrapping. Looks like it could fit in two lines if properly wrapped.

Having said that, I would leave the comment fixing out of this PR. We will handle them in bulk as part of https://github.com/backdrop/backdrop-issues/issues/3213

core/modules/views_ui/views_ui.admin.inc
```diff
@@ -2290,12 +2290,29 @@ function views_ui_edit_form_get_bucket($type, $view, $display) {
...
+ *
+ * @param view $view
+ *   The view object to work on.
+ * @param array $output
+ *   The output array.
+ * @param string $display_id
+ *   The ID of the display in the $view that is being viewed.
  */
-function views_ui_regenerate_tab(&$view, &$output, $display_id) {
+function views_ui_regenerate_tab(&$view, array &$output, $display_id) {
```
quicksketch: This code hinting will cause non-array parameters to throw a fatal error (and would make the loose checking below never executed).

quicksketch (changes requested):

core/modules/views_ui/views_ui.admin.inc
jenlampton: done.

core/modules/views_ui/views_ui.admin.inc
jenlampton: removed.

quicksketch (approved):

PR #2280, opened 2018-08-14
Issue #3213: Run phpcs only on the changed files.

serundeputy:


quicksketch:
Let's give this a shot, at this point we're just adding the testing script. We're not actually executing it until https://github.com/backdrop-ops/backdropcms.org/pull/460 is completed.

PR #2283, opened 2018-08-17
Issue #2570: Remove ViewsTestCase and ViewsSqlTest entries that point to the wrong file from the autoloader.

jlfranklin:
Addresses backdrop/backdrop-issues#2530 per @quicksketch 's [comment](https://github.com/backdrop/backdrop-issues/issues/2530#issuecomment-297588133).


quicksketch:
Thanks @jlfranklin! Merged into 1.x and 1.10.x.

PR #2284, opened 2018-08-17
Issue #3256: Override of layout display renderer should work on existing paths.

quicksketch:
Fixes https://github.com/backdrop/backdrop-issues/issues/3256.


docwilmot (approved):
Yep, we missed that. Makes sense.

quicksketch:

PR #2288, opened 2018-08-30
Issue #3141: Add display mode to page title on manage display pages.

jenlampton:
Fixes https://github.com/backdrop/backdrop-issues/issues/3141


quicksketch:
Looks great. Though I don't like putting backdrop_set_title() in form definitions, this isn't likely to be used outside of a page. Merged into 1.x and 1.10.x. Thanks!

PR #2299, opened 2018-09-08
Issue #3209: Adding default color schemes for Basis.

oadaeh:
This adds a few default color schemes to Basis for https://github.com/backdrop/backdrop-issues/issues/3209


herbdool:
Looks like DNS isn't working? Maybe close/open the PR and see if it helps.

herbdool:
Overall looks ok. The mobile menu link on Metro is not visible. Black on black I think

oadaeh:
It's using the color for Borders. The same is true for Tropical. It is also barely visible for Ice, and slightly more visible for Goldfish.

oadaeh:
I tweaked the Borders color for those four schemes, so that both the menu item and the borders (seen under Drop) are (more) visible.

herbdool:
The menus look better, thanks @oadaeh.

Another odd one is Tropical:

screenshot from 2018-09-09 23-16-17

The tab bar has a blended background that doesn't look quite tropical. Is there an easy tweak for this?

oadaeh:
That is being color blended. That area on Goldfish is slightly off, too.
We would need to add another color option to the scheme, so that it could be set.

herbdool:
I think it would be worth adding, though not sure if it can make it into 1.11.0. If it does make it in then we have to update this PR with the new color option.

herbdool:
@oadaeh I added a PR with the primary tab background color option https://github.com/backdrop/backdrop/pull/2303

oadaeh:
@herbdool I'll try to take a look at it this evening. It should be more successful than the last two evenings.

oadaeh:
Actually, maybe @hosef can take a look, too.

herbdool:
@oadaeh looks great! One small change: the hex should be lower case.

jenlampton:
My color review:
* Basis - love it
* ice - like it - potentially not accessible
* plum - love ot
* goldfish - potentially not accessible
* mocha - like it, maybe too much red? (swap header bg color w/ logo bg color?)
* metro - love it
* polar - love it (but isn't this the same as Basis?)
* tropical - like it - potentially not accessible

hosef:
I took a look at modifying the colors to be more accessible, however I ran into many issues where 2 or more elements were using the same color, and I couldn't make it accessible in both cases.

I think what we need to do is add more settings, but with the current interface, that would be very messy. This could be addressed by getting backdrop/backdrop-issues/issues/3257 done.

jenlampton:
> I ran into many issues where 2 or more elements were using the same color, and I couldn't make it accessible in both cases.

Does this mean we shouldn't have been using the same color for both those elements?

PR #2304, opened 2018-09-13
Issue #3260: Remove module_load_include() call to file the call is in.

jenlampton:
Fixes https://github.com/backdrop/backdrop-issues/issues/3260


serundeputy:
looks good. works good. thanks @jenlampton

PR #2305, opened 2018-09-15
Issue #3208: Deprecate Authorize.php.

quicksketch:
Initial PR for https://github.com/backdrop/backdrop-issues/issues/3208.


core/modules/installer/installer.browser.inc
```diff
@@ -324,7 +324,10 @@ function _installer_browser_batch_install_release($release_name, $project, &$con
 
   module_load_include('inc', 'installer', 'installer.manager');
   $url = file_create_url($release['download_link']);
-  $result = installer_manager_download_project($url, TRUE);
+  $result = installer_manager_download_project($url);
+  if ($result['success']) {
```
docwilmot, line 328: Why is this separate from the same conditional `if ($result['success'])` just below this? Can this be combined?

PR #414, opened 2014-08-12
Update admin_menu so it doesn't use hook_page_alter().

jenlampton:
first attempt at a PR for https://github.com/backdrop/backdrop-issues/issues/294

PR #2307, opened 2018-09-21
Issue #414: Automatic core updates.

serundeputy:


core/modules/installer/installer.manager.inc
```diff
@@ -273,6 +273,8 @@ function installer_manager_update_form($form, $form_state, $context) {
   if (!empty($projects['manual'])) {
     $prefix = '<h2>' . t('Manual updates required') . '</h2>';
     $prefix .= '<p>' . t('Updates of Backdrop CMS core using the web interface are disabled.') . '</p>';
+    $prefix .= '<p>' . t('To enable Backdrop CMS core updates change update settings: ') .
```
jenlampton, line 276: Is this for automatic updates? If so, shouldn't we say so?

serundeputy: no; this is just to enable updates via UI.

PR #2308, opened 2018-09-22
Issue #3056: added permission to view and use "Clear log messages" form.

### LINKED ISSUE #3056: Add permissions to view and use button "Clear log messages"
Inappropriate user can accidentally use this button. On multi user site this can cause administration related problem.

### Expected behavior
 Only administrator level (or user with special role) can see and use this button.

afreemannc:
Resolves https://github.com/backdrop/backdrop-issues/issues/3056

### DIFF
--- a/core/modules/dblog/dblog.admin.inc
+++ b/core/modules/dblog/dblog.admin.inc
@@ -30,7 +30,10 @@ function dblog_overview() {
     WATCHDOG_DEPRECATED => 'dblog-deprecated',
   );
 
-  $build['dblog_clear_log_form'] = backdrop_get_form('dblog_clear_log_form');
+  if (user_access('clear log messages')) {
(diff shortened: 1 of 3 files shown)


core/modules/dblog/dblog.admin.inc
```diff
@@ -30,7 +30,10 @@ function dblog_overview() {
     WATCHDOG_DEPRECATED => 'dblog-deprecated',
   );
 
-  $build['dblog_clear_log_form'] = backdrop_get_form('dblog_clear_log_form');
+  if (user_access('clear log messages')) {
```
opi, line 33: Maybe the check would be better on the `dblog_clear_log_form` form itself (like an `#access` key on the fieldset), in case this form is called somewhere else.

PR #2320, opened 2018-10-13
Issue #2583: [UX] Revamp the maintenance mode page

### LINKED ISSUE #2583: [UX] Revamp the maintenance mode page
I usually only ever visit this page if my site has been left in maintenance mode accidentally, and I need to disable that setting. In that case, I have clicked a link that says "Go online" and I expect by clicking that link, I will go online. Instead, I am taken to a form, where I usually blindly hit "Save configuration" since I can't tell that I need to UNcheck the box for `Put site into maintenance mode`.

A checkbox with a label that's an action is not the right tool here. The says "Put..." and that's not something my brain things can be toggled. An action has a yes/no answer, not an on/off value. 

What if the label were simply `Maintenance mode` and the form element was a set of radio buttons: `On` or `Off`?

Also, we should do a better job of letting people know the site is in maintenance mode:
1) how about a message at the top of this page saying my site is in maintenance mode
2) how about a message in the status report, saying my site is in maintenance mode

klonos:
Fixes https://github.com/backdrop/backdrop-issues/issues/2583

### DIFF
--- a/core/modules/system/system.admin.inc
+++ b/core/modules/system/system.admin.inc
@@ -2003,23 +2003,23 @@ function system_site_maintenance_mode($form, &$form_state) {
     '#title' => t('Maintenance mode'),
     '#default_value' => $maintenance_mode,
     '#options' => array(
-      '1' => t('On'),
       '0' => t('Off'),
+      '1' => t('On'),
     ),
   );
-  $form['maintenance_mode']['1']['#description'] = t('Visitors will only see the maintenance mode message.');
   $form['maintenance_mode']['0']['#description'] = t('Visitors will see the site as per usual.');
+  $form['maintenance_mode']['1']['#description'] = t('Visitors will only see the maintenance mode message.');
 
   $form['maintenance_mode_message'] = array(
     '#type' => 'textarea',
-    '#title' => t('Message to display when in maintenance mode'),
+    '#title' => t('Maintenance mode message'),
     '#default_value' => $config->get('maintenance_mode_message'),
     '#description' => theme('token_tree_link'),
     '#element_validate' => array('token_element_validate'),
   );
 
   if ($maintenance_mode) {
-    backdrop_set_message(t('<strong>The site is currently in maintenance mode</strong>. Visitors will only see the maintenance mode message.'), 'warning');
+    backdrop_set_message(t('<strong>The site is currently in maintenance mode</strong>. Visitors will only see the maintenance mode message (as configured below).'), 'warning');
 
     $form['permissions_message'] = array(
       '#type' => 'container',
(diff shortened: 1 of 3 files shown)


core/modules/system/system.admin.inc
```diff
@@ -2010,32 +2019,130 @@ function system_site_maintenance_mode($form, &$form_state) {
...
+  // Retrieve all system-related permissions.
+  $systmem_permissions = system_permission();
+
+  // We currently need only the "Use the site in maintenance mode" permission.
+  $permissions['access site in maintenance mode'] = $systmem_permissions['access site in maintenance mode'];
+
+  // Render role/permission overview:
+  $options = array();
+  $status = array();
+  foreach ($permissions as $permisson => $permission_item) {
```
amoutiers: There's a typo here, it's permission not permisson

```suggestion
  foreach ($permissions as $permission => $permission_item) {
```

amoutiers (changes requested):

klonos:
...refreshing the PR sandbox.

core/modules/system/system.admin.inc
```diff
@@ -1996,46 +1996,162 @@ function system_regional_settings_submit($form, &$form_state) {
...
-    '#description' => t('Visitors will only see the maintenance mode message. Only users with the "Access site in maintenance mode" <a href="@permissions-url">permission</a> will be able to access the site. Authorized users can log in directly via the <a href="@user-login">user login</a> page.', array('@permissions-url' => url('admin/config/people/permissions'), '@user-login' => url('user'))),
+    '#type' => 'radios',
+    '#title' => t('Maintenance mode'),
+    '#default_value' => $maintenance_mode,
+    '#options' => array(
+      '0' => t('Off'),
+      '1' => t('On'),
+    ),
   );
+  $form['maintenance_mode']['0']['#description'] = t('Visitors will see the site as per usual.');
```
jenlampton: remove "per"

core/modules/system/system.admin.inc
```diff
@@ -1996,46 +1996,162 @@ function system_regional_settings_submit($form, &$form_state) {
...
+
+    $permissions_message_type = 'warning';
+  }
+
+  $form['permissions_message'] = array(
+    '#type' => 'container',
+    '#attributes' => array('class' => array('messages', $permissions_message_type)),
+  );
+  $form['permissions_message']['info'] = array(
+    '#markup' => t('While in maintenance mode, authorized users can still log in directly via the <a href="@user-login">user login</a> page, but they need to have the "Use site in maintenance mode" permission in order to be able to access the site.', array('@user-login' => url('user'))),
```
jenlampton: This could use to be less verbose, and use the word "user" less...  how about: 
`People with the "Use site in maintenance mode" permission will still be able to log in while the site is in maintenance mode.`

Also, if we are going to use a link here, we should probably use `/user/login` over `/user`.

klonos: Thanks. Done 😉

klonos:
...time to refresh the sandbox

PR #2324, opened 2018-10-17
Issue #3316: [UX] Add human name to fields list at admin/reports/fields

klonos:
https://github.com/backdrop/backdrop-issues/issues/3316

### DIFF
--- a/core/modules/field_ui/field_ui.admin.inc
+++ b/core/modules/field_ui/field_ui.admin.inc
@@ -66,7 +66,8 @@ function field_ui_fields_list() {
     }
   }
   foreach ($rows as $field_name => $cell) {
-    // Render the items of the "Used in" column as a list.
+    // A reused field can have different labels across the various bundles it is
+    // being used in. So we render the items of the "Used as" column as a list.
     $rows[$field_name]['data'][3] = theme('item_list', array('items' => $rows[$field_name]['data'][3]));
   }
 
--- a/core/modules/field_ui/field_ui.admin.inc
+++ b/core/modules/field_ui/field_ui.admin.inc
@@ -32,7 +32,7 @@ function field_ui_fields_list() {
     t('Field name'),
     array('data' => t('Field type'), 'class' => array(RESPONSIVE_PRIORITY_MEDIUM)),
     array('data' => t('Module'), 'class' => array(RESPONSIVE_PRIORITY_LOW)),
-    t('Used in'),
+    t('Used as'),
   );
 
   $rows = array();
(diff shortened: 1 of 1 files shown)


core/modules/field_ui/field_ui.admin.inc
```diff
@@ -45,26 +46,39 @@ function field_ui_fields_list() {
...
+          $rows[$field_name]['data'][1] = $field_types[$field['type']]['label'];
+          $rows[$field_name]['data'][2] = $modules[$module_name]->info['name'];
         }
 
         // Add the current instance.
         $admin_path = _field_ui_bundle_admin_path($entity_type, $bundle);
-        $rows[$field_name]['data'][2][] = $admin_path ? l($bundles[$entity_type][$bundle]['label'], $admin_path . '/fields') : $bundles[$entity_type][$bundle]['label'];
+        $field_label = $instances[$entity_type][$bundle][$field_name]['label'];
+        $bundle_label = $bundles[$entity_type][$bundle]['label'];
+        if ($admin_path) {
```
klonos, line 56: This is to account for the previous `$rows[$field_name]['data'][2][] = $admin_path ? ...` conditional that would render the bundle name/label as a link to the "Manage fields" page of that bundle, or render it as plain text.

Not sure which cases that was there to cover. Was it permissions (current used not allowed to edit the bundle)? ...or are there cases of bundles/fields (perhaps in contrib if not in core?) that cannot be configured? Anyway, I though that there must have been a reason, so I have retained it. If there is no reason, then we could remove the conditional, and have the field label and bundle label be rendered as links in all cases.

core/modules/field_ui/field_ui.admin.inc
```diff
@@ -45,26 +46,39 @@ function field_ui_fields_list() {
...
+        else {
+          $rows[$field_name]['data'][3][] = t('%field_label in @bundle_label', array('%field_label' => $field_label, '@bundle_label' => $bundle_label));
+        }
       }
     }
   }
   foreach ($rows as $field_name => $cell) {
-    $rows[$field_name]['data'][2] = implode(', ', $cell['data'][2]);
+    // Render the items of the "Used in" column as a list.
+    $rows[$field_name]['data'][3] = theme('item_list', array('items' => $rows[$field_name]['data'][3]));
```
klonos: As explained in the issue, a reused filed can have different label across the various bundles it is being used in.

klonos:
Time for a sandbox refresh 😄

core/modules/field_ui/field_ui.admin.inc
```diff
@@ -57,28 +56,22 @@ function field_ui_fields_list() {
...
 
-  if (empty($rows)) {
-    $output = t('No fields have been defined yet.');
-  }
-  else {
-    // Sort rows by field name.
-    ksort($rows);
-    $output = theme('table', array('header' => $header, 'rows' => $rows, 'empty' => t('No fields available.'),));
-  }
+  $output = theme('table', array('header' => $header, 'rows' => $rows, 'empty' => t('No fields have been defined yet.'),));
```
opi, line 74: get sort+order with `tablesort_get_sort` & `tablesort_get_order` and the properly sort the `$rows` array before rendering as table.

klonos:
Closing in favor of https://github.com/backdrop/backdrop/pull/2528

PR #2327, opened 2018-10-19
Issue #3325: Use WATCHDOG_* contants instead of numerical value.

opi:
Disabled values are commented out for a better comprehension of the list. Please tell me if it should be removed.


klonos (approved):
👍

quicksketch:
Simple change, merged into 1.x and 1.11.x. Thanks @opi!

PR #2340, opened 2018-10-27
Issue #3345: [UX] Layouts: render the "Configure" link only if the block is not broken.

klonos:
https://github.com/backdrop/backdrop-issues/issues/3345

### DIFF
--- a/core/modules/layout/plugins/renderers/layout_renderer_editor.inc
+++ b/core/modules/layout/plugins/renderers/layout_renderer_editor.inc
@@ -205,7 +205,7 @@ class LayoutRendererEditor extends LayoutRendererStandard {
     $links = array();
 
     // Only render the "Configure" link if the block is not broken.
-    if (!$block->isBroken()) {
+    if (!is_a($block, 'BlockBroken')) {
       $links['configure'] = array(
         'title' => t('Configure'),
         'href' => $this->getUrl('configure-block', $block->uuid),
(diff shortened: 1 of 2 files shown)


core/modules/layout/plugins/renderers/layout_renderer_editor.inc
```diff
@@ -203,15 +203,19 @@ class LayoutRendererEditor extends LayoutRendererStandard {
...
-      'href' => $this->getUrl('configure-block', $block->uuid),
-      'attributes' => array(
-        'class' => array('use-ajax'),
-        'data-dialog' => TRUE,
-        'data-dialog-options' => json_encode(array('dialogClass' => 'layout-dialog')),
-      ),
-    );
+
+    // Only render the "Configure" link if the block is not broken.
+    if (!$block->isBroken()) {
```
quicksketch: Instead of a new method, we could simply check the class type and replace this with `if (!is_a($block, 'BlockBroken'))`

quicksketch (changes requested):

core/modules/layout/plugins/renderers/layout_renderer_editor.inc
opi: nitpick :  `$block instanceof BlockBroken` is slightly faster (and I personally prefer this syntax, but it's a matter of personal taste).

klonos: PR updated.

klonos:
...sandbox refresh

klonos:
...refreshing sandbox.

quicksketch (approved):

quicksketch:

PR #2349, opened 2018-11-01
Issue #3352: Allow multiple default values

hosef:
Modifies the field settings form to use the multi value version for the default value field if the cardinality of the field is above 1.

### DIFF
--- a/core/modules/field_ui/field_ui.admin.inc
+++ b/core/modules/field_ui/field_ui.admin.inc
@@ -2096,7 +2096,10 @@ function field_ui_field_edit_form_validate($form, &$form_state) {
     if ($cardinality === 'number' && $non_empty_items_count > $cardinality_number) {
       $errors[$field['field_name']][LANGUAGE_NONE][0][] = array(
         'error' => 'cardinality_smaller_than_existing_default_values',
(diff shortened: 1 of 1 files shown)


core/modules/field_ui/field_ui.admin.inc
```diff
@@ -2085,6 +2085,20 @@ function field_ui_field_edit_form_validate($form, &$form_state) {
...
+    $non_empty_items_count = 0;
+    foreach ($items as $item) {
+      if (!empty($item['value'])) {
+        $non_empty_items_count++;
+      }
+    }
+    if ($cardinality === 'number' && $non_empty_items_count > $cardinality_number) {
+      $errors[$field['field_name']][LANGUAGE_NONE][0][] = array(
+        'error' => 'cardinality_smaller_than_existing_default_values',
+        'message' => t('%name: you have entered more default values (@count_non_empty_items) than allowed by the %cardinality_setting_label setting.', array('%name' => $instance['label'], '@count_non_empty_items' => $non_empty_items_count, '@cardinality' => $cardinality_number, '%cardinality_setting_label' => 'Number of values')),
```
oadaeh: The array for the string has one more item than is in the string: `'@cardinality' => $cardinality_number,`

oadaeh: Also, `'%cardinality_setting_label' => 'Number of values'` does not need to be escaped and can just be placed in the string.
Also, also, the array probably should be broken out into multiple lines, but that's a Drupal coding standard, and I don't know it is the same with Backdrop.

oadaeh:
I added a couple of comments in the PR.

PR #2350, opened 2018-11-02
Issue #3223: Use a string when rewriting database prefix.

quicksketch:
Fixes https://github.com/backdrop/backdrop-issues/issues/3223.

### DIFF
--- a/core/includes/install.inc
+++ b/core/includes/install.inc
@@ -473,7 +473,7 @@ abstract class DatabaseTasks {
             'required' => TRUE,
           );
           $settings['database_prefix'] = array(
-            'value' => $db_settings['prefix'],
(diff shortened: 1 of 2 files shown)


core/includes/install.inc
```diff
@@ -473,7 +473,7 @@ abstract class DatabaseTasks {
             'required' => TRUE,
           );
           $settings['database_prefix'] = array(
-            'value' => $db_settings['prefix'],
+            'value' => $db_settings['prefix']['default'],
```
opi, line 476: `$db_settings` came from `getConnectionInfo` which indeed define `prefix` as an array with a `default` key.

opi (approved):

PR #2356, opened 2018-11-05
Issue #3371: Show "You do not have any admin items" message in '/admin/config'

klonos:
https://github.com/backdrop/backdrop-issues/issues/3371

### DIFF
--- a/core/modules/system/system.theme.inc
+++ b/core/modules/system/system.theme.inc
@@ -404,8 +404,6 @@ function theme_admin_page($variables) {
     }
   }
 
-  $output = '<div class="admin clearfix">';
-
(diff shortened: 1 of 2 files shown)


core/modules/system/system.theme.inc
```diff
@@ -406,12 +406,18 @@ function theme_admin_page($variables) {
 
   $output = '<div class="admin clearfix">';
```
quicksketch: Looks like this line should be removed as it's repeated below. Otherwise we'll have an unclosed div.

klonos: Indeed. Fixed.

PR #2361, opened 2018-11-07
Issue #3355: Update CKEditor to 4.11.1

klonos:
https://github.com/backdrop/backdrop-issues/issues/3355

New plugin added automatically by the build: `core/misc/ckeditor/plugins/clipboard`


core/modules/ckeditor/ckeditor.module
```diff
@@ -4,7 +4,7 @@
  * Provides integration with the CKEditor WYSIWYG editor.
  */
 
-define('CKEDITOR_VERSION', '4.7.3+2018-04-18-security-patch');
+define('CKEDITOR_VERSION', '4.10.1');
```
quicksketch: Looks like this should be updated to 4.11.0

klonos: 4.11.1 is out actually ...and changes were only limited to `CHANGES.md`

quicksketch:
Thanks! I've merge this into 1.x for 1.12.0 so we can get further testing prior to the release.

PR #2375, opened 2018-11-09
Issue #3375: IE11 & Chrome(PointerEvents enabled) & some Firefox scroll to the top of the page after dragging the bottom item with jQuery 1.5 <-> 1.11

klonos:
http://drupal.org/node/2843240

### DIFF
--- a/core/misc/tabledrag.js
+++ b/core/misc/tabledrag.js
@@ -618,26 +618,47 @@ Backdrop.tableDrag.prototype.dropRow = function (event, self) {
   }
 };
 
-
 /**
  * Get the coordinates from the event (allowing for browser differences).
  */
 Backdrop.tableDrag.prototype.pointerCoords = function (event) {
-  // Complete support for pointer events was only introduced to jQuery in
-  // version 1.11.1; between versions 1.7 and 1.11.0 pointer events have the
-  // clientX and clientY properties undefined. In those cases, the properties
-  // must be retrieved from the event.originalEvent object instead.
-  var clientX = event.clientX || event.originalEvent.clientX;
-  var clientY = event.clientY || event.originalEvent.clientY;
 
-  if (event.pageX || event.pageY) {
-    return { x: event.pageX, y: event.pageY };
+  // Match both null and undefined, but not zero, by using != null.
+  // See https://stackoverflow.com/questions/2647867/how-to-determine-if-variable-is-undefined-or-null
+  if (event.pageX != null && event.pageY != null) {
+    return {x: event.pageX, y: event.pageY};
   }
 
-  return {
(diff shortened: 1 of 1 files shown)


core/misc/tabledrag.js
```diff
@@ -618,26 +618,47 @@ Backdrop.tableDrag.prototype.dropRow = function (event, self) {
...
+    return {x: event.originalEvent.pageX, y: event.originalEvent.pageY};
+  }
 
-  if (event.pageX || event.pageY) {
-    return { x: event.pageX, y: event.pageY };
+  // Some old browsers do not support MouseEvent.pageX and *.pageY at all.
+  // See https://developer.mozilla.org/en-US/docs/Web/API/MouseEvent/pageY
+  // For those, we look at event.clientX and event.clientY instead.
+  if (event.clientX == null || event.clientY == null) {
+    // In some jQuery versions, some events created by jQuery do not have
```
klonos, line 644: Not sure if this applies to Backdrop, since we are using an up-to-date jQuery version.

@quicksketch ??

core/misc/tabledrag.js
```diff
@@ -618,26 +618,47 @@ Backdrop.tableDrag.prototype.dropRow = function (event, self) {
...
+      throw new Error("The original event has no coordinates.");
+    }
   }
 
-  return {
-    x: clientX + document.body.scrollLeft - document.body.clientLeft,
-    y: clientY + document.body.scrollTop  - document.body.clientTop
-  };
+  // Copied from jQuery.event.fix() in jQuery 1.4.1.
+  // In newer jQuery versions, this code is in jQuery.event.mouseHooks.filter().
```
klonos, line 656: Does this mean that we do not need this entire chunk of code here @quicksketch ??

core/misc/tabledrag.js
quicksketch: Yeah I think we may not need this entire PR, as it only affects earlier versions of jQuery we were not using. We should read the original Drupal issue and see if we can reproduce the problem as described there in Backdrop. If not, I think we should skip this whole PR.

quicksketch:
I cannot reproduce this problem as described in the Drupal.org issue on Backdrop. I tried in the latest Chrome (71) and Firefox (64) with and without touch events on both and I didn't have any problems. This seems to have been a widespread problem for Drupal users from the looks of the issue, but we haven't had any similar reports that I know of where the tabledrag was broken as described. We may have also slightly different code than Drupal that avoids this problem.

We could adjust our code to match Drupal's anyway, just for consistency. But in this case where it is only fixing a single function that has shims we apparently don't need, I think we may want to skip this.

PR #2378, opened 2018-11-12
Issue #3304: Add current path as a "source" parameter to the "Add redirect..." link in 404's

klonos:
https://github.com/backdrop/backdrop-issues/issues/3304

### DIFF
--- a/core/modules/redirect/redirect.admin.inc
+++ b/core/modules/redirect/redirect.admin.inc
@@ -320,8 +320,9 @@ function redirect_list_form_operations_confirm_form($form, &$form_state, $operat
 /**
  * Page callback; Redirect to the form for creating a redirect.
  */
-function redirect_add_redirect_page() {
-  backdrop_goto('admin/config/urls/redirect/add');
+function redirect_add_redirect_page(array $source = array()) {
+  $options = array('query' => $source);
+  backdrop_goto('admin/config/urls/redirect/add', $options);
 }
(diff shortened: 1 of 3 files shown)


klonos:
...refreshing the PR sandbox.

klonos:
...refreshing the PR sandbox.

klonos:
...refreshing the PR sandbox.

klonos:
...refreshing the PR sandbox.

herbdool:
@klonos you're close. I think you should use `backdrop_get_destination()` since it'll handle case where destination is not present. This works for me:

```
  $items[$site_404 . '/add-redirect'] = array(
    'title' => 'Add a URL redirect from this page to another location',
    'type' => MENU_LOCAL_ACTION,
    'page callback' => 'redirect_add_redirect_page',
    'page arguments' => array(backdrop_get_destination()),
    'access callback' => 'redirect_access',
    'access arguments' => array('create', 'redirect'),
    'file' => 'redirect.admin.inc',
  );
```

And in `redirect_add_redirect_page`

```
function redirect_add_redirect_page($source) {
  backdrop_goto('admin/config/urls/redirect/add', array('query' => array('source' => $source['destination'])));
}
```

`backdrop_get_destination()` returns an array formatted with destination so we need to reformat that which I do above. Plus note that I added in "query" => ... which you missed in yours.

PR #2379, opened 2018-11-12
Issue #3377: Fix PHP error Undefined index in path path_entity_insert…

andybroomfield:
… and path_delete_all_by_entity

Some entities (eg. Paragraphs) won't contain a uri.
Adds a check to see if the entity has a path set before creating or deleting paths.


klonos (approved):
👍

quicksketch:

PR #2395, opened 2018-11-27
Issue #3398: Default the status code to 500 before any other code is run.

klonos:
https://github.com/backdrop/backdrop-issues/issues/3398


quicksketch:
Great, merged into 1.x and 1.11.x. Thanks!

PR #2396, opened 2018-11-28
Fixed random sort weirdness.

ghost:
Fixes https://github.com/backdrop/backdrop-issues/issues/1909

### DIFF
--- a/core/modules/views/handlers/views_handler_sort_random.inc
+++ b/core/modules/views/handlers/views_handler_sort_random.inc
@@ -19,4 +19,8 @@ class views_handler_sort_random extends views_handler_sort {
     parent::options_form($form, $form_state);
     $form['order']['#access'] = FALSE;
   }
+
+  function can_expose() { return FALSE; }
+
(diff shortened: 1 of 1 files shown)


core/modules/views/handlers/views_handler_sort_random.inc
```diff
@@ -19,4 +19,8 @@ class views_handler_sort_random extends views_handler_sort {
     parent::options_form($form, $form_state);
     $form['order']['#access'] = FALSE;
   }
+
+  function can_expose() { return FALSE; }
+
+  function admin_summary() { }
```
klonos: ...wondering if we should return something here. Perhaps `return '';`

klonos (approved):
Looks good 👍

core/modules/views/handlers/views_handler_sort_random.inc
ghost: Ah yes, probably. Though I think `return;` will suffice.

PR #2402, opened 2018-12-04
Issue #3384 : Adding a link to an image.

Graham-72:
Prevent addition of data-file-id and text attributes.


quicksketch:
Thanks! I've merge this into 1.x and 1.11.x.

PR #2408, opened 2018-12-07
Issue #3386 : Thumbnails not showing in Library View.

### LINKED ISSUE #3386: No thumbnails in library
### In library media there are empty thumbnails just the filenames

---
PR by @Graham-72 : /backdrop/backdrop/pull/2408

Graham-72:
Fix by providing for customised public file system path.

Fixes https://github.com/backdrop/backdrop-issues/issues/3386

### DIFF
--- a/core/modules/file/views/views_handler_field_file_uri.inc
+++ b/core/modules/file/views/views_handler_field_file_uri.inc
@@ -43,12 +43,15 @@ class views_handler_field_file_uri extends views_handler_field_file {
     if (!empty($this->options['file_download_path']) && $data !== NULL && $data !== '') {
       $data = file_create_url($data);
       if (!empty($this->options['image_style'])) {
-      // $data contains url of image.
-      // Find insertion position of '/files'.
-        $insert_at = strpos($data, '/files');
-      // Text to insert in order to modify url.
+        // $data contains url of image.
+        // Get public file system path and its length.
+        $file_public_path = '/' . config_get('system.core','file_public_path');
+        $file_public_path_strlen = strlen($file_public_path);
+        // Find insertion position of public file system path.
+        $insert_at = strpos($data, $file_public_path);
+        // Text to insert in order to modify url.
         $add_to_url = '/styles/' . $this->options['image_style'] . '/public';
-        $data = substr_replace($data, $add_to_url, $insert_at + 6, 0);
+        $data = substr_replace($data, $add_to_url, $insert_at + $file_public_path_strlen, 0);
       }
     }
     return $this->render_link($data, $values);
--- a/core/modules/file/views/views_handler_field_file_uri.inc
+++ b/core/modules/file/views/views_handler_field_file_uri.inc
@@ -45,7 +45,7 @@ class views_handler_field_file_uri extends views_handler_field_file {
       if (!empty($this->options['image_style'])) {
         // $data contains url of image.
         // Get public file system path and its length.
-        $file_public_path = '/' . config_get('system.core','file_public_path');
+        $file_public_path = '/' . config_get('system.core', 'file_public_path');
         $file_public_path_strlen = strlen($file_public_path);
         // Find insertion position of public file system path.
         $insert_at = strpos($data, $file_public_path);


core/modules/file/views/views_handler_field_file_uri.inc
```diff
@@ -43,12 +43,15 @@ class views_handler_field_file_uri extends views_handler_field_file {
...
       $data = file_create_url($data);
       if (!empty($this->options['image_style'])) {
-      // $data contains url of image.
-      // Find insertion position of '/files'.
-        $insert_at = strpos($data, '/files');
-      // Text to insert in order to modify url.
+        // $data contains url of image.
+        // Get public file system path and its length.
+        $file_public_path = '/' . config_get('system.core','file_public_path');
+        $file_public_path_strlen = strlen($file_public_path);
```
klonos, line 49: @Graham-72 I was also surprised to find that `strlen()` returns the number of bytes rather than the number of characters in a string. So in some languages it might return a different result than what is expected: http://php.net/manual/en/function.strlen.php#118484

I realise that usually people will name their directories using English characters, but there might be an odd case where they don't. AFAIK strings are treated as arrays, so you can use `count()` on them. Might wanna consider that 😉

Some also recommend using `utf8_decode()` on the string before then running it trough `strlen()`: http://php.net/manual/en/function.strlen.php#45407
> `utf8_decode()` converts characters that are not in ISO-8859-1 to '?', which, for the purpose of counting, is quite alright.

Graham-72: @klonos this is interesting and a bit worrying. This render function makes use of php `substr_replace()` in combination with `strlen()`, the latter to provide the correct insertion point for the revised file URL. If both these PHP functions calculate numbers of bytes then perhaps the right answer is obtained. But if `substr_replace() `calculates the start position in characters whereas `strlen()` provides the number of bytes we would seem to have a mismatch.

And I am not sure that utf8_decode() would solve this dilemma. 

Maybe we need to test with 'abnormal' values for the public file system path?

Graham-72: I have now done a simple test by modifying the public file system path in a test installation, by adding the character é to the name (this being a two byte character). The `strlen()` function does indeed count bytes rather than characters but `substr_replace()` would seem to be doing the same and the image library is showing the thumbnail correctly.
So there does not seem to be a problem.

klonos: Good on ya for testing this @Graham-72 👍

core/modules/file/views/views_handler_field_file_uri.inc
```diff
@@ -43,12 +43,15 @@ class views_handler_field_file_uri extends views_handler_field_file {
     if (!empty($this->options['file_download_path']) && $data !== NULL && $data !== '') {
       $data = file_create_url($data);
       if (!empty($this->options['image_style'])) {
-      // $data contains url of image.
-      // Find insertion position of '/files'.
-        $insert_at = strpos($data, '/files');
```
jenlampton, line 48: nice catch!

core/modules/file/views/views_handler_field_file_uri.inc
```diff
@@ -43,12 +43,15 @@ class views_handler_field_file_uri extends views_handler_field_file {
...
-      // Text to insert in order to modify url.
+        // $data contains url of image.
+        // Get public file system path and its length.
+        $file_public_path = '/' . config_get('system.core','file_public_path');
```
jenlampton: looks like we may be missing a space after this comma.

Graham-72: Now added to PR.

core/modules/file/views/views_handler_field_file_uri.inc
quicksketch: Could this be rewritten to use `backdrop_strpos` and `backdrop_substr()`? Though we don't have a `backdrop_substr_replace()`, seems as though we could make do. Those functions count the number of characters, not the number of bytes, so they act more intuitively than the PHP native versions. However... if we're getting the right result perhaps we don't have a problem.

Graham-72: @quicksketch we don't seem to have a `backdrop_strpos()` do we?

Graham-72: I think it is OK as it is, using the PHP native functions, since they seem to be consistent with each other, and one as it were cancels the other. I think we would have a problem if we just used backdrop_substr() because we would be mixing counting in characters with counting in bytes.

quicksketch:
Thanks @Graham-72, merged into 1.x and 1.12.x.

PR #2411, opened 2018-12-09
Issue #3365: BackdropCMS v. 1.11.1 is changing underscores in classes to hyphens

Radcliffe:
Don't replace underscores with hyphens in custom class names.


quicksketch:
Thanks @Radcliffe! Merged into 1.x and 1.11.x.

PR #2412, opened 2018-12-09
Issue #3413: [UX] Better action button and page title for the layout reorder page.

klonos:
https://github.com/backdrop/backdrop-issues/issues/3413

### DIFF
--- a/core/modules/layout/layout.module
+++ b/core/modules/layout/layout.module
@@ -91,7 +91,7 @@ function layout_menu() {
   ) + $base;
 
   $items['admin/structure/layouts/reorder'] = array(
-    'title' => 'Configure layout priority',
+    'title' => 'Layout priority',
     'page callback' => 'backdrop_get_form',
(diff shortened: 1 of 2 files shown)


core/modules/layout/layout.module
```diff
@@ -91,10 +91,10 @@ function layout_menu() {
   ) + $base;
 
   $items['admin/structure/layouts/reorder'] = array(
-    'title' => 'Reorder',
+    'title' => 'Configure layout priority',
     'page callback' => 'backdrop_get_form',
     'page arguments' => array('layout_reorder_form'),
-    'type' => MENU_CALLBACK,
+    'type' => MENU_NORMAL_ITEM,
```
klonos, line 97: http://drupal.org/update/modules/6/7#menu_breadcrumb

> ...you should only use `'type' => MENU_CALLBACK` for items that are real callbacks (i.e., hidden, internal callbacks, typically used for API calls, with no menu or breadcrumbs)

PR #2415, opened 2018-12-10
Issue #3323: Typing a "/" in a field label stops the machine name autocomplete

amoutiers:
I choose to be as close as possible to the original code (ie lower case in js)


klonos (approved):
It works! 👍

quicksketch:

PR #2420, opened 2018-12-12
Issue #2461: Add option to delete content when deleting a content type

### LINKED ISSUE #2461: When deleting a content type add option to delete content
A checkbox on the confirm form would do.

klonos:
Fixes https://github.com/backdrop/backdrop-issues/issues/2461


klonos:
...refreshing the PR sandbox.

alanmels:
I've just tested this on http://2420.backdrop.backdrop.qa.backdropcms.org/admin/structure/types and unfortunately it still does not delete the content type itself. The website's log shows: 

```
Error
Argument 2 passed to t() must be an array, integer given, called in /home/qa/github/backdrop/backdrop/pr/2420/core/modules/dblog/dblog.admin.inc on line 310 and defined
```

core/modules/node/node.types.inc
```diff
@@ -683,11 +727,26 @@ function node_type_delete_confirm($form, &$form_state, $type) {
  * @see node_type_delete_confirm()
  */
 function node_type_delete_confirm_submit($form, &$form_state) {
-  $type_name = $form_state['values']['type'];
-  node_type_delete($type_name);
+  $type = node_type_load($form['type']['#value']);
+
+  $nodes = db_select('node', 'n')
```
indigoxela, line 732: Instead of db_select() you could use db_query() here, which is faster.

See the explanation [here](https://api.backdropcms.org/api/backdrop/core%21includes%21database%21database.inc/function/db_query/1)

Another thought: you need the same data in node_form_node_type_delete_confirm_alter(). Do you see any chance to get it from there via form storage or the like in $form_state?

core/modules/node/node.types.inc
```diff
@@ -683,11 +727,26 @@ function node_type_delete_confirm($form, &$form_state, $type) {
...
+            ->fields('n', array('nid'))
+            ->condition('type', $type->type, '=')
+            ->execute();
+  foreach ($nodes as $node) {
+    $nids[] = $node->nid;
+  }
+
+  if (!empty($nids)) {
+    node_delete_multiple($nids);
```
indigoxela, line 741: What worries me a bit here is the direct call to node_delete_multiple(), which potentially has to delete *a lot* of nodes and attached fields at once.

I'm not sure, but a batch job might make sense.

PR #2422, opened 2018-12-15
Issue #3430: Views: add padding to messages within dialogs.

klonos:
https://github.com/backdrop/backdrop-issues/issues/3430


quicksketch:

PR #2426, opened 2018-12-21
Remove the 16px message icons from the misc directory.

jenlampton:
Fixes https://github.com/backdrop/backdrop-issues/issues/3437

This PR also adds our new `info` color to info type messages.


core/modules/dblog/css/dblog.css
```diff
@@ -68,18 +68,22 @@
...
 }
 .admin-dblog .dblog-error .icon:before,
 .admin-dblog .dblog-critical .icon:before,
 .admin-dblog .dblog-alert .icon:before,
 .admin-dblog .dblog-emergency .icon:before {
-  background: #EE3D23 url(../../../misc/message-16-error.png) no-repeat center center;
-}
-.admin-dblog .dblog-debug .icon:before,
-.admin-dblog .dblog-info .icon:before,
-.admin-dblog .dblog-notice .icon:before {
```
opi, line 83: we loose notice ?

jenlampton:
The default is to use the notice icon now.

core/modules/dblog/css/dblog.css
quicksketch: Looks like it...

jenlampton: No, the default is notice now, instead. See line 71.

jenlampton: > we loose notice ?

Nope, see line 71.

quicksketch:
Thanks! As this isn't a bug fix I've only merged into 1.x, and not the current minor release.

PR #2438, opened 2018-12-31
Issue #3408: remove top padding from custom blocks when styled as messages.

klonos:
https://github.com/backdrop/backdrop-issues/issues/3408


jenlampton:
Closing because this is too specific a request to support in core. Changes like this should be done in a custom theme, if needed.

PR #2429, opened 2018-12-22
Issue #3428: Allow underscores and all valid characters in Views clas…

jenlampton:
Fixes https://github.com/backdrop/backdrop-issues/issues/3428

PR #2439, opened 2018-12-31
Revert "Issue #3428: Allow underscores and all valid characters in Views clas…"

quicksketch:
Reverts backdrop/backdrop#2429


quicksketch:
Reverting because this may cause a regression. Explanation posted in https://github.com/backdrop/backdrop-issues/issues/3428.

PR #2440, opened 2018-12-31
Issue #3271: Follow-up, set in-browser self-updates to FALSE.

jenlampton:
Fixes https://github.com/backdrop/backdrop-issues/issues/3271


quicksketch:
Thanks, merged in to leave the previous default at FALSE.

PR #2449, opened 2019-01-02
Issue #1992: Add support for signing and verifying modules

jlfranklin:
Prototype to address backdrop/backdrop-issues#1992

Required by backdrop-contrib/project#29

See also https://github.com/sd-backdrop/codesign_gnupg as a functional signing module.


core/modules/code_sign/code_sign.module
```diff
@@ -0,0 +1,241 @@
...
+
+    foreach ($all_signers as $engine => $info) {
+      $key_list = code_sign_get_private_keychain($engine);
+      foreach ($key_list as $key_id => $key_info) {
+        $signing_profiles[$key_id] = array(
+          '#crypto_engine' => $engine,
+          '#name' => $key_info['name'],
+          '#profile_id' => $key_id,
+          '#signing_key' => $key_info,
+        );
```
quicksketch, line 129: The addition of hash signs in front of each key is unnecessary here. Form API uses hashes because of the nested, arbitrary structure of forms. But in a known data structure, no hash signs are needed.

quicksketch:
I reworked this module significantly to use a class-based system (similar to Blocks, Layouts, or Views) and made a new updated PR at https://github.com/backdrop/backdrop/pull/5304. I kept the original history in place and adapted the examples and the GnuPG sub-module to work with the new approach.

PR #2451, opened 2019-01-02
3011 entity access

### LINKED ISSUE #3011: Add function to entity: entity_access
Paragraphs module uses this function, and it's likely other modules converted from Drupal would use it as well. We should add it to core.

---
PR by @jenlampton: https://github.com/backdrop/backdrop/pull/2451

jenlampton:
Fixes https://github.com/backdrop/backdrop-issues/issues/3011
Replaces https://github.com/backdrop/backdrop/pull/2119


core/modules/comment/comment.module
```diff
@@ -1518,20 +1508,36 @@ function comment_config_create_validate(Config $staging_config, $all_changes) {
...
- *   The operation that is to be performed on the comment. Only 'edit' is
- *   recognized now.
- * @param $comment
- *   The comment object.
+ * @param string $op
+ *   The operation to be performed on the comment. Possible values are:
+ *   - create
+ *   - view
+ *   - update
+ *   - delete
```
quicksketch, line 1516: "approve" should also be added here.

core/modules/entity/entity.class.inc
```diff
@@ -73,6 +75,37 @@ interface EntityInterface {
...
+  /**
+   * Determines whether the given user has access to create this type of entity.
+   *
+   * @param string $bundle
+   *   (optional) The entity's bundle to check create access against.
+   * @param User|AnonymousUser|object $account
+   *   (optional) The user to check for. Leave it to NULL to check for the
+   *   global user.
+   *
+   * @return bool
```
quicksketch, line 87:
```suggestion
   * @return bool|NULL
```

core/modules/entity/entity.class.inc
```diff
@@ -73,6 +75,37 @@ interface EntityInterface {
...
+   * Determines whether the given user has access to this entity.
+   *
+   * @param string $op
+   *   The operation being performed. One of 'view', 'update', 'create' or
+   *   'delete'.
+   * @param User|AnonymousUser|object $account
+   *   (optional) The user to check for. Leave it to NULL to check for the
+   *   global user.
+   *
+   * @return bool
```
quicksketch:
```suggestion
   * @return bool|NULL
```

core/modules/file/file.entity.inc
```diff
@@ -126,6 +128,58 @@ class File extends Entity {
...
+  /**
+   * Overrides Entity::createAccess().
+   */
+  public static function createAccess($bundle = NULL, $account = NULL) {
+    // If no user object is supplied, the access check is for the current user.
+    if (empty($account)) {
+      $account = $GLOBALS['user'];
+    }
+
+    return user_access('access content', $account);
```
quicksketch: Seems like this should always just return TRUE, as we don't have any requirements on this entity type. Since `Entity` defaults to returning TRUE, we can just delete this whole method.

core/modules/file/file.entity.inc
```diff
@@ -126,6 +128,58 @@ class File extends Entity {
...
+  }
+
+  /**
+   * Overrides Entity::access().
+   *
```
quicksketch, line 134: It would be good to note that "view" is not supported on the File Entity itself because it has no page callback (yet), but there is separately `file_download_access()` to check if a file is downloadable.

core/modules/node/node.entity.inc
```diff
@@ -206,6 +216,172 @@ class Node extends Entity {
...
+   *   (optional) The user to check for. Leave it to NULL to check for the
+   *   global user.
+   *
+   * @return bool
+   *   TRUE if access is granted, FALSE otherwise.
+   */
+  public function access($op, $account = NULL) {
+    // Casting class with private property causes errors due to added prefix.
+    // e.g. "\0" . 'Node' . "\0". So use static array instead.
+    $rights = &backdrop_static(__METHOD__, array());
```
quicksketch: Looks like this would create a static just with the name `access` because that is the value of `__METHOD__`, which isn't very identifying. If needed just use a hard-coded string like `'node_access'`, matching what it was before.

I'm also not understanding why we have a `private $rights` property defined above, but then we're using a static here. Should the property be removed?

core/modules/node/node.entity.inc
```diff
@@ -156,6 +155,15 @@ class Node extends Entity {
...
   public $in_preview;
 
+  /**
+   * Cached access values.
+   *
+   * @var array
+   *
+   * @see Node::access()
+   */
+  private $rights = array();
```
quicksketch: I don't see `$this->rights` being used below, should this be removed?

core/modules/node/node.entity.inc
```diff
@@ -206,6 +216,172 @@ class Node extends Entity {
     );
   }
 
+  /**
+   * Overrides Entity::createAccess().
+   */
+  public static function createAccess($bundle = NULL, $account = NULL) {
+    $rights = &backdrop_static(__METHOD__, array());
```
quicksketch: See other comment on `__METHOD__` not being unique enough.

core/modules/node/node.entity.inc
jenlampton: Should this be updating the private $rights?

core/modules/node/node.entity.inc
jenlampton: yes.

core/modules/node/node.entity.inc
jenlampton: ok.

quicksketch:
Thanks for getting this revised! Merged into 1.x for 1.13.0.

PR #2455, opened 2019-01-05
Issue #3459: file_unmanaged_move() should issue rename() where possible instead of copy() & unlink()

klonos:
https://github.com/backdrop/backdrop-issues/issues/3459

### DIFF
--- a/core/includes/file.inc
+++ b/core/includes/file.inc
@@ -1158,11 +1158,8 @@ function file_unmanaged_move($source, $destination = NULL, $replace = FILE_EXIST
   if (!file_unmanaged_prepare($source, $destination, $replace)) {
     return FALSE;
   }
-  // Ensure compatibility with Windows.
-  // @see backdrop_unlink()
-  if ((substr(PHP_OS, 0, 3) == 'WIN') && (!file_stream_wrapper_valid_scheme(file_uri_scheme($source)))) {
-    chmod($source, 0600);
-  }
+  // The file must be writable in the old location before moving.
+  backdrop_chmod($source);
   // Attempt to resolve the URIs. This is necessary in certain configurations
   // (see above) and can also permit fast moves across local schemes.
   $real_source = ($real_source = backdrop_realpath($source)) ? $real_source : $source;
--- a/core/includes/file.inc
+++ b/core/includes/file.inc
@@ -1140,7 +1140,7 @@ function file_move(File $source, $destination = NULL, $replace = FILE_EXISTS_REN
  * @param $destination
  *   A URI containing the destination that $source should be moved to. The
  *   URI may be a bare filepath (without a scheme) and in that case the default
- *   scheme (file://) will be used. If this value is omitted, Drupal's default
+ *   scheme (file://) will be used. If this value is omitted, Backdrop's default
  *   files scheme will be used, usually "public://".
  * @param $replace
  *   Replace behavior when the destination file already exists:
--- a/core/includes/file.inc
+++ b/core/includes/file.inc
@@ -1168,8 +1168,7 @@ function file_unmanaged_move($source, $destination = NULL, $replace = FILE_EXIST
   if (!@rename($real_source, $real_destination)) {
     // Fall back to slow copy and unlink procedure. This is necessary for
     // renames across schemes that are not local, or where rename() has not been
-    // implemented. It's not necessary to use drupal_unlink() as the Windows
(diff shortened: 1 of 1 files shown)


core/includes/file.inc
```diff
@@ -1107,16 +1150,37 @@ function file_move(File $source, $destination = NULL, $replace = FILE_EXISTS_REN
...
-  if ($filepath == FALSE || file_unmanaged_delete($source) == FALSE) {
+  if (!file_unmanaged_prepare($source, $destination, $replace)) {
     return FALSE;
   }
-  return $filepath;
+  // Ensure compatibility with Windows.
+  // @see backdrop_unlink()
+  if ((substr(PHP_OS, 0, 3) == 'WIN') && (!file_stream_wrapper_valid_scheme(file_uri_scheme($source)))) {
+    chmod($source, 0600);
+  }
```
quicksketch: This entire bit is confusing. I noted that David_Rothstein said the same thing in his code review on Drupal. It's not clear why this would affect Windows only.

In order to move a file, write permissions are necessary both on the source and the destination, since the file is going to be moved from the previous location. Since we use `backdrop_chmod($destination)` on the destination, the moved file is going to have the standard Backdrop permissions. I think if we want to avoid permissions problems, we should simply do the same thing on `$source` prior to the move. It would accomplish the same thing to ensure the file is writable for Windows computers, if they actually have a problem in the first place.

So I would suggest the following:

```
  // The file must be writable in the old location before moving.
  backdrop_chmod($source);
```

core/includes/file.inc
```diff
@@ -1107,16 +1150,37 @@ function file_move(File $source, $destination = NULL, $replace = FILE_EXISTS_REN
...
+  // Attempt to resolve the URIs. This is necessary in certain configurations
+  // (see above) and can also permit fast moves across local schemes.
+  $real_source = ($real_source = backdrop_realpath($source)) ? $real_source : $source;
+  $real_destination = ($real_destination = backdrop_realpath($destination)) ? $real_destination : $destination;
+  // Perform the move operation.
+  if (!@rename($real_source, $real_destination)) {
+    // Fall back to slow copy and unlink procedure. This is necessary for
+    // renames across schemes that are not local, or where rename() has not been
+    // implemented. It's not necessary to use drupal_unlink() as the Windows
+    // issue has already been resolved above.
```
quicksketch: I think if we make the above change, we can omit "It's not necessary to use drupal_unlink() as the Windows issue has already been resolved above." (And we don't want drupal_unlink() in there anyway).

core/includes/file.inc
```diff
@@ -894,6 +893,54 @@ function file_valid_uri($uri) {
...
+ *   is reported.
+ * - If file already exists in $destination either the call will error out,
+ *   replace the file or rename the file based on the $replace parameter.
+ *
+ * @param $source
+ *   A string specifying the filepath or URI of the source file.
+ * @param $destination
+ *   A URI containing the destination that $source should be moved/copied to.
+ *   The URI may be a bare filepath (without a scheme) and in that case the
+ *   default scheme (file://) will be used. If this value is omitted, Drupal's
```
```suggestion
 *   default scheme (file://) will be used. If this value is omitted, Backdrop's
```

core/includes/file.inc
```diff
@@ -1092,13 +1124,24 @@ function file_move(File $source, $destination = NULL, $replace = FILE_EXISTS_REN
...
  * @param $source
- *   A string specifying the filepath or URI of the original file.
+ *   A string specifying the filepath or URI of the source file.
  * @param $destination
- *   A string containing the destination that $source should be moved to.
- *   This must be a stream wrapper URI. If this value is omitted, Backdrop's
- *   default files scheme will be used, usually "public://".
+ *   A URI containing the destination that $source should be moved to. The
+ *   URI may be a bare filepath (without a scheme) and in that case the default
+ *   scheme (file://) will be used. If this value is omitted, Drupal's default
```
```suggestion
 *   scheme (file://) will be used. If this value is omitted, Backdrop's default
```

quicksketch (changes requested):
Mostly docblock concerns. Could you adjust?

core/includes/file.inc
klonos: Agreed. Done.

core/includes/file.inc
klonos: ...also done 😉

quicksketch (approved):

PR #2456, opened 2019-01-07
Issue #1841: Add icons to the admin bar.

jenlampton:
Fixes https://github.com/backdrop/backdrop-issues/issues/1841

The one part of the PR that could use work is the addition of the classes. I'm currently adding a class to every link based on it's URL. It might be smarter to only add classes where they are needed (top-level only) and maybe name them something different? I liked these names because the paths will never change, where something like mlid could be different on every site.


core/modules/admin_bar/css/admin_bar.css
```diff
@@ -137,7 +215,7 @@
 [dir="rtl"] #admin-bar .dropdown li li a {
   border-left: none;
 }
-#admin-bar .dropdown a:hover {
+#admin-bar .dropdown ul a:hover {
```
opi: This prevent "user" and "logout" links to have a hover style.
Moreover, every other link (nested, nested expandable, ...) have a `text-decoration: underline;`, so I'd like first level link to have it too. #a11y

opi (changes requested):

docwilmot:
The icon for the Admin user looks funny on mobile. And I would like icons on mobile as well. Why not?

core/modules/admin_bar/css/admin_bar.css
jenlampton: I did that on purpose (removed the underline) as I thought it was a visual improvement. I didn't realize it was an #a11y regression -- putting back!

jenlampton:
Added icons back for mobile.

> I'm currently adding a class to every link

Fixed this too.

jenlampton:
Closing for new sandbox.

jenlampton:
Sandbox?

docwilmot:
`#admin-bar-menu.dropdown li > span.menu {    background: pink` ??

core/modules/admin_bar/admin_bar.theme.inc
```diff
@@ -65,6 +65,12 @@ function theme_admin_bar_links($variables) {
       }
       if (isset($elements[$path]['#attributes']['class'])) {
         $elements[$path]['#attributes']['class'] = $elements[$path]['#attributes']['class'];
+
+        // Only add additional classes to top-level links.
+        if (substr_count($path, '/') <=1 ) {
```
klonos:
```suggestion
        if (substr_count($path, '/') <= 1 ) {
```

core/modules/admin_bar/css/admin_bar.css
```diff
@@ -68,14 +68,142 @@
...
+  background-size: 16px;
+}
+[dir="rtl"] #admin-bar #admin-bar-menu > li > .dropdown > li > a.admin-content,
+[dir="rtl"] #admin-bar #admin-bar-menu > li > .dropdown > li > a.admin-people,
+[dir="rtl"] #admin-bar #admin-bar-menu > li > .dropdown > li > a.admin-appearance,
+[dir="rtl"] #admin-bar #admin-bar-menu > li > .dropdown > li > a.admin-modules,
+[dir="rtl"] #admin-bar #admin-bar-menu > li > .dropdown > li > a.admin-structure,
+[dir="rtl"] #admin-bar #admin-bar-menu > li > .dropdown > li > a.admin-config,
+[dir="rtl"] #admin-bar #admin-bar-menu > li > .dropdown > li > a.admin-devel,
+[dir="rtl"] #admin-bar #admin-bar-menu > li > .dropdown > li > a.admin-reports {
```
klonos: ...wondering whether we could replace all these with a single `[dir="rtl"] #admin-bar #admin-bar-menu > li > .dropdown > li > a`. This list here could get pretty huge and unmanageable if we decide to add icons to more menu items.

core/modules/admin_bar/admin_bar.theme.inc
jenlampton: Fixed!

core/modules/admin_bar/css/admin_bar.css
jenlampton: I didn't think that would work due to selector specificity, but it totally did! 
screen shot 2019-01-11 at 3 57 30 pm

core/modules/admin_bar/css/admin_bar.css
```diff
@@ -68,14 +68,135 @@
...
+}
+[dir="rtl"] #admin-bar .dropdown .admin-bar-account > a {
+  background-position: 10% center;
+  padding-right: 10px;
+  padding-left: 28px;
+}
+
+/* Log Out link */
+#admin-bar .dropdown .admin-bar-logout > a {
+  background: transparent url(../images/sign-out--white--64.png) 95% center no-repeat; /* LTR */
```
opi: background-position should be "90% center" to match other links in the #admin-bar-extra list.

jenlampton:
This link is narrower. The icon position looked better at this position.

opi:
> This link is narrower. The icon position looked better at this position.

Not in mobile view, though. Maybe 95% for all ?

quicksketch:
Seems like part of the problem here is that the icons are different sizes, which is odd considering they're all named `--64.png`, which I would have thought meant 64x64 pixels. I think a better solution to using a background-image directly on the links would be to use a pseudo-element that is placed with `:after`, then it can be made a consistent size and not have to worry about percentage differences.

But in any case, I think that can happen later. For now I went ahead and went with the 90% across the board as suggested by opi, which looks better on mobile where they're all lined up vertically and doesn't make a discernible difference on desktop (to my eye). Let's go with this and refine as needed.

opi:
> a better solution to using a background-image directly on the links would be to use a pseudo-element 

+1000 ; I keep this tasks in my head, for a rainy day (which come really often in Brittany, France).

thanks @quicksketch for the merge

PR #2462, opened 2019-01-11
Issue #3466: ETags should be quoted.

quicksketch:
Fixes https://github.com/backdrop/backdrop-issues/issues/3466.


klonos (approved):
LGTM and works 👍

PR #2467, opened 2019-01-11
Issue #3468: User role drop-buttons should read Configure instead of …

jenlampton:
Fixes https://github.com/backdrop/backdrop-issues/issues/3468


jenlampton:
Looks like this failure might be random? I'm not getting it on my local.

quicksketch:
Thanks, merged into 1.x and 1.11.x.

PR #2071, opened 2018-01-16
Issue #2923: Add configuration options to the search block.

jenlampton:
Fixes https://github.com/backdrop/backdrop-issues/issues/2923

PR #2468, opened 2019-01-11
Issue #2923: Add configuration options to the search block.

jenlampton:
Fixes https://github.com/backdrop/backdrop-issues/issues/2923
Replaces https://github.com/backdrop/backdrop/pull/2071

### DIFF
--- a/core/modules/search/search.module
+++ b/core/modules/search/search.module
@@ -112,12 +112,92 @@ function search_block_info() {
 }
 
 /**
+ * Implements hook_block_configure().
+ */
+function search_block_configure($delta = '', $settings = array()) {
+  $form = array();
+  if ($delta == 'form') {
+    $search_settings = !empty($settings['search'])? $settings['search'] : array();
+    $search_settings += array(
+      'label' => '',
+      'label_hide' => TRUE,
+      'placeholder' => '',
+      'button_text' => 'Search',
+    );
+
+    $form['search'] = array(
+      '#type' => 'fieldset',
+      '#title' => t('Search form settings'),
+      '#collapsible' => TRUE,
+    );
+
+    $form['search']['label'] = array(
+      '#type' => 'textfield',
+      '#title' => t('Search box label'),
+      '#default_value' => check_plain($search_settings['label']),
+    );
+    $form['search']['label_hide'] = array(
+      '#type' => 'checkbox',
+      '#title' => t('Hide the search box label'),
+      '#default_value' => $search_settings['label_hide'],
+    );
+    $form['search']['placeholder'] = array(
+      '#type' => 'textfield',
+      '#title' => t('Placeholder text'),
+      '#default_value' => check_plain(trim($search_settings['placeholder'])),
+      '#description' => t('This text will appear in the search box until text is entered.'),
+    );
+    $form['search']['button_text'] = array(
+      '#type' => 'textfield',
+      '#title' => t('Button text'),
+      '#default_value' => check_plain(trim($search_settings['button_text'])),
+      '#description' => t('Leave this field empty to hide the submit button.'),
+    );
+  }
+  return $form;
+}
+
+/**
  * Implements hook_block_view().
  */
-function search_block_view($delta = '') {
+function search_block_view($delta = '', $settings = array(), $contexts = array()) {
   if (user_access('search content')) {
+    $search_settings = !empty($settings['search'])? $settings['search'] : array();
+    $search_settings += array(
+      'label' => '',
+      'label_hide' => TRUE,
+      'placeholder' => '',
+      'button_text' => 'Search',
+    );
+    $form = backdrop_get_form('search_block_form');
+
+    // Show the label.
+    if (!$search_settings['label_hide']) {
+      $form['search_block_form']['#title_display'] = 'before';
+
+      // Update the label text.
+      if ($search_settings['label'] != '') {
+        $form['search_block_form']['#title'] = t(check_plain($search_settings['label']));
+      }
+    }
+
+    // Add a placeholder.
+    if ($search_settings['placeholder'] != '') {
+      $form['search_block_form']['#attributes']['placeholder'] = t(trim($search_settings['placeholder']));
+    }
+
+    // Change the button text.
+    if ($search_settings['button_text'] != 'Search') {
+      if ($search_settings['button_text'] == '') {
+        $form['actions']['submit']['#access'] = FALSE;
+      }
+      else {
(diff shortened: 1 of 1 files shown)


docwilmot:
The block cant be added to layouts due to a PHP error: `Undefined index: search in search_block_configure()` where there is a `$search_settings = $settings['search'];`

core/modules/search/search.module
```diff
@@ -111,13 +111,94 @@ function search_block_info() {
...
+    $settings += array(
+      'label' => '',
+      'label_hide' => TRUE,
+      'placeholder' => '',
+      'button' => 'Search',
+    );
+    $form = backdrop_get_form('search_block_form');
+
+    // Show the label.
+    if (!$search_settings['label_hide']) {
```
docwilmot, line 175: if the "search box label will always be visible to screen readers", then shouldnt there always be one? are we missing an `else`?

core/modules/search/search.module
```diff
@@ -111,13 +111,94 @@ function search_block_info() {
...
+
+    $form['search']['label'] = array(
+      '#type' => 'textfield',
+      '#title' => t('Search box label'),
+      '#default_value' => check_plain($search_settings['label']),
+    );
+    $form['search']['label_hide'] = array(
+      '#type' => 'checkbox',
+      '#title' => t('Hide the search box label'),
+      '#description' => t('The search box label will always be visible to screen readers.'),
```
docwilmot: Is this an a11y requirement?

core/modules/search/search.module
jenlampton: Hm. Yes, good point!

core/modules/search/search.module
jenlampton: yes, exactly.

core/modules/search/search.module
jenlampton: In the function [`search_box()`](https://api.backdropcms.org/api/backdrop/core%21modules%21search%21search.module/function/search_box/1) the title display is already defined as `invisible`. Only when the title is visible does value needs to be changed to `before`

jenlampton:
Updated PR to address PHP notices.

core/modules/search/search.module
```diff
@@ -111,13 +111,98 @@ function search_block_info() {
   return $blocks;
 }
 
+/**
+ * Implements hook_block_configure().
+ */
+function search_block_configure($delta = '', $settings = array()) {
+  $form = array();
+  if ($delta == 'form') {
+    if (array_key_exists('search', $settings) && !empty($settings['search'])) {
```
docwilmot: Isn't this the same as `if (!empty($settings['search'])) {`?

jenlampton: /me goes to look that up...

docwilmot: Youd still need to declare `$search_settings = array()` before this. The PHP error persists.

jenlampton: yes. And it also looks like `array_key_exists()` is a lot slower than `isset()`! Still learning things :)

jenlampton: > The PHP error persists.

What are your steps to reproduce the error? I'm not getting it...  I thought the `$search_settings += ` should handle the assignment.

jenlampton: Nevermind, I got it! pushing a fix now :)

core/modules/search/search.module
```diff
@@ -111,13 +111,93 @@ function search_block_info() {
...
+
+    $form['search']['label'] = array(
+      '#type' => 'textfield',
+      '#title' => t('Search box label'),
+      '#default_value' => check_plain($search_settings['label']),
+    );
+    $form['search']['label_hide'] = array(
+      '#type' => 'checkbox',
+      '#title' => t('Hide the search box label'),
+      '#default_value' => $search_settings['label_display'],
```
docwilmot: this should be `$search_settings['label_hide']`

core/modules/search/search.module
```diff
@@ -111,13 +111,93 @@ function search_block_info() {
...
+    );
+    $form['search']['placeholder'] = array(
+      '#type' => 'textfield',
+      '#title' => t('Placeholder text'),
+      '#default_value' => check_plain($search_settings['placeholder']),
+      '#description' => t('This text will appear in the search box until text is entered.'),
+    );
+    $form['search']['button'] = array(
```
docwilmot: this would probably be best as `$form['search']['button_text']`

core/modules/search/search.module
jenlampton: yep, fixed.

core/modules/search/search.module
jenlampton: okay. fixed.

core/modules/search/search.module
```diff
@@ -111,13 +111,103 @@ function search_block_info() {
...
+      '#description' => t('Leave this field empty to hide the submit button.'),
+    );
+  }
+  return $form;
+}
+
+/**
+ * Implements hook_form_FORM_ID_alter().
+ */
+function search_form_layout_block_configure_form_alter(&$form, &$form_state) {
```
docwilmot: Another way would have been for `hook_block_view()` set `$block['subject'] = '';`
Users can set a block title in the block config UI title settings dropdown if they want. That would need an update hook to explicitly reset existing blocks with `"title_display": "default"` to `"title_display": "custom"` and `"title": "Search"` (translated).

jenlampton: Hm, I like your way better. It's much cleaner. The update hook might be tricky though. I think maybe I'll move that to another issue so we can wrap this up and get it done :) 
Tah Dah: https://github.com/backdrop/backdrop-issues/issues/3476

jenlampton: I've removed the form alter from this PR. Will work on that over in https://github.com/backdrop/backdrop-issues/issues/3476

findlabnet:
An element `$search_settings['button']` was not declared/defined, so I got notice instead rendered button:
`Notice: Undefined index: button in search_block_view() (line 191 of /home/qa/github/backdrop/backdrop/pr/2468/core/modules/search/search.module).`

jenlampton:
Nice catch @findlabnet, I have pushed a fix to the PR.

findlabnet:
I think values of fields "Placeholder text" and "Button text" should be saved via trim() function, for now we can have saved empty spaces before and after real text.

jenlampton:
> I think values of fields "Placeholder text" and "Button text" should be saved via trim()

good call 👍  I've placed a `trim()` in both the form and the display.

PR #2470, opened 2019-01-12
Issue #3476: Remove the search block default title. Update old blocks…

### LINKED ISSUE #3476: Remove search block default title
In https://github.com/backdrop/backdrop-issues/issues/2923 we discussed how there are a lot of instances of the word "Search". Ideally, the search form block would have no title. We can remove it, but we'd also need an update hook to update all existing layouts to add a custom block title for any block that had previously used the default. This could be a bit tricky, so in the hopes of getting https://github.com/backdrop/backdrop-issues/issues/2923 in for Backdrop 1.12, I've split this off into a separate issue. 

The problem: 
screen shot 2019-01-12 at 11 08 14 am

The solution:
screen shot 2019-01-12 at 11 09 14 am

Steps to test:
1. Fresh install of backdrop (or, untouched sandbox)
2. Confirm there is no search block on any layout
3. Add a search block -- ignore all settings on the configure block form, just add it blindly.
4. Confirm there is no title shown for the search block.

---
PR: https://github.com/backdrop/backdrop/pull/2470

jenlampton:
Fixes https://github.com/backdrop/backdrop-issues/issues/3476

### DIFF
--- a/core/modules/search/search.install
+++ b/core/modules/search/search.install
@@ -202,6 +202,29 @@ function search_update_1001() {
 }
 
 /**
+ * Set all search blocks with 'default' title to 'custom' title with 'Search'.
+ */
+function search_update_1002() {
+  $config_names = config_get_names_with_prefix('layout.layout.');
+  foreach ($config_names as $config_file) {
+    $config = config($config_file);
+    $data = $config->get();
+    foreach ($data['content'] as $hash => $block) {
+      if ($block['plugin'] == 'search:form') {
+        $has_title_display = isset($block['data']['settings']['title_display']);
+        if ($has_title_display && $block['data']['settings']['title_display'] == 'default') {
+          $data['content'][$hash]['data']['settings']['title_display'] = 'custom';
+          $data['content'][$hash]['data']['settings']['title'] = 'Search';
+        }
+      }
+    }
+    // Save the config.
+    $config->setData($data);
+    $config->save();
+  }
+}
+
+/**
  * @} End of "addtogroup updates-7.x-to-1.x"
  * The next series of updates should start at 2000.
  */
--- a/core/modules/search/search.module
+++ b/core/modules/search/search.module
@@ -120,7 +120,7 @@ function search_block_configure($delta = '', $settings = array()) {
     $search_settings = !empty($settings['search'])? $settings['search'] : array();
     $search_settings += array(
       'label' => '',
-      'label_hide' => TRUE,
+      'label_hide' => FALSE,
       'placeholder' => '',
       'button_text' => 'Search',
     );
@@ -198,6 +198,7 @@ function search_block_view($delta = '', $settings = array(), $contexts = array()
 
     $block['subject'] = t('Search');
     $block['content'] = $form;
+
     return $block;
   }
 }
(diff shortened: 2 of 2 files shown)


core/modules/search/search.install
```diff
@@ -201,6 +201,27 @@ function search_update_1001() {
...
+ * Set all search blocks with 'default' title to 'custom' title with 'Search'.
+ */
+function search_update_1002() {
+  $config_names = config_get_names_with_prefix('layout.layout.');
+  foreach ($config_names as $config_file) {
+    $config = config($config_file);
+    $data = $config->get();
+    foreach ($data['content'] as $hash => $block) {
+      if ($block['plugin'] == 'search:form'
+          && $block['data']['settings']['title_display'] == 'default') {
```
klonos: Not sure if such splitting of expressions within conditionals is according to our coding standards.

jenlampton: I'm not sure either, but I thought I'd send it by @quicksketch and see what he thought. I do this all the time in my custom code (so I can read it clearly) and did it here by force-of-habit. After I looked at it I thought, eh, maybe? and subimtted the PR anyway :)

quicksketch: While less preferable than using separate variables, I found that we have at least 45 instances of doing this in code currently. I think updating the coding standards is in order to include reasonable separation of conditions on different lines by `&&` or `||` operators.

jenlampton: If i wanted to use separate variables, would it need to add 4 extra lines, like this?
```php
      $block_title_default = FALSE;
      if ($block['data']['settings']['title_display'] == 'default') {
        $block_title_default = TRUE;
      }
      if ($block['plugin'] == 'search:form' && $block_title_default) {
```

Or could we do use a ternary even though it's more than 80 characters?
```php
      $default = $block['data']['settings']['title_display'] == 'default'? TRUE: FALSE;
      if ($block['plugin'] == 'search:form' && $default) {
```

core/modules/search/search.install
```diff
@@ -201,6 +201,27 @@ function search_update_1001() {
...
+ * Set all search blocks with 'default' title to 'custom' title with 'Search'.
+ */
+function search_update_1002() {
+  $config_names = config_get_names_with_prefix('layout.layout.');
+  foreach ($config_names as $config_file) {
+    $config = config($config_file);
+    $data = $config->get();
+    foreach ($data['content'] as $hash => $block) {
+      $default = $block['data']['settings']['title_display'] == 'default'? TRUE: FALSE;
+      if ($block['plugin'] == 'search:form' && $default) {
```
docwilmot: This is failing on upgrade tests only. Presumably the upgrade test may create blocks without settings and thus no `'settings'` key. This should work better:
```
    foreach ($data['content'] as $hash => $block) {
      if ($block['plugin'] == 'search:form' && isset($block['data']['settings']['title_display']) && $block['data']['settings']['title_display'] == 'default') {
         ...
```
Or maybe split into two lines:
```
    foreach ($data['content'] as $hash => $block) {
      if ($block['plugin'] == 'search:form') {
        if (isset($block['data']['settings']['title_display']) && $block['data']['settings']['title_display'] == 'default') {
         ...
```

jenlampton: I ended up going with 3 lines to keep the character limit as close to 80 as seemed reasonable: 
```php
      if ($block['plugin'] == 'search:form') {
        $has_title_display = isset($block['data']['settings']['title_display']);
        if ($has_title_display && $block['data']['settings']['title_display'] == 'default') {
```

core/modules/search/search.install
ghost: One option is:

```php
$block_title_display = $block['data']['settings']['title_display'];
if ($block['plugin'] == 'search:form' && $block_title_display == 'default') {
```

However I tend to steer away from using a variable like this when it's just used in one place only. Not sure if that's right or wrong though...

> Or could we do use a ternary even though it's more than 80 characters?

Code can be longer than 80 characters, it's only comments that are affected by that restriction (to my knowledge).

jenlampton:
closing for new sandbox.

jenlampton:
sandbox please

laryn:
Thanks @jenlampton, @quicksketch, @docwilmot, @BWPanda, and @klonos! Merged into 1.x and 1.19.x.

PR #2474, opened 2019-01-13
Issue #3431 by shabibalm: Fix RTL formatting options in Seven.

jenlampton:
Fixes https://github.com/backdrop/backdrop-issues/issues/3431

### DIFF
--- a/core/modules/system/css/system.theme.css
+++ b/core/modules/system/css/system.theme.css
@@ -240,6 +240,7 @@ abbr.ajax-changed {
 .js[dir="rtl"] fieldset.collapsed .fieldset-title {
   background-image: url(../../../misc/menu-collapsed-rtl.png);
   background-position: 98% 50%;
+  background-repeat: no-repeat;
 }
 .fieldset-legend span.summary {
   color: #999;
--- a/core/themes/seven/css/style.css
+++ b/core/themes/seven/css/style.css
@@ -692,6 +692,10 @@ fieldset .fieldset-wrapper {
   border-radius: 4px;
   background: transparent;
 }
+.js[dir="rtl"] fieldset.collapsible .fieldset-title {
+  background-image: none;
+  padding: .3em 1.8em .3em .3em;
+}
 .js fieldset.collapsible .fieldset-legend a:before {
   content: "";
   position: absolute;
@@ -703,7 +707,10 @@ fieldset .fieldset-wrapper {
   border-top-color: #000;
   border-bottom: 0;
 }
-
+.js[dir="rtl"] fieldset.collapsible .fieldset-legend a:before {
+  left: auto;
+  right: .6em;
+}
 .fieldset-legend span.summary {
   position: absolute;
   top: .4em;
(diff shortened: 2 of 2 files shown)


core/themes/seven/css/style.css
```diff
@@ -692,6 +692,9 @@ fieldset .fieldset-wrapper {
   border-radius: 4px;
   background: transparent;
 }
+.js[dir="rtl"] fieldset.collapsible .fieldset-title {
+  background-image: none;
```
quicksketch, line 696: It's not clear to me how or why this would do anything. Is there a background image in the first place that needs hiding?

core/themes/seven/css/style.css
```diff
@@ -703,6 +706,9 @@ fieldset .fieldset-wrapper {
   border-top-color: #000;
   border-bottom: 0;
 }
+.js[dir="rtl"] fieldset.collapsible .fieldset-legend a:before {
+  right: 0;
```
quicksketch: To make this symmetrical, seems as though this should be `right: .6em;`. Also when specifying the RTL version, the left value should be reset, such as `left: auto;`.

core/themes/seven/css/style.css
```diff
@@ -725,6 +731,12 @@ fieldset .fieldset-wrapper {
   border: .32em solid transparent;
   border-left-color: black;
 }
+.js[dir="rtl"] fieldset.collapsed .fieldset-legend a:before {
+  right: 0;
```
quicksketch: As above, this should be:
```
left: auto;
right: .8em;
```

core/themes/seven/css/style.css
```diff
@@ -725,6 +731,12 @@ fieldset .fieldset-wrapper {
...
+  top: .75em;
+  border: .32em solid transparent;
```
quicksketch: `top` rule can be removed entirely.

`border` rule is slightly redundant but it works to reset the border. Perhaps it should just be `border-left-color: transparent;`?

quicksketch (changes requested):

core/modules/system/css/system.theme.css
```diff
@@ -240,6 +240,7 @@ abbr.ajax-changed {
 .js[dir="rtl"] fieldset.collapsed .fieldset-title {
   background-image: url(../../../misc/menu-collapsed-rtl.png);
   background-position: 98% 50%;
+  background-repeat: no-repeat;
```
opi, line 243: the background-repeat rule should be by default, in the upper selector

opi (changes requested):

core/themes/seven/css/style.css
jenlampton: > Is there a background image in the first place that needs hiding?

yes. It is added in system.theme.css:
```
.js[dir="rtl"] fieldset.collapsed .fieldset-title {
    background-image: url(../../../misc/menu-collapsed-rtl.png);
    background-position: 98% 50%;
    background-repeat: no-repeat;
}
``

core/themes/seven/css/style.css
jenlampton: fixed

core/themes/seven/css/style.css
jenlampton: fixed.

core/themes/seven/css/style.css
jenlampton: fixed.

core/modules/system/css/system.theme.css
jenlampton: the upper selector is for `background-image` not `background`. This copies what other selectors in this file are doing. We can re-work in a separate issue if necessary.

quicksketch:
Thanks, merged into 1.x and 1.12.x!

PR #2488, opened 2019-01-26
Issue #3424: [UX] Fix RTL vertical tabs style

klonos:
https://github.com/backdrop/backdrop-issues/issues/3424

### DIFF
--- a/core/themes/seven/css/vertical-tabs.css
+++ b/core/themes/seven/css/vertical-tabs.css
@@ -7,7 +7,10 @@
   border: 1px solid #ccc;
   background: #fff;
 }
- @media (min-width: 750px) {
+[dir="rtl"] .vertical-tabs:before {
+  right: 0;
+}
+@media (min-width: 750px) {
   .vertical-tabs {
     padding: 0 0 0 15em; /* LTR */
   }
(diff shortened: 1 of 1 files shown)


core/themes/seven/css/vertical-tabs.css
```diff
@@ -7,7 +7,10 @@
   border: 1px solid #ccc;
   background: #fff;
 }
- @media (min-width: 750px) {
+[dir="rtl"] .vertical-tabs:before {
+  right: 0;
```
quicksketch: Hm, where is the corresponding LTR rule for this? I suppose it's implied just because left: 0 is the default? I think we need to clarify this by adding:

```
.vertical-tabs:before {
  left: 0; /* LTR */
}
```

core/themes/seven/css/vertical-tabs.css
```diff
@@ -34,16 +37,16 @@
...
     float: left; /* LTR */
     margin: -1px 0 -1px -15em; /* LTR */
   }
-   [dir="rtl"] .vertical-tabs-list {
-    margin-left: 0;
-    margin-right: -15em;
+  [dir="rtl"] .vertical-tabs-list {
     float: right;
+    margin: 0 -15em -1px 0;
+    border-top: none;
```
quicksketch: Why did the top margin change from `-1px` to `0` here? It looks like setting `border-top: none` might balance that all out, but the rules shouldn't show/hide the border based on language direction.

quicksketch (changes requested):

core/themes/seven/css/vertical-tabs.css
klonos: There already was one, further down, within the `@media (min-width: 750px)` ...moved things around.

core/themes/seven/css/vertical-tabs.css
klonos: ...oops 😅

quicksketch (approved):

PR #2494, opened 2019-01-31
Issue #3512: Test that array index '#type' is set in form_process_act…

bradbulger:
…ions().


bradbulger:
See https://github.com/backdrop/backdrop-issues/issues/3512

quicksketch:

PR #2495, opened 2019-01-31
Issue #3493: Prevent PHP warnings when classes are givein for views l…

jenlampton:
Fixes https://github.com/backdrop/backdrop-issues/issues/3493


findlabnet:
Fix from this PR helped me clean up same warning message fired from the node template by using function "views_embed_view" to print some block.

opi (approved):

quicksketch:

PR #2496, opened 2019-02-01
Issue #3517: Add a search relevancy factor for content title match.

jenlampton:
Fixes https://github.com/backdrop/backdrop-issues/issues/3517

PR #2499, opened 2019-02-03
Issue #3517: Add a search relevancy factor for content title match.

### LINKED ISSUE #3517: [UX] Add content title match as search relevancy factor
**Is your feature request related to a problem? Please describe.**

**Describe the solution you'd like**
I'd like to be able to weight a title match on the search settings page:

**Describe alternatives you've considered**
Have you seen this work in a better way in another CMS?

Is there a contributed module that accomplishes this? ...if so, then has this
been ported to Backdrop?: https://github.com/backdrop-contrib

...if not, then look here to see if anyone has already requested it to be
ported: https://github.com/backdrop-ops/contrib/issues

...if not, then create a request for it to be ported:
https://github.com/backdrop-ops/contrib/issues/new

**Additional information**

jenlampton:
Fixes backdrop/backdrop-issues#3517
replaces https://github.com/backdrop/backdrop/pull/2496

### DIFF
--- a/core/modules/node/node.module
+++ b/core/modules/node/node.module
@@ -1322,8 +1322,8 @@ function node_views_api() {
  * @param SearchQuery|SelectQueryExtender $query
  *   A query object that has been extended with the Search DB Extender.
  */
-function _node_rankings(SelectQueryExtender $query) {
-  if ($ranking = module_invoke_all('ranking')) {
+function _node_rankings(SelectQueryExtender $query, $keywords = NULL) {
+  if ($ranking = module_invoke_all('ranking', $keywords)) {
     $tables = &$query->getTables();
     $config = config('search.settings');
     foreach ($ranking as $rank => $values) {
@@ -1431,7 +1431,7 @@ function node_search_execute($keys = NULL, $conditions = NULL) {
   }
 
   // Add the ranking expressions.
-  _node_rankings($query);
+  _node_rankings($query, $keys);
 
   // Load results.
   $find = $query
@@ -1470,8 +1470,15 @@ function node_search_execute($keys = NULL, $conditions = NULL) {
 /**
  * Implements hook_ranking().
  */
-function node_ranking() {
+function node_ranking($keywords = NULL) {
   // Create the ranking array and add the basic ranking options.
+  $words = explode(' ', $keywords);
+  $conditions = array();
+  foreach($words as $word) {
+    $conditions[] = "n.title LIKE '%" . $word . "%'";
+  }
+  $sql_string = implode(' OR ', $conditions);
+
   $ranking = array(
     'relevance' => array(
       'title' => t('Keyword relevance'),
@@ -1487,6 +1494,11 @@ function node_ranking() {
       'title' => t('Content is promoted'),
       // The promote flag is either 0 or 1, which is automatically normalized.
       'score' => 'n.promote',
+    ),
+    'title_match' => array(
+      'title' => t('Content title match'),
+      // If a keyword appears in the title, return 1, else 0.
+      'score' => "IF((" . $sql_string . "), 1, 0)",
     ),
   );
 
--- a/core/modules/node/node.module
+++ b/core/modules/node/node.module
@@ -1474,7 +1474,7 @@ function node_ranking($keywords = NULL) {
   // Create the ranking array and add the basic ranking options.
   $words = explode(' ', $keywords);
   $conditions = array();
-  foreach($words as $word) {
+  foreach ($words as $word) {
     $conditions[] = "n.title LIKE '%" . $word . "%'";
   }
   $sql_string = implode(' OR ', $conditions);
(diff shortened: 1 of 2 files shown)


jenlampton:
This has the same PDO error as https://github.com/backdrop/backdrop/pull/2496:
```
PDOException: SQLSTATE[HY093]: Invalid parameter number: number of bound variables does not match number of tokens: SELECT i.type AS type, i.sid AS sid, SUM(CAST(:multiply_0 AS DECIMAL) * COALESCE(( IF((CONCAT(',', n.title, ',') REGEXP ',(:words),' > 0), '1','0')), 0) / CAST(:total_0 AS DECIMAL)) AS calculated_score FROM {search_index} i INNER JOIN {node} n ON n.nid = i.sid INNER JOIN {search_total} t ON i.word = t.word INNER JOIN {search_dataset} d ON i.sid = d.sid AND i.type = d.type WHERE (n.status = :db_condition_placeholder_0) AND( (i.word = :db_condition_placeholder_1) )AND (i.type = :db_condition_placeholder_2) AND( (d.data LIKE :db_condition_placeholder_3 ESCAPE '\\') ) GROUP BY i.type, i.sid HAVING (COUNT(*) >= :matches) ORDER BY calculated_score DESC LIMIT 10 OFFSET 0; Array ( [:db_condition_placeholder_0] => 1 [:db_condition_placeholder_1] => webform [:db_condition_placeholder_2] => node [:db_condition_placeholder_3] => % webform % [:matches] => 1 [:words] => webform [:multiply_0] => 5 [:total_0] => 5 ) in PagerDefault->execute() (line 76 of /home/qa/github/backdrop/backdrop/pr/2499/core/includes/pager.class.inc).
```

It's strange the error is about the number of placeholders in the query, but the number is identical to the number of values provided. My guess is that there's a SQL syntax error here that might be triggering this error -- but maybe only for MariaDB version  5.5.48-37.8.

edit: aha! I got it to fail on my local... working on this now.

jenlampton:
aha! I got it to fail on my local...  working on this now.

jenlampton:
here we go: http://2499.backdrop.backdrop.qa.backdropcms.org/search/node/delete%20post

core/modules/node/node.module
```diff
@@ -1470,8 +1470,15 @@ function node_search_execute($keys = NULL, $conditions = NULL) {
 /**
  * Implements hook_ranking().
  */
-function node_ranking() {
+function node_ranking($keywords = NULL) {
   // Create the ranking array and add the basic ranking options.
+  $words = explode(' ', $keywords);
+  $conditions = array();
+  foreach($words as $word) {
```
quicksketch:
```suggestion
  foreach ($words as $word) {
```

core/modules/node/node.module
```diff
@@ -1470,8 +1470,15 @@ function node_search_execute($keys = NULL, $conditions = NULL) {
...
+    $conditions[] = "n.title LIKE '%" . $word . "%'";
+  }
+  $sql_string = implode(' OR ', $conditions);
```
quicksketch, line 1480: One thing about this approach is that **any** word that matches in the title gives a score of `1`. Since the score range is a decimal, I think the more words that match, the greater the score should be, correct?

So ideally, if the query had 3 words but one matched, the score would be `0.333`. If two matched, it would be `0.666`. And if all words match it's `1`.

jenlampton: I started there too, but couldn't figure out how to do it with SQL.

quicksketch: So this gets complicated, but I think the query would end up looking like:

$score_per_word = 1 / count($keywords);

`SUM(IF((n.title LIKE '%foo%'), $score_per_word, 0), IF((n.title LIKE '%bar%'), $score_per_word, 0), IF((n.title LIKE '%baz%'), $score_per_word, 0))`

core/modules/node/node.module
```diff
@@ -1470,8 +1470,15 @@ function node_search_execute($keys = NULL, $conditions = NULL) {
 /**
  * Implements hook_ranking().
  */
-function node_ranking() {
+function node_ranking($keywords = NULL) {
   // Create the ranking array and add the basic ranking options.
+  $words = explode(' ', $keywords);
+  $conditions = array();
+  foreach ($words as $word) {
+    $conditions[] = "n.title LIKE '%" . $word . "%'";
```
quicksketch, line 1478: This looks like a prime candidate for SQL-injection. I'm not sure we can safely use user-entered values directly in a SQL query like this. We probably could however, pass placeholders like `array(':keyword1', ':keyword2', ':keyword3')`, and let the rest of the search query populate those placeholders.

quicksketch: Actually on further thought yet, that wouldn't work because parameters have to be wrapped in `db_like()` if being passed into a LIKE statement like this. Further compounding this problem this approach would not respect search's querying structure, such as doing `term1 AND term2` vs `term1 OR term2`. The AND/OR would end up being part of the query, rather than an operator as designed.

Overall I think we might be chasing the wrong approach here. Simply giving the H1 a greater weight in the index or adding the title in a different way to `_node_index_node()` might accomplish this goal just as effectively.

core/modules/node/node.module
quicksketch: Resolving this conversation since I don't think this approach is feasible.

PR #2500, opened 2019-02-06
Issue #3523: [UX] Fix RTL table headers

klonos:
https://github.com/backdrop/backdrop-issues/issues/3523


quicksketch:
Thanks, I like this approach to RTL issues, just make it always symmetrical and it results in fewer CSS rules. Thanks @klonos! Merged into 1.x and 1.12.x.

PR #2501, opened 2019-02-07
Issue #3525: [UX] Fix RTL options element

klonos:
https://github.com/backdrop/backdrop-issues/issues/3525


quicksketch:

PR #2506, opened 2019-02-11
Issue #3537: Fix error using entity functions in install files

### LINKED ISSUE #3537: Cannot call entity_create in hook_install
**Describe the bug**
If a module provides a new Entity - that is it implements `hook_entity_info()` and has an implementation of `hook_autoload_info()` to load its classes - then `entity_create()` cannot be called in the `hook_install()` for that same module.

**Steps To Reproduce**
Find any contrib module that implements `hook_entity_info()`, note the id of one of its entities (one of the keys from `hook_entity_info()`), and add `entity_create('the_entity_id', array())` to that module's install file in `hook_install()` 

**Actual behavior**
`Fatal error: Class name must be a valid object or a string in <ROOT>\core\modules\entity\entity.module...`

**Expected behavior**
No error, you should be able to create an instance of your entity in the install file.

**Additional information**
Backdrop 1.13
This error suggests that the classes defined in `hook_entity_info()` are not yet available during `module_enable()`. 

---
PR by @docwilmot: https://github.com/backdrop/backdrop/pull/2506

docwilmot:
Fixes https://github.com/backdrop/backdrop-issues/issues/3537

### DIFF
--- a/core/includes/module.inc
+++ b/core/includes/module.inc
@@ -474,6 +474,8 @@ function module_enable($module_list, $enable_dependencies = TRUE) {
       backdrop_get_schema(NULL, TRUE);
       // Update the theme registry to include it.
       backdrop_theme_rebuild();
+      // Refresh Entity info static cache.
+      entity_info_cache_clear();
 
       // Allow modules to react prior to the installation of a module.
       module_invoke_all('modules_preinstall', array($module));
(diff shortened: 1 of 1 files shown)


core/includes/module.inc
```diff
@@ -576,6 +578,9 @@ function module_disable($module_list, $disable_dependents = TRUE) {
     $module_list = array_keys($module_list);
   }
 
+  // Refresh Entity info static cache.
+  entity_info_cache_clear();
```
quicksketch, line 582: Is there a reason why this is here instead of 20 lines down, where all the other caches are cleared?

docwilmot: The cache has to be cleared before `hook_disable()` is invoked in `module_invoke($module, 'disable')`.

quicksketch:
Thanks @docwilmot! Merged into 1.x and 1.12.x.

PR #2508, opened 2019-02-12
#3249: Help Views Wizard connect to vocabulary correctly

laryn:


laryn (commented):
> I'm not sure if it's the best way to go about it, but I figured out a way to make this work. Would appreciate a code review -- it seems there was some D7 cruft in there yet and there may be further related clean up that can happen.

core/modules/views_ui/wizards/views_ui_base_views_wizard.php
```diff
@@ -687,9 +687,9 @@ protected function default_display_filters_user($form, $form_state) {
 
     if (!empty($form_state['values']['show']['type']) && $form_state['values']['show']['type'] != 'all') {
       $bundle_key = $this->entity_info['bundle keys']['bundle'];
```
quicksketch: I think we just need to be checking `entity keys` here instead of `bundle keys`, and then we'll get the right key value from `taxonomy_entity_info()`.

laryn: Thanks @quicksketch! Great improvement.

laryn:
Rebooting tests.

quicksketch:

PR #2510, opened 2019-02-13
Issue #3543: [UX] Fix RTL password field style

klonos:
https://github.com/backdrop/backdrop-issues/issues/3543

### DIFF
--- a/core/modules/user/css/user.css
+++ b/core/modules/user/css/user.css
@@ -45,10 +45,15 @@
   max-width: 100%;
 }
 .password-toggle {
-  right: 0;
   position: absolute;
-  font-size: 80%;
   top: -2em;
+  right: 0; /* LTR */
+  font-size: 80%;
+}
+[dir="rtl"] .password-toggle {
+  position: absolute;
+  right: unset;
+  left: 0;
 }
 
 /**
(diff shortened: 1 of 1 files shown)


core/modules/user/css/user.css
```diff
@@ -45,10 +45,14 @@
...
 .password-toggle {
-  right: 0;
   position: absolute;
-  font-size: 80%;
   top: -2em;
+  right: 0; /* LTR */
+  font-size: 80%;
+}
+[dir="rtl"] .password-toggle {
+  right: unset;
```
quicksketch: `unset` is relatively new (CSS 3) and doesn't have any IE support (even IE11). For the time being, could we just switch to `auto` (which is the default: https://www.w3schools.com/cssref/pr_pos_left.asp)?

core/modules/user/css/user.css
```diff
@@ -68,6 +72,18 @@
...
+  float: left; /* LTR */
+}
+.password-strength-text {
+  padding-left: 5px; /* LTR */
+}
+[dir="rtl"] .password-strength-title {
+  float: right;
+}
+[dir="rtl"] .password-strength-text {
+  padding-right: 5px; /* LTR */
```
quicksketch: The padding-left needs to be set back to 0 here.

core/modules/user/css/user.css
klonos: Agreed 👍 ...fixed 😉

core/modules/user/css/user.css
klonos: Yep, ...and that ` /* LTR */` should not be there. Also fixed.

PR #2515, opened 2019-02-21
#3550: Set maintenance mode session var correctly

laryn:


herbdool (changes requested):
I think you should set the default too like this `state_get('maintenance_mode', FALSE)`. Other than that it's a no-brainer to merge.

quicksketch:

PR #2517, opened 2019-02-21
#3470: Fix PHP Deprecated Warning with Link module  

andybroomfield:
Fix #3470

Adds variant `INTL_IDNA_VARIANT_UTS46` to the IDNA Check, if defined.
This is to prevent the PHP Deprecated warnings in PHP 7.2 and 7.3.
I've added a check to see if its defined first, which 'should' allow this to still work in PHP 5.3, however I am unable to check this.


quicksketch:

PR #2519, opened 2019-02-25
Issue #3557: [PHP 7.3] strpos explicit string needle warnings

klonos:
https://github.com/backdrop/backdrop-issues/issues/3557

https://www.drupal.org/node/3020771

When the test suite is run on PHP 7.3, there is a warning due to changes in PHP 7.3:

```
exception: [Deprecated] Line 3015 of modules/simpletest/drupal_web_test_case.php:
strpos(): Non-string needles will be interpreted as strings in the future. Use an explicit chr() call to preserve the current behavior
```

If the second parameter (needle) of the `strpos` call is an integer, PHP uses the `chr()` value of that integer, which is likely not the intended behavior. Because of this, PHP now warnings when it sees an integer. The caller is now responsible to call `chr()` if intended.


klonos:
...checked the rest of the codebase for occurrences of `strpos()` and it seems that it is only used in comments 👍

opi:
LGTM

quicksketch:
Thanks folks! Merged into 1.x and 1.12.x.

PR #2521, opened 2019-02-26
Issue #3558: [DX] Consolidate theme_debug warning

klonos:
https://github.com/backdrop/backdrop-issues/issues/3558

### DIFF
--- a/core/includes/theme.inc
+++ b/core/includes/theme.inc
@@ -2849,3 +2849,16 @@ function template_preprocess_maintenance_page(&$variables) {
     $variables['theme_hook_suggestion'] = 'maintenance_page__offline';
   }
 }
+
+/**
+ * Check if theme_debug is enabled, and throw a warning if so.
+ */
+function theme_debug_enabled_warning() {
+  $theme_debug_enabled = config_get('system.core', 'theme_debug');
+  if ($theme_debug_enabled) {
+    backdrop_set_message(t('Theme debugging is enabled. See the <a href="!url">Status report</a> page for more info.', array('!url' => url('admin/reports/status'))), 'warning', FALSE);
+  }
+  else {
+    return;
+  }
+}
--- a/core/includes/theme.inc
+++ b/core/includes/theme.inc
@@ -2849,16 +2849,3 @@ function template_preprocess_maintenance_page(&$variables) {
     $variables['theme_hook_suggestion'] = 'maintenance_page__offline';
   }
 }
-
-/**
- * Check if theme_debug is enabled, and throw a warning if so.
- */
-function theme_debug_enabled_warning() {
-  $theme_debug_enabled = config_get('system.core', 'theme_debug');
-  if ($theme_debug_enabled) {
-    backdrop_set_message(t('Theme debugging is enabled. See the <a href="!url">Status report</a> page for more info.', array('!url' => url('admin/reports/status'))), 'warning', FALSE);
-  }
-  else {
-    return;
-  }
-}
--- a/core/modules/system/system.admin.inc
+++ b/core/modules/system/system.admin.inc
@@ -105,7 +105,7 @@ function system_admin_index() {
  */
 function system_themes_page() {
   // Inform the user if theme_debug is enabled.
-  theme_debug_enabled_warning();
+  system_theme_debug_enabled_warning();
 
   // Get current list of themes.
   $themes = system_rebuild_theme_data();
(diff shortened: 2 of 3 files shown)


klonos:
This seems like a straight-forward change, but I keep getting **different** random test failures with each run. Keeping a log here for future reference:

## php70:

```
fail "Generated 20 URL aliases." found path_pattern.test:929
```

...despite the above failure being logged, the test still passes:

screen shot 2019-02-26 at 7 11 32 pm

## php53

1st run:
```
fail The Install Dependencies page was loaded installer.test:134
fail "eee_installer_test 1.1.1" found installer.test:135
fail Failed to set field maintenance_mode to installer.test:141
fail Found the Install button installer.test:141
fail Found the requested form fields at installer.test:141
fail EEE Installer test exists in the modules directory installer.test:148
fail Failed to set field modules[ddd_installer_test] to 1 installer.test:155
fail Failed to set field modules[eee_installer_test] to 1 installer.test:155
fail Found the Enable modules button installer.test:155
fail Found the requested form fields at installer.test:155
```

2nd run:
```
fail "Generated 1 URL alias." found path_pattern.test:943
```

core/includes/theme.inc
```diff
@@ -2849,3 +2849,16 @@ function template_preprocess_maintenance_page(&$variables) {
...
+
+/**
+ * Check if theme_debug is enabled, and throw a warning if so.
+ */
+function theme_debug_enabled_warning() {
+  $theme_debug_enabled = config_get('system.core', 'theme_debug');
+  if ($theme_debug_enabled) {
+    backdrop_set_message(t('Theme debugging is enabled. See the <a href="!url">Status report</a> page for more info.', array('!url' => url('admin/reports/status'))), 'warning', FALSE);
+  }
+  else {
```
opi: not needed.

core/includes/theme.inc
```diff
@@ -2849,3 +2849,16 @@ function template_preprocess_maintenance_page(&$variables) {
...
+    return;
```
opi: But maybe we could `return $theme_debug_enabled;` anyway, this could help people calling this function to have the theme debug status...

opi (approved):

core/modules/system/system.admin.inc
```diff
@@ -105,10 +105,7 @@ function system_admin_index() {
  */
 function system_themes_page() {
   // Inform the user if theme_debug is enabled.
-  $theme_debug_enabled = config_get('system.core', 'theme_debug');
-  if ($theme_debug_enabled) {
-    backdrop_set_message(t('Theme debugging is enabled. See the <a href="!url">Status report</a> page for more info.', array('!url' => url('admin/reports/status'))), 'warning', FALSE);
-  }
+  theme_debug_enabled_warning();
```
quicksketch: Any function that starts with `theme_` you would expect to be override-able by the theme system, but this is not. For clarity, I recommend we rename this to `system_theme_debug_enabled_warning()`, unless this actually is intended to be overridden by a theme, in which case we need `system_theme()` to be updated and call this using `theme()` like other theme functions and templates.

klonos: I think that `system_theme_debug_enabled_warning()` as you suggest is the right way to go @quicksketch for all the reasons. One question though: I have placed the actual function in theme.inc (because it was called `theme_`something), but I feel that it should now be moved elsewhere. Should it be in `common.inc`, or perhaps `errors.inc`, or `utility.inc`? Elsewhere?

opi: `theme.inc` is always loaded ?

herbdool (commented):
system.module is also always loaded (I think) so why not keep it in the system module for simplicity. And make sure it is named `system_theme_debug_enabled_warning()`

klonos:
...OK, I have:

- moved the function in system.module
- renamed the function to `system_theme_debug_enabled_warning()`
- removed the `else { return; }` bit

core/modules/system/system.admin.inc
klonos: ..moved the function in system.module

core/includes/theme.inc
klonos: removed it 😉

core/includes/theme.inc
klonos: I have now removed the return, but can re-add it if others think that it would help to return `$theme_debug_enabled`. ...although I think that since the function name includes `warning` in it, it might be misleading.

PR #2525, opened 2019-03-03
Issue #3519: Fix image link class being removed when using image caption

ksthomas:
https://github.com/backdrop/backdrop-issues/issues/3519


quicksketch:
Thanks @ksthomas! Merged into 1.x and 1.12.x.

PR #2535, opened 2019-03-17
Issue #3592: Change the weight field in the taxonomy term form to a number field

klonos:
https://github.com/backdrop/backdrop-issues/issues/3592


quicksketch:

PR #2541, opened 2019-03-23
Issue #3613: Support 'type = profile' in .info files

klonos:
https://github.com/backdrop/backdrop-issues/issues/3553

Alternative to https://github.com/backdrop/backdrop/pull/2540


quicksketch:

PR #2546, opened 2019-03-24
Issue #3617: [UX] Better messages and clear caches when changing the default theme

klonos:
https://github.com/backdrop/backdrop-issues/issues/3617

### DIFF
--- a/core/modules/system/system.admin.inc
+++ b/core/modules/system/system.admin.inc
@@ -273,6 +273,7 @@ function system_themes_admin_form_submit($form, &$form_state) {
   }
 
   $config->set('admin_theme', $new_admin_theme);
+  $config->save();
 }
 
 /**
--- a/core/modules/system/system.admin.inc
+++ b/core/modules/system/system.admin.inc
@@ -258,8 +258,9 @@ function system_themes_admin_form_submit($form, &$form_state) {
   backdrop_set_message(t('The configuration options have been saved.'));
 
   $themes = list_themes();
-  $current_default_theme = config_get('system.core', 'theme_default');
-  $current_admin_theme = config_get('system.core', 'admin_theme');
+  $config = config('system.core');
+  $current_default_theme = $config->get('theme_default');
+  $current_admin_theme = $config->get('admin_theme');
   $new_admin_theme = $form_state['values']['admin_theme'];
 
   if (!$new_admin_theme) {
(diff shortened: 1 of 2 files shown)


klonos:
...refreshing the sandbox

core/modules/system/system.admin.inc
```diff
@@ -322,30 +337,41 @@ function system_theme_default() {
...
+      // The status message depends on whether an admin theme is currently in
+      // use: a value of 0 means the admin theme is set to be the default theme.
       $admin_theme = config_get('system.core', 'admin_theme');
+
+      // If the admin theme is set to a different theme than the one the user
+      // has just switched to, show an info message to explain why.
       if ($admin_theme && $admin_theme != $theme) {
-        backdrop_set_message(t('Please note that the administration theme is still set to the %admin_theme theme; consequently, the theme on this page remains unchanged. All non-administrative sections of the site, however, will show the selected %selected_theme theme by default.', array(
+        backdrop_set_message(t('%theme is now the default theme for all non-administrative pages.', array('%theme' => $themes[$theme]->info['name'])));
+        backdrop_set_message(t('The administration theme is still set to the %admin_theme theme. <a href="@change_admin_theme_link">Change the administration theme', array(
```
quicksketch: IMO we should just remove this second message. This message makes it sound as though you _should_ change the admin theme, even though that's an uncommon task.

klonos: The `The administration theme is still set to ...` bit is still required, in order to let people understand that the fact that they are not seeing the new theme applied is normal/expected. ...I see how the `Change the administration theme` link may make it seem as what you describe.

We need to either come up with a better wording, or remove that link (and unfortunately assume that people know that they need to scroll to the bottom of the page to change that - if they want).

klonos: ...I have removed the `Change the administration theme` link from the info message, and made the `administration theme` bit a link to the settings at the bottom of the page instead 😉

core/modules/system/system.admin.inc
```diff
@@ -256,7 +256,22 @@ function system_themes_admin_form($form, &$form_state, $theme_options) {
  */
 function system_themes_admin_form_submit($form, &$form_state) {
   backdrop_set_message(t('The configuration options have been saved.'));
-  config_set('system.core', 'admin_theme', $form_state['values']['admin_theme']);
+
+  $themes = list_themes();
+  $current_default_theme = config_get('system.core', 'theme_default');
```
ghost: There are 3 times that the `system.core` config is used here. I think it'd be better to have a single `$config = config('system.core');` then adjust the rest accordingly (e.g. `$config->get(...)`).

ghost (changes requested):

core/modules/system/system.admin.inc
klonos: Well, only 2 of these instances were within the same function, but I have updated the PR for those 2 nevertheless. Thanks 👍

quicksketch:

PR #2562, opened 2019-04-02
Issue #2595: Prevent double-click on form submit

klonos:
https://github.com/backdrop/backdrop-issues/issues/2595

https://www.drupal.org/node/1705618

### DIFF
--- a/core/includes/form.inc
+++ b/core/includes/form.inc
@@ -3393,6 +3393,30 @@ function form_process_actions($element, &$form_state) {
 }
 
 /**
+ * Processes a form button element.
+ *
+ * @param $element
+ *   An associative array containing the properties and children of the
+ *   form button.
+ * @param $form_state
+ *   The $form_state array for the form this element belongs to.
+ *
+ * @return
+ *   The processed element.
+ */
+function form_process_button($element, &$form_state) {
+  // We normally want to add backdrop.form so that the double submit protection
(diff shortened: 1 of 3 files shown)


core/includes/form.inc
```diff
@@ -3392,6 +3392,30 @@ function form_process_actions($element, &$form_state) {
...
+ * @return
+ *   The processed element.
+ */
+function form_process_button($element, &$form_state) {
+  // We normally want to add backdrop.form so that the double submit protection
+  // can be added to the site. However, with the addition of
+  // javascript_always_use_jquery, this would make most pages with a login
+  // block or a search form have jquery always added, changing what people who
+  // enabled javascript_always_use_jquery would have expected.
+  if (variable_get('javascript_always_use_jquery', TRUE)) {
```
quicksketch, line 3413: `variable_get()` is deprecated and I think we don't even have this setting anyway in Backdrop. I think just removing this check for the time being may be appropriate.

klonos: Sorry I missed this @quicksketch ...I've left a comment on the issue. I believe that we should hold this off till 1.13, and implement the missing pieces first. Then we should implement this as a proper `config_get()`/`settings_get()`.

quicksketch:
Replaced by https://github.com/backdrop/backdrop/pull/2595

PR #2568, opened 2019-04-06
Issue #3657: Add vertical tabs to the taxonomy term edit form.

### LINKED ISSUE #3657: [UX] Add vertical tabs to the taxonomy term edit form
I find it confusing that on node edit forms my Metatags, URL settings, and Redirects are nicely organized into vertical tabs, but on taxonomy term edit forms they are all in separate fieldsets. I think we should make the UI consistent across all edit forms.

Term after:
Screen Shot 2019-04-06 at 3 37 34 PM

Term before:
Screen Shot 2019-04-06 at 2 33 36 PM

Node, for comparison: 
Screen Shot 2019-04-06 at 2 33 05 PM

To get to the term edit form on the sandbox to test, go to http://2572.backdrop.backdrop.qa.backdropcms.org/taxonomy/term/3/edit

---
~PR: https://github.com/backdrop/backdrop/pull/2568~
PR: https://github.com/backdrop/backdrop/pull/2572

jenlampton:
Fixes https://github.com/backdrop/backdrop-issues/issues/3657

### DIFF
--- a/core/modules/taxonomy/js/taxonomy.term.js
+++ b/core/modules/taxonomy/js/taxonomy.term.js
@@ -8,10 +8,19 @@ Backdrop.behaviors.Term = {
     // Relation settings.
     $context.find('#edit-relations').backdropSetSummary(function(context) {
       var vals = [];
-
       var info = Backdrop.t('No relations.');
-      if (parents = $(context).find('select[name="parent[]"] option:selected').text()) {
-        var info = 'Parent: ' + Backdrop.checkPlain(parents);
+
+      // Get a list of all selected options and concatenate with a comma.
+      var parents = $.map(
+        $context.find('select[name="parent[]"] option:selected'),
+        function(element) {
+          // Remove leading hyphens on indented terms.
+          return $(element).text().replace(/^(\-)+/, '');
+        })
+        .join(', ');
+
+      if (parents) {
+        info = Backdrop.t('Parents: @parents', { '@parent': parents });
       }
 
       vals.push(info);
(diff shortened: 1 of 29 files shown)


core/modules/taxonomy/js/taxonomy.term.js
```diff
@@ -0,0 +1,35 @@
...
+  attach: function (context) {
+    var $context = $(context);
+
+    // Relation settings.
+    $context.find('#edit-relations').backdropSetSummary(function(context) {
+      var vals = [];
+
+      var info = Backdrop.t('No relations.');
+      if (parents = $(context).find('select[name="parent[]"] option:selected').text()) {
+        var info = 'Parent: ' + Backdrop.checkPlain(parents);
```
quicksketch: Needs to be using `Backdrop.t('Parents: @parents', { '@parents': parents })` here. (Not sure if that syntax is 100% correct, so this isn't a GitHub suggestion).

klonos: This works for me:

```suggestion
        var info = Backdrop.t('Parent: @parents', {'@parents': Backdrop.checkPlain(parents)});
```

quicksketch: The use of the `@` prefix makes it escaped already, so there's no need to keep the `checkPlain()` call.

quicksketch: I ran into another issue where multiple parents weren't being handled correctly and instead all the parents were being concatenated without any separator. I pushed up an update that should fix both the translation issue and the multiple parents issue.

quicksketch:
Dang, I think I sort of broke this PR when I merged in 1.x. I've made a new PR at https://github.com/backdrop/backdrop/pull/2572 with my merge and fix to the parents vertical tab summary.

PR #2578, opened 2019-04-14
Issue #3673: improve documentation of settings.local.php in settings.php

klonos:
https://github.com/backdrop/backdrop-issues/issues/3673


quicksketch:
Looks good! Because we have to note if settings.php changes each release, I've only merged this into 1.x to minimize changes in release notes for the bug fix versions.

PR #2579, opened 2019-04-14
Issue #3665: New layout templates *installed manually* require theme cache clear

klonos:
https://github.com/backdrop/backdrop-issues/issues/3665


core/modules/layout/layout.admin.inc
```diff
@@ -584,9 +584,14 @@ function layout_settings_form_update_layout($form, &$form_state) {
...
   }
   if (isset($form_state['values']['layout_template'])) {
-    $new = $form_state['values']['layout_template'];
-    $region_mapping = layout_get_legacy_region_mapping($layout->layout_template, $new);
-    $layout->setLayoutTemplate($form_state['values']['layout_template'], $region_mapping);
+    $old_template = $layout->layout_template;
+    $new_template = $form_state['values']['layout_template'];
+    if ($old_template != $new_template) {
+      $region_mapping = layout_get_legacy_region_mapping($old_template, $new_template);
+      $layout->setLayoutTemplate($form_state['values']['layout_template'], $region_mapping);
```
docwilmot: why not `$layout->setLayoutTemplate($new_template, $region_mapping)`?

klonos: Indeed, why not. Fixed1 😉

klonos:
...the usual false-positive `fail comment-new class not found. comment.test:628`

core/modules/layout/layout.admin.inc
```diff
@@ -589,9 +589,14 @@ function layout_settings_form_update_layout($form, &$form_state) {
...
   }
   if (isset($form_state['values']['layout_template'])) {
-    $new = $form_state['values']['layout_template'];
-    $region_mapping = layout_get_legacy_region_mapping($layout->layout_template, $new);
-    $layout->setLayoutTemplate($form_state['values']['layout_template'], $region_mapping);
+    $old_template = $layout->layout_template;
+    $new_template = $form_state['values']['layout_template'];
+    if ($old_template != $new_template) {
+      $region_mapping = layout_get_legacy_region_mapping($old_template, $new_template);
+      $layout->setLayoutTemplate($new_template, $region_mapping);
```
docwilmot, line 596: Perhaps, as a compromise, instead we could simply set a message here saying Backdrop detected a template change, please flash your theme cache if you just added a new template and your regions are not showing what you expect. Personally I prefer to flush caches automatically. 80% would UX benefit surely.

docwilmot: Or better still, is there any way to check if there is a theme entry for the new template and then we'd know for sure that this is the issue, then just explicitly message "you need to flush your theme cache" in nicer words of course.

klonos:
Closing in favor of the merged https://github.com/backdrop/backdrop/pull/3107

PR #2590, opened 2019-04-19
Views: Add class to table in table format

klonos:
https://www.drupal.org/node/1689472

https://git.drupalcode.org/project/views/commit/f3ce4242c2697bf17b1d8213a93c555e37c5921a

### DIFF
--- a/core/modules/views/templates/views.theme.inc
+++ b/core/modules/views/templates/views.theme.inc
@@ -453,7 +453,7 @@ function template_preprocess_views_view_table(&$variables) {
   $variables['rows'] = array();
   $variables['field_classes'] = array();
   $variables['header'] = array();
-  $variables['classes_array'] = array();
+  $variables['classes'] = array();
 
   $options = $view->style_plugin->options;
   $handler = $view->style_plugin;
@@ -470,7 +470,7 @@ function template_preprocess_views_view_table(&$variables) {
       foreach ($classes as $key => $class) {
         if (!empty($class)) {
           $classes[$key] = backdrop_clean_css_identifier($class, $filter);
-          $vars['classes_array'][] = $classes[$key];
+          $variables['classes'][] = $classes[$key];
         }
       }
     }
--- a/core/modules/views/templates/views.theme.inc
+++ b/core/modules/views/templates/views.theme.inc
@@ -460,7 +460,11 @@ function template_preprocess_views_view_table(&$variables) {
 
   if (!empty($handler->options['class'])) {
     $classes = explode(' ', $handler->options['class']);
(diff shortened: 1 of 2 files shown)


herbdool:
@klonos doesn't work for me. No class is visible

klonos:
@jenlampton you have recently worked on tables/grid for views. Can you help or provide any pointers?

core/modules/views/templates/views.theme.inc
```diff
@@ -453,10 +453,25 @@ function template_preprocess_views_view_table(&$variables) {
...
   $variables['field_classes'] = array();
   $variables['header'] = array();
+  $variables['classes_array'] = array();
 
   $options  = $view->style_plugin->options;
   $handler  = $view->style_plugin;
 
+  if (!empty($handler->options['class'])) {
+    $classes = explode(' ', $handler->options['class']);
+    $classes = array_map('views_clean_css_identifier', $classes);
```
jenlampton: This function doesn't exist in backdrop. Change this to `backdrop_clean_css_identifier()` and see if that works?

jenlampton: Related: 
[Add back views_clean_css_identifier as a wrapper for backdrop_clean_css_identifier](https://github.com/backdrop/backdrop-issues/issues/3716)

klonos: Thanks @jenlampton ...but hmmm 🤔 ...`backdrop_clean_css_identifier()` is in place of `drupal_clean_css_identifier()`, so wondering what the difference between that and `views_clean_css_identifier()` is. Will look into it.

jenlampton: I noted it in the referenced issue.

klonos: I have changed this to use `backdrop_clean_css_identifier()`, but was not smart enough to figure how to use `array_walk()` (which allows to pass a parameter to the callback function), so I have used a `foreach` instead. I then merged that with the `foreach` that was just below, as they both seemed to be iterating over the same array, and using the same key/values.

I hope I have done things right.

klonos:
> doesn't work for me. No class is visible

@herbdool there were 2 issues:

- a stray `$vars` array instead of `$variables`
- the fact that we use `classes` for the array key, whereas Views 7.x uses `classes_array`

Fixed those on my local. Pushing...

klonos:
Shouldn't we keep parity with Views 7.x and use `$variables['classes_array']` instead of `$variables['classes']` everywhere in `views.theme.inc`?

jenlampton:
No that already been changed everywhere in Core from D7.

PR #2593, opened 2019-04-20
Issue #3609 - removing duplicate content on field date description

Wylbur:
Simpl UI change, just removing the desription which duplicates the label above the date select elements.  This change just sets the label to '', which keeps the remaining description logic intact.  

I've attached an image of the field description in action after the change.  

Screenshot_2019-04-20_14-19-38

### DIFF
--- a/core/modules/date/date.admin.inc
+++ b/core/modules/date/date.admin.inc
@@ -492,7 +492,7 @@ function _date_field_settings_form($field, $instance, $has_data) {
   }
   $tz_handling = $settings['tz_handling'];
 
-  $description = t('Select the date attributes to collect and store.');
+  $description = t('');
   if ($has_data) {
     $description .= ' ' . t('Changes to date attributes only effects new or updated content.');
   }
--- a/core/modules/date/date.admin.inc
+++ b/core/modules/date/date.admin.inc
@@ -492,7 +492,7 @@ function _date_field_settings_form($field, $instance, $has_data) {
   }
   $tz_handling = $settings['tz_handling'];
 
-  $description = t('');
+  $description = t();
   if ($has_data) {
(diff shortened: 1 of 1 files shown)


core/modules/date/date.admin.inc
```diff
@@ -492,7 +492,7 @@ function _date_field_settings_form($field, $instance, $has_data) {
   }
   $tz_handling = $settings['tz_handling'];
 
-  $description = t('Select the date attributes to collect and store.');
+  $description = t('');
   if ($has_data) {
     $description .= ' ' . t('Changes to date attributes only effects new or updated content.');
```
serundeputy: @Wylbur I think we should remove the `' ' . ` part of this now that the description is an empty string.

Wylbur: So just change it to empty function 
```t();```

Comment (serundeputy) [reply core/modules/date/date.admin.inc]: on line 497; remove the first part that adds a space and concatenates; so it will just be 

```php
$description .= t('Changes to date attributes only effects new or updated content.');
```

Wylbur: @serundeputy I changed to empty t(); but now it fails tests.

matthui:
Code syntax looks good.

core/modules/date/date.admin.inc
jenlampton: @Wylbur you can remove the empty `t()` and instead just set the variable to an empty string. Then, if there's data, overwrite it, like this:
```
$description = '';
if ($has_data) {
  $description = t('Changes to date.....');
}
```

quicksketch:
Thanks all! Merged into 1.x and 1.12.x.

PR #2602, opened 2019-04-20
2456 changed layout content saved to layout saved

### LINKED ISSUE #2456: [UX] Status message displayed when saving a layout is confusing
When I save a layout I see the message `Layout content saved.`  Did I save the layout? or did I jusrt save content somehow? 

We should remove the word "content" from this message. We could add the word "configuration" if necessary.

isabelhaas:
Fixed https://github.com/backdrop/backdrop-issues/issues/2456


isabelhaas:
Closed because I forgot to change the label in the tests so some of the tests fail, will do another pull request with that fixed.

PR #2606, opened 2019-04-21
2456 layout saved label

### LINKED ISSUE #2456: [UX] Status message displayed when saving a layout is confusing

isabelhaas:
Fixes https://github.com/backdrop/backdrop-issues/issues/2456


herbdool (changes requested):
All DS_Store files need to be removed from this PR. Otherwise looks good.

PR #2610, opened 2019-04-22
Issue #3700: php 7.3: session_id() cannot be changed after session is started

klonos:
https://github.com/backdrop/backdrop-issues/issues/3700


quicksketch:

PR #2160, opened 2018-04-30
Hide comments section when no comments

quicksketch:
Fixes https://github.com/backdrop/backdrop-issues/issues/2926

Replaces https://github.com/backdrop/backdrop/pull/2048

PR #2617, opened 2019-04-27
Issue #2160: Adding a background hover effect to improve usability of…

hosef:
… Bartik.

### DIFF
--- a/core/themes/bartik/color/color.inc
+++ b/core/themes/bartik/color/color.inc
@@ -16,7 +16,6 @@ $info = array(
     'text' => t('Text color'),
     'link' => t('Link color'),
     'menu' => 'Menu link',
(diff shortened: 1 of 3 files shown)


core/themes/bartik/color/color.inc
```diff
@@ -16,7 +16,6 @@ $info = array(
     'text' => t('Text color'),
     'link' => t('Link color'),
     'menu' => 'Menu link',
-    'activemenu' => t('Active menu link color'),
```
jenlampton, line 19: Was this removal intended?

hosef: Yes. I forgot to mention it in my comment, but the theme doesn't actually use that setting, so I removed it.

PR #2626, opened 2019-04-29
Cache views plugin data

klonos:
https://www.drupal.org/node/1376400

https://git.drupalcode.org/project/views/commit/580ff35b315b8cf0eed6e33ab97e985634f829b1

https://github.com/backdrop/backdrop-issues/issues/3719

### DIFF
--- a/core/modules/views/includes/cache.inc
+++ b/core/modules/views/includes/cache.inc
@@ -127,8 +127,25 @@ function _views_data_process_entity_types(&$data) {
 function _views_fetch_plugin_data($type = NULL, $plugin = NULL, $reset = FALSE) {
   static $cache = NULL;
   if (!isset($cache) || $reset) {
-    views_include_handlers();
-    $cache = views_discover_plugins();
+    // Load necessary code once.
+    if (!isset($cache)) {
+      views_include('plugins');
+      views_include_handlers();
+    }
+    // Because plugin data contains translated strings, and as such can be
+    // expensive to build, the results are cached per language.
+    global $language;
+    $cache_key = 'views:plugin_data:' . $language->langcode;
+    if (!$reset) {
+      if ($cache = cache_get($cache_key)) {
+        $cache = $cache->data;
+      }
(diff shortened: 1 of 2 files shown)


core/modules/views/includes/cache.inc
```diff
@@ -127,8 +127,25 @@ function _views_data_process_entity_types(&$data) {
...
-    $cache = views_discover_plugins();
+    // Load necessary code once.
+    if (!isset($cache)) {
+      views_include('plugins');
+      views_include_handlers();
+    }
+    // Because plugin data contains translated strings, and as such can be
+    // expensive to build, the results are cached per language.
+    global $language;
+    $cache_key = 'views:plugin_data:' . $language->langcode;
```
klonos, line 138: Wondering if we should be checking if `$language->langcode` is set here 🤔

quicksketch: It is always set  on a full bootstrap. I think this is safe.

quicksketch:
Not sure why this one is failing tests. I restarted it to check.

core/modules/views/includes/cache.inc
```diff
@@ -127,8 +127,25 @@ function _views_data_process_entity_types(&$data) {
 function _views_fetch_plugin_data($type = NULL, $plugin = NULL, $reset = FALSE) {
   static $cache = NULL;
   if (!isset($cache) || $reset) {
-    views_include_handlers();
-    $cache = views_discover_plugins();
+    // Load necessary code once.
+    if (!isset($cache)) {
+      views_include('plugins');
```
quicksketch, line 132: I think this line is unnecessary and might be the reason for the failure?

```suggestion
```

quicksketch:
Gah, I accidentally merged this one even though it's failing tests. I've reverted it (https://github.com/backdrop/backdrop/pull/2801) and I'll merge https://github.com/backdrop/backdrop/pull/2800 instead.

PR #2629, opened 2019-04-29
Typo in the description text on date fields

klonos:
https://github.com/backdrop/backdrop-issues/issues/3609

Follow-up to https://github.com/backdrop/backdrop/pull/2593, to fix the typo.

### DIFF
--- a/core/modules/date/date.admin.inc
+++ b/core/modules/date/date.admin.inc
@@ -494,7 +494,7 @@ function _date_field_settings_form($field, $instance, $has_data) {
 
   $description = '';
   if ($has_data) {
-    $description = t('Changes to collected date attributes affect only new or updated content.');
+    $description = t('Changes to date granularity affect only new or updated content.');
   }
   $options = date_granularity_names();
   $checkbox_year = array(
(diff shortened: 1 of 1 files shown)


core/modules/date/date.admin.inc
```diff
@@ -494,7 +494,7 @@ function _date_field_settings_form($field, $instance, $has_data) {
 
   $description = '';
   if ($has_data) {
-    $description = t('Changes to date attributes only effects new or updated content.');
+    $description = t('Changes to collected date attributes affect only new or updated content.');
```
klonos:
```suggestion
    $description = t('Changes to date granularity affect only new or updated content.');
```

matthui:
The code syntax looks correct, but I think the grammar in this fix (`Changes to date granularity`) is not as good as #2630 since its only stating a fact rather than providing instruction.

However, I can't locate where the fix is on the website, so I don't fully know the context where this is used in.

quicksketch:
Replaced by https://github.com/backdrop/backdrop/pull/2630.

PR #2630, opened 2019-04-29
Typo in the description text on date fields

klonos:
https://github.com/backdrop/backdrop-issues/issues/3609

Follow-up to https://github.com/backdrop/backdrop/pull/2593, to fix the typo.

Alternative to https://github.com/backdrop/backdrop/pull/2629

### DIFF
--- a/core/modules/date/date.admin.inc
+++ b/core/modules/date/date.admin.inc
@@ -494,7 +494,7 @@ function _date_field_settings_form($field, $instance, $has_data) {
 
   $description = '';
   if ($has_data) {
-    $description = t('Changes to date attributes only effects new or updated content.');
+    $description = t('Changing collected date attributes will affect only new or updated content.');
   }
   $options = date_granularity_names();
   $checkbox_year = array(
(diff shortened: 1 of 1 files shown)


core/modules/date/date.admin.inc
```diff
@@ -494,7 +494,7 @@ function _date_field_settings_form($field, $instance, $has_data) {
 
   $description = '';
   if ($has_data) {
-    $description = t('Changes to date attributes only effects new or updated content.');
+    $description = t('Changing collected date attributes will affect only new or updated content.');
```
klonos:
```suggestion
    $description = t('Changing date granularity will affect only new or updated content.');
```

matthui:
The code syntax looks correct, and I think the grammar in this fix (`Changing date granularity`) is better than #2629 , as it is more instructive/active, rather than plainly stating a fact.

However, I can't locate where the fix is on the website, so I don't fully know the context where this is used in.

quicksketch:
Agreed, thanks @matthui! Merged this into 1.x and 1.12.x. Thanks!

PR #2524, opened 2019-03-03
495 dashboard

jenlampton:
Fixes https://github.com/backdrop/backdrop-issues/issues/495

Replaces https://github.com/backdrop/backdrop/pull/2452

PR #2632, opened 2019-04-30
Add Dashboard to core

quicksketch:
Replaces https://github.com/backdrop/backdrop/pull/2524

In-progress work does a few things differently:

- Added constructors to all the new blocks so that we can define defaults rather than checking isset() all over the place. (I didn't get to DashboardCreateBlock, that still needs this change)
- Fixed a number of XSS holes.
- Added a new Layout context for "dashboard" so that multiple dashboards can be created and still have all the behaviors attached, such as the CSS file and class additions.


jenlampton:
In my initial testing, this works as expected. The only recommended change is to make it enabled for new sites. I'll do a quick code review next.

jenlampton (commented):
Looks pretty good! I found a few issues with the config forms for blocks not saving, or having their saved values not respected. Pushing up an update now.

jenlampton:
Replaced by https://github.com/backdrop/backdrop/pull/2633

PR #2641, opened 2019-05-02
#3745: Access arguments for Dashboard menus

laryn:


laryn:
Superceded by https://github.com/backdrop/backdrop/pull/2681

stpaultim:
Does "superceded" mean "fixed" in this case? @laryn

laryn:
Yes, it was fixed in an alternate way and that alternate way was merged.

PR #2643, opened 2019-05-03
#3751: Trying to get property of non-object

laryn:


quicksketch:
Merged into 1.x and 1.13.x.

PR #2646, opened 2019-05-04
Issue #3755: Flexible templates

docwilmot:
https://github.com/backdrop/backdrop-issues/issues/3755


docwilmot:
This is a rough draft. The code is in a separate inc file for convenience; it would need to be merged into `layout.admin.inc`. 
To use, go to `admin/structure/layouts/settings` and add a new template, then configure it, then save it, then use it as an ordinary layout template.

docwilmot:
Resetting

docwilmot:
Sandbox broken

jenlampton:
The add-row dialog is getting there :) Some feedback:

Screen Shot 2019-06-15 at 5 41 43 PM

* The 'Region titles' fieldset should be renamed `Region names`
* The 'Region names' fieldset should be moved up on the form above classes, wrapper, and behavior (so that the form elements are in approximate order of importance with the things that are most likely to go unused at the bottom of the form).
* "Region name" numbering should start with 1 and not zero
* The Human-readable names for regions should never be random hashes (even the machine names should be region_1, region_2, region_3, etc).
* The Human-readable region names should default to "Region 1", "Region 2", "Region 3" 
* Placeholder text for Human-readable region names is unnecessary (as long as good defaults are provided)

Hm... The current PR isn't working for me, I get only a single region in each row regardless of my settings.

klonos:
I agree with @jenlampton's feedback.

> - Placeholder text for Human-readable region names is unnecessary (as long as good defaults are provided)

Not sure if we are thinking the same thing here, but when I was testing this, having the auto-generated names be placeholder text made the UX odd. I expected those to be default, editable values, but I could not "edit" or "delete" them. It felt as if the field was locked/disabled. I think that these should be actual default text values, that the user can edit/delete, instead of placeholder text. The placeholder text should be something like "this cannot be left blank" ...in case the user for some reason deletes the region name(s).

But perhaps @jenlampton means the same thing.

jenlampton:
Yep!

herbdool:
Looking good! My bit of feedback:

After entering name it should redirect to configuring the regions. Instead it goes back to the main list.

On mobile it hides the icon and name, making the list useless. I haven't checked if this is core to the theme or in this PR.

Probably shouldn't let people change the machine name lest they break integrations. Views and fields don't let you edit the machine name.

core/modules/layout/plugins/renderers/layout_renderer_flexible.inc
```diff
@@ -0,0 +1,119 @@
...
+
+  /**
+   * {@inheritdoc}
+   */
+  function renderRegion($region_id, $blocks) {
+    $output = '';
+    if ($region_id == 'title') {
+      return '';
+    }
+    else {
```
herbdool, line 113: Don't need an "else" here since it'll skip anyway with the return just above this.

Generally it's good to avoid using else and ifelse to help flatten the logic. The first "if"  is like a gate. If we pass it we're good to go on.

core/modules/layout/layout.admin.inc
```diff
@@ -2730,10 +2775,46 @@ function layout_settings_page_validate(&$form, &$form_state) {
...
+function _layout_settings_get_group_operations($template_info, $is_default) {
+  $links = array();
+
+  if ($is_default) {
+    $links['disable'] = array(
+      'title' => t('Disable'),
+      'href' => 'admin/structure/layouts/',
+    );
+  }
+  else {
```
herbdool: This can go at the beginning without the "else". It'll get overwritten by the if statement which is what we want.

core/modules/layout/layout.flexible.inc
```diff
@@ -0,0 +1,1030 @@
...
+  $form_state['region_style'] = $region_style;
+
+  $row_styles = layout_flexible_row_styles();
+  $selected_style = $row_styles[$region_style]['name'];
+  $region_count = $row_styles[$region_style]['region_count'];
+
+  if ($original_row == 'add') {
+    backdrop_set_title(t('Add new row'));
+  }
+  else {
```
herbdool, line 460: This can go before the if statement without the else.

core/modules/layout/layout.flexible.inc
```diff
@@ -0,0 +1,1030 @@
...
+
+  if (backdrop_is_ajax()) {
+    $commands[] = ajax_command_remove('#flexible-row-id-' . $original_row);
+
+    return array(
+      '#type' => 'ajax',
+      '#commands' => $commands,
+    );
+  }
+  else {
```
herbdool, line 758: This can be above with logic `if (! backdrop_is_ajax())`.

core/modules/layout/plugins/renderers/layout_renderer_flexible.inc
docwilmot: Thanks, this was a modification: everything after the else is actually unnecessary, I just havent cleaned up yet.

core/modules/layout/layout.admin.inc
docwilmot: Not really, the two keys are different ('enable' and 'disable'), so the wouldnt overwrite.

core/modules/layout/layout.flexible.inc
docwilmot: I would prefer an if/else if each block called a function. We'd be calling `backdrop_set_title()` twice.

core/modules/layout/layout.flexible.inc
docwilmot: It would still require an else statement if the negation came first.

herbdool (commented):
Just want to mention where I got the idea of simplifying if else statements @docwilmot. From https://carlalexander.ca/mastering-php-conditionals/ he describes "guard clauses" and returning early rather than nesting. And from here https://laracasts.com/series/simple-rules-for-simpler-code about not using else statements if we can help it. Sadly, the episode isn't free.

core/modules/layout/layout.flexible.inc
herbdool: Or even better just remove the else conditional. It has an early return which already serves the same purpose.

core/modules/layout/layout.flexible.inc
herbdool: But you don't need to call it twice. Create a `$title = ...` and then use that in `backdrop_set_title()`.

core/modules/layout/layout.flexible.inc
```diff
@@ -0,0 +1,1030 @@
...
+ * @param $new_key
+ *   The key to insert.
+ * @param $new_value
+ *   An value to insert.
+ *
+ * @return
+ *   The new array if the key exists, FALSE otherwise.
+ */
+function layout_flexible_array_insert_key_value($key, array &$array, $new_key, $new_value, $position) {
+  if (array_key_exists($key, $array)) {
```
herbdool: This one has many layers of conditionals. We should try to flatten it as much as possible.

core/modules/layout/layout.flexible.inc
docwilmot: > It has an early return

True. This is just a copy and paste though. This repeats a few times in `layout.admin`

core/modules/layout/layout.flexible.inc
docwilmot: Yes but we dont gain anything from that, we still require a conditional and introduce an unnecessary variable.

docwilmot:
Thanks @herbdool I agree with the principles. I was all for simplifying when I saw your comments first time, but on review, I'm not sure these principles necessarily apply in all the codeblocks you commented on.

docwilmot:
Rebooting

docwilmot:
This is quite broken now since the last commit.

core/modules/layout/layout.admin.inc
```diff
@@ -2653,75 +2654,86 @@ function layout_ajax_form_update($form, $form_state) {
...
     }
   }
 
-  $form['templates'] = array(
-    '#title' => t('Available layout templates'),
-    '#description' => t('Only templates that are selected here will be available when configuring layouts.'),
-    '#type' => 'checkboxes',
+  $form['#attached']['css'][] = backdrop_get_path('module', 'layout') . '/css/layout.admin.css';
+  $form['table'] = array(
+    '#type' => 'tableselect',
```
quicksketch, line 2746: I'm not sure tableselect is the right tool here? Would you ever want to disable all layout templates at the same time? This is sort of similar to the modules page, where a select-all checkbox might not make sense.

It's also not clear based on the current form what the checkbox even does. Maybe replacing the select-all checkbox with just a label for enabled would be more clear:

image

core/modules/layout/layout.admin.inc
```diff
@@ -2653,75 +2654,86 @@ function layout_ajax_form_update($form, $form_state) {
...
 /**
  * Validation for layout settings page.
  *
  * Ensures that active layouts are not disabled while in use.
  */
 function layout_settings_page_validate(&$form, &$form_state) {
-  $options = array_keys($form_state['values']['templates']);
-  $included = array_values($form_state['values']['templates']);
+  $options = array_keys($form_state['values']['table']);
+  $included = array_values($form_state['values']['table']);
```
quicksketch: If we could leave this as `templates` that lowers risk for other modules that may have form_alter'd this form (unlikely as that may be).

core/modules/layout/layout.module
```diff
@@ -90,6 +92,105 @@ function layout_menu() {
...
+    'page callback' => 'backdrop_get_form',
+    'page arguments' => array('layout_flexible_template_edit_row_form', 5, 7, 9),
+    'type' => MENU_CALLBACK,
+    'weight' => -10,
+    'access arguments' => array('administer layouts'),
+    'file' => 'layout.flexible.inc',
+    'theme callback' => 'ajax_base_page_theme',
+  );
+
+  $items['admin/structure/layouts/settings/flexible-template-enable/%/%'] = array(
```
quicksketch: This seems fine but I would suggest making two menu entries at the same `/flexible-template/` path.

```
   $items['admin/structure/layouts/settings/flexible-template/%/enable'] = array(
...
   $items['admin/structure/layouts/settings/flexible-template/%/disable'] = array(
```

core/modules/layout/layout.module
```diff
@@ -90,6 +92,105 @@ function layout_menu() {
...
+    'title' => 'Add flexible layout template',
+    'page callback' => 'backdrop_get_form',
+    'page arguments' => array('layout_flexible_template_settings_form'),
+    'type' => MENU_LOCAL_ACTION,
+    'weight' => -10,
+    'access arguments' => array('administer layouts'),
+    'file' => 'layout.flexible.inc',
+  );
+
+  $items['admin/structure/layouts/settings/flexible-template/%/edit'] = array(
```
quicksketch: In all cases where a flexible layout template is being loaded, using a magic menu loader might be better in each menu path.
```suggestion
  $items['admin/structure/layouts/settings/flexible-template/%layout_flexible_template/edit'] = array(
```

That will automatically load the flexible template into the menu callback instead of getting the `$flexible_template_id`, it will be the `$flexible_template` object. It will also automatically make any paths that are not a valid template ID return a 404 not found if the ID does not exist.

core/modules/layout/layout.flexible.inc
```diff
@@ -0,0 +1,1118 @@
...
+    '#maxlength' => 128,
+    '#default_value' => $template_data['title'],
+    '#required' => TRUE,
+  );
+  $form['machine_name'] = array(
+    '#type' => 'machine_name',
+    '#maxlength' => 21,
+    '#default_value' => $template_data['name'],
+    '#machine_name' => array(
+      'exists' => 'layout_flexible_template_name_exists',
```
quicksketch: Looks like in other places we just use the loader function instead of making a wrapper for `_exists()`.

```suggestion
      'exists' => 'layout_flexible_template_load',
```

core/modules/layout/layout.flexible.inc
```diff
@@ -0,0 +1,1118 @@
...
+    // Check which layout templates are in use.
+    foreach (layout_load_all() as $layout) {
+      $info = layout_get_layout_template_info($layout->layout_template);
+      if (in_array($layout->layout_template, $data['excluded_templates'])) {
+        $errors = TRUE;
+        break;
+      }
+    }
+    if ($errors) {
+      backdrop_set_message(t('The "@layout" layout template is currently in use and may not be excluded.', array('@layout' => $info['title'])), 'error');
```
quicksketch:
```suggestion
      backdrop_set_message(t('The "@layout" layout template is currently in use and may not be disabled.', array('@layout' => $info['title'])), 'error');
```

core/modules/layout/layout.flexible.inc
```diff
@@ -0,0 +1,1118 @@
...
+
+  $template_data['rows'][$original_row] = $new_row;
+
+  $form_state['template_data'] = $template_data;
+
+  $form_state['ajax_rebuild_form'] = 'layout_flexible_template_configure_form';
+  $form_state['ajax_rebuild_args'] = array($form_state['flexible_template_id']);
+  $form_state['ajax_update'] = array('content');
+  layout_set_flexible_tempstore($template_data, $form_state['flexible_template_id']);
+  layout_flexible_reset_caches();
```
quicksketch, line 671: Upon saving this form without JavaScript enabled (needed for automated testing). I don't get redirected back to the layout template configuration form.

core/modules/layout/layout.flexible.inc
```diff
@@ -0,0 +1,1118 @@
...
+  $new_row = array(
+    'contains' => $form_state['region_style'],
+    'element' => $form_state['values']['element'],
+    'classes' => $form_state['values']['row_classes'],
+    'container' => $form_state['values']['container'],
+    'region_names' => $form_state['values']['region_names'],
+  );
+
+  if ($original_row == 'add') {
+    $original_row = uniqid();
```
quicksketch: Maybe use a UUID instead of `uniqid()`? UUIDs are guaranteed to be unique with no conflicts and it what we use in some other places, e.g. the UUID for placed blocks.

```
    $uuid = new Uuid();
    $original_row = $uuid->generate();
```

core/modules/layout/layout.admin.inc
docwilmot: Done. Converted to plain table and an enabled indicator.

docwilmot: On the other hand, reverted this back to a tableselect but without the select all checkbox. We'd lose the ability to enable/disable multiple templates at once unless we had checkboxes. If its unclear might need a `#description`?

core/modules/layout/layout.admin.inc
docwilmot: Changed back.

core/modules/layout/layout.module
docwilmot: Done

core/modules/layout/layout.module
docwilmot: This was the plan all along. Done.

core/modules/layout/layout.flexible.inc
docwilmot: Done

core/modules/layout/layout.flexible.inc
docwilmot: Done

core/modules/layout/layout.flexible.inc
docwilmot: Uniqids are shorter, and looked better as array keys. But agreed nonetheless. Done.

docwilmot:
Changes made as above. Rebooting as sandbox is broken

core/modules/layout/layout.admin.inc
```diff
@@ -2653,54 +2654,66 @@ function layout_ajax_form_update($form, $form_state) {
...
-    '#type' => 'checkboxes',
+    '#type' => 'tableselect',
+    '#js_select' => FALSE,
+    '#header' => $header,
+    '#options' => $rows,
+    '#empty' => t('There are no layout templates.'),
     '#default_value' => $default,
-    '#options' => $options,
-    '#wrapper_attributes' => array('class' => array('clearfix', 'layout-options')),
+    '#attributes' => array('class' =>   array('layout-list')),
```
klonos:
```suggestion
    '#attributes' => array('class' => array('layout-list')),
```

core/modules/layout/layout.theme.inc
```diff
@@ -244,6 +297,66 @@ function layout_preprocess_layout(&$variables) {
       $variables['tabs'] =  menu_local_tabs();
```
klonos:
```suggestion
      $variables['tabs'] = menu_local_tabs();
```

core/modules/layout/layout.theme.inc
```diff
@@ -244,6 +297,66 @@ function layout_preprocess_layout(&$variables) {
...
+      $variables['flexible_editor'] = TRUE;
+      $data = layout_flexible_tempstore_load($variables['layout_info']['name']);
+    }
+    else {
+      $variables['flexible_editor'] = FALSE;
+      $data = layout_flexible_template_load($variables['layout_info']['name']);
+    }
+    $variables['region_buttons'] = array();
+    if (isset($variables['renderer']->region_buttons)) {
+      $variables['region_buttons'] =  $variables['renderer']->region_buttons;
```
klonos:
```suggestion
      $variables['region_buttons'] = $variables['renderer']->region_buttons;
```

core/modules/layout/layout.theme.inc
```diff
@@ -244,6 +297,66 @@ function layout_preprocess_layout(&$variables) {
...
+    }
+
+
+    $variables['column_data'] = layout_flexible_row_styles();
+
+    $variables['rows'] =  $data['rows'];
```
klonos:
```suggestion
    $variables['rows'] = $data['rows'];
```

core/modules/layout/plugins/renderers/layout_renderer_flexible.inc
```diff
@@ -0,0 +1,172 @@
...
+    $output .= $this->renderEditor();
+    $output .= '</div>';
+
+    return $output;
+  }
+
+  /**
+   * Perform display-level render operations.
+   *
+   * This method renders the flexible editor rows, passes this  to the flexible
```
klonos:
```suggestion
   * This method renders the flexible editor rows, passes this to the flexible
```

core/modules/layout/templates/layout--flexible.tpl.php
```diff
@@ -0,0 +1,57 @@
...
+ *
+ * Variables:
+ * - $messages: Status and error messages. Should be displayed prominently.
+ * - $tabs: Tabs linking to any sub-pages beneath the current page
+ *   (e.g., the view and edit tabs when displaying a node).
+ * - $action_links: Array of actions local to the page, such as 'Add menu' on
+ *   the menu administration interface.
+ * - $classes: Array of classes to be added to the layout wrapper.
+ * - $attributes: Additional attributes to be added to the layout wrapper.
+ * - $row_data: An array of information about each row. Each item  in the array
```
klonos:
```suggestion
 * - $row_data: An array of information about each row. Each item in the array
```

core/themes/seven/css/style.css
```diff
@@ -1233,7 +1233,7 @@ div.filter-options select {
 /**
  * Pretty block radios, `:not(#foo) >` rules out browsers that can't support this
  */
-:not(#foo) > .layout-settings-form .layout-options .form-type-radio input {
+:not(#foo) > .layout-settings-form .layout-options .form-type-radio input, .form-item-region-style  .form-type-radio input {
```
klonos:
```suggestion
:not(#foo) > .layout-settings-form .layout-options .form-type-radio input, .form-item-region-style .form-type-radio input {
```

core/themes/seven/css/style.css
```diff
@@ -1244,6 +1244,7 @@ div.filter-options select {
   clip: rect(0, 0, 0, 0);
   border: 0;
 }
+.form-item-region-style  .form-type-radio label,
```
klonos:
```suggestion
.form-item-region-style .form-type-radio label,
```

core/themes/seven/css/style.css
```diff
@@ -1260,17 +1261,25 @@ div.filter-options select {
   transition: border-color 0.25s ease-in-out, background-color 0.25s ease-in-out;
   will-change: border-color, background-color;
 }
+.form-item-region-style  .form-type-radio label {
```
klonos:
```suggestion
.form-item-region-style .form-type-radio label {
```

core/themes/seven/css/style.css
```diff
@@ -1260,17 +1261,25 @@ div.filter-options select {
...
+  min-height: inherit;
+}
 .layout-settings-page .layout-options.no-flexbox .form-type-checkbox label,
 .layout-settings-form .layout-options.no-flexbox .form-type-radio label {
   min-height: 150px;
 }
+.form-item-region-style  .form-type-radio:active label,
```
klonos:
```suggestion
.form-item-region-style .form-type-radio:active label,
```

core/themes/seven/css/style.css
```diff
@@ -1260,17 +1261,25 @@ div.filter-options select {
...
+.form-item-region-style  .form-type-radio input:checked + label,
```
klonos:
```suggestion
.form-item-region-style .form-type-radio input:checked + label,
```

core/themes/seven/css/style.css
```diff
@@ -1260,17 +1261,25 @@ div.filter-options select {
...
 .layout-settings-page .layout-options .form-type-checkbox:active label,
 .layout-settings-page .layout-options .form-type-checkbox input:checked + label,
 .layout-settings-form .layout-options .form-type-radio:active label,
 .layout-settings-form .layout-options .form-type-radio input:checked + label {
   background-color: #ffffff;
   border-color: #d0d0d0;
 }
+.form-item-region-style  .form-type-radio input:focus + label,
```
klonos:
```suggestion
.form-item-region-style .form-type-radio input:focus + label,
```

core/themes/seven/css/style.css
```diff
@@ -1260,17 +1261,25 @@ div.filter-options select {
...
+.form-item-region-style  .form-type-radio input:hover + label,
```
klonos:
```suggestion
.form-item-region-style .form-type-radio input:hover + label,
```

core/themes/seven/css/style.css
```diff
@@ -1260,17 +1261,25 @@ div.filter-options select {
...
+.form-item-region-style  .form-type-radio:hover input:checked + label,
```
klonos:
```suggestion
.form-item-region-style .form-type-radio:hover input:checked + label,
```

core/modules/layout/layout.flexible.inc
```diff
@@ -0,0 +1,1103 @@
...
+ * Form to configure a flexible template.
+ *
+ * @param $flexible_template
+ *   The loaded flexible template object.
+ *
+ * @ingroup forms
+ */
+function layout_flexible_template_settings_edit_form($form, &$form_state, $flexible_template) {
+  $form_state['flexible_template'] = $flexible_template;
+  $exists = layout_flexible_template_load($flexible_template['name']);
```
klonos: Can we please rename this to something less vague, like `$template_exists`?

herbdool:
Should the checkbox have a header too here `admin/structure/layouts/settings`? Something like "Enabled"? Not intuitive what it does.

(Not sure if I'm rehashing stuff other people already brought up.)

herbdool:
I'm configuring a layout (e.g. admin/structure/layouts/settings/flexible-template/watery_world/configure). I'm not sure what the cancel button does except go back to the layout list. When I add a region it seems to automatically save so I don't have a chance to cancel anything anyway.

Can we either not save automatically or change the cancel button to something else?

herbdool:
When I click "Change region widths" it just shows me the default rather than my current configuration. I think it would be better to have it show the current configuration. For example, if I had 4 regions it should show all the different region widths in the 4 regions option, not the 1 region.

Screenshot from 2019-08-23 10-13-06

docwilmot:
> Should the checkbox have a header too here admin/structure/layouts/settings? Something like "Enabled"? Not intuitive what it does.

this is exactly what the modules table looks like right now though. Can this be a f/u? I'd have to rewrite the form completely to do this.
> I'm configuring a layout (e.g. admin/structure/layouts/settings/flexible-template/watery_world/configure). I'm not sure what the cancel button does except go back to the layout list. When I add a region it seems to automatically save so I don't have a chance to cancel anything anyway.

Just like Views or Layouts your changes are save in tempstore until you save or cancel. Save -> permanent, cancel-> delete all changes.

Maybe we need a tempstore message like in Views and Layouts "This form has unsaved changes. Click "Save layout template" to make changes permanent or "Cancel" to discard changes.". Would that make things clearer?
> When I click "Change region widths" it just shows me the default rather than my current configuration. I think it would be better to have it show the current configuration

Good point, will work on that.

olafgrabienski:
> Maybe we need a tempstore message like in Views and Layouts "This form has unsaved changes. Click "Save layout template" to make changes permanent or "Cancel" to discard changes.". Would that make things clearer?

Yes!

core/modules/layout/layout.flexible.inc
```diff
@@ -0,0 +1,1167 @@
...
+
+/**
+ * Menu callback; Delete a flexible template.
+ *
+ * @ingroup forms
+ */
+function layout_flexible_template_delete_form($form, &$form_state, $flexible_template) {
+  $form_state['template'] = $flexible_template;
+  $question = t('Delete flexible template @title?', array('@title' => $flexible_template['title']));
+  $description = t('This action cannot be undone.');;
```
klonos:
```suggestion
  $description = t('This action cannot be undone.');
```

core/modules/layout/layout.flexible.inc
```diff
@@ -0,0 +1,1167 @@
...
+  $errors = array();
+  foreach (layout_load_all() as $layout) {
+    if ($layout->layout_template == $flexible_template['name']) {
+      $errors[] = $layout->title;
+    }
+  }
+
+  if (!empty($errors)) {
+    $form['template_in_use'] = array(
+      '#title' => t('This template is in use in the following layouts') . ':',
```
klonos: We include punctuation inside the `t()`s

```suggestion
      '#title' => t('This template is in use in the following layouts:'),
```

core/modules/layout/layout.flexible.inc
```diff
@@ -0,0 +1,1167 @@
...
+      '#type' => 'item',
```
klonos: Can we please move this to be the first item, above the `#title`?

core/modules/layout/layout.flexible.inc
```diff
@@ -0,0 +1,1167 @@
...
+      '#markup' => theme('item_list', array('items' => $errors)),
+    );
+    backdrop_set_message(t('The "@layout" layout template is currently in use and may not be deleted.', array('@layout' => $flexible_template['title'])), 'error');
```
klonos: Any reason why we are using both a form element + an error message here? I would ditch the form element, and include the list of layouts that are using the template within the error message. This is what we are doing elsewhere.

Also, instead of wrapping the layout name placeholder in double quotes, you should use `%layout` instead of the regular `@layout`; this would render it wrapped in `<em>`.

core/modules/layout/layout.flexible.inc
```diff
@@ -0,0 +1,1167 @@
...
+ * Menu callback to delete a row from a flexible template.
+ *
+ * @param $flexible_template
+ *   The loaded flexible template object.
+ * @param $original_row
+ *   The row above or below which a new row is being inserted.
+ *
+ * @ingroup forms
+ */
+function layout_flexible_template_delete_row($flexible_template, $original_row) {
```
klonos: Any specific reason why we are not using `confirm_form()` for this?

docwilmot: This action is saved in tempstore and not made permanent until the final "Save layout template". So no need for a confirm form as its already a 2-step process.

core/modules/layout/layout.flexible.inc
docwilmot: Good call

quicksketch:
New Sandbox.

quicksketch:
New sandbox.

core/modules/layout/layout.admin.inc
```diff
@@ -1909,6 +1910,54 @@ function layout_enable_layout($status, Layout $layout) {
   backdrop_goto('admin/structure/layouts');
 }
 
+/**
+ * Menu callback; Enable or disable a layout template.
+ */
+function layout_enable_layout_template($status, $template_name) {
```
klonos: I liked better how @jenlampton has named the respective function that enables/disables file types, and how the parameter for the thing being toggled comes before the one for the status it's being toggled to. Can we please change this to:

```suggestion
function layout_layout_template_toggle_enable($template_name, $status) {
```

docwilmot: We dont have a convention for this, I'm OK with `layout_enable_layout_template()`.

quicksketch: > Can we please change this to:

Yeah I wasn't crazy about the parameter order either. I made it that order because it matches the function directly above it, which is:

```
function layout_enable_layout($status, Layout $layout) {
```

So that's why I did the same thing for a layout template:
```
function layout_enable_layout_template($status, $template_nname) {
```

I think we could flip both though. It's a menu callback that isn't likely used anywhere else.

core/modules/layout/layout.flexible.inc
```diff
@@ -0,0 +1,826 @@
...
+  // Todo use token, like in block add page.
```
herbdool: Should we check a token before deleting, like this comment suggests? I haven't checked block add page but I assume that's what it's referring to. @docwilmot

core/modules/layout/layout.module
```diff
@@ -1796,3 +1916,368 @@ function layout_get_title_description(Layout $layout) {
...
   return $description;
 }
+
+/**
+ * Store changes to a flexible layout template in the temporary store.
+ *
+ * @param $flexible_template
+ *   The loaded flexible template object.
+ */
+function layout_set_flexible_tempstore($flexible_template) {
```
herbdool: Maybe this could be `layout_flexible_tempstore_set()` to match `layout_flexible_tempstore_load()`.

herbdool (changes requested):

core/modules/layout/layout.flexible.inc
quicksketch: I think this was removed in an earlier iteration, not sure why it still showing up here. It's no longer in the latest version of code.

core/modules/layout/layout.flexible.inc
quicksketch: Just added this in. Thanks!

core/modules/layout/layout.module
quicksketch: Thanks that's better. Changed.

PR #2647, opened 2019-05-08
Issue #3760: [DX] Add a utility-like function that removes contrib modules that were merged into core.

klonos:
https://github.com/backdrop/backdrop-issues/issues/3760

### DIFF
--- a/core/includes/update.inc
+++ b/core/includes/update.inc
@@ -444,11 +444,11 @@ function update_uninstall_merged_module($module) {
       // this entry will be recreated, but it will have status 0 (disabled), and
       // schema -1 (uninstalled). So it is safe to remove from the file system.
       db_query("DELETE FROM {system} WHERE name = :module AND type = 'module'", array(':module' => $module));
-      backdrop_set_message(t('The contributed module %module has been located at <code>@path</code>. The module has been disabled, since its functionality is now provided by Backdrop core. It is recommended that you remove this module from your site.', array('%module_name' => $module_name, '@path' => BACKDROP_ROOT . '/' . $module_path)), 'warning');
+      backdrop_set_message(t('The contributed module %module has been located at <code>@path</code>. The module has been disabled, since its functionality is now provided by Backdrop core. It is recommended that you remove this module from your site.', array('%module' => $module_name, '@path' => BACKDROP_ROOT . '/' . $module_path)), 'warning');
       // Throw a warning if the merged contrib module has not been added to the
       // list of modules returned by backdrop_merged_modules().
       if (!in_array($module, backdrop_merged_modules())) {
-        backdrop_set_message(t('The %module module is not included in <code>backdrop_merged_modules()</code>.', array('%module_name' => $module_name)), 'warning');
+        backdrop_set_message(t('The %module module is not included in <code>backdrop_merged_modules()</code>.', array('%module' => $module_name)), 'warning');
       }
       return TRUE;
     }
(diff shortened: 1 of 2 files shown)


core/includes/update.inc
```diff
@@ -418,6 +418,46 @@ function update_variable_del($name) {
...
+ * Use this to remove contrib modules that have been merged into Backdrop core.
+ * Be sure to add an entry for that module in backdrop_merged_modules().
+ *
+ * @param string $module
+ *   The name of the contrib module to uninstall.
+ *
+ * @return
+ *   TRUE if the module has been found and uninstalled; otherwise FALSE.
+ */
+function update_uninstall_merged_module($module) {
```
serundeputy, line 433: @klonos you could make `$module` an `$modules` array here and foreach over it.

klonos: Thanks @serundeputy 👍 ...I imagine also rename the function to `update_uninstall_merged_modules()` (use plural for "modules")?

klonos: ...my only concern is that as it is currently, I return TRUE if a module has been found and uninstalled, and FALSE if not. What should happen in the case of multiple modules, where some were found, while others not?

klonos: ...should we perhaps leave it as is (single module instead of an array), and then use foreach when we call this function? As in:

```php
  update_uninstall_merged_module($module);
}
```

serundeputy: that works

otherwise you could return an array something like:

```php
return array(
  'module_one' => FALSE,
  'module_two' => TRUE,
  .
  .
  .
);
```

quicksketch: It's pretty unusual we merge more than one module into core at a time (hasn't happened yet), so IMO just handling the single-module use-case seems fine.

core/modules/field/field.install
```diff
@@ -122,19 +122,7 @@ function field_update_1002() {
...
-    // to be anything else left to clean up or migrate.
-    // While the contrib module remains in the filesystem, this entry will be
-    // recreated, but it will have status 0 (disabled), and schema -1
-    // (uninstalled), so it is safe to remove.
-    db_query("DELETE FROM {system} WHERE name = 'options_element' AND type = 'module'");
-  }
-  else {
+  if (!update_uninstall_merged_module('options_element')) {
     return t('<em>Options Element</em> not found. Nothing to be done.');
   }
```
quicksketch, line 127: I find this to be weird that we return a message when nothing happens, but display nothing when we do something. I know there is the big message at the top of the update page, but seems like we should return a message in the update results area as well?

klonos: Yeah, I thought of that, but in my head this works like this:

- if a module has been found and was disabled, then we need to warn the user and let them know that it is advised to remove the contrib module from the filesystem.
- if nothing was found, we just communicate this in a less prominent way, via the update results area.

The update results area has traditionally been used to add log-type errors (like SQL errors for example). We could deprecate it, and use `backdrop_set_message()` for everything(?).

In any way, we could return something like `'<em>Options Element</em> found and disabled.'`.

I think that this is another bit that could be included in `update_uninstall_merged_module()` ...I mean, we introduced this function in order to not be duplicating code, and now we are returning the "not found. Nothing to be done." message in each update hook. Unsure how to implement it properly though. What if I added another `&$update_results_message` parameter to it, which was passed by reference? ...we could then return that too. Would that work though? I need to experiment more 😅

klonos:
...in the meantime, I have tweaked the inline documentation in this function, to explain why we are doing direct db queries. This was actually a collation of the answers to questions posted by other devs that reviewed the PRs that originally included the implementation of this function in the update hooks. Hope that there will be less head-scratching in the future when people revisit this code 😉

PR #2654, opened 2019-05-13
Views strings not translatable when loaded from a feature

### LINKED ISSUE #4137: Views strings not translatable when loaded from an export
This issue is a spin-off from a views cross-port that likely won't apply to Backdrop, but we may want to discuss. 

from @quicksketch in https://github.com/backdrop/backdrop-issues/issues/3720#issuecomment-541466138

> I don't think we'll need/want a direct port of that code from Drupal 7.

From @quicksketch in https://github.com/backdrop/backdrop/pull/2654#issuecomment-541464146

> I think we may need to think about this one wholistically. The concept of "exported" views doesn't exist in Backdrop; we don't have hook_views_default_views() like D7 did. Instead all views are always stored in config. So the idea of using a different translation system based on whether a view is exported or now doesn't make sense in Backdrop. That may also mean that the view::is_translatable() method needs updating in Backdorp.

---
PR by @klonos: https://github.com/backdrop/backdrop/pull/2654

klonos:
https://www.drupal.org/node/1248716

https://git.drupalcode.org/project/views/commit/fba32b8efcff01e5fd03a0905aa1647d6dfea52b

Fixes https://github.com/backdrop/backdrop-issues/issues/4137 (?)

### DIFF
--- a/core/modules/views/includes/view.inc
+++ b/core/modules/views/includes/view.inc
@@ -1992,6 +1992,10 @@ class view {
    * Determine whether a view supports admin string translation.
    */
   function is_translatable() {
+    // Use translation no matter what type of view.
+    if (config_get('views.settings', 'localize_all')) {
+      return TRUE;
+    }
     // If the view is normal or overridden, use admin string translation.
     // A newly created view won't have a type. Accept this.
     return (!isset($this->type) || in_array($this->type, array(t('Normal'), t('Overridden')))) ? TRUE : FALSE;
--- a/core/modules/views/views.install
+++ b/core/modules/views/views.install
@@ -173,6 +173,17 @@ function views_update_1005() {
 }
 
 /**
+ * Convert the views_localize_all variable to config.
+ */
+function views_update_1006() {
+  $config = config('views.settings');
+  $config->set('localize_all', update_variable_get('views_localize_all', FALSE));
+  $config->save();
+
+  update_variable_del('views_localize_all');
+}
+
+/**
  * @} End of "addtogroup updates-7.x-to-1.x"
  * The next series of updates should start at 2000.
  */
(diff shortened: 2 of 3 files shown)


klonos:
We seem to be missing this entire set of radios in `views_ui.admin.inc`:

```php
  $form['locale']['views_localization_plugin'] =  array(
    '#type' => 'radios',
    '#title' => t('Translation method'),
    '#options' => views_fetch_plugin_names('localization', NULL, array()),
    '#default_value' => views_get_localization_plugin(),
    '#description' => t('Select a translation method to use for Views data like header, footer, and empty text.'),
  );
```

Was this specific plugin not implemented in the Backdrop version of Views? Shouldn't we include `views_get_localization_plugin()` (even if only to have it be a wrapper for `watchdog_deprecated_function()`)?

klonos:
...refreshing PR sandbox.

jenlampton:
The PR looks good. RTBC. 

@klonos you make a good point about [views_get_localization_plugin](https://api.drupal.org/api/views/views.module/function/views_get_localization_plugin/7.x-3.x) Can you make a separate issue for that?

quicksketch:
I think we may need to think about this one wholistically. The concept of "exported" views doesn't exist in Backdrop; we don't have hook_views_default_views() like D7 did. Instead all views are always stored in config. So the idea of using a different translation system based on whether a view is exported or now doesn't make sense in Backdrop. That may also mean that the `view::is_translatable()` method needs updating in Backdorp.

klonos:
> @klonos you make a good point about views_get_localization_plugin Can you make a separate issue for that?

@jenlampton it seems that time has made that happen 😅 ... @indigoxela created https://github.com/backdrop/backdrop-issues/issues/4382 and we re-introduced all that missing section in the form with https://github.com/backdrop/backdrop/pull/3132 which was merged and included since 1.16.3 👍

PR #2660, opened 2019-05-13
Check for entity in views_handler_field_field->get_value() before accessing it.

klonos:
https://www.drupal.org/node/2115269

https://git.drupalcode.org/project/views/commit/21fd8f9f89deaf875e768b6fb97717297f0f6967

### DIFF
--- a/core/modules/field/views/views_handler_field_field.inc
+++ b/core/modules/field/views/views_handler_field_field.inc
@@ -767,6 +767,10 @@ class views_handler_field_field extends views_handler_field {
   }
 
   function get_value($values, $field = NULL) {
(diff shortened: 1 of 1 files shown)


klonos:
...not entirely sure if this needs additional work compared to the Views 7.x commit (because of the recent changes in https://github.com/backdrop/backdrop/pull/2596 and/or https://github.com/backdrop/backdrop/pull/2560)

jenlampton:
@klonos I don't think so. Those were utility functions anyway, everything else still works the same way. 

RTBC.

quicksketch:

PR #2679, opened 2019-05-15
Views: Schema mismatch after upgrade D6 to D7/Backdrop

klonos:
https://www.drupal.org/node/1154830

https://git.drupalcode.org/project/views/commit/4edefe170aec2df46a59a3b5a17e67429f9108fb

### DIFF
--- a/core/modules/views/views.install
+++ b/core/modules/views/views.install
@@ -173,6 +173,20 @@ function views_update_1005() {
 }
 
 /**
+ * Remove headers field from cache tables.
+ *
+ * @see https://www.drupal.org/node/1154830.
+ */
+function views_update_1006() {
+  if (db_field_exists('cache_views', 'headers')) {
+    db_drop_field('cache_views', 'headers');
+  }
+  if (db_field_exists('cache_views_data', 'headers')) {
+    db_drop_field('cache_views_data', 'headers');
+  }
+}
+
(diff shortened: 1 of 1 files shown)


jenlampton:
Looks good, RTBC.

jenlampton:
Question from @klonos on the main issue:

> Re https://www.drupal.org/node/1154830 and [the respective 7.x commit](https://git.drupalcode.org/project/views/commit/4edefe170aec2df46a59a3b5a17e67429f9108fb), do we support D6 straight to Backdrop, and in general, do we advise people to first upgrade to the latest version of the respective 7.x module before doing D7 to Backdrop upgrades?

> I'm asking because the commit adds a 
`views_update_7302()`, which would have already run if the module on a D6/D7 site was using the latest version of Views 7.x. Other wise, do we need to add a new 100x update hook on our end?

Yes we advise people upgrade from D6 to D7 and then D7 to backdrop from the latest version of views. But I don't see any harm in adding it anyway, better safe than sorry?

But I'll leave it to the core committers to decide...

quicksketch:
Thanks! I had a concern about this one but seems like a good idea to me. See https://github.com/backdrop/backdrop-issues/issues/3720#issuecomment-541465329.

Merged into 1.x and 1.14.x.

PR #2677, opened 2019-05-14
Issue #3781: [PS] Vocabulary overview page: use a single query with a `GROUP BY`

klonos:
https://github.com/backdrop/backdrop-issues/issues/3781

PR #2682, opened 2019-05-15
Issue #3781: [PS] Vocabulary overview page: use a single query with a `GROUP BY`

### LINKED ISSUE #3781: [PS] Vocabulary overview page: use a single query with a `GROUP BY`, instead of one query per vocabulary
This is a follow-up to #3655.

@quicksketch:

> This could be optimized using a single query with a `GROUP BY` instead of one query per vocabulary, but in the scheme of overall site speed, this seems like it would be negligible. I've merged https://github.com/backdrop/backdrop/pull/2594 into 1.x.

---
PR by @klonos: https://github.com/backdrop/backdrop/pull/2682

klonos:
Fixes: https://github.com/backdrop/backdrop-issues/issues/3781

Alternative to: https://github.com/backdrop/backdrop/pull/2677:


quicksketch:
Looks great! Tested out the sandbox with flying colors to double-check.

PR #2693, opened 2019-05-20
Set sticky, promote and revision default settings regardless of 'shown' status.

### LINKED ISSUE #3800: Node default values ignored if 'Show option...' disabled
**Description of the bug**
If you set a content type's "Show option to promote content" setting to disabled (meaning that option isn't shown on the node forms) but you set the "Promote content by default" setting to enabled, when you create a node it's not promoted.

The reasons I believe it should be promoted in this case are:

1. The first setting is called '*Show* option to promote content', nothing about enabling/disabling the next setting
2. The 'Promote content by default' setting is always shown, regardless of the first setting (as a result of https://github.com/backdrop/backdrop-issues/issues/1536)

The same is true for the Sticky and Revision settings. This is basically a follow-up issue to https://github.com/backdrop/backdrop-issues/issues/1536.

**Steps To Reproduce**
To reproduce the behavior:
1. Edit a content type and untick 'Show option to promote content' but tick 'Promote content by default'
2. Create a piece of content of that type
3. Check the frontpage (if set to 'node') or the database to see the value of 'promote'

**Actual behavior**
Content is not promoted.

**Expected behavior**
Content is promoted.

-----

PR: https://github.com/backdrop/backdrop/pull/2693

ghost:
Fixes https://github.com/backdrop/backdrop-issues/issues/3800


stpaultim:
@BWPanda - Sandbox needs refreshing.

ghost:
@stpaultim Please try again.

PR #2695, opened 2019-05-21
Added a bottom border to Admin tables in seven

matthui:
Fixes #3741, also deleted a repeated CSS line (532) because it is a repeat of line 527


jenlampton:
Replaced by https://github.com/backdrop/backdrop/pull/2714

PR #2696, opened 2019-05-21
Issue #3804: [UX] Search result list is hidden

### LINKED ISSUE #3804: [UX] Admin bar: Search result list is hidden in RTL view
**Description of the bug**
Bug in 'Administration bar' core module: If an RTL language is default, the search result list is hidden under search bar.

**Steps To Reproduce**
To reproduce the behavior:
1. Enable the 'Language' core module.
2. Set an RTL language as default. For example: Hebrew
3. Write a key word in the search bar. For example: url

**Actual behavior**
There isn't result list.

**Expected behavior**
Result list is displayed with links.

**Additional information**
I will create a PR: We need to remove a line (307) from admin_bar.css, and the result list appears.

---
PR by @VasasA: https://github.com/backdrop/backdrop/pull/2696

VasasA:
Fix: https://github.com/backdrop/backdrop-issues/issues/3804
Bug in 'Administration bar' core module: If an RTL language is default, the search result list is hidden under search bar. We need to remove a line (307) from admin_bar.css, and the result list appears.

### DIFF
--- a/core/modules/admin_bar/css/admin_bar.css
+++ b/core/modules/admin_bar/css/admin_bar.css
@@ -304,7 +304,6 @@
 }
 [dir="rtl"] #admin-bar .dropdown li ul {
   left: auto;
-  right: -999em;
 }
 
 /* Third-and-above-level lists */


quicksketch:
Thanks @VasasA! Merged into 1.x and 1.13.x.

PR #2706, opened 2019-05-28
Display a Backdrop logo in the Powered By footer.

### LINKED ISSUE #576: Add our logo to the "powered by backdrop" footer.
...and link it up for good SEO juice.

ghost:
Fixes https://github.com/backdrop/backdrop-issues/issues/576


core/modules/system/system.theme.inc
```diff
@@ -10,7 +10,7 @@
  * @ingroup themeable
  */
 function theme_system_powered_by() {
-  return '<span>' . t('Powered by <a href="@poweredby">Backdrop CMS</a>', array('@poweredby' => 'https://backdropcms.org')) . '</span>';
+  return '<span>' . t('<a href="@backdrop" class="@class" title="Powered by Backdrop CMS">Powered by Backdrop CMS</a>', array('@backdrop' => 'https://backdropcms.org', '@class' => 'logo backdrop-mark')) . '</span>';
```
```suggestion
  $powered_by = t('Powered by Backdrop CMS');
  $powered_by_link = l($powered_by, 'https://backdropcms.org', array('attributes' => array('class' => array('logo', 'backdrop-mark'), 'title' => $powered_by)));

  return '<span>' . $powered_by_link . '</span>';
```

core/modules/system/system.module
```diff
@@ -3222,6 +3238,35 @@ function system_breadcrumb_block($settings = array()) {
...
+function system_powered_by_block($settings = array()) {
+  if (isset($settings['color']) && $settings['color'] == 'black') {
+    $primary = '#000';
+    $secondary = '#fff';
+  }
+  else {
+    $primary = '#fff';
+    $secondary = '#000';
+  }
+  backdrop_add_css('
```
klonos: Can we add these to a .css file and load it in `backdrop_add_css()` with `'file'` instead of `'inline'`?

...either that, or use heredoc perhaps?

ghost:
@klonos Well that makes much more sense and is much cleaner! Thanks 😁

core/modules/system/system.module
ghost: @klonos Confirming that it's better to have two additional CSS files, and then choose between them based on the color setting, and then load that one, instead of adding the CSS inline? I know inline CSS doesn't get aggregated, but I thought having two additional, and very small, CSS files would be a bit pointless. Good to hear your advice though.

klonos (approved):
👍

PR #2710, opened 2019-05-31
Issue #3832: Prevent PHP warning about backdrop_clean_css_identifier …

### LINKED ISSUE #3832: PHP Warning: strtr(): The second argument is not an array in backdrop_clean_css_identifier
**Description of the bug**
On a views page, I see the following error in the logs:
```
Warning: strtr(): The second argument is not an array in backdrop_clean_css_identifier() (line 4166 of backdrop/core/includes/common.inc).
```

**Steps to reproduce**
1) Enable all errors in `admin/config/development/logging`
1) Create a view using a list style display
1) add a css class for the wrapper
1) add a css class for the list
1) save the view
1) view the view -> error 👎 ...if you have enabled errors in step 1 above; if not, then...
1) check the logs -> error 👎 

**Additional information**
This happens because we are using `array_walk()` in `views.theme.inc` to pass classes in views through `backdrop_clean_css_identifier()` but we are not passing an equal number of argument-sets to the number of classes, and `array_walk()` requires that they match. When they don't, NULL is passed as the second argument, and `backdrop_clean_css_identifier()` expects an array - which causes this error.

---
PR: https://github.com/backdrop/backdrop/pull/2710

jenlampton:
Fixes https://github.com/backdrop/backdrop-issues/issues/3832

### DIFF
--- a/core/modules/views/templates/views.theme.inc
+++ b/core/modules/views/templates/views.theme.inc
@@ -56,7 +56,7 @@ function template_preprocess_views_view(&$variables) {
 
   $variables['classes'] = array();
   $variables['classes'][] = 'view';
-  $variables['classes'][] = 'view-' . backdrop_clean_css_identifier($variables['name']);
+  $variables['classes'][] = 'view-' . $variables['css_name'];
   $variables['classes'][] = 'view-id-' . $variables['name'];
   $variables['classes'][] = 'view-display-id-' . $variables['display_id'];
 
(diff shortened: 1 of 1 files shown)


core/modules/views/templates/views.theme.inc
```diff
@@ -930,10 +930,14 @@ function template_preprocess_views_view_list(&$variables) {
   $handler  = $variables['view']->style_plugin;
 
   $class = explode(' ', $handler->options['class']);
-  $class = array_map('backdrop_clean_css_identifier', $class, array(array()));
+  foreach ($class as $key => $individual) {
```
klonos: Wondering if we can rename the `$class` variable above this line to `$classes` instead? That way, we could do this, which makes more sense to me:
```suggestion
  foreach ($classes as $key => $class) {
```

core/modules/views/templates/views.theme.inc
```diff
@@ -930,10 +930,14 @@ function template_preprocess_views_view_list(&$variables) {
...
+    $class[$key] = backdrop_clean_css_identifier($individual, array());
+  }
 
   $wrapper_class = explode(' ', $handler->options['wrapper_class']);
-  $wrapper_class = array_map('backdrop_clean_css_identifier', $wrapper_class, array(array()));
+  foreach ($wrapper_class as $key => $individual) {
```
klonos: Same as my previous comment; can we rename `$wrapper_class` to `$wrapper_classes` instead, and do:
```suggestion
  foreach ($wrapper_classes as $key => $wrapper_class) {
```

klonos:
@jenlampton close/reopen the PR to refresh the sandbox please?

jenlampton:
Thanks @quicksketch

klonos (approved):
WFM 👍

quicksketch:

PR #2712, opened 2019-05-31
Changed misleading description for 'Trimmed limit' setting.

### LINKED ISSUE #3828: Misleading description for 'Trimmed limit' setting on long text fields
When configuring the 'Trimmed' field format, the 'Trimmed limit' setting has the following description:

> If the summary is not set, the trimmed [FIELD] field will be shorter than this character limit.

This description is the same for both the 'Summary or trimmed' and 'Trimmed' formats. It makes sense for the 'Summary or trimmed' format, but the 'Trimmed' format doesn't take the text's summary into consideration when trimming, so this is misleading. THe description should be changed for the 'Trimmed' format.

image

ghost:
Fixes https://github.com/backdrop/backdrop-issues/issues/3828


quicksketch:
Merged into 1.x and 1.25.x. Thanks @BWPanda!

PR #2713, opened 2019-05-31
Issue #3788: Document when #access should be used in hide().

### LINKED ISSUE #3788: Document where `#access = FALSE` should be used instead of `hide()`
This came up in #3768 (credits to @BWPanda for figuring that out). Perhaps we should document this in [`hide()`](https://api.backdropcms.org/api/backdrop/core%21includes%21common.inc/function/hide/1) so that people don't have to waste time re-researching things:

> https://stackoverflow.com/a/21041471/4168619 suggests that `hide()`-ing a field should only be done if you intend to display it elsewhere:
>
>>I.e. hiding it (somewhere I saw someone suggesting that): `hide($form['field_myfield']);` really hides the field, but after that, when form is saved this field has empty value, validation fails, etc, so that's not a good way to do this. Hiding makes sense only if you want to print separately that field later, at some other place.
>
>Also, https://api.drupal.org/comment/59900#comment-59900 says:
>
>>Using `hide()` on field of type Date will give you a validation error, so `#access = FALSE` has to be used.
>
>It seems that I was trying to use `hide()` incorrectly, and that I should be using `'#access' => FALSE` instead. Closing this issue then.

---

PR by @quicksketch: https://github.com/backdrop/backdrop/pull/2713

quicksketch:
Fixes https://github.com/backdrop/backdrop-issues/issues/3788 if it looks good to everyone else.

### DIFF
--- a/core/includes/common.inc
+++ b/core/includes/common.inc
@@ -6725,6 +6725,15 @@ function render(&$element) {
  * rendered for the first time, as it will have no effect on subsequent
  * renderings of the parent tree.
  *
+ * Note that hide() should only be used when doing the final printing of a
+ * render element, such as in a template. If wanting to hide a form element
+ * earlier in the process, such as in hook_form_alter(), you should set #access
+ * instead, like this:
+ *
+ * @code
+ * $element['#access'] = FALSE
+ * @endcode
+ *
  * @param $element
  *   The element to be hidden.
  *


klonos (approved):
👍

quicksketch:
Merged into 1.x and 1.13.x. Thanks @klonos for the review!

PR #2716, opened 2019-06-01
Issue #3841: Replace table borders with box-shadow borders in admin t…

### LINKED ISSUE #3841: Improve table border color in Seven (admin theme)
This is a follow-up to https://github.com/backdrop/backdrop-issues/issues/3741

---

We were just noticing that in the [styleguide](http://backdrop-ops.github.io/styleguide/#table) the tables didn't have a bottom border. There they looked great because the styleguide shows everything on a light gray background:
styleguide

But on the dashboard in Backdrop core, we have tables on a white background. When the bottom row of a table also has a white background, this caused the tables to have this weird floaty feel:
floaty

@jenlampton recommended adding a bottom border to the tables, one that matched the left & right borders. When we tried that, we noticed that color gray was so light that the border was still almost invisible:
invisible

@klonos recommended that we darken the gray color, so it would be easier to see where the table ends:
ends

@quicksketch pointed out that this made the left-right borders now a different color from the bottom border. And that maybe we should bring in @dariusgarza since this gets into the area of changing the design.
@klonos had also noticed that when you can see the borders, it becomes clear that the left/right borders are 1px wider than the header, which gets distracting. 
sideburns

@jenlampton proposes that we replace these actual borders with "box-shadow borders" instead, which will not cause the 1px difference in width:
better

jenlampton:
Fixes https://github.com/backdrop/backdrop-issues/issues/3841

### DIFF
--- a/core/themes/seven/css/style.css
+++ b/core/themes/seven/css/style.css
@@ -511,14 +511,21 @@ table {
   border: 0;
 }
 tbody {
-  border-bottom: 1px solid #f7f7f7;
+  -webkit-box-shadow: inset 0 1px 0 0 #eee;
+  box-shadow: inset 0 1px 0 0 #eee;
 }
 tr.even,
 tr.odd {
-  border: 1px solid #f7f7f7;
-  border-top: 0;
-  border-bottom: 0;
+  border: 0;
   background: #f7f7f7;
+  -webkit-box-shadow:
+    inset -1px 0 0 #eee,
+    inset 0 -1px 0 0 #eee,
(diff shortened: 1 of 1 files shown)


docwilmot:
Can we reboot please. Sandbox gone.

jenlampton:
new sandbox?

klonos (approved):
LGTM 👍

quicksketch:
Thanks! Looks good to me in testing as well. Merged into 1.x and 1.16.x.

PR #2720, opened 2019-06-02
Remove views-more.tpl.php and use theme_more_link() instead.

### LINKED ISSUE #547: Remove views-more.tpl.php (use theme_more_link instead)
Views provides it's own theme template for a more link.  We already have one of these in core.  Let's consolidate.

ghost:
Fixes https://github.com/backdrop/backdrop-issues/issues/547


drupol (approved):
I agree with the idea, however, there is currently no way to add more attributes to the link. The wrapping DIV also dissapeared. Is this the intended behavior?

ghost:
Yes, see https://github.com/backdrop/backdrop-issues/issues/547 for the discussion.

core/includes/theme.inc
```diff
@@ -2476,7 +2476,7 @@ function theme_head_tag($variables) {
  *   - title: A descriptive verb for the link, like 'Read more'.
  */
 function theme_more_link($variables) {
-  return '<div class="more-link">' . l(t('More'), $variables['url'], array('attributes' => array('title' => $variables['title']))) . '</div>';
+  return l(t('More'), $variables['url'], array('attributes' => array('title' => $variables['title'], 'class' => 'more-link')));
```
drupol, line 2479: After a second thought, I definitely think that this is an unrelated changes and should be reverted for now.

core/modules/system/css/system.theme.css
```diff
@@ -164,6 +164,7 @@ abbr.ajax-changed {
  * Markup generated by theme_more_link().
  */
 .more-link {
+  display: block;
```
drupol, line 167: Unrelated changes.

core/modules/views/plugins/views_plugin_display.inc
```diff
@@ -2537,10 +2537,9 @@ class views_plugin_display extends views_plugin {
         if (!empty($this->view->exposed_raw_input)) {
           $url_options['query'] = $this->view->exposed_raw_input;
         }
-        $theme = views_theme_functions('views_more', $this->view, $this->display);
```
drupol, line 2540: I would keep this line and modify it like this:

`$theme = views_theme_functions('more_link', $this->view, $this->display);`

core/modules/views/plugins/views_plugin_display.inc
```diff
@@ -2537,10 +2537,9 @@ class views_plugin_display extends views_plugin {
...
         $path = check_url(url($path, $url_options));
 
-        return theme($theme, array('more_url' => $path, 'link_text' => check_plain($this->use_more_text()), 'view' => $this->view));
+        return theme('more_link', array('url' => $path, 'title' => check_plain($this->use_more_text())));
```
drupol, line 2542:
```suggestion
        return theme($theme, array('url' => $path, 'title' => check_plain($this->use_more_text())));
```

drupol (changes requested):
I left my review, let me know what you think.

core/modules/views/plugins/views_plugin_display.inc
drupol: This `views_theme_functions()` is what makes people love views and why it's so easy to extend it through templates.

Basically, when you call `$theme = views_theme_functions('more_link', 'view_custom', 'display_custom');`, the value of `$theme` will be: `more_link__view_custom__display_custom`.

Then, the templates that will be available are:
* `more-link--view-custom--display-custom.tpl.php`
* `more-link--view-custom.tpl.php`
* `more-link.tpl.php`

In this case, as we are using a theme function to render the link, those templates won't be used, but if one day we decide to convert theme functions and use templates exclusively, then it will be needed.

And this, is exactly what Atomium is doing, see the more_link: https://github.com/backdrop-contrib/atomium/tree/1.x-1.x/atomium/templates/more_link

PR #2721, opened 2019-06-02
Issue #3846: Prevent PHP fatal on update when staging directory is mi…

### LINKED ISSUE #3846: [UX][DX] Update.php throws error about install.php (and says to delete live config)
I'm experiencing some server issue (missing staging directory) that is preventing update.php from running. 

When this happens, I would expect to see an error message letting me know that my staging directory is missing. 

Unfortunately, instead I see an error about the installer that asks me to **delete all my live config** which is the last thing I should do. 

```
Exception: The <em class="placeholder">../config/live-active</em> directory must be completely empty to proceed with installation. Empty this directory or specify a different directory in settings.php. in backdrop_install_config_directories() (line 851 of core/includes/install.inc).
```

It looks like `update_prepare_bootstrap()` calls `backdrop_install_config_directories()` even when the staging directory is missing, and that checks to confirm that the live-active directory is empty, since it's intended to be used for the installer.

Related: 
* https://github.com/backdrop/backdrop-issues/issues/2262
* https://github.com/backdrop/backdrop-issues/issues/2538

---

PR: https://github.com/backdrop/backdrop/pull/2721

jenlampton:
Fixes https://github.com/backdrop/backdrop-issues/issues/3846

### DIFF
--- a/core/includes/update.inc
+++ b/core/includes/update.inc
@@ -89,8 +89,15 @@ function update_prepare_bootstrap() {
   $active_config = config_get_config_storage('active');
   $staging_config = config_get_config_storage('staging');
 
-  if (!$settings_exist || !$active_config->isInitialized() || !$staging_config->isInitialized()) {
+  if (!$settings_exist || !$active_config->isInitialized()) {
     backdrop_install_config_directories();
+  }
+
+  // Check for a staging directory.
+  if (!$staging_config->isInitialized()) {
+    $t = get_t();
+    backdrop_set_message($t('The staging configuration directory (%directory) could not be found. Please make sure it exists and is properly referenced in settings.php.',
+      array('%directory' => config_get_config_directory('staging'))), 'warning', FALSE);
   }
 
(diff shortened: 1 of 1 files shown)


core/includes/update.inc
```diff
@@ -89,10 +89,17 @@ function update_prepare_bootstrap() {
   $active_config = config_get_config_storage('active');
   $staging_config = config_get_config_storage('staging');
 
-  if (!$settings_exist || !$active_config->isInitialized() || !$staging_config->isInitialized()) {
+  if (!$settings_exist || !$active_config->isInitialized()) {
     backdrop_install_config_directories();
   }
 
+  // Check for a staging directory
```
klonos: There's a period missing:

```suggestion
  // Check for a staging directory.
```

klonos: ...alternative for consideration:

```suggestion
  // Warn if staging directory is not found.
```

Not sure if we want to add a "but do not prevent updates" there.

quicksketch:

PR #2723, opened 2019-06-03
Moved 'flush caches' permission to system.module.

### LINKED ISSUE #3844: Move "flush caches" permission from admin_bar to system module
As per the discussion in #3617 I think we should have a new permission for viewing any messages related to caches being cleared/flushed. These aren't something most users would care, or even know, about, so I think being able to show them more specifically (e.g. just Administrators by default) would be good.

---
PR by @BWPanda: https://github.com/backdrop/backdrop/issues/2723

ghost:
Fixes https://github.com/backdrop/backdrop-issues/issues/3844

### DIFF
--- a/core/modules/system/system.module
+++ b/core/modules/system/system.module
@@ -179,6 +179,10 @@ function system_permission() {
     'block IP addresses' => array(
       'title' => t('Block IP addresses'),
     ),
+    'flush caches' => array(
+      'title' => t('Flush caches'),
+      'description' => t('Access links to flush caches.'),
+    ),
   );
 }
 
--- a/core/modules/system/system.module
+++ b/core/modules/system/system.module
@@ -181,7 +181,7 @@ function system_permission() {
     ),
     'flush caches' => array(
       'title' => t('Flush caches'),
-      'description' => t('Access links to flush caches.'),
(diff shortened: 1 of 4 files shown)


core/modules/system/system.module
```diff
@@ -179,6 +179,10 @@ function system_permission() {
     'block IP addresses' => array(
       'title' => t('Block IP addresses'),
     ),
+    'flush caches' => array(
+      'title' => t('Flush caches'),
+      'description' => t('Access links to flush caches.'),
```
jenlampton: These aren't always links anymore. Maybe `links and buttons` or maybe we leave that bit out? `Access to flush caches`? `Permission to flush caches`? 

why are the words the hardest part of any PR?

ghost: How about 'Provides access to flush caches from the user interface'?

jenlampton: Maybe manually?  
`Provides access to manually flush the caches from the user interface.` 
or even `Provides access to manually flush the caches.`

klonos (approved):
:+1:

quicksketch:
Looks good! Merged into 1.x and 1.13.x.

PR #2725, opened 2019-06-04
Issue 3849: Slight changes to config_install_default_config().

ghost:
Addresses https://github.com/backdrop/backdrop-issues/issues/3849

### DIFF
--- a/core/includes/config.inc
+++ b/core/includes/config.inc
@@ -252,6 +252,7 @@ function config_load_multiple($names, $type = 'active') {
 function config_install_default_config($module, $config_name = NULL) {
   $module_config_dir = backdrop_get_path('module', $module) . '/config';
   if (is_dir($module_config_dir)) {
+    $storage = new ConfigFileStorage($module_config_dir);
     $files = glob($module_config_dir . '/' . '*.json');
     foreach ($files as $file) {
       // Load config data into the active store and write it out to the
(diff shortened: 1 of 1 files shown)


core/includes/config.inc
```diff
@@ -251,8 +251,7 @@ function config_load_multiple($names, $type = 'active') {
  */
 function config_install_default_config($module, $config_name = NULL) {
   $module_config_dir = backdrop_get_path('module', $module) . '/config';
-  $storage = new ConfigFileStorage($module_config_dir);
-  if (is_dir(backdrop_get_path('module', $module) . '/config')) {
+  if (is_dir($module_config_dir)) {
```
klonos, line 254: I see how I would do the same here @BWPanda 👍

drupol (approved):
:+1:

core/includes/config.inc
```diff
@@ -262,6 +261,7 @@ function config_install_default_config($module, $config_name = NULL) {
       $file = array_pop($parts);
       $file_config_name = str_replace('.json', '', $file);
       if (is_null($config_name) || $file_config_name === $config_name) {
+        $storage = new ConfigFileStorage($module_config_dir);
```
klonos: Not sure what this does and how things work, so cannot comment on this change (other than making this comment here that is 😄)

drupol: I don't know either yet, but the proposed change make sense to me.

klonos (approved):
👍

core/includes/config.inc
klonos: This seems to have been a mishap. @BWPanda fixed it. Everything makes sense now 😄

quicksketch:
Merged into 1.x and 1.13.x. Thanks @BWPanda, @drupol, and @klonos!

PR #2738, opened 2019-06-15
Issue #3869: [UX] Taxonomy term list: fix table headers when Language module is enabled

klonos:
https://github.com/backdrop/backdrop-issues/issues/3869


quicksketch:

PR #2750, opened 2019-06-24
Issue #3643: CKEditor: Link to page points to file with corresponding file extension

### LINKED ISSUE #3643: CKEditor: Link to existing content gets confused by paths that match file extensions (like /pdf and .pdf)
Using CKEditor's link dialog autocomplete, a link to a page with a title corresponding to an allowed and used file extension doesn't link to the page but to a file.

**Example:** A link to a page with title "PDF" doesn't point to the "PDF" page but to a `.pdf` file.

(The issue happens also with other file types, e.g. with a page called "XML" and a `.xml` file.)

**Steps To Reproduce**
1. Check if your site manages at least one `.pdf` file on `admin/content/files`.
2. If there's no `.pdf` file:
  Make sure you have a content type where you can upload a PDF file via a *File* field.
  Create a content item, and attach a PDF file to it.

3. Add a page with title "PDF", e.g. `example.com/pdf`.
4. Add a post, write some body text, and use the link dialog to link to the "PDF" page.
5. Save the post.

**Actual behavior**
The link points to *one of the PDF files* of the site, e.g. `example.com/files/dummy.pdf`.

**Expected behavior**
The link points to *the PDF page* which had been chosen in the link dialog.

**Additional information**
- Backdrop 1.12.5 (also earlier versions)
- Web server and OS: happens in different environments

---

PR by @olafgrabienski: https://github.com/backdrop/backdrop/pull/2750

olafgrabienski:
Fixes https://github.com/backdrop/backdrop-issues/issues/3643.

### DIFF
--- a/core/modules/filter/filter.pages.inc
+++ b/core/modules/filter/filter.pages.inc
@@ -630,7 +630,7 @@ function _filter_get_file_id($href) {
   $result = 0;
   
   // Check URI for valid file scheme.
-  if(!file_valid_uri($href)){
+  if (!file_valid_uri($href)){
     return $result;
   }
 
--- a/core/modules/filter/filter.pages.inc
+++ b/core/modules/filter/filter.pages.inc
@@ -628,6 +628,11 @@ function _filter_format_editor_link_url_validate(&$element, &$form_state) {
  */
 function _filter_get_file_id($href) {
   $result = 0;
+  
+  // Check URI for valid file scheme.
+  if(!file_valid_uri($href)){
+    return $result;
+  }
 
   // Extract file name and directory from $href.
   $pos = strrpos($href, "/");


core/modules/filter/filter.pages.inc
```diff
@@ -628,6 +628,11 @@ function _filter_format_editor_link_url_validate(&$element, &$form_state) {
  */
 function _filter_get_file_id($href) {
   $result = 0;
+  
+  // Check URI for valid file scheme.
+  if(!file_valid_uri($href)){
```
quicksketch:
```suggestion
  if (!file_valid_uri($href)){
```
Just a missing space here.

olafgrabienski: Oh yes, thanks! Can I just hit „Commit suggestion“ in such a case? I forgot. And the term „commit“ is irritating me.

quicksketch: Yep that works. Thanks!

quicksketch:

PR #2751, opened 2019-06-24
Issue #3566: Add a Backdrop News block to the dashboard.

serundeputy:


serundeputy:
closing ... to reopen to build a new sandbox.

core/modules/dashboard/config/layout.layout.dashboard.json
```diff
@@ -7,7 +7,7 @@
     "renderer_name": "standard",
     "module": "dashboard",
     "weight": 0,
-    "storage": 4,
+    "storage": 2,
```
docwilmot: Dont get why change this.

core/modules/dashboard/includes/block.news.inc
```diff
@@ -0,0 +1,94 @@
...
+  public function getAdminPreview() {
+    return t('Displays a feed of news from BackdropCMS.org.');
+  }
+
+  /**
+   * {@inheritdoc}
+   */
+  public function getContent() {
+    $feed = backdrop_http_request('https://backdropcms.org/api/backdrop-news');
+    $news = json_decode($feed->data);
```
docwilmot: Should move this inside the `if ($feed->code === '200') {` conditional

core/modules/dashboard/includes/block.news.inc
```diff
@@ -0,0 +1,94 @@
...
+    $links = array();
+
+    if ($feed->code === '200') {
+      foreach ($news->nodes as $key => $node) {
+        $links[] = l(
+          $node->node->title,
+          $node->node->field_link
+        ) . ' - ' . $node->node->field_tagline;
+      }
+    }
```
docwilmot, line 51: Would it be worth checking if the error is 404 etc and changing the message to coulndt connect or offline or something?

core/modules/dashboard/includes/block.news.inc
```diff
@@ -0,0 +1,94 @@
...
+        '#items' => $links,
+      );
+    }
+
+    return $build;
+  }
+  /**
+   * {@inheritdoc}
+   */
+  public function form(&$form, &$form_state) {
```
docwilmot, line 75: Could we just not override the form, and leave no message?

core/modules/dashboard/tests/dashboard.test
```diff
@@ -108,6 +108,9 @@ class DashboardTest extends BackdropWebTestCase {
     // The manage user accounts link is present.
     $this->assertLink('Manage user accounts', 0, 'The Manage user accounts link was found.');
 
+    // The bckdrop-news block was found.
```
klonos: typo

core/modules/dashboard/includes/block.news.inc
```diff
@@ -0,0 +1,94 @@
...
+    parent::form($form, $form_state);
+
+    $form['url'] = array(
+      '#type' => 'markup',
```
klonos:
```suggestion
      '#type' => 'help',
```

core/modules/dashboard/includes/block.news.inc
```diff
@@ -0,0 +1,94 @@
...
+      '#markup' => t(
+        'This block is providing a news feed from BackdropCMS.org and the
```
klonos: Coding standards please - we do not wrap text.

core/modules/dashboard/includes/block.news.inc
```diff
@@ -0,0 +1,94 @@
...
+    else {
+      $build['list'] = array(
+        '#theme' => 'item_list',
+        '#items' => $links,
+      );
+    }
+
+    return $build;
+  }
+  /**
```
klonos, line 72: Add empty line before.

core/modules/dashboard/includes/block.news.inc
```diff
@@ -0,0 +1,94 @@
...
+   */
+  public function getAdminTitle() {
+    return t('Backdrop News');
+  }
+
+  /**
+   * {@inheritdoc}
+   */
+  public function getAdminPreview() {
+    return t('Displays a feed of news from BackdropCMS.org.');
```
klonos: Lets be consistent with the description text:
```suggestion
    return t('Displays a news feed from BackdropCMS.org.');
```

core/modules/dashboard/config/layout.layout.dashboard.json
quicksketch: Yeah, accidental I think. `4` means "default" storage, which it should be on new installs.

core/modules/dashboard/includes/block.news.inc
serundeputy: that happens in the `if (empty($links))`

core/modules/dashboard/includes/block.news.inc
serundeputy: we can; should we?

core/modules/dashboard/includes/block.news.inc
docwilmot: I mean change the message. If 200 and empty, `'#markup' => t('No news at this time.')`, if 404 `'#markup' => t('Could not connect to server.')` etc. Not sure its worth all that though.

core/modules/dashboard/includes/block.news.inc
docwilmot: This would be a new pattern as I dont recall blocks without form input elements having markup. Usually theyre just blank, like the Local Items block or the Powered By block. Plus this block may be used anywhere, not just on the dashboard, so I think the message might be redundant.

klonos (approved):
LGTM 👍

core/modules/dashboard/config/layout.layout.dashboard.json
```diff
@@ -170,6 +171,30 @@
...
+            "data": {
+                "module": "dashboard",
+                "delta": "news",
+                "settings": {
+                    "title_display": "default",
+                    "title": "",
+                    "style": "default",
+                    "block_settings": [],
+                    "contexts": [],
+                    "url": "https://api.serundeputy.io/api/views/events_past_api"
```
klonos: ...this will need to be changed too.

core/modules/dashboard/includes/block.news.inc
serundeputy: I've never noticed that about the `Powered by` block, but honestly I don't like it.  It seems confusing, i.e. I wish that one had some message as well saying what the purpose of the block is, why I can't edit it, and telling me that I can remove it.

I know I can remove it but does everyone?  Seems a bad UX to me.

core/modules/dashboard/includes/block.news.inc
```diff
@@ -0,0 +1,93 @@
...
+  /**
+   * {@inheritdoc}
+   */
+  public function form(&$form, &$form_state) {
+    parent::form($form, $form_state);
+
+    $form['url'] = array(
+      '#type' => 'help',
+      '#markup' => t(
+        'This block is providing a news feed from BackdropCMS.org and the contents can not be editted. You can remove it from the dashboard if you wish.'
```
klonos: Typo: `editted`. Also, as @docwilmot pointed, people might use this block elsewhere, so we should remove the reference to the dashboard. Finally, it is common to use double quotes for strings that are to be translated, as single quotes are often used as apostrophes, and would need to be escaped otherwise.

Suggestion:

```suggestion
        "This block provides a news feed from BackdropCMS.org. Its contents can not be edited."
```

...also, the contents of the `t()` function should not be wrapped onto a separate line.

core/modules/dashboard/includes/block.news.inc
klonos: I agree. I do find it bad UX to see "blank" blocks; so I like having a short description in such cases.

core/modules/dashboard/includes/block.news.inc
serundeputy: This string is not wrapped; the string itself is moved to a separate line; `'` are faster to parse for PHP and if there are not `'`s in the string itself the single quote wrapper is preferred in Backdrop.  

Thanks for spotting the typo.

jenlampton:
Replaced by https://github.com/backdrop/backdrop/pull/3026

PR #2763, opened 2019-06-29
Issue #1990: Equal height regions/columns in Layout UI.

### LINKED ISSUE #1990: Regions that are side-by-side should have the same height, independently of contained blocks.
Not sure if this is a feature request or a bug report. It's been bothering me for a while, but never filed an issue for it till now. Here's what I mean:

backdrop-layout-regions-different_heights

In my custom front-end themes in D7, I used to use https://www.drupal.org/project/equalheights ...especially when I wanted the main content to be of equal height as the sidebars.

Here's how things would look instead:

backdrop-layout-regions-equal_heights

I personally feel that it's much better visually.

---
PR by @BWPanda (using CSS `flex`): https://github.com/backdrop/backdrop/pull/2763

ghost:
Fixes https://github.com/backdrop/backdrop-issues/issues/1990

### DIFF
--- a/core/modules/layout/css/layout.admin.css
+++ b/core/modules/layout/css/layout.admin.css
@@ -324,7 +324,8 @@
   display: inline-block;
 }
 .layout-content-form main {
-  margin-bottom: 0; /* Override Seven's styling. */
+  /* Override Seven's styling. */
+  margin-bottom: 0;
 }
 
 @media (min-width: 768px) {
--- a/core/modules/layout/css/layout.admin.css
+++ b/core/modules/layout/css/layout.admin.css
@@ -132,6 +132,12 @@
   border-bottom: none;
 }
 
+/* Layout columns */
+.layout-content-form .row > div,
+.layout-content-form .row > main {
+  display: flex;
+}
+
 /* Layout editor. */
 .layout-editor-region {
   width: 100%;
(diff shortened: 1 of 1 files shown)


core/modules/layout/css/layout.admin.css
```diff
@@ -317,6 +323,9 @@
 .layout-access-list .layout-condition .layout-operations {
   display: inline-block;
 }
+.layout-content-form main {
+  margin-bottom: 0; /* Override Seven's styling. */
```
klonos: Coding standards say that only `/* LTR */` is allowed next to CSS properties. This should go above the ruleset (in which case, a blank line above it is also required).

ghost: You're right about the LTR inline comments, thanks for pointing that out. But I disagree with putting the comment above the ruleset as it just applies to that one property.

> Place the comment on the line immediately above the **property** or ruleset it describes. The comment should be indented the same amount as the **property** or ruleset it describes.
**If** the comment is describing a ruleset, place a blank line before the comment.

(emphasis mine)

So I think it's ok to put where I did, and without a blank line before it. PR updated.

klonos (approved):
👍

klonos (approved):
👍

quicksketch:
Thanks @BWPanda! This is a great fix. Merged into 1.x and 1.13.x.

PR #2773, opened 2019-07-02
Issue #1366: Added validation to max/min settings.

### LINKED ISSUE #1366: Add validation to prevent maximum value being less than minimum value
The core Number module has validation functions in `number_field_instance_settings_form()` to make sure maximum and minimum values are numbers. But there's nothing to stop the maximum value being set less than the minimum value...

I'm not sure if you could simply add another 'element_validate' function to the maximum field, or if you'd have to use a form-wide validation function instead since you'll need to get the value of the minimum field too to compare.

I noticed this in Drupal, but figured it'd have a better chance of being fixed here instead ;)

---
PR by @BWPanda: https://github.com/backdrop/backdrop/pull/2773

ghost:
Fixes https://github.com/backdrop/backdrop-issues/issues/1366


klonos (approved):
LGTM 👍

quicksketch:

PR #2775, opened 2019-07-02
Issue #1887: Display correct default value for views' block titles.

### LINKED ISSUE #1887: [UX] Improve the admin label for Views block displays
backdrop-views_ui-block_label-none_should_read_default

We should at the very least have it say "Default". We should definitely change that, but if we are to change it, we might as well take it up a notch...

Since "default" might seem cryptic (because it does not actually convey what the actual name will be), I propose to either describe what elements are going to be used, or display the actual value.

So instead of "none" or "default", it should either be "Use view and view display name" or better have it be "View: [view_name]: [view_display_name]" (with the tokens dynamically replaced by the actual values of course and we should fix #1889 too).

ghost:
Fixes https://github.com/backdrop/backdrop-issues/issues/1887


stpaultim:
@BWPanda - Sandbox isn't working for me. Can you restart it?

PR #2776, opened 2019-07-02
Issue #1926: Allow prevention of automatic linking of URLs and email addresses.

### LINKED ISSUE #1926: Prevent URLs or email addresses turning into links
Scenario: I'm writing a page and in the Body field I want to enter a dummy email address (e.g. info@mybusiness.com). When I save the page, the text filters are run and the email address is turned into a clickable link. I don't want that. However, I cannot remove the 'Convert URLs into links' filter from the text format, as I need other pages to have clickable links (e.g. URLs or real email addresses), and I cannot create a new text format without the URL filter to use for this specific page as I may need other URLs or email addresses on the same page to be turned into links...

So I'm wondering if there's a way (or if not yet, can we make one) where I can tell a specific URL or email address to not turn into a link (either before or after saving the node)?

---
PR by @BWPanda: https://github.com/backdrop/backdrop/pull/2776

Advocate: @BWPanda

ghost:
Fixes https://github.com/backdrop/backdrop-issues/issues/1926


core/modules/filter/filter.module
```diff
@@ -2161,6 +2162,10 @@ function _filter_url($text, $filter) {
           if (preg_match("`<($ignore_tags)(?:\s|>)`i", $chunks[$i], $matches)) {
             $open_tag = $matches[1];
           }
+          // Check whether this tag has a class contained in $ignore_classes.
+          elseif (preg_match("`<([a-z0-9]+)\s+[\s\S]*?class=[\"'][\s\w-]*(?:$ignore_classes)[\"'\s]+(?:\s|>)`i", $chunks[$i], $matches)) {
```
klonos:
```suggestion
          elseif (preg_match("`<[a-z0-9-]+\s.*class=[\"']{1}.*\b(?<!\-)$ignore_classes\b(?!\-).*[\"']{1}\s?>`i", $chunks[$i], $matches)) {
```

...see my comment in https://github.com/backdrop/backdrop-issues/issues/1926#issuecomment-653849722

ghost:
Merged into 1.x.

PR #2778, opened 2019-07-04
Issue #3834: Use human name for vocabularies on URL alias bulk actions page

### LINKED ISSUE #3834: [UX] The url alias bulk actions page uses machine name instead of human name
**Description of the bug**
I can't tell which vocabulary is which:
Screen Shot 2019-05-31 at 10 41 38 AM

**Steps To Reproduce**
To reproduce the behavior:
1. Create a vocabulary with differnt machine name from human name
2. Visit admin/config/urls/path/bulk-update
3. View taxonomy section

**Actual behavior**
Machine names visible

**Expected behavior**
Human names visible

**Additional information**
My screenshot shows a site that was upgraded from Drupal 6, where the machine names are indistinguishable. This is a worst case scenario, but the problem should still be fixed. On the patterns page it uses the human name:
Screen Shot 2019-05-31 at 10 44 20 AM

isabelhaas:
Fixes backdrop/backdrop-issues#3834

The URL bulk actions page appears to be using the machine name instead of the human name because the pattern items descriptions are keyed by machine name. This PR takes a substring of the pattern items description string, “URL alias pattern for all [human name] URLs”, to get the human name, but I’m wondering if there’s a better way to do this. I didn’t see another way to access the human name from within the path_bulk_update_form() function that makes this page, because the info it receives from path_get_info() does not appear to include the human name outside of that string, but I’m wondering if there is another way to get the human name.

### DIFF
--- a/core/modules/filter/filter.pages.inc
+++ b/core/modules/filter/filter.pages.inc
@@ -175,7 +175,7 @@ function filter_format_editor_image_form($form, &$form_state, $format) {
     // Remember this is editing, not inserting. In the event the image is
     // removed and re-uploaded.
     $form_state['storage']['editing_image'] = TRUE;
-    backdrop_set_title(t('Edit image details'));
+    backdrop_set_title(t('Edit image'));
     $submit_button_label = t('Update');
   }
   else {
--- a/core/modules/filter/filter.pages.inc
+++ b/core/modules/filter/filter.pages.inc
@@ -175,7 +175,7 @@ function filter_format_editor_image_form($form, &$form_state, $format) {
     // Remember this is editing, not inserting. In the event the image is
     // removed and re-uploaded.
     $form_state['storage']['editing_image'] = TRUE;
-    backdrop_set_title(t('Edit image'));
+    backdrop_set_title(t('Edit image details'));
     $submit_button_label = t('Update');
   }
   else {
--- a/core/modules/path/path.admin.inc
+++ b/core/modules/path/path.admin.inc
@@ -644,9 +644,10 @@ function path_bulk_update_form($form, &$form_state) {
       if (isset($path_info['pattern items'])) {
         foreach ($path_info['pattern items'] as $item => $item_label) {
           $human_name = $item;
-          // Use the human name in the row title for vocabularies, not the machine name they are keyed by.
-          if ($type == taxonomy_term) {
-            $human_name = substr(strip_tags($item_label), 26, -5);
+          if ($type == 'taxonomy_term') {
+            //For vocabularies, use their human name in the checkbox label instead of their 
+            //machine name.
+            $human_name = $path_info[$item . '-human'];
           }
           $form['update'][$type][$item] = array(
             '#type' => 'checkbox',
--- a/core/modules/path/path.admin.inc
+++ b/core/modules/path/path.admin.inc
@@ -643,9 +643,14 @@ function path_bulk_update_form($form, &$form_state) {
       // vocabularies.
       if (isset($path_info['pattern items'])) {
         foreach ($path_info['pattern items'] as $item => $item_label) {
+          $human_name = $item;
+          // Use the human name in the row title for vocabularies, not the machine name they are keyed by.
+          if ($type == taxonomy_term) {
+            $human_name = substr(strip_tags($item_label), 26, -5);
+          }
           $form['update'][$type][$item] = array(
             '#type' => 'checkbox',
-            '#title' => t('All !label URL aliases', array('!label' => $item)),
+            '#title' => t('All !label URL aliases', array('!label' => $human_name)),
             '#attributes' => array(
               'class' => array('path-type', 'path-' . $type),
               'data-path-type' => array($type),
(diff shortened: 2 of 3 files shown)


core/modules/filter/filter.pages.inc
```diff
@@ -175,7 +175,7 @@ function filter_format_editor_image_form($form, &$form_state, $format) {
     // Remember this is editing, not inserting. In the event the image is
     // removed and re-uploaded.
     $form_state['storage']['editing_image'] = TRUE;
-    backdrop_set_title(t('Edit image'));
+    backdrop_set_title(t('Edit image details'));
```
serundeputy: Is this from a different PR you are working on seems unrelated to this issue?

core/modules/path/path.admin.inc
```diff
@@ -643,9 +643,14 @@ function path_bulk_update_form($form, &$form_state) {
       // vocabularies.
       if (isset($path_info['pattern items'])) {
         foreach ($path_info['pattern items'] as $item => $item_label) {
+          $human_name = $item;
+          // Use the human name in the row title for vocabularies, not the machine name they are keyed by.
+          if ($type == 'taxonomy_term') {
+            $human_name = substr(strip_tags($item_label), 26, -5);
```
serundeputy: your intuition is correct; what we need here is to expose the Human Name as part of the `pattern items` array.

The `$path_info` variable is set via line 624:

```php
  $all_path_info = path_get_info();
```

That function in turn calls all the callbacks from any module that implements that hook; in this case the `taxonomy` module; The hook implementation in `taxonomy` is in the file: `taxonomy.path.inc` on line 10.  In the for loop that is populating the `$info` variable we need something like this in both branches of the `if`:

```php
$info['taxonomy_term']['pattern items'][$vocabulary_name . '-human'] = $vocabulary->name;
```

That will expose the human name here so you can just grab it instead of having to to the string manipulations here.

serundeputy (changes requested):

core/modules/path/path.admin.inc
klonos: ...or just do this:

```suggestion
            $item = taxonomy_vocabulary_load($item)->name;
```

...and don't introduce the `$human_name` variable at all 😉

core/modules/filter/filter.pages.inc
isabelhaas: Sorry, my mistake, this was from a different PR, I didn't mean to include it in this PR.

core/modules/path/path.admin.inc
```diff
@@ -643,9 +643,14 @@ function path_bulk_update_form($form, &$form_state) {
       // vocabularies.
       if (isset($path_info['pattern items'])) {
         foreach ($path_info['pattern items'] as $item => $item_label) {
+          $human_name = $item;
+          // Use the human name in the row title for vocabularies, not the machine name they are keyed by.
```
klonos: This comment exceeds the 80 chars max limit. Also, since this is not a table in an end-user context, may I suggest changing it to something like this:

> For vocabularies, use their human name in the checkbox label instead of their machine name.

...and while at it, perhaps also move it inside the `if` conditional?

core/modules/path/path.admin.inc
serundeputy: @klonos @isabelhaas  the issue here is performance; this submit handler is already passing off to any module that implements `hook_path_info` and in `taxonomy_path_info()` we are already loading the vocabularies in `$vocabularies = taxonomy_get_vocabularies();` so loading it again at this step is more memory and less efficient.

It would work but is not optimized.

serundeputy: @isabelhaas I also see that putting the human name here:

```php
$info['taxonomy_term']['pattern items'][$vocabulary_name . '-human'] = $vocabulary->name;
```

makes too many checkboxes; :) ; we can put it here instead:

```php
$info['taxonomy_term'][$vocabulary_name . '-human'] = $vocabulary->name;
```

core/modules/path/path.admin.inc
isabelhaas: @klonos Thanks for the feedback! I changed the comment to what you suggested and moved it inside the `if`.

core/modules/path/path.admin.inc
isabelhaas: @serundeputy Thanks for the help! I did what you suggested, could you take a look and see if what's there now looks correct?

core/modules/path/path.admin.inc
```diff
@@ -643,9 +643,15 @@ function path_bulk_update_form($form, &$form_state) {
       // vocabularies.
       if (isset($path_info['pattern items'])) {
         foreach ($path_info['pattern items'] as $item => $item_label) {
+          $human_name = $item;
+          if ($type == 'taxonomy_term') {
+            //For vocabularies, use their human name in the checkbox label instead of their 
+            //machine name.
+            $human_name = $path_info[$item . '-human'];
+          }
```
klonos: There should be a space between each leading `//` and the beginning of the respective comment line.

Other than that, looks good 👍 ...if I was to suggest anything, perhaps this should be an `if`/`else`? ...or something like this?:

```
// For vocabularies, use their human name in the checkbox label instead of their 
// machine name.
$human_name = $type == 'taxonomy_term' ? $path_info[$item . '-human'] : $item;
```

isabelhaas: Thanks @klonos! I made the changes you suggested.

PR #2779, opened 2019-07-05
Addresses Issue #3911

ak2243:
I tried to change the Save configuration button too redirect to the “Uninstall modules” page but it didn’t work

### DIFF
--- a/core/modules/system/system.admin.inc
+++ b/core/modules/system/system.admin.inc
@@ -1006,7 +1006,7 @@ function system_modules_submit($form, &$form_state) {
       array('query' => array('search' => $parameter))
     );
   }
-  elseif(sizeof($actions['enable']) != 0) {
+  elseif(sizeof($actions['enable']) != 0)) {
     $form_state['redirect'] = 'admin/modules';
   }
   else {
--- a/core/modules/system/system.admin.inc
+++ b/core/modules/system/system.admin.inc
@@ -676,8 +676,10 @@ function system_modules($form, $form_state = array()) {
 
   $form['actions'] = array('#type' => 'actions');
   $form['actions']['submit'] = array(
-    '#type' => 'submit',
+    '#type' => 'link',
     '#value' => t('Save configuration'),
+    #href => 'admin/modules/uninstall',
+
   );
   $form['#action'] = url('admin/modules/list/confirm');
 
--- a/core/modules/system/system.admin.inc
+++ b/core/modules/system/system.admin.inc
@@ -999,15 +999,15 @@ function system_modules_submit($form, &$form_state) {
   }
 
   // Check if a previous search is present and preserve it on redirect.
-  if (!empty($form_state['values']['search']) && sizeof($actions['enable']) != 0)) {
+  if(sizeof($actions['enable']) != 0) {
+    $form_state['redirect'] = 'admin/modules';
+  }
+  elseif (!empty($form_state['values']['search']) && sizeof($actions['enable']) != 0) {
     $parameter = $form_state['values']['search'];
     $form_state['redirect'] = array(
       'admin/modules',
       array('query' => array('search' => $parameter))
     );
-  }
-  elseif(sizeof($actions['enable']) != 0)) {
-    $form_state['redirect'] = 'admin/modules';
   }
   else {
     $form_state['redirect'] = 'admin/modules/uninstall';
--- a/core/modules/system/system.admin.inc
+++ b/core/modules/system/system.admin.inc
@@ -676,8 +676,9 @@ function system_modules($form, $form_state = array()) {
 
   $form['actions'] = array('#type' => 'actions');
   $form['actions']['submit'] = array(
-    '#type' => 'submit',
+    '#type' => 'link',
     '#value' => t('Save configuration'),
+    '#href' => 'admin/modules/uninstall',
   );
   $form['#action'] = url('admin/modules/list/confirm');
 
--- a/core/modules/system/system.admin.inc
+++ b/core/modules/system/system.admin.inc
@@ -676,10 +676,8 @@ function system_modules($form, $form_state = array()) {
 
   $form['actions'] = array('#type' => 'actions');
   $form['actions']['submit'] = array(
-    '#type' => 'link',
+    '#type' => 'submit',
     '#value' => t('Save configuration'),
-    #href => 'admin/modules/uninstall',
-
   );
   $form['#action'] = url('admin/modules/list/confirm');
 
--- a/core/modules/system/system.admin.inc
+++ b/core/modules/system/system.admin.inc
@@ -679,7 +679,7 @@ function system_modules($form, $form_state = array()) {
     '#type' => 'submit',
     '#value' => t('Save configuration'),
   );
-  $form['#action'] = url('admin/modules/uninstall');
+  $form['#action'] = url('admin/modules/list/confirm');
   return $form;
 }
 
@@ -1009,6 +1009,7 @@ function system_modules_submit($form, &$form_state) {
   else {
     $form_state['redirect'] = 'admin/modules';
   }
+  $form_state['redirect'] = 'admin/modules/uninstall';
 }
 
 /**
--- a/core/modules/system/system.admin.inc
+++ b/core/modules/system/system.admin.inc
@@ -676,12 +676,10 @@ function system_modules($form, $form_state = array()) {
 
   $form['actions'] = array('#type' => 'actions');
   $form['actions']['submit'] = array(
-    '#type' => 'link',
+    '#type' => 'submit',
     '#value' => t('Save configuration'),
-    '#href' => 'admin/modules/uninstall',
   );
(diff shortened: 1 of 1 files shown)


docwilmot:
@arjunk2004 thanks for your help. The the button doesnt show up because the `link` form element is not meant to have a `#value`, it should instead have a `#title`; if you changed that it should appear. But of course it would not be a button anymore, it would just be a hyperlink.

Check out [our Form API reference](https://api.backdropcms.org/form_api) for info about form elements.

We cant however have a link there because this is a form, and users expect to take action to submit a form, and a link isnt really an action, it just takes you to another page, so that would be confusing to users. So for forms, we want buttons.

Also if you had a link instead of a button the form actions wouldnt work. Buttons should have a `#type => 'submit'` so that when you click the button a submit handler would execute to process the form. For this function the submit handler would be `system_modules_submit()`, and that does all the sorting, then at the end calls `module_enable()` or `module_disable()` on each module. 

What we would want in this case then is to leave the submit button, to allow `system_modules_submit()` to run and do its thing, but then at the end of `system_modules_submit()`, have a line that causes the page to redirect to `admin/modules/uninstall`, usually by setting `$form_state['redirect'] = 'admin/modules/uninstall'` or rarely calling `backdrop_goto('admin/modules/uninstall')`. So in that function as well we would have some logic that checks if the module(s) being disabled have uninstall hooks, and do the redirect only if that is true.

I must say though that `system_modules()` is a real beast of a function, and same for its submit handler. This would be difficult if you wanted to try.

arjunk2004:
Thanks for the help. I did realize, however, that there was a way to do this because there was a line that reloaded to and showed a confirmation message. So, I changed the url on that to the uninstall page. I think that should do it for now.

ak2243:
So, I just added a redirect at the end of the submit handler. The result should be that it saves, and then shows the confirmation message on the uninstall page.

serundeputy:
Nice work @arjunk2004 !

Looks like the fails are:

```bash
---- Module: Filter modules and reset filter (FilterResetTestCase) ----


Fail      Other      system.test        114 FilterResetTestCase->testFilterRese
    Current URL is '/admin/modules?search=image'.
Fail      Browser    system.test        117 FilterResetTestCase->testFilterRese
    Link Reset does not exist on http://localhost/admin/modules/uninstall
Fail      Other      system.test        118 FilterResetTestCase->testFilterRese
    Current URL is '/admin/modules'.
```

this is because we added some module filter functionality for a workflow like:

* search for `devel` on `/admin/modules`
* check of `devel` (to enable it)
* submit the form
  * form enables `devel` and returns you to `/admin/modules` pre-sorted to `devel` on the filter

So, it seems there is another couple of use cases here:

  * User enables some module after using the filter
    * Expectation is to return to `/admin/modules` sorted to that filter
  * User enables several modules
    * Expectation return user to `/admin/modules`
    * sorted if they had a filter
  * User enables some module(s) and disables some module(s)
    * not sure but I think the enabling should take precedence and return them to `/admin/modules`

So you'll want to find out where those decisions happen in the submit handler and make sure we account for each.

jenlampton:
So @serundeputy do you think all the test fails come from not checking for modules being disabled before doing the redirect? 

I agree, but I didn't get that from your comment :)

ak2243:
@serundeputy is that what you meant? I accounted for the three scenarios, I think.

core/modules/system/system.admin.inc
```diff
@@ -680,7 +680,6 @@ function system_modules($form, $form_state = array()) {
     '#value' => t('Save configuration'),
   );
   $form['#action'] = url('admin/modules/list/confirm');
-
```
jenlampton, line 683: looks like a line was removed from here, probably by mistake?

jenlampton:
@arjunk2004 you're getting closer, but we're not quite there.

We don't need to check if any modules were enabled, we need to see if any modules were disabled - and only add the redirect to the uninstall page if they were.

We should pass along the search query regardless of where they are being redirected. This will enable the following workflow:

1) Visit the modules page
2) check/uncheck boxes to both enable and disable a module at the same time
3) Submit the modules form
4) Land on the uninstall page to uninstall the module that was disabled
5) Submit the uninstall form
6) Land back on the modules page, with the search query pre-filled

core/modules/system/system.admin.inc
```diff
@@ -1000,16 +999,21 @@ function system_modules_submit($form, &$form_state) {
   }
 
   // Check if a previous search is present and preserve it on redirect.
-  if (!empty($form_state['values']['search'])) {
+  if (!empty($form_state['values']['search']) && sizeof($actions['enable']) != 0)) {
```
jenlampton: this is fine as-is as we probably don't need the search term if the module is being uninstalled...

core/modules/system/system.admin.inc
```diff
@@ -1000,16 +999,21 @@ function system_modules_submit($form, &$form_state) {
...
     $form_state['redirect'] = array(
       'admin/modules',
       array('query' => array('search' => $parameter))
     );
   }
-  else {
+  elseif(sizeof($actions['enable']) != 0) {
     $form_state['redirect'] = 'admin/modules';
   }
+  else {
```
jenlampton, line 1012: This shouldn't be inside an else, it needs to be inside a check if any modules were disabled...

Maybe something like ...
```
if (sizeof($actions['disable']) >= 1) {
  $form_state['redirect'] = 'admin/modules/uninstall';
} 
else {
  $form_state['redirect'] = 'admin/modules';
}

if (!empty($form_state['values']['search']) && sizeof($actions['enable']) != 0)) {
  $form_state['redirect'][] = array('query' => array('search' => $parameter));
}

```

We'll also need to add something to the uninstall form to catch this search value and pass it back to the 'admin/modules' page.

core/modules/system/system.admin.inc
ak2243: Doo you mean the original is fine or my change is fine?

ak2243:
> @arjunk2004 you're getting closer, but we're not quite there.
> 
> We don't need to check if any modules were enabled, we need to see if any modules were disabled - and only add the redirect to the uninstall page if they were.
> 
> We should pass along the search query regardless of where they are being redirected. This will enable the following workflow:
> 
> 1. Visit the modules page
> 2. check/uncheck boxes to both enable and disable a module at the same time
> 3. Submit the modules form
> 4. Land on the uninstall page to uninstall the module that was disabled
> 5. Submit the uninstall form
> 6. Land back on the modules page, with the search query pre-filled

So if they enabled it shouldn’t we stay on this page as @serundeputy suggested?

core/modules/system/system.admin.inc
ak2243: would that make a difference?

jenlampton (commented):
I meant that adding the addition of `sizeof($actions['enable']) != 0` is fine :)

jenlampton:
> So if they enabled it shouldn’t we stay on this page as @serundeputy suggested?

I'm not sure that's what @serundeputy suggested, but had a hard time following his comment...

If anyone disables any module it needs to redirect to the uninstall page. 

If they also enabled a module, we can redirect back to the modules page after the uninstall action happens (and retain the search query!) but the uninstall step cannot be skipped.

serundeputy:
@jenlampton @arjunk2004 the tests fail because there is a test for the query parameter on the module filter page that expects to land on the `/admin/modules` page after submitting the form; so with this change it does not land there and the checks the tests perform fail because of that; so those tests would need to be modified accordingly.

ak2243:
@jenlampton how do I pass on the filter?

core/modules/system/system.admin.inc
klonos: Code execution-wise, it would not matter @arjunk2004 ...usually for clarity (especially in long functions), we keep the return in a separate line + an empty line before it. It helps visually when going through code.

core/modules/system/system.admin.inc
klonos: I find repetition of the `sizeof($actions['enable']) != 0` condition in 2 separate conditionals superfluous. I think that this is a cleaner way to do it:

```
if (sizeof($actions['enable']) != 0) {
  // If there was a search query, preserve it when redirecting...
  if (!empty($form_state['values']['search'])) {
    $parameter = $form_state['values']['search'];
    $form_state['redirect'] = array('admin/modules', array('query' => array('search' => $parameter)));
  }
  // ...otherwise simply redirect to the modules list page.
  else {
    $form_state['redirect'] = 'admin/modules';
  }
}
else {
  $form_state['redirect'] = 'admin/modules/uninstall';
}
```

This is purely from looking at the code from a novice dev PoV; this may need to be rethought, considering the logic we are introducing.

ghost:
Since the discussion in https://github.com/backdrop/backdrop-issues/issues/3911 has moved away from issuing a redirect, I'm going to close this PR. See https://github.com/backdrop/backdrop/pull/4351 for a new PR that just sets a message instead.

PR #2781, opened 2019-07-08
Issue #1881: Allow teaser titles to be hidden in views.

### LINKED ISSUE #1881: Allow title to be hidden when displaying a teaser in views
I'd like to display a node teaser in a view, but without the node title being added and linking to the node itself. Currently you can choose whether or not links and comments are displayed, so I'd like to propose another option to display (or hide) the node title.

Something like this:
firefox_screenshot_2016-05-17t05-26-02 783z

ghost:
Fixes https://github.com/backdrop/backdrop-issues/issues/1881

### DIFF
--- a/core/modules/node/views/views_plugin_row_node_view.inc
+++ b/core/modules/node/views/views_plugin_row_node_view.inc
@@ -59,7 +59,7 @@ class views_plugin_row_node_view extends views_plugin_row {
     );
     $form['links'] = array(
       '#type' => 'checkbox',
-      '#title' => t("Display links (e.g. 'Read more')"),
+      '#title' => t("Display links (e.g. 'Read more', 'Add comment', etc.)"),
       '#default_value' => $this->options['links'],
     );
     $form['comments'] = array(
--- a/core/modules/node/views/views_plugin_row_node_view.inc
+++ b/core/modules/node/views/views_plugin_row_node_view.inc
@@ -55,11 +55,13 @@ class views_plugin_row_node_view extends views_plugin_row {
     $form['title'] = array(
       '#type' => 'checkbox',
       '#title' => t('Display title'),
+      '#description' => t('The title will display as a link to each item.'),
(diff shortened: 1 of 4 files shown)


core/modules/node/views/views_plugin_row_node_view.inc
```diff
@@ -50,10 +51,15 @@ class views_plugin_row_node_view extends views_plugin_row {
...
+    );
+    $form['title'] = array(
+      '#type' => 'checkbox',
+      '#title' => t('Display title'),
+      '#default_value' => $this->options['title'],
+    );
     $form['links'] = array(
       '#type' => 'checkbox',
-      '#title' => t('Display links'),
+      '#title' => t("Display links (e.g. 'Read more')"),
```
klonos:
```suggestion
      '#title' => t('Display links (e.g. "Read more", "Reply" etc.)'),
```

klonos (approved):
👍

core/modules/node/views/views_plugin_row_node_view.inc
ghost: I don't see any 'Reply' links, but there's an 'Add comment' one, so I've used that terminology instead.

core/modules/views_ui/wizards/views_ui_node_views_wizard.php
```diff
@@ -32,25 +32,22 @@ protected function build_form_style(&$form, &$form_state, $type) {
...
-          '#title_display' => 'invisible',
-          '#title' => t('Should comments be displayed below each node'),
-          '#options' => array(
-            1 => t('with comments'),
-            0 => t('without comments'),
-          ),
-          '#default_value' => 0,
+          '#type' => 'checkbox',
+          '#title' => t('Display comments'),
+          '#default_value' => FALSE,
```
klonos, line 50: Good call here 👍 ...it's just silly to have selects for dual on/off settings.

PR #2782, opened 2019-07-08
Issue #2933: Removed unused 'property_type' key.

### LINKED ISSUE #2933: Remove unused 'property_type' key from Email module
The [documentation](https://api.backdropcms.org/api/backdrop/core%21modules%21field%21field.api.php/function/hook_field_info/1) for `hook_field_info()` doesn't mention anything about a `property_type` key, but [`email_field_info()`](https://api.backdropcms.org/api/backdrop/core%21modules%21email%21email.module/function/email_field_info/1) uses it...

Can we add this to the documentation please?

---
PR by @BWPanda: https://github.com/backdrop/backdrop/pull/2782

ghost:
Fixes https://github.com/backdrop/backdrop-issues/issues/2933

### DIFF
--- a/core/modules/email/email.module
+++ b/core/modules/email/email.module
@@ -14,7 +14,6 @@ function email_field_info() {
       'description' => t('This field stores and renders email addresses.'),
       'default_widget' => 'email_textfield',
       'default_formatter' => 'email_default',
-      'property_type' => 'text',
     ),
   );
 }


klonos:
@BWPanda it seems that if we remove that line, we will most likely break integration with other modules (core and contrib). See my comment in https://github.com/backdrop/backdrop-issues/issues/2933#issuecomment-509272581

quicksketch:
Per the discussion this has been marked RTBC. As it may affect contrib needing to re-add this same property via an alter hook, I'll only merge into 1.x for 1.14.0.

PR #1910, opened 2017-06-11
Issue #2731: [UX] Manage blocks UI: No required indicator in "Admin l…

klonos:
…abel" field for custom blocks.

Fixes https://github.com/backdrop/backdrop-issues/issues/2731

PR #2783, opened 2019-07-08
Issue #2731: [UX] Manage blocks UI: Add required indicator in "Admin label" field for custom blocks.

### LINKED ISSUE #2731: [UX] Manage blocks UI: No required indicator in "Admin label" field for custom blocks.

While creating custom blocks from Structure -> Custom blocks -> Add custom block (`/admin/structure/block/add`) shows the "Admin label" field as required:

image

...when adding a custom block via the Manage blocks UI, it is not shown as required:

image

1. Edit any layout (the Default layout for example: `admin/structure/layouts/manage/default/blocks`)
2. Add a custom block in any region by hitting the "Add block" button.
3. Click the "Make this block reusable" checkbox -> The "Admin label" field is shown, but it does not have the red asterisk that signifies a required field:

    image

4. **Without adding an admin label**, hit the "Add block" button -> Validation error saying that the "Admin label" field is required:

image

The "Admin label" field is shown as an optional field in the add/edit custom block dialog.

That field should have a red asterisk, indicating that it is in fact required.

---
PR upcoming...

klonos:
Fixes https://github.com/backdrop/backdrop-issues/issues/2731

Replaces https://github.com/backdrop/backdrop/pull/1910


core/modules/layout/css/layout.admin.css
```diff
@@ -326,3 +326,19 @@
     float: left;
   }
 }
+
+/* Make sure that the machine name wraps under the "Admin label" field. */
+.layout-block-configure-form .form-item-label span.field-suffix {
+  display: block;
```
klonos: This is to fix this:

Screen Shot 2019-07-09 at 7 49 50 am

...happens only on high-res, wide screens.

core/modules/layout/css/layout.admin.css
```diff
@@ -337,6 +337,22 @@
   }
 }
 
+/* Make sure that the machine name wraps under the "Admin label" field. */
+.layout-block-configure-form .form-item-label span.field-suffix {
```
jenlampton: This CSS is now specific enough, it will target all suffixes for all fields on this form. If all we're after is the machine name, we should target that single item specifically.

core/modules/layout/css/layout.admin.css
klonos: FTR, we have a dedicated issue to fix this globally: https://github.com/backdrop/backdrop-issues/issues/1349, so I may have to leave this out of the PR. Let's see how things go in that issue though.

core/modules/layout/css/layout.admin.css
klonos: I've removed this CSS, and left it to be handled in https://github.com/backdrop/backdrop-issues/issues/1349

PR #2788, opened 2019-07-10
Issue #3923: Installer: Hide the "Choose profile" step if only one exclusive profile is present.

klonos:
https://github.com/backdrop/backdrop-issues/issues/3923

### DIFF
--- a/core/includes/install.core.inc
+++ b/core/includes/install.core.inc
@@ -535,16 +535,19 @@ function install_tasks($install_state) {
   // Determine whether translation import tasks will need to be performed.
   $needs_translations = count($install_state['translations']) > 1 && !empty($install_state['parameters']['langcode']) && $install_state['parameters']['langcode'] != 'en';
 
-  // Display profile task only when there is more than one non-hidden profile.
+  // Display the "Choose profile" task only when there is more than one
+  // non-hidden profiles; hide it if there is only a single exclusive profile.
   $hidden_profile_count = 0;
+  $exclusive_profile_count = 0;
   foreach ($install_state['profiles'] as $profile) {
     $profile_info = install_profile_info($profile->name);
     $hidden_profile_count += empty($profile_info['hidden']);
+    $exclusive_profile_count += !empty($profile_info['exclusive']);
   }
-  $display_select_profile = $hidden_profile_count > 1 && count($install_state['profiles']) != 1;
+  $display_select_profile = $hidden_profile_count > 1 && count($install_state['profiles']) != 1 && $exclusive_profile_count !=1;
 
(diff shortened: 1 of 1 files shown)


core/includes/install.core.inc
```diff
@@ -535,16 +535,19 @@ function install_tasks($install_state) {
...
   $needs_translations = count($install_state['translations']) > 1 && !empty($install_state['parameters']['langcode']) && $install_state['parameters']['langcode'] != 'en';
 
-  // Display profile task only when there is more than one non-hidden profile.
+  // Display the "Choose profile" task only when there is more than one
+  // non-hidden profiles; hide it if there is only a single exclusive profile.
   $hidden_profile_count = 0;
+  $exclusive_profile_count = 0;
   foreach ($install_state['profiles'] as $profile) {
     $profile_info = install_profile_info($profile->name);
     $hidden_profile_count += empty($profile_info['hidden']);
```
ghost, line 544: Can we fix this while we're at it? This seems counter-intuitive to me, in that it's not a count of hidden profiles, but non-hidden ones... I think we should either rename this variable (to `$visible_profile_count`), or change the logic to actually count and act on the number of hidden profiles.

core/includes/install.core.inc
```diff
@@ -535,16 +535,19 @@ function install_tasks($install_state) {
...
+    $exclusive_profile_count += !empty($profile_info['exclusive']);
   }
-  $display_select_profile = $hidden_profile_count > 1 && count($install_state['profiles']) != 1;
+  $display_select_profile = $hidden_profile_count > 1 && count($install_state['profiles']) != 1 && $exclusive_profile_count !=1;
```
ghost, line 547: Missing a space between the `=` and `1` :slightly_smiling_face:

ghost (changes requested):
Wow, guess who started a review but didn't actually post it... :roll_eyes:

core/includes/install.core.inc
```diff
@@ -535,16 +535,19 @@ function install_tasks($install_state) {
   // Determine whether translation import tasks will need to be performed.
   $needs_translations = count($install_state['translations']) > 1 && !empty($install_state['parameters']['langcode']) && $install_state['parameters']['langcode'] != 'en';
 
-  // Display profile task only when there is more than one non-hidden profile.
+  // Display the "Choose profile" task only when there is more than one
+  // non-hidden profiles; hide it if there is only a single exclusive profile.
```
ghost, line 539: There shouldn't be an 's' on the end of 'profiles' here (see original comment you're replacing).

ghost (changes requested):

PR #2789, opened 2019-07-10
Issue #3924: Installer: Do not assume only modules for dependencies.

klonos:
https://github.com/backdrop/backdrop-issues/issues/3924

### DIFF
--- a/core/includes/install.inc
+++ b/core/includes/install.inc
@@ -904,7 +904,7 @@ make this configuration active, see admin/config/development/configuration/sync.
  *   An array of information about the current installation state.
  *
  * @return
- *   The list of modules to install.
+ *   The list of projects to install/enable.
  *
  * @throws Exception
  */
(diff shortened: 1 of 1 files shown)


core/includes/install.inc
```diff
@@ -920,31 +920,37 @@ function backdrop_verify_profile($install_state) {
...
   $requirements = array();
 
-  if (count($missing_modules)) {
-    $modules = array();
-    foreach ($missing_modules as $module) {
-      $modules[] = '<span class="admin-missing">' . backdrop_ucfirst($module) . '</span>';
+  if (count($missing_projects)) {
+    $projects = array();
+    foreach ($missing_projects as $project) {
+      $projects[] = '<span class="admin-missing">' . $project . '</span>';
```
klonos: We should not be capitalizing things here, as missing projects means no .info file, which means no pretty name - only machine name.

core/includes/install.inc
```diff
@@ -920,31 +920,37 @@ function backdrop_verify_profile($install_state) {
...
-  $missing_modules = array_diff($info['dependencies'], $present_modules);
+  // Verify that all of the profile's required projects are present.
+  $missing_projects = array_diff($info['dependencies'], $present_projects);
 
   $requirements = array();
 
-  if (count($missing_modules)) {
-    $modules = array();
-    foreach ($missing_modules as $module) {
-      $modules[] = '<span class="admin-missing">' . backdrop_ucfirst($module) . '</span>';
```
klonos, line 941: We should not be capitalizing things here, as missing projects means no .info file, which means no pretty name - only machine name.

klonos:
Been some time. Refreshing the PR sandbox, and running tests again...

PR #2790, opened 2019-07-11
Issue #3876: Coding standards: changed 'else if' to 'elseif'

### LINKED ISSUE #3876: Coding standards: use `elseif` instead of `else if`
I have done a grep and found ~many~ a few instances of `else if (` in our codebase. See https://www.drupal.org/docs/develop/standards/coding-standards#controlstruct and https://api.backdropcms.org/php-standards#control-structures

> (Note: Don't use `else if` -- always use `elseif`.)

...we should make sure that we don't change any externally-maintained, 3rd-party files, nor .js files (https://api.backdropcms.org/js-standards#control-structures does not mention not using `else if`).

---
PR by @isabelhaas: backdrop/backdrop#2790

isabelhaas:
Fixes backdrop/backdrop-issues#3876

### DIFF
--- a/core/modules/layout/layout.module
+++ b/core/modules/layout/layout.module
@@ -554,7 +554,7 @@ function layout_theme() {
       if (is_array($style['hook theme'])) {
         $items += $style['hook theme'];
       }
-      else if (function_exists($style['hook theme'])) {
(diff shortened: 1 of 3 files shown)


klonos (approved):
👍

quicksketch:

PR #2792, opened 2019-07-16
avoid display recursion

robertgarrigos:
fixes this problem: https://github.com/backdrop-contrib/headless/issues/12
already being reported and fixed in drupal here: https://www.drupal.org/project/link/issues/2295071


klonos (approved):
LGTM 👍

PR #2794, opened 2019-07-20
issue #3941 install dbtasks now checks for utf8mb4 support

tabroughton:
This solves https://github.com/backdrop/backdrop-issues/issues/3941
I haven't tested with other databases other than mariadb but I can't see why it wouldn't work.  I wonder why it has been left out in the first place.

### DIFF
--- a/core/includes/install.inc
+++ b/core/includes/install.inc
@@ -305,6 +305,10 @@ abstract class DatabaseTasks {
       'arguments'   => array(),
     ),
     array(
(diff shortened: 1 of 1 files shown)


oadaeh:
The code changes like fine to me.
I think it might need tests for it, since the problem has been there for a lone time and no one knew about it.

klonos (approved):
LGTM 👍

quicksketch:

PR #2633, opened 2019-04-30
495 dashboard quicksketch

jenlampton:
Replaces https://github.com/backdrop/backdrop/pull/2632
Fixes https://github.com/backdrop/backdrop-issues/issues/495

PR #2796, opened 2019-07-22
Issue  #2633: [UX] Add a file upload UI

klonos:
https://github.com/backdrop/backdrop-issues/issues/2633

Continues work started by @jenlampton in https://github.com/backdrop/backdrop/pull/1970


jenlampton:
This change is included in https://github.com/backdrop/backdrop-issues/issues/2632, I would prefer that we focus our efforts on one copy of the same change.

klonos:
...will shortly close this. Just curious if the sandbox will work this time.

PR #2800, opened 2019-07-26
3719/cache plugin data

quicksketch:
https://www.drupal.org/node/1376400

https://git.drupalcode.org/project/views/commit/580ff35b315b8cf0eed6e33ab97e985634f829b1

backdrop/backdrop-issues#3719

Replaces https://github.com/backdrop/backdrop/pull/2626

### DIFF
--- a/core/modules/views/includes/cache.inc
+++ b/core/modules/views/includes/cache.inc
@@ -128,23 +128,25 @@ function _views_fetch_plugin_data($type = NULL, $plugin = NULL, $reset = FALSE) 
   static $cache = NULL;
   if (!isset($cache) || $reset) {
     // Load necessary code once.
-    if (!isset($cache)) {
-      views_include('plugins');
-      views_include_handlers();
-    }
+    views_include_handlers();
+
     // Because plugin data contains translated strings, and as such can be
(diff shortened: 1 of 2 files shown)


core/modules/views/includes/cache.inc
```diff
@@ -127,8 +127,27 @@ function _views_data_process_entity_types(&$data) {
...
     views_include_handlers();
-    $cache = views_discover_plugins();
+
+    // Because plugin data contains translated strings, and as such can be
+    // expensive to build, the results are cached per language.
+    global $language;
+    $cache_key = 'views_plugin_data:' . $language->langcode;
+    if ($reset) {
+      $cache = NULL;
+    }
```
quicksketch, line 139: I changed this line from the port. Looking at what was committed to Drupal, I can't see how it would work properly because the static cache value is still used even if the `$reset` parameter is passed in. Perhaps our test coverage was simply different than Drupal's and caught this problem which may have existed in the original patch.

quicksketch:
Merged into 1.x and 1.13.x. Thanks @klonos for the original PR at https://github.com/backdrop/backdrop/pull/2626!

PR #2801, opened 2019-07-26
Revert "Cache views plugin data"

quicksketch:
Reverts backdrop/backdrop#2626


quicksketch:
I merged the wrong PR previously. Reverting so that https://github.com/backdrop/backdrop/pull/2800 can be merged instead.

PR #2802, opened 2019-07-31
Issue #2982: Added 'clearfix' class to node content div.

### LINKED ISSUE #2982: Should add a clearfix class to the node `content` DIV
By default, node content looks like this, based on `node.tpl.php` in `node.module` itself:
```
<main class="l-content col-md-9" role="main" aria-label="Main content">
  <article id="node-18" class="node clearfix" role="article">
      <footer> stuff </footer>
          <div class="content">
          <section class="comments">
```
Bartik overrides `node.tpl.php` to add clearfix, so that `<div class="content clearfix">` doesnt bleed into the comments.

Basis doesnt do this, so I just had a node with a small image field as the last item, and the comments DIV rides up into the node content.

I can fix with custom CSS, but Basis should fix that.

---
PR by @BWPanda: https://github.com/backdrop/backdrop/pull/2802

ghost:
Fixes https://github.com/backdrop/backdrop-issues/issues/2982

### DIFF
--- a/core/modules/node/templates/node.tpl.php
+++ b/core/modules/node/templates/node.tpl.php
@@ -99,7 +99,7 @@
     </footer>
   <?php endif; ?>
 
-  <div class="content"<?php print backdrop_attributes($content_attributes); ?>>
+  <div class="content clearfix"<?php print backdrop_attributes($content_attributes); ?>>
     <?php
       // We hide the links now so that we can render them later.
       hide($content['links']);


core/modules/node/templates/node.tpl.php
```diff
@@ -99,7 +99,7 @@
     </footer>
   <?php endif; ?>
 
-  <div class="content"<?php print backdrop_attributes($content_attributes); ?>>
+  <div class="content clearfix"<?php print backdrop_attributes($content_attributes); ?>>
```
klonos, line 102: ...wondering if there should be a space between the last double quote, after the classes declaration, and before the printing of `$content_attributes`?

klonos: ...and also whether the `content` and `clearfix` classes should be merged into any possible classes provided via `$content_attributes`.

quicksketch:

Yeah agreed on maybe rethinking the classes being hard-coded here instead of being part of the `$content_attributes`... but since we already have it separated, we'll roll with this fix first.

PR #2803, opened 2019-08-02
3956 stuck in maint mode

### LINKED ISSUE #3956: [UX] Updates: Maintenance mode enabled automatically, but never disabled
**Description of the bug**
When I choose to update my site using the self-update feature (but maybe this happens with regular updates too, I haven't checked) step 2 is "Put your site into maintenance mode." but when I run the update my site enters into maintenance mode even when I don't want it to.

We either need to remove step 2 from this list, or honor the setting during updates.

Either way we need to turn off the setting at the end of updates if we turned it on at the beginning. 

**Steps To Reproduce**
To reproduce the behavior:
1. Visit update.php
2. Update core without setting the site into maintenance mode
3. After the update, confirm that your site IS in maintenance mode.

**Actual behavior**
Site IS in maintenance mode, even though you didn't set it to be.

**Expected behavior**
Site should not be in maintenance mode.

---
PR by @jenlampton: https://github.com/backdrop/backdrop/pull/2803

jenlampton:
Fixes https://github.com/backdrop/backdrop-issues/issues/3956

HOW TO TEST:
* apply the patch before the core update (to get the right session var set)
* then apply the patch again after the core code is replaced, but before the updates have run (to get the un-setting of the session var right.)
* then run the database updates

### DIFF
--- a/core/modules/installer/installer.authorize.inc
+++ b/core/modules/installer/installer.authorize.inc
@@ -1,5 +1,4 @@
 <?php
-
 /**
  * @file
  * Callbacks and related functions invoked by authorize.php to update projects.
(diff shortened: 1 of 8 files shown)


core/modules/installer/installer.manager.inc
```diff
@@ -424,7 +431,8 @@ function installer_manager_update_ready_form($form, &$form_state) {
  */
 function installer_manager_update_ready_form_submit($form, &$form_state) {
   // Store maintenance_mode setting so we can restore it when done.
-  $_SESSION['maintenance_mode'] = config_get('system.core', 'maintenance_mode');
+  $_SESSION['maintenance_mode'] = state_get('system.core', 'maintenance_mode');
```
klonos: @jenlampton I think I have faced similar problems when working on https://github.com/backdrop/backdrop-issues/issues/2583 (sorry, it's been almost a year, so can't recall what exactly). Can you please try something like this:

```suggestion
  $previous_maintenance_mode_state = state_get('maintenance_mode', FALSE);
  $maintenance_mode = $form_state['values']['maintenance_mode'];
  state_set('maintenance_mode', (bool) $maintenance_mode);
  // Also store maintenance_mode setting so we can restore it when done.
  $_SESSION['previous_maintenance_mode'] = $previous_maintenance_mode_state;
```


Notice how the proper way to use `state_get()` is `state_get('maintenance_mode', FALSE);`.

Also noting that in one of my PRs I have this:

```php
  // In some cases, the site might seem as still online/offline, due to cache.
  cache_clear_all();
```

core/modules/installer/installer.manager.inc
```diff
@@ -364,6 +364,13 @@ function installer_manager_download_batch_finished($success, $results) {
     // but just in case, we have to tell them something.
     backdrop_set_message(t('Fatal error trying to download.'), 'error');
   }
+
+  // Now that the update is done, we can put the site back online if it was
+  // previously in maintenance mode.
+  if (isset($_SESSION['maintenance_mode']) && $_SESSION['maintenance_mode'] == FALSE) {
```
klonos: Can we please change this to `$_SESSION['previous_maintenance_mode']`? (see my next comment)

core/modules/installer/installer.authorize.inc
```diff
@@ -201,37 +201,26 @@ function installer_authorize_update_batch_finished($success, $results) {
...
-      );
-    }
-    else {
-      $page_message = array(
-        'message' => t('Update was completed successfully.'),
-        'type' => 'status',
-      );
-    }
+    $page_message = array(
+      'message' => t('Code replacement completed successfully.'),
```
klonos: Can we please use something like `Files updated successfully.` here (same for the rest of the messages following)? The average Joe/Joette would not know the difference here, and "files" makes more sense than "code".

jenlampton:
Files has a diffferent meaning: the files directory. What if we were more
specific? Something like "X module" or "Backdrop core"

core/modules/installer/installer.authorize.inc
jenlampton: The word `Files` has a different meaning than what we are talking about here. The `files` directory is sort-of the opposite of what we are talking about here, which is code.  Any other recommendations?

jenlampton: What about `Code replaced successfully` or `Code updated successfully` ?

core/modules/installer/installer.manager.inc
jenlampton: This is used all over the place, I don't want to break something because we wanted to change an array key.

core/modules/installer/installer.authorize.inc
jenlampton: I've tried `Code updated...`

jenlampton:
closing for new sandbox.

jenlampton:
new sandbox please :)

core/modules/installer/installer.manager.inc
jenlampton: outdated... but I added a cache_clear_all() just in case.

PR #2806, opened 2019-08-05
Issue #512: [UX] update.php: Revamp the "Overview" step

### LINKED ISSUE #512: [UX] Streamline the messages and workflow for the "Backdrop site updates" form
We should also run a check to see if any updates are pending to begin with (edit: this is being handled separately in #3245). If no updates, then communicate that to the user, instead of asking them to back things up, set the site to maintenance mode, and continue.

Nitpicking, I know, but these are the suggested steps as it is now:

> 1. Create backups. This update utility will alter your database and config files. In case of an emergency you may need to revert to a recent backup; make sure you have one.
>     - **Database**: Create a database dump of the 'my_backdrop_db' database.
>     - **Config files**: Back up the entire directory at './files/config_random_hash_here/active'.
> 2. Put your site into [maintenance mode](url).
> 3. Install your new files into the appropriate location, as described in [the handbook](https://backdropcms.org/upgrade).

Related:

- As per #3960, we should provide a checkbox instead of point 2; to allow setting the site to maintenance mode without having to leave this page.
- As per #3959, we should remove point 3, as it has no value.

## Status (Aug 2019)

The user gets the following, whether there are pending updates or not:

Screen Shot 2019-08-04 at 7 19 24 am

## Proposed change

If there are no pending updates:

Screen Shot 2019-08-05 at 6 59 05 pm

If there are updates pending:

Screen Shot 2019-08-05 at 4 59 18 am

If there are pending updates, but the site has been set to maintenance mode before running update.php:

Screen Shot 2019-08-05 at 7 11 22 pm

---
PR by @klonos: https://github.com/backdrop/backdrop/pull/2806

klonos:
Fixes https://github.com/backdrop/backdrop-issues/issues/512


klonos:
...resetting the PR sandbox.

core/update.php
```diff
@@ -283,61 +285,120 @@ function update_results_page() {
...
+  if ($maintenance_mode) {
+    backdrop_set_message(t('The site is already in maintenance mode.'));
+  }
+
+  $form['maintenance_mode'] = array(
+    '#type' => 'checkbox',
+    '#id' => 'maintenance_mode',
+    '#title' => 'Perform updates with site in maintenance mode (<strong>strongly recommended</strong>)',
+    '#default_value' => TRUE,
+    '#disabled' => $maintenance_mode,
```
avpaderno, line 361: Why would users be allowed to decide if putting the site in maintenance mode when it is not already in maintenance mode, but they cannot decide about that when the site is already in maintenance mode?  
The checkbox should be always enabled.

core/update.php
```diff
@@ -283,61 +285,120 @@ function update_results_page() {
...
+ */
+function update_script_overview_form_submit($form, &$form_state) {
+  global $base_url;
+
+  // Store maintenance_mode setting so we can restore it when done.
+  $_SESSION['maintenance_mode'] = config_get('system.core', 'maintenance_mode');
+  if ($form_state['values']['maintenance_mode'] == TRUE) {
+    state_set('maintenance_mode', TRUE);
+  }
+
```
avpaderno, line 395: The code should set *maintenance_mode* to the value selected from users. If they do not want the site to be in maintenance mode, the site should not be in maintenance mode.

core/update.php
```diff
@@ -347,18 +408,18 @@ function update_info_page() {
...
   backdrop_add_http_header('Status', '403 Forbidden');
   watchdog('access denied', 'update.php', NULL, WATCHDOG_WARNING);
-  backdrop_set_title(t('Access denied'));
+  backdrop_set_title('Access denied');
 
   $output = '';
   $steps = array();
 
-  $output .= t('You are not authorized to access this page. Log in using either an account with the <em>administer software updates</em> permission, or the site maintenance account (the account you created during installation). If you cannot log in, you will have to edit <code>settings.php</code> to bypass this access check. To do this:');
+  $output .= 'You are not authorized to access this page. Log in using either an account with the <em>administer software updates</em> permission, or the site primary account (the account you created during installation). If you cannot log in, you will have to edit <code>settings.php</code> to bypass this access check. To do that:';
```
avpaderno, line 416: To reference something just described or mentioned, *this* is correct. *this access check* does not seem correct, as *Log in* is at the beginning of the previous sentence and that is not an *access check*.

klonos:
@kiamlaluno thank you for the review. I'm closing this PR, since there are to many changes in the meantime, so fixing conflicts is not practical. I'll file a new PR, taking your review/suggestions into account.

PR #2809, opened 2019-08-07
SA-CORE-2019-011-013

quicksketch:
Fixes for today's security release.


quicksketch:
Looks like we're getting the random failure from https://github.com/backdrop/backdrop-issues/issues/3794, which doesn't seem to be related. This seems good to go.

PR #2814, opened 2019-08-10
Issue #3975: [UX] Installer: Revamp the language selection page

klonos:
https://github.com/backdrop/backdrop-issues/issues/3975

### DIFF
--- a/core/includes/install.core.inc
+++ b/core/includes/install.core.inc
@@ -1407,32 +1407,30 @@ function install_select_language_form($form, &$form_state, $files) {
     '#default_value' => !empty($browser_langcode) ? $browser_langcode : 'en',
   );
 
-  if (count($files) == 1) {
-    $directory = settings_get('locale_translate_file_directory', conf_path() . '/files/translations');
+  $directory = settings_get('locale_translate_file_directory', conf_path() . '/files/translations');
 
-    $steps = array();
-    $steps[] = st('Download translation files from the <a href="https://localize.backdropcms.org/translate/languages" target="_blank">translation server</a>.');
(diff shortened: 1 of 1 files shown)


klonos:
With the entire `$output` moved into the form as help `$markup`, and without the need for a `'translate'` URL parameter (previously required to redirect to the dedicated page with the instructions for the installer translations), the logic within the `if (!empty($install_state['parameters']['translate']))` conditional was not required, which left the remaining `else` identical to the  `if ($install_state['interactive'])` conditional, which in turn lead to simplification of the whole thing to:

- interactive installation:
  - always show the language selection menu + translation help
- non interactive installation:
  - if one language (English) -> proceed with installation
  - if multiple languages -> throw error about requiring a language to be specified

I'm really happy with this code reduction and the simplification of the logic ...that is on top of the UX improvements of keeping the translation instructions in the collapsed fieldset instead of a separate page.

olafgrabienski:
I'm not sure which URL to the translation server we should use. Currently (and in the PR) we link to https://localize.backdropcms.org/translate/languages.

Possible alternatives:

- https://localize.backdropcms.org (pro: URL won't change; con: links in language list point to the "Translate" page of the single languages, so language files are hard to find)
- https://localize.backdropcms.org/translate (more blocks in the sidebar than the current linked page)
- https://localize.backdropcms.org/translate/downloads (shows the language files directly, but of all languages, looks quite crowded)

klonos:
...replacing with #2930. @olafgrabienski can you please move your comment from above to the issue queue in https://github.com/backdrop/backdrop-issues/issues/3975 so we can get more feedback re the URL to use? Thanks.

PR #2815, opened 2019-08-10
Issue #3974 More flexible regex for detecting .po files

klonos:
https://github.com/backdrop/backdrop-issues/issues/3974


quicksketch:

PR #2819, opened 2019-08-14
#3972: Do not translate am/pm to prevent validation errors.

indigoxela:
See https://github.com/backdrop/backdrop-issues/issues/3972 for details.

Note: skipping translation is probably not the right approach.


indigoxela:
Moving this out of the way, as skipping translation is the wrong approach anyway.

PR #2821, opened 2019-08-15
Issue #1575: Allow disabled text formats to be enabled again.

### LINKED ISSUE #1575: [UX] Text formats: provide a way to re-enable disabled formats via the admin UI
A 'Disabled' text format is different than every other 'Disabled' thing in the admin Ui. A text format, once disabled, can not be enabled again. Everything else can. This can be a pain-point for admins who think they know what to expect.

Possible solutions that have been considered:

1) Allow admins to re-enable disabled text formats, making the operation reversible.
2) Change the name from 'Disabled' to something more like 'Hidden'
3) Allow admins to completely 'Delete' a text format (possibly dangerous!)

The final approach taken in this issue was to keep disabled formats in the text formats list in `/admin/config/content/formats`, denote them as disabled, and allow for them to be re-enabled (via a newly-introduced "Enable" operation). See https://github.com/backdrop/backdrop-issues/issues/1575#issuecomment-527534721 for a screenshot of this implementation.

This fixes the actual UX issue here, which is that previously we were telling people that we are going to disable the text format, but what we were actually doing instead was to completely remove it from the UI, leaving the user wondering how to recover from this situation. Advanced users knew how to revert this, by manually editing the config files; the new implementation makes it so that even novice users can accomplish the same thing via the admin UI.

The actual "Remove" functionality, and whether we should implement that as complete removal of the text format config files, or simply hide them from the admin UI and keep the config files in place, was left as a follow-up task in #4013.

---

## Original Issue

Most site admins think that something "disabled" can be "enabled" again.
Ok, there is a confirm screen, but not everybody read them, especially if they are written in a foreign language (and drupal-7.41.##.po does not include text of `admin/config/content/formats/{format}/disable`)...

----

Advocate:  @klonos

----

Weekly update?

----

~~PR https://github.com/backdrop/backdrop/pull/2821~~ by @BWPanda
~~PR https://github.com/backdrop/backdrop/pull/2826~~ by @klonos (based on @BWPanda's original PR): 
~~PR https://github.com/backdrop/backdrop/pull/2845~~ by @quicksketch 
~~PR https://github.com/backdrop/backdrop/pull/2846 by @jenlampton~~
~~PR  https://github.com/backdrop/backdrop/pull/3024~~ by @klonos
PR https://github.com/backdrop/backdrop/pull/3071 by @docwilmot

ghost:
Fixes https://github.com/backdrop/backdrop-issues/issues/1575

### DIFF
--- a/core/modules/filter/filter.module
+++ b/core/modules/filter/filter.module
@@ -128,6 +128,14 @@ function filter_menu() {
     'access arguments' => array('administer filters'),
     'file' => 'filter.admin.inc',
   );
+  $items['admin/config/content/formats/%filter_format/enable'] = array(
+    'title' => 'Enable text format',
+    'page callback' => 'filter_enable',
+    'page arguments' => array(4),
+    'access arguments' => array('administer filters'),
+    'type' => MENU_CALLBACK,
+    'file' => 'filter.pages.inc',
+  );
   $items['admin/config/content/formats/%filter_format/disable'] = array(
     'title' => 'Disable text format',
     'page callback' => 'backdrop_get_form',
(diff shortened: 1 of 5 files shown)


core/modules/filter/filter.module
```diff
@@ -725,8 +739,7 @@ function filter_formats($account = NULL) {
         );
         $filter_format = (object) $filter_format;
 
-        // We don't include disabled formats.
-        if (!$filter_format->status) {
+        if (!$include_disabled && !$filter_format->status) {
```
klonos, line 742: Shouldn't this be an OR instead?

ghost: Nope, that'd mean all formats get ignored when `$include_disabled` is FALSE...

herbdool:
Replaced by https://github.com/backdrop/backdrop/pull/2846

PR #2822, opened 2019-08-15
Issue #3748: Replace hover background color on account links with hov…

### LINKED ISSUE #3748: [UX] Remove the hover color from header account links in Basis
While testing https://github.com/backdrop/backdrop-issues/issues/3209 I found it distracting that the account links in the Header area had a hover color. I would prefer that they have an underline while hovered.

I was distracted by the big color block, especially when it does not reach the top of the menu area. I don't think we should treat them the same way as the main menu.

---
PR by @jenlampton: https://github.com/backdrop/backdrop/pull/2822

jenlampton:
Fixes https://github.com/backdrop/backdrop-issues/issues/3748

### DIFF
--- a/core/themes/basis/css/skin.css
+++ b/core/themes/basis/css/skin.css
@@ -170,6 +170,7 @@ fieldset .fieldset-legend {
 .header-menu a:hover,
 .l-header a:active {
   text-decoration: underline;
+  background: unset;
 }
 
 .block-system-header {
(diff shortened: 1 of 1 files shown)


core/themes/basis/css/skin.css
```diff
@@ -169,7 +169,7 @@ fieldset .fieldset-legend {
 
 .header-menu a:hover,
 .l-header a:active {
-  background: #114a75;
+  text-decoration: underline;
```
klonos, line 172: The `:active` color falls back to `#fffffe` (either click and hold, or use the inspector 	to activate):

Screen Shot 2019-08-15 at 10 40 12 pm

```suggestion
  text-decoration: underline;
  background: unset;
```

jenlampton:
nice catch :)

PR #2826, opened 2019-08-24
Issue #1575: [UX] Allow disabling/enabling text formats via the admin UI

klonos:
https://github.com/backdrop/backdrop-issues/issues/1575

### DIFF
--- a/core/modules/filter/filter.admin.inc
+++ b/core/modules/filter/filter.admin.inc
@@ -18,6 +18,12 @@ function filter_admin_overview($form) {
   $fallback_format = filter_fallback_format();
   $editors = filter_get_editors();
 
+  $form['help'] = array(
+    '#type' => 'help',
+    '#markup' => t('The order of the text formats here reflects the order in which they will show up in the <em>Editor</em> selector, when creating content (still subject to which text formats the user has access to).'),
+    '#weight' => -100,
+  );
+
   $form['#tree'] = TRUE;
   foreach ($formats as $id => $format) {
     $links = array();
--- a/core/modules/filter/filter.admin.inc
+++ b/core/modules/filter/filter.admin.inc
@@ -20,7 +20,7 @@ function filter_admin_overview($form) {
 
   $form['help'] = array(
     '#type' => 'help',
-    '#markup' => t('The order of the text formats here reflects the order in which they will show up in the <em>Editor</em> selector, when creating content (still subject to which text formats the user has access to).'),
+    '#markup' => t('The order of the text formats here reflects the order in which they will show up in the <em>Editor</em> selector when creating content (subject to the text formats to which a user has access.'),
     '#weight' => -100,
   );
 
--- a/core/modules/filter/filter.admin.inc
+++ b/core/modules/filter/filter.admin.inc
@@ -25,46 +25,39 @@ function filter_admin_overview($form) {
   );
 
   $form['#tree'] = TRUE;
+
   foreach ($formats as $id => $format) {
-    $links = array();
-    if (!$format->status) {
-      $links['enable'] = array(
-        'title' => t('Enable'),
-        'href' => "admin/config/content/formats/$id/enable",
-      );
+    // Prepare roles for each fomrat.
+    $roles = array();
+    foreach (filter_get_roles_by_format($format) as $role_name) {
+      $role = user_role_load($role_name);
+      $roles[] = check_plain($role->label);
     }
+    $roles_markup = $roles ? theme('item_list', array('items' => $roles)) : t('No roles specified for this format.');
+
+    // Prepare operations links for each format.
+    $links = array();
+
     $links['configure'] = array(
       'title' => t('Configure'),
       'href' => "admin/config/content/formats/$id",
     );
-    // Check whether this is the fallback text format. This format is available
-    // to all roles and cannot be disabled via the admin interface.
-    $form['formats'][$id]['#is_fallback'] = ($id == $fallback_format);
-    if ($form['formats'][$id]['#is_fallback']) {
-      $form['formats'][$id]['name'] = array('#markup' => $format->name);
-      $roles_markup = t('Fallback format.<br>Used if a role has access to no other formats.');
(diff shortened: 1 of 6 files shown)


klonos:
...php70 test acting up.

klonos:
CI acting up again 😞

core/modules/filter/filter.admin.inc
```diff
@@ -14,17 +14,50 @@
...
 function filter_admin_overview($form) {
   // Overview of all formats.
-  $formats = filter_formats();
+  $formats = filter_formats(NULL, TRUE, TRUE);
   $fallback_format = filter_fallback_format();
   $editors = filter_get_editors();
 
+  $form['help'] = array(
+    '#type' => 'help',
+    '#markup' => t('The order of the text formats here reflects the order in which they will show up in the <em>Editor</em> selector, when creating content (still subject to which text formats the user has access to).'),
```
quicksketch:
```suggestion
    '#markup' => t('The order of the text formats here reflects the order in which they will show up in the <em>Editor</em> selector when creating content (subject to the text formats to which a user has access.'),
```

core/modules/filter/filter.admin.inc
```diff
@@ -474,6 +503,20 @@ function filter_admin_format_form_submit($form, &$form_state) {
   $form_state['redirect'] = 'admin/config/content/formats';
 }
 
+/**
+ * Page callback: Enable a text format.
+ */
+function filter_admin_enable($format = NULL) {
+  if (isset($format)) {
```
quicksketch, line 517: We need to add token validation on these links to prevent CSRF attacks. We just did this over in Flexible Layouts in this commit: https://github.com/backdrop/backdrop/pull/2646/commits/45ab3a029a9f593fdeea024ba2a1df420a7d0d43#diff-1b45663e4cee17ec13f8e83d295a118cR740

quicksketch: Looks good!

core/modules/filter/filter.module
```diff
@@ -692,20 +697,26 @@ function filter_modules_disabled($modules) {
...
+ *   Defaults to FALSE.
  *
  * @return stdClass[]
  *   An array of text format objects, keyed by the format ID and ordered by
  *   weight.
  *
  * @see filter_formats_reset()
  */
-function filter_formats($account = NULL) {
+function filter_formats($account = NULL, $include_disabled = FALSE, $reset = FALSE) {
```
quicksketch, line 713: I think you can eliminate the `$reset` parameter, as it seems `filter_formats_reset()` provides the same functionality.

klonos: You mean that we should be specifically calling `filter_formats_reset()` before `filter_formats()` if the intention was to bypass cache?

docwilmot: also note if `filter_formats(NULL,, TRUE)` is called, any subsequent calls on the same page load will ignore `$include_disabled` because thats checked inside the cache loop.

docwilmot:
To follow up my comment above, mostly because I found this an interesting situation I hadnt seen happen before, I'm posting a longer explanation here. Hope you dont mind me banging on about this @klonos :

```
function filter_formats($account = NULL, $include_disabled = FALSE, $reset = FALSE) {
  global $language;
  $formats = &backdrop_static(__FUNCTION__, array()); <-- this means the `$formats` variable persists if this function is called multiple times for the same page.

  // All available formats are cached for performance.
  if (!isset($formats['all']) || $reset) {    <-- this `$formats['all']` is set the FIRST time you call the function, so if you call twice on same page load, its already set after the first call, and on the second call this entire code block is not executed.
    ...
        if (!$include_disabled && !$filter_format->status) {  <-- which means we never reach here on the second call
          continue;
        }
    ...
  }

  // Build a list of user-specific formats.
  if (isset($account) && !isset($formats['user'][$account->uid])) {
    $formats['user'][$account->uid] = array();
    foreach ($formats['all'] as $format) {
      // Include only enabled formats, to which the user has access to.
	  
	  // if (!$include_disabled && !$filter_format->status) ... <-- so you'd need to do the check here, and use `unset()` to remove disabled ones


      if ($format->status && filter_access($format, $account)) {
        $formats['user'][$account->uid][$format->format] = $format;
      }
    }
  }

  return isset($account) ? $formats['user'][$account->uid] : $formats['all'];
}
```

If this is called twice on the same page load and both calls are `filter_formats()` or both calls are `filter_formats(NULL, TRUE)`, no problem, you'd get the same return each time, but if the first call is to return everything including disabled, ie `filter_formats(NULL, TRUE)`, and the second call was to exclude disabled (ie `filter_formats()` ) then the second call would not be what you wanted, because `$formats['all']` would be populated with everything already from the first call, and would skip the check and the `continue()`.

core/modules/filter/filter.admin.inc
```diff
@@ -14,17 +14,53 @@
...
+    else {
+      $links['enable'] = array(
+        'title' => t('Enable'),
+        'href' => "admin/config/content/formats/$id/enable",
+        'query' => array(
+          'token' => backdrop_get_token('format-' . $format->format)
+        ),
+      );
+      // Move "Enable" to be the first/default operation (for disabled formats).
+      $links = array_merge(array('enable' => $links['enable']), $links);
```
jenlampton, line 61: This can be done by adding 'weight' instead.

core/modules/filter/filter.admin.inc
```diff
@@ -14,17 +14,53 @@
...
+
     $links['configure'] = array(
       'title' => t('Configure'),
       'href' => "admin/config/content/formats/$id",
     );
+
+    if ($format->status) {
+      $links['disable'] = array(
+        'title' => t('Disable'),
+        'href' => "admin/config/content/formats/$id/disable",
```
jenlampton, line 49: what you have here is very slightly slower to process than `'admin/config/content/formats/' . $id . '/disable'`. Let's be as consistent as we can in core, and choose the faster way :)

jenlampton:
Replaced by https://github.com/backdrop/backdrop/pull/2846

PR #2827, opened 2019-08-24
Issue #3936: Fix group operators in views "rearrange filter criteria" dialog

### LINKED ISSUE #3936: Views filters: Group `OR` operator shown as `AND` in the "Rearrange filter criteria" dialog
Filters overview 👍:

Screen Shot 2019-07-17 at 11 57 14 am

"Rearrange filter criteria" dialog 👎:

Screen Shot 2019-07-17 at 11 57 03 am

---
PR by @isabelhaas: backdrop/backdrop#2827

isabelhaas:
Fixes backdrop/backdrop-issues#3936

It seemed like the reason both operators were showing up as the value of the first operator was because in views-admin.js, when getting the operator value, `find()` returns the elements in the order they are on the page, so the value found was always the value of the first operator, not the one relating to the current group. I added `.last()` after `find(‘option:selected’)` to access the operator value that is associated with the current row’s group.

### DIFF
--- a/core/modules/views_ui/js/views-admin.js
+++ b/core/modules/views_ui/js/views-admin.js
@@ -724,7 +724,7 @@ $.extend(Backdrop.viewsUi.RearrangeFilterHandler.prototype, {
         // first operator dropdown we encounter, going backwards from the current
         // row. This dropdown is the one associated with the current row's filter
         // group.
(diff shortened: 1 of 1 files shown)


ghost:

PR #2828, opened 2019-08-24
Issue #3979: Selfupdate fails

indigoxela:
Adds a test and fixes failed check for core directory in core/includes/updater.inc resulting in an exception when running self-update.

See https://github.com/backdrop/backdrop-issues/issues/3979 for details.


core/includes/updater.inc
```diff
@@ -186,8 +186,11 @@ abstract class Updater implements BackdropUpdaterInterface {
    * @throws UpdaterException
    */
   public static function getProjectType($directory) {
-    if ($directory == BACKDROP_ROOT) {
-      return 'core';
+    if (basename($directory) == 'backdrop') {
```
serundeputy, line 189: I think this will only work if the directory backdrop is installed in is named `backdrop` if for example you install backdrop in a directory named `www` this will fail.

indigoxela: Nope, $directory is the stream wrapper to the extracted zip file (temporary://) and NOT the install dir. ;)

core/includes/updater.inc
```diff
@@ -175,18 +175,19 @@ abstract class Updater implements BackdropUpdaterInterface {
...
-   * Return the project type from a Backdrop info file.
+   * Return the project type from a Backdrop info file or directory.
    *
    * @param string $directory
-   *   Directory to search for the info file.
+   *   Stream wrapper to directory to get the type from.
    *
    * @return string
-   *   The type of the project, either "module", "theme", or "layout".
+   *   The type of the project, either "core", "module", "theme", or "layout".
```
klonos, line 184: Good catch 👍

core/includes/updater.inc
```diff
@@ -175,19 +175,23 @@ abstract class Updater implements BackdropUpdaterInterface {
...
    *
    * @throws UpdaterException
    */
   public static function getProjectType($directory) {
-    if ($directory == BACKDROP_ROOT) {
-      return 'core';
+    $basename = basename($directory);
+    if (substr($basename, 0, 8) == 'backdrop') {
```
klonos: ...this may match things like `backdrop_cool_module` or `backdrop_neat_theme`. I think that it would be better to use `preg_match`:

```suggestion
    if preg_match('/^backdrop(-[\d\.]+)?/', $basename) {
```

Note: untested ^^

herbdool: @klonos might be right if `$directory` is being checked for any project. Though perhaps instead of crafting the perfect regex for a use case that @quicksketch says will *never* happen, we should go back to just checking for only "backdrop".

> However, since the self-updater only ever uses packages provided by backdropcms.org, the self-updater would never use one of these "Source code" archives anyway. And if we get something unexpected, just failing the entire process is safer than attempting to proceed.

We check for "backdrop" > if that succeeds, we then check for "core" directory. If that succeeds, great. If that fails, shouldn't we be throwing an error? Currently the function does nothing.

klonos: Agreed 👍

klonos: Good catch @herbdool 👍 ...were you thinking a `continue` perhaps?

indigoxela: We found a consensus, that it's either exactly "backdrop", or it can't be backdrop core.
Regarding @herbdool's remark, that if the check for "core" fails, nothing happens. That's not the case. An UpdaterExeption is thrown by getUpdaterFromDirectory(), as getProjectType() returns nothing then.

core/includes/updater.inc
```diff
@@ -175,19 +175,26 @@ abstract class Updater implements BackdropUpdaterInterface {
...
-    if ($directory == BACKDROP_ROOT) {
-      return 'core';
+    if (basename($directory) == 'backdrop') {
+      $realpath = backdrop_realpath($directory);
+      if (is_dir($realpath . '/core')) {
+        return 'core';
+      }
+      else {
+        $message = 'Updating core from @directory was attempted, but either it doesn\'t contain a "core" directory, or it isn\'t accessible. You may need to update manually.';
+        throw new UpdaterException(t($message, array('@directory' => $directory)));
```
klonos: This may be a technicality, but we usually would do `$message = t('...', array('@directory' => $directory));` and then:

```suggestion
        throw new UpdaterException($message);
```

...point being that the actual message and the array that holds placeholders should stick together.

indigoxela: Agreed, keeping them together makes sense.

klonos (approved):
LGTM 👍

quicksketch:

PR #2709, opened 2019-05-30
Issue #2632: Core fieldable file entity bundles.

jenlampton:
Fixes https://github.com/backdrop/backdrop-issues/issues/2632

Replaces https://github.com/backdrop/backdrop/pull/2238

PR #2832, opened 2019-08-28
File 2632

### LINKED ISSUE #2632: Core fieldable file entity bundles
Were working on adding all of the file_entity module to Backdrop core, including fieldable files. 

### Before PR merge:

- [x] Convert file type info to config (move from DB table)
- [x] Remove the default alt and title fields provided out of box on the Image file type.
- [x] Make [UI changes](https://github.com/backdrop/backdrop-issues/issues/2632#issuecomment-526890267) to simplify the "Manage File Display" modes form.
- [x] Add the upgrade path from Backdrop 1.x
- [x] Test the upgrade path from Backdrop 1.x
- [x] Add an upgrade path from Drupal 7 for D7 file types and file_display settings.
- [x] Test the upgrade path from Drupal 7

### Before release (if possible):
- [ ] Move "Manage File Display" settings to the "Manage displays" tab (as settings for the "File" psuedofield.
- [x] Add a wrapper function for file_entity_access with watchdog_deprecated_function()
- [ ] Integrate with Path module (add to vertical tabs + Provide default path aliases for files)

### Optional UX improvements:
- [x] Remove the confirm form when disabling / enabling file types
- [x] Make the disabled types gray, and below the enabled ones.
- [ ] File type config form: Mimetypes should be a checklist or something helpfully clickable like tokens.
- [ ] Add the "Add file" link into admin bar?

### At a later time
* [Discuss adding back Alt and Title fields](https://github.com/backdrop/backdrop-issues/issues/4007)
* Add back-end breadcrumbs for front-end pages `/file/%/manage` and `/file/%/usage`, to allow easy navigation back to `/admin/content/files`
* add something like inline entity form so that we can get additional file fields to show up in other places like node forms and rich-text editor forms.
* there is no obvious way to add a video to a node, now that we allow uploading videos. Should there be an issue for that?

---
#3971 #Original Issue

We do have this task in File management meta issue #1448, but I didn't find separate issue for it.

Looks like we can make file entities fieldable with just one line of code:
`$entity_info['file']['fieldable'] = TRUE;`

So I merged "file types" and "fieldable file types" in one issue. Feel free to separate them if I'm wrong.
 
---
#Advocates: @docwilmot @jenlampton 

----
~PR: https://github.com/backdrop/backdrop/pull/1924~ by @docwilmot
~PR: https://github.com/backdrop/backdrop/pull/2238~ by @jenlampton
~PR: https://github.com/backdrop/backdrop/pull/2832~ by @docwilmot
PR: https://github.com/backdrop/backdrop/pull/2838 by @jenlampton

docwilmot:
Fixes https://github.com/backdrop/backdrop-issues/issues/2632

Might replace https://github.com/backdrop/backdrop/pull/2709
PR to test in sandbox for now.


docwilmot:
Ah, just remembered why I'd renamed those fields to `field_fiat` (and `field_fitt`): because simpletest creates tables called `simpletest_cache_standard_field_revision_field_file_image_alt_text` and that is too long and crashes the sandbox!

> Names for databases, tables, columns, and indexes can be up to 64 characters long.

But `simpletest_cache_standard_field_revision_field_file_image_alt_text` is 66.

jenlampton:
> Ah, just remembered why I'd renamed those fields to field_fiat (and field_fitt): because simpletest creates tables called simpletest_cache_standard_field_revision_field_file_image_alt_text and that is too long and crashes the sandbox!

Hm, well that's a problem! Unfortunately, I don't think we should sacrifice code quality because of limitations on our sandbox. @Gormartsen is there any chance we can adjust the sandbox so these table names don't crash it?

docwilmot:
> because of limitations on our sandbox

Its not our sandbox. Its a MySQL limitation apparently.

jenlampton:
Hm. Maybe our tests need to be reworked then?  @quicksketch have you run
into this before? I'm sure this isn't the first time this has happened. :)

docwilmot:
> Hm. Maybe our tests need to be reworked then? 

Might be worth considering shortening the simpletest prefixes. For a field table name to not crash, the field name has to be less than 17 characters, which seems too limiting to me. 

This because the first bit of the resulting table name `simpletest_cache_standard_field_revision_field_`, (47 characters) will be created for every field during a test run, which leaves only 17 more characters before you hit the MySQL length limit if I read this correctly.

jenlampton:
> For a field table name to not crash, the field name has to be less than 17 characters, which seems too limiting to me.

I agree. 

> This because the first bit of the resulting table name simpletest_cache_standard_field_revision_field_, (47 characters) will be created for every field during a test run

😮 that seems crazy, no?

jenlampton:
@quicksketch says there is a mysql setting that allows for long tablenames. We should have @Gormartsen enable this on the CI system -- but I still think we should fix the too-long table names in our tests :)

klonos:
Interesting discussion! ...I have captured all that in https://github.com/backdrop/backdrop-issues/issues/4004 😉

quicksketch:
Replaced by https://github.com/backdrop/backdrop/pull/2838.

PR #2382, opened 2018-11-13
Issue #2568: Add support for "trusted_host_patterns".

klonos:
https://github.com/backdrop/backdrop-issues/issues/2568

PR #2835, opened 2019-08-31
Issue #2568: Add support for "trusted_host_patterns".

### LINKED ISSUE #2568: [D8][SR] Add support for "trusted_host_patterns" (but do not enforce).
https://www.drupal.org/node/1992030

> ## Trusted host security setting in Drupal 8
> 
> As of January 2015, [Drupal 8 supports "trusted host patterns"](https://www.drupal.org/node/2410395), where you can (and should) specify a set of regular expressions that the domains on incoming requests *must* match. Example configuration in settings.php would read:
> 
> ```php
> $settings['trusted_host_patterns'] = array(
>   '^www\.example\.com$',
> );
> ```
> 
> See the above change record for more details. Note that, if you're doing local development, you might get (temporarily) locked out of your site by the above configuration on its own. [You should add a trusted host pattern](http://drupal.stackexchange.com/a/145994/) for `'^localhost$'` in this case.

https://www.drupal.org/node/2410395

> ...
> **Description**: 
> It is possible to spoof the HTTP Host header for nefarious purposes, and trick Drupal into using a different domain name in several subsystems (particularly link generation). The Symfony framework provides a trusted host mechanism, where site administrators can whitelist hostnames. This mechanism can now be configured through settings.php.
>...

d8-status_report-trusted_host_patterns

---
Advocates: @klonos 

---

~~PR by @klonos: https://github.com/backdrop/backdrop/pull/2382~~
updated PR by @quicksketch: https://github.com/backdrop/backdrop/pull/2835

---

NOTE: The current PR relies on things being defined in settings.php, which we cannot edit in the PR sandboxes. So this will need to be tested locally:

1. Pull that PPR changes locally.
2. Run the Backdrop installer -> confirm no errors that would block the installation
3. Once the installation is finished, go to `/admin/reports/status` -> there should be a warning about trusted host patterns not being configured 👍 
4. Edit your `settings.php` file, and un-comment the `$settings['trusted_host_patterns'] = array('^www\.example\.com$');` line. Save.
5. Refresh the "Status report" page -> the warning should now be telling you that the host is not trusted according to the `trusted_host_patterns` setting, and also show you a list of currently configured patterns 👍 
6. Keep changing the `trusted_host_patterns` setting in your `settings.php` file, saving, and then refreshing the "Status report" page. Test what happens to the "Trusted Host Settings" entry with none/one/multiple patterns, and with a variety of matching and non-matching ones. Report back if it mis-reports trusted or untrusted hosts during your tests.

quicksketch:
Fixes https://github.com/backdrop/backdrop-issues/issues/2568

Replaces https://github.com/backdrop/backdrop/pull/2382

### DIFF
--- a/core/modules/system/system.install
+++ b/core/modules/system/system.install
@@ -125,7 +125,7 @@ function system_requirements($phase) {
   }
 
   if (!empty($missing_extensions)) {
-    $description = $t('Backdrop requires you to enable the PHP extensions in the following list (see the <a href="@system_requirements">system requirements page</a> for more information):', array(
+    $description = $t('Backdrop requires you to enable the PHP extensions in the following list (see the <a href="@system_requirements" target="_blank">system requirements page</a> for more information):', array(
       '@system_requirements' => 'https://backdropcms.org/requirements',
     ));
 
@@ -148,24 +148,21 @@ function system_requirements($phase) {
 
     // Make sure PDO is available.
     if (!extension_loaded('pdo')) {
-      $pdo_message = $t('Your web server does not appear to support PDO (PHP Data Objects). Ask your hosting provider if they support the native PDO extension. See the <a href="@link">system requirements</a> page for more information.', array(
-        '@link' => 'https://backdropcms.org/requirements',
-      ));
+      $pdo_message = $t('Your web server does not appear to support PDO (PHP Data Objects). Ask your hosting provider if they support the native PDO extension.');
     }
     else {
       // Make sure the native PDO extension is available, not the older PEAR
       // version. (See install_verify_pdo() for details.)
       if (!defined('PDO::ATTR_DEFAULT_FETCH_MODE')) {
-        $pdo_message = $t('Your web server seems to have the wrong version of PDO installed. Backdrop CMS requires the PDO extension from PHP core. This system has the older PECL version. See the <a href="@link">System Requirements</a> page for more information.', array(
-          '@link' => 'https://backdropcms.org/requirements',
-        ));
+        $pdo_message = $t('Your web server seems to have the wrong version of PDO installed. Backdrop CMS requires the PDO extension from PHP core. This system has the older PECL version. See the <a href="@link" target="_blank">System Requirements</a> page for more information.');
       }
     }
 
     if (isset($pdo_message)) {
+      $pdo_more_info = $t('See the <a href="@link" target="_blank">System Requirements</a> page for more information.', array('@link' => 'https://backdropcms.org/requirements'));
       $requirements['database_extensions']['value'] = $t('Disabled');
       $requirements['database_extensions']['severity'] = REQUIREMENT_ERROR;
-      $requirements['database_extensions']['description'] = $pdo_message;
+      $requirements['database_extensions']['description'] = $pdo_message . ' ' . $pdo_more_info;
     }
     else {
       $requirements['database_extensions']['value'] = $t('Enabled');
@@ -228,7 +225,7 @@ function system_requirements($phase) {
         $description .= ' ' . $t('Contact your system administrator or hosting provider for assistance with increasing your PHP memory limit.');
       }
 
-      $requirements['php_memory_limit']['description'] = $description . ' ' . $t('See the <a href="@url">Backdrop requirements</a> for more information.', array('@url' => 'https://backdropcms.org/requirements'));
+      $requirements['php_memory_limit']['description'] = $description . ' ' . $t('See the <a href="@url" target="_blank">Backdrop requirements</a> for more information.', array('@url' => 'https://backdropcms.org/requirements'));
       $requirements['php_memory_limit']['severity'] = REQUIREMENT_WARNING;
     }
   }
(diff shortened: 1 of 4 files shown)


core/modules/system/system.install
```diff
@@ -602,6 +600,43 @@ function system_requirements($phase) {
...
+  // Warn the user if trusted hostnames have not been configured.
+  if ($phase == 'runtime') {
+    $trusted_host_patterns = settings_get('trusted_host_patterns', array());
+    $more_info = $t('See <a href="@url" target="_blank">Protecting against HTTP HOST Header attacks</a> for more information.', array('@url' => 'https://api.backdropcms.org/documentation/trusted-host-settings'));
+    // @todo: upgrade this to an error when we provide a way to configure
+    // $trusted_host_patterns via the admin UI.
+    if (empty($trusted_host_patterns)) {
+      $requirements['trusted_host_patterns'] = array(
+        'title' => $t('Trusted Host Setting'),
+        'value' => $t('Not enabled'),
```
```suggestion
        'value' => $t('Not configured'),
```

core/modules/system/system.install
```diff
@@ -602,6 +600,43 @@ function system_requirements($phase) {
...
+        $description .= ' ' . $trusted_host_patterns[0] . '<br>';
+      }
+      // ...otherwise, render the patterns as a list.
+      else {
+        $description .= theme('item_list', array('items' => $trusted_host_patterns));
+      }
+      $description .= $t('You can change this by editing the <code>trusted_host_patterns</code> setting in <code>settings.php</code>.') . ' ' . $more_info;
+      $requirements['trusted_host_patterns'] = array(
+        'title' => $t('Trusted Host Settings'),
+        'value' => $t('Enabled'),
```
```suggestion
        'value' => $t('Current hostname is trusted'),
```

core/modules/system/system.install
```diff
@@ -148,24 +148,21 @@ function system_requirements($phase) {
...
+      $pdo_message = $t('Your web server does not appear to support PDO (PHP Data Objects). Ask your hosting provider if they support the native PDO extension.');
     }
     else {
       // Make sure the native PDO extension is available, not the older PEAR
       // version. (See install_verify_pdo() for details.)
       if (!defined('PDO::ATTR_DEFAULT_FETCH_MODE')) {
-        $pdo_message = $t('Your web server seems to have the wrong version of PDO installed. Backdrop CMS requires the PDO extension from PHP core. This system has the older PECL version. See the <a href="@link">System Requirements</a> page for more information.', array(
-          '@link' => 'https://backdropcms.org/requirements',
-        ));
+        $pdo_message = $t('Your web server seems to have the wrong version of PDO installed. Backdrop CMS requires the PDO extension from PHP core. This system has the older PECL version. See the <a href="@link" target="_blank">System Requirements</a> page for more information.');
```
klonos: The array was removed, but not the `@link` from the link in `t()`. Not sure if you meant to remove the entire "See the..." bit there @quicksketch

klonos, line 157:
```suggestion
        $pdo_message = $t('Your web server seems to have the wrong version of PDO installed. Backdrop CMS requires the PDO extension from PHP core. This system has the older PECL version.');
```

...we're adding the "`"See the System Requirements page..." bit later on in the code.

core/modules/system/system.install
```diff
@@ -602,6 +600,43 @@ function system_requirements($phase) {
...
+    $trusted_host_patterns = settings_get('trusted_host_patterns', array());
+    $more_info = $t('See <a href="@url" target="_blank">Protecting against HTTP HOST Header attacks</a> for more information.', array('@url' => 'https://api.backdropcms.org/documentation/trusted-host-settings'));
+    // @todo: upgrade this to an error when we provide a way to configure
+    // $trusted_host_patterns via the admin UI.
+    if (empty($trusted_host_patterns)) {
+      $requirements['trusted_host_patterns'] = array(
+        'title' => $t('Trusted Host Setting'),
+        'value' => $t('Not configured'),
+        'description' => $t('The <code>trusted_host_patterns</code> setting is not configured in <code>settings.php</code>. It is highly recommended that you configure this to protect your site.') . ' ' . $more_info,
+        'severity' => $trusted_host_patterns === FALSE ? REQUIREMENT_INFO : REQUIREMENT_WARNING,
```
klonos, line 614: This didn't seem to be doing anything during my testing. I always saw a warning when there was no pattern match - not any info message ...if this is meant for the installer, then we need `if ($phase == 'runtime' || $phase == 'install') {`. Right?

klonos: ...oh well, maybe this works. The `backdrop` recipe lando is using skips the requirements steps (unless there are errors) 🤔

quicksketch: Per the updated documentation in settings.php:
```
 * If you do not need this functionality (such as in development environments or
 * if protection is at another layer), you can suppress the status report
 * warning by setting this value to FALSE:
 * @code
 * $settings['trusted_host_patterns'] = FALSE;
 * @endcode
 *
```

So you have to set it explicitly to FALSE. If you don't define it at all, then you get the warning.

PR #2838, opened 2019-09-01
2632 fieldable files new

### LINKED ISSUE #2632: Core fieldable file entity bundles

jenlampton:
Fixes https://github.com/backdrop/backdrop-issues/issues/2632
Replaces https://github.com/backdrop/backdrop/pull/2832


core/modules/file/file.module
```diff
@@ -4,9 +4,92 @@
...
+ * Implements hook_hook_info().
+ */
+function file_hook_info() {
+  $hooks = array(
+    // File API hooks
+    'file_copy',
+    'file_move',
+    'file_validate',
+    'file_type_info',
+    'file_type_info_alter',
```
quicksketch, line 58: I can't find any documentation or usage of any of these above hooks, e.g. `hook_file_copy()`, `hook_file_move()`, `hook_file_validate()`, `hook_file_type_info()`, and `hook_file_type_info_alter()`. Do these hooks actually exist?

quicksketch: Oh I see `hook_file_copy/move/validate()` are existing hooks that aren't modified in this PR. Where's `file_type_info()` used though?

jenlampton: I removed the hooks we don't have anymore from this list.

klonos: ...and those comments need periods 😄 ...OCDs, sorry 😅

core/modules/file/file.install
```diff
@@ -713,3 +852,106 @@ function file_update_1002() {
...
+}
+
+/**
+ * Add the default file types for existing Backdrop sites.
+ */
+function file_update_1006() {
+  $default_file_types = array('audio', 'document', 'image', 'video');
+  foreach ($default_file_types as $type) {
+    if (!file_type_get_type($type)) {
+      config_install_default_config('file', 'file.type.' . $type);
```
quicksketch: We can't assume that the default config will always be the same. Similar to when we have to make DB changes or views, you have to duplicate or write the whole config manually here.

core/modules/file/file.install
```diff
@@ -713,3 +852,106 @@ function file_update_1002() {
...
+  variable_del('file_entity_default_file_directory');
+  variable_del('file_entity_allow_insecure_download');
+  variable_del('file_entity_file_upload_wizard_skip_file_type');
+  variable_del('file_entity_file_upload_wizard_skip_scheme');
+  variable_del('file_entity_file_upload_wizard_skip_fields');
+  variable_del('file_entity_protect_repeated_render');
+
+  // Delete unused variables.
+  variable_del('file_entity_alt');
+  variable_del('file_entity_title');
```
quicksketch: These should all be switched to `update_variable_del()`, which is not deprecated.

core/modules/file/file.install
```diff
@@ -713,3 +852,106 @@ function file_update_1002() {
...
+ */
+function file_update_1006() {
+  $default_file_types = array('audio', 'document', 'image', 'video');
+  foreach ($default_file_types as $type) {
+    if (!file_type_get_type($type)) {
+      config_install_default_config('file', 'file.type.' . $type);
+    }
+  }
+  // Clear the necessary caches.
+  file_type_cache_reset();
```
quicksketch: All caches are cleared after running update.php, this should not be necessary.

jenlampton: removed.

core/modules/file/file.install
jenlampton: fixed.

core/modules/file/file.install
jenlampton: reworked.

core/modules/file/file.install
```diff
@@ -713,3 +857,179 @@ function file_update_1002() {
...
+      'mimetypes' => array(
+          'video/*',
+      ),
+      'disabled' => FALSE,
+      'storage' => 4,
+      'module' => 'file',
+    ),
+  );
+  foreach ($default_file_types as $machine => $type) {
+    if (!file_type_get_type($machine)) {
```
quicksketch: Instead of using an API function which may change in the future, or in some weird situations the .module file might not be loaded, use `$config = config('file.type.' . $machine);` and then check for `$config->isNew()`.

core/modules/file/file.install
```diff
@@ -105,11 +105,23 @@ function file_schema() {
...
+        'description' => 'The type of this file.',
+        'type' => 'varchar',
+        'length' => 50,
+        'not null' => TRUE,
+        // If the FILE_TYPE_NONE constant ever changes, then change the value here
+        // too, and add an update function to deal with existing records. The
+        // constant isn't used here, because there may be cases where this function
+        // runs without the module file loaded.
+        'default' => 'undefined',
+      ),
```
quicksketch, line 118: We need another update hook that adds this new column, and the new `file_type` index to this table. I tried updating from Backdrop 1.13 to this PR and the `type `column appears to be missing.

core/modules/file/file.install
```diff
@@ -695,9 +740,108 @@ function file_update_1001() {
 }
 
 /**
- * Switch the 'edit files' permission from file_entity to 'manage files'.
+ * Add {file_managed}.type field
  */
 function file_update_1002() {
+  $schema = array();
+  $schema['file_metadata'] = array(
```
quicksketch, line 767: Coming from Backdrop 1.13, this update hook isn't run. Why are you replacing `file_update_1002()` instead of just making this `file_update_1003()`?

jenlampton: file update 1002 has been there since the very begining. This is the one that is broken for D7 sites that have file_entity module. The change here fixes it.

core/modules/file/file.install
jenlampton: re-numbered the updates, running test update again now.

core/modules/file/file.install
jenlampton: ah, this is because the PR predates that update hook. fixed now.

core/modules/file/file.install
jenlampton: done, though an API function shouldn't change.

core/modules/file/file.module
```diff
@@ -1139,3 +1956,1333 @@ function file_get_file_references(File $file, $field = NULL, $age = FIELD_LOAD_R
...
+
+/**
+ * Deprecated wrapper function. Determines if a user may perform the given
+ * operation on the specified file.
+ *
+ * @see file_access()
+ * @deprecated since 1.14.0
+ */
+function file_entity_access($op, $file = NULL, $account = NULL) {
+  watchdog_deprecated_function('file', __FUNCTION__);
```
```suggestion
  watchdog_deprecated_function('file', __FUNCTION__, t('Change record:') . ' ' . l(t('File access functions have been renamed from <code>file_entity</code> module in Drupal 7'), 'https://api.backdropcms.org/change-records/file-access-functions-have-been-renamed-from-file_entity-module-in-drupal-7', array('attributes' => array('target' => '_blank'))));
```

jenlampton:
closing for new sandbox.

jenlampton:
sandbox me?

quicksketch:
Yay! Merged into 1.x for 1.14.0.

PR #2841, opened 2019-09-02
Issue #3931: Remove default stop words from URL aliases.

### LINKED ISSUE #3931: Remove all default words (stop words) from path alias "Strings to Remove" settings
**Description of the need**
It's bothered me for a while that Pathauto removes prepositions (of, over, up, etc) and articles (the, an, a) from URLs by default. I ran across an article describing exactly why this is a bad idea:

https://onlinemediamasters.com/why-you-shouldnt-remove-stop-words-from-urls/

> Removing stop words can actually hurt your SEO because it can make your URLs read differently. For example, /growing-up-with-hearing-loss/ is NOT the same thing as /growing-hearing-loss/ (the version where stop words are removed). 

I also noticed that StackOverflow always includes all words:

Title: "How to make a whole 'div' clickable in html and css without JavaScript?"
URL:   https://stackoverflow.com/questions/8160494/how-to-make-a-whole-div-clickable-in-html-and-css-without-javascript

Whereas if this were using Backdrop's default rules it would read `how-make-whole-div-clickable-html-css-without-javascript`, which IMO is both less readable and harder to remember.

**Proposed solution**
All default words should be removed from the "Words to Remove" setting on Path Alias Settings.

---
PR by @BWPanda: https://github.com/backdrop/backdrop/pull/2841

ghost:
Fixes https://github.com/backdrop/backdrop-issues/issues/3931

### DIFF
--- a/core/modules/path/config/path.settings.json
+++ b/core/modules/path/config/path.settings.json
@@ -9,7 +9,7 @@
     "update_action": 2,
     "transliterate": false,
     "reduce_ascii": false,
-    "ignore_words": "a, an, as, at, before, but, by, for, from, is, in, into, like, of, off, on, onto, per, since, than, the, this, that, to, up, via, with",
+    "ignore_words": "",
     "punctuation_double_quotes": 0,
     "punctuation_quotes": 0,
     "punctuation_backtick": 0,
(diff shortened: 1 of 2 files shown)


quicksketch:
Merged into 1.x for 1.15.0.

PR #2843, opened 2019-09-03
Issue #871: Fixed node type operation links overlap.

### LINKED ISSUE #871: Content types owned by modules other than "node" have incorrect "Delete" link
After installing [Project module](https://github.com/backdrop-contrib/project) I wanted to delete the "Project" content type it came with to set up my own different types. However, I found that on `admin/structure/types`, the link to Delete the "Project" content type actually was linking to the form to delete the Page content type! This obviously could cause some accidental data loss if the user were not to read the confirmation form.

My guess is that this behavior is caused by project module specifying a "module" in the content type when it creates it hook_install(). This is supposed to prevent the content type from being renamed, but looks like it's also having some odd behavior when deleting the content type.

ghost:
Fixes https://github.com/backdrop/backdrop-issues/issues/871

### DIFF
--- a/core/modules/node/node.types.inc
+++ b/core/modules/node/node.types.inc
@@ -23,6 +23,7 @@ function node_overview_types() {
     $type = $types[$key];
     $type_url_str = str_replace('_', '-', $type->type);
     $row = array(theme('node_admin_overview', array('name' => $name, 'type' => $type)));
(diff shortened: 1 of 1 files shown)


klonos (approved):
RTBC 👍

PR #2851, opened 2019-09-06
Issue #2655: Fixed inconsistent sorting issues with backdrop_sort().

### LINKED ISSUE #2655: Fix ordering issues in admin menu
**Status update (8th Jan)**

The PR is passing all tests and seems to be working nicely. There have been multiple reviews (of different versions), but the latest version has been marked RTBC, so I think this is good to go!

---

In #2502 @olafgrabienski pointed out that the content type ordering is inconsistent in the admin bar:

> Looks good at first view. Then I've seen another place where the content types don't sort correctly: The *Add content* links in the Admin bar seem to not being ordered at all:

screen-backdrop-pr-1809-sort-order-add-content-admin-bar

Let's examine, identify weird or inconsistent ordering in admin menu and decide what to do about it in this issue.

---

PR by @BWPanda: https://github.com/backdrop/backdrop/pull/2851

ghost:
Fixes https://github.com/backdrop/backdrop-issues/issues/2655


ghost:
Updated PR with latest core changes to see if it affects tests at all...

herbdool:
Restarting test box

core/includes/common.inc
```diff
@@ -7132,81 +7149,39 @@ function backdrop_sort(array &$array, array $keys = array('weight'), $dir = SORT
...
-      if (!isset($b[$key])) {
-        $b[$key] = 0;
-      }
-      if ($a[$key] == $b[$key]) {
-        // Sort on the next key if one exists.
-        if (isset($keys_map[$key_index + 1])) {
-          return $recursive_callback($a, $b, $key_index + 1);
-        }
+  // Create an array to use for sorting by each key.
+  $sort = array();
```
herbdool, line 7149: Above `$sort` is assigned a constant (ie. an integer) but here it becomes an array. I'm guessing one or the other variable needs to be renamed.

core/includes/common.inc
```diff
@@ -7132,81 +7149,39 @@ function backdrop_sort(array &$array, array $keys = array('weight'), $dir = SORT
   }
   $keys = $new_keys;
 
-  // If sorting on a single value, optimize the sorting to sort only on that
```
herbdool, line 7135: I'm curious if there's a way to simplify for case where sorting by a single value, like before.

herbdool (changes requested):

core/includes/common.inc
ghost: Depends... Are you worried about the variable types getting mixed up, or just the readability of the code?
If the former, no worries as the first `$sort` is scoped to the `foreach()` statement, while the second is outside it.
If the latter, I guess it might be confusing for some, but the first `$sort` has done its job and is then re-assigned for a different purpose further on in the function...

core/includes/common.inc
ghost: I'm not sure. You can read about the PHP sorting function we're now using here: https://www.php.net/manual/en/function.array-multisort But I couldn't find any references there to optimisation, performance, etc...

core/includes/common.inc
herbdool: Alright, I get it. I didn't see that you assigned it to `$new_key` so you don't need it anymore. Maybe for legibility they could be different.

ghost: New commit changes previous `$sort` to be consistent with the rest of the code (i.e. all `foreach()` iterating over the `$keys` array now uses `$key` and `$key_sort`. All iterations of the `$array` array use `$key` and `$value` (with the exception of one where that'd conflict with the `$keys` array iteration so uses `$k` and `$v` instead)).

core/includes/common.inc
```diff
@@ -7112,101 +7112,72 @@ function element_info_property($type, $property_name, $default = NULL) {
...
-        if (isset($keys_map[$key_index + 1])) {
-          return $recursive_callback($a, $b, $key_index + 1);
-        }
+  // Create an array to use for sorting by each key.
+  $sort = array();
+  foreach ($array as $k => $v) {
+    $v = (array) $v;
+    foreach ($keys as $key => $key_sort) {
+      if (isset($v[$key])) {
+        $sort[$key][$k] = ($key_sort === SORT_STRING) ? backdrop_strtolower($v[$key]) : $v[$key];
```
herbdool, line 7154: @BWPanda it looks like if we're certain that the strings are going to be ASCII only then we can use `strtolower()` otherwise `backdrop_strtolower()` is the best option. I guess we can't predict that every use case will be ASCII only.

ghost: > I guess we can't predict that every use case will be ASCII only.

Yeah, and with the possibility for emoji in the database, eventually they might make it into core... So since this doesn't seem like a regression performance-wise, I'd say leave it as `backdrop_strtolower()`.

quicksketch:

PR #2856, opened 2019-09-09
Issue #4021: Property langcode of non-object

indigoxela:
See details in https://github.com/backdrop/backdrop-issues/issues/4021

If language isn't set, $languages[$options['language']] is no object.

### DIFF
--- a/core/includes/locale.inc
+++ b/core/includes/locale.inc
@@ -422,6 +422,7 @@ function locale_language_url_rewrite_url(&$path, &$options) {
   if (!isset($languages)) {
     // Get the enabled languages only.
     $languages = language_list(TRUE);
+    $languages = array_keys($languages);
   }
 
   // Array key doesn't even exist in the options.
--- a/core/includes/locale.inc
+++ b/core/includes/locale.inc
@@ -431,7 +431,7 @@ function locale_language_url_rewrite_url(&$path, &$options) {
     $options['language'] = $language_url;
   }
   // We allow only enabled languages here.
-  elseif (!isset($languages[$options['language']])) {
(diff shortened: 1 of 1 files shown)


core/includes/locale.inc
```diff
@@ -422,21 +422,27 @@ function locale_language_url_rewrite_url(&$path, &$options) {
...
     // Get the enabled languages only.
     $languages = language_list(TRUE);
-    $languages = array_flip(array_keys($languages));
+    $languages = array_keys($languages);
   }
 
-  // Language can be passed as an option, or we go for current URL language.
+  // Array key doesn't even exist in the options.
   if (!isset($options['language'])) {
+    $options['language'] = FALSE;
```
quicksketch: I don't think this is safe... it could have effects elsewhere because `$options` is passed by reference, but setting `$options['language'] = FALSE` isn't something that is expected by the `url()` function. It's documentation says that it's either `NULL` or an object.

> 'language': An optional language object. If the path being linked to is internal to the site, $options['language'] is used to look up the alias for the URL. If $options['language'] is omitted, the global $language_url will be used.

indigoxela: While working on this, it turned out that `$options['language']` can be a lot of things or it doesn't even exist. (See my [comment](https://github.com/backdrop/backdrop-issues/issues/4021#issuecomment-529926814).)
This really just tries to prevent "undefined index" warnings.

Anyway - re-reading the code, I see that setting it to NULL would be an option, too.
And maybe I should do the same for disabled languages (line 440).

In line 445 I check for an _object_ and ignore anything else.

quicksketch:
Thanks! This looks great. Marked RTBC in the issue and I don't see any issues. Merged into 1.x and 1.14.x.

PR #2860, opened 2019-09-09
Issue #4022: [UX] Missing translation calls fixed

VasasA:
Missing translation function calls in `node.module`
Issue: https://github.com/backdrop/backdrop-issues/issues/4022

### DIFF
--- a/core/modules/node/node.module
+++ b/core/modules/node/node.module
@@ -569,7 +569,7 @@ function node_add_body_field($type, $label = 'Body') {
       'field_name' => 'body',
       'entity_type' => 'node',
       'bundle' => $type->type,
-      'label' => $label,
+      'label' => t($label),
       'widget' => array('type' => 'text_textarea_with_summary'),
       'settings' => array('display_summary' => TRUE),
       'display' => array(


core/modules/node/node.module
```diff
@@ -569,7 +569,7 @@ function node_add_body_field($type, $label = 'Body') {
       'field_name' => 'body',
       'entity_type' => 'node',
       'bundle' => $type->type,
-      'label' => $label,
+      'label' => t($label),
```
klonos, line 572: You cannot translate variables, so you cannot do this. Perhaps try adding the `t()` function in the function declaration? (I have not tested this before though):

```php
function node_add_body_field($type, $label = t('Body')) {
...
```

If that doesn't work, then we will need to remove the default value from the function declaration, and check if empty within the function:

```php
function node_add_body_field($type, $label) {
  if (empty($label)) {
    $label = t('Body');
  }
```

docwilmot: Search for ` t($` in core yields 81 results.

klonos: > Search for `t($` in core yields 81 results.

That is sad, if the only thing contained in those `t()` functions is a variable 😞 ...translators will literally see "$label", and how are they to translate that?

klonos: ...this is the patch that went in for D9: https://www.drupal.org/files/issues/t-in-fields-is-evil.patch ...I am guessing that there must be a reason why they named that patch that way 😄

klonos: Screen Shot 2019-09-10 at 4 29 24 am

...there's more to it: https://api.backdropcms.org/api/backdrop/core%21includes%21bootstrap.inc/function/t/1

VasasA:
This PR is bad. Sorry!
Replaced by: https://github.com/backdrop/backdrop/pull/2866

PR #2865, opened 2019-09-09
#4019: Reassigning blocks with flexible layout templates -v2

laryn:


core/modules/layout/includes/layout.class.inc
```diff
@@ -675,7 +675,12 @@ class Layout {
 
     $old_layout_info = layout_get_layout_template_info($this->layout_template);
     $new_layout_info = layout_get_layout_template_info($layout_name);
+     // Set up the array with region names for the new layout.
```
klonos:
```suggestion
    // Set up the array with region names for the new layout.
```

core/modules/layout/includes/layout.class.inc
```diff
@@ -710,11 +712,12 @@ class Layout {
         }
       }
     }
+
+    // Assign orphan blocks to the default or last region,
```
klonos:
```suggestion
    // Assign orphan blocks to the default or last region, if the former doesn't
```

core/modules/layout/includes/layout.class.inc
```diff
@@ -710,11 +712,12 @@ class Layout {
...
+    // if the former doesn't exist.
```
klonos:
```suggestion
    // exist.
```

PR #2871, opened 2019-09-13
Issue #4032: Remove 'navigation' role from 'nav' elements.

### LINKED ISSUE #4032: [A11Y]  Remove 'navigation' role from 'nav' elements
`<nav role="navigation" class="breadcrumb">`
This code gives the following warning when [validated](https://validator.w3.org/):

> The `navigation` role is unnecessary for element `nav`.

We should change instances of the above code to `<nav class="breadcrumb">` instead.

For more information, see: https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/Navigation_Role

---

PR: https://github.com/backdrop/backdrop/pull/2871

ghost:
Fixes https://github.com/backdrop/backdrop-issues/issues/4032

### DIFF
--- a/core/includes/theme.inc
+++ b/core/includes/theme.inc
@@ -1995,7 +1995,7 @@ function theme_breadcrumb($variables) {
   $breadcrumb = $variables['breadcrumb'];
   $output = '';
   if (!empty($breadcrumb)) {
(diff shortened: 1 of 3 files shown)


laryn:
Thanks @BWPanda for filing and fixing! Merged into 1.x and 1.19.x.

PR #2872, opened 2019-09-13
Issue #4033: Remove 'article' role from 'article' elements.

### LINKED ISSUE #4033: [A11Y] Remove 'article' role from 'article' elements
Similar to https://github.com/backdrop/backdrop-issues/issues/4032...

`<article ... role="article">`
This code gives the following warning when validated:

> The `article` role is unnecessary for element `article`.

We should change instances of the above code to remove `role="article"`. Note that we can't do this in every situation, as some `role` attributes are added in preprocess functions, but the element type is set in the template file (e.g. `node` module).

For more information, see: https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/Article_Role

---

PR: https://github.com/backdrop/backdrop/pull/2872

ghost:
Fixes https://github.com/backdrop/backdrop-issues/issues/4033


klonos (approved):
LGTM 👍🏼

quicksketch:
Merged into 1.x and 1.24.x. Thanks @BWPanda!

PR #2873, opened 2019-09-13
Issue #4034: Replace 'footer' roles with 'contentinfo'.

### LINKED ISSUE #4034: No 'footer' ARIA role
There doesn't seem to be any such ARIA role as `footer`. Instead, `footer` elements should be given the `contentinfo` role: https://developer.mozilla.org/en-US/docs/Web/HTML/Element/footer#Accessibility_concerns

Instances of `<footer ... role="footer">` should be fixed.

---

PR: https://github.com/backdrop/backdrop/pull/2873

ghost:
Fixes https://github.com/backdrop/backdrop-issues/issues/4034


klonos (approved):
LGTM ...and thanks for fixing those double spaces while at it @BWPanda 👍

klonos (changes requested):
edit: actually no; please see my comment in https://github.com/backdrop/backdrop-issues/issues/4034#issuecomment-593107560

klonos (approved):
Thanks @BWPanda ...now LGTM 😉

quicksketch:
Merged into 1.x and 1.15.x for 1.15.1. Thanks!

PR #2882, opened 2019-09-15
Issue #4039: Grant the Admin role all new file related permissions.

### LINKED ISSUE #4039: Core fieldable file permission issues after upgrade from 1.13.3
**Description of the bug**

Not yet sure if this is a bug. Looking for some help to determine if I'm misunderstanding something.

I just upgraded live version of Triplo.co from 1.13.3 to 1.14.x-dev and I am not seeing any of the UI options for core fieldable files. 

I spun up a fresh new site with the same code and on the fresh new site I have the ability to:
- add a file on the manage files page
- In the structure menu, I have a file types menu

Neither of these options are available to me on my upgraded Triplo.co site. 

I ran the database updates and got the message about "Existing files must be classified according to the currently defined file types"

Is there anything else I need to do?

I have not configured Trusted Host Patterns, but I don't think these are related.

---

PR: https://github.com/backdrop/backdrop/pull/2882

jenlampton:
Fixes https://github.com/backdrop/backdrop-issues/issues/4039

### DIFF
--- a/core/modules/file/file.install
+++ b/core/modules/file/file.install
@@ -1141,3 +1141,24 @@ function file_update_1006() {
     db_drop_table('file_display');
   }
 }
+
+/**
+ * Grant the admin role permission to manage files and file types.
+ */
+function file_update_1007() {
+  $admin_role = config_get('system.core', 'user_admin_role');
+
+  $config = config('user.role.' . $admin_role);
+  $permissions = $config->get('permissions');
+  // Add the new permissions.
+  $permissions[] = 'bypass file access';
+  $permissions[] = 'administer file types';
+  $permissions[] = 'create files';
+  $permissions[] = 'view own private files';
+  $permissions[] = 'view own files';
+  $permissions[] = 'view private files';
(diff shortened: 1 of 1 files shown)


core/modules/file/file.install
```diff
@@ -1141,3 +1141,24 @@ function file_update_1006() {
...
 }
+
+/**
+ * Grant the admin role permission to manage files and file types.
+ */
+function file_update_1007() {
+  $admin_role = config_get('system.core', 'user_admin_role');
+
+  $config = config('user.role.' . $admin_role);
+  $permissions = $config->get('permissions');
```
quicksketch, line 1152: Can we do a safety-check here to ensure the config file exists?
```
if (!$config->isNew()) {
```

quicksketch: The admin role _should_ always exist but having a partial config file for a user role would probably cause all sorts of issues.

jenlampton:
I got all of the file permissions!

PR #2886, opened 2019-09-17
Issue #4061: Fixed comment wording in filter.module

### LINKED ISSUE #4061: Fix ambiguous wording in filter.module comment
In `filter.module` there's the following comment and code:

```php
// Programmatic saves may not contain any filters.
if (!isset($format->filters)) {
  $format->filters = array();
}
```

This sounds like it means "Programmatic saves are not allowed to contain any filters", but the code seems to suggest it's actually "Programmatic saves might not contain any filters."

I suggest changing this to avoid confusion.

ghost:
Fixes https://github.com/backdrop/backdrop-issues/issues/4061


ghost:
Merged into 1.x and 1.21.x.

PR #2887, opened 2019-09-19
#4066 Fix default classes in header region for Flexible Layouts

stpaultim:
Adds l-[element] class to each row.


klonos (approved):
LGTM 👍

PR #2889, opened 2019-09-20
Issue #4072: Use labels instead of UUIDs for region names in the blocks listing.

klonos:
https://github.com/backdrop/backdrop-issues/issues/4072

### DIFF
--- a/core/modules/layout/layout.admin.inc
+++ b/core/modules/layout/layout.admin.inc
@@ -763,12 +763,12 @@ function layout_settings_form_save_submit($form, &$form_state) {
 }
 
 /**
- * Form callback; Main form for editing a layout's content.
+ * Form callback; Main form for editing a layout's content ("Manage blocks").
  *
- * No validation function is necessary, as all 'validation' is handled
- * either in the lead-up to form rendering (through the selection of
- * specified content types) or by the validation functions specific to
- * the ajax modals & content types.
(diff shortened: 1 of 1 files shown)


core/modules/layout/layout.admin.inc
```diff
@@ -2999,9 +2999,28 @@ function layout_block_list() {
...
+      // Layouts that use flexible templates return cryptic UUIDs as region
+      // names. Replace these UUIDs with the respective user-friendly labels.
+      if ($layout_has_flexible_template) {
+        $flexi_regions = array_flip($all_template_info[$layout_template]['regions']);
+
+        foreach ($flexi_regions as $flexi_region_name => $uuid) {
+          $flexi_regions[$flexi_region_name] = $layout->positions[$uuid];
+        }
+
+        $layout->positions = $flexi_regions;
```
quicksketch, line 3021: Let's use a different variable instead of overridding `$layout->positions`. There's a chance that if this layout were saved back to config with `$layout->save()`, it would corrupt the config data.

klonos:
...this PR originates from a wrongly-named branch, which makes it confusing. Replaced by https://github.com/backdrop/backdrop/pull/2925

core/modules/layout/layout.admin.inc
klonos: 👍 ...I have fixed this in https://github.com/backdrop/backdrop/pull/2925

PR #2894, opened 2019-09-22
Issue #2615: Allow restricting text formats per field.

### LINKED ISSUE #2615: [UX] Add the option to limit text formats for each field instance.
This came up in a Gitter discussion between @laryn and @klonos and it is basically getting https://www.drupal.org/project/better_formats in core. There is a respective D8 core issue: https://www.drupal.org/node/784672 (from where https://www.drupal.org/project/allowed_formats was born). From the issue summary in d.org (some emphasis mine):

> **Problem/Motivation**
> 
> Currently text fields can either be restricted to plain text, or the user may select between **all accessible text formats** independently of the context. This means that a privileged user who needs access to a permissive text formats (for example, to put tables or embedded remote content in basic pages) will get access to that format **on every formatted text field** (for example on a comment field).
> 
> There are three problems with this approach, and most experience Drupal developers have faced at least one of those in the past:

>   - **Consistency**
>     At the moment you have to count on competence, good will and diligence of privileged users not to put certain markup in certain places. It would be convenient if a text field could be forced to use a specific text format (other than plain text). For example, you may want to make sure that comments only allow a very limited set of HTML tags ("filtertered HTML" for example) independently of the user's role, even if the same user has access to less restrictive formats in other places. 
>   - **Usability**
>     The ability to select text formats is a common source of confusion. By limiting the available text formats we also remove confusing user interface elements.
>   - **Security**
>     If a privileged user account is taken over (for example, through social engineering), the attack surface is large due to the fact that every single text field can be used for XSS injections. By limiting where a dangerous text format can be used, we restrict the possibilities to inject malicious content.

> **Proposed resolution**
> 
> Add an optional setting to the text field types that lets the site-builder determine if the text formats should be restricted. This setting is then used in the default textfield and textarea widgets to remove any non-allowed text formats. If nothing is set, the current behavior is unchanged.
> 
> Note that as it uses the underlying `'#allowed_formats'` form API property, the settings can't be used to give access to text formats that the user wouldn't have access to otherwise.
> 
> ...

> **User interface changes**
> 
> Checkboxes on list of available formats on text field configuration. Reduced set of allowed formats on content edit forms, where used. No fields use the new setting by default, so the patch doesn't affect the user interface for those who don't do anything with this functionality.
> 
> **API changes**
> 
> None
> 
> **Data model changes**
> 
> One setting is added to the field settings. The structure of the field data is unchanged.

---

PR: https://github.com/backdrop/backdrop/pull/2894

Advocate: @quicksketch

ghost:
Fixes https://github.com/backdrop/backdrop-issues/issues/2615

### DIFF
--- a/core/modules/filter/filter.module
+++ b/core/modules/filter/filter.module
@@ -50,6 +50,7 @@ function filter_element_info() {
   $type['text_format'] = array(
     // The text format selected by default for this element.
     '#format' => NULL,
+    '#allowed_formats' => array(),
     // Whether file uploads are permitted through the editor (if supported).
     // Disable by default, modules must specifically implement saving files as
     // permanent when they save their text data. See filter_entity_insert().
@@ -1218,8 +1219,13 @@ function filter_process_format($element) {
   $element['value']['#type'] = $element['#base_type'];
   $element['value'] += element_info($element['#base_type']);
 
-  // Get a list of formats to which the current user has access.
+  // Get a list of allowed formats to which the current user has access.
   $formats = filter_formats($user);
+  foreach ($formats as $format) {
+    if (!in_array($format->format, array_filter($element['#allowed_formats']))) {
+      unset($formats[$format->format]);
+    }
+  }
 
   // Turn original element into a text format wrapper.
   if (isset($element['#attached'])) {
--- a/core/modules/filter/filter.module
+++ b/core/modules/filter/filter.module
@@ -1238,7 +1238,7 @@ function filter_process_format($element) {
   }
 
   // Remove the Plain text format if not set and other options are available.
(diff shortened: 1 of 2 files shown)


core/modules/filter/filter.module
```diff
@@ -1238,7 +1244,7 @@ function filter_process_format($element) {
   }
 
   // Remove the Plain text format if not set and other options are available.
-  $fallback_format = config_get('system.core', 'filter_fallback_format');
+  $fallback_format = filter_fallback_format();
```
klonos, line 1247: 👍

quicksketch:
Replaced by https://github.com/backdrop/backdrop/pull/4423

PR #2896, opened 2019-09-23
Issue #1505: Lowercase file names by default.

### LINKED ISSUE #1505: Transliteration: enable the "Lowercase transliterated file names" setting by default
This is the recommended setting, yet it is not activated by default. Allowing both lower and upper case will cause problems for Windows users.

---
PR by @BWPanda: https://github.com/backdrop/backdrop/pull/2896

ghost:
Fixes https://github.com/backdrop/backdrop-issues/issues/1505


quicksketch:

PR #2900, opened 2019-09-25
Issue #3973: Stay on same page after running cron from admin bar.

### LINKED ISSUE #3973: [UX] Running cron via the Admin Bar menu always redirects to the "Status report" page
**Steps To Reproduce**
Run cron via the "Run cron" menu item, under "Home"...

**Actual behavior**
You get redirected to `/admin/reports/status` and get a "Cron ran successfully." message

**Expected behavior**
You should stay on the same page.

- - - - - - - 
PR by @BWPanda - https://github.com/backdrop/backdrop/pull/2900

ghost:
Fixes https://github.com/backdrop/backdrop-issues/issues/3973


core/modules/system/system.admin.inc
```diff
@@ -2242,7 +2235,12 @@ function system_run_cron() {
     backdrop_set_message(t('Cron run failed.'), 'error');
   }
 
-  backdrop_goto('admin/reports/status');
+  if ($_SERVER['HTTP_REFERER']) {
```
klonos, line 2274: Perhaps explore using `backdrop_get_destination()` here instead?

ghost: I looked into that originally, but [it says](https://api.backdropcms.org/api/backdrop/core%21includes%21common.inc/function/backdrop_get_destination/1) "By default the current URL is returned." So that would only work if each and every place that calls this set a destination parameter in the URL. Currently none do, so I thought it easier to do it this way.

PR #2903, opened 2019-09-26
Issue #1610: Add Pages to menu by default.

### LINKED ISSUE #1610: [UX] 'Auto Menu Settings' in core and add Pages to the main menu by default
### Weekly update (19th Dec)

https://github.com/backdrop/backdrop/pull/2999 is RTBC, but more feedback/wording changes requested...

----

Drupal has long had this user-nightmare of lost pages. You made one, now where did it go? Yes, the experienced user knows about the content orphanage where all content is listed, but to a new user losing these pages can be traumatic. Backdrop now has this problem too.

If we placed basic pages into the primary navigation menu by default, then first-time users would be able to find their recently created content without needing to visit the content orphanage.

Experienced users could disable this behavior, or simply move pages on a case-by-case basis out of this menu.

## Proposed solution

* On the Page node form, check the box for "create menu item" by default.
* Add 2 per-content type settings: 1) create menu item by default, and 2) which menu.

---

Advocate for this issue: @BWPanda 

----

Original PR by @BWPanda: ~https://github.com/backdrop/backdrop/pull/2903~
Alternate PR by @jenlampton: ~https://github.com/backdrop/backdrop/pull/2951~
New PR by @BWPanda: https://github.com/backdrop/backdrop/pull/2999

ghost:
Fixes https://github.com/backdrop/backdrop-issues/issues/1610

### DIFF
--- a/core/modules/menu/tests/menu.test
+++ b/core/modules/menu/tests/menu.test
@@ -681,6 +681,7 @@ class MenuNodeTestCase extends BackdropWebTestCase {
     $edit = array(
       "title" => $node_title,
       "body[$language][0][value]" => $this->randomString(),
+      "menu[enabled]" => 0,
     );
     $this->backdropPost('node/add/page', $edit, t('Save'));
     $node = $this->backdropGetNodeByTitle($node_title);
--- a/core/modules/menu/tests/menu.test
+++ b/core/modules/menu/tests/menu.test
@@ -662,17 +662,17 @@ class MenuNodeTestCase extends BackdropWebTestCase {
     $edit = array(
       'menu_options[main-menu]' => 1,
     );
-    $this->backdropPost('admin/structure/types/manage/page', $edit, t('Save content type'));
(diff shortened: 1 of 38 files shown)


jenlampton:
In the sandbox, the `Provide a menu link` checkbox was not checked when creating a new Page node.

core/modules/menu/tests/menu.test
```diff
@@ -681,6 +681,7 @@ class MenuNodeTestCase extends BackdropWebTestCase {
     $edit = array(
       "title" => $node_title,
       "body[$language][0][value]" => $this->randomString(),
+      "menu[enabled]" => 0,
```
herbdool, line 684: can't tell if anything is wrong but since the test is using backdropPost it means it's simulating using the browser so you could look at HTML to see what options are available. If that's a node add form perhaps menu isn't even there? You'd then need to set that separately before this

jenlampton:
Replaced by https://github.com/backdrop/backdrop/pull/2951

PR #2910, opened 2019-09-28
BD-4103: Check before and after file_delete() to show metadata is really deleted.

jlfranklin:
Addresses backdrop/backdrop-issues#4103


klonos (approved):
LGTM 👍

PR #2911, opened 2019-09-28
Issue #1828: [UX] Seven: Fix font-weight for labels inside fieldsets

### LINKED ISSUE #1828: [UX] Fix font-weight for labels inside fieldsets in Seven theme
In https://github.com/backdrop/backdrop/blob/1.x/core/modules/system/css/system.theme.css#L103 we set the font to bold for labels, but in https://github.com/backdrop/backdrop/blob/1.x/core/themes/seven/css/style.css#L780 we override it with normal.

This causes the following:
- forms without any fieldsets get bold labels for all fields
- forms with _all fields_ wrapped in fieldsets get normal labels for all fields
- forms with _some_ of their fields wrapped in fieldsets get a mix of both bold and normal labels

Was this intentional? If yes, what is the reasoning?

---
PR by @klonos: https://github.com/backdrop/backdrop/pull/2911

klonos:
Fixes https://github.com/backdrop/backdrop-issues/issues/1828

### DIFF
--- a/core/themes/seven/css/style.css
+++ b/core/themes/seven/css/style.css
@@ -822,9 +822,6 @@ div.teaser-checkbox .form-item,
 .form-item label {
   margin: 0 0 10px;
 }
-fieldset label {
-  font-weight: normal;
-}
 .field-label label {
   margin: 0;
   padding: 0;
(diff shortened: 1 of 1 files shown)


jenlampton:
This code looks great, only two CSS removals. RTBC.

quicksketch:

PR #2914, opened 2019-10-02
Issue #4105: `backdrop_mkdir` does not set permissions to directories it created recursively

klonos:
https://github.com/backdrop/backdrop-issues/issues/4105

https://www.drupal.org/project/drupal/issues/2789723
https://www.drupal.org/files/issues/2789723-1.patch

### DIFF
--- a/core/includes/file.inc
+++ b/core/includes/file.inc
@@ -2526,8 +2526,7 @@ function backdrop_basename($uri, $suffix = NULL) {
 }
 
 /**
- * Creates a directory, optionally creating missing components in the path to
- * the directory.
+ * Creates a directory; optionally creating any missing components in the path.
  *
  * When PHP's mkdir() creates a directory, the requested mode is affected by the
  * process's umask. This function overrides the umask, and sets the mode
--- a/core/includes/file.inc
+++ b/core/includes/file.inc
@@ -2604,7 +2604,7 @@ function backdrop_mkdir($uri, $mode = NULL, $recursive = FALSE, $context = NULL)
  * @see backdrop_mkdir()
  */
 function _backdrop_mkdir_call($uri, $mode, $recursive, $context) {
-  if (is_null($context)) 
+  if (is_null($context)) {
     return mkdir($uri, $mode, $recursive);
   }
   else {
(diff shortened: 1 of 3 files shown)


core/includes/file.inc
```diff
@@ -2526,19 +2526,21 @@ function backdrop_basename($uri, $suffix = NULL) {
 }
 
 /**
- * Creates a directory using Backdrop's default mode.
+ * Creates a directory, optionally creating missing components in the path to
+ * the directory.
```
oadaeh: According to https://api.backdropcms.org/doc-standards#general, this should be shortened. Maybe leave out " to the directory".

core/includes/file.inc
```diff
@@ -2554,7 +2556,55 @@ function backdrop_mkdir($uri, $mode = NULL, $recursive = FALSE, $context = NULL)
...
+}
+
+/**
+ * Helper function. Ensures we don't pass a NULL as a context resource to
+ * mkdir().
+ *
+ * @see backdrop_mkdir()
+ */
+function _backdrop_mkdir_call($uri, $mode, $recursive, $context) {
+  if (is_null($context))
```
oadaeh: There is a missing `{` at the end of this line.

oadaeh (changes requested):
Other than the two things I commented on, this seems fine to me and does what it claims.

core/includes/file.inc
klonos:
```suggestion
  if (is_null($context)) {
```

klonos: Huh!? ...I thought that something like this would trigger some error, or make the tests fail. Anyway, good catch @oadaeh 👍 ...fixed!

core/includes/file.inc
klonos: Thanks 👍 ...fixed!

PS: for my benefit, to which of the bullet points in that documentation are you referring?

klonos:
...ZenCI again 🙄

core/includes/file.inc
oadaeh: At https://api.backdropcms.org/doc-standards#general, the second to last item in the bulleted list.
It's not easy to find the documentation (it took me a few minutes to find it when I was reviewing), but it's been ingrained in head for so long, I just know and recognize it.

klonos: Ah, I get it now ...`must be under 80 characters` ...I was reading that as "less than 80 characters **per line**" (same as regular non-summary documentation rules). So many years I have not noticed this detail. Thanks for pointing that out 👍

PR #2915, opened 2019-10-03
Issue #4104: Automatically generate icons for flexible layout templates

### LINKED ISSUE #4104: Automatically generate icons for flexible layout templates
Part of the #4046 

In #3755 we have used a single icon for all flexible templates. The idea here is that upon saving a flexible layout template (create/update), the amount of rows and columns/regions per row is calculated, and then these numbers are used to generate an icon/thumbnail to be used the layout template listing in `/admin/structure/layouts/settings`.

---
PR by @klonos: https://github.com/backdrop/backdrop/pull/2915

klonos:
Fixes https://github.com/backdrop/backdrop-issues/issues/4104


klonos:
...refreshing sandbox

klonos:
...reopening

core/modules/layout/layout.module
```diff
@@ -2035,7 +2035,7 @@ function layout_layout_info() {
       'flexible' => TRUE,
       'type' => 'layout',
       'path' => 'templates',
-      'preview' => 'flexible_template.png',
+      'preview' => layout_template_generate_thumbnail($flexible_template),
```
quicksketch, line 2038: Rather than calling this function that generates an image in layout_layout_info(), it would be preferable to do this in the submit handler. It also gives you an opportunity to delete the old image, or generate a new file name.

core/modules/layout/layout.theme.inc
```diff
@@ -127,9 +127,17 @@ function theme_layout_template_option($variables) {
 function theme_layout_template_info($variables) {
   $template_info = $variables['template_info'];
 
-  // Build template thumbnail preview
-  $preview_path = $template_info['path'] . '/' . $template_info['preview'];
-  $preview_img = theme('image', array('path' => $preview_path));
+  // Build template thumbnail preview.
+  if (!empty($template_info['flexible'])) {
```
quicksketch, line 131: Rather than checking if it's a flexible template, perhaps use `file_valid_uri($template_info['preview'])` to allow using a URI on all templates?

PR #2916, opened 2019-10-03
Issue #4107: Allow cache busting in `theme_image()` (only for managed images)

### LINKED ISSUE #4107: Allow cache busting string in `theme_image()`
This is the respective issue for https://www.drupal.org/node/2313539, which was originally filed against D7 core; currently against D8 core.

The problem came up while working in #4104, where thumbnail images generated for the flexible layout templates would not refresh, unless you did a hard-refresh on your browser (CRTL+F5 / CMD+SHIFT+R).

> **Problem/Motivation**
>
> When working on a module that uses File Entity to handle media assets we've discovered that currently, there is no way to bypass browser file cache in `theme_image()` and `theme_image_style()`. File Entity provides users with a way to replace their file with a new one, without changing the name of the file. Because filename remains the same, it's still cached in browser and will not show up until clearing page cache with CTRL+R.
>
> This might be fine and we're used to it on the user facing site, but it might confuse authors who just replaced their image with a new one, and it's not showing the latest image in admin interface. It would be nice to have a way to bust browser cache that works with Drupal's theming functions, so we don't reinvent the wheel.

PS: also mentioned in https://www.drupal.org/project/file_entity/issues/1701924

**Update**: It turns out that the D7/8 solution only adds timestamps to managed files, whereas the same problem where the browser cache may serve stale versions of images can occur with unmanaged files too. This is common with files used in the admin UI, and it's the situation in #4104; it has also been reported as a problem by contrib modules (https://www.drupal.org/project/drupal/issues/2800731 for example).

---

- PR by @klonos (adds a timestamp only to managed images): https://github.com/backdrop/backdrop/pull/2916
- Alternative PR by @klonos (adds timestamp to both managed and unmanaged images): https://github.com/backdrop/backdrop/pull/2920

klonos:
Fixes https://github.com/backdrop/backdrop-issues/issues/4107

See: https://www.drupal.org/node/2313539
This is a backport of the respective D8 patch in https://www.drupal.org/files/issues/2019-09-10/2313539-allow-cache-busting-string-21.patch


core/includes/theme.inc
```diff
@@ -1940,6 +1940,13 @@ function template_preprocess_image(&$variables) {
   $path_or_uri = $variables['uri'] ? $variables['uri'] : $variables['path'];
   $variables['attributes']['src'] = file_create_url($path_or_uri);
 
+  if (isset($variables['timestamp'])) {
+    $variables['attributes']['src'] = $variables['attributes']['src'] . (strpos($variables['attributes']['src'], '?') !== FALSE ? '&' : '?') . backdrop_http_build_query(array('timestamp' => $variables['timestamp']));
+  }
+  else {
+    $variables['attributes']['src'] = file_create_url($variables['uri']);
```
indigoxela: This fails for images provided via path (from modules or core), hence feed icon is broken (and many other images).

Using `file_create_url($path_or_uri);` fixes it. And fixes at least some of the broken tests.

klonos: Thanks @indigoxela 👍 ...one step closer 🙂

core/includes/theme.inc
```diff
@@ -1940,6 +1940,13 @@ function template_preprocess_image(&$variables) {
   $path_or_uri = $variables['uri'] ? $variables['uri'] : $variables['path'];
   $variables['attributes']['src'] = file_create_url($path_or_uri);
 
+  if (isset($variables['timestamp'])) {
+    $variables['attributes']['src'] = $variables['attributes']['src'] . (strpos($variables['attributes']['src'], '?') !== FALSE ? '&' : '?') . backdrop_http_build_query(array('timestamp' => $variables['timestamp']));
+  }
+  else {
+    $variables['attributes']['src'] = file_create_url($path_or_uri);
```
indigoxela: Hm... isn't that _else_ redundant? The same value gets set in line 1941.

klonos: Was just looking at that.

klonos: Fixed. Waiting on tests...

bugfolder:
@klonos, can you edit your first comment to add the magic word "Fixes", which will bind this PR to the issue?

PR #2918, opened 2019-10-03
Issue #3833: Do not make entries in tempstore when viewing forms.

### LINKED ISSUE #3833: Do not make entries in "cache_form" when viewing forms that use #ajax['callback']

quicksketch:
May fix https://github.com/backdrop/backdrop-issues/issues/3833.

### DIFF
--- a/core/includes/ajax.inc
+++ b/core/includes/ajax.inc
@@ -693,7 +693,10 @@ function ajax_footer() {
  */
 function ajax_process_form($element, &$form_state) {
   $element = ajax_pre_render_element($element);
-  if (!empty($element['#ajax_processed']) && !isset($element['#ajax']['no_form_cache'])) {
+  // Custom callback paths set in $element['#ajax']['path'] are assumed to
+  // require form caching. Using $element['#ajax']['callback'] on the other hand
+  // does not require the form to be cached, and is the preferred method.
+  if (!empty($element['#ajax_processed']) && !isset($element['#ajax']['callback'])) {
     $form_state['cache'] = TRUE;
   }
   return $element;
--- a/core/includes/ajax.inc
+++ b/core/includes/ajax.inc
@@ -696,7 +696,8 @@ function ajax_process_form($element, &$form_state) {
   // Custom callback paths set in $element['#ajax']['path'] are assumed to
   // require form caching. Using $element['#ajax']['callback'] on the other hand
   // does not require the form to be cached, and is the preferred method.
-  if (!empty($element['#ajax_processed']) && !isset($element['#ajax']['callback'])) {
+  $has_ajax_path = isset($element['#ajax']['path']) && $element['#ajax']['path'] !== $_GET['q'];
+  if (!empty($element['#ajax_processed']) && $has_ajax_path) {
     $form_state['cache'] = TRUE;
   }
   return $element;
--- a/core/includes/ajax.inc
+++ b/core/includes/ajax.inc
@@ -382,7 +382,16 @@ function ajax_get_form() {
 function ajax_form_callback() {
   list($form, $form_state, $form_id, $form_build_id, $commands) = ajax_get_form();
   backdrop_process_form($form['#form_id'], $form, $form_state);
+  return ajax_form_return_commands($form, $form_state, $commands);
+}
 
+/**
+ * Return a set of AJAX commands for updating a form.
+ *
+ * @see ajax_form_callback()
+ * @see backdrop_get_form()
+ */
+function ajax_form_return_commands($form, $form_state, $commands = array()) {
   // We need to return the part of the form (or some other content) that needs
   // to be re-rendered so the browser can update the page with changed content.
   // Since this is the generic menu callback used by many Ajax elements, it is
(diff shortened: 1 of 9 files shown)


core/includes/ajax.inc
```diff
@@ -792,7 +805,7 @@ function ajax_pre_render_element($element) {
     // #ajax['callback'] for determining which part of the form needs
     // re-rendering. For that, we have a special 'system/ajax' route.
```
oadaeh, line 806: I think this comment might need to be updated.

core/modules/simpletest/backdrop_web_test_case.php
```diff
@@ -2222,18 +2222,29 @@ protected function backdropGetAJAX($path, array $options = array(), array $heade
    *   the form, which is typically the same thing but with hyphens replacing
    *   the underscores.
    * @param $extra_post
-   *   (optional) A string of additional data to append to the POST submission.
+   *   (optional) An array of additional data to append to the POST submission.
```
oadaeh, line 2225: Should this also mention it handles string values for backwards compatibility?

core/modules/views_ui/views_ui.admin.inc
```diff
@@ -609,8 +609,12 @@ function views_ui_add_ajax_trigger(&$wrapping_element, $trigger_key, $refresh_pa
   $seen_ids = &backdrop_static(__FUNCTION__ . ':seen_ids', array());
   $seen_buttons = &backdrop_static(__FUNCTION__ . ':seen_buttons', array());
 
-  // Add the AJAX behavior to the triggering element.
+  // Add the AJAX behavior to the triggering element. We specifically set
+  // 'system/ajax' as the path to use the legacy way of building the form.
+  // @todo: Remove the 'system/ajax' path to use the new rebuilding approach
+  // introduced in https://github.com/backdrop/backdrop-issues/issues/3833.
```
oadaeh, line 615: What marker is in place to address these things in the future?

oadaeh (commented):
I reviewed the code changes and I did not see anything that seemed obviously questionable.
I did add a few notes where I think comments might need updating.
I have not tested the functionality, but I will attempt to do that next.

core/includes/form.inc
```diff
@@ -128,7 +128,36 @@ function backdrop_get_form($form_id) {
   array_shift($args);
   $form_state['build_info']['args'] = $args;
 
-  return backdrop_build_form($form_id, $form_state);
+  // Disable redirection if the request was made via AJAX.
+  $ajax_post = isset($_POST['ajax_page_state']) && $_POST['form_id'] === $form_id;
```
oadaeh: Okay, `$_POST['form_id']` needs an isset or empty check.
When managing blocks on a layout, if you click on any "Configure" link or click on any item after clicking any "Add block" link, a watchdog notice similar to this is generated:
`Notice: Undefined index: form_id in backdrop_get_form() (line 132 of /home/qa/github/backdrop/backdrop/pr/2918/core/includes/form.inc).`
The form appears to work as expected otherwise.

oadaeh (changes requested):
I found a problem.

oadaeh:
The failed test appears to be unrelated to this:

```
Fail      Other      update.test        198 UpdateCoreTestCase->testModulePageS
    "Checked available update data for one project." found
```

core/includes/form.inc
oadaeh: I found a place where this error does affect the behavior of Backdrop:
1. Navigate to the Install modules page at /admin/modules/install
2. Click the name of a module
3. Click the Add to Installation queue link

The module does not get added to the queue.

PR #2920, opened 2019-10-05
Issue  #4107: Allow cache busting string in `theme_image()` (for both managed and unmanaged images)

### LINKED ISSUE #4107: Allow cache busting string in `theme_image()`

klonos:
Fixes https://github.com/backdrop/backdrop-issues/issues/4107

Alternative to https://github.com/backdrop/backdrop/pull/2916 ...adds a timestamp to unmanaged images as well (as opposed to https://github.com/backdrop/backdrop/pull/2916 which only does that for managed images).


core/includes/theme.inc
```diff
@@ -1938,8 +1938,21 @@ function theme_dropbutton_wrapper($variables) {
 function template_preprocess_image(&$variables) {
   // Accept either URI or path key. But both are handled by file_create_url().
   $path_or_uri = $variables['uri'] ? $variables['uri'] : $variables['path'];
+  $image_real_path = backdrop_realpath($path_or_uri);
   $variables['attributes']['src'] = file_create_url($path_or_uri);
 
+  // If the image is an unmanaged file, it will not have an entry in the
+  // file_managed table, hence no timestanp attribute either.
```
```suggestion
  // file_managed table, hence no timestamp attribute either.
```

core/modules/layout/layout.module
```diff
@@ -2055,6 +2055,100 @@ function layout_layout_info() {
   return $layouts;
 }
 
+/**
+ * Generate an image to be used as a layout template thumbnail.
```
quicksketch: This should be in a separate PR. I think this may be part of https://github.com/backdrop/backdrop-issues/issues/4104 ?

core/includes/theme.inc
klonos: Thanks 👍 😅

core/modules/layout/layout.module
klonos: You're right; I was merging changes from both issues, so to file combined PRs. Fixed.

PR #2927, opened 2019-10-08
Issue #74 add entity cache to core

### LINKED ISSUE #74: [D8][PS] Integrate the functionality of the entity_cache module into core
...but not as a module... just as part of Backdrop, maybe part of the entity system? 

---

Advocate for this issue @hosef

---

Weekly update?

---

PR by @hosef: https://github.com/backdrop/backdrop/pull/2927
Change record: https://api.backdropcms.org/change-records/entity-caching-added-to-core

hosef:
Fixes https://github.com/backdrop/backdrop-issues/issues/74

This PR implements the core functionality of the Entity Cache module from D7 by integrating it directly into `DefaultEntityController`. I have tried to make the new functionality as transparent as possible to the rest of Backdrop or any contributed modules.

I did not port any of the tests because they just extended core tests from modules that added entities with the entitycache module enabled. By integrating the functionality into `DefaultEntityController` we have effectively the same result.

### DIFF
--- a/core/modules/entity/entity.controller.inc
+++ b/core/modules/entity/entity.controller.inc
@@ -159,12 +159,14 @@ class DefaultEntityController implements EntityControllerInterface {
    */
   public function resetCache(array $ids = NULL) {
     // Reset the persistent cache.
-    if (!empty($ids)) {
-      cache_clear_all($ids, 'cache_entity_' . $this->entityType);
-    }
-    else {
-      // Force all cached entries to be deleted.
-      cache_clear_all('*', 'cache_entity_' . $this->entityType, TRUE);
+    if (isset($this->entityInfo['entity cache']) && $this->entityInfo['entity cache']) {
+      if (!empty($ids)) {
+        cache_clear_all($ids, 'cache_entity_' . $this->entityType);
+      }
+      else {
+        // Force all cached entries to be deleted.
+        cache_clear_all('*', 'cache_entity_' . $this->entityType, TRUE);
+      }
     }
 
     $this->resetStaticCache($ids);
@@ -251,7 +253,7 @@ class DefaultEntityController implements EntityControllerInterface {
     }
 
     // Try to load entities from the persistent cache.
-    if ($this->entityInfo['entity cache'] && !$revision_id && $ids && !$conditions) {
+    if (isset($this->entityInfo['entity cache']) && $this->entityInfo['entity cache'] && !$revision_id && $ids && !$conditions) {
       $cached_entities = array();
       if ($ids && !$conditions) {
         $cached = cache_get_multiple($ids, 'cache_entity_' . $this->entityType);
@@ -301,7 +303,7 @@ class DefaultEntityController implements EntityControllerInterface {
     }
 
     // Add entities to the entity cache if we are not loading a revision.
-    if ($this->entityInfo['entity cache'] && !empty($queried_entities) && !$revision_id) {
+    if (isset($this->entityInfo['entity cache']) && $this->entityInfo['entity cache'] && !empty($queried_entities) && !$revision_id) {
       // Only cache the entities which were loaded by ID. Entities that were
       // loaded based on conditions will never be found via cacheGet() and we
       // would keep on caching them.
--- a/core/modules/entity/entity.controller.inc
+++ b/core/modules/entity/entity.controller.inc
@@ -28,6 +28,15 @@ interface EntityControllerInterface {
   public function resetCache(array $ids = NULL);
 
   /**
+   * Resets the internal, static entity cache.
+   *
+   * @param string[] $ids
+   *   (optional) If specified, the cache is reset for the entities with the
+   *   given ids only.
+   */
+  public function resetStaticCache(array $ids = NULL);
+
+  /**
    * Loads one or more entities.
    *
    * @param $ids
@@ -149,6 +158,22 @@ class DefaultEntityController implements EntityControllerInterface {
    * Implements EntityControllerInterface::resetCache().
    */
   public function resetCache(array $ids = NULL) {
+    // Reset the persistent cache.
+    if (!empty($ids)) {
+      cache_clear_all($ids, 'cache_entity_' . $this->entityType);
+    }
+    else {
+      // Force all cached entries to be deleted.
+      cache_clear_all('*', 'cache_entity_' . $this->entityType, TRUE);
+    }
+
+    $this->resetStaticCache($ids);
+  }
+
+  /**
+   * Implements EntityControllerInterface::resetStaticCache().
+   */
+  public function resetStaticCache(array $ids = NULL) {
     if (isset($ids)) {
       foreach ($ids as $id) {
         unset($this->entityCache[$id]);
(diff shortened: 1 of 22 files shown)


core/modules/file/file.module
```diff
@@ -461,7 +461,7 @@ function file_entity_info() {
     );
   }
 
-  return array(
+  $return = array(
```
klonos: Can we give a more meaningful name to this variable?

hosef: We can. I was just trying to keep it consistent with all of the other implementations of hook_entity_info in core.

hosef: @klonos Do you think I should change that for all 5 core entities, or just the new instances from this PR?

klonos: I would prefer all of them please. We need to keep in mind that DX should be improved for novice devs too; and having meaningful, self-explanatory names for variables helps to that end.

docwilmot: To me, this is an info hook, so all it does is return info. It seems appropriate that the variable holding the return be called `$return` I think.

klonos: To me, what is returned here (as the name of the function implies) is information; so I'd expect the variable to be `$info`. The keyword `return` covers the "return" bit.

As a novice developer, it'd make more sense to me to see `return $info;` or `return $entity_info`, rather than `return $return`.

My 2c

core/modules/entity/entity.controller.inc
```diff
@@ -27,6 +27,15 @@ interface EntityControllerInterface {
    */
   public function resetCache(array $ids = NULL);
 
+  /**
+   * Resets the internal, static entity cache.
+   *
+   * @param string[] $ids
```
docwilmot: Not sure this is Backdrop standard for denoting arrays.

core/modules/entity/entity.controller.inc
```diff
@@ -108,14 +117,23 @@ class DefaultEntityController implements EntityControllerInterface {
...
   protected $revisionTable;
 
+  /**
+   * Whether this entity type should use the persistent entity cache.
+   *
+   * Set by entity info.
+   *
+   * @var boolean
+   */
+  protected $persistanctCache;
```
docwilmot: This should be `persistentCache`. Five more instances need changing in this PR.

core/modules/entity/entity.controller.inc
```diff
@@ -142,14 +160,42 @@ class DefaultEntityController implements EntityControllerInterface {
...
 
     // Check if the entity type supports static caching of loaded entities.
-    $this->cache = !empty($this->entityInfo['static cache']);
+    $this->staticCache = !empty($this->entityInfo['static cache']);
+
+    // Check if the entity type supports usage of the persistant entity cache.
+    if (isset($this->entityInfo['entity cache'])
+      && $this->entityInfo['entity cache']
+      && db_table_exists('cache_entity_' . $this->entityType)) {
+      $this->persistanctCache = TRUE;
```
docwilmot: This should be `persistentCache`.

core/modules/entity/entity.controller.inc
```diff
@@ -142,14 +160,42 @@ class DefaultEntityController implements EntityControllerInterface {
...
+    }
+    else {
+      $this->persistanctCache = FALSE;
```
docwilmot: This should be `persistentCache`.

core/modules/entity/entity.controller.inc
```diff
@@ -142,14 +160,42 @@ class DefaultEntityController implements EntityControllerInterface {
...
+    }
   }
 
   /**
    * Implements EntityControllerInterface::resetCache().
    */
   public function resetCache(array $ids = NULL) {
-    if (isset($ids)) {
+    // Reset the persistent cache.
+    if ($this->persistanctCache) {
```
docwilmot: This should be `persistentCache`.

core/modules/entity/entity.controller.inc
```diff
@@ -181,16 +227,71 @@ class DefaultEntityController implements EntityControllerInterface {
...
+    if ($this->staticCache && !$revision_id) {
       $entities += $this->cacheGet($ids, $conditions);
       // If any entities were loaded, remove them from the ids still to load.
       if ($passed_ids) {
         $ids = array_keys(array_diff_key($passed_ids, $entities));
       }
     }
 
+    // Try to load entities from the persistent cache.
+    if ($this->persistanctCache && !$revision_id && $ids && !$conditions) {
```
docwilmot: This should be `persistentCache`.

core/modules/entity/entity.controller.inc
```diff
@@ -215,7 +316,22 @@ class DefaultEntityController implements EntityControllerInterface {
       $entities += $queried_entities;
     }
 
-    if ($this->cache) {
+    // Add entities to the entity cache if we are not loading a revision.
+    if ($this->persistanctCache && !empty($queried_entities) && !$revision_id) {
```
docwilmot: This should be `persistentCache`.

core/modules/entity/entity.controller.inc
```diff
@@ -142,14 +160,42 @@ class DefaultEntityController implements EntityControllerInterface {
     }
 
     // Check if the entity type supports static caching of loaded entities.
-    $this->cache = !empty($this->entityInfo['static cache']);
+    $this->staticCache = !empty($this->entityInfo['static cache']);
+
+    // Check if the entity type supports usage of the persistant entity cache.
```
docwilmot: This should be `persistent`.

core/modules/entity/entity.controller.inc
```diff
@@ -142,14 +160,42 @@ class DefaultEntityController implements EntityControllerInterface {
...
+    }
+
+    $this->resetStaticCache($ids);
+  }
+
+  /**
+   * Implements EntityControllerInterface::resetStaticCache().
+   */
+  public function resetStaticCache(array $ids = NULL) {
+    if (isset($ids) && !empty($ids)) {
```
docwilmot: Minor, but trying to think of why `!empty()` wouldnt suffice, without the `isset`.

core/modules/entity/entity.controller.inc
```diff
@@ -181,16 +227,71 @@ class DefaultEntityController implements EntityControllerInterface {
     // and we need to know if it's empty for this reason to avoid querying the
     // database when all requested entities are loaded from cache.
     $passed_ids = !empty($ids) ? array_flip($ids) : FALSE;
+
+    // Use an entity field query to transform the list of conditions into
+    // the set of entity IDs which the conditions admit, then re-enter this
+    // method with that set as the $ids constraint.
+    if ($conditions && !$revision_id) {
```
docwilmot, line 235: Noting that this PR will result in two DB queries now, if `$conditions` is set: this one to sort out the conditions and the next one, starting on line 295, to load the entities. But we have also made it clear that using `$conditions` to load entities is deprecated, so this would not be a problem for any core usage of `Entity::load()`, just any contrib that isnt paying attention.

core/modules/entity/entity.controller.inc
quicksketch: Hm, yeah we do this when returning an array of objects sometimes, e.g. `Node[]` or `File[]` indicates an array of nodes or files. I haven't seen this for basic types like strings though. Still it seems as though it may be more informative than just saying `array`.

core/modules/entity/entity.controller.inc
quicksketch: I could see this being helpful in the event you were going to call resetStaticCache() with a list of IDs but you weren't sure if there was going to be something in the array. For example this may or may not have values in the array:

```
$this->resetStaticCache($unknown_array);
```

If it contained one ID, it would clear one item from the cache. If it was an empty array, it shouldn't clear anything at all. Versus if NULL is passed in or the parameter omitted, everything is cleared.

quicksketch: I thought the current code was good initially but after writing the above, it seems like only `isset()` should be used, and remove the `!empty()`. Because an empty array shouldn't clear anything. Only omitting the array entirely or passing in NULL should clear the entire cache.

core/modules/entity/entity.controller.inc
quicksketch: For now I just set this to `array` to match the rest of core.

quicksketch:
I updated this PR based on @docwilmot's feedback.

quicksketch:
Thanks @hosef! Merged into 1.x for 1.16.0!

PR #1849, opened 2017-04-28
Issue #2637: Transliteration doesn't respect site default language

olafgrabienski:
Fixes https://github.com/backdrop/backdrop-issues/issues/2637.

PR #2931, opened 2019-10-12
Issue #1849: [UX] Admin bar: remove empty `class=""` attribute when removing the `active-search-item` class

klonos:
https://github.com/backdrop/backdrop-issues/issues/1849

### DIFF
--- a/core/modules/admin_bar/js/admin_bar.js
+++ b/core/modules/admin_bar/js/admin_bar.js
@@ -431,13 +431,13 @@ Backdrop.adminBar.behaviors.search = function (context, settings, $adminBar) {
     $this.trigger(show ? 'showPath' : 'hidePath', [this]);
   }
 
-  /**	
-   * Closes the search results and clears the search input.	
-   */	
-  function resultsClickHandler(e, link) {	
-    var $original = $(this).data('original-link');	
-    $original.trigger('mouseleave');	
-    $input.val('').trigger('keyup');	
+  /**
+   * Closes the search results and clears the search input.
+   */
+  function resultsClickHandler(e, link) {
+    var $original = $(this).data('original-link');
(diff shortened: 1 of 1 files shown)


core/modules/admin_bar/js/admin_bar.js
```diff
@@ -412,11 +412,21 @@ Backdrop.adminBar.behaviors.search = function (context, settings, $adminBar) {
...
     }
     if (show) {
-      $adminBar.find('.active-search-item').removeClass('active-search-item');
+      if ($adminBar.find('.active-search-item').classList.length === 1) {
+        $adminBar.find('.active-search-item').removeAttr('class');
+      }
+      else {
+        $adminBar.find('.active-search-item').removeClass('active-search-item');
+      }
       $(this).addClass('active-search-item');
```
indigoxela, line 421: Where does this "classList" in line 415 come from?

Another approach: after `removeClass` check for an empty attribute and remove then:

`$('*[class=""]').removeAttr('class');`

Also untested... ;)

klonos: Sorry for the late response @indigoxela 😓 

Here's some info on `classList`:

https://developer.mozilla.org/en-US/docs/Web/API/Element/classList
https://www.w3schools.com/jsref/prop_element_classlist.asp

klonos: ...this basically searches for the element with a class of `.active-search-item`, and if that is the only class, it removes the entire `class=` attribute instead of only the class.

indigoxela: Ah, got it, thanks for the hint.

klonos:
Closing this one, since https://github.com/backdrop/backdrop/pull/2932 has been merged.

PR #2932, opened 2019-10-12
Issue #1849: Remove empty class attributes from Admin Bar.

### LINKED ISSUE #1849: [UX] Admin bar: empty `class=""` attribute left behind when the `active-search-item` class is removed.
Notice how there is a `class=""` left behind while the `active-search-item` class is added to the currently hovered-over item:

backdrop-admin_bar-empty_left-over-class

---
~~PR by @klonos: https://github.com/backdrop/backdrop/pull/2931~~
PR by @BWPanda: https://github.com/backdrop/backdrop/pull/2932

ghost:
Fixes https://github.com/backdrop/backdrop-issues/issues/1849


klonos (approved):
LGTM 👍

klonos (approved):
LGTM 👍

core/modules/admin_bar/js/admin_bar.js
```diff
@@ -418,6 +418,7 @@ Backdrop.adminBar.behaviors.search = function (context, settings, $adminBar) {
     else {
       $(this).removeClass('active-search-item');
```
quicksketch, line 415:
```suggestion
      $this.removeClass('active-search-item');
```

quicksketch: The same change could be made on line 416, GitHub won't let me comment on that line though.

core/modules/admin_bar/js/admin_bar.js
```diff
@@ -418,6 +418,7 @@ Backdrop.adminBar.behaviors.search = function (context, settings, $adminBar) {
...
     }
+    $adminBar.find('[class=""]').removeAttr('class');
```
quicksketch: Instead of searching the entire admin bar, we could just check the affected item after a class is removed. Something like this:

```
$this.removeClass('active-search-item');
if ($this.is('[class=""]')) {
  $this.removeAttr('class');
}
```

ghost: Tried that exact cde, but it didn't work. I think because there's another place just above that removes that class as well. See updated PR for my solution that seems to work, and hopefully isn't too bad re. performance.

core/modules/admin_bar/js/admin_bar.js
```diff
@@ -411,12 +411,14 @@ Backdrop.adminBar.behaviors.search = function (context, settings, $adminBar) {
...
     }
-    if (show) {
-      $adminBar.find('.active-search-item').removeClass('active-search-item');
-      $(this).addClass('active-search-item');
-    }
-    else {
+    $adminBar.find('.active-search-item').each(function () {
       $(this).removeClass('active-search-item');
+      if ($(this).is('[class=""]')) {
+        $(this).removeAttr('class');
```
klonos, line 417: I guess what @quicksketch meant was this:

```suggestion
      $this.removeClass('active-search-item');
      if ($this.is('[class=""]')) {
        $this.removeAttr('class');
```

ghost: Yes, however the code changed since then. You can't use that here since I added a new function, so now `$this != $(this)`.

quicksketch:
Thanks @BWPanda and @klonos! Merged into 1.x and 1.17.x.

PR #2939, opened 2019-10-18
4144 link field order ux

### LINKED ISSUE #4144: [UX] Link text field should come first in ckeditor link dialog
Three times in a row I just opened the link dialog, typed in some text, then a url, then hit save. Only to be shown a validation error that I got the two fields backwards. Does anyone else have this happen?

The problem:
Screen Shot 2019-10-17 at 3 35 03 PM

The solution:
Screen Shot 2019-10-17 at 3 38 12 PM

jenlampton:
Fixes https://github.com/backdrop/backdrop-issues/issues/4144

### DIFF
--- a/core/modules/views/plugins/views_plugin_display_page.inc
+++ b/core/modules/views/plugins/views_plugin_display_page.inc
@@ -264,7 +264,7 @@ class views_plugin_display_page extends views_plugin_display {
     $options['path'] = array(
       'category' => 'page',
       'title' => t('Path'),
-      'value' => views_ui_truncate($path, 24),
+      'value' => $path,
     );
 
     $menu = $this->get_option('menu');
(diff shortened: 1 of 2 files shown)


core/modules/views/plugins/views_plugin_display_page.inc
```diff
@@ -264,7 +264,7 @@ class views_plugin_display_page extends views_plugin_display {
     $options['path'] = array(
       'category' => 'page',
       'title' => t('Path'),
-      'value' => views_ui_truncate($path, 24),
+      'value' => $path,
```
klonos, line 267: This seems to have snuck in from https://github.com/backdrop/backdrop/pull/2940

Otherwise, this LGTM 👍

PR #2940, opened 2019-10-18
2843 longer urls

### LINKED ISSUE #2843: CKEditor: Long link URLs being truncated
## Issue
Editors using a basic subset of ckeditor tools are finding that when they paste long query style URL's into the modal box that is presented by the link function, that URL is truncated, which breaks the link.  Shorter links work as expected.

## Expected behavior
The link function accepts any valid link.

## Observed Workaround
Switching to HTML Editor allows the complete URL to be entered into the code and in view it works as expected.  Since exposing HTML Editor can be a security issue, it is preferred that the WYSIWYG editor work as expected.

##  Comments
Seems that input filtering controls this either in core or at the plugin level.  Would be nice if this was exposed as an admin setting so that it can be en/dis/abled or the value that controls the length of the truncate can be modified.

---

~PR by @alexfinnarn : https://github.com/backdrop/backdrop/pull/2032~ merged
PR by @jenlampton: https://github.com/backdrop/backdrop/pull/2940

jenlampton:
Fixes https://github.com/backdrop/backdrop-issues/issues/2843

### DIFF
--- a/core/modules/filter/filter.pages.inc
+++ b/core/modules/filter/filter.pages.inc
@@ -234,6 +234,7 @@ function filter_format_editor_image_form($form, &$form_state, $format) {
       'data-editor-image-toggle' => t('Select from library'),
     ),
     '#weight' => -8,
+    '#maxlength' => 256,
   );
 
   // If no current value or an existing file exists, default to showing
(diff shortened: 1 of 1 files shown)


klonos (approved):
LGTM 👍

core/modules/views/plugins/views_plugin_display_page.inc
```diff
@@ -264,7 +264,7 @@ class views_plugin_display_page extends views_plugin_display {
     $options['path'] = array(
       'category' => 'page',
       'title' => t('Path'),
-      'value' => views_ui_truncate($path, 24),
+      'value' => $path,
```
quicksketch: This change seems unrelated. And it's a display-only presentation, not a field. Should this be reverted?

quicksketch:
Merged into 1.x and 1.14.x. Thanks!

PR #2942, opened 2019-10-19
#4145 - Fixing admin/structure/contact/list menu link

### LINKED ISSUE #4145: Contact form categories menu item generates error
**Description of the bug**
If on a fresh install of Backdrop CMS a user enables the contact form module and goes to 'admin/structure/contact/list,' they will get an error message:

```
Warning: Missing argument 1 for contact_category_list() in contact_category_list()
(line 166 of /core/modules/contact/contact.admin.inc).
Warning: Missing argument 2 for contact_category_list() in contact_category_list()
(line 166 of /core/modules/contact/contact.admin.inc).
```

**Steps To Reproduce**

To reproduce the behavior:
1. Spin up fresh installation of Backdrop CMS
2. Enable the contact form module
3. Go to Structure > Contact form > Categories
4. Observe error message

"Too few arguments to function contact_category_list(), 0 passed in /var/lib/tugboat/core/modules/system/system.module on line 2464 and exactly 2 expected"

**Additional information**

The problem seems to be that the form is trying to take the user to: /admin/structure/contact/list
I think that the user is just trying to get here: /admin/structure/contact/

I'm not sure if the original contact/list page exists or even needs to exist. 

I'm guessing this is an easy fix to this file: core/modules/contact/contact.module

contact_module_-_backdrop-starter_-____Sites_lando_backdrop-starter_

I thought about it some more and wondered if we need that link at all. Because, if the user clicks on the parent menu item "Contact form" they are taken to the categories page anyway.

Contact_form___Fish

But, then I see that all the top level items under Structure have a sub-menu item that is "list," when the top menu item already takes you there. So, I guess we need to fix this Category link to be consistent. 

**We should probably also name it "List categories" to be consistent with the other menu items.**

- - - - - - - - -

PR by @stpaultim - https://github.com/backdrop/backdrop/pull/2942

stpaultim:
Fixes https://github.com/backdrop/backdrop-issues/issues/4145

### DIFF
--- a/core/modules/contact/contact.module
+++ b/core/modules/contact/contact.module
@@ -85,12 +85,8 @@ function contact_menu() {
     'file' => 'contact.admin.inc',
   );
   $items['admin/structure/contact/list'] = array(
-    'title' => 'Categories',
-    'page callback' => 'contact_category_list',
-    'access arguments' => array('administer contact forms'),
+    'title' => 'List categories',
     'type' => MENU_DEFAULT_LOCAL_TASK,
-    'weight' => 1,
-    'file' => 'contact.admin.inc',
   );
   $items['admin/structure/contact/settings'] = array(
     'title' => 'Settings',


quicksketch:

PR #2944, opened 2019-10-21
Issue #4149: Remove duplication in `install.sh`.

### LINKED ISSUE #4149: Remove duplication in `install.sh`
[`install.sh`](https://github.com/backdrop/backdrop/blob/1.x/core/scripts/install.sh) has a duplicated `site-name` [array key](https://github.com/backdrop/backdrop/blob/1.x/core/scripts/install.sh#L51).

We should remove it.

ghost:
Fixes https://github.com/backdrop/backdrop-issues/issues/4149


ghost:
Merged into 1.x and 1.17.x.

PR #2946, opened 2019-10-22
Issue #4151: Support multisites in install.sh.

### LINKED ISSUE #4151: Support multisites in `install.sh`
**Description of the need**
If you want to install Backdrop from the command line using the core `install.sh` script, you can only do a regular install. You can't do a multisite install.

Setting up a Backdrop installation for multisite, and then accessing the GUI installer works fine - the `/sites/[SITE]/settings.php` is updated with the DB info, config directories, etc.
But setting up the same Backdrop multisite installation, and then accessing the command line installer, doesn't work - the core `/settings.php` is updated with the DB info, config directories, etc., not the site-specific one.

**Proposed solution**
The command line installer should work the same as the GUI one; however it is that the GUI one knows to check for site-specific directories, the command line one should do the same.

For now multisites have to be setup via the GUI, but that's not always possible (e.g. setting up a non-interactive test environment).

**Additional information**
To test, here's a Lando file for setting up the multisite environment:

```yaml
name: multisite
recipe: backdrop
config:
  webroot: www
services:
  appserver:
    build:
      - git clone https://github.com/backdrop/backdrop.git /app/www
      - mkdir /app/www/sites/multisite && cp /app/www/settings.php /app/www/sites/multisite/
```
I couldn't do this as part of the Lando setup, so one you run `lando start` you have to manually add this to the `/sites/sites.php` file:
`$sites['multisite.lndo.site'] = 'multisite';`

---

PR by @BWPanda: https://github.com/backdrop/backdrop/pull/2946

ghost:
Fixes https://github.com/backdrop/backdrop-issues/issues/4151

### DIFF
--- a/core/includes/bootstrap.inc
+++ b/core/includes/bootstrap.inc
@@ -611,6 +611,7 @@ function find_conf_path($http_host, $script_name, $require_settings = TRUE) {
 
   if ($sites) {
     $uri = explode('/', $script_name);
+    $uri[0] = ($uri[0] == '.') ? '' : $uri[0];
     $server = explode('.', implode('.', array_reverse(explode(':', rtrim($http_host, '.')))));
     for ($i = count($uri) - 1; $i > 0; $i--) {
       for ($j = count($server); $j > 0; $j--) {
--- a/core/scripts/install.sh
+++ b/core/scripts/install.sh
@@ -102,7 +102,19 @@ unset($options['root']);
 
 // Set the site URL.
 if ($options['url']) {
-  $_SERVER['HTTP_HOST'] = $options['url'];
+  $url_parts = parse_url($options['url']);
+  if (!empty($url_parts['host'])) {
+    // E.g.: 'http://example.com/'
+    $_SERVER['HTTP_HOST'] = $url_parts['host'];
+  }
+  elseif (!empty($url_parts['path'])) {
+    // E.g.: 'example.com'
+    $_SERVER['HTTP_HOST'] = $url_parts['path'];
+  }
+  else {
+    print "--url option is invalid. Specify the URL of your site as --url=http://example.com/ or --url=example.com.\n";
+    exit;
+  }
 }
 unset($options['url']);
 
--- a/core/scripts/install.sh
+++ b/core/scripts/install.sh
@@ -20,6 +20,7 @@ key=value...                              Any additional settings you wish to pa
 
 Options:
 --root                                    Set the working directory for the script to the specified path. Required if running this script from a directory other than the Backdrop root.
+--url                                     Set the URL of the site to install, as defined in sites.php. Required for multisite installations.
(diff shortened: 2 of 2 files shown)


core/scripts/install.sh
```diff
@@ -98,6 +100,12 @@ if ($options['root'] && is_dir($options['root'])) {
 }
 unset($options['root']);
 
+// Set the site URL.
+if ($options['url']) {
+  $_SERVER['HTTP_HOST'] = $options['url'];
```
quicksketch: Isn't HTTP_HOST the domain name only, without the protocol? So if `$options['url']` was `https://example.com`, I think `HTTP_HOST` would just be `example.com`. Perhaps this should be checked with `parse_url()` to pull that out.

Also, multisite can be configured not just on domains but also sub-directories. Are those handled in the PR as well?

core/includes/bootstrap.inc
```diff
@@ -611,6 +611,7 @@ function find_conf_path($http_host, $script_name, $require_settings = TRUE) {
 
   if ($sites) {
     $uri = explode('/', $script_name);
+    $uri[0] = ($uri[0] == '.') ? '' : $uri[0];
```
quicksketch, line 614: What does this change accomplish?

core/scripts/install.sh
ghost: Good idea. I've updated the code to parse URL strings better.

Re. sub-directories, are you referring to `$sites['8080.www.example.com.mysite.test'] = 'example';`? If so, I've never used this and am not sure how to go about allowing for that scenario...

core/includes/bootstrap.inc
ghost: So the `$uri` variable in the line above has the following array when the UI installer is used:

```
0 => '',
1 => 'core',
2 => 'install.php',
```

When the `install.sh` script is used though, the array is:

```
0 => '.',
1 => 'core',
2 => 'scripts',
3 => 'install.sh',
```

The following `$dir` variable is then set to the following values (during the loop), looking for a matching domain entry in the `sites.php` file...

UI installer:

```
example.com.core
com.core
example.com
com
```

And the `install.sh` script:

```
example.com..core.scripts
com..core.scripts
example.com..core
com..core
example.com.
com.
```

As you can see, the UI installer will eventually find and match `example.com`, whereas the `install.sh` script won't (the closest it gets is `example.com.` (note the end `.`)).

So this line of code addition in the PR basically checks to see if the first `$uri` array item is a single `.` and, if so, replaces it with an empty string. This makes it possible to properly find a matching domain in `sites.php`.

quicksketch: Huh, that seems dumb that directories like "core" or "core/scripts" get added to the search keys in the first place. But your explanation makes sense to get consistent results. And that would be an issue with the sites.php parsing, rather than the install.sh script.

quicksketch:

PR #2947, opened 2019-10-22
Issue #4155: Added some styling from color.admin.inc in the Color mod…

### LINKED ISSUE #4155: Improve the styling of color input elements in the admin theme
**Description of the need**
When a contributed module attempts to add a color input, the current styling in Seven causes it to look very weird. Because of this, any module that has one will need to provide it's own styling of the form.

**Proposed solution**
We can move some of the CSS styles from the Color module to Seven, and that should provide a reasonable base for module developers to use.

- - - - - - - - - - - - 

PR by @hosef - https://github.com/backdrop/backdrop/pull/2947

hosef:
Fixes: https://github.com/backdrop/backdrop-issues/issues/4155

…ule to style.css in Seven.


core/themes/seven/css/style.css
```diff
@@ -1160,6 +1160,27 @@ textarea.form-textarea:focus {
   color: #000;
   border-color: #43afe4;
 }
+input[type=color] {
```
quicksketch: All other selectors here use a class name. Can `input.form-color` be used instead?

PR #2949, opened 2019-10-24
Ref #4147 - Rename system.performance config key to system.core

chellman:
I feel like there must be something more I should do with this, but I renamed the core config key as mentioned in #4147.  I spun up a fresh Backdrop off this, and everything went well.


gitressa:
Thanks @chellman! The issue https://github.com/backdrop/backdrop-issues/issues/4147 is closed, so can this PR also be closed @jenlampton?

jenlampton:
Yes, thank you both!

PR #2950, opened 2019-10-24
#4093 - Basis: Remove the default color change behind the site name

stpaultim:


stpaultim:
```
.header-identity-wrapper {
  /* BG color for slogan, and logo/site name if they don't have a bg */
  background: #000000;
}
```
I deleted this, because I fail to see where it's necessary and it was causing a problem. 

I think it's better to set the background color to transparent on screens greater than 34em, but this default background color wouldn't let me do that. 

I could set the title background to the same thing as the rest of the header, but this means that if a themer overrides the header background, they will also have to override the background for the title. That seems unnecessary. 

Is there any reason for this default background color that I can't see. Under normal circumstances, this is never visible.

jenlampton:
I like the change to transparent too. Nice and clean!

klonos (approved):
LGTM 👍

quicksketch:

PR #2954, opened 2019-10-28
#4166 - set default margins on 'p' tag in Basis

stpaultim:


olafgrabienski (commented):
IIRC, Basis uses `rem`, not `em`.

PR #2955, opened 2019-10-28
3824 php max input vars

### LINKED ISSUE #3824: [UX] Throw a warning if max_input_vars is 1000 (default PHP setting)
**Description of the bug**

I just got bitten by the problem where PHP's `max_post_vars` is set to 1000 by default. This is a really frustrating bug because there are no errors or warnings, things just act funny when not all the values in a node form (for example) are passed through `$_POST`.

For backdrop sites, we should warn people when this value hasn't been adjusted appropriately. I was thinking of putting this in the status report, perhaps with a recommended value in the `ini_set()` section of settings.php

jenlampton:
Fixes https://github.com/backdrop/backdrop-issues/issues/3824

### DIFF
--- a/core/includes/form.inc
+++ b/core/includes/form.inc
@@ -909,6 +909,11 @@ function backdrop_process_form($form_id, &$form, &$form_state) {
   $unprocessed_form = $form;
   $form = form_builder($form_id, $form, $form_state);
 
+  // Set the message type depending on whether the form has been submitted or not.
+  $error_level = ($form_state['submitted']) ? 'error' : 'warning';
+  // Count the number of values and alert when max_input_vars is exceeded.
+  _form_check_input_vars($form['#element_count'], $error_level);
+
   // Only process the input if we have a correct form submission.
   if ($form_state['process_input']) {
     backdrop_validate_form($form_id, $form, $form_state);
@@ -1931,6 +1936,7 @@ function form_builder($form_id, &$element, &$form_state) {
 
   // Recurse through all child elements.
   $count = 0;
+  static $count_total;
   foreach (element_children($element) as $key) {
     // Prior to checking properties of child elements, their default properties
     // need to be loaded.
@@ -1948,6 +1954,15 @@ function form_builder($form_id, &$element, &$form_state) {
     if (isset($element['#access']) && !$element['#access']) {
       $element[$key]['#access'] = FALSE;
     }
+
+    // Count the elements at this level with IDs.
+    if (!isset($count_total)) {
+      $count_total = 0;
+    }
+    elseif (isset($element['#id'])) {
+      $count_total++;
+    }
+
 
     // Make child elements inherit their parent's #disabled and #allow_focus
     // values unless they specify their own.
@@ -2015,6 +2030,9 @@ function form_builder($form_id, &$element, &$form_state) {
         ),
       ),
     );
+
+    // Add a count of all child elements to the form itself.
+    $element['#element_count'] = $count_total;
 
     // If a form contains a single textfield, and the ENTER key is pressed
     // within it, Internet Explorer submits the form with no POST data
@@ -4983,6 +5001,96 @@ function _form_set_class(&$element, $class = array()) {
   }
   if (isset($element['#parents']) && form_get_error($element) !== NULL && !empty($element['#validated'])) {
     $element['#attributes']['class'][] = 'error';
+  }
+}
+
+/**
+ * Checks the number of input form elements to ensure that the PHP
+ * max_input_vars limit is not exceeded.
+ *
+ * @param array $count
+ *   The number of elements in the form.
+ * @param string $error_level
+ *   The error level for the message displayed on screen if the input vars value
+ *   has been exceeded. Either 'warning' or 'error'.
+ *
+ * @see form_pre_render()
+ */
+function _form_check_input_vars($count, $error_level = 'warning') {
+  // Determine the limit on input vars for this server configuration.
+  $limit = ini_get('max_input_vars');
+  if ($limit) {
+    // Leave a 5% buffer for ajax elements or similar.
+    if ($count > $limit * 0.95) {
+      $substitutions = array(
+        '@count' => $count,
+        '@limit' => $limit,
+      );
+
+      // Translatable strings.
+      $warning = t('This form may be too long to work properly.');
+      $notice = t('The last form processed was longer than this system can safely support.');
+      $please_increase = t('This form contains @count input elements. PHP <code>max_input_vars</code> allows only @limit and should be increased.', $substitutions);
+      $data_loss = t('The form has been submitted, but data may heve been lost.');
+
+      switch ($error_level) {
+        case 'error':
+          if (user_access('administer site configuration')) {
+            $message = $data_loss . ' ' . $please_increase;
+          }
+          else {
+            $message = $notice . ' ' . $data_loss;
+          }
+          break;
+
+        // 'warning'
+        default:
+          if (user_access('administer site configuration')) {
+            $message = $please_increase;
+          }
+          else {
+            $message = $warning;
+          }
+      }
+
+      // Remove existing messages that could be confusing.
+      $set_new_message = TRUE;
+      if (!empty($_SESSION['messages'])) {
+        $current_path = current_path();
+        $translations = array(
+          'admin/config/people/permissions' => t('The changes have been saved.'),
+          'admin/structure/taxonomy' => t('The configuration options have been saved.'),
+          'admin/structure/menu' => t('Your configuration has been saved.'),
+        );
+
+        foreach ($_SESSION['messages'] as $level => $set) {
+          foreach ($set as $key => $previous_message) {
+            // Look for a specific message on the permissions page.
+            if (array_key_exists($current_path, $translations) && $previous_message == $translations[$current_path]) {
+              unset($_SESSION['messages'][$level][$key]);
+            }
+
+            // If we are setting a warning now, but previously set an error...
+            if ($error_level == 'warning' && ($previous_message == $notice . ' ' . $data_loss || $previous_message == $data_loss . ' ' . $please_increase)) {
+              // ...don't set the warning.
+              $set_new_message = FALSE;
+            }
+          }
+
+          // Don't leave an empty set behind.
+          if (empty($_SESSION['messages'][$level])) {
+            unset($_SESSION['messages'][$level]);
+          }
+        }
+      }
+
+      if ($set_new_message) {
+        backdrop_set_message($message, $error_level);
+      }
+
+      $watchdog_message = 'Form contains @count input elements but PHP max_input_vars allows only @limit. Increase max_input_vars.';
+      watchdog('form', $watchdog_message, $substitutions, WATCHDOG_ERROR);
+    }
   }
 }
 


core/modules/system/system.install
```diff
@@ -72,6 +72,8 @@ function system_requirements($phase) {
     );
   }
 
+
+
```
klonos: Remove the extraneous lines here.

core/modules/system/system.install
```diff
@@ -98,6 +100,21 @@ function system_requirements($phase) {
     $requirements['php_register_globals']['value'] = $t('Disabled');
   }
 
+  // Test PHP max_input_vars setting.
+  $max_input_vars = ini_get('max_input_vars');
+  $requirements['php_max_input_vars'] = array(
+    'title' => $t('PHP form input max'),
```
```suggestion
    'title' => $t('PHP form input limit'),
```

core/modules/views/plugins/views_plugin_display_page.inc
```diff
@@ -264,7 +264,7 @@ class views_plugin_display_page extends views_plugin_display {
     $options['path'] = array(
       'category' => 'page',
       'title' => t('Path'),
-      'value' => views_ui_truncate($path, 24),
+      'value' => $path,
```
klonos: Is this change from another issue?

core/modules/system/system.install
```diff
@@ -98,6 +100,21 @@ function system_requirements($phase) {
...
+    'value' => $t("PHP limits the number of form values to @value items.", array('@value' => number_format($max_input_vars, 0, '.', ','))),
+  );
+  // 1000 is the default value, and indicates potential issues.
+  if ($max_input_vars == 1000) {
+    $requirements['php_max_input_vars']['description'] = $t('When set too low, PHP\'s <code>max_input_vars</code> setting can cause silent failures on large forms. Use caution when using the permissions page, ordering large menus, arraging large vocabularies, or editing content with lots of fields.');
+    $requirements['php_max_input_vars']['severity'] = REQUIREMENT_ERROR;
+  }
+  else {
```
klonos: This should be an `elseif`, and also check for more than a specific number. Something like:

```suggestion
  elseif ($max_input_vars >= 10000) {
```

Not sure what the exact number should be, but because the previous `if` statement checks for `== 1000`, this means that as is now, this condition will validate as TRUE even with a limit set to 1, or 50, or 100, telling people that that is sufficient.

core/modules/system/system.install
```diff
@@ -98,6 +100,21 @@ function system_requirements($phase) {
...
 
+  // Test PHP max_input_vars setting.
+  $max_input_vars = ini_get('max_input_vars');
+  $requirements['php_max_input_vars'] = array(
+    'title' => $t('PHP form input max'),
+    'value' => $t("PHP limits the number of form values to @value items.", array('@value' => number_format($max_input_vars, 0, '.', ','))),
+  );
+  // 1000 is the default value, and indicates potential issues.
+  if ($max_input_vars == 1000) {
+    $requirements['php_max_input_vars']['description'] = $t('When set too low, PHP\'s <code>max_input_vars</code> setting can cause silent failures on large forms. Use caution when using the permissions page, ordering large menus, arraging large vocabularies, or editing content with lots of fields.');
```
```suggestion
    $requirements['php_max_input_vars']['description'] = $t('When set too low, PHP\'s <code>max_input_vars</code> setting can cause silent failures on large forms. Use caution when using the <em>Permissions</em> page, ordering large menus, arranging large vocabularies, or editing content with lots of fields.');
```

klonos (changes requested):
This looks good in general, but please see my comments.

core/modules/system/system.install
jenlampton: done.

core/modules/system/system.install
jenlampton: I don't really like either of these. It's not a limit on forms, or even a limit on input. It's an input on items per form. How do we convey that?

core/modules/views/plugins/views_plugin_display_page.inc
jenlampton: yes. I think this is on my 1.x branch, it keeps getting into other things...

jenlampton: found it.

core/modules/system/system.install
jenlampton: good catch.

core/modules/system/system.install
jenlampton: I don't see a reason for emphasizing permissions (?) but I've made the other change.

core/modules/system/system.install
jenlampton: I tried `PHP form value limit` let's try that?

klonos: Another option: `Form input PHP limit`?

core/modules/system/system.install
klonos: 🙂

core/modules/system/system.install
klonos: The recommendation is as per other places in the UI, where we render things that have a name/title in italics. It's not "some permissions page"; it's a "the _Permissions_ page".

core/modules/system/system.install
```diff
@@ -98,6 +98,18 @@ function system_requirements($phase) {
     $requirements['php_register_globals']['value'] = $t('Disabled');
   }
 
+  // Test PHP max_input_vars setting.
+  $max_input_vars = ini_get('max_input_vars');
+  $requirements['php_max_input_vars'] = array(
+    'title' => $t('PHP form value limit'),
+    'value' => $t("PHP limits the number of values in a single form to @value items.", array('@value' => number_format($max_input_vars, 0, '.', ','))),
```
klonos: "PHP limits" might be interpreted as a general statement about PHP; we should make it clear that we are referring to something that is currently configured for the site in question. Something like this perhaps?:

```suggestion
    'value' => $t('PHP is currently configured to limit the number of values in each form to @value items.', array('@value' => number_format($max_input_vars, 0, '.', ','))),
```

core/modules/system/system.install
```diff
@@ -98,6 +98,18 @@ function system_requirements($phase) {
...
+  );
+  // 1000 is the default value, and indicates potential issues.
+  if ($max_input_vars <= 1000) {
+    $requirements['php_max_input_vars']['description'] = $t('When set too low, PHP\'s <code>max_input_vars</code> setting can cause silent failures on large forms. Use caution when using the permissions page, ordering large menus, arranging large vocabularies, or editing content with lots of fields.');
```
```suggestion
    $requirements['php_max_input_vars']['description'] = $t('When set too low, the <code>max_input_vars</code> PHP setting can cause silent failures on forms with a large amount of input fields, such as checkboxes. Read more about this <a href="@documentation" target="_blank">here</a>.', array('@documentation' => url('https://api.backdropcms.org/max-input-vars)));
```

We are just warning people here about a value being "too low"; should we add some accepted/sufficient value? Something like "A value of 10000 is generally found to be sufficient"

core/modules/system/system.install
```diff
@@ -98,6 +98,18 @@ function system_requirements($phase) {
...
+    $requirements['php_max_input_vars']['severity'] = REQUIREMENT_ERROR;
```
klonos, line 110: Given that not all sites are affected by this, should we change this to a warning instead? ...or perhaps leave it be an error if bellow 1000, and a warning if between 1001 and 10007? We could have the message say that a safe value is anything above 10k.

.htaccess
```diff
@@ -28,6 +28,14 @@ DirectoryIndex index.php index.html index.htm
...
   php_flag mbstring.encoding_translation    off
 </IfModule>
+# Support PHP7 as well as PHP5.
+<IfModule mod_php7.c>
+  php_flag session.auto_start               off
+  php_value mbstring.http_input             pass
+  php_value mbstring.http_output            pass
+  php_value max_input_vars                  10007
+  php_flag mbstring.encoding_translation    off
+</IfModule>
```
klonos: This section should be moved before the php5 block, as per @quicksketch's feedback on the issue.

.htaccess
```diff
@@ -21,6 +21,10 @@ DirectoryIndex index.php index.html index.htm
 # Override PHP settings that cannot be changed at runtime. See
 # settings.php and backdrop_environment_initialize() in includes/bootstrap.inc
 # for settings that can be changed at runtime.
+<IfModule mod_php7.c>
+  php_value max_input_vars                  10007
+</IfModule>
+# Same settings again for PHP 5 versions.
```
klonos: With https://github.com/backdrop/backdrop/pull/3399 merged, this change should be removed, as it causes conflicts.

core/includes/form.inc
```diff
@@ -4889,6 +4892,61 @@ function _form_set_class(&$element, $class = array()) {
...
+ *
+ * @return int
+ *   Number of non-array elements within $a.
+ */
+function _form_count_terminals($element) {
+  return is_array($element)
+    ? array_reduce($element, function ($carry, $item) {
+      return $carry + _form_count_terminals($item);
+    }, 0)
+    : 1;
```
klonos: Can we please simplify this to a nice if/else?

core/includes/form.inc
```diff
@@ -4889,6 +4892,61 @@ function _form_set_class(&$element, $class = array()) {
...
+    }
+  }
+}
+
+/**
+ * Counts terminals in an array.
+ *
+ * Useful for counting how many input_vars were returned in $_POST.
+ *
+ * @param $a
```
klonos:
```suggestion
 * @param $element
```

core/includes/form.inc
```diff
@@ -4889,6 +4892,61 @@ function _form_set_class(&$element, $class = array()) {
...
+ *   Array or array element to be counted
+ *
+ * @return int
+ *   Number of non-array elements within $a.
```
klonos:
```suggestion
 *   Number of non-array elements within $element.
```

core/includes/form.inc
```diff
@@ -4889,6 +4906,97 @@ function _form_set_class(&$element, $class = array()) {
...
+            // Look for a specific message on the permissions page.
+            if ($current_path == 'admin/config/people/permissions' && $previous_message == t('The changes have been saved.')) {
+              unset($_SESSION['messages'][$level][$key]);
+            }
+            elseif (strstr($current_path, 'admin/structure/taxonomy') && $previous_message == t('The configuration options have been saved.')) {
+              unset($_SESSION['messages'][$level][$key]);
+            }
+            elseif (strstr($current_path, 'admin/structure/menu') && $previous_message == t('Your configuration has been saved.')) {
+              unset($_SESSION['messages'][$level][$key]);
+            }
```
klonos: Since we're always doing the same `unset($_SESSION['messages'][$level][$key]);` for every conditional, it feels like this bit of code here could be improved/consolidated.

Also, this feels like it should be a more generic/global duplicate message prevention functionality (in `backdrop_set_message()`, where we check whether the `$repeat` parameter is set to `TRUE`/`FALSE`), so that we can benefit in other cases/places from it.

core/includes/form.inc
```diff
@@ -4973,6 +4990,97 @@ function _form_set_class(&$element, $class = array()) {
...
+      $substitutions = array(
+        '@count' => $count,
+        '@limit' => $limit,
+      );
+
+      // Translatable strings.
+      $warning = t('This form may be too long to work properly.');
+      $notice = t('The last form processed was longer than this system can safely support.');
+      $please_increase = t('This form contains @count input elements. PHP max_input_vars allows only @limit and should be increased.', $substitutions);
+      $data_loss = t('The form has been submitted, but data may heve been lost.');
```
ghost, line 5034: Typo: 'heve'

```suggestion
      $data_loss = t('The form has been submitted, but data may have been lost.');
```

core/includes/form.inc
```diff
@@ -4973,6 +4990,97 @@ function _form_set_class(&$element, $class = array()) {
...
 
+/**
+ * Checks the number of input form elements to ensure that the PHP
+ * max_input_vars limit is not exceeded.
+ *
+ * @param array $count
+ *   The number of elements in the form.
+ * @param string $error_level
+ *   The error level for the message displayed on screen if the input vars value
+ *   has been exceeded.
```
ghost: Would be good to list the possible values here, since there are only two.

```suggestion
 *   has been exceeded. Either 'warning' or 'error'.
```

core/includes/form.inc
```diff
@@ -4973,6 +4990,97 @@ function _form_set_class(&$element, $class = array()) {
...
+          break;
+      }
+
+      // Remove existing messages that could be confusing.
+      $set_new_message = TRUE;
+      if (!empty($_SESSION['messages'])) {
+
+        $current_path = current_path();
+        foreach ($_SESSION['messages'] as $level => $set) {
+          foreach ($_SESSION['messages'][$level] as $key => $previous_message) {
```
ghost: Couldn't this just be:

```suggestion
          foreach ($set as $key => $previous_message) {
```

core/includes/form.inc
```diff
@@ -909,6 +909,10 @@ function backdrop_process_form($form_id, &$form, &$form_state) {
   $unprocessed_form = $form;
   $form = form_builder($form_id, $form, $form_state);
 
+  // Count the number of values and alert when max_input_vars is exceeded.
+  $error_level = ($form_state['submitted'])? 'error' : 'warning';
```
ghost: This line could use its own comment explaining what it's doing (wasn't immediately obvious to me). And there's a missing space before the '?'.

```suggestion
  // Set the message type depending on whether the form has been submitted or not.
  $error_level = ($form_state['submitted']) ? 'error' : 'warning';
  // Count the number of values and alert when max_input_vars is exceeded.
```

core/includes/form.inc
```diff
@@ -4973,6 +4990,97 @@ function _form_set_class(&$element, $class = array()) {
...
+          break;
+
+        case 'error':
+          if (user_access('administer site configuration')) {
+            $message = $data_loss . ' ' . $please_increase;
+          }
+          else {
+            $message = $notice . ' ' . $data_loss;
+          }
+          break;
```
ghost: I think 'warning' should be the default for the switch, otherwise there might be issues with an undefined `$message` variable later if an error level other than 'warning' or 'error' is passed in...

```suggestion
        case 'error':
          if (user_access('administer site configuration')) {
            $message = $data_loss . ' ' . $please_increase;
          }
          else {
            $message = $notice . ' ' . $data_loss;
          }
          break;

        // 'warning'
        default:
          if (user_access('administer site configuration')) {
            $message = $please_increase;
          }
          else {
            $message = $warning;
          }
```

core/modules/system/system.install
```diff
@@ -110,6 +110,18 @@ function system_requirements($phase) {
     $requirements['php_register_globals']['value'] = $t('Disabled');
   }
 
+  // Test PHP max_input_vars setting.
+  $max_input_vars = ini_get('max_input_vars');
+  $requirements['php_max_input_vars'] = array(
+    'title' => $t('PHP form value limit'),
+    'value' => $t('PHP limits the number of values in a single form to @value items.', array('@value' => number_format($max_input_vars, 0, '.', ','))),
```
ghost: Should be called 'PHP form *elements* limit' instead...? Also I think the term 'max_input_vars' should be written here somewhere in case people search for that.

```suggestion
    'title' => $t('PHP form elements limit') . ' (<code>max_input_vars</code>)',
    'value' => $t('PHP limits the number of elements in a single form to @value items.', array('@value' => number_format($max_input_vars, 0, '.', ','))),
```

core/modules/system/system.install
```diff
@@ -110,6 +110,18 @@ function system_requirements($phase) {
...
+  );
+  // 1000 is the default value and indicates potential issues.
```
ghost: Might be good to link to where this is stated:

```suggestion
  // 1000 is the default value and indicates potential issues.
  // @see https://www.php.net/manual/en/info.configuration.php
```

core/modules/system/system.install
```diff
@@ -110,6 +110,18 @@ function system_requirements($phase) {
...
+  if ($max_input_vars <= 1000) {
+    $requirements['php_max_input_vars']['description'] = $t('When set too low, the <code>max_input_vars</code> PHP setting can cause silent failures on forms with a large number of input fields, such as the permissions page. <a href="@documentation" target="_blank">Read more about increasing the <code>max_input_vars setting</code></a>.', array('@documentation' => url('https://api.backdropcms.org/max-input-vars')));
```
ghost: Updated URL.

```suggestion
    $requirements['php_max_input_vars']['description'] = $t('When set too low, the <code>max_input_vars</code> PHP setting can cause silent failures on forms with a large number of input fields, such as the permissions page. <a href="@documentation" target="_blank">Read more about increasing the <code>max_input_vars</code> setting</a>.', array('@documentation' => url('https://docs.backdropcms.org/max-input-vars')));
```

core/includes/form.inc
```diff
@@ -1944,6 +1949,15 @@ function form_builder($form_id, &$element, &$form_state) {
       $element[$key]['#access'] = FALSE;
     }
 
+    // Count the elements at this level with IDs.
+    if (!isset($count_total)) {
+      $count_total = 0;
+    }
+    elseif (isset($element['#id'])) {
```
ghost, line 1962: How certain are we that all elements on the form will have an ID? Wouldn't it be better to just count the number of `<input>` elements...?

ghost (changes requested):

core/includes/form.inc
klonos: @BWPanda the code suggestion above has an empty line with whitespace in it. Can you please fix it, so that it can be merged directly if the change accepted?

ghost: @klonos Done

core/includes/form.inc
jenlampton: All form API elements will have an ID since the ID is derived from the form API key. There are a lot of form elements that aren't `<input>` elements so I think using IDs will be more accurate (it won't always be perfect, but closer than leaving out selects, textareas, etc).

core/modules/system/system.install
jenlampton: I think saying "form elements" might mislead people. They might think the limit is on the form, and not the submission.

Maybe we should use the same language PHP uses on  https://www.php.net/manual/en/info.configuration.php#ini.max-input-vars?

`How many input variables may be accepted (limit is applied to $_GET, $_POST and $_COOKIE superglobal separately). `

```php
'title' => $t('PHP <code>max_input_vars</code> setting') )',
'value' => $t('PHP limits the number of input variables from each form to @value items.', array('@value' => number_format($max_input_vars, 0, '.', ','))),
```

core/includes/form.inc
ghost: @jenlampton I'm not seeing this change on https://github.com/backdrop/backdrop/pull/2955/files...

core/modules/system/system.install
ghost: Good point. New suggestion:

```suggestion
    'title' => $t('PHP <code>max_input_vars</code> setting'),
    'value' => $t('PHP limits the number of input variables from each form to @value items.', array('@value' => number_format($max_input_vars, 0, '.', ','))),
```

core/includes/form.inc
jenlampton: I've cleaned this up a little the latest PR.

core/modules/system/system.install
```diff
@@ -110,6 +110,19 @@ function system_requirements($phase) {
...
+  // Test PHP max_input_vars setting.
+  $max_input_vars = ini_get('max_input_vars');
+  $requirements['php_max_input_vars'] = array(
+    'title' => $t('PHP <code>max_input_vars</code> setting'),
+    'value' => $t('PHP limits the number of input variables in each form to @value items.', array('@value' => number_format($max_input_vars, 0, '.', ','))),
+  );
+  // 1000 is the default value and indicates potential issues.
+  // @see https://www.php.net/manual/en/info.configuration.php
+  if ($max_input_vars <= 1000) {
+    $requirements['php_max_input_vars']['description'] = $t('When set too low, the <code>max_input_vars</code> PHP setting can cause silent failures on large forms, such as the permissions page. A value of <code>10000</code> is recommended. <a href="@documentation" target="_blank">Read more about increasing the <code>max_input_vars</code> setting</a>.', array('@documentation' => url('https://docs.backdropcms.org/max-input-vars')));
```
jenlampton: @klonos  I included your recommended value here.

ghost (approved):

core/includes/form.inc
```diff
@@ -4973,6 +4991,96 @@ function _form_set_class(&$element, $class = array()) {
...
+    if ($count > $limit * 0.95) {
+      $substitutions = array(
+        '@count' => $count,
+        '@limit' => $limit,
+      );
+
+      // Translatable strings.
+      $warning = t('This form may be too long to work properly.');
+      $notice = t('The last form processed was longer than this system can safely support.');
+      $please_increase = t('This form contains @count input elements. PHP max_input_vars allows only @limit and should be increased.', $substitutions);
```
klonos:
```suggestion
      $please_increase = t('This form contains @count input elements. PHP <code>max_input_vars</code> allows only @limit and should be increased.', $substitutions);
```

core/includes/form.inc
```diff
@@ -4986,6 +5004,96 @@ function _form_set_class(&$element, $class = array()) {
...
+
+      // Remove existing messages that could be confusing.
+      $set_new_message = TRUE;
+      if (!empty($_SESSION['messages'])) {
+        $current_path = current_path();
+        $translations = array(
+          'admin/config/people/permissions' => t('The changes have been saved.'),
+          'admin/structure/taxonomy' => t('The configuration options have been saved.'),
+          'admin/structure/menu' => t('Your configuration has been saved.'),
+        );
```
klonos, line 5064: Is there any better way to do this? (instead of duplicating anticipated text here) 🤔 ...if these messages happen to change in their respective submit handlers, then we need to somehow remember to change them here too. ...is there a way to reference them?

Thoughts/ideas:
- use constants (seems like a messy workaround and I don't like it)
- a helper function that generates these messages based on path (passed as a param to the function), and then we reuse that function here, as well as in the respective submit handlers for the forms (those that are prone to this issue). The function would be generating two sets of messages: one for when `max_input_vars` is sufficient, and one for when not.
- something that intercepts these messages in `backdrop_set_message()`? 🤔

PR #2958, opened 2019-10-29
Issue #4170: Prevent PHP notice and SQL fatal by not passing image id.

### LINKED ISSUE #4170: SQL fatal an PHP Notice: Undefined variable: id in file_field_update()
After upgrading to 1.14, I'm getting the following PHP notice when removing, and then adding a new image to an existing multifield (to an image subfield). 

```
Notice: Undefined variable: id in file_field_update() (line 292 of backdrop/core/modules/file/file.field.inc).
```
I'm also seeing the following SQL fatal:
```
SQLSTATE[23000]: Integrity constraint violation: 1048 Column 'id' cannot be null
```

There may be an oversight here:
```php
  // Add new usage entries for newly added files.
  foreach ($items as $item) {
    if (!in_array($item['fid'], $original_fids)) {
      if ($file = file_load($item['fid'])) {
        file_usage_add($file, 'file', $entity_type, $id);
      }
    }
  }
```
Shouldn't this `$id` be `$item['fid']` instead?
```php
  // Add new usage entries for newly added files.
  foreach ($items as $item) {
    if (!in_array($item['fid'], $original_fids)) {
      if ($file = file_load($item['fid'])) {
        file_usage_add($file, 'file', $entity_type, $item['fid']);
      }
    }
  }
```

---
PR: https://github.com/backdrop/backdrop/pull/2958

jenlampton:
Fixes https://github.com/backdrop/backdrop-issues/issues/4170

### DIFF
--- a/core/modules/file/file.field.inc
+++ b/core/modules/file/file.field.inc
@@ -282,7 +282,7 @@ function file_field_update($entity_type, $entity, $field, $instance, $langcode, 
   foreach ($items as $item) {
     if (!in_array($item['fid'], $original_fids)) {
       if ($file = file_load($item['fid'])) {
-        file_usage_add($file, 'file', $entity_type, $id);
+        file_usage_add($file, 'file', $entity_type, $item['fid']);
       }
     }
   }


core/modules/file/file.field.inc
```diff
@@ -282,7 +282,7 @@ function file_field_update($entity_type, $entity, $field, $instance, $langcode,
   foreach ($items as $item) {
     if (!in_array($item['fid'], $original_fids)) {
       if ($file = file_load($item['fid'])) {
-        file_usage_add($file, 'file', $entity_type, $id);
+        file_usage_add($file, 'file', $entity_type, $item['fid']);
```
indigoxela, line 285: I suspect, this is not the right approach.

$id gets set in line 247 and should be - according to the comment block of function file_usage_add():

> The unique, numeric ID of the object containing the referenced file.

Not the file fid.

PR #2960, opened 2019-10-30
Issue #4175: Add the complete file path to the manage file form.

### LINKED ISSUE #4175: Add the complete file path to the manage file form.
When I'm managing my files, I really want to know the file name (not the human-readable label we already show in the form, but the actual name of the document on disk). I also sometimes wonder where that file is stored, because I'd like to test it directly in my browser. We don't reveal any of this information to the user, and it seems both are a valuable part of managing files.

I'd like to propose the following improvements to the Manage File form:
* ~Change the label of the "File name" element to indicate this is a label~
* Change the vertical tab heading for "Destination" to "File location"
* Display the complete file path in the "File location" vertical tab. 
* The filepath field should be read-only and not allow changes

Screenshot:
Screen Shot 2019-10-30 at 1 55 35 PM

~I went with "File link text" in the screenshot (and first PR) but am open to recommendations on what to call the human-friendly name for a file~

~I went with `Description` as the label for that field, since that is what we are calling the same thing in other interfaces. See https://github.com/backdrop/backdrop-issues/issues/4178 for better text to use there.~

---
PR: https://github.com/backdrop/backdrop/pull/2960

---

Translation updates:
* "Destination" replaced with "File location"

jenlampton:
Fixes https://github.com/backdrop/backdrop-issues/issues/4175

### DIFF
--- a/core/modules/file/file.pages.inc
+++ b/core/modules/file/file.pages.inc
@@ -575,7 +575,7 @@ function file_manage_form($form, &$form_state, File $file) {
     $form['destination']['url'] = array(
       '#type' => 'item',
       '#title' => t('File URL'),
-      '#markup' => url($filepath, array('absolute' => TRUE)),
+      '#markup' => url(check_plain($filepath), array('absolute' => TRUE)),
     );
   }
 
--- a/core/modules/file/file.pages.inc
+++ b/core/modules/file/file.pages.inc
@@ -498,10 +498,17 @@ function file_manage_form($form, &$form_state, File $file) {
     );
   }
 
+  $form['name'] = array(
+    '#type' => 'item',
+    '#title' => t('Filename'),
+    '#markup' => check_plain($file->filename),
+    '#weight' => -11,
+  );
+
   $form['filename'] = array(
     '#type' => 'textfield',
-    '#title' => t('File display title'),
-    '#description' => t('This text will be used in links to the uploaded file.'),
+    '#title' => t('Description'),
+    '#description' => t('This description will be used as the text in links to the file.'),
     '#default_value' => $file->filename,
     '#required' => TRUE,
     '#maxlength' => 255,
@@ -524,6 +531,7 @@ function file_manage_form($form, &$form_state, File $file) {
     );
   }
 
+
   $form['additional_settings'] = array(
     '#type' => 'vertical_tabs',
     '#weight' => 99,
@@ -533,7 +541,7 @@ function file_manage_form($form, &$form_state, File $file) {
   $form['destination'] = array(
     '#type' => 'fieldset',
     '#access' => user_access('manage files'),
-    '#title' => t('Destination'),
+    '#title' => t('File location'),
     '#collapsible' => TRUE,
     '#collapsed' => TRUE,
     '#group' => 'additional_settings',
@@ -558,6 +566,18 @@ function file_manage_form($form, &$form_state, File $file) {
     '#options' => $options,
     '#default_value' => file_uri_scheme($file->uri),
   );
+
+
+  if (!empty($file->uri)) {
+    $directory_path = file_stream_wrapper_get_instance_by_uri($file->uri)->getDirectoryPath();
+    $filename = file_uri_target($file->uri);
+    $filepath = $directory_path . '/' . $filename;
+    $form['destination']['url'] = array(
+      '#type' => 'item',
+      '#title' => t('File URL'),
+      '#markup' => url($filepath, array('absolute' => TRUE)),
+    );
+  }
 
   // File user information for administrators.
   $anonymous = config_get_translated('system.core', 'anonymous');
--- a/core/modules/file/file.pages.inc
+++ b/core/modules/file/file.pages.inc
@@ -507,8 +507,8 @@ function file_manage_form($form, &$form_state, File $file) {
 
   $form['filename'] = array(
     '#type' => 'textfield',
-    '#title' => t('Description'),
-    '#description' => t('This description will be used as the text in links to the file.'),
+    '#title' => t('File display title'),
+    '#description' => t('This text will be used in links to the uploaded file.'),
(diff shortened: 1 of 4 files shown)


core/modules/file/file.pages.inc
```diff
@@ -553,6 +555,17 @@ function file_manage_form($form, &$form_state, File $file) {
     '#default_value' => file_uri_scheme($file->uri),
   );
 
+  if (!empty($file->uri)) {
+    $form['destination']['path'] = array(
+      '#type' => 'textfield',
```
klonos: **Technically** this is a URL - not a path. Any reason why this cannot be `'#type' => 'markup'`, or even better:

```suggestion
      '#type' => 'item',
```

https://api.backdropcms.org/form_api#item

>**Description**: Generate a display-only form element allowing for an optional title and description.

core/modules/file/file.pages.inc
```diff
@@ -527,7 +529,7 @@ function file_manage_form($form, &$form_state, File $file) {
   $form['destination'] = array(
     '#type' => 'fieldset',
     '#access' => user_access('manage files'),
-    '#title' => t('Destination'),
+    '#title' => t('File location'),
```
klonos, line 544: 👍

core/modules/file/file.pages.inc
```diff
@@ -495,7 +495,8 @@ function file_manage_form($form, &$form_state, File $file) {
 
   $form['filename'] = array(
     '#type' => 'textfield',
-    '#title' => t('File name'),
+    '#title' => t('Description'),
```
klonos:
```suggestion
    '#title' => t('File display title'),
```

😉

...and perhaps lose the `'#description'`?

core/modules/file/file.pages.inc
jenlampton: why item? Does it still provide an easy-to-copy-paste input field?

core/modules/file/file.pages.inc
jenlampton: This is handled in the other PR, and that one updates all the other places too. Think we should keep them separate.

core/modules/file/file.pages.inc
jenlampton: I tried item, and the input disappeared entirely. https://api.backdropcms.org/form_api#item

Looks like `#item` would need `#markup`. instead of `#default_value`. trying that...

core/modules/file/file.pages.inc
klonos: Then revert all these changes here?

jenlampton: Description is better than what we have now because it's consistent with everything else in the UI :) I'd like either of these to be able to get in w/o the other incase there's a hangup.

core/modules/file/file.pages.inc
jenlampton: Pushed a new PR

core/modules/file/file.pages.inc
```diff
@@ -498,6 +498,13 @@ function file_manage_form($form, &$form_state, File $file) {
     );
   }
 
+  $form['name'] = array(
+    '#type' => 'item',
+    '#title' => t('Filename'),
+    '#markup' => $file->filename,
```
quicksketch: Looks likely to possibly create an XSS issue here. Wrapping in a `check_plain()` wouldn't hurt.

core/modules/file/file.pages.inc
```diff
@@ -559,6 +567,18 @@ function file_manage_form($form, &$form_state, File $file) {
...
 
+
+  if (!empty($file->uri)) {
+    $directory_path = file_stream_wrapper_get_instance_by_uri($file->uri)->getDirectoryPath();
+    $filename = file_uri_target($file->uri);
+    $filepath = $directory_path . '/' . $filename;
+    $form['destination']['url'] = array(
+      '#type' => 'item',
+      '#title' => t('File URL'),
+      '#markup' => url($filepath, array('absolute' => TRUE)),
```
quicksketch: I think another check plain is needed here, in the event a file name attempts to include JS exploit.

```suggestion
      '#markup' => url(check_plain($filepath), array('absolute' => TRUE)),
```

Alternatively, this could run through `l()` which already escapes by default, but it makes it an `<a>` tag, not sure if that's desired here:
```suggestion
      '#markup' => l($filepath, $filepath, array('absolute' => TRUE)),
```

core/modules/file/file.pages.inc
quicksketch: These changes should be removed. We just changed these strings in https://github.com/backdrop/backdrop-issues/issues/4178 (in fact in a PR by @jenlampton at https://github.com/backdrop/backdrop/pull/2961).

core/modules/file/file.pages.inc
```diff
@@ -559,6 +567,18 @@ function file_manage_form($form, &$form_state, File $file) {
...
 
+
+  if (!empty($file->uri)) {
+    $directory_path = file_stream_wrapper_get_instance_by_uri($file->uri)->getDirectoryPath();
+    $filename = file_uri_target($file->uri);
+    $filepath = $directory_path . '/' . $filename;
+    $form['destination']['url'] = array(
+      '#type' => 'item',
+      '#title' => t('File URL'),
+      '#markup' => url(check_plain($filepath), array('absolute' => TRUE)),
```
quicksketch: Sorry, I know you just changed this, but I agree with @klonos's feedback in the issue that this should probably be a link, not just the URL as text.

image

```suggestion
      '#markup' => l($filepath, $filepath, array('absolute' => TRUE)),
```

Though this could cause a new UX issue that clicking the link may loose changes on the form. We have an existing behavior in `Backdrop.behaviors.filePreviewLinks` that might help, but that might be out of scope?

herbdool:
Replaced by https://github.com/backdrop/backdrop/pull/4619

PR #1999, opened 2017-09-08
Issue #2740: Fix views dropbuttons.

opi:
Based on @quicksketch's https://github.com/backdrop/backdrop/pull/1971

PR #2962, opened 2019-10-31
Issue #1999 - Removing the minimal install profile

stpaultim:
Just wanted to party like it's #1999, so I removed the install profile to be a bit provocative and see what happens. 

Not sure if this is a good idea and since I didn't understand most of the changes I was making at in the "test" section, this needs to be looked at closely and potentially disregarded.


stpaultim:
OK that didn't go well. I didn't really expect it to be that easy, but gave it a try. Tempted to simply close PR, but I'll leave it here just in case my failed experiment is useful to anyone else.

PR #2966, opened 2019-11-01
Issue #4181: Fix stub entities

### LINKED ISSUE #4181: Remove entity_create_stub_entity() in favor of entity_create()
In https://github.com/backdrop/backdrop-issues/issues/1475 I propose to add forward revision capability to core. The corresponding issue in D7 listed [Remove stub entities, replace entity_create_stub_entity()](https://www.drupal.org/project/drupal/issues/1551140) as a blocker, subsequently fixed. This issue is to decide if this needs to follow the D7 path.

`entity_create_stub_entity()` is called only in four places: 3 times in Field tests and also in `EntityFieldQuery::finishQuery()` and just creates a `stdClass()` Object to initialize a stub entity, which was fine when all D7 Entities were not classed, but now in Backdrop all Entities are classed Objects.

The original resolution to stub entities when D7 started classing entities was to just change  `entity_create_stub_entity()` 
```
function entity_create_stub_entity($entity_type, $ids) {
- $entity = new stdClass();
...
-  return $entity;
+ return isset($info['entity class']) ? entity_create($entity_type, $values) : (object) $values;
}
```

But ultimately the patch committed just before forward revisions was committed was not that "simple" fix but [this one](https://www.drupal.org/files/stub-entity-concept-removed-1551140-38.patch) which removes stub entities completely. 

This issue is to track which route would be best for Backdrop. 

Would appreciate some advice from @quicksketch on how he feels we should move forward as this as noted above is a blocker.

---
PR by @docwilmot https://github.com/backdrop/backdrop/pull/2966

docwilmot:
Fixes https://github.com/backdrop/backdrop-issues/issues/4181

### DIFF
--- a/core/modules/entity/entity.module
+++ b/core/modules/entity/entity.module
@@ -334,6 +334,38 @@ function entity_extract_ids($entity_type, $entity) {
 }
 
 /**
+ * Assembles an object structure with initial IDs.
+ *
+ * This function can be seen as reciprocal to entity_extract_ids().
+ *
+ * @param $entity_type
+ *   The entity type; e.g. 'node' or 'user'.
+ * @param $ids
+ *   A numerically indexed array, as returned by entity_extract_ids(),
+ *   containing these elements:
+ *   - 0: Primary ID of the entity.
+ *   - 1: Revision ID of the entity, or NULL if $entity_type is not versioned.
+ *   - 2: Bundle name of the entity, or NULL if $entity_type has no bundles.
+ *
+ * @return
+ *   An entity object, initialized with the IDs provided.
+ *
+ * @deprecated since 1.15.0.
+ */
+function entity_create_stub_entity($entity_type, $ids) {
+  $values = array();
+  $info = entity_get_info($entity_type);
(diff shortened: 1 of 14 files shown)


herbdool (commented):
This code looks good to me. I haven't tested but looks like it's mostly about tests passing.

core/modules/entity/entity.module
```diff
@@ -349,18 +349,20 @@ function entity_extract_ids($entity_type, $entity) {
  *
  * @return
  *   An entity object, initialized with the IDs provided.
+ *
+ * @deprecated since 1.15.0.
  */
 function entity_create_stub_entity($entity_type, $ids) {
-  $entity = new stdClass();
```
quicksketch, line 354: Could you add a deprecated watchdog call in here as the first line?
```
watchdog_deprecated_function('entity', __FUNCTION__);
```

docwilmot: Done

docwilmot:
Sandbox did funny things, so re-opening.

quicksketch:
Thanks @docwilmot! Merged into 1.x for 1.15.0.

PR #2967, opened 2019-11-01
Issue #4181: Fix stub entities

docwilmot:
Fixes https://github.com/backdrop/backdrop/pull/2966


docwilmot:
This is approaching the previous PR, except with keeping the function `entity_create_stub_entity()`. So will close and resume on the previous PR.

PR #2969, opened 2019-11-02
Issue #4191: Added the missing @see lines to config_* function doc blocks

### LINKED ISSUE #4191: Add missing `@see` lines in the `config_*()` functions' docblocks.
The `config_set()`, `config_diff()`, `config_clear()` and `config_menu()` functions do not have the @see lines in the doc blocks. As a result, API documentation pages like https://api.backdropcms.org/api/backdrop/core%21includes%21config.inc/function/config_get/1 are missing **See also** sections.

The issue was first reported on https://github.com/backdrop-ops/api.backdropcms.org/issues/71, however because PR titles automatically pull up issue IDs from https://github.com/backdrop I had to create this one.

---
~~PR by @alanmels: https://github.com/backdrop/backdrop/pull/2969~~
PR by @alanmels: https://github.com/backdrop/backdrop/pull/2979

alanmels:
Fixes https://github.com/backdrop/backdrop-issues/issues/4191

### DIFF
--- a/core/includes/config.inc
+++ b/core/includes/config.inc
@@ -64,6 +64,11 @@ function config($config_file, $type = 'active') {
  *   The name of the config option within the file to delete. The config option
  *   may contain periods to indicate levels within the config file.
  *
+ * @see config_set()
+ * @see config_get()
+ * @see config_diff()
+ * @see config_menu()
+ *
  * @since 1.7.0
  */
 function config_clear($config_file, $option) {
@@ -87,6 +92,11 @@ function config_clear($config_file, $option) {
  * @return mixed
  *   The contents of the requested config option. Returns NULL if the specified
  *   option was not found in the file at all.
+ *
+ * @see config_set()
+ * @see config_diff()
+ * @see config_clear()
+ * @see config_menu()
  */
 function config_get($config_file, $option = NULL) {
   $config = config($config_file);
@@ -145,6 +155,11 @@ function config_get_translated($config_file, $option = NULL, $args = array(), $o
  *   may contain periods to indicate levels within the config file.
  * @param mixed $value
  *   The value to save into the config file.
+ *
+ * @see config_get()
+ * @see config_diff()
+ * @see config_clear()
+ * @see config_menu()
  */
 function config_set($config_file, $option, $value) {
   $config = config($config_file);
--- a/core/modules/config/config.admin.inc
+++ b/core/modules/config/config.admin.inc
@@ -542,6 +542,11 @@ function config_get_modified_times() {
  *
  * @return array
  *   An array of formatted strings showing the diffs between the two storages.
+ * 
+ * @see config_set()
+ * @see config_get()
+ * @see config_clear()
+ * @see config_menu()
  */
 function config_diff($name) {
   $active_storage = config_get_config_storage('active');
--- a/core/modules/config/config.module
+++ b/core/modules/config/config.module
@@ -6,6 +6,11 @@
 
 /**
  * Implements hook_menu().
+ * 
+ * @see config_set()
+ * @see config_get()
+ * @see config_clear()
+ * @see config_diff()
  */
 function config_menu() {
   $items['admin/config/development/configuration'] = array(


klonos (approved):
LGTM 👍

klonos:
@alanmels give it some time (say 1hr), then close this PR, then wait a couple of mins (for the CI to destroy the PR sandbox etc.), then reopen the PR. This usually solves the weird timezone error for me.

core/includes/config.inc
```diff
@@ -64,6 +64,11 @@ function config($config_file, $type = 'active') {
  *   The name of the config option within the file to delete. The config option
  *   may contain periods to indicate levels within the config file.
  *
+ * @see config_set()
+ * @see config_get()
+ * @see config_diff()
+ * @see config_menu()
```
quicksketch, line 70: I don't think `config_diff()` and `config_menu()` are necessary here.

core/includes/config.inc
```diff
@@ -87,6 +92,11 @@ function config_clear($config_file, $option) {
  * @return mixed
  *   The contents of the requested config option. Returns NULL if the specified
  *   option was not found in the file at all.
+ *
+ * @see config_set()
+ * @see config_diff()
+ * @see config_clear()
+ * @see config_menu()
```
quicksketch, line 99: Same here, I don't think `config_diff()` and `config_menu()` are necessary here.

core/includes/config.inc
```diff
@@ -145,6 +155,11 @@ function config_get_translated($config_file, $option = NULL, $args = array(), $o
  *   may contain periods to indicate levels within the config file.
  * @param mixed $value
  *   The value to save into the config file.
+ *
+ * @see config_get()
+ * @see config_diff()
+ * @see config_clear()
+ * @see config_menu()
```
quicksketch, line 162: And again here, I don't think `config_diff()` and `config_menu()` are necessary here.

core/modules/config/config.module
```diff
@@ -6,6 +6,11 @@
 
 /**
  * Implements hook_menu().
+ * 
+ * @see config_set()
+ * @see config_get()
+ * @see config_clear()
+ * @see config_diff()
```
quicksketch, line 13: `config.inc` in core provides the main `config_set/get/clear()` functions. We don't do this for any other hook_menu() in Backdrop, e.g. `node_menu()` doesn't `@see` reference `node_load()` and `views_menu()` doesn't reference `views_get_view()`.

I think all of these should be removed for consistency with other hook_menu() implementations.

core/modules/config/config.admin.inc
```diff
@@ -542,6 +542,11 @@ function config_get_modified_times() {
  *
  * @return array
  *   An array of formatted strings showing the diffs between the two storages.
+ * 
+ * @see config_set()
+ * @see config_get()
+ * @see config_clear()
+ * @see config_menu()
```
quicksketch, line 549: I'm not sure these `@see` values are directly related to this. Perhaps a `@see config_diff_page()` instead of all of these?

quicksketch (commented):
I definitely like the cross-linking between the main `config_set()`, `config_get()`, and `config_clear()` functions. But a lot of the other `@see` references don't seem strongly linked enough to give them special treatment.

The `config_diff()` function is not a common config function to call, and it's exclusively part of `config.admin.inc` rather than one of the main functions provided by `config.inc()` (which is always loaded).

`config_menu()` is an implementation of `hook_menu()`. Usually we only `@see` hook_menu() entries from the page callbacks. For example `@see config_menu()` would be appropriate on the page callbacks that are referenced in that function, including: 

- `config_sync_form()`
- `config_diff_page()`
- `config_import_single_form()`
- `config_export_single_form()`
- `config_export_full_form()`
- `config_download_full_export()`

core/includes/config.inc
alanmels: I had doubts about adding those two from the very beginning, but thought would be useful to users to have a quick jump links to view them too.

I will remove the links, but I am afraid I have to create a new PR, because the original branch on which this PR was based is gone since long and I forked Backdrop several times over to the same path, so could not restore the branch.

core/includes/config.inc
alanmels: Agree.

core/includes/config.inc
alanmels: Agree

core/modules/config/config.admin.inc
alanmels: Done

core/modules/config/config.module
alanmels: done too

alanmels:
All the comments are totally accepted, but I am afraid I had to create a new PR at https://github.com/backdrop/backdrop/pull/2979, because the original branch on which this PR was based is gone since long and I forked Backdrop several times over to the same path, so could not restore the branch.

PR #2975, opened 2019-11-04
Issue #4194: Open the rewrite fieldset in Views UI if any rewriting i…

jenlampton:
Fixes https://github.com/backdrop/backdrop-issues/issues/4194

PR #2976, opened 2019-11-04
Issue #4194: Allow any fieldset to avoide collapse if default values …

### LINKED ISSUE #4194: [UX] Views: open the "rewrite results" fieldset if the results are rewritten
When I'm configuring a view, I want to know at first glance whether a field has been rewritten, without needing to click on the fieldset label to expand it. When all fieldsets are closed all the time, finding the field with the settings I need to change is time-consuming.

Screen Shot 2019-11-04 at 1 17 17 PM

---
PR w stand-alone fix: https://github.com/backdrop/backdrop/pull/2975
PR with API change: https://github.com/backdrop/backdrop/pull/2976

jenlampton:
Fixes https://github.com/backdrop/backdrop-issues/issues/4194#issuecomment-549588451

# API CHANGE:

Adds an additional `#defaults` form API property for fieldsets. When provided, a collapsed fieldset will not be collapsed if any of these defaults have been overridden.

### DIFF
--- a/core/includes/form.inc
+++ b/core/includes/form.inc
@@ -3814,8 +3814,22 @@ function form_process_fieldset(&$element, &$form_state) {
   if (!empty($element['#collapsible'])) {
     $element['#attached']['library'][] = array('system', 'backdrop.collapse');
     $element['#attributes']['class'][] = 'collapsible';
+    $add_class = TRUE;
+
     if (!empty($element['#collapsed'])) {
-      $element['#attributes']['class'][] = 'collapsed';
+      if (!empty($element['#defaults'])) {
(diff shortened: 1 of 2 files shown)


quicksketch:
I can't really see this as making sense from a generic `#type = fieldset` property. It doesn't make sense to me that any `TRUE` value in a `#defaults` property would cause the `#expanded` property to change. The approach over in https://github.com/backdrop/backdrop/pull/2975 where you just directly set `#collapsed` seems better in that it doesn't introduce any new FormAPI properties which can be in conflict with each other potentially.

Closing this in favor of https://github.com/backdrop/backdrop/pull/2975.

PR #2980, opened 2019-11-08
#4200: Allow click menus 

laryn:


core/modules/system/system.menu.inc
```diff
@@ -120,6 +121,18 @@ function system_menu_block_form($config) {
     '#default_value' => $config['toggle'],
     '#description' => t('On devices with small screens, this option reduces the menu to a toggle button, commonly known as <a href="https://en.wikipedia.org/wiki/Hamburger_button" target="_blank"><em>hamburger button</em></a>. Clicking the icon reveals or hides the menu, saving screen space.'),
   );
+  
+  $form['clickdown'] = array(
+    '#type' => 'checkbox',
+    '#title' => t('Use a click to open instead of a hover'),
```
klonos: Punctuation matters:

```suggestion
    '#title' => t('Use a click to open, instead of a hover'),
```

core/modules/system/system.menu.inc
```diff
@@ -120,6 +121,18 @@ function system_menu_block_form($config) {
...
+    '#default_value' => $config['clickdown'],
+    '#description' => t('When checked, the menu will expand when clicked rather than when hovered.'),
```
klonos: Instead of simply repeating the label of the checkbox, how about something like this?:

```suggestion
    '#description' => t('This option may work better on mobile, where there is no cursor, and menu expand/collapse relies on touch events.'),
```

laryn: @klonos I wasn't sure how much of the debate about when this improves usability to try to pack into the description. (e.g. mobile use, older users with less steady hands, avoiding "hover tunnels"...) How about this? 

> Menus that don't appear or disappear on hover can be better for usability in some cases.

quicksketch:
Thanks! Merged into 1.x for 1.15.0.

PR #2989, opened 2019-11-16
Issue #4210: Display the default search setting option only if more than one search modules are active

### LINKED ISSUE #4210: [UX] Display the default search setting option only if more than one search modules are active
**Sub-issue**

Per discussion on https://github.com/backdrop/backdrop-issues/issues/4197 AltaGrade team has started to work on getting some features of the Search Index Options module into Backdrop core. And because the module incorporates several different features, not all of which might be liked and accepted by Backdrop community, I've decided to split the project into several issues with respective PRs.

**Description of the need**

This one is for hiding the options for setting the default search module in case if only one such module is enabled. It makes perfect sense to display such a configuration option only if more than one search modules are activated.

I will later create a PR which provides the following features shown on the demo screencast below:

1. If there are more than one search modules available (for example, both `Content` and `User accounts`) then it makes the default one greyed out (disabled);

2. If there is only one search module enabled (for example, only `Content`), then it hides the `Default search` fieldset, because no need to display the only choice, which is chosen by default anyway.

Search_settings

Note that currently the [Search Index Options](https://github.com/backdrop-contrib/search_index_options) module behaves slightly differently: namely it does not effectively hide the `Default search` fieldset, but shows the single option like so:

Screen Shot 2019-11-16 at 8 57 09 PM

However, I thought to make it even better for core and completely hide that fieldset, which is useless and just unnecessarily takes place on the page.

alanmels:
Resolves https://github.com/backdrop/backdrop-issues/issues/4210

### DIFF
--- a/core/modules/search/search.admin.inc
+++ b/core/modules/search/search.admin.inc
@@ -48,13 +48,11 @@ function _search_get_module_names() {
  *
  * @ingroup forms
  *
- * @see search_admin_settings_validate()
  * @see search_admin_settings_submit()
  * @see search_admin_reindex_submit()
  */
 function search_admin_settings($form, &$form_state) {
   $config = config('search.settings');
-
 
   $form['active'] = array(
     '#type' => 'fieldset',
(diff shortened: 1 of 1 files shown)


alanmels:
The `zenci/test/php70/pr-2989` test failed with the following details:

>  /home/test/github/pr/2989/core/misc/zen-ci/init_test.sh0.00
> There is no such a script
>  /usr/bin/php /home/test/github/pr/2989/core/misc/zen-ci/run_tests.php4 sec
> Could not open input file: core/scripts/run-tests.sh
> PHP Warning:  file_get_contents(/tmp/summary): failed to open stream: No such file or directory in /home/test/github/pr/2989/core/misc/zen-ci/run_tests.php on line 37

Apparently, not because of some kind of issues in the code. So I wonder if passing all of these tests is a strict requirement or some failures could be just ignored?

I am closing the PR to re-open it after a minute as @klonos recommended to me once.

alanmels:
Closed in favor of https://github.com/backdrop/backdrop-issues/issues/3385

PR #2992, opened 2019-11-18
Issue #4214: "Reply-To" header should be capitalized

klonos:
https://github.com/backdrop/backdrop-issues/issues/4214

### DIFF
--- a/core/includes/mail.inc
+++ b/core/includes/mail.inc
@@ -151,7 +151,10 @@ function backdrop_mail($module, $key, $to, $language, $params = array(), $reply 
     $headers['From'] = $headers['Sender'] = $headers['Return-Path'] = $from;
   }
   if ($reply) {
-    $headers['Reply-to'] = $reply;
+    // The "To" bit in the "Reply-To" header needs to be capitalized.
+    // See: https://tools.ietf.org/html/rfc5322 and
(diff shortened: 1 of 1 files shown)


core/includes/mail.inc
```diff
@@ -151,7 +151,10 @@ function backdrop_mail($module, $key, $to, $language, $params = array(), $reply
     $headers['From'] = $headers['Sender'] = $headers['Return-Path'] = $from;
   }
   if ($reply) {
-    $headers['Reply-to'] = $reply;
+    // The "To" bit in the "Reply-To" header needs to be capitalized.
+    // See: https://tools.ietf.org/html/rfc5322 and
+    // https://github.com/backdrop/backdrop-issues/issues/4214
```
quicksketch: Could we move this code comment up to line 140 (`// Build the default headers`) and rephrase it cover all headers? e.g. `// Build the default headers. Note that the first letter of each word is always capitalized.` Linking to the email spec is helpful but I think we can drop the issue link, that can be pulled from the git log.

klonos: Sure. Done 👍

quicksketch:
Thanks! Looks great, merged into 1.x and 1.14.x.

PR #2998, opened 2019-11-21
Issue #4221: hide help text for min resolution if there is no limit set.

### LINKED ISSUE #4221: Don't show image minimum dimension help text if min restriction isn't set
**Description of the bug**
In image field the help text for minimum dimension restriction shows even if there isn't a limit set.

**Steps To Reproduce**

1. Set a dimension limit for "max" but none for "min" on a node image field
2. Open a node form of that type

**Actual behavior**
The help text for minimum dimension shows, without number of course.

screenshot-image-field-help

**Expected behavior**
If there's no limit set, the help text shouldn't appear.

**Additional information**

indigoxela:
Fixes https://github.com/backdrop/backdrop-issues/issues/4221


quicksketch:

PR #3003, opened 2019-11-24
Issue #4180: DateThemeTestCase fails at certain times of day

klonos:
https://github.com/backdrop/backdrop-issues/issues/4180

### DIFF
--- a/core/modules/date/tests/date_themes.test
+++ b/core/modules/date/tests/date_themes.test
@@ -99,7 +99,7 @@ class DateThemeTestCase extends BackdropWebTestCase {
     $expected = '<span class="date-display-range"><span class="date-display-start">2019-03-12T12:00+0100</span> to ';
     $expected .= '<span class="date-display-end">2019-04-12T12:00+0200</span></span>';
     $output = theme('date_display_combination', $variables);
-    $success = $this->assertEqual($expected, $output, 'Start and end date on different days with offset renders correctly.');
+    $success = $this->assertEqual($expected, $output, 'Start and end date on different days, with offset is rendered correctly.');
     if (!$success) {
       $this->verbose('<br />Expected:<pre>' . check_plain($expected) . '</pre>Got:<pre>' . check_plain($output) . '</pre>');
     }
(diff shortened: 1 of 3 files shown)


core/modules/date/tests/date_themes.test
```diff
@@ -108,18 +108,12 @@ class DateThemeTestCase extends BackdropWebTestCase {
     // with timezone identifier, with remaining days.
     $tz = 'Europe/Moscow';
     $dateTimeStart = new DateTime('now', new DateTimeZone($tz));
-    $currentHour = $dateTimeStart->format('H');
+    $currentHour = $dateTimeStart->format('G');
     if ($currentHour > 22) {
       // Make sure start and end are the same day.
       $dateTimeStart->modify('- 1 hour');
     }
-    elseif ($currentHour < 1) {
```
indigoxela, line 116: You revert the previous fix - please don't.

klonos: @indigoxela as I have explained in https://github.com/backdrop/backdrop-issues/issues/4180#issuecomment-557864981 I have reverted this **intentionally**, as **it is not needed** (and also **it does not fix the 7 vs 6 days issue**, which is actually caused by using `floor()` when calculating `$remaining_days` - not by the time diff we're trying to fix here).

Here's why:

- `$dateTimeStart` is set to now
- `$currentHour` is set to the hour of the time of `$dateTimeStart`
- if `$currentHour` is 23, then later when we set `$dateTimeEnd` to +1hr, we'd end up with the hour of `$dateTimeEnd` being 00 (which is next day). That's why the first conditional is required: it sets `$dateTimeStart` to be -1hr (so basically 22), so that `$dateTimeEnd` becomes 23hr **of the same day**.
- the next conditional which was added by your PR is not needed, because if `$currentHour` is less than 1, it will essentially be 00hrs, so when we set `$dateTimeEnd` to +1hr, it'll be 01hrs of the same day - no need to manipulate things.

I hope that this is clear now. I have tested this on my local, and confirmed that the only change required in order to fix the 6 vs 7 days issue is to simplify the calculation of `$remaining_days`, and use standard php-supplied time/date functions, instead of custom-crafted php math/calculations (which are the culprit).

indigoxela: @klonos many thanks for your explanation. So, let's give your approach a try.

quicksketch:

PR #3009, opened 2019-12-01
Issue #4227: Image fields: Do not allow min resolution to be larger than max resolution

klonos:
https://github.com/backdrop/backdrop-issues/issues/4227

### DIFF
--- a/core/modules/image/image.field.inc
+++ b/core/modules/image/image.field.inc
@@ -85,6 +85,10 @@ function image_field_instance_settings_form($field, $instance) {
   // Use the file field instance settings form as a basis.
   $form = file_field_instance_settings_form($field, $instance);
 
+  $form = array(
+    '#element_validate' => array('image_field_settings_form_validate'),
+  );
+
   // Add maximum and minimum resolution settings.
   $max_dimensions = explode('x', $settings['max_dimensions']) + array('', '');
   $max_dimensions[0] = empty($max_dimensions[0]) ? NULL : $max_dimensions[0];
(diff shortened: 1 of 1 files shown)


core/modules/image/image.field.inc
```diff
@@ -200,6 +204,23 @@ function image_field_instance_settings_form($field, $instance) {
...
+  $max_dimensions_x = $form_state['input']['instance']['settings']['max_dimensions']['x'];
+  $max_dimensions_y = $form_state['input']['instance']['settings']['max_dimensions']['y'];
+  $min_dimensions_x = $form_state['input']['instance']['settings']['min_dimensions']['x'];
+  $min_dimensions_y = $form_state['input']['instance']['settings']['min_dimensions']['y'];
+
+  // Ensure that min dimentions are not set to be bigger than max dimentions,
+  // but only if max dimentions have been specified. This allows to set min
+  // dimentions, without having to restrict max dimentions.
+  if (!empty($max_dimensions_x) && $min_dimensions_x > $max_dimensions_x || !empty($max_dimensions_y) && $min_dimensions_y > $max_dimensions_y) {
+    form_set_error('', t('The minimum image dimentions cannot be bigger than its maximum dimentions.'));
```
indigoxela: A red border around a form field would be nice.
Something like `form_set_error('instance][settings][min_dimensions', t('The min....` should do the trick.
While that's not a must, it's helpful for people not knowing, what this form error means.

And - ehem - I'm not a native speaker, but I think it's dimen**s**ions, not dimentions. :wink:

klonos: Thanks for the suggestion and the catch on that typo @indigoxela 👍 ...fixed!

indigoxela (approved):
Code looks good.

docwilmot:
Tests would be useful here.

core/modules/image/image.field.inc
```diff
@@ -85,6 +85,10 @@ function image_field_instance_settings_form($field, $instance) {
   // Use the file field instance settings form as a basis.
   $form = file_field_instance_settings_form($field, $instance);
 
+  $form = array(
+    '#element_validate' => array('image_field_settings_form_validate'),
+  );
+
```
quicksketch, line 91: Instead of replacing all of `$form`, you should append the additional element validate:

```
$form['#element_validate'] = isset($form['#element_validate']) : $form['#element_validate'] : array();
$form['#element_validate'][] = 'image_field_settings_form_validate';
```

core/modules/image/image.field.inc
```diff
@@ -200,6 +204,23 @@ function image_field_instance_settings_form($field, $instance) {
...
 }
 
+/**
+ * #element_validate handler for image_field_instance_settings_form().
+ */
+function image_field_settings_form_validate($element, &$form_state, $complete_form) {
+  $max_dimensions_x = $form_state['input']['instance']['settings']['max_dimensions']['x'];
+  $max_dimensions_y = $form_state['input']['instance']['settings']['max_dimensions']['y'];
+  $min_dimensions_x = $form_state['input']['instance']['settings']['min_dimensions']['x'];
+  $min_dimensions_y = $form_state['input']['instance']['settings']['min_dimensions']['y'];
```
quicksketch, line 214: We should be using `$form_state['values']` rather than `$form_state['input']`. The "values" are checked by Drupal to make sure corresponding elements exist, where "input" can be anything sent by the user through a `POST` request, even things that don't have form elements, so using "values" is safer.

quicksketch:
Replaced by https://github.com/backdrop/backdrop/pull/3109.

PR #3011, opened 2019-12-04
Issue #3297 Add Image Library modal to Image fields

Graham-72:
A partial solution that makes use of Ajax modal dialog.


Graham-72:
No sandbox! Will close this and re-open.

laryn:
@Graham-72 - I applied these updates locally since this sandbox isn't cooperating at the moment. Some thoughts as I go:

1. The auto-upload functionality on the image field doesn't seem to be working. (e.g. I have to save the content in order to see the image appear after selecting it, so this seems to break the Ajax uploading at the moment). Interestingly, after editing the image field settings and adjusting Alt/Title field settings it started to work as expected, even after I put the settings back to the way they were initially. Maybe this was just something on my local, **or** it required field settings to be saved after the patch was added.
2. The separate fieldset/button feels a little big in terms of page space. If it was possible to shrink all that to something more like the [original graphical mockup](https://github.com/backdrop/backdrop-issues/issues/3297#issuecomment-423606336), I think it would be preferable and would also be more consistent with the UI that is provided in the WYSIWYG interface.
3. After I expose a "File Name" / "Contains" filter on the Image Library view, I can successfully filter on a partial string for filename in this modal. However, this breaks the functionality of the WYSIWYG image editor modal -- trying to filter there reloads the modal as a stand alone page, which is obviously jarring and also loses the content that was being worked on.
4. As you noted via PM (adding here for others to see) the current status does not yet allow the image to actually be selected/inserted -- that needs to be added yet.

Other suggestions that come to mind:

1. In the exposed filters, it might be good to provide exposed sorting options for Date and Name. I was having trouble saving that on mine -- I could add sort options, but the setting to expose them didn't seem to save correctly. [That is a core bug](https://github.com/backdrop/backdrop-issues/issues/1324), and there is a workaround in the linked issue to get the settings to save.
2. For the exposed filter(s), it would be nice to have them auto-submit, perhaps, so the filters apply immediately as typed/selected. That may be personal preference so I'm interested to hear what you and others think about that. That setting saved and functions on my local, but it is a little unfriendly on the text typing field since focus keeps getting pulled away from what is being typed -- maybe that won't work if that is default behavior. I just think a list that filters on the fly feels more up to date, if it's possible in a user-friendly way.

**This is looking really good! Thanks again for all your work.**

Graham-72:
Closing this and re-opening in the hope that a sandbox will be generated now that I have closed the previous PR for this issue.

Graham-72:
Re-opening.

Graham-72:
Another step forward, with modifications to the UI, including provision for autocompletion of image file names though not in the library view. The input of full URLs is not yet working but it will accept input like `field/image/DSCF2069.jpg`

The next step will be to get selection of an image from the modals to work and to close modals.

Finally it will no doubt be necessary to add some tests - something I know little about.

Graham-72:
If you have visited this sandbox before, please be sure to empty your browser cache and hard reload `file.js` in order to view changes correctly.

klonos:
https://github.com/backdrop/backdrop-issues/issues/3297 has been merged into core and been enjoying the Image Library since Backdrop v1.15.0 back in Jan. 2020! ❤️  ...closing this old PR here.

PR #3023, opened 2019-12-23
#1519 - Removing Full HTML filter from standard install profile

### LINKED ISSUE #1519: [UX] Remove the FULL HTML format from the Standard install profile
As discussed over in https://github.com/backdrop/backdrop-issues/issues/1091, we think it would be better for users to have fewer text formats to need to understand out-of-the-box.  Since the Full HTML format is the easiest to configure (as it is now, there are no filters and no editors on it) and the name is less than ideal for a lot of use cases (some people like to add a rich-text editor, so having 'HTML' in the name is counter-intuitive for end-users who are not actually able to enter HTML).

- - - - - - - 
PR by @stpaultim = https://github.com/backdrop/backdrop/pull/3023

stpaultim:
Fixes https://github.com/backdrop/backdrop-issues/issues/1519

### DIFF
--- a/core/profiles/standard/standard.install
+++ b/core/profiles/standard/standard.install
@@ -63,42 +63,6 @@ function standard_install() {
   $filtered_html_format = (object) $filtered_html_format;
   filter_format_save($filtered_html_format);
 
-  $full_html_format = array(
-    'format' => 'full_html',
-    'name' => 'Full HTML',
-    'weight' => 1,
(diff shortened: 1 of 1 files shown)


jenlampton:
Code change looks good +1

...but it looks like we have some tests that assume this filter is there. Each test that uses this filter will either need to add it at the beginning of the test, or switch to using the other filter (if not required by the test)

PR #3025, opened 2019-12-23
#3205 Add profiles directory

stpaultim:


gitressa:
This is an older PR, but since https://github.com/backdrop/backdrop-issues/issues/3205 still is open, maybe add `Fixes URL_issue`?

profiles/README.md
```diff
@@ -0,0 +1,24 @@
...
+themes, and pre-defined configuration. They make it possible to quickly set up a 
+complex, use-specific site in fewer steps than if installing and configuring 
+elements individually. 
+
+What to place in this directory?
+--------------------------------
+
+A great way to get started with installation profiles for Backdrop CMS would be
+to copy an existing installation profile from the core/profiles directory and 
+begin to edit it to meet your sepcific needs.
```
yorkshire-pudding, line 12:
```suggestion
begin to edit it to meet your specific needs. 
```

yorkshire-pudding (changes requested):
@stpaultim - minor typo spotted.
In the issue, BWPanda suggested removing the `README.md` from `/core/profiles` as part of this. This would be in line with the other core folders. 

I would suggest incorporating more from that file so every point is covered.

PR #3026, opened 2019-12-27
#3566 - Updating code for dashboard news notifications

stpaultim:
Thanks to @serundeputy - https://github.com/backdrop/backdrop/pull/2751


stpaultim:
This is a modified version of https://github.com/backdrop/backdrop/pull/2751 - (by @serundeputy). I created a new PR to expedite review and discussion.

core/modules/dashboard/includes/block.news.inc
```diff
@@ -0,0 +1,93 @@
...
+ * @file
+ * Dashboard block displaying BackdropCMS.org news items.
+ */
+
+class DashboardNewsBlock extends Block {
+  /**
+   * {@inheritdoc}
+   */
+
+  public function __construct($plugin_name, array $data) {
```
docwilmot: If there is no new code, then the parent method is already inherited. So you dont need the constructor at all.

core/modules/dashboard/includes/block.news.inc
```diff
@@ -0,0 +1,93 @@
...
+/**
+ * @file
+ * Dashboard block displaying BackdropCMS.org news items.
+ */
+
+class DashboardNewsBlock extends Block {
+  /**
+   * {@inheritdoc}
+   */
+
```
docwilmot: No space between the comment and its target function

core/modules/dashboard/includes/block.news.inc
```diff
@@ -0,0 +1,93 @@
...
+      foreach ($news->notifications as $key => $notifications) {
+        $links[] = '<p>' . l(
+            $notifications->notification->title,
+            $notifications->notification->url
+          ) . ' - ' . $notifications->notification->summary . '</p>';
+      }
+    }
+
+    $build = array(
+      '#theme' => 'dashboard_panel__news',
```
docwilmot: I dont see this theme function actually implemented? You'd need a `hook_theme()` entry as well. And as a matter of fact, if we're adding a theme function, we should just pass the value of `$news` to the theme function, and build the whole output in the theme function, so site builders can override the output.

core/modules/dashboard/includes/block.news.inc
```diff
@@ -0,0 +1,93 @@
...
+        'This block provides a news feed from BackdropCMS.org. Its contents can not be edited.'
+      ),
+    );
+
+  }
+
+  /**
+   * {@inheritdoc}
+   */
+  public function formSubmit($form, &$form_state) {
```
docwilmot: Same as for the constructor, we dont need this.

core/modules/dashboard/includes/block.news.inc
```diff
@@ -0,0 +1,93 @@
...
+   */
+  public function getAdminPreview() {
+    return t('Displays a news feed from BackdropCMS.org.');
+  }
+
+  /**
+   * {@inheritdoc}
+   */
+  public function getContent() {
+    $feed = backdrop_http_request('https://backdropcms.org/api/notifications');
```
docwilmot, line 34: We could save the URL for the feed as a config value, in `layout.settings.json` so that we can have proper tests.

core/modules/dashboard/tests/dashboard.test
```diff
@@ -111,6 +111,9 @@ class DashboardTest extends BackdropWebTestCase {
     // The manage user accounts link is present.
     $this->assertLink('Manage user accounts', 0, 'The Manage user accounts link was found.');
 
+    // The backdrop-news block was found.
+    $this->assertText('Backdrop News', 'Backdrop News block found.');
```
docwilmot, line 115: If the feed URL was in settings, we could have a bit more tests, like 
- load the URL in the test, check its contents, and verify that the block displays the feed contents
- change the saved URL in settings, then test to see what the block displays (the proper "No news at this time", or if it blows up the site if the feed is not found for example)

Maybe not absolutely necessary, but would harden our testing a bit more.

core/themes/seven/css/style.css
```diff
@@ -1233,6 +1233,9 @@ div.filter-options select {
 .block-dashboard-updates .block-title {
   background-image: url('../images/info-circle--black--64.png');
 }
+.block-dashboard-news .block-title {
+  background-image: url('../images/bell--black--64.png');
```
docwilmot, line 1237: this image is missing

stpaultim:
Thanks @docwilmot - Most of this code was from the PR by @serundeputy. I'll edit my PR with your comments in mind.

core/themes/seven/css/style.css
stpaultim: Not any more.

core/modules/dashboard/includes/block.news.inc
stpaultim: Deleted

core/modules/dashboard/includes/block.news.inc
stpaultim: Deleted

core/modules/dashboard/includes/block.news.inc
stpaultim: Fixed

core/modules/dashboard/includes/block.news.inc
jenlampton: I don't think `layout.settings.json` is the right place for it, as that's where we store settings for layouts (not for individual blocks). Why do we need it in config for tests? Could we use a constant instead?

core/modules/dashboard/tests/dashboard.test
jenlampton: > load the URL in the test, check its contents, and verify that the block displays the feed contents

We can do this with a hard-coded URL, too. :)

> change the saved URL in settings, then test to see what the block displays

Hm, this is probably something we should test manually before release, but since it's not something anyone running backdrop could do (yet) I don't see why we need a test for it (yet). If we ever make it possible to change the feed URL we can add that test to go with that feature at the time we add it.

core/modules/dashboard/includes/block.news.inc
```diff
@@ -0,0 +1,76 @@
...
+    $links = array();
+    if ($feed->code === '200') {
+      $news = json_decode($feed->data);
+      foreach ($news->notifications as $key => $notifications) {
+        $links[] = '<p>' . l(
+            $notifications->notification->title,
+            $notifications->notification->url
+          ) . ' - ' . $notifications->notification->summary . '</p>';
```
quicksketch: Although this is coming from backdropcms.org, we should do some basic filtering here. I suggest wrapping in a `filter_xss()` check.

core/modules/dashboard/includes/block.news.inc
quicksketch: Yeah, `dashboard.settings.json` would be a better place. But I agree that we should have this in config. We do the same thing for update module, that allows us to use a different location for the XML feed (such as a local copy). It might also be a good idea to be able to disable this feed entirely during test runs so we're not hitting backdropcms.org on every test.

core/modules/dashboard/includes/block.news.inc
quicksketch:
```suggestion
          ) . ' - ' . filter_xss($notifications->notification->summary) . '</p>';
```

stpaultim: Done

quicksketch:
Thanks! Merged into 1.x for 1.15.0.

PR #3029, opened 2019-12-29
#4156 - Changing default on comments

stpaultim:
Turning off comments by default on new content types.


quicksketch:
Thanks @stpaultim! Merged into 1.x for 1.15.0.

PR #3031, opened 2020-01-02
Issue #4113 - Better default positioning for breadcrumbs

### LINKED ISSUE #4113: [UX] Better defaults position for breadcrumbs on installation of new sites.
**Revised Description** - UPDATED: Aug 14,2020 by @stpaultim 

The original issue suggested changing the order of title, breadcrumbs, and tabs. The only portion of this that I've noticed real support for (and the most urgent part of this in my own mind) is moving the breadcrumbs above the title. We can revisit other ideas for improving the layout in the future. 

The original PR changed the core file: core/modules/layout/config/layout.layout.default.json which meant that this change would effect all news sites regardless of install profile.

**Reasons for change:**

The default theme in Backdrop CMS standard install profile is Basis and positioning the breadcrumbs below the Title doesn't seem practical or appealing for Basis. While there are definitely use cases for positioning the breadcrumbs where they are, I believe that the default settings should be configured to look good and be useful in the default theme. Site admins using other themes and/or creating their own themes can easily move the breadcrumbs to the position that works best for them. The goal here is to make Basis look good and work good out of the box. 

The approach in the existing PR is simply to edit the install file in the Standard Install Profile. This should only effect new sites using the Standard Install Profile and not have any effect on existing sites. 

We could make the same changes directly here: core/modules/layout/config/layout.layout.default.json, but in this case these changes would effect all new sites and may have an unintended effect on users who have created their own custom install profile. 

PROPOSED SOLUTION:

Based upon some feedback, I've modified the PR to only effect the standard install profile. With this PR, the standard install profile will modify the layout.layout.default.json config file to do the following:

1) It implements the "Page Title Combo" block, instead of the default hard-coded Title on the page. I don't see any other way of getting the breadcrumbs above the title in the install profile. 
2) It positions the breadcrumbs above the title. 

CURRENT STATUS OF ISSUE:

Looking for feedback on this new approach. Also, need review of PR to check my syntax and variables. 

---
Below this issue is the original issue and proposed solutions, some of which is no longer relavent.

**ORIGINAL - Description of the need**

Backdrop out of the box starts with a page layout that puts breadcrumbs below the page title. This seems REALLY weird and unhelpful (even though it is easy to fix). 

I think that we need a better place to put the default breadcrumbs block than the current location and current layouts do not have a suitable place. We can't move the default position of the block to the header, because it just doesn't work and we can't move the default position of the title, because it's hard-coded into core templates. 

While it is relatively easy to solve this problem by turning off breadcrumbs or adding your own title block in another region, I think it's poor user experience to require everyone to make this change and sets a poor first impression of Backdrop CMS (out of the box) to find your breadcrumbs in such a strange position. 

**ORIGINAL - Proposed solution**

I propose adding another region (need ideas on what to name this region) to all core layouts between the header and the default title. Adding a region will not break existing sites and provides the opportunity to move the default breadcrumb positioning above the page title where most people would expect it to be. 

1) We could turn off the breadcrumbs block by default. Requiring users to turn it on would remove the poor first impression, but also force new users to figure out how to place breadcrumbs properly on their own (if they want breadcrumbs). 
2) Placing a default TITLE block in the content region for new sites. This would turn off the current title block and by default the breadcrumbs would be above the page title. 
3) Create new sites with a custom Flexible Layout Template that resolves this problem by default, instead of the current core layouts. 
4) **(ADDED)** Add a single NEW layout that can used in the default install profile to address these issues. 

**Additional information**

1) I don't think this is a critical issue, but if this solution makes sense, I think would be easy to implement and improve the out of the box experience for new users. 
2) I'm not convinced that this is the right solution, but would like to start a discussion. 
3) I think that this would be a good easy issue for a beginner to create a PR. If I am correct, we'd just need to add a region to all core layouts. 
4) I think that an argument can be made that there are lots of other things that a user might want between the header and the page title for which these new regions would be helpful. 
5) Best practices in breadcrumbs design? = https://www.nngroup.com/articles/breadcrumbs/

Your_first_post____Demo

stpaultim:
Fixes backdrop/backdrop-issues#4113

### DIFF
--- a/core/profiles/standard/standard.install
+++ b/core/profiles/standard/standard.install
@@ -106,6 +106,41 @@ function standard_install() {
   theme_enable(array($default_theme));
   theme_disable(array('stark'));
 
+  // Enables the title_combo block and positions breadcrumbs
+  // immediately above title_combo block.
+
+  $title_combo_uuid = 'a549d0bc-51f7-46b0-9b41-653788410dd4';
+  $title_combo_block = array(
+    'content.' . $title_combo_uuid => array(
+      'plugin' => 'system:page_components:title_combo',
+      'data' => array(
+        'module' => 'system',
+        'delta' => 'page_components',
+        'settings' => array(
+          'title_display' => 'default',
+          'title' => '',
+          'style' => 'default',
+          'block_settings' => [],
+          'contexts' => [],
+          'title_tag' => 'h1',
+          'title_classes' => 'page-title',
+          'tab-type' => 'both',
+        ),
+        'uuid' => $title_combo_uuid,
+        'style' => array(
+          'plugin' => 'default',
+          'data' => array(
+            'settings' => array(
+              'classes' => '',
+            ),
+          ),
+        ),
+      ),
+    ),
+  );
+  config_set_multiple('layout.layout.default', $title_combo_block );
+  config_set('layout.layout.default','positions.top', array('09f11b4e-877e-45ab-9000-a4c6254faa04', 'a549d0bc-51f7-46b0-9b41-653788410dd4'));
+
   // Set front page to "home".
   config_set('system.core', 'site_frontpage', 'home');
   backdrop_flush_all_caches();
--- a/core/modules/layout/config/layout.layout.default.json
+++ b/core/modules/layout/config/layout.layout.default.json
@@ -19,10 +19,6 @@
       "84c43df4-1165-4faf-8700-c362a7153c0b",
       "13b2f325-af26-40bb-8c61-97fe1d93f19f"
     ],
-    "top": [
-      "09f11b4e-877e-45ab-9000-a4c6254faa04",
-      "a549d0bc-51f7-46b0-9b41-653788410dd4"
-    ],
     "content": [
       "default"
     ],
@@ -30,7 +26,9 @@
       "cd4d033d-3f67-4191-8655-85ee758877c0"
     ],
     "sidebar": [],
-    "title": []
+    "top": [
+      "09f11b4e-877e-45ab-9000-a4c6254faa04"
+    ]
   },
   "content": {
     "84c43df4-1165-4faf-8700-c362a7153c0b": {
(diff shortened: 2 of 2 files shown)


stpaultim:
Fixes #4113

core/profiles/standard/standard.install
```diff
@@ -106,6 +106,41 @@ function standard_install() {
   theme_enable(array($default_theme));
   theme_disable(array('stark'));
 
+  // Enables the title_combo block and positions breadcrumbs
+  // immediately above title_combo block.
+
+  $title_combo_uuid = 'a549d0bc-51f7-46b0-9b41-653788410dd4';
```
klonos, line 112: Perhaps we should be generating a random UUID here, instead of hard-coding it:

```suggestion
  $uuid = new Uuid();
  $title_combo_uuid = $uuid->generate();
```

core/profiles/standard/standard.install
```diff
@@ -106,6 +106,41 @@ function standard_install() {
...
+            'settings' => array(
+              'classes' => '',
+            ),
+          ),
+        ),
+      ),
+    ),
+  );
+  config_set_multiple('layout.layout.default', $title_combo_block );
+  config_set('layout.layout.default','positions.top', array('09f11b4e-877e-45ab-9000-a4c6254faa04', 'a549d0bc-51f7-46b0-9b41-653788410dd4'));
```
klonos, line 142:
```suggestion
  config_set('layout.layout.default', 'positions.top', array('09f11b4e-877e-45ab-9000-a4c6254faa04', $title_combo_uuid));
```

core/profiles/standard/standard.install
stpaultim: > Perhaps we should be generating a random UUID here, instead of hard-coding

Maybe. I didn't think so, because we are not generating random uuids in our current core layouts. They are hard-coded. See:

core/modules/layout/config/layout.layout.default.json
core/modules/layout/config/layout.layout.admin_default.json

Is this a security issue or simply a tool to prevent conflicting ids? If this is just about avoiding conflicts, then I don't see that a new random id is necessary in every case.

stpaultim: I think I would defer to a core committer regarding this change. I don't object to this change. It's not clear to me if this is helpful or not.

quicksketch:
Closing/Reopening for new sandbox.

PR #3033, opened 2020-01-02
Issue #4254: Dashboard News block follow-ups. Add cache, customizable URL.

### LINKED ISSUE #4254: 1.15: News Block Followup: Switch URL to a config setting and cache the news updates
**Description of the need**
Follow up issue to: https://github.com/backdrop/backdrop-issues/issues/3566

We added a block to the Dashboard that will ingest a feed of public notifications from BackdropCMS.org. @quicksketch asked for the following followup issues:

There are a few other things that I think should happen, but can be done in a follow-up:

- The feed URL should come from a config setting in `dashboard.settings.json`.
- We should disable checking for updates at all in all test runs, probably by blanking out that setting in the config file.
- We should add a cache to the HTTP request and update the feed on cron runs rather than doing it on-the-fly.
- Add a test that loads the URL, checks its contents, and verifies that the block displays the feed contents

---

PR https://github.com/backdrop/backdrop/pull/3033

quicksketch:
Fixes https://github.com/backdrop/backdrop-issues/issues/4254.

~Not 100% ready. We should update tests to set the value to FALSE to suppress hitting BackdropCMS.org on cron runs.~ Tests fixed and now checking the feed is disabled during test runs.

### DIFF
--- a/core/modules/dashboard/includes/block.news.inc
+++ b/core/modules/dashboard/includes/block.news.inc
@@ -31,15 +31,14 @@ class DashboardNewsBlock extends Block {
    * {@inheritdoc}
    */
   public function getContent() {
-    $feed = backdrop_http_request('https://backdropcms.org/api/notifications');
+    $news = $this->getNews();
     $links = array();
-    if ($feed->code === '200') {
-      $news = json_decode($feed->data);
-      foreach ($news->notifications as $key => $notifications) {
+    if ($news) {
+      foreach ($news as $key => $notifications) {
         $links[] = '<p>' . l(
-            $notifications->notification->title,
-            $notifications->notification->url
-          ) . ' - ' . filter_xss($notifications->notification->summary) . '</p>';
+            $notifications['title'],
+            $notifications['url']
+          ) . ' - ' . filter_xss($notifications['summary']) . '</p>';
       }
     }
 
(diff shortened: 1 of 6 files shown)


core/modules/dashboard/includes/block.news.inc
```diff
@@ -73,4 +72,53 @@ class DashboardNewsBlock extends Block {
...
+   * with FALSE as the saved value until the next time the cache is cleared or
+   * cron updates the news list.
+   *
+   * @return FALSE|array
+   */
+  public static function getNews() {
+    $cache = cache()->get('dashboard_news_feed');
+    // If no cache is set at all, fetch results immediately.
+    if ($cache === FALSE) {
+      $news = DashboardNewsBlock::refreshNewsFeed();
```
jenlampton, line 90: Do we have to use this syntax here? It looks strange compared to the rest of the code in core. Is this the same as `$this->refreshNewsFeed();`?

jenlampton: Hm. I see we are using the same `syntax in dashboard_cron()`. I might be tempted to add a procedural wrapper function if I thought this was going to be something contrib might leverage. If that turns out to be the case, we can add one later.

quicksketch:
Super! Merged into 1.x for 1.15.0.

PR #3036, opened 2020-01-08
Issue #4263: Convert {} to [] for PHP 7.4.

### LINKED ISSUE #4263: PHP 7.4: Array and string offset access syntax with curly braces is deprecated
I get the following warning when running Backdrop on PHP 7.4:

> Deprecated: Array and string offset access syntax with curly braces is deprecated in /{SNIP}/core/modules/views/handlers/views_handler_filter_string.inc on line 275

I'm pretty sure we don't officially support PHP v7.4 yet, but considering the fix is to use a syntax that works in older versions too then I don't see why we can't fix this easily now.

https://stackoverflow.com/questions/12769982/reference-what-does-this-error-mean-in-php/59158847#59158847 says that it's just a matter of replacing `{}` with `[]` when accessing string offsets or array elements.

---

PR: https://github.com/backdrop/backdrop/pull/3036

ghost:
Fixes https://github.com/backdrop/backdrop-issues/issues/4263

### DIFF
--- a/core/modules/search/search.extender.inc
+++ b/core/modules/search/search.extender.inc
@@ -208,7 +208,7 @@ class SearchQuery extends SelectQueryExtender {
       }
       $phrase = FALSE;
       // Strip off phrase quotes.
-      if ($match[2]{0} == '"') {
+      if ($match[2][0] == '"') {
         $match[2] = substr($match[2], 1, -1);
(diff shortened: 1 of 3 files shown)


klonos (approved):
LGTM 👍

quicksketch:

PR #3037, opened 2020-01-13
Issue #4270: Fix regression when adding blocks with file fields

### LINKED ISSUE #4270: Possible regression in 1.15.x-dev when adding blocks with file fields
**Description of the bug**

When newly adding certain types of blocks to a layout, they don't appear on the "Manage blocks" page after clicking "Add block".

Also the message "This form has unsaved changes. ..." won't show.
After flushing the layout cache via admin menu, or even simply reloading the layout form page, both show in the form (block and message).

**Steps To Reproduce**

1. Go to a layout and add a hero block
2. Save the block

Please note: not all types of blocks seem affected. It only affects block types with a file field like the core Hero block type.

Nothing in dblog, nothing in error_log.

**Expected behavior**
Block and message should show up.

**Additional information**

- Backdrop CMS version: 1.15.x-dev (latest)
- PHP version 7.2

---

PR by @indigoxela: https://github.com/backdrop/backdrop/pull/3037

indigoxela:
Fixes https://github.com/backdrop/backdrop-issues/issues/4270

`$browserContainer` is set but has zero length.

### DIFF
--- a/core/modules/file/js/file.js
+++ b/core/modules/file/js/file.js
@@ -203,7 +203,7 @@ Backdrop.file = Backdrop.file || {
    */
   dialogCloseEvent: function(e, dialog, $element) {
     var $browserContainer = $element.find(".file-browser");
-    if ($browserContainer) {
+    if ($browserContainer.length > 0) {
       // These two variables are set server-side when submitting the dialog, in
       // file_managed_file_browser_submit().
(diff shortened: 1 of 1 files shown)


quicksketch:
Thanks! Great fix. Merged into 1.x for 1.15.

PR #3038, opened 2020-01-15
Issue #3174: Base/sub install profiles.

### LINKED ISSUE #3174: Base/sub install profiles
**Update:** It was decided that this feature won't go into Backdrop core, but it is easy enough to create sub-profiles without any core changes. Sub-profiles just need to run the base profile manually as the first step in their install hook, and list the base profile's dependencies in their info file. See the attached test profile here as an example: https://github.com/backdrop/backdrop-issues/issues/3174#issuecomment-629072580

---

I'd like to propose adding the ability for install profiles to define a 'base profile'. Here's the equivalent Drupal issue for this feature request: https://www.drupal.org/project/drupal/issues/1356276

The way I see this working is that an install profile defines a 'base profile' in it's .info file (like themes can define a 'base theme'), then when that install profile is used to setup a new website, the base profile is run first, then the regular profile.

This would make it easier, for example, in https://github.com/backdrop/backdrop-issues/issues/3126 to create additional install profiles based off the Standard one without having to re-invent the wheel as it were.

Also, my use case (admittedly probably only useful to me) is wanting to create install profiles that setup packaged websites for my clients. Clients can purchase the basic package, which sets up a website the way I like to build them for my clients (content types, configuration, etc.), or they can purchase the advanced package, which (in addition the the basic package) sets up additional features (extra modules, advanced configuration, etc.). This would save me having to duplicate my basic install profile.

---

Advocate: @BWPanda

---

PR by @BWPanda: https://github.com/backdrop/backdrop/pull/3038

ghost:
Fixes https://github.com/backdrop/backdrop-issues/issues/3174

### DIFF
--- a/core/includes/install.core.inc
+++ b/core/includes/install.core.inc
@@ -165,8 +165,8 @@ function install_state_defaults() {
     'profile_info' => array(),
     // An array of available installation profiles.
     'profiles' => array(),
-    // An array of base profiles of the chosen installation profile
-    // (e.g. chosen profile > base_profiles[0] > base_profiles[1] > etc.)
+    // An array of base profiles of the chosen installation profile (e.g.,
+    // chosen profile > base_profiles[0] > base_profiles[1] > etc.).
     'base_profiles' => array(),
     // An array of server variables that will be substituted into the global
     // $_SERVER array via backdrop_override_server_variables(). Used by
@@ -1577,8 +1577,8 @@ function install_profile_modules(&$install_state) {
   $required = array();
   $non_required = array();
   // Although profile module(s) are marked as required, they need to go after
-  // every dependency, including non-required ones. So clear their required
-  // flag for now to allow them to install late.
+  // every dependency, including non-required ones. So clear their required flag
+  // for now to allow them to install late.
   $files[$install_state['parameters']['profile']]->info['required'] = FALSE;
   if (!empty($install_state['base_profiles'])) {
     foreach ($install_state['base_profiles'] as $base_name => $base_profile) {
--- a/core/includes/install.core.inc
+++ b/core/includes/install.core.inc
@@ -543,6 +543,17 @@ function install_tasks($install_state) {
   }
   $display_select_profile = $hidden_profile_count > 1 && count($install_state['profiles']) != 1;
 
+  // Generate list of base profiles if necessary.
+  if (!empty($install_state['parameters']['profile'])) {
+    $langcode = !empty($install_state['parameters']['langcode']) ? $install_state['parameters']['langcode'] : NULL;
+    $info = install_profile_info($install_state['parameters']['profile'], $langcode);
+    while (isset($info['base'])) {
+      $base_name = $info['base'];
+      $info = install_profile_info($base_name, $langcode);
+      $install_state['base_profile'][$base_name] = $info;
+    }
+  }
+
   // Start with the core installation tasks that run before handing control
   // to the installation profile.
   $tasks = array(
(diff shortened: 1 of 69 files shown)


core/includes/install.core.inc
```diff
@@ -1531,10 +1576,15 @@ function install_profile_modules(&$install_state) {
   // the modules.
   $required = array();
   $non_required = array();
-  // Although the profile module is marked as required, it needs to go after
-  // every dependency, including non-required ones. So clear its required
-  // flag for now to allow it to install late.
+  // Although profile module(s) are marked as required, they need to go after
+  // every dependency, including non-required ones. So clear their required
+  // flag for now to allow them to install late.
```
```suggestion
  // Although profile module(s) are marked as required, they need to go after
  // every dependency, including non-required ones. So clear their required flag
  // for now, to allow them to install late.
```

core/modules/system/system.module
```diff
@@ -2824,22 +2824,44 @@ function system_get_info($type, $name = NULL) {
...
 function _system_rebuild_module_data() {
+  require_once BACKDROP_ROOT . '/core/includes/install.inc';
+
   // Find modules
   $modules = backdrop_system_listing('/^' . BACKDROP_PHP_FUNCTION_PATTERN . '\.module$/', 'modules', 'name', 0);
 
-  // Include the installation profile in modules that are loaded.
+  // Include the installation profile(s) in modules that are loaded.
+  // This will include the current profile, then loop through any base profiles
+  // as well.
```
klonos:
```suggestion
  // Include the installation profile(s) in modules that are loaded. This will
  // include the current profile, then loop through any base profiles as well.
```

core/modules/system/system.module
```diff
@@ -2824,22 +2824,44 @@ function system_get_info($type, $name = NULL) {
...
+    $weight--;
+
+    // Get base profile, if any.
+    $profile_info = install_profile_info($profile);
+    if (isset($profile_info['base profile'])) {
+      $profile = $profile_info['base profile'];
+      $has_base = TRUE;
+    }
+    else {
+      // Set $profile back to the current one for use later in this function.
```
klonos, line 2865:
```suggestion
      // Set $profile back to the current one, for use later in this function.
```

core/includes/install.core.inc
```diff
@@ -165,6 +165,9 @@ function install_state_defaults() {
     'profile_info' => array(),
     // An array of available installation profiles.
     'profiles' => array(),
+    // An array of base profiles of the chosen installation profile
+    // (e.g. chosen profile > base_profiles[0] > base_profiles[1] > etc.)
```
```suggestion
    // An array of base profiles of the chosen installation profile (e.g.:
    // chosen profile > base_profiles[0] > base_profiles[1] > etc.).
```

core/modules/system/system.module
ghost: I disagree with the use of a comma here.

core/includes/install.core.inc
ghost: I changed the ':' to a ',' to match other instances of 'e.g.' in this file.

core/modules/system/system.module
klonos: OK, consider this (when reading):

> Set $profile back to the current one, for use later in this function.

vs:

> Set $profile back to the current one for use, later in this function.

...in the last version, the person is actually reading/understanding this:

> Later in this function, set $profile back to the current one for use.

ghost: I don't think we need a comma at all. I think it reads fine as is.

PR #3041, opened 2020-01-17
Issue #4073: Add module_set_weight() function.

### LINKED ISSUE #4073: Add module_set_weight() API function
**Description of the need**
Right now when we need to set module weight (which determines hook execution order), a module has to manually update its weight in the system table in hook_install() like this:

```
/**
 * Implements hook_install().
 */
function contextual_install() {
  // Weight must be greater than that of views module.
  db_query("UPDATE {system} SET weight = 11 WHERE name = 'contextual'");
}
```

**Proposed solution**
Drupal 8 provided a [module_set_weight()](https://api.drupal.org/api/drupal/core%21includes%21module.inc/function/module_set_weight/8.2.x) function. Though ironically it's intended to be deprecated and possibly removed.

We could just keep using SQL queries, but providing an API function could allow us to more easily switch the system table to a config file later.

ghost:
Fixes https://github.com/backdrop/backdrop-issues/issues/4073

### DIFF
--- a/core/includes/module.inc
+++ b/core/includes/module.inc
@@ -1192,9 +1192,9 @@ function backdrop_alter($type, &$data, &$context1 = NULL, &$context2 = NULL, &$c
  * @param string $module
  *   The name of the module (without the .module extension).
  * @param integer $weight
- *   The weight to set as an integer. Defaults to 0.
+ *   The weight to set, as an integer.
  */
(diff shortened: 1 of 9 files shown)


quicksketch:
Thanks @BWPanda! Merged into 1.x and 1.17.x.

PR #3048, opened 2020-01-24
Issue #2353: Dblog clear form

### LINKED ISSUE #2353: [UX] move the clear log messages button to bottom of dblog messages page
The question here is: **Where should we put this "Clear Log Messages" Button?** 

There are several UX improvements we are addressing here:
* avoid accidental clicking/deleting FOR ADMINS + DEVS
* prevent confusion from too many available options above the listing FOR ADMINS + DEVS
* reduce the amount of scrolling needed to read the listing FOR ADMINS + DEVS
* make this interface more consistent with the rest of core FOR ADMINS + DEVS
* reduce #of clicks to delete messages ? FOR DEVS (only)

We will eventually want to remove the fieldset around the filters to reduce the number of clicks needed to filter the listing, too. (This will also make this interface more consistent with the rest of core). Our decision here should take that into account. 

I indicated the ROLE of the person completing each task above because in Backdrop we put the needs of ADMINS above the needs of the needs of DEVeloperS. 

Options as follows.

1) Leave it where it is, at the top of the page, buried in a fieldset:
Screen Shot 2020-02-20 at 10 57 54 AM

2) Move it to the bottom of the page, but NOT in a fieldset: 
screen shot 2018-04-22 at 11 22 12 am

3) Move it to another page, accessed by a tab, like Drupal 8 did:
screenshot-drupal8

4) Combine the two fieldsets for filtering the current list, and deleting all items to save space
(no screenshot available)

5) move that button to the Log settings page: `/admin/config/development/logging`?

### Objections to 1:

> The clear log messages is inside a fieldset. I understand that we don't want people clicking this by accident, but I'm not sure a fieldset is the way to deal with the problem

> I don't know of anywhere else in core where we make users go to the listing page in order to delete items. The delete option is always on another page/tab.

### Objections to option 2

> [DEVS will] need to scroll down this page 

> it might seem like a "save" button or something people are supposed to do here, and might increase the probability of people doing it by accident.

### Objections to option 3

> After doing UX reviews/studies, D8 have decided to remove all tabs and convert them to action links

> introduction of [a new] UI pattern of having an action live as a primary/secondary tab. It just does not feel right UX-wise to me. 

> There have been extensive discussions and user studies in Drupal-land, concluding that things that are actions (such as the "delete" operation), should be buttons/links and not tabs; tabs are being reserved for navigational operations.

### Objections to option 4

> taking [the Clear Log Messages button] out of the fieldset and then having it anywhere near the buttons for the filters (above, below, to the right) would also be confusing [for ADMINS and DEVS]

> If the button can be moved away from the top of the page (above what you're trying to access 99%* of the time), then I agree that a fieldset is not needed. But as long as the button's above the content, it should be in a collapsed fieldset.

### Objections to option 5

>  it's convenient for DEVS to have here - right where you are anyway when you are debugging

> Definitely wrong place with too long path, we need clear log button on same screen where we read it.

---

Original Issue

---

The 'Recent log messages' page has several UI patterns that are unusual for core. 

The clear log messages is inside a fieldset. I understand that we don't want people clicking this by accident, but I'm not sure a fieldset is the way to deal with the problem, especially since it separates the filters at the top of the page with the list the filters are acting on, below the fieldset.  How about we 1) remove the fieldset and 2) move the button to the bottom of the page. 

~~First, the filters for the log messages are hidden away in a fieldset. This means you have to open the fieldset to filter the messages.  Second, the `Severity` filter is a multi-select widget, which is nearly unusable. This should be replaced with checkboxes instead.~~ See https://github.com/backdrop/backdrop-issues/issues/3051 for these issues

----
~PR: https://github.com/backdrop/backdrop/pull/2142~
PR by @opi: https://github.com/backdrop/backdrop/pull/2148 
Alternate PR by @indigoxela: https://github.com/backdrop/backdrop/pull/3048 (closed in favor of a contrib module)
Alternate PR by @klonos: https://github.com/backdrop/backdrop/pull/3055

indigoxela:
Fixes https://github.com/backdrop/backdrop-issues/issues/2353

Alternative approach: move form to an extra tab.

### DIFF
--- a/core/modules/dblog/dblog.admin.inc
+++ b/core/modules/dblog/dblog.admin.inc
@@ -10,7 +10,6 @@
  * Messages are truncated at 56 chars. Full-length messages can be viewed on the
  * message details page.
  *
- * @see dblog_clear_log_form()
  * @see dblog_event()
  * @see dblog_filter_form()
  * @see dblog_menu()
@@ -30,7 +29,6 @@ function dblog_overview() {
     WATCHDOG_DEPRECATED => 'dblog-deprecated',
   );
 
-  $build['dblog_clear_log_form'] = backdrop_get_form('dblog_clear_log_form');
   $build['dblog_filter_form'] = backdrop_get_form('dblog_filter_form');
 
   $header = array(
@@ -88,6 +86,16 @@ function dblog_overview() {
   );
   $build['dblog_pager'] = array('#theme' => 'pager');
 
+  return $build;
+}
+
+/**
+ * Page callback to clear all dblog messages.
+ *
+ * @see dblog_clear_log_form()
+ */
+function dblog_clear() {
+  $build['dblog_clear_log_form'] = backdrop_get_form('dblog_clear_log_form');
   return $build;
 }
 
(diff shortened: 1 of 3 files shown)


indigoxela:
Closed in favor of a contrib module.

PR #3050, opened 2020-01-25
Issue #4283: Remove broken javascript disabled setting

### LINKED ISSUE #4283: Disabling JavaScript with Views doesn't work properly
This might be related to https://github.com/backdrop/backdrop-issues/issues/3734, but since I was not sure I decided to file a separate bug report.

**Description of the bug**
Disabling JavaScript on Views' Advanced settings page (http://view.docksal/admin/structure/views/settings/advanced) does not have full effect: it **does**  open the option links on a regular window (instead of a pop-up modal one), but it **does not** close the option page to return to the initial page.

**Steps To Reproduce**

To reproduce the behavior:
1. Go to `/admin/structure/views/settings/advanced`, select the `Disable JavaScript with Views` option and press on the `Save Settings` button.
2. Go to any views configuration page (for example to `Administer content (Content)`  at `admin/structure/views/view/node_admin_content`).
3. Click on any option (for example Path :/admin/content/node), which will take you to `/admin/structure/views/nojs/display/node_admin_content/page/path`.
4. Click on Apply or Cancel button to be stuck at the same page.

views-ajax-fail

**Expected behavior**
Expected behavior was returning to previous screen as on Drupal 7:

works-on-d7

indigoxela:
Fixes https://github.com/backdrop/backdrop-issues/issues/4283

### DIFF
--- a/core/modules/views/config/views.settings.json
+++ b/core/modules/views/config/views.settings.json
@@ -1,6 +1,5 @@
 {
     "_config_name": "views.settings",
-    "no_javascript": false,
     "exposed_filter_any_label": false,
     "show_additional_queries": false,
     "sql_signature": false,
--- a/core/modules/views/views.install
+++ b/core/modules/views/views.install
@@ -186,13 +186,13 @@ function views_update_1006() {
 }
 
 /**
+ * Remove obsolete variable from config.
+ */
+function views_update_1007() {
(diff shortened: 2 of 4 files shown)


quicksketch:
Thanks @indigoxela! Merged into 1.x and 1.16.x.

PR #3052, opened 2020-01-28
Issue #4225: Fix broken per vocabulary permission

### LINKED ISSUE #4225: Per vocabulary permissions do not appear to do anything
giving a role permission to edit terms in a particular vocabulary does not allow users with that role to edit terms in the vocabulary UNLESS they also have the "administer vocabularies and terms" permission, which defeats the purpose of having per-vocabulary permissions.

my scenario:

there are permissions defined for editing and deleting each defined vocabulary, on the permissions page (/admin/config/people/permissions).

i have checked the edit permission on a vocabulary for a defined role "content editor" (i do NOT want to give them access to any other vocabularies), and saved. i have cleared all caches.

but when i log in as a user with the "content editor" role, and i try to access the term edit form for a term in that vocabulary (/taxonomy/term/47/edit), i get an access denied from backdrop.

hopefully this is enough information. if not, please let me know what more i can provide.

---
~~PR by @indigoxela: https://github.com/backdrop/backdrop/pull/3052 (rename permission, closed in favor of the more lightweight change)~~
PR by @quicksketch: https://github.com/backdrop/backdrop/pull/3067 (update access check)

indigoxela:
Fixes https://github.com/backdrop/backdrop-issues/issues/4225

Closed in favor of the alternate PR.

### DIFF
--- a/core/modules/taxonomy/taxonomy.module
+++ b/core/modules/taxonomy/taxonomy.module
@@ -66,7 +66,7 @@ function taxonomy_permission() {
 function taxonomy_vocabulary_list_permissions($vocabulary) {
   $permissions = array(
     'update terms in ' . $vocabulary->machine_name => array(
-      'title' => t('%vocabulary: Update terms', array('%vocabulary' => $vocabulary->name)),
+      'title' => t('%vocabulary: Edit terms', array('%vocabulary' => $vocabulary->name)),
     ),
     'delete terms in ' . $vocabulary->machine_name => array(
       'title' => t('%vocabulary: Delete terms', array('%vocabulary' => $vocabulary->name)),
--- a/core/modules/taxonomy/taxonomy.module
+++ b/core/modules/taxonomy/taxonomy.module
@@ -65,8 +65,8 @@ function taxonomy_permission() {
  */
 function taxonomy_vocabulary_list_permissions($vocabulary) {
   $permissions = array(
-    'edit terms in ' . $vocabulary->machine_name => array(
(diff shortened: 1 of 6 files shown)


core/modules/taxonomy/taxonomy.module
```diff
@@ -65,8 +65,8 @@ function taxonomy_permission() {
  */
 function taxonomy_vocabulary_list_permissions($vocabulary) {
   $permissions = array(
-    'edit terms in ' . $vocabulary->machine_name => array(
-      'title' => t('%vocabulary: Edit terms', array('%vocabulary' => $vocabulary->name)),
```
ghost, line 69: The array key needs to change (obviously), but I don't think the 'title' should change. I like the current matching of the '*Tags*: Edit terms' wording with the 'Edit' tab on the term page. Renaming this to 'Update' would create a discrepancy I think...

ghost (changes requested):
One small change I think needs to be removed, otherwise very nice code-wise!

core/modules/taxonomy/taxonomy.module
ghost: What I mean is that the 'Edit terms' permission directly controls a user's access to the 'Edit' tab. So calling the permission 'Update terms' and having it control access to the 'Edit' tab could be confusing...

indigoxela: Good point, let's stick with "Edit terms" in titles.

PR #3054, opened 2020-01-29
Issue #4289: Add cache clear buttons to the theme overview page.

### LINKED ISSUE #4289: [UX] Add a button for clearing css/js cache to the appearence page
We had a forum post that highlighted how people new to Backdrop get frustrated when they are not able to see their theme changes right away. Even when someone mentioned **looking for theme update/refresh options** they were unable to find the "Flush CSS/JS cache" link.

css:js

I would like to propose that we add a button to the top of the Appearance page for clearing the CSS/JS cache.

Screen Shot 2020-01-23 at 11 46 14 AM

jenlampton:
Fixes https://github.com/backdrop/backdrop-issues/issues/4289

### DIFF
--- a/core/modules/system/css/system.admin.css
+++ b/core/modules/system/css/system.admin.css
@@ -270,8 +270,14 @@ table.screenshot {
   text-align: right;
   border-bottom: 1px solid #cdcdcd;
 }
-.system-themes-cache-buttons p{
+[dir="rtl"] .system-themes-cache-buttons {
   text-align: left;
+}
+.system-themes-cache-buttons p {
+  text-align: left;
+}
+[dir="rtl"] .system-themes-cache-buttons p {
+  text-align: right;
 }
 .system-themes-list {
   margin-bottom: 20px;
--- a/core/modules/system/css/system.admin.css
+++ b/core/modules/system/css/system.admin.css
@@ -266,6 +266,13 @@ table.screenshot {
 .theme-info p {
   margin-top: 0;
 }
+.system-themes-cache-buttons {
+  text-align: right;
+  border-bottom: 1px solid #cdcdcd;
+}
+.system-themes-cache-buttons p{
+  text-align: left;
+}
 .system-themes-list {
   margin-bottom: 20px;
 }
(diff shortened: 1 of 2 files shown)


core/modules/system/css/system.admin.css
```diff
@@ -270,6 +270,13 @@ table.screenshot {
 .theme-info p {
   margin-top: 0;
 }
+.system-themes-cache-buttons {
+  text-align: right;
+  border-bottom: 1px solid #cdcdcd;
+}
+.system-themes-cache-buttons p{
```
klonos:
```suggestion
.system-themes-cache-buttons p {
```

klonos: We should be adding the respective RTL rules for these.

core/modules/system/system.theme.inc
```diff
@@ -88,11 +88,31 @@ function theme_system_themes_page($variables) {
 
   $output = '<div id="system-themes-page">';
 
+  // @todo make these not dependant on admin_bar?
```
klonos: We should have a follow-up issue for this.

core/modules/system/system.theme.inc
```diff
@@ -88,11 +88,31 @@ function theme_system_themes_page($variables) {
...
+    // If we have a cached menu for the current user, only output the hash
+    // for the client-side HTTP cache callback URL.
+    global $user, $language;
+    $cid = 'admin_bar:' . $user->uid . ':' . session_id() . ':' . $language->langcode;
+    if ($hash = admin_bar_cache_get($cid)) {
+      $options = array(
+        'attributes' => array('class' => array('button')),
+        'query' => array('token' => $hash, backdrop_get_destination()),
+      );
+      $output .= '<div class="system-themes-cache-buttons message-info">';
```
klonos, line 102: Why isn't this output via `backdrop_set_message()`, with `info` as its type instead?

jenlampton:
I've created https://github.com/backdrop/backdrop-issues/issues/5664 based on @klonos 's recommendation.

core/modules/system/system.theme.inc
jenlampton: Because it's not a message. It will always be on the page, regardless of previous actions taken. It's more like a #help, but because this page has special output (not just a rendered array) it was concatenated this way instead.

PR #3057, opened 2020-02-01
Issue #669: Remove RSS settings page.

### LINKED ISSUE #669: Remove the page for RSS settings
Once the front page and the taxonomy/term/% pages have been replaced with views, we'll need to remember to remove the page for RSS settings, since these won't be used anymore.

Also related:
https://github.com/backdrop/backdrop-issues/issues/668

ghost:
Fixes https://github.com/backdrop/backdrop-issues/issues/669


core/modules/system/system.install
```diff
@@ -3101,6 +3101,17 @@ function system_update_1070() {
   }
 }
 
+/**
+ * Remove RSS publishing settings (these are controlled by a view now).
+ */
+function system_update_1071() {
```
klonos: Nice 👍 ...love the cleanup.

PR #3059, opened 2020-02-02
Issue #4298: UX: better admin titles and preview for layout field blocks

klonos:
https://github.com/backdrop/backdrop-issues/issues/4298

### DIFF
--- a/core/modules/field/field.block.inc
+++ b/core/modules/field/field.block.inc
@@ -23,7 +23,7 @@ class FieldBlock extends Block {
    */
   function getAdminTitle() {
     list($entity_type, $field_name) = explode('-', $this->childDelta, 2);
-    return t('Field %field', array('%field' => $field_name));
+    return t('%field field', array('%field' => $field_name));
   }
 
   /**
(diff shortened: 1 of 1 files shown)


core/modules/field/field.block.inc
```diff
@@ -32,11 +32,13 @@ class FieldBlock extends Block {
   function getAdminPreview() {
     list($entity_type, $field_name) = explode('-', $this->childDelta, 2);
     $field = field_info_field($field_name);
+    $field_cardinality = $field['cardinality'];
     $formatter = field_info_formatter_types($this->settings['formatter']);
     $pseudo_instance = $this->createPseudoInstance($this->settings['formatter'], $this->settings['formatter_settings']);
     $summary = module_invoke($formatter['module'], 'field_formatter_settings_summary', $field, $pseudo_instance, '_custom');
 
-    $preview = '<p>' . t('Displays values of "%field" field.', array('%field' => $field_name)) . '</p>';
+    $preview = format_plural($field_cardinality, 'Displays the value of the %field field.', 'Displays values of the %field field.', array('%field' => $field_name));
```
jenlampton: This is a great improvement, but while we are changing translated strings I'd also like to make this more accurate :) This block does not only display the value of the field (example `foo`), it displays the complete field, both formatted and rendered (example: `<div class="field field-body><p>foo</p></div>`).

Let's remove "Values of the" and make the description more truthful :)

```
$preview = t('Displays the %field field.', array('%field' => $field_name));
```

jenlampton:
@klonos I was unable to make a PR against your PR for some reason, so I committed the change to your branch. I think this may need a rebase and will see if I can work on that later. (sorry!)

PR #3066, opened 2020-02-05
Issue #3764: Autocomplete all link/path fields.

### LINKED ISSUE #3764: [UX] Enable autocomplete on all link/path fields
**Description of the need**
One of the things I hate when teaching a client how to use their new Backdrop site is linking to internal pages. The conversation goes something like this:

"And here's a link field where you can link to any page on your site, or to any other website. For linking to pages on your site, you need to go and open that page separately, then copy the URL and then paste it here. Now when you paste the URL here, the best way to do it is to paste everything after your domain name. In other words, if you want to link to https://example.com/projects/my-awesome-project, paste in just the '/projects/my-awesome-project' part (and don't forget the beginning '/'). You can paste in the full URL, but if your domain name ever changes then your links might not work anymore."

This is convoluted and unnecessary. Firstly, having to open a new tab to find the content you want to link to so you can copy the URL, Secondly, having to only paste in part of the URL (and getting frustrated if you do it wrong).

**Proposed solution**
A better solution would be an autocomplete field that shows pages from your site when you start entering keywords. This solves both of the above problems: no need to open a separate tab to find the page yourself, and when you click on a matching result, it'll enter the proper URL for you.

This is actually already in core: https://github.com/backdrop/backdrop-issues/issues/1206 But apparently only in CKEditor. I'd like to see this feature expanded as follows:

- Enable this on any and all link/URL/path fields where internal links are supported (e.g. Link fields, menu item Path fields, Default Front Page (in Site Information), Existing System Path (when adding a URL alias), redirect From & To fields, etc.).
- Support matching all paths/URLs on the site. This includes menu items, views pages, layout pages, the contact page, etc.

Drupal's [Linkit](https://www.drupal.org/project/linkit) module provides a lot of this functionality, but I think most of what it provides is already in Backdrop as part of the CKEditor link feature. I believe there are open issues for supporting menu items in general though (rather than just nodes, etc.). Also, [Linkit Views](https://www.drupal.org/project/linkit_views) seems to allow linking to views pages, and [Linkit Panel Pages](https://www.drupal.org/project/linkit_panel_pages) seems to do the same for Page Manager Pages.

---

PR by @BWPanda: https://github.com/backdrop/backdrop/pull/3066

ghost:
Fixes https://github.com/backdrop/backdrop-issues/issues/3764


core/modules/filter/filter.pages.inc
```diff
@@ -383,7 +262,7 @@ function filter_format_editor_link_form($form, &$form_state, $format) {
     ),
     '#weight' => -10,
     '#maxlength' => 256,
-    '#autocomplete_path' => 'filter-link/autocomplete',
+    '#autocomplete_path' => 'path/autocomplete',
```
docwilmot: could this be just `path-autocomplete`? It isn't part of a `path` module and is in generic use, so no real need for fragments.

core/includes/path.inc
```diff
@@ -581,3 +581,145 @@ function backdrop_clear_path_cache($source = NULL) {
...
   backdrop_static_reset('backdrop_lookup_path');
 }
+
+/**
+ * Autocomplete callback for paths.
+ *
+ * @param string $string
+ *   The string to find matches for.
+ */
+function path_autocomplete($string = '') {
```
docwilmot: Should this be in System Module? It seems curious that the menu link is there but this is in an inc file. And although its a function to find internal paths, it doesnt really have anything to do with core path handling and so on.

docwilmot (commented):
Very minor comments, mostly personal preference. Otherwise RTBC for me.

core/modules/filter/filter.pages.inc
BWPanda: If I do a search of core for `autocomplete_path` I see other ones such as `taxonomy/autocomplete`, `user/autocomplete`, `node/autocomplete`, etc. So that's why I did it this way. There's no other uses of `*-autocomplete` in core, so this'd be the first if we did it that way... I'm not opposed to it necessarily, but just explaining my reasons. Also, I consider 'path' to be generic, not specific to a module/piece-of-functionality.

ghost: Sure, will do.

core/includes/path.inc
ghost: Yeah, I think I was thinking about this backwards originally, but your suggestion makes sense. Will update.

ghost:
Merged into 1.x, thanks!

PR #3070, opened 2020-02-11
Issue #2339: Make Views wizard display names unique.

### LINKED ISSUE #2339: [DX] Views display machine names and template file overrides.
When you create a block or page from the simplified Views UI screen, the machine name is set to `block` or `page` respectively. However when you create your first block, page, feed, etc. from the main (advanced) Views UI screen, the machine names are appended with a number (e.g. `page_1`).

The basic machine names are problematic when trying to override that display's template file if another display of the same type exists. For example, if I have a block created from the simplified UI screen, it's machine name is `block` and so the most specific template file I can use is `views-view-unformatted--test--block.tpl.php`. However, if I have another block display, it's machine name is `block_1` and I can override it using `views-view-unformatted--test--block-1.tpl.php`. The problem is that the template file override for`block` (`views-view-unformatted--test--block.tpl.php`) applies to `block_1` as well, since it's a block-type too. There's therefore no way to override the template file for `block` without it affecting `block_1` as well.

I therefore propose that the machine names get changed to `[display-type]_1` by default (from the simplified UI screen) to avoid this problem, though I'm not sure what ongoing affects this change might have...

ghost:
Fixes https://github.com/backdrop/backdrop-issues/issues/2339


herbdool:
Restarting

PR #3076, opened 2020-02-14
Issue #4306: Adding code to time the test runs and display them in the summary

### LINKED ISSUE #4306: Enhancing testing performance. Also, PHPUnit?
I haven't timed it, but running tests (either via the Testing module in Backdrop or on GitHub when submitting a PR) seems to take a long time. I'm wondering if there's a way of enhancing the performance of tests to make them run better/faster...?

I'm no expert in this area (hence the 'Question' label for this issue), but is there any room for re-writing tests to be more performant?

Also, we use SimpleTest which, if I understand correctly, was custom written for Drupal. What about moving to PHPUnit instead? I know we don't want to use existing technologies just because they exist (e.g. Symphony, Composer, etc.) like Drupal 8 does, but would there be any performance improvement there?

---

PR by @hosef: https://github.com/backdrop/backdrop/pull/3076

hosef:
Fixes backdrop/backdrop-issues#4306


core/modules/simpletest/simpletest.module
```diff
@@ -539,6 +576,45 @@ function simpletest_clean_results_table($test_id = NULL) {
   return 0;
 }
 
+
+/**
+ * Removes cached profile tables from the database.
+ */
+function simpletest_clean_profile_cache_tables(){
+  $tables = db_find_tables(Database::getConnection()->prefixTables('{simpletest_cache_}') . '%');
+  $count = 0;
```
klonos: Instead of using `$count = 0;` and `$count++` here, you could do something like `$count = count($tables);`.

core/modules/simpletest/simpletest.module
```diff
@@ -539,6 +576,45 @@ function simpletest_clean_results_table($test_id = NULL) {
...
+  foreach ($tables as $table) {
+    db_drop_table($table);
+    $count++;
+  }
+
+  if ($count > 0) {
+    backdrop_set_message(format_plural($count, 'Removed 1 profile cache table.', 'Removed @count profile cache tables.'));
```
klonos, line 598: Should we be showing which tables were dropped here? ...like an `implode()` on `$tables` or something?

```suggestion
    backdrop_set_message(t('Removed the following profile cache tables: @tables', array('@tables' => implode(",", $tables))));
```

core/modules/simpletest/simpletest.module
```diff
@@ -539,6 +576,45 @@ function simpletest_clean_results_table($test_id = NULL) {
...
+  else {
+    backdrop_set_message(t('No profile cache tables to remove.'));
+  }
+}
+
+/**
+ * Removes cached profile folders from the files directory.
+ */
+function simpletest_clean_profile_cache_folders(){
+  $profiles = array(
```
klonos: We are using the same list of profiles earlier, in `simpletest_run_tests()`. Should we define those in a `simpletest_get_profiles()` function, or a constant?

core/modules/simpletest/simpletest.pages.inc
```diff
@@ -308,6 +308,12 @@ function simpletest_settings_form($form, &$form_state) {
     '#description' => t('The verbose data will be printed along with the standard assertions and is useful for debugging. The verbose data will be erased between each test suite run. The verbose data output is very detailed and should only be used when debugging.'),
     '#default_value' => $config->get('simpletest_verbose'),
   );
+  $form['general']['simpletest_cache'] = array(
+    '#type' => 'checkbox',
+    '#title' => t('Cache instal profiles'),
```
klonos: Typo: "instal" -> "install" ...or better use the full word "installation" instead:

```suggestion
    '#title' => t('Cache installation profiles'),
```

core/modules/simpletest/simpletest.pages.inc
```diff
@@ -308,6 +308,12 @@ function simpletest_settings_form($form, &$form_state) {
...
+    '#description' => t('Saves the database tables and config files from each instal profile so that each one only has to be run once. If you are making database changes, you will need to clean the test environment for the changes to show up in the tests.'),
```
klonos: A bit more wordy, but better describes what is happening I believe:

```suggestion
    '#description' => t('Instead of multiple installations of Backdrop once for each test, each installation profile is installed only once, and its initial state (database tables and config files) is cached and reused. If you are making database changes, you will need to clean the test environment for the changes to show up in the tests.'),
```

core/modules/simpletest/tests/batch_test.module
```diff
@@ -357,7 +357,7 @@ function _batch_test_batch_0() {
 function _batch_test_batch_1() {
   // Ensure the batch takes at least two iterations.
   $total = 10;
-  $sleep = (1000000 / $total) * 2;
+  $sleep = 100000 / $total;
```
klonos: I believe that these were set to 2sec for good reasons. We may have failures in slower systems (not 100% though - someone else please verify?)

core/modules/simpletest/simpletest.module
hosef: I think we should probably not do that, because there are over a hundred tables. It is also just a copy of the tables that are available with a default install.

klonos: Fair enough 👍

core/modules/simpletest/backdrop_web_test_case.php
```diff
@@ -1217,34 +1225,44 @@ protected function backdropCreateContentType($settings = array()) {
...
+    // Make sure type is valid.
+    if (in_array($type, array('binary', 'html', 'image', 'javascript', 'php', 'sql', 'text'))) {
 
-      // Generate text test files.
-      $lines = array(16, 256, 1024, 2048, 20480);
-      $count = 0;
-      foreach ($lines as $line) {
-        simpletest_generate_file('text-' . $count++, 64, $line, 'text');
-      }
+      if (!in_array($type, $this->generatedTestFiles)) {
```
quicksketch, line 1232: Smart. I like how this only generates the individual test file requested instead of generating all of them.

core/modules/simpletest/simpletest.module
```diff
@@ -539,6 +576,44 @@ function simpletest_clean_results_table($test_id = NULL) {
...
+    'minimal',
+    'standard',
+    'testing',
+  );
+
+  $file_public_path = config_get('system.core', 'file_public_path', 'files');
+
+  foreach($profiles as $profile) {
```
quicksketch:
```suggestion
  foreach ($profiles as $profile) {
```

core/modules/simpletest/simpletest.module
```diff
@@ -539,6 +576,44 @@ function simpletest_clean_results_table($test_id = NULL) {
   return 0;
 }
 
+
+/**
+ * Removes cached profile tables from the database.
+ */
+function simpletest_clean_profile_cache_tables(){
```
quicksketch:
```suggestion
function simpletest_clean_profile_cache_tables() {
```

core/modules/simpletest/simpletest.module
```diff
@@ -539,6 +576,44 @@ function simpletest_clean_results_table($test_id = NULL) {
...
+  }
+  else {
+    backdrop_set_message(t('No profile cache tables to remove.'));
+  }
+}
+
+/**
+ * Removes cached profile folders from the files directory.
+ */
+function simpletest_clean_profile_cache_folders(){
```
quicksketch:
```suggestion
function simpletest_clean_profile_cache_folders() {
```

core/modules/simpletest/simpletest.module
quicksketch: I think @klonos's suggestion is good here. It would also be a good idea to make it so that new profiles or custom/contrib profiles worked with profile caching.

We already have a function in install.core.inc that finds all profiles, `install_find_profiles()`. Unfortunately this isn't a commonly included file. I think making a wrapper function like `simpletest_get_profiles()` that wraps around that other function and include would be appropriate to make it dynamic and avoid a hard-coded list.

hosef:
@quicksketch Thanks for the review. I made the changes that you recommended, and added the function that Klonos suggested.

Unfortunately, I also had to add some of the `sleep()` calls back in, but I made them all 1 second instead of the 3-6 seconds they were before, since some of the tests were failing because they were too fast.

quicksketch:
Thanks @hosef! Yeah I reran the tests a few times and was getting failures depending on how fast they executed. I'll run a few more times to check the latest changes.

quicksketch:
I got a random failure on one of the modified tests, a failure I haven't seen any time recently:

image

`Connection header not set keep-alive when serving cached page. bootstrap.test:270`

Maybe it would be best to leave the 5 second sleep time in there instead of reducing it to 1 second. Sometimes if the page takes too long to load, 1 second cuts it too close.

hosef:
@quicksketch I reverted a bunch of the sleep changes and they all passed now.

PR #3078, opened 2020-02-18
Issue #4218: Replaced double hyphens with middots in menu parent select list.

verygladden:
https://github.com/backdrop/backdrop-issues/issues/4218


ghost:
Thanks! Merged into 1.x.

PR #3079, opened 2020-02-18
Issue #4308: Fixing PHP 7.4 compatibility problems.

hosef:


hosef:
backdrop/backdrop-issues#4308

core/includes/batch.inc
```diff
@@ -453,7 +453,7 @@ function _batch_finished() {
 
   // Clean-up the session. Not needed for CLI updates.
   if (isset($_SESSION)) {
-    unset($_SESSION['batches'][$batch['id']]);
+    unset($_SESSION['batches'][$_batch['id']]);
```
quicksketch, line 456: Nice bug fix here. If `$batch` is set to NULL just three lines above, there's no way we can check `$batch['id']`.

core/includes/bootstrap.inc
```diff
@@ -3258,7 +3258,13 @@ function _backdrop_bootstrap_sanitize_request() {
     }
 
     // If there is a query string, check its query parameters.
-    $destination_parts = backdrop_parse_url($_GET['destination']);
+    if (isset($_GET['destination'])) {
+      $destination_parts = backdrop_parse_url($_GET['destination']);
+    }
+    else if (isset($_REQUEST['destination'])) {
+      $destination_parts = backdrop_parse_url($_REQUEST['destination']);
+    }
```
quicksketch: The addition of `$_REQUEST['destination']` seems like a change in functionality? Can we remove this `else if` statement?

quicksketch: Also a few lines below, looks like we need to check if each exists when doing:
```
      unset($_GET['destination']);
      unset($_REQUEST['destination']);
```

Even though that triggers a fatal error, it shouldn't also cause a PHP warning.

core/includes/common.inc
```diff
@@ -471,7 +471,7 @@ function backdrop_add_feed($url = NULL, $title = '') {
  */
 function backdrop_get_feeds($delimiter = "\n") {
   $feeds = backdrop_add_feed();
-  return implode($feeds, $delimiter);
+  return implode($delimiter, $feeds);
```
quicksketch, line 474: Wow. Huh. Backwards parameters since 2006. How did you run across this?

core/includes/common.inc
```diff
@@ -3993,7 +3995,12 @@ function _backdrop_build_css_path($matches, $base = NULL) {
   }
 
   // Prefix with base and remove '../' segments where possible.
-  $path = $_base . $matches[1];
+  if (is_array($matches)) {
+    $path = $_base . $matches[1];
+  }
+  else {
+    $path = $_base . $matches;
+  }
```
quicksketch, line 4003: Looks like `$matches` is either an array or `NULL`. I think this would be better to be more specific and be written as:

```
  if (isset($matches)) {
    $path = $_base . $matches[1];
  }
  else {
    $path = $_base;
  }
```

core/modules/image/image.module
```diff
@@ -881,16 +881,18 @@ function image_style_transform_dimensions($style_name, array &$dimensions) {
  *   An image style array.
  */
 function image_style_flush($style) {
-  // Delete the style directory in each registered wrapper.
-  $wrappers = file_get_stream_wrappers(STREAM_WRAPPERS_WRITE_VISIBLE);
-  foreach ($wrappers as $wrapper => $wrapper_data) {
-    if (file_exists($directory = $wrapper . '://styles/' . $style['name'])) {
-      file_unmanaged_delete_recursive($directory);
+  if (!empty($style)) {
```
quicksketch: Hm, this seems like it _should_ throw a PHP warning. `$style` should always be something. If it's being called with `NULL` somewhere, we should fix that call, not suppress the warning.

core/modules/image/image.module
```diff
@@ -1245,7 +1247,7 @@ function image_effect_save($style_name, $effect) {
   }
 
   $config->save();
-  $style = image_style_load(NULL, $style_name);
+  $style = image_style_load($style_name);
```
quicksketch, line 1248: Oh, maybe here's where you fixed it? If so, perhaps just revert the above change in `image_style_flush()`

core/modules/path/path.module
```diff
@@ -687,19 +687,21 @@ function path_load_multiple_by_entities(array $entities) {
...
-    // Provide a default path in the event no path is found.
-    $entity->path = array(
-      'pid' => NULL,
-      'source' => $uri['path'],
-      'alias' => NULL,
-      'langcode' => $langcode,
-      'auto' => FALSE,
-    );
+    if (!empty($entity->uri)) {
+      $uri = $entity->uri();
```
quicksketch: I'm not sure this is correct. Entities like `Node` do not have a `$node->uri` property, but they still return something when calling `$node->uri()`. I think this should be:
```
    $uri = $entity->uri()
    if (!empty($uri)) {
```

core/modules/system/system.admin.inc
```diff
@@ -1499,11 +1499,12 @@ function system_site_information_settings_submit($form, &$form_state) {
...
   if ($form_state['values']['site_logo_path']) {
+    $form_state['values']['site_logo_attributes'] = array();
     $dimensions = @getimagesize($form_state['values']['site_logo_path']);
-    $form_state['values']['site_logo_attributes'] = array(
-      'width' => $dimensions[0],
-      'height' => $dimensions[1],
-    );
+    if (!empty($dimensions)) {
+      $form_state['values']['site_logo_attributes']['width'] = $dimensions[0];
+      $form_state['values']['site_logo_attributes']['height'] = $dimensions[0];
```
quicksketch: Looks like height should be coming from `$dimensions[1]`
```suggestion
      $form_state['values']['site_logo_attributes']['height'] = $dimensions[1];
```

core/modules/taxonomy/taxonomy.admin.inc
```diff
@@ -310,7 +310,9 @@ function taxonomy_form_vocabulary_submit($form, &$form_state) {
   }
 
   $vocabulary = $form_state['vocabulary'];
-  entity_form_submit_build_entity('taxonomy_vocabulary', $vocabulary, $form, $form_state);
+  foreach ($form_state['values'] as $key => $value) {
+    $vocabulary->$key = $value;
+  }
```
quicksketch, line 316: Nice thanks. As vocabularies are no longer entities themselves, this seems correct. This might also end up saving things like the form_token though. Could you call `form_state_values_clean($form_state)` first?

core/modules/views_ui/views_ui.admin.inc
```diff
@@ -2978,7 +2978,7 @@ function views_ui_reorder_displays_form($form, &$form_state) {
       'weight' => array(
         '#type' => 'weight',
         '#value' => $display->position,
-        '#delta' => count($last_display->position),
+        '#delta' => (empty($last_display->position)) ? 1 : 0,
```
quicksketch: This would result in a select list that only had either 0 or 3 options (-1, 0, and 1). I think the right thing here is probably:
```suggestion
        '#delta' => count($view->display),
```

quicksketch:
Merged into 1.x and 1.15.x. Thanks @hosef!

PR #3082, opened 2020-02-22
Issue #4295: Color Module: Rebuild CSS

### LINKED ISSUE #4295: [DX] Themes with color support: CSS changes not taking effect unless theme settings are re-saved.
**Description of the need**
I just ran into a situation in which I made a css change and could not get it to take effect, regardless of how many times I cleared the browser cache or the Backdrop cache. 

I examined the css and the new rule was there in the css, the theme was loading the css sheet, but I could not find the rule in the inspector. 

I finally realized that I am using the color module for this site and that the color module is rewriting that particular css sheet and storing it in files/color

The ONLY way I could get my new css change to take effect, was to resave the color settings for the theme. It's always weird when you are forced to resave something that you haven't changed (or are not aware that you have changed). 

**Proposed solution**
Add a "rebuild color scheme" link in the admin menu that basically just resaves (and rebuilds) the color files. Put this near the "clear cache" link/button. 

Then if you are trouble shooting a theme issue and doing the obvious things, like clearing the cache, you'll also be reminded to rebuild your color settings - if it makes sense. 

???

I would love to hear alternative ideas to help new users in this kind of situation. 

**Additional information**
Most experienced developers don't use the color module, they hard code their color changes into their theme. I ran into this while working on Tatsu, which I'm specifically building for users that are likely to use the color module. 

This type of developer (or site architect) is not likely to think about how the [color module works](https://www.triplo.co/blog/color-module-backdrop-cms) and recognize the need to resave their settings. 

___

PR by docwilmot: https://github.com/backdrop/backdrop/pull/3110

docwilmot:
Fixes https://github.com/backdrop/backdrop-issues/issues/4295

### DIFF
--- a/core/modules/color/color.module
+++ b/core/modules/color/color.module
@@ -582,6 +582,42 @@ function color_save_configuration($theme_name, $theme_info, $palette) {
 }
 
 /**
+ * Implements hook_admin_bar_cache_info().
+ */
+function color_admin_bar_cache_info() {
+  global $theme;
+  $cache = array();
+  $theme_config_name = $theme . '.settings';
+  $theme_config = config($theme_config_name);
+  if ($theme_config && $color = $theme_config->get('color')) {
+    $cache['color'] = array(
+      'title' => 'Color scheme',
+      'callback' => '_color_rebuild_settings',
+    );
+  }
+  return $cache;
+}
+
+/**
+ * Rebuild color scheme info.
+ */
+function _color_rebuild_settings() {
+  global $theme;
+  $theme_config_name = $theme . '.settings';
+  $theme_config = config($theme_config_name);
+  if ($theme_config && $new_color_config = $theme_config->get('color')) {
+    $color_info = color_get_info($theme);
+    if ($color_info) {
+      $palette = isset($new_color_config['palette']) ? $new_color_config['palette'] : NULL;
+      $new_color_info = color_save_configuration($theme, $color_info, $palette);
+      $theme_config->set('color', $new_color_info);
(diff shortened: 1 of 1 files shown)


core/modules/color/color.module
```diff
@@ -581,6 +581,42 @@ function color_save_configuration($theme_name, $theme_info, $palette) {
...
+/**
+ * Implements hook_admin_bar_cache_info().
+ */
+function color_admin_bar_cache_info() {
+  global $theme;
+  $cache = array();
+  $theme_config_name = $theme . '.settings';
+  $theme_config = config($theme_config_name);
+  if ($theme_config && $color = $theme_config->get('color')) {
+    $cache['color'] = array(
```
klonos, line 593: The other implementations of this hook return a `$caches` array (plural). Perhaps that's why this one is not shown under the "Clear caches" menu in the admin bar(?).

Edit: that's not it, but I think that we should still be using plural, for consistency.

core/modules/color/color.module
```diff
@@ -581,6 +581,42 @@ function color_save_configuration($theme_name, $theme_info, $palette) {
...
+      'title' => 'Color scheme',
```
klonos, line 594:
```suggestion
      'title' => t('Color scheme'),
```

core/modules/color/color.module
docwilmot: The link only appears if there are color schemes saved. I changed the color scheme now, it appears now.

docwilmot:
Testing closing

PR #1692, opened 2016-12-27
Issue #2432: Remove "core": entries from default views config files.

klonos:

PR #3090, opened 2020-03-05
Issue #2432: Remove "core": entries from default views config files.

### LINKED ISSUE #2432: What is the "core": entry used for in views .json config files?
Now that I am creating PRs that involve creating default views configs, I noticed that there are these lines added in the config files. What are they used for? Should we remove them if not used? Should we also remove the function that adds them in the first place?

klonos:
Fixes https://github.com/backdrop/backdrop-issues/issues/2432
Replaces https://github.com/backdrop/backdrop/pull/1692

### DIFF
--- a/core/modules/views/includes/view.inc
+++ b/core/modules/views/includes/view.inc
@@ -5,15 +5,15 @@
  */
 
 /**
- * @defgroup views_objects Objects that represent a View, or part of a view
+ * @defgroup views_objects Objects that represent a View or part of a view
  * @{
- * These objects are the core of Views, do the bulk of the direction and
+ * These objects are the core of Views do the bulk of the direction and
  * storing of data. All database activity is in these objects.
  */
 
 /**
  * An object to contain all of the data to generate a view, plus the member
- * functions to build the view query, execute the query, and render the output.
+ * functions to build the view query, execute the query and render the output.
  */
 class view {
   var $base_table = 'node';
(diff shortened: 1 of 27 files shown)


core/modules/views/includes/view.inc
```diff
@@ -49,12 +50,6 @@ class view {
    */
   var $human_name;
 
-  /**
-   * The core version the view was created for.
-   * @var int
-   */
-  var $core;
```
klonos: This is basically the only substantial change in this file; the rest is just fixing comments.

core/modules/views/views.install
```diff
@@ -186,6 +186,21 @@ function views_update_1006() {
...
 }
 
+/**
+ * Remove the unused "core" key from active config files.
+ */
+function views_update_1007() {
+  $configs = config_get_names_with_prefix('views.view.');
+
+  foreach ($configs as $config_file) {
+    $core_key_in_config = config_get($config_file, 'core');
```
ghost: Would it be better to use:

```php
$config = config($config_file);
$core_key_in_config = $config->get('core');
```

here to avoid reading the same config file twice (once for getting the value and once for clearing it)?

core/modules/views/views.install
```diff
@@ -186,6 +186,21 @@ function views_update_1006() {
...
+    if ($core_key_in_config !== NULL) {
+      config_clear($config_file, 'core');
+    }
+  }
+
```
ghost: Remove blank line here.

ghost (changes requested):

core/modules/views/views.install
klonos: You're right. Done...

core/modules/views/views.install
klonos: ...and done 🙂

PR #3094, opened 2020-03-07
Issue #3880: Update the html5shiv library from 3.7.2 to 3.7.3.

hosef:


laryn:
Thanks @hosef, @klonos, @herbdool and @BWPanda!

PR #3101, opened 2020-03-24
Issue #1425: Token display format is not configurable

### LINKED ISSUE #1425: Token display format is not configurable (and date token creates very long paths)
I've created a date field, and want to include the start date of an event in its url alias pattern:
events/[node:field_event_date]-[node:title]

However, when the path is generated, it uses the complete start and end dates and times, like:
http://test-datera.pantheon.io/events/2015-12-18-220000-2015-12-18-230000-second-website-demo

I don't remember having trouble adding the simple start date when doing this in D7. How are we handling things differently, now that we've moved the token module to Backdrop core?

---

~PR by @docwilmot: https://github.com/backdrop/backdrop/pull/2476~
PR by @docwilmot: https://github.com/backdrop/backdrop/pull/3101

docwilmot:
Fixes https://github.com/backdrop/backdrop-issues/issues/1425

### DIFF
--- a/core/modules/entity/entity.module
+++ b/core/modules/entity/entity.module
@@ -282,6 +282,16 @@ function entity_get_info($entity_type = NULL) {
           }
         }
 
+        // Add a token view mode if it does not already exist. Only work with
+        // fieldable entities.
+        if (!empty($entity_info[$name]['fieldable'])) {
+          if (!isset($entity_info[$name]['view modes']['token'])) {
+            $entity_info[$name]['view modes']['token'] = array(
+              'label' => t('Tokens'),
+            );
+          }
+        }
+
         foreach ($entity_info[$name]['view modes'] as $view_mode => $view_mode_info) {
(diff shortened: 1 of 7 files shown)


core/modules/entity/entity.module
```diff
@@ -282,6 +282,16 @@ function entity_get_info($entity_type = NULL) {
           }
         }
 
+        // Add a token view mode if it does not already exist. Only work with
+        // fieldable entities.
+        if (!empty($entity_info[$name]['fieldable'])) {
+          if (!isset($entity_info[$name]['view modes']['token'])) {
+            $entity_info[$name]['view modes']['token'] = array(
+              'label' => t('Tokens'),
+            );
```
quicksketch, line 292: Other places, `'custom settings' => FALSE,` is also added. Should that be here too?

Additionally, if the `token` view mode is added to all entities here, do they need to be specified separately in node, comment, user, and taxonomy modules?

docwilmot: Done

PR #3102, opened 2020-03-25
Issue #4330: Fixed the permission check in the color module so that u…

hosef:
…sers other than uid 1 can see the color preview.


klonos (approved):
LGTM 👍

quicksketch:

I accidentally cherry-picked into 1.14.x as well. I could revert it in the git history but this shouldn't do any harm. So although 1.14.3 is a security release, it will also fix this bug.

PR #3103, opened 2020-03-25
Update CKEditor to version 4.14.0.

quicksketch:
Fixes https://github.com/backdrop-ops/security/issues/87.

Updates CKEditor, so is related to https://github.com/backdrop/backdrop-issues/issues/3355


jenlampton:
In my testing the image relocation feature (cross-hairs-drag-and-dropper) wont relocate the image in either FireFox or Chrome, on Mac OS. 
Screen Shot 2020-03-25 at 12 59 14 PM

jenlampton:
In my testing, it was only the second embedded image that would not relocate. The first one moved around just fine. This problem was not present on 1.15.0

quicksketch:
New sandbox with center-alignment fixed in the editor when using a caption.

jenlampton:
On the new site, I had fewer problems, with dragging the second image. It mostly worked. When it didn't I could switch to source view, then switch back, and it would work.

quicksketch:
Reopening to re-run tests.

quicksketch:
We're seeing one of the intermittent test failures here: https://github.com/backdrop/backdrop-issues/issues/4248

That test aside, we're passing everything and neither @jenlampton nor I have found regressions specific to the upgrade.

PR #3104, opened 2020-03-25
Update CKEditor to version 4.14.0.

quicksketch:
Fixes backdrop-ops/security#87 for 1.14.x.

The commit from https://github.com/backdrop/backdrop/pull/3103 applies cleanly to both 1.x and 1.14.x. This separate PR is being created so we can get a separate sandbox to test.


quicksketch:
Merged in https://github.com/backdrop/backdrop/pull/3103 into 1.14.x.

PR #3106, opened 2020-03-29
Issue #4353: Changing the test setUp and tearDown functions to only r…

### LINKED ISSUE #4353: Reuse test prefixes instead of rebuilding them for each test
**Description of the need**

Currently, when a test is done it deletes all the tables and files for it's prefix. During most tests, only a few tables are modified, which means that most of the table dropping and creating is unnecessary.
**Proposed solution**

Change the setUp and tearDown functions from the test class to not delete the class prefix, but instead replace the tables that were changed.

hosef:
…eplace changed files

Fixes https://github.com/backdrop/backdrop-issues/issues/4353


hosef:
Closed and reopened to re-run the tests.

PR #3109, opened 2020-04-01
Issue #4227: Image fields: Do not allow min resolution to be larger than max resolution

### LINKED ISSUE #4227: Image fields: Do not allow min resolution to be larger than max resolution
Filing this as a follow-up to #4221.

When creating image fields, there is no validation for the resolution when min > max. So you end up with this:

69861930-80763100-12ed-11ea-813c-f39916cb9f6b

---
~~PR by @klonos https://github.com/backdrop/backdrop/pull/3009~~
PR by docwilmot https://github.com/backdrop/backdrop/pull/3109

docwilmot:
Fixes https://github.com/backdrop/backdrop-issues/issues/4227
Replaces https://github.com/backdrop/backdrop/pull/3009

### DIFF
--- a/core/modules/image/image.field.inc
+++ b/core/modules/image/image.field.inc
@@ -204,7 +204,7 @@ function image_field_instance_settings_form($field, $instance) {
 }
 
 /**
- * #element_validate handler for image_field_instance_settings_form().
+ * Element validate handler for image_field_instance_settings_form().
  */
 function image_field_instance_settings_form_validate($element, &$form_state, $complete_form) {
   $min_dimensions = explode('x', $form_state['values']['instance']['settings']['min_dimensions']) + array('', '');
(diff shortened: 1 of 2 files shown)


core/modules/image/tests/image.test
```diff
@@ -872,6 +872,32 @@ class ImageFieldDisplayTestCase extends ImageFieldTestCase {
       '%max' => $schema['fields'][$field_name .'_title']['length'],
       '%length' => $test_size,
     )));
+
+    // Check that an image minimum size cant be larger than its maximum size.
```
jenlampton: can we change `Check that an image minimum size cant be larger than its maximum size.` into `Check that image minimum size is not larger than maximum size.`?

docwilmot: Resolved

core/modules/image/image.field.inc
```diff
@@ -200,6 +203,26 @@ function image_field_instance_settings_form($field, $instance) {
   return $form;
 }
 
+/**
+ * #element_validate handler for image_field_instance_settings_form().
```
jenlampton: If you're still nearby, can we also change this to `Element validate` instead of `#element_validate`?

docwilmot: Done

quicksketch:
Merged into 1.x and 1.15.x. Thanks!

PR #3110, opened 2020-04-02
Issue #4295: Color Module: Rebuild CSS

### LINKED ISSUE #4295: [DX] Themes with color support: CSS changes not taking effect unless theme settings are re-saved.

docwilmot:
Fixes https://github.com/backdrop/backdrop-issues/issues/4295
Replaces https://github.com/backdrop/backdrop/pull/3082

### DIFF
--- a/core/includes/common.inc
+++ b/core/includes/common.inc
@@ -8407,6 +8407,10 @@ function backdrop_implode_tags($tags) {
  * invokes a hook so that other modules' cache data can be cleared as well.
  */
 function backdrop_flush_all_caches() {
+  // Rebuild all color palettes.
+  if (module_exists('color')) {
+    color_rebuild_settings();
+  }
   // Change query-strings on css/js files to enforce reload for all users.
   _backdrop_flush_css_js();
 
--- a/core/includes/common.inc
+++ b/core/includes/common.inc
@@ -8407,10 +8407,6 @@ function backdrop_implode_tags($tags) {
  * invokes a hook so that other modules' cache data can be cleared as well.
  */
 function backdrop_flush_all_caches() {
-  // Rebuild all color palettes.
-  if (module_exists('color')) {
-    color_rebuild_settings();
-  }
   // Change query-strings on css/js files to enforce reload for all users.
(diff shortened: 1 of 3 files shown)


core/includes/common.inc
```diff
@@ -8407,6 +8407,10 @@ function backdrop_implode_tags($tags) {
  * invokes a hook so that other modules' cache data can be cleared as well.
  */
 function backdrop_flush_all_caches() {
+  // Rebuild all color palettes.
+  if (module_exists('color')) {
+    color_rebuild_settings();
+  }
```
quicksketch: Instead of hard-coding color module into `backdrop_flush_all_caches()` could we instead use `hook_flush_caches()` and put it in `color.module`?

quicksketch:

PR #3113, opened 2020-04-07
Issue #484: Added comment blocks to be more complete with arg specs et-al fo…

### LINKED ISSUE #484: Update all preprocess docblocks
Initial PR for this issue: https://github.com/backdrop/backdrop/pull/3113

The documentation we use for preprocess functions is inconsistent all over core.  [Drupal 8 adopted a new standard](https://www.drupal.org/node/1913208).  Below are two examples of the documentation I'm proposing for Backdrop.

1) When there are only a few variables (5 or fewer) provided to the template we should list them all in the docs for the preprocess function.  For example:

```
/**
 * Prepares variables for block templates.
 *
 * @param $variables
 *   An array containing (but not limited to) the following:
 *   - content: The actual content of the block.
 *   - block: The block object being displayed.
 *
 * @see block.tpl.php
 */
function template_preprocess_block(&$variables) {
```

2) When there are a lot of variables (6 or more) provided to the template (node, for example) we can point people to the template file where all available default variables should be listed.  For example:

```
/**
 * Prepares variables for node templates.
 *
 * Most themes use their own copy of node.tpl.php. The default is located at
 * "core/modules/node/node.tpl.php". The full list of variables is documented
 * in that file.
 *
 * @param $variables
 *   An array containing, but not imited to, the following:
 *   - node: Full node entity. Contains data that may not be safe.
 *   - view_mode: View mode; for example, "full", or "teaser".
 *
 * @see node.tpl.php
 */
function template_preprocess_node(&$variables) {
```

ps23Rick:
Fixes backdrop/backdrop-issues#484 (partial)
Initial rundown with alternate comment blocks.  Seeking feedback on whether to continue or if further alterations are necessary/wanted.

### DIFF
--- a/core/themes/basis/template.php
+++ b/core/themes/basis/template.php
@@ -7,6 +7,18 @@
 /**
  * Prepares variables for page templates.
  *
+ * Many themes provide their own copy of page.tpl.php. The default is located at
+ * "core/modules/system/templates/page.tpl.php". The full list of variables is 
+ * documented in that file.
+ *
+ * @param $variables
+ *   An array containing (but not limited to) the following:
+ *   - css: The array of CSS files to be used for this page.
+ *   - page: The rendered page content, as output from Layout module.
+ *   - page_bottom: Final closing markup from any modules that have altered the
+ *   page. This variable should always be output last, after all other dynamic
+ *   content.
+ * 
  * @see page.tpl.php
  */
 function basis_preprocess_page(&$variables) {
@@ -30,6 +42,18 @@ function basis_preprocess_page(&$variables) {
 /**
  * Prepares variables for maintenance page templates.
  *
+ * Many themes provide their own copy of maintenance-page.tpl.php. The default is located at
+ * "core/modules/system/templates/maintenance-page.tpl.php". The full list of variables is 
+ * documented in that file or files that it references.
+ *
+ * @param $variables
+ *   An array containing (but not limited to) the following:
+ *   - css: The array of CSS files to be used for this page.
+ *   - page: The rendered page content, as output from Layout module.
+ *   - page_bottom: Final closing markup from any modules that have altered the
+ *   page. This variable should always be output last, after all other dynamic
+ *   content.
+ * 
  * @see maintenance-page.tpl.php
  */
 function basis_preprocess_maintenance_page(&$variables) {
@@ -40,6 +64,16 @@ function basis_preprocess_maintenance_page(&$variables) {
 /**
  * Prepares variables for layout templates.
  *
+ * This is the theme specific layout method for single column layouts. 
+ * The default is located at "core/modules/layout/templates/layout.tpl.php". 
+ * The full list of variables is documented in that file.
+ *
+ * @param $variables
+ *   An array containing (but not limited to) the following:
+ *   - title: The page title, for use in the actual HTML content.
+ *   - classes: Array of classes to be added to the layout wrapper.
+ *   - content: An array of content, each item in the array is keyed to one
+ * 
  * @see layout.tpl.php
  */
 function basis_preprocess_layout(&$variables) {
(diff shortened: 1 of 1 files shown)


core/themes/basis/template.php
```diff
@@ -7,6 +7,18 @@
 /**
  * Prepares variables for page templates.
  *
+ * Many themes provide their own copy of page.tpl.php. The default is located at
+ * "core/modules/system/templates/page.tpl.php". The full list of variables is 
+ * documented in that file.
+ *
+ * @param $variables
+ *   An array containing (but not limited to) the following:
+ *   - css: The array of CSS files to be used for this page.
```
jenlampton, line 16: I don't think there is a $css variable in Backdrop. We may need to fix the docs, these look like they are from Drupal 7...  Can you check `template_preprocess_page()` in `system.module`?

core/themes/basis/template.php
```diff
@@ -30,6 +42,18 @@ function basis_preprocess_page(&$variables) {
 /**
  * Prepares variables for maintenance page templates.
  *
+ * Many themes provide their own copy of maintenance-page.tpl.php. The default is located at
+ * "core/modules/system/templates/maintenance-page.tpl.php". The full list of variables is 
+ * documented in that file or files that it references.
+ *
+ * @param $variables
+ *   An array containing (but not limited to) the following:
+ *   - css: The array of CSS files to be used for this page.
```
jenlampton, line 51: same here... Backdrop calls `backdrop_get_css()` in the template file. I don't think this variable exists before then since we have no process layer to alter it, so if used, this variable might print an incomplete set of CSS files.

core/themes/basis/template.php
```diff
@@ -40,6 +64,16 @@ function basis_preprocess_maintenance_page(&$variables) {
 /**
  * Prepares variables for layout templates.
  *
+ * This is the theme specific layout method for single column layouts.
```
jenlampton, line 67: I don't think the term `layout method` is defined anywhere? How about...  `This template file will be used by this theme instead of the default template file for a single column layout.`

jenlampton: Or we could stick with the standard...  `Some themes provide their own copy of layout.tpl.php.` ... without elaborating. The layout template is special though, so maybe more here is useful?

jenlampton:
@ps23Rick these changes look great! I've added some comments about docs that need to be updated (I suspect they were incorrect before this PR) and provided some options for alternate wording in one case...  What do you think of the options?

ps23Rick:
Hey @jenlampton.. I'm assuming you are referring to updates to the original issue#484.. It looks like the original post was updated since I looked at it yesterday/last night.  It looks good. 

If you'd like me to tackle more, I'm assuming you'll want something along the lines of a separate PR for maybe each directory or some other grouping that isn't obvious in my head at the moment.  There are 181 instances of `_preprocess_` in the entire tree -- I believe its very much a subset of that in which changes are needed.  If you're OK with this thought, I'm thinking that I'll tackle a specific area of the tree, target 3-5 files (likely with each file having 3-5 changes in it) that need changes and create a new PR for that.  Does that seem reasonable?  Thx!

jenlampton:
> I'm assuming you are referring to updates to the original issue#484

I left them as comments here, on the PR.  See above?  Hmm.. why do they say pending...  I'll work on that...

> I'm thinking that I'll tackle a specific area of the tree, target 3-5 files (likely with each file having 3-5 changes in it) that need changes and create a new PR for that. Does that seem reasonable? Thx!

Yep!  The separate-PRs idea was primarily to keep you sane while working on them, so do whatever division feels reasonable to you :)

jenlampton (changes requested):
how do I submit this thing?

core/themes/basis/template.php
ps23Rick: Ok.. I'm not sure how to proceed on this.. I assumed the comments in page.tpl.php were accurate but as I well know in coding that is no guarantee. I found the offending `template_preprocess_page()` method located in `./core/includes/theme.inc`.  I didn't find any reference in system.module.  Unfortunately without me having more knowledge of this body of code I'm not sure how to adjust this.

core/themes/basis/template.php
ps23Rick: I'll take door #2 and just be a bit more consistent with the other mods.  I've adjusted my local codebase with this fix.

PR #3122, opened 2020-04-22
Issue #4373: Fix incorrect cache bin.

### LINKED ISSUE #4373: Incorrect cache bin used in 'views_handler_relationship_groupwise_max::query'
in `views_handler_relationship_groupwise_max::query()` is has this code:

```php
// Get the stored subquery SQL string.
$cid = 'views_relationship_groupwise_max:' . $this->view->name . ':' . $this->view->current_display . ':' . $this->options['id'];
$cache = cache('views_data')->get($cid);
if (isset($cache->data)) {
  $def['left_query'] = $cache->data;
}
else {
  $def['left_query'] = $this->left_query($this->options);
  cache('cache_views_data')->set($cid, $def['left_query']);
}
```

The last line sets the value in cache bin `cache_views_data`, but the bin `views_data` is used above to get the value.

`cache_views_data` needs to be changed to `views_data`.

I noticed this when setting up a 'Taxonomy term: Representative node' relationship in Views. The sorting didn't work unless I ticked 'Generate subquery each time view is run' (which is for testing only).

ghost:
Fixes https://github.com/backdrop/backdrop-issues/issues/4373

### DIFF
--- a/core/modules/views/handlers/views_handler_relationship_groupwise_max.inc
+++ b/core/modules/views/handlers/views_handler_relationship_groupwise_max.inc
@@ -358,7 +358,7 @@ class views_handler_relationship_groupwise_max extends views_handler_relationshi
       }
       else {
         $def['left_query'] = $this->left_query($this->options);
-        cache('cache_views_data')->set($cid, $def['left_query']);
+        cache('views_data')->set($cid, $def['left_query']);
       }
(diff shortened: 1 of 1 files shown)


ghost:
Merged into 1.x and 1.15.x.

PR #3123, opened 2020-04-22
Issue #4375: Change number of images in Image Library view.

### LINKED ISSUE #4375: Image Library view: Better number of items per page
In the Image Library view, 20 thumbnail images are shown by default. These are displayed in a responsive grid, so the number of columns depends on the user's screen/window size.

For example, here's what it looks like on my screen:
image
It looks weird that the bottom row isn't filled...

Now I know this is because we need to specify a set number of images, and so the last row will be often only partially filled (depending on the number of columns). However I believe we can optimise the number of images displayed to have the least amount of empty last rows as possible.

I made a spreadsheet to calculate the number of columns a given number of images would fill completely for different widths. Here're the results:
image

The numbers down the left side are the number of images in the view (e.g. the pager setting). The numbers across the top are columns (different screen sizes would have different columns). The check marks show which number of images would completely fill which columns (e.g. 10 is divisible by 2, 5 & 10). The numbers down the right are simply a count of the check marks for each row.

The current number of images (20) is highlighted yellow. There are 4 column widths where the bottom row is completely filled.

I've highlighted the 24 images in green as I believe that might be a better number to use. It has 6 columns that are filled (which means less columns that aren't). Obviously there's no perfect answer (unless we're able to make a paging system that dynamically moves images to the next page that don't fit in the current screen width), but using 24 images in the Image Library will look less weird than 20.

ghost:
Fixes https://github.com/backdrop/backdrop-issues/issues/4375

### DIFF
--- a/core/modules/filter/config/views.view.image_library.json
+++ b/core/modules/filter/config/views.view.image_library.json
@@ -30,7 +30,7 @@
                 "pager": {
                     "type": "mini",
                     "options": {
-                        "items_per_page": "20",
+                        "items_per_page": "24",
                         "offset": "0",
                         "id": "0",
                         "total_pages": "",


quicksketch:
@BWPanda could you remake this PR as a branch in your personal repository fork? Right now it looks like this is a branch pushed into the main backdrop core repository?

ghost:
@quicksketch Yeah, it was just a quick change I made from the GitHub UI. Is that not allowed?

ghost:
Closing this PR/branch until I can add it properly to my personal fork.

PR #3124, opened 2020-04-23
Issue #4354: Forward revisions

### LINKED ISSUE #4354: [D8][WP][UX] Forward Revisions

docwilmot:
Fixes https://github.com/backdrop/backdrop-issues/issues/4354

### DIFF
--- a/core/modules/entity/entity.class.inc
+++ b/core/modules/entity/entity.class.inc
@@ -190,13 +190,20 @@ interface EntityInterface {
   /**
    * Checks if this entity is the active revision.
    *
-   * @param bool $new_value
-   *   (optional) A Boolean to (re)set the is_active_revision flag.
-   *
    * @return bool
    *   TRUE if the entity is the active revision, FALSE otherwise.
    */
-  public function isActiveRevision($new_value = NULL);
+  public function isActiveRevision();
+
+  /**
+   * Gets active revision for this entity.
+   */
+  public function getActiveRevision();
+
+  /**
+   * Sets this revision as the active revision.
+   */
+  public function setActiveRevision();
 }
 
 /**
@@ -210,6 +217,13 @@ abstract class Entity extends stdClass implements EntityInterface {
    * @var bool
    */
   public $is_active_revision = TRUE;
+
+  /**
+   * The active revision for this entity.
+   *
+   * @var int
+   */
+  public $active_revision;
 
   /**
    * Constructs a new entity object.
@@ -316,12 +330,22 @@ abstract class Entity extends stdClass implements EntityInterface {
   /**
    * Implements Backdrop\entity\EntityInterface::isActiveRevision().
    */
-  public function isActiveRevision($new_value = NULL) {
-    $return = $this->is_active_revision;
-    if (isset($new_value)) {
-      $this->is_active_revision = (bool) $new_value;
-    }
-    return $return;
+  public function isActiveRevision() {
+    return $this->is_active_revision;
+  }
+
+  /**
+   * Implements Backdrop\entity\EntityInterface::getActiveRevision().
+   */
+  public function getActiveRevision() {
+    return $this->active_revision;
+  }
+
+  /**
+   * Implements Backdrop\entity\EntityInterface::setActiveRevision().
+   */
+  public function setActiveRevision() {
+    $this->is_active_revision = TRUE;
   }
 }
 
--- a/core/modules/entity/entity.class.inc
+++ b/core/modules/entity/entity.class.inc
@@ -188,15 +188,15 @@ interface EntityInterface {
   public function getRevisionId();
 
   /**
-   * Checks if this entity is the default revision.
+   * Checks if this entity is the active revision.
    *
    * @param bool $new_value
-   *   (optional) A Boolean to (re)set the is_default_revision flag.
+   *   (optional) A Boolean to (re)set the is_active_revision flag.
    *
    * @return bool
-   *   TRUE if the entity is the default revision, FALSE otherwise.
+   *   TRUE if the entity is the active revision, FALSE otherwise.
    */
-  public function isDefaultRevision($new_value = NULL);
+  public function isActiveRevision($new_value = NULL);
 }
 
 /**
@@ -205,11 +205,11 @@ interface EntityInterface {
 abstract class Entity extends stdClass implements EntityInterface {
 
   /**
-   * Indicates whether this is the default revision.
+   * Indicates whether this is the active revision.
    *
    * @var bool
    */
-  public $is_default_revision = TRUE;
+  public $is_active_revision = TRUE;
 
   /**
    * Constructs a new entity object.
@@ -314,12 +314,12 @@ abstract class Entity extends stdClass implements EntityInterface {
   }
 
   /**
-   * Implements Backdrop\entity\EntityInterface::isDefaultRevision().
+   * Implements Backdrop\entity\EntityInterface::isActiveRevision().
    */
-  public function isDefaultRevision($new_value = NULL) {
-    $return = $this->is_default_revision;
+  public function isActiveRevision($new_value = NULL) {
+    $return = $this->is_active_revision;
     if (isset($new_value)) {
-      $this->is_default_revision = (bool) $new_value;
+      $this->is_active_revision = (bool) $new_value;
     }
     return $return;
   }
--- a/core/modules/entity/entity.class.inc
+++ b/core/modules/entity/entity.class.inc
@@ -177,12 +177,36 @@ interface EntityInterface {
    * @since 1.13.0 Method added.
    */
   public function getFieldValues($field_name, $value_key = 'value', $langcode = NULL);
+
+  /**
+   * Returns the revision identifier of the entity.
+   *
+   * @return
+   *   The revision identifier of the entity, or NULL if the entity does not
+   *   have a revision identifier.
+   */
+  public function getRevisionId();
+
+  /**
+   * Checks if this entity is the current revision.
+   *
+   * @return bool
+   *   TRUE if the entity is the current revision, FALSE otherwise.
+   */
+  public function isCurrentRevision();
 }
 
 /**
  * Defines a base entity class.
  */
 abstract class Entity extends stdClass implements EntityInterface {
+
+  /**
+   * Indicates whether this is the current revision.
+   *
+   * @var bool
+   */
+  public $isCurrentRevision = TRUE;
 
   /**
    * Constructs a new entity object.
@@ -277,6 +301,20 @@ abstract class Entity extends stdClass implements EntityInterface {
       }
     }
     return $values;
+  }
+
+  /**
+   * Implements Drupal\entity\EntityInterface::getRevisionId().
+   */
+  public function getRevisionId() {
+    return NULL;
+  }
+
+  /**
+   * Implements Drupal\entity\EntityInterface::isCurrentRevision().
+   */
+  public function isCurrentRevision() {
+    return $this->isCurrentRevision;
   }
 }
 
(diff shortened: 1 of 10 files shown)


core/modules/node/node.entity.inc
```diff
@@ -462,7 +478,14 @@ class NodeStorageController extends EntityDatabaseStorageController {
...
       else {
         $op = 'update';
-        $return = backdrop_write_record($this->entityInfo['base table'], $entity, $this->idKey);
+        // Update the base node table, but only if this revision is marked as
+        // the default.
+        if ($entity->isDefaultRevision()) {
+          $return = backdrop_write_record($this->entityInfo['base table'], $entity, $this->idKey);
+        }
+        else {
+          $return = SAVED_REVISION;
```
quicksketch: This may cause API breaks in some situations. Instead of a new return value, I think it may be preferable to return `SAVED_UPDATED` as before, even though we're not updating the entity base table.

This also would make it so that nodes wouldn't have a special 4th return value when compared to other types of entities.

core/modules/node/node.module
```diff
@@ -75,6 +75,12 @@ define('NODE_ACCESS_DENY', 'deny');
  */
 define('NODE_ACCESS_IGNORE', NULL);
 
+/**
+ * Return status for saving which involved an update to an existing node which
+ * was not the approved revision.
+ */
+define('SAVED_REVISION', 4);
```
quicksketch: Not sure about this being in Node module, but if we don't return a new value from `node_save()`, we don't need this new constant in the first place.

core/modules/entity/entity.class.inc
```diff
@@ -177,13 +177,40 @@ interface EntityInterface {
...
+  /**
+   * Checks if this entity is the default revision.
+   *
+   * @param bool $new_value
+   *   (optional) A Boolean to (re)set the is_default_revision flag.
+   *
+   * @return bool
+   *   TRUE if the entity is the default revision, FALSE otherwise.
+   */
+  public function isDefaultRevision($new_value = NULL);
```
quicksketch: I think separating the parameter for `$new_value` into a different method would be more clear. e.g. add a `setDefaultRevision()` in addition to `isDefaultRevision()`.

quicksketch (commented):
Thanks for making this API-only. It's a lot more manageable to review at this scope.

I'm not sure about the terminology here (sorry!) I'll post my thoughts on the naming of methods (e.g. `isDefaultRevision()`) in the main issue at https://github.com/backdrop/backdrop-issues/issues/4354

core/modules/entity/entity.class.inc
```diff
@@ -177,13 +177,54 @@ interface EntityInterface {
...
+
+  /**
+   * Gets active revision for this entity.
+   */
+  public function getActiveRevision();
+
+  /**
+   * Sets this revision as the active revision.
+   */
+  public function setActiveRevision();
```
quicksketch: Needs `@param` docblock and the default parameter.

core/modules/entity/entity.class.inc
```diff
@@ -278,6 +319,34 @@ abstract class Entity extends stdClass implements EntityInterface {
...
+   */
+  public function isActiveRevision() {
+    return $this->is_active_revision;
+  }
+
+  /**
+   * Implements Backdrop\entity\EntityInterface::getActiveRevision().
+   */
+  public function getActiveRevision() {
+    return $this->active_revision;
```
quicksketch: I didn't see this in the earlier iterations. I think this should be renamed to `getActiveRevisionId()` and `$this->active_revision_id` to make it clear you're getting an ID (or VID really) and not  an entire object. That would also better line up with the existing `$this->revisionId()`

core/modules/entity/entity.class.inc
```diff
@@ -278,6 +319,34 @@ abstract class Entity extends stdClass implements EntityInterface {
...
+  }
+
+  /**
+   * Implements Backdrop\entity\EntityInterface::setActiveRevision().
+   */
+  public function setActiveRevision() {
+    $this->is_active_revision = TRUE;
```
quicksketch, line 335: If we're updating `is_active_revision`, I think we should set `active_revision_id` to be the `$this->getRevisionId()` so that they match. In addition to that, I think that if `is_active_revision` and the revision ID do not match up when saving the entity, an exception should be thrown. For example this should not be allowed to save:

```
$node_revision = node_load(NULL, 10);
$node->setActiveRevision();
$node->active_revision_id = 9;
$node->save();
```

Because `is_active_revision` does not agree with the `vid` and `active_revision_id` -- both of which should be 10 at this point.

core/modules/entity/entity.controller.inc
```diff
@@ -437,7 +442,7 @@ class DefaultEntityController implements EntityControllerInterface {
    * @param $queried_entities
    *   Associative array of query results, keyed on the entity ID.
    * @param $revision_id
-   *   ID of the revision that was loaded, or FALSE if the most current revision
+   *   ID of the revision that was loaded, or FALSE if the most default revision
```
quicksketch: Let's use "active" in the code comments too.

core/modules/entity/entity.class.inc
docwilmot: But `ENTITY->active_revision_id` would be read-only so that code block wouldnt have an effect. 

However as it is now though, the `ENTITY->active_revision_id` property and the `getActiveRevisionId()` are both unused in the code, just added as utilities for UX. We could drop both of those if we think it isnt necessary, and skip having this discussion.

Aside, should `setActiveRevision()` really be `setAsActiveRevision()`?

core/modules/entity/entity.controller.inc
docwilmot: I saw that but didnt think it makes sense to say the "most active". That comment means the most current or most recent I think, not the most active/default/approved revision.

quicksketch: Ah I see, so maybe just leave it "most current revision" as it was before. Or "most recent" if that's actually more accurate.

core/modules/entity/entity.class.inc
quicksketch: > Aside, should setActiveRevision() really be setAsActiveRevision()?

Maybe `setIsActiveRevision()` would be the best alignment.

> But ENTITY->active_revision_id would be read-only so that code block wouldnt have an effect.

PHP doesn't have read-only properties, so you _could_ write to it even if you weren't supposed to. The only way to prevent writing is to it private and only accessible via `getActiveRevisionId()`.

But yeah, if we're not using it in the first place, perhaps save it for later.

core/modules/entity/entity.class.inc
docwilmot: I dont get this. THere are no parameters to this function?

quicksketch: Sorry, my mistake. I thought there was a boolean TRUE/FALSE parameter that was missing.

core/modules/entity/entity.controller.inc
quicksketch: I'm reading this more deeply now and I think it needs to be updated still. I think the correct phrasing is:

"ID of the revision that was loaded, or FALSE if no revision was specified."

We could add further that "If no revision is specified, the active revision is being loaded." But it's probably unnecessary for this particular method to get that in-depth on this behavior.

core/modules/node/node.module
```diff
@@ -1078,9 +1078,9 @@ function node_delete_multiple($nids) {
  */
 function node_revision_delete($revision_id) {
   if ($revision = node_load(NULL, $revision_id)) {
-    // Prevent deleting the current revision.
+    // Prevent deleting the default revision.
```
quicksketch: "default" should be "active" in these code comments. (This one and about a dozen more below it).

quicksketch (commented):
Looks great. The only complaints I have are code comments' use of the word "default". I think these should be "active" to be consistent.

docwilmot:
Sandbox!

quicksketch:
Merged into 1.x for 1.16.0. Woot!

PR #3128, opened 2020-04-27
3402 admin label

### LINKED ISSUE #3402: [UX] Standardize Label and `machine name` usage in admin listings

In https://github.com/backdrop/backdrop-issues/issues/3395 the `machine name` value was moved into the `View name` column:

screen shot 2018-11-30 at 2 57 31 pm

I've followed that pattern in [recent changes in the Paragraphs module in contrib](https://github.com/backdrop-contrib/paragraphs/issues/23#issuecomment-443291869), and I propose it would be good to use it more broadly in core and make things consistent. We'd first need to get a list of all the pages that list `machine name` as a separate column, but for starters here are some suggestions/questions:

 - all the `Manage fields` pages have separate columns (on content types, user account fields, taxonomy vocabulary fields)
screen shot 2018-11-29 at 5 30 11 pm

 - Content types list machine name after the label but include the words "Machine name: " which could maybe be removed:
screen shot 2018-11-30 at 3 05 38 pm

 - On the other hand, Taxonomy vocabularies don't list the machine name at all (although it can be useful to have):
screen shot 2018-11-30 at 3 11 27 pm

---

**Final interfaces after the PR:**

Content types list:
Screen Shot 2021-08-29 at 3 48 28 PM

Taxonomy vocabularies:
Screen Shot 2021-08-29 at 3 48 49 PM

Custom blocks:
Screen Shot 2021-08-29 at 3 50 15 PM

Manage displays:
Screen Shot 2021-08-29 at 3 51 08 PM

File types:
Screen Shot 2021-08-29 at 3 51 08 PM

jenlampton:
Fixes https://github.com/backdrop/backdrop-issues/issues/3402

This PR adds a new theme function `theme_admin_info()` into system module that can be used universally. 

* The function `theme_node_admin_overview()` is marked as deprecated.
* Node types listing now calls `theme_admin_info()` instead.
* Layout listing now calls `theme_admin_info()`.
* The function `theme_file_type_overview` is marked as deprecated.
* File types listing now calls `theme_admin_info()` instead.
* The function `theme_menu_admin_overview()' is marked as deprecated.
* Taxonomy vocabulary listing now calls `theme_admin_info()`.
* The menu listing now calls `theme_admin_info()` instead.
* The function `theme_views_ui_view_name()` is marked as deprecated.
* The views listing now calls `theme_admin_info()`.
* The image style listing now calls `theme_admin_info()`.
* The text format listing now calls `theme_admin_info()`.
* The date format listing now calls `theme_admin_info()`.

### DIFF
--- a/core/modules/menu/menu.admin.inc
+++ b/core/modules/menu/menu.admin.inc
@@ -14,7 +14,12 @@ function menu_overview_page() {
   $rows = array();
   foreach ($menus as $menu) {
     $row = array();
-    $row[] = theme('menu_admin_overview', array('title' => $menu['title'], 'name' => $menu['menu_name'], 'description' => $menu['description']));
+    $label_variables = array(
+      'label' => $menu['title'],
+      'machine' => $menu['menu_name'],
+      'description' => $menu['description'],
+    );
+    $row[] = theme('machine_name_label__menu', $label_variables);
     $links = array();
     $links['list'] = array(
       'title' => t('Manage links'),
--- a/core/modules/system/system.module
+++ b/core/modules/system/system.module
@@ -86,6 +86,9 @@ function system_theme() {
     'status_report' => array(
       'render element' => 'requirements',
     ) + $base,
+    'machine_name_label' => array(
+      'variables' => array('label' => NULL, 'machine' => NULL, 'description' => NULL, 'inline' => TRUE),
+    ) + $base,
     'admin_page' => array(
       'variables' => array('blocks' => NULL),
     ) + $base,
--- a/core/modules/system/system.theme.inc
+++ b/core/modules/system/system.theme.inc
@@ -506,6 +506,42 @@ function theme_admin_block_content($variables) {
 }
 
 /**
+ * Returns HTML for the display of label with both Human and Machine names.
+ *
+ * @param $variables
+ *   An associative array containing:
+ *   - label: The human-readable name or label.
+ *   - machine: The machine-safe name or label.
+ *   - description: (optional) The description of the item.
+ *   - inline: TRUE if the machine name should be on the same line. Defaults to TRUE.
+ *
+ * @ingroup themeable
+ */
+function theme_admin_info($variables) {
+  $classes = array('admin-label');
+  if ($variables['inline']) {
+    $classes[] = 'inline';
+  }
+
+  $output  = '<div class="' . implode(' ', $classes) . '">';
+  $output .=   '<strong>' . check_plain($variables['label']) . '</strong>';
+  if (!empty($variables['machine'])) {
+    $machine_name = t('Machine name: @machine', array('@machine' => $variables['machine']));
+    if ($variables['inline']) {
+      $output .= '&nbsp;';
+    }
+    $output .=   '<small>(' . $machine_name . ')</small>';
+  }
+  $output .= '</div>';
+
+  if (isset($variables['description'])) {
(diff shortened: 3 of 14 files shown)


core/modules/menu/menu.admin.inc
```diff
@@ -14,7 +14,7 @@ function menu_overview_page() {
   $rows = array();
   foreach ($menus as $menu) {
     $row = array();
-    $row[] = theme('menu_admin_overview', array('title' => $menu['title'], 'name' => $menu['menu_name'], 'description' => $menu['description']));
+    $row[] = theme('admin_info', array('label' => $menu['title'], 'machine' => $menu['menu_name'], 'description' => $menu['description']));
```
quicksketch: If we're no longer calling `theme_menu_admin_overview` at all, maybe it should be removed entirely? Seems unlikely that another place would actually call that theme function.

core/modules/system/system.module
```diff
@@ -86,6 +86,9 @@ function system_theme() {
     'status_report' => array(
       'render element' => 'requirements',
     ) + $base,
+    'admin_info' => array(
+      'variables' => array('label' => NULL, 'machine' => NULL, 'inline' => TRUE),
+    ) + $base,
```
quicksketch, line 91: The name "admin_info" seems ambiguous. How about "machine_name_label"?

Also, I think everywhere this is called it should be done with a theme suggestion (double-underscore), e.g. `theme('machine_name_label__node_type')` or `theme('machine_name_label__menu')`

core/modules/system/system.theme.inc
```diff
@@ -505,6 +505,42 @@ function theme_admin_block_content($variables) {
...
+  }
+
+  $output  = '<div class="' . implode(' ', $classes) . '">';
+  $output .=   '<strong>' . check_plain($variables['label']) . '</strong>';
+  if (!empty($variables['machine'])) {
+    $machine_name = t('Machine name: @machine', array('@machine' => $variables['machine']));
+    if ($variables['inline']) {
+      $output .= '&nbsp;';
+    }
+    $output .=   '<small>(' . $machine_name . ')</small>';
```
quicksketch, line 533:
```suggestion
    $output .= '<small>(' . $machine_name . ')</small>';
```

core/modules/system/system.theme.inc
```diff
@@ -505,6 +505,42 @@ function theme_admin_block_content($variables) {
...
+ * @ingroup themeable
+ */
+function theme_admin_info($variables) {
+  $classes = array('admin-label');
+  if ($variables['inline']) {
+    $classes[] = 'inline';
+  }
+
+  $output  = '<div class="' . implode(' ', $classes) . '">';
+  $output .=   '<strong>' . check_plain($variables['label']) . '</strong>';
```
quicksketch, line 527: I see what's happening here but these indentations are unusual.

```suggestion
  $output .= '<strong>' . check_plain($variables['label']) . '</strong>';
```

core/modules/menu/menu.admin.inc
jenlampton: removed.

core/modules/system/system.module
jenlampton: good ideas. done.

jenlampton:
Closing in favor of https://github.com/backdrop/backdrop/pull/3473

PR #3152, opened 2020-05-25
Issue #4420: Allow blocks to be removed when changing layout templates.

### LINKED ISSUE #4420: Cannot remove blocks when switching layouts anymore
**Description of the bug**

The [documentation](https://api.backdropcms.org/api/backdrop/core%21modules%21layout%21includes%21layout.class.inc/function/Layout%3A%3AsetLayoutTemplate/1) for the `$region_mapping` parameter of `Layout::setLayoutTemplate()` says:

> An array of region names, each keyed by the old region name and the value of the new region name. If empty, all known regions will be copied from one to the other. **To discard the contents of a region, set the value to FALSE.**

However this doesn't always seem to be the case...

**Steps To Reproduce**

To reproduce the behavior:
1. On a new Backdrop site, install and enable the Devel module
1. Note the blocks/regions in the Default layout: `/admin/structure/layouts/manage/default`
1. Go to `/admin/devel/php` and run this code (it changes the layout for the Default layout to 'Boxton' and says to discard all blocks in the process):
   ```php
   $layout = layout_load('default', TRUE);
   $layout->setLayoutTemplate('boxton', array(
     'header' => FALSE,
     'top' => FALSE,
     'content' => FALSE,
     'footer' => FALSE,
   ));
   $layout->save();
   ```
1. Note the blocks/regions in the Default layout now: `/admin/structure/layouts/manage/default`

**Actual behavior**

Blocks are removed from all but the 'Content' region. Blocks from other regions moved to 'Content'.

**Expected behavior**

All regions empty, no blocks remaining.

ghost:
Fixes https://github.com/backdrop/backdrop-issues/issues/4420

### DIFF
--- a/core/modules/layout/includes/layout.class.inc
+++ b/core/modules/layout/includes/layout.class.inc
@@ -718,16 +718,19 @@ class Layout {
         $new_region_name = $old_region_name;
       }
       else {
-        $new_region_name = FALSE;
+        $new_region_name = '';
       }
 
       if (isset($this->positions[$old_region_name])) {
         foreach ($this->positions[$old_region_name] as $uuid) {
-          if ($new_region_name) {
+          if (!empty($new_region_name)) {
             $new_positions[$new_region_name][] = $uuid;
           }
-          else {
+          elseif ($new_region_name === '') {
             $this->orphaned_blocks[] = $uuid;
+          }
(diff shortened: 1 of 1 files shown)


quicksketch:

PR #3153, opened 2020-05-26
Issue #4423: Fix broken t() usage in function update_mail

### LINKED ISSUE #4423: Warning: htmlspecialchars() ... on cron runs when an update is available
**Description of the bug**

When an update is available, whenever Backdrop checks for them on my sites (according to the update interval setting (daily / weekly...)) via cron run, a php warning appears twice in dblog:

`Warning: htmlspecialchars() expects parameter 1 to be string, array given in check_plain() (... 2054 ...core/includes/bootstrap.inc).`

**Steps To Reproduce**

That's a bit tricky, as this only appears when updates are available. Not if there aren't any and not if running the update check via UI. Also, it might be necessary to install an additional language. Note: the sites aren't multilingual, it's just that the default language isn't English.

**Additional information**

This warning isn't new, I'm not even sure, if it hasn't been there ever since I'm using Backdrop. At least I don't remember when I first saw that in dblog.

This time I was smart and put a `debug($text)` in bootstrap.inc of a dev site. This is what check_plain() gets as array then (for the two notices):

1. `array()`
2. `array('langcode' => 'en',)`

indigoxela:
Fixes https://github.com/backdrop/backdrop-issues/issues/4423

### DIFF
--- a/core/modules/update/update.module
+++ b/core/modules/update/update.module
@@ -445,7 +445,7 @@ function _update_get_cached_available_releases() {
 function update_mail($key, &$message, $params) {
   $language = $message['language'];
   $langcode = $language->langcode;
-  $site_name_localized =  config_get_translated('system.core', 'site_name', array(), array('langcode' => $langcode));
+  $site_name_localized = config_get_translated('system.core', 'site_name', array(), array('langcode' => $langcode));
   $message['subject'] .= t('New release(s) available for !site_name', array('!site_name' => $site_name_localized), array('langcode' => $langcode));
   foreach ($params as $msg_type => $msg_reason) {
     $message['body'][] = _update_message_text($msg_type, $msg_reason, FALSE, $language);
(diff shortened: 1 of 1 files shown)


core/modules/update/update.module
```diff
@@ -445,7 +445,8 @@ function _update_get_cached_available_releases() {
 function update_mail($key, &$message, $params) {
   $language = $message['language'];
   $langcode = $language->langcode;
-  $message['subject'] .= t('New release(s) available for !site_name', array('!site_name' => config_get_translated('system.core', 'site_name'), array(), array('langcode' => $langcode)), array('langcode' => $langcode));
+  $site_name_localized =  config_get_translated('system.core', 'site_name', array(), array('langcode' => $langcode));
```
ghost: Extra space here after '=' can be removed.

ghost (changes requested):

core/modules/update/update.module
indigoxela: Oops, correct, I'll remove that.

ghost (approved):

klonos (approved):
LGTM 👍

ghost:
Thank you! Merged into 1.x and 1.16.x.

PR #3157, opened 2020-06-03
Issue #4428: Allow acount administratve approval email address to be …

### LINKED ISSUE #4428: User module still sends messages TO `site_mail`, the FROM email address
Follow up to https://github.com/backdrop/backdrop-issues/issues/4403

> When testing this, make sure your site mail (admin/config/system/site-information) is a real address able to receive mail.

no no we shouldn't be telling people to do this. This is likely to make all messages coming from the site undeliverable :(

> Which leads me to a question: shouldn't we use user number one's (uid 1) email address instead as recipient?

Yes, Contact module also had this problem, and now uses `user/1` as a default when no other email address has been set.

There should be nothing anywhere in backdrop core that sends mail TO `site_mail` that is a FROM email address only. If there is, that is an oversight and was not corrected when we removed `site_mail` from the installation form. 

> Or even better: make the recipient address configurable. The site email is a bad choice for sure, but uid 1 might also be inappropriate.

I would like to see this be configureable! Perhaps on the same page where the text for the emails can be configured.

jenlampton:
Fixes https://github.com/backdrop/backdrop-issues/issues/4428

### DIFF
--- a/core/modules/user/user.module
+++ b/core/modules/user/user.module
@@ -1,4 +1,4 @@
-<?php
+ <?php
 /**
  * @file
  * Enables the user registration and login system.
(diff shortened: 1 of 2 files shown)


core/modules/user/user.module
```diff
@@ -1,4 +1,4 @@
-<?php
+ <?php
```
klonos, line 1:
```suggestion
<?php
```

PR #3164, opened 2020-06-08
Issue #52: Add assignable roles setting for role assign feature.

### LINKED ISSUE #4437: [D7] Add setting for which roles can be assigned
**Description of the need**

In Drupal 7, the "assign roles" module had a setting to define which roles could be assigned. This made it possible to grant someone the permission to `Assign roles` without allowing themselves to increase their own role beyond what was indented.

I would love to see this setting in Backdrop, with the `Administrative role` initially excluded. This would make for a safer set-up out-of-the box.

**Proposed solution**

Add a setting for roles that are allowed to be assigned. 

Port the assign roles module, and have it include only the setting (since the feature has been added to core)

**Additional information**

We initially created this as a security issue, since it has security implications. However, since the permission is no more dangerous than "Administer users" -- which was the only way people could assign roles before this feature was added, we decided to move the PR to the public queue for more reviews.

jenlampton:
Fixes https://github.com/backdrop/backdrop-issues/issues/4437


core/modules/user/user.admin.inc
```diff
@@ -500,6 +500,22 @@ function user_admin_roles($form, $form_state) {
...
+    '#collapsed' => TRUE,
+    '#weight' => -20,
+  );
+  unset($roles[0]);
+  $form['assign_roles']['user_assignable_roles'] = array(
+    '#type' => 'checkboxes',
+    '#title' => t('Roles available for assignment'),
+    '#default_value' => $system_config->get('user_assignable_roles'),
+    '#options' => $roles,
+    '#description' => t('Anyone with the <em>assign roles</em> permission will be able to grant these roles themselves, as well as to others.'),
```
klonos:
```suggestion
    '#description' => t('Anyone with the <em>Assign roles</em> permission will be able to grant these roles to themselves, as well as to others.'),
```

PR #3166, opened 2020-06-11
Issue #4439: Denote which theme has been set as the admin theme

### LINKED ISSUE #4439: [UX][D8] Denote which theme has been set as the admin theme
We denote which one has been set as the default theme:

Screen Shot 2020-06-11 at 9 45 07 pm

We should also denote which one is the admin theme, same as D8/9:

Screen Shot 2020-06-11 at 9 47 18 pm

klonos:
Fixes https://github.com/backdrop/backdrop-issues/issues/4439

### DIFF
--- a/core/modules/system/system.admin.inc
+++ b/core/modules/system/system.admin.inc
@@ -111,7 +111,10 @@ function system_themes_page() {
   $themes = system_rebuild_theme_data();
   uasort($themes, 'system_sort_modules_by_info_name');
 
-  $theme_default = config_get('system.core', 'theme_default');
+  $config = config('system.core');
+  $theme_default = $config->get('theme_default');
+  $admin_theme = $config->get('admin_theme');
+
   $theme_groups  = array(
     'enabled' => array(),
     'disabled' => array(),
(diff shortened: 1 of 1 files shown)


core/modules/system/system.admin.inc
```diff
@@ -123,6 +126,7 @@ function system_themes_page() {
     }
     $admin_theme_options[$theme->name] = $theme->info['name'];
     $theme->is_default = ($theme->name == $theme_default);
+    $theme->is_admin = ($theme->name == $admin_theme);
```
klonos, line 129: The way this is implemented these properties are available only within the context of the themes listing page. I would like us to move these into `system_rebuild_theme_data()` and/or `list_themes()`, so that they become available info on themes in general. Would this be an API change?

quicksketch: I think moving it into `list_themes()` would be appropriate. It would be an API expansion to do so, since these properties would be made available in more places than they had been previously. That's fine to do in our 1.x release. In our docblock for `list_themes()` we'd just have to include documentation on the new properties and include `@since` sections for the new properties.

Then instead of this function (`system_themes_page()`) calling `$themes = system_rebuild_theme_data();`, it should pull from `list_themes()` instead so it doesn't have to duplicate the work.

PR #3169, opened 2020-06-17
Harden form processing when tokens are invalid.

quicksketch:
Resolves https://github.com/backdrop-ops/security/issues/97.


quicksketch:
Merged into 1.x, 1.16.x, and 1.15.x.

PR #3172, opened 2020-06-19
Issue #1478: Avoid race-condition in page header test. Fixing regression from #4306.

quicksketch:
Reverts some of the changes from issue https://github.com/backdrop/backdrop-issues/issues/4306 to prevent race conditions.


quicksketch:
Replaced by https://github.com/backdrop/backdrop/pull/3173.

PR #3176, opened 2020-06-28
Issue #2637: Transliteration for language neutral content

### LINKED ISSUE #2637: Transliteration doesn't respect site default language
In German, certain characters are supposed to be transliterated language-specific. Example: The letter `ä` should be transliterated to `a` for English but to `ae` for German. Backdrop doesn't deliver the language-specific transliterations unless you add *Multilingual support* to a content type and select the relevant language on the content items.

**Steps to reproduce:**

- Enable Language and Locale modules.
- *Don't* enable Content Translation module.
- Add Language: German
- Make German default.
- Disable the Detection Method "URL" to make sure that the default language is used in any case.
- Enable "Transliterate prior to creating URL aliases" (`admin/config/urls/path/patterns/settings`)
- Enable "Reduce URL alias characters to letters and numbers".
- *Don't* enable Multilingual Support for Posts.
- Add a Post with Title: Ärger im Büro

Resulting URL alias: `posts/arger-im-buro`

Expected alias: `posts/aerger-im-buero`

---

Current PR by @indigoxela: https://github.com/backdrop/backdrop/pull/3176 (transliterate neutral paths in current lang, like D7)

---

Outdated PRs:

<strike>PR by @olafgrabienski: https://github.com/backdrop/backdrop/pull/1849</strike> (outdated)
<strike>PR by @quicksketch: https://github.com/backdrop/backdrop/pull/1850</strike> (outdated, conflicts)
<strike>PR by @indigoxela: https://github.com/backdrop/backdrop/pull/3175</strike> (set node langcode, closed in favor of different approach)

---

**Original issue description:**

On a Backdrop 1.6.3 site which uses German as default language, transliteration for URL aliases and files are enabled and work generally, but they do not respect language-specific variations by default.

Certain characters are supposed to be transliterated language-specific. Example:

- **ä** => **a** (English)
- **ä** => **ae** (German)

At the moment, Backdrop doesn't deliver language-specific transliterations of URL aliases, unless you add Multilingual support to a content type and select the relevant language on the content items.

(Note: For localized sites which use only one language there's usually no need to add multilingual support to content types, and it's not an acceptable workaround to do so.)

At the top of `core/includes/transliteration.inc` I see the following comment which makes me think that Backdrop should use the site's default language for language-specific transliterations:

```
 * @param $source_langcode
 *   Optional ISO 639 language code that denotes the language of the input and
 *   is used to apply language-specific variations. If the source language is
 *   not known at the time of transliteration, it is recommended to set this
 *   argument to the site default language to produce consistent results.
 *   Otherwise the current display language will be used.
```

Unfortunately I don't have the skills to check `$source_langcode` in the code itself. Could someone have a look at it and give feedback how it's supposed to work and/or why it doesn't?

indigoxela:
Fixes https://github.com/backdrop/backdrop-issues/issues/2637

This PR handles transliteration for language neutral content paths like the D7 module does - by choosing the current language.

### DIFF
--- a/core/modules/path/path.inc
+++ b/core/modules/path/path.inc
@@ -184,6 +184,12 @@ function path_clean_string($string, array $options = array()) {
   elseif (!empty($options['langcode'])) {
     $langcode = $options['langcode'];
   }
+  if ($langcode == LANGUAGE_NONE) {
+    // Paths for language neutral content get transliterated according to
+    // current language.
+    global $language;
+    $langcode = $language->langcode;
+  }
 
   // Check if the string has already been processed, and if so return the
   // cached result.
--- a/core/modules/path/path.inc
+++ b/core/modules/path/path.inc
@@ -188,6 +188,9 @@ function path_clean_string($string, array $options = array()) {
     // Paths for language neutral content get transliterated according to
     // current language.
     global $language;
+    // We are intentionally not using config_get('system.core', 'language_default')
+    // here. That can have unexpected behavior in cases such as on a multilingual
+    // site with language-neutral content types.
     $langcode = $language->langcode;
   }
 
--- a/core/modules/path/tests/path_pattern.test
+++ b/core/modules/path/tests/path_pattern.test
@@ -813,13 +813,19 @@ class PathPatternLocaleTestCase extends PathPatternFunctionalTestHelper {
     $modules[] = 'translation';
     parent::setUp($modules, array('administer languages'));
 
-    // Add predefined French language and reset the locale cache.
-    $language = (object) array(
+    // Add predefined French and German languages and reset the locale cache.
+    $french = (object) array(
       'langcode' => 'fr',
       'name' => 'French',
       'direction' => LANGUAGE_LTR,
     );
-    language_save($language);
+    language_save($french);
+    $german = (object) array(
+      'langcode' => 'de',
+      'name' => 'German',
+      'direction' => LANGUAGE_LTR,
+    );
+    language_save($german);
     backdrop_language_initialize();
   }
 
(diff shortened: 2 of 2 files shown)


core/modules/path/path.inc
```diff
@@ -184,6 +184,12 @@ function path_clean_string($string, array $options = array()) {
   elseif (!empty($options['langcode'])) {
     $langcode = $options['langcode'];
   }
+  if ($langcode == LANGUAGE_NONE) {
+    // Paths for language neutral content get transliterated according to
+    // current language.
+    global $language;
+    $langcode = $language->langcode;
```
quicksketch: Would this work instead? It makes me uncomfortable that the language in which the interface is displayed would affect how a path is created. This would instead check the default language, which in the case of single-language sites, would result in the same outcome.

```suggestion
    $langcode = config_get('system.core', 'language_default');
```

indigoxela: `language_default` - That's what @olafgrabienski suggested, but it has side effects. It could be the wrong choice on a multilingual site with language neutral content types.

`global $language` That's exactly how D7 Transliteration module does it. And no one seemed to have trouble with it the past years.

https://git.drupalcode.org/project/transliteration/-/blob/7.x-3.x/transliteration.inc#L182

indigoxela:
Requesting fresh sandbox.

indigoxela:
Requesting fresh sandbox, second try. (This time: "Could not open input file: core/scripts/run-tests.sh"). Github, please show mercy!

core/modules/path/path.inc
klonos: At the very least, we should add an inline comment with what you said @indigoxela:

```suggestion
    global $language;
    // We are intentionally not using config_get('system.core', 'language_default')
    // here. That can have unexpected behavior in cases such as on a multilingual
    // site with language-neutral content types.
    $langcode = $language->langcode;
```

...that way, next person that reads this and wonders the same thing as @quicksketch will get an explanation.

core/modules/path/tests/path_pattern.test
```diff
@@ -848,6 +864,27 @@ class PathPatternLocaleTestCase extends PathPatternFunctionalTestHelper {
...
   }
+
+  /**
+   * Language neutral German node has the correct transliteration applied.
+   */
+  public function testNeutralTransliteration() {
+    config_set('system.core', 'language_default', 'de');
+    config_set('path.settings', 'transliterate', TRUE);
+    // Note: "ü" -> "ue", "ß" -> "ss".
+    $title = 'Lückenbüßer';
```
quicksketch: Could you save two nodes, both with this title? Or `Lückenbüßer DE` and `Lückenbüßer EN` if you want to load them by title. I think it's important to test that the exact same title is transliterated differently based on the site default language.

indigoxela: That's a good idea.

indigoxela:
Requesting fresh sandbox

klonos (approved):
Code looks good (and includes the change requested by @quicksketch re the tests) 👍

ghost:

PR #3178, opened 2020-06-29
Issue #4461: Dedicated setting for admin theme same as default

### LINKED ISSUE #4461: [UX][DX] Remove the option for configuring the admin theme to always be the same as the default theme.
This annoys me on both a UX and DX level...

## UX

Screen Shot 2020-06-28 at 11 30 24 am

This ^^ is a Drupalism/Backdropism, and we should fix it! The "Choose "Default theme" to ..." help text is similarly annoying to all the "leave blank to ..." help texts; we shouldn't have to be explaining "special" options, when we can use clearly distinct UI elements that convey clear messages of what each form element does.

As a newcomer, when I click that dropdown select, I am assuming that all the available options there are themes ...the "Default theme" option is not though; and what if there is an actual theme, with "Default theme" as its name? (but more on that in the DX section).

Here are the 2 alternatives that would make much more sense from a UX PoV:

- We could make it clear that the first option is a "special option" (and remove the help text):

    Screen Shot 2020-06-28 at 12 30 44 pm

- Or ideally remove the "special option", and replace it with a distinct checkbox instead:

    Screen Shot 2020-06-28 at 12 01 13 pm

    Notice how the vertical space is not increased ^^ (so no additional scrolling required), because we have practically converted the help text into a checkbox 😉 ...if the checkbox is ticked, the admin theme select would be hidden via `#states`:

    Screen Shot 2020-06-28 at 12 01 37 pm

## DX

This is the "magic" we use to inject that "special" option in the select options:

```php
  $form['admin_theme']['admin_theme'] = array(
    '#type' => 'select',
    '#options' => array(0 => t('Default theme')) + $theme_options,
    '#title' => t('Administration theme'),
    '#description' => t('Choose "Default theme" to always use the same theme as the rest of the site.'),
    '#default_value' => config_get('system.core', 'admin_theme'),
  );
```

...so when that option is selected, `config_get('system.core', 'admin_theme')` returns a string `0` instead of a theme name (what is actually saved in `system.core.json` is `"admin_theme": "0",`) 🤦 ...some may say that this is fine, but then instead of being able to do simple, straight-forward things like this (note that almost no comments to explain what's going on):

```php
  $theme_default = $config->get('theme_default');
  $theme_admin = $config->get('admin_theme');
  ...
  // Add notes to default and administration theme.
  $theme->notes = array();
  $theme->classes = array();
  if ($theme->is_default) {
    $theme->classes[] = 'theme-default';
    $theme->notes[] = t('default theme');
  }
  if ($theme->is_admin) {
    $theme->classes[] = 'theme-admin';
    $theme->notes[] = t('administration theme');
  }
```

...one has to do this:

```php
  $theme_default = $config->get('theme_default');
  $theme_admin = $config->get('admin_theme');
  ...
  // Add notes to default and administration theme.
  $theme->notes = array();
  $theme->classes = array();
  if ($theme->is_default) {
    $theme->classes[] = 'theme-default';
    $theme->notes[] = t('default theme');
  }
  // When the admin theme is set to be the same as the default theme, then
  // admin_theme in system.core.json is set to the string "0".
  if ($theme->is_admin || ($admin_theme == FALSE && $theme->is_default)) {
    $theme->classes[] = 'theme-admin';
    $theme->notes[] = t('administration theme');
  }
```

klonos:
Fixes https://github.com/backdrop/backdrop-issues/issues/4461

### DIFF
--- a/core/modules/simpletest/tests/batch.test
+++ b/core/modules/simpletest/tests/batch.test
@@ -287,11 +287,11 @@ class BatchPageTestCase extends BackdropWebTestCase {
   function testBatchProgressPageTheme() {
     // Make sure that the page which starts the batch (an administrative page)
     // is using a different theme than would normally be used by the batch API.
-    config('system.core')
-      ->set('admin_theme_same_as_default', 0)
-      ->set('theme_default', 'bartik')
-      ->set('admin_theme', 'seven')
-      ->save();
+    $config = config('system.core');
+    $config->set('admin_theme_same_as_default', 0);
+    $config->set('theme_default', 'bartik');
+    $config->set('admin_theme', 'seven');
+    $config->save();
     // Log in as an administrator who can see the administrative theme.
     $admin_user = $this->backdropCreateUser(array('view the administration theme'));
     $this->backdropLogin($admin_user);
--- a/core/modules/simpletest/tests/batch.test
+++ b/core/modules/simpletest/tests/batch.test
@@ -287,8 +287,11 @@ class BatchPageTestCase extends BackdropWebTestCase {
   function testBatchProgressPageTheme() {
     // Make sure that the page which starts the batch (an administrative page)
     // is using a different theme than would normally be used by the batch API.
-    config_set('system.core', 'theme_default', 'bartik');
-    config_set('system.core', 'admin_theme', 'seven');
+    $config = config('system.core');
+    $config->set('admin_theme_same_as_default', 0);
+    $config->set('theme_default', 'bartik');
+    $config->set('admin_theme', 'seven');
+    $config->save();
     // Log in as an administrator who can see the administrative theme.
     $admin_user = $this->backdropCreateUser(array('view the administration theme'));
     $this->backdropLogin($admin_user);
--- a/core/modules/simpletest/tests/batch.test
+++ b/core/modules/simpletest/tests/batch.test
@@ -287,11 +287,11 @@ class BatchPageTestCase extends BackdropWebTestCase {
   function testBatchProgressPageTheme() {
     // Make sure that the page which starts the batch (an administrative page)
     // is using a different theme than would normally be used by the batch API.
-    $config = config('system.core');
-    $config->set('admin_theme_same_as_default', 0);
-    $config->set('theme_default', 'bartik');
-    $config->set('admin_theme', 'seven');
-    $config->save();
+    config('system.core')
+      ->set('admin_theme_same_as_default', 0)
+      ->set('theme_default', 'bartik')
+      ->set('admin_theme', 'seven')
+      ->save();
     // Log in as an administrator who can see the administrative theme.
     $admin_user = $this->backdropCreateUser(array('view the administration theme'));
     $this->backdropLogin($admin_user);
--- a/core/modules/simpletest/tests/menu.test
+++ b/core/modules/simpletest/tests/menu.test
@@ -129,11 +129,11 @@ class MenuRouterTestCase extends BackdropWebTestCase {
     // Make the tests below more robust by explicitly setting the default theme
     // and administrative theme that they expect.
     theme_enable(array('bartik'));
-    config('system.core')
-      ->set('admin_theme_same_as_default', 0)
-      ->set('theme_default', 'bartik')
-      ->set('admin_theme', 'seven')
-      ->save();
+    $config = config('system.core');
+    $config->set('admin_theme_same_as_default', 0);
+    $config->set('theme_default', 'bartik');
+    $config->set('admin_theme', 'seven');
+    $config->save();
   }
 
   /**
--- a/core/modules/simpletest/tests/menu.test
+++ b/core/modules/simpletest/tests/menu.test
@@ -129,8 +129,11 @@ class MenuRouterTestCase extends BackdropWebTestCase {
     // Make the tests below more robust by explicitly setting the default theme
     // and administrative theme that they expect.
     theme_enable(array('bartik'));
-    config_set('system.core', 'theme_default', 'bartik');
-    config_set('system.core', 'admin_theme', 'seven');
+    $config = config('system.core');
+    $config->set('admin_theme_same_as_default', 0);
+    $config->set('theme_default', 'bartik');
+    $config->set('admin_theme', 'seven');
+    $config->save();
   }
 
   /**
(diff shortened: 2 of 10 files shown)


core/modules/simpletest/tests/batch.test
```diff
@@ -287,8 +287,11 @@ class BatchPageTestCase extends BackdropWebTestCase {
   function testBatchProgressPageTheme() {
     // Make sure that the page which starts the batch (an administrative page)
     // is using a different theme than would normally be used by the batch API.
-    config_set('system.core', 'theme_default', 'bartik');
-    config_set('system.core', 'admin_theme', 'seven');
+    config('system.core')
```
jenlampton: Can we switch this to the syntax where every action completes on a single line. Like this:

```php
$config->set('admin_theme_same_as_default', 0);
$config->set('theme_default', 'bartik');
$config->set('admin_theme', 'seven');
$config->save();
```

It's both more intuitive and will be consistent with the rest of core.

core/modules/simpletest/tests/menu.test
```diff
@@ -129,8 +129,11 @@ class MenuRouterTestCase extends BackdropWebTestCase {
     // Make the tests below more robust by explicitly setting the default theme
     // and administrative theme that they expect.
     theme_enable(array('bartik'));
-    config_set('system.core', 'theme_default', 'bartik');
-    config_set('system.core', 'admin_theme', 'seven');
+    config('system.core')
```
jenlampton: Same thing here:
```php
$config->set('admin_theme_same_as_default', 0)
$config->set('theme_default', 'bartik')
$config->set('admin_theme', 'seven')
$config->save();
```

core/modules/system/system.admin.inc
```diff
@@ -234,14 +234,27 @@ function system_themes_admin_form($form, &$form_state, $theme_options) {
   // Administration theme settings.
   $form['admin_theme'] = array(
     '#type' => 'fieldset',
-    '#title' => t('Administration theme'),
+    '#title' => t('Administration theme settings'),
+  );
+  $form['admin_theme']['admin_theme_same_as_default'] = array(
+    '#type' => 'checkbox',
+    '#title' => t('Always use the default theme as the administration theme'),
```
jenlampton: Do we need the "Always" here? What about just `Use the default theme as the administration theme`.

core/modules/taxonomy/tests/taxonomy.test
```diff
@@ -1782,8 +1782,11 @@ class TaxonomyThemeTestCase extends TaxonomyWebTestCase {
 
     // Make sure we are using distinct default and administrative themes for
     // the duration of these tests.
-    config_set('system.core', 'theme_default', 'bartik');
-    config_set('system.core', 'admin_theme', 'seven');
+    config('system.core')
```
jenlampton: same thing here...
```php
$config->set('admin_theme_same_as_default', 0);
$config->set('theme_default', 'bartik');
$config->set('admin_theme', 'seven');
$config->save();
```

core/profiles/standard/standard.install
```diff
@@ -440,17 +434,25 @@ function standard_install() {
...
-    ->condition('name', 'seven')
-    ->execute();
-  config_set('system.core', 'admin_theme', 'seven');
-  config_set('system.core', 'node_admin_theme', '1');
+  $default_theme = 'basis';
+  $admin_theme = 'seven';
+  // Enable the Basis and Seven themes, and disable the Stark theme.
+  theme_enable(array($default_theme, $admin_theme));
+  theme_disable(array('stark'));
+  config('system.core')
```
jenlampton: same thing here...  especially since we have comments interspersed.
```php
// Set the Basis theme as the default theme.
// @see system_install()
$config->set('theme_default', $default_theme);
// Set the Seven theme as the admin theme.
$config->set('admin_theme', $admin_theme);
$config->set('admin_theme_same_as_default', 0);
$config->set('node_admin_theme', 1);
```

See my note about the save, below :)

core/profiles/standard/standard.install
```diff
@@ -440,17 +434,25 @@ function standard_install() {
...
+    // Set the Basis theme as the default theme.
+    // @see system_install()
+    ->set('theme_default', $default_theme)
+    // Set the Seven theme as the admin theme.
+    ->set('admin_theme', $admin_theme)
+    ->set('admin_theme_same_as_default', 0)
+    ->set('node_admin_theme', 1)
+    ->save();
```
jenlampton: We don't need this save here, since we save on line 457.

core/profiles/standard/standard.install
```diff
@@ -440,17 +434,25 @@ function standard_install() {
...
 
   // Enable the page cache and aggregation.
-  config_set('system.core', 'preprocess_css', '1');
-  config_set('system.core', 'preprocess_js', '1');
-  config_set('system.core', 'cache', '1');
+  config('system.core')
```
jenlampton: Since these are also variables in the same config file, we don't need to load the file again.
```php
$config->set('preprocess_css', 1);
$config->set('preprocess_js', 1);
$config->set('cache', 1);

$config->save();
```

jenlampton (commented):
Some minor code style feedback :)

core/modules/simpletest/tests/batch.test
klonos: Done 👍

core/modules/simpletest/tests/menu.test
klonos: ...done 👍

core/modules/system/system.admin.inc
klonos: ...done 👍

core/modules/taxonomy/tests/taxonomy.test
klonos: ...and done 🙂

core/profiles/standard/standard.install
klonos: ...oooh, there's more 😅

core/profiles/standard/standard.install
klonos: ...done 👍

core/profiles/standard/standard.install
klonos: ...and finally done 🙂

core/modules/system/system.install
```diff
@@ -3123,6 +3123,25 @@ function system_update_1072() {
...
+  // Only set a value if not already set.
+  if ($config->get('admin_theme_same_as_default' == NULL)) {
+    // If admin_theme is set to either the numeric 0 or the string "0".
+    if ($config->get('admin_theme') == FALSE) {
+      $config->set('admin_theme_same_as_default', 1);
+    }
+    else {
+      $config->set('admin_theme_same_as_default', 0);
+    }
+
```
jenlampton, line 3140: extra whitespace?

klonos:
Replaced by https://github.com/backdrop/backdrop/pull/3199

PR #3179, opened 2020-06-30
Issue #2306: Make the field-as-block feature more discoverable

### LINKED ISSUE #2306: Figure a way to make the field-as-block feature more discoverable.
None of the layouts that we ship with core has `node/%` as path, thus people trying to add a field as a block won't be able to figure it out unless they have good knowledge of the internals of layout paths/contexts. This is something that has made me scratch my head more than once and it honestly is a shame to have in-built such an amazing feature, but not expose it prominently or at least in a way that's easy to figure out.

Possible solutions:

1. We could ship core with another layout that has its path set to `node/%` and includes a field block. A good candidate would be a "Post" layout that has the visibility set to only post nodes and the main content is a display without the image field so we can have that as a field block placed in a sidebar region.

2. We could show all blocks in the block selection dialog and have the ones that are not available be in a disabled state (grayed out) with a note in the form of "This block is available only in layouts that have the *Node* context. For example their path is set to `node/%`". That should be enough to give people a hint as to why the blocks they are looking for are not available and also as to what they need to do.

3. ???

klonos:
Fixes https://github.com/backdrop/backdrop-issues/issues/2306

Replaces https://github.com/backdrop/backdrop/pull/1854


core/modules/layout/layout.admin.inc
```diff
@@ -1051,12 +1044,50 @@ function layout_block_add_page(Layout $layout, $renderer_name, $region_name, $bl
...
+      }
+
+      $link = $block_info['info'];
+      $requirements_class = 'missing-required-contexts';
+
+      if (count($required_contexts) == 1) {
+        $required_context = reset($required_contexts);
+        $context_info = layout_get_context_info($required_context);
+        // Grab the first value in the menu paths, to use as an example.
+        $example_path = $context_info['menu paths'][0];
```
klonos: This is addressing @quicksketch's comment from https://github.com/backdrop/backdrop/pull/1854#discussion_r114206079

The contexts returned now are in this form (as returned by `layout_get_context_info()`):

```
node/%node
user/%user
taxonomy/term/%term
```

...whereas @jenlampton had previously hard-coded them to:

```
node/%
user/%
taxonomy/term/%
```

PR #3183, opened 2020-07-07
Issue #4470: Clarify reason that the server does not support automatic updates.

mikemccaffrey:


gitressa:
Thanks! Just a tip: You can link an issue with its PR by adding `Fixes *ISSUE_URL*` in the first PR comment. So if you update the syntax in this PR, your issue will probably also be linked, where it now says "No branches or pull requests".

izmeez:
@mikemccaffrey As noted in previous comment it must be the first comment that is edited to include `Fixes [issue_url]`. This will connect the PR to the issue. The suggested change of text is good. Thanks.

izmeez:
A minor suggestion: change `you should` to `you can manually ...`

PR #3192, opened 2020-07-20
Issue #4480: Display layout names properly in link autocomplete dropdowns.

### LINKED ISSUE #4480: [UX] Link autocomplete suggestions: Show human-readable names instead of machine names for layouts
This is a follow-up of #3764

image

It should be "Home (Layout: _home_)" instead

ghost:
Fixes https://github.com/backdrop/backdrop-issues/issues/4480


core/modules/system/system.module
```diff
@@ -4278,7 +4278,12 @@ function path_autocomplete($string = '') {
         if ($match_count < $range) {
           $layout_title = '';
           $found_match = FALSE;
-          if (!empty($menu_item['menu']['title']) && stripos($menu_item['menu']['title'], $string) !== FALSE) {
+          $layout = layout_load($menu_item['name'], TRUE);
+          if ($layout && $layout->title) {
```
jenlampton, line 4282: In more modern versions of PHP these non-specific evaluators may cause problems. These should probably be 

`if ($layout != FALSE && !empty($layout->title)) {`

PR #3194, opened 2020-07-22
Issue #4404: syntax highlighting on github for additional file ext

### LINKED ISSUE #4404: Get all our php files showing up with syntax highlighting here on Github
**Description of the need**

Syntax highlighting is a nice feature when viewing code, but per default Github doesn't show some of Backdrop's typical php file extensions highlighted (.module, .install, .test).

**Proposed solution**

It's easy to tell the Github highlighter, which files are also supposed to be php with a **.gitattributes** file.
See an example in this [contrib project](https://github.com/backdrop-contrib/i18n/blob/1.x-1.x/.gitattributes)

Having syntax highlighting for all our php files here on Github is a nice-to-have thing, but not really essential.

What do people think?

indigoxela:
Fixes https://github.com/backdrop/backdrop-issues/issues/4404

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,4 @@
+# get php syntax highlighting for these file extensions
+*.module linguist-language=php
+*.install linguist-language=php
+*.test linguist-language=php
--- a/.gitattributes
+++ b/.gitattributes
@@ -2,3 +2,4 @@
 *.module linguist-language=php
 *.install linguist-language=php
 *.test linguist-language=php
+*.profile linguist-language=php


.gitattributes
```diff
@@ -0,0 +1,4 @@
+# get php syntax highlighting for these file extensions
+*.module linguist-language=php
+*.install linguist-language=php
+*.test linguist-language=php
```
klonos, line 4: Let's please add .profile to the list:

```suggestion
*.test linguist-language=php
*.profile linguist-language=php
```

klonos, line 4:
```suggestion
*.test linguist-language=php
*.profile linguist-language=php
```

indigoxela: Yep, nice catch, let's also add .profile to the list. :+1:

quicksketch:

PR #3195, opened 2020-07-22
Issue #4492: CKEditor settings not respecting lang

### LINKED ISSUE #4492: CKEditor settings not respecting language.
**Description of the bug**

The code is the description :wink: 

https://github.com/backdrop/backdrop/blob/c98bf5148875a4e829754d30afd940d79c8ee87f/core/modules/ckeditor/ckeditor.module#L520

```
 global $language;
...
...
'language' => isset($language->language) ? $language->language : '',
```

Anyone else having a deja-vu here? That should be:

1. `$language->langcode`
2. No need for isset() - that's a global. (OK, it has no _language_ property, it's *langcode*)

What's the problem for users, if the editor's ui lang isn't set correctly?

1. The tooltips on editor toolbar don't show in the right language (en is the fallback)
2. Web browsers also depend on that lang="XX" and possibly mark everything as misspelled inside the editor

I did a recursive grep through core and found several more instances of `$language->language`.
Every here and there there's another _language vs. langcode_ issue - shouldn't we fix this once and for all throughout core?

indigoxela:
Fixes https://github.com/backdrop/backdrop-issues/issues/4492

Uses the correct property of the global $language object. No isset needed.

### DIFF
--- a/core/modules/ckeditor/ckeditor.module
+++ b/core/modules/ckeditor/ckeditor.module
@@ -517,7 +517,7 @@ function ckeditor_get_settings($format, $existing_settings) {
     'coreStyles_strike' => array('element' => 'del'),
     'coreStyles_underline' => array('element' => 'span', 'attributes' => array('class' => 'underline')),
     //'format_tags' => implode(';', $format->editor_settings['format_list']),
-    'language' => isset($language->language) ? $language->language : '',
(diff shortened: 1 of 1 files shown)


quicksketch:
Thanks @indigoxela! Merged into 1.x and 1.16.x.

PR #3197, opened 2020-07-22
Issue #4457: Move error page settings to 'Logging and errors' section…

### LINKED ISSUE #4457: [UX] Move the 403/404 path settings to a better suited place.
It feels that these settings have been "thrown" in that page for lack of a better place...

I feel that somewhere under "URL handling" may be a better place.

---
PR that moves error settings into `URL redirect` section: https://github.com/backdrop/backdrop/pull/3187
PR that moves error settings into `Error handling` section: https://github.com/backdrop/backdrop/pull/3197

(Triaged as possible accessibility issue on 08/17/2921. This one only came up because a possible solution might be bad for accessibility, please add the [A11Y] tag if you think it qualifies.)

jenlampton:
Fixes https://github.com/backdrop/backdrop-issues/issues/4457


core/modules/system/system.admin.inc
```diff
@@ -1610,19 +1588,30 @@ function system_run_cron_submit($form, &$form_state) {
  * @see system_logging_settings_validate()
  */
 function system_logging_settings($form, &$form_state) {
+  $site_config = config('system.core');
   $form['#config'] = 'system.core';
-  $form['error_level'] = array(
+
+  $form['error_display'] = array(
+    '#type' => 'fieldset',
+    '#title' => t('Errors displayed'),
```
klonos: Perhaps change this to "Error messages to display" and remove the title from the radios contained in this fieldset?

core/modules/system/system.module
```diff
@@ -820,9 +820,16 @@ function system_menu() {
     'file' => 'system.admin.inc',
     'weight' => -20,
   );
+  // @deprecated: Remove backwards-compatibility redirect in 2.0.
```
klonos, line 823: 👍

core/modules/system/system.module
```diff
@@ -820,9 +820,16 @@ function system_menu() {
...
   $items['admin/config/development/logging'] = array(
-    'title' => 'Logging and errors',
-    'description' => "Settings for logging and alerts modules. Various modules can route Backdrop's system events to different destinations, such as syslog, database, email, etc.",
+    'page callback' => 'backdrop_goto_deprecated',
+    'page arguments' => array('admin/config/development/errors'),
```
klonos, line 826: 👍

core/modules/system/system.module
```diff
@@ -928,7 +935,7 @@ function system_menu() {
   );
   $items['admin/config/system/site-information'] = array(
     'title' => 'Site information',
-    'description' => 'Change site name, logo, favicon, slogan, e-mail address, default front page, and number of posts per page, error pages.',
+    'description' => 'Change site name, logo, favicon, slogan, and from e-mail address.',
```
klonos: The home page setting can still be configured here + while at it, let's reorder these to reflect their order in the form.

```suggestion
    'description' => 'Change site name and slogan, from e-mail address, logo, favicon, and default home page.',
```

core/modules/system/system.admin.inc
jenlampton: I don't think it's possible to have a form element without a title (?) but even if it were, it would be bad for accessibility. Can you think of another fieldset label that might be better, or should I still try this suggestion?

klonos: 🤔 ...I actually felt the same way, and was to do some research re what would be an acceptable solution (I was in need to do the same for the 404/403 path fields in the PR I've filed against yours) ...but then got distracted and never did that.

klonos: ...oh joy! ...turns out you can hide element titles in a way that still retains accessibility, by using `'#title_display' => 'invisible'` 😀 ...I've seen that in the past somewhere; just never had the need for it till now.

core/modules/system/system.admin.inc
```diff
@@ -1610,19 +1588,30 @@ function system_run_cron_submit($form, &$form_state) {
...
+  );
+  $form['error_display']['error_level'] = array(
     '#type' => 'radios',
     '#title' => t('Error messages to display'),
```
klonos, line 1600:
```suggestion
    '#title' => t('Error messages to display'),
    '#title_display' => 'invisible',
```

core/modules/system/system.admin.inc
klonos: So change this to:

```suggestion
    '#title' => t('Error messages to display'),
```

core/modules/system/system.admin.inc
klonos, line 1600: ...and then also do this:

```suggestion
    '#title' => t('Error messages to display'),
    '#title_display' => 'invisible',
```

PR #3199, opened 2020-07-26
Issue 4461: [UX] Remove the special 'Default theme' option

### LINKED ISSUE #4461: [UX][DX] Remove the option for configuring the admin theme to always be the same as the default theme.

klonos:
Fixes https://github.com/backdrop/backdrop-issues/issues/4461

Replaces https://github.com/backdrop/backdrop/pull/3178


quicksketch:
Thanks @klonos! Merged into 1.x for 1.20.0.

PR #3203, opened 2020-07-29
Issue #4504: Add help text in the roles management page

klonos:
UPDATE: This was for https://github.com/backdrop/backdrop-issues/issues/4504 but should probably be on a new follow-up issue.


core/modules/user/css/user.css
```diff
@@ -14,26 +14,17 @@
...
-  float: left; /* LTR */
-  margin-right: 1em; /* LTR */
-}
-[dir="rtl"] #user-admin-roles .form-item-name {
-  float: right;
-  margin-left: 1em;
-  margin-right: 0;
+#user-roles .description {
+  margin-left: 2em;
 }
```
klonos: This entire section of CSS is moot, as `theme_user_admin_roles()` as it is currently implemented renders the text field and the "Add role" button into separate cells/columns in the table. In D7, both the text field and the button were rendered in the same table cell, with `colspan="4"`. In Backdrop, the "Add role" button was moved in the "Operations" column though.

klonos: Filed a follow-up: https://github.com/backdrop/backdrop-issues/issues/4506

core/modules/user/user.admin.inc
```diff
@@ -552,7 +562,8 @@ function user_admin_roles($form, $form_state) {
         ),
       );
     }
-    if ($role_name !== BACKDROP_ANONYMOUS_ROLE && $role_name !== BACKDROP_AUTHENTICATED_ROLE) {
```
jenlampton, line 580: looks like there's an accidental change here. This is the kind of thing that could introduce errors, so it's best not to change anything other than what the issue says you're changing.

core/modules/user/user.admin.inc
```diff
@@ -606,7 +616,7 @@ function user_admin_roles($form, $form_state) {
  * Form submit function. Update the role weights.
  */
 function user_admin_roles_order_submit($form, &$form_state) {
-  // Save admin role and anon name.
```
jenlampton, line 609: why fix `anon` but not `admin`?

core/modules/user/user.admin.inc
```diff
@@ -628,6 +638,8 @@ function user_admin_roles_add_submit($form, &$form_state) {
   $role = (object) array(
     'name' => $form_state['values']['name'],
     'label' => $form_state['values']['label'],
+    // Add new roles directly after the authenticated role.
+    'weight' => $form_state['values']['roles'][BACKDROP_AUTHENTICATED_ROLE]['weight'] + 1,
```
jenlampton: I don't think this will work reliably unless we increment all the other roles by 1 also. I'll test.

core/modules/user/user.theme.inc
```diff
@@ -194,7 +204,9 @@ function theme_user_admin_roles($variables) {
 
   backdrop_add_tabledrag('user-roles', 'order', 'sibling', 'role-weight');
 
-  $output = theme('table', array('header' => $header, 'rows' => $rows, 'attributes' => array('id' => 'user-roles')));
+  $output = backdrop_render($form['roles_help_before']);
+  $output .= theme('table', array('header' => $header, 'rows' => $rows, 'attributes' => array('id' => 'user-roles')));
+  $output .= backdrop_render($form['roles_help_after']);
```
jenlampton: As long as we get the correct `#weight` for this element, we don't need to add it separately. The only reason the `before` help was separate is because the table goes between it and the rest of the form.

core/modules/user/user.admin.inc
jenlampton: Yeah, the current weighting behaves strangely. To the untrained eye, new roles are arbitrarily placed somewhere between `Anonymous` and `Administrator`.

core/modules/user/user.theme.inc
```diff
@@ -177,15 +177,25 @@ function theme_user_admin_roles($variables) {
     $label = $form['roles'][$role_name]['#role']->label;
     $row = array();
     if (in_array($role_name, array(BACKDROP_ANONYMOUS_ROLE, BACKDROP_AUTHENTICATED_ROLE))) {
-      $row[] = t('@name <em>(required)</em>', array('@name' => $label));
```
jenlampton, line 180: I think this change is a step backwards. 
We've removed something that was short, easy to understand, and consistent with the rest of core, and added something foreign and confusing. Let's put this back unless we can agree on something that's better than what we had before.

core/modules/user/user.admin.inc
klonos: I was trying to be consistent with https://github.com/backdrop/backdrop/pull/3203/files#diff-2b236f6288c8e14c97892a1e54258197R179 ...happy to revert.

core/modules/user/user.admin.inc
klonos: You're right! 😅 ...will fix 👍

core/modules/user/user.admin.inc
klonos: I know 🤔 ...this is why I said "I've implemented this in a simple way..." in the issue comment. Perhaps this will need to be a follow up if the logic is too complicated.

core/modules/user/user.theme.inc
klonos: Not married to this change, but for argument's sake, this was my rationale:

>Now, as a curious user, if I click the drop-button on the first role in the list, I see no delete action available. If I try the next role, it's the same ...so do I keep going, or assume that none of the roles can be deleted? We (experienced Drupal/Backdrop users) may know what/why, but we shouldn't be assuming that it is common knowledge that the anonymous and authenticated roles cannot be deleted, whereas the rest can. To that effect, I have changed the vague "required" note next to these special-cased roles with "system role - cannot be deleted", which I think helps this situation (the respective D7 text for this was "locked").

core/modules/user/user.theme.inc
klonos: I may need some help with this. Before I even implemented things this way, I've tried setting the weight of the help text to something insane, like `'#weight' => -99999999,`, but the highest I could get it to render in the page was before the "Administrator role" fieldset 😓

core/modules/user/user.admin.inc
klonos: ...I'll revert this. I've filed a separate follow-up here: #https://github.com/backdrop/backdrop-issues/issues/4507

klonos: ...reverted 👍

core/modules/user/user.admin.inc
klonos: Done! 👍

core/modules/user/user.theme.inc
jenlampton: Yeah, I read that comment when you posted it. I still think this new text isn't as good as the vague "required" that was there before. 

If someone adds a role and wants to delete it later, they'll be able to. I don't think there is a problem that they can't delete the required roles -- I mean, **it says right there that they are required.**  But `system role`? That doesn't mean anything, it's just more arbitrary words. And `cannot be deleted` is already covered by `Required`. I see this change as a more verbose and more confusing version of what was there before.

core/modules/user/user.theme.inc
jenlampton: > but the highest I could get it to render in the page was before the "Administrator role" fieldset

That's where we want it, isn't it? (That's where it is now)

Screen Shot 2020-07-29 at 5 33 51 PM

Can you delete `$output .= backdrop_render($form['roles_help_after']);` and see what happens?

klonos: Oh, I see what you men now; I misread your initial comment and thought that you were suggesting to remove the `backdrop_render()` for `roles_help_before` as well.

Fixed 👍

core/modules/user/user.theme.inc
klonos: I've reverted back to "required", but have removed the name placeholder and html tags from the translatable string - no need for that to be in the face of translators 😉

core/modules/user/user.admin.inc
```diff
@@ -500,6 +500,16 @@ function user_admin_permissions_submit($form, &$form_state) {
 function user_admin_roles($form, $form_state) {
   $system_config = config('system.core');
 
+  $form['roles_help_before'] = array(
+    '#type' => 'help',
+    '#markup' => t('Roles provide a way to organize people into groups that have similar duties. Each user account can be given one or more roles, and <a href="@permissions">permissions</a> can be granted to those roles (instead of individually to each user account).', array('@permissions' => url('admin/config/people/permissions'))),
+  );
+  $form['roles_help_after'] = array(
+    '#type' => 'help',
+    '#markup' => t('To make the permissions page easier to navigate, we recommended ordering roles from the least permissive (Anonymous) to the most permissive (usually Administrator).'),
```
quicksketch, line 509: Help text usually speaks declarative (I think that's the right term?) rather than in first person.

```suggestion
    '#markup' => t('To make the permissions page easier to navigate, it is recommended to order roles from the least permissive (Anonymous) to the most permissive (usually Administrator).'),
```

quicksketch:
Ah I see this is replaced by https://github.com/backdrop/backdrop/pull/3888. Closing in favor of that PR.

PR #3207, opened 2020-07-31
Issue #4425: Allow selecting an existing date format for log messages.

### LINKED ISSUE #4425: Allow selecting an existing date format for log messages
**Description of the need**

The "Date format for log messages" field at `/admin/config/development/logging` requires entry of PHP codes, but we have existing date/time formats at `/admin/config/regional/date-time` we could use instead.

image

**Proposed solution**

Change the "Date format for log messages" text field to a select list where you can choose an existing date format instead. This would provide more consistency without requiring people to enter their preferred date format in two seperate places.

ghost:
Fixes https://github.com/backdrop/backdrop-issues/issues/4425

### DIFF
--- a/core/modules/dblog/dblog.install
+++ b/core/modules/dblog/dblog.install
@@ -107,3 +107,33 @@ function dblog_update_1001() {
   $short = system_date_format_load('short');
   config_set('system.core', 'log_date_format', $short['pattern']);
 }
+
+/**
+ * Convert the format for dates in the database log.
+ */
+function dblog_update_1002() {
+  $t = get_t();
+  $config = config('system.core');
+  $log_format = $config->get('log_date_format');
+
+  // Look for a matching date format to use.
+  foreach (system_get_date_formats() as $date_format) {
+    if ($log_format == $date_format['pattern']) {
+      $config->set('log_date_format', $date_format['name']);
+      $config->save();
+      break;
+    }
+  }
+
+  // No matching date format found, so create a custom one.
+  if ($config->get('log_date_format') == $log_format) {
+    system_date_format_save(array(
+      'name' => 'dblog_date_format',
+      'label' => $t('Database log messages date'),
+      'pattern' => $log_format,
+      'module' => 'dblog',
+    ));
+    $config->set('log_date_format', 'dblog_date_format');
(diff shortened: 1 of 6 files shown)


core/modules/dblog/dblog.install
```diff
@@ -107,3 +107,33 @@ function dblog_update_1001() {
...
+  foreach (system_get_date_formats() as $date_format) {
+    if ($log_format == $date_format['pattern']) {
+      $config->set('log_date_format', $date_format['name']);
+      $config->save();
+      break;
+    }
+  }
+
+  // No matching date format found, so create a custom one.
+  if ($config->get('log_date_format') == $log_format) {
```
klonos, line 129: Not sure this conditional is required 🤔 ...we're setting `$log_format = $config->get('log_date_format');` earlier, and we also `break;` from the `foreach` if `$log_format` matches an existing pattern. I think that there's no point - we should directly do the `system_date_format_save()`.

ghost: This is needed. We first look for a matching date format and, if found, use it. Then we're done, nothing more needed. If, after going through all existing date formats, a match isn't found (i.e. log_date_format in config still equals the old pattern value) then we need to create and save a new date format, then save its name in config.

Does that explain better?

klonos: Yeah, sorry, I confused `break` with `return` here. Ignore me.

ghost: Ah yes, a `return;` would simplify things, but when I looked into the `hook_update_N()` function, it said returns will be printed to the user as a message, so I thought it better not to do a return then.

ghost:
Merged into 1.x and 1.17.x.

jenlampton:
@BWPanda Is there a policy about core committers not merging their own PRs? I see @quicksketch mention that a lot but I don't know if it's official policy or just best practice.

ghost:
@jenlampton I'm not sure... I know that I don't merge PRs if I'm the one who's marked it RTBC. Maybe it's the same for own PRs, don't know. Let me know if you find anything documented about it...

jenlampton:
I don't think we have any documentation :)

PR #3209, opened 2020-08-03
PoC: Detect "forked" core themes

klonos:
https://github.com/backdrop/backdrop-issues/issues/4167#issuecomment-667767923


jenlampton:
@klonos what does this accomplish? If someone copies a core theme out of core, and then when core is updated, the site indicates that it also needs to be updated -- what was the point of copying it out of core?

PR #3218, opened 2020-08-07
Issue #3730: Add NID class to node blocks.

### LINKED ISSUE #3730: Add NID to node block class
**Description of the need**
When displaying a node in a block, there's no way to differentiate one node block from another based on the block classes. I was able to fix this by adding the following to my theme's `template.php` file, but I think something similar in core would be better:

```
/**
 * Prepares variables for block templates.
 */
function THEME_preprocess_block(&$variables) {
  if ($variables['theme_hook_original'] == 'block__node__content') {
    $nid = $variables['block']->settings['nid'];
    $variables['classes'][] = "block-node-$nid";
  }
}
```

ghost:
Fixes https://github.com/backdrop/backdrop-issues/issues/3730

### DIFF
--- a/core/modules/layout/layout.theme.inc
+++ b/core/modules/layout/layout.theme.inc
@@ -459,6 +459,11 @@ function template_preprocess_block(&$variables) {
     }
   }
 
+  // Add a NID class to node blocks.
+  if ($block->plugin == 'node:content') {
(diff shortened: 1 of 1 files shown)


ghost:
Merged into 1.x.

PR #3219, opened 2020-08-07
Issue #3708: Add current character count to text length error message.

### LINKED ISSUE #3708: Improve "The text may not be longer than..." error message
**Description of the need**
When entering too much text into a field with a maximum length limit, you get an error stating "The text may not be longer than [N] characters." I'm guessing most users wouldn't know how (and certainly can't be expected) to copy/paste the text into an editor to see the character count, especially when HTML tags are counted as part of the limit.

**Proposed solution**
I think it'd be handy to add to the error message the actual length of the text entered, so the user doesn't have to guess how much they have to delete (e.g. "The text may not be longer than [N] characters. The text is currently [X] characters long." It'll at least give them an idea if they're just over the limit, double the limit, etc.

Alternatively, or in addition to, we could offer some sort of a confirmation to proceed regardless, stating that doing so will mean the text will be truncated to the character limit.

ghost:
Fixes https://github.com/backdrop/backdrop-issues/issues/3708


bugfolder:
Closing in favor of https://github.com/backdrop/backdrop/pull/4559.

PR #3222, opened 2020-08-07
Issue #4526: Prevent Fatal when GD cannot determine the size of an im…

### LINKED ISSUE #4526: Saving an Image field with an image that has undeterminable dimensions throws PDOException
**Description of the bug**

I see a lot of issues about this error in various contrib modules, but it is not exclusively a contrib problem and can happen with the core integer field as well as with contrib field types such as location field and node reference that store integer values in their respective field_data_field tables. To reproduce the problem, create an integer field (or location or node reference or...) with no default value, then create content without specifying a value for that field. The error is as follows:

PDOException: SQLSTATE[HY000]: General error: 1366 Incorrect integer value: '' for column 'field_[field name here]' at row 1: INSERT INTO {field_data_field_[field name here]} (entity_type, entity_id, revision_id, bundle, delta, language, field_[field name]_value) VALUES (:db_insert_placeholder_0, :db_insert_placeholder_1, :db_insert_placeholder_2, :db_insert_placeholder_3, :db_insert_placeholder_4, :db_insert_placeholder_5, :db_insert_placeholder_6); Array ( [:db_insert_placeholder_0] => node [:db_insert_placeholder_1] => 5060 [:db_insert_placeholder_2] => 5310 [:db_insert_placeholder_3] => directory [:db_insert_placeholder_4] => 0 [:db_insert_placeholder_5] => und [:db_insert_placeholder_6] => ) in field_sql_storage_field_storage_write() (line 451 of /home/[account]/public_html/drupal-7.9/modules/field/modules/field_sql_storage/field_sql_storage.module).

This is one we inherited from Drupal.
See [This Drupal 8 issue](https://www.drupal.org/project/drupal/issues/1330952) and [This drupal 7 issue](https://www.drupal.org/project/drupal/issues/1546024), that got marked as dupe.

**Proposed resolution**

Check that GD is installed.

Placeholder 6 is not an integer and is not NULL. There is some code in the function (lines 433-435 of field_sql_storage.module) that attempts to handle this:
```php
  foreach ($field['columns'] as $column => $attributes) {
     $record[_field_sql_storage_columnname($field_name, $column)] = isset($item[$column]) ? $item[$column] : NULL;
  }
```
but for some reason this is not sufficient to convert these non-values to NULL before sending them to the database. I replaced this with the following:
```php
  foreach ($field['columns'] as $column => $attributes) {
     //convert non-numeric values (such as empty strings) in integer fields to NULL
     if ($attributes['type'] == 'int' && !$attributes['not null'] && !is_numeric($item[$column])) {
       $item[$column] = NULL;
     }
     $record[_field_sql_storage_columnname($field_name, $column)] = isset($item[$column]) ? $item[$column] : NULL;
  }
```
and that has resolved the problem. I am terrible at rolling patches, so I'd appreciate it if someone who is more competent in that regard could review this fix and turn it into a proper patch. Thank you in advance.

**Steps To Reproduce**

1. Download the file [Problem.jpg](https://www.drupal.org/files/Problem.jpg) (this is critical because only certain images generate this error)
2. Log in to your fresh install, click the "Add Content" link, and create a new Article
3. Enter a title and then browse for the image you just downloaded
4. Save the node (notice you don't see the image when you view the node)
5. Click the node's Edit tab (notice you see the file name/icon but again no image thumbnail)
6. Save the node and you'll see the error

jenlampton:
Fixes https://github.com/backdrop/backdrop-issues/issues/4526

### DIFF
--- a/core/modules/image/image.field.inc
+++ b/core/modules/image/image.field.inc
@@ -315,12 +315,16 @@ function image_field_presave($entity_type, $entity, $field, $instance, $langcode
 
   // Determine the dimensions if necessary.
   foreach ($items as &$item) {
-    if (!isset($item['width']) || !isset($item['height'])) {
+    if (!isset($item['width']) || !isset($item['height']) || empty($item['width']) || empty($item['height'])) {
       $info = image_get_info(file_load($item['fid'])->uri);
 
       if (is_array($info)) {
         $item['width'] = $info['width'];
         $item['height'] = $info['height'];
+      }
+      else {
+        // Width and height need to be NULL when image_get_info() fails.
+        $item['width'] = $item['height'] = NULL;
       }
     }
   }


core/modules/image/image.field.inc
```diff
@@ -315,13 +315,17 @@ function image_field_presave($entity_type, $entity, $field, $instance, $langcode
 
   // Determine the dimensions if necessary.
   foreach ($items as &$item) {
-    if (!isset($item['width']) || !isset($item['height'])) {
+    if (!isset($item['width']) || !isset($item['height']) || empty($item['width']) || empty($item['height'])) {
```
indigoxela, line 318: No need to use both, isset() and empty() here. Using empty() is sufficient.

core/modules/image/image.field.inc
```diff
@@ -315,13 +315,17 @@ function image_field_presave($entity_type, $entity, $field, $instance, $langcode
...
       $info = image_get_info(file_load($item['fid'])->uri);
 
       if (is_array($info)) {
         $item['width'] = $info['width'];
         $item['height'] = $info['height'];
       }
+      else {
```
indigoxela, line 325: At this point it is clear that this file is severely broken. It might be smarter to not handle it as image anymore.

PR #3223, opened 2020-08-08
Issue #3419: Allow 'Content: Link' views field to just output URL.

### LINKED ISSUE #3419: [UX][D8] Allow 'Content: Link' views field to just output URL
Often when creating a view that requires a URL to the content (e.g. for adding to the 'Output this field as a link' setting on other fields), I'll add the 'Content: Link' field, forgetting that this is meant to display an actual link (e.g. 'View more') as opposed to outputting just the URL/path that I can use elsewhere. I need to add the separate 'Content: Path' field instead.

It'd be nice if we merged these two very similar fields into one, by adding an option to the 'Content: Link' field for outputting just the URL (and therefore hiding the 'Text to display' setting).

Drupal 8 does this nicely:
screenshot_2018-12-11

- - - - - - - 

PR by @BWPanda: https://github.com/backdrop/backdrop/pull/3223

ghost:
Fixes https://github.com/backdrop/backdrop-issues/issues/3419


core/modules/node/views/views_handler_field_node_link.inc
```diff
@@ -14,21 +14,41 @@ class views_handler_field_node_link extends views_handler_field_entity {
...
     $options['text'] = array('default' => '', 'translatable' => TRUE);
     return $options;
   }
 
   function options_form(&$form, &$form_state) {
+    $form['url'] = array(
+      '#type' => 'checkbox',
+      '#title' => t('Output URL as text'),
+      '#default_value' => $this->options['url'],
+      '#description' => t('Enable this option to output just the URL (i.e. not a clickable link).'),
```
jenlampton: ~This text could be shortened to "Output only the URL, not a clickable link."~

I think we should remove the description entirely, and change the title to `Output the URL as text, not a clickable link`

I like to avoid using text like "Enable this option to" (because it's not necessary) and also "i.e."(because it makes the software feel slightly less friendly). I know these are all over Drupal so we still inherit a lot from core and conrib :)

core/modules/node/views/views_handler_field_node_link.inc
```diff
@@ -14,21 +14,41 @@ class views_handler_field_node_link extends views_handler_field_entity {
...
+        'visible' => array(
+          ':input[name="options[url]"]' => array('checked' => FALSE),
+        ),
+      ),
+    );
+    $form['absolute'] = array(
+      '#type' => 'checkbox',
+      '#title' => t('Use absolute link (begins with "http(s)://")'),
+      '#default_value' => $this->options['absolute'],
+      '#description' => t('Enable this option to output an absolute link.'),
```
jenlampton: ~Here, the title and the description seem to be switched. Title should be "Use an absolute link." and the description should be "An absolute link begins with "https://"~

Let's remove the description here, it doesn't add any value. 

The title could be improved as follows:
`Use an absolute link beginning with "http(s)://"`

ghost: Changed title based on your feedback, slightly different wording though.

core/modules/node/views/views_handler_field_node_link.inc
ghost: @jenlampton I felt title would be too long, so made some other changes. Please review.

quicksketch:
This is great! I merged into 1.x for 1.27.0. Thanks @BWPanda!

PR #3224, opened 2020-08-09
Issue #4512: Initial proof of concept on adding supplemental stylesheets to Basis

stpaultim:
I'm not sure if this is a good idea yet, but wanted to get something started so that I can get constructive feedback.

### DIFF
--- a/core/themes/basis/theme-settings.php
+++ b/core/themes/basis/theme-settings.php
@@ -75,3 +75,18 @@ else {
     '#markup' => '<p>' . t('This theme supports custom color palettes if the Color module is enabled on the <a href="!url">modules page</a>. Enable the Color module to customize this theme.', array('!url' => url('admin/modules'))) . '</p>',
   );
 }
+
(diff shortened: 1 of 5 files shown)


core/themes/basis/theme-settings.php
```diff
@@ -75,3 +75,18 @@
...
 }
+
+function basis_form_system_theme_settings_alter(&$form, &$form_state) {
+  $form['supplemental'] = array(
+    '#type'      => 'fieldset',
+    '#title'     => t('CSS Updates'),
+    '#collapsible' => TRUE,
+    '#collapsed' => TRUE,
+  );
+  $form['supplemental']['basis_supplemental_css'] = [
```
indigoxela: `[]` isn't optimal because Backdrop still tries to be backwards compatible. Use `array()` instead - like above.
That should fix the failing tests on php 5.3.

stpaultim: Thanks for this info. I'm a php hack and copied code from different sources and this was the result.

PR #3225, opened 2020-08-09
Issue #4527: Image browser broken without filter fields

### LINKED ISSUE #4527: Image browser does not load correctly when content type has not WYSIWYG field
**Description of the bug**

Whenever I have content types without any HTML formatte field, the image browser does not load correctly.

**Steps To Reproduce**

To reproduce the behavior:
- Create a new content type
- Delete the body field
- add an image field
- optional: add some non-WYSIWYG fields
- create the node and try to choose an existing image ("Select from Image Library")

Screenshot 2020-08-08 at 11 30 27

- Backdrop Version 1.16.2 
- various browsers on OSX

As soon as any WISYWIG field is added, browser works correctly, shows images in matrix

Screenshot 2020-08-08 at 11 35 28

indigoxela:
Fixes https://github.com/backdrop/backdrop-issues/issues/4527

Attach libraries required for functional image browser.

### DIFF
--- a/core/modules/file/file.module
+++ b/core/modules/file/file.module
@@ -1264,6 +1264,7 @@ function file_admin_paths() {
  */
 function file_managed_file_process($element, &$form_state, $form) {
   $fid = isset($element['#value']['fid']) ? $element['#value']['fid'] : 0;
+  $filter_path = backdrop_get_path('module', 'filter');
 
   // Set some default element properties.
   $element['#progress_indicator'] = empty($element['#progress_indicator']) ? 'none' : $element['#progress_indicator'];
@@ -1316,9 +1317,6 @@ function file_managed_file_process($element, &$form_state, $form) {
   );
 
   // Add a button that will be used to select existing files if enabled.
-  // Attaching the assets from filter module makes sure, the browser works
-  // even if no field on that page has a filter set.
-  $filter_path = backdrop_get_path('module', 'filter');
   $element['browse_button'] = array(
     '#name' => implode('_', $element['#parents']) . '_browse_button',
     '#type' => 'submit',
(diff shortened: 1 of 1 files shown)


core/modules/file/file.module
```diff
@@ -1316,6 +1316,9 @@ function file_managed_file_process($element, &$form_state, $form) {
   );
 
   // Add a button that will be used to select existing files if enabled.
+  // Attaching the assets from filter module makes sure, the browser works
+  // even if no field on that page has a filter set.
```
ghost: I feel like this comment should be moved down to just above the `#attached` line (where it's referring to). Also, remove the comma please.

klonos: The comma feels out of place here. Perhaps it was meant to go after the word "works"(?).

While at it, can we please be more detailed? How about this?:

```suggestion
  // Attaching the assets from the Filter module makes sure that the image
  // browser works, even if no field on that form has a filter set. This can
  // happen on forms that have only image field(s), without any WYSIWYG-enabled
  // text field.
```

core/modules/file/file.module
```diff
@@ -1316,6 +1316,9 @@ function file_managed_file_process($element, &$form_state, $form) {
...
+  $filter_path = backdrop_get_path('module', 'filter');
```
ghost: There's a comment on line 1318 that explains this whole block of code, but then this line sits between the two. Maybe this line can be moved above the comment, or even to the top of the function...?

ghost (changes requested):

core/modules/file/file.module
indigoxela: WYSIWYG-enabled is misleading here. The problem has nothing to do with the editor, but with a filter (text format) applied or not applied.

core/modules/file/file.module
indigoxela: Ehm.. so where should I put it? Above? Next to the `#attached` - I'm confused. :wink:

ghost:
@indigoxela Sorry for the confusion. Your PR adds 7 lines of code. Let's refer to them as letters (A-G respectively).

I'm suggesting that A & B be placed immediately before D. And that C be moved to the top of the function.

Does that make more sense?

ghost (approved):
Looks all good to me!

quicksketch:

PR #3229, opened 2020-08-10
Issue #4252: Replace 'js-only' class with 'js-show'.

### LINKED ISSUE #4252: [DX] Consolidate "js-show" and "js-only" classes
**Description of the bug**
Both views_ui.admin.css and system.css provide a way to show contents on pages only if JS is enabled:

system.css:
```
/**
 * For anything you want to show on page load only when JS is enabled.
 */
.js-show {
  display: none;
}
.js .js-show {
  display: block;
}
```

views_ui.admin.css:
```
.js-only {
  display: none;
}

html.js .js-only {
  display: inherit;
}

html.js span.js-only {
  display: inline;
}
```

We should remove the Views-specific one and update all instances of the "js-only" class to use "js-show".

This came up as I was trying to make the Browse button for the image library hide when JS is disabled in #3297.

**Steps to test:**
* Edit the admin_content view
* Click `Rearrange` for filters.
* Click `+ Create new filter group`
* Confirm that the text `Drag to add filters` appears in the new row

ghost:
Fixes https://github.com/backdrop/backdrop-issues/issues/4252

### DIFF
--- a/core/modules/installer/installer.pages.inc
+++ b/core/modules/installer/installer.pages.inc
@@ -20,7 +20,7 @@ function installer_browser_page($type) {
   // select to view available releases. Install blocked on select versions page.
   installer_browser_check_zip_loaded();
 
-  // Show link to enable page if recently installed modules were not yet 
+  // Show link to enable page if recently installed modules were not yet
   // enabled.
   $recent_projects = installer_browser_get_installed_projects();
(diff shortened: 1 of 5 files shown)


quicksketch:
Merged into 1.x only for 1.18.0. Thanks!

PR #3231, opened 2020-08-11
Issue #1500: Prevent current and user 1 accounts from being blocked.

### LINKED ISSUE #1500: VBO: Do not allow blocking the current user, or user with UID 1.
I think if a user is bulk deleting users and User 1 is selected it should not block the user 1 account.

Steps to reproduce (Backdrop 1.x):
- Loggin as user 1
- Create some users
- visit: `/admin/people/list`
- Select all the users
- Under Operations dropdown select Block User(s)
- Click execute

You are logged out and the DB shows user 1 status = 0;

screen shot 2016-01-09 at 1 29 43 pm

It blocks all the users including you (user 1); I don't think this is good or desired behaviour.

We should add an exception for user 1.

ghost:
Fixes https://github.com/backdrop/backdrop-issues/issues/1500

### DIFF
--- a/core/modules/user/user.actions.inc
+++ b/core/modules/user/user.actions.inc
@@ -16,7 +16,7 @@ function user_block_user_action(User $account, &$context) {
   global $user;
 
   // Don't block User 1, or the user performing the action.
-  if ($account->uid === 1) {
+  if ($account->uid === '1') {
     backdrop_set_message(t("User 1 cannot be blocked."), 'warning');
     return FALSE;
   }
--- a/core/modules/user/user.actions.inc
+++ b/core/modules/user/user.actions.inc
@@ -13,6 +13,18 @@
  * @ingroup actions
  */
 function user_block_user_action(User $account, &$context) {
+  global $user;
+
+  // Don't block User 1, or the user performing the action.
+  if ($account->uid === 1) {
+    backdrop_set_message(t("User 1 cannot be blocked."), 'warning');
+    return FALSE;
+  }
+  elseif ($account->uid == $user->uid) {
+    backdrop_set_message(t("You cannot block your own user account."), 'warning');
+    return FALSE;
+  }
+
   if ($account->status == 1) {
     $account->status = 0;
     $account->save();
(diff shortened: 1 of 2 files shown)


core/modules/user/user.actions.inc
```diff
@@ -13,6 +13,18 @@
  * @ingroup actions
  */
 function user_block_user_action(User $account, &$context) {
+  global $user;
+
+  // Don't block User 1, or the user performing the action.
+  if ($account->uid === 1) {
```
indigoxela: `$account->uid` is actually a string, not an integer

ghost: Fixed, thanks!

core/modules/user/user.actions.inc
```diff
@@ -13,6 +13,18 @@
  * @ingroup actions
  */
 function user_block_user_action(User $account, &$context) {
+  global $user;
+
+  // Don't block User 1, or the user performing the action.
+  if ($account->uid === '1') {
+    backdrop_set_message(t("User 1 cannot be blocked."), 'warning');
```
klonos: https://api.backdropcms.org/php-standards#quotes

For strings within `t()` single quotes are preferred over double:

```suggestion
    backdrop_set_message(t('User 1 cannot be blocked.'), 'warning');
```

While at it, and because we may be showing this to people that do not understand what "user 1" is, may I suggest something like the following instead:

```suggestion
    backdrop_set_message(t('The user account %admin_name (user with ID 1) cannot be blocked.', array('%admin_name' => $account->name)), 'warning');
```

core/modules/user/user.actions.inc
```diff
@@ -13,6 +13,18 @@
...
+    return FALSE;
+  }
+  elseif ($account->uid == $user->uid) {
+    backdrop_set_message(t("You cannot block your own user account."), 'warning');
```
klonos: Same as previous comment, re single/double quotes:

```suggestion
    backdrop_set_message(t('You cannot block your own user account.'), 'warning');
```

klonos (approved):
LGTM 👍

ghost:
Merged into 1.x and 1.16.x.

PR #3237, opened 2020-08-11
Issue #4532: Fix undefined variable in ConfigFileStorage::initializeStorage().

### LINKED ISSUE #4532: Undefined variable in ConfigFileStorage::initializeStorage()
**Description of the bug**

The [`ConfigFileStorage::initializeStorage()`](https://api.backdropcms.org/api/backdrop/core%21includes%21config.inc/function/ConfigFileStorage%3A%3AinitializeStorage/1) has an undefined variable: `$config_directory`.

**Steps To Reproduce**

To reproduce the behavior setup a new Backdrop site with the `$config_directories` variables in `settings.php` set to something ridiculous (like `'../../../../../../../../../config/active'`, i.e. this directory isn't possible on my filesystem). Now try installing Backdrop.

**Actual behavior**

Error:

> Notice: Undefined variable: config_directory in ConfigFileStorage->initializeStorage() (line 1252 of /app/backdrop/core/includes/config.inc).

**Expected behavior**

Just the exception:

> The directory '../../../../../../../../../config/active' could not be created or could not be made writable.

ghost:
Fixes https://github.com/backdrop/backdrop-issues/issues/4532

### DIFF
--- a/core/includes/config.inc
+++ b/core/includes/config.inc
@@ -63,7 +63,7 @@ function config($config_file, $type = 'active') {
  * @param string $option
  *   The name of the config option within the file to delete. The config option
  *   may contain periods to indicate levels within the config file.
- * 
+ *
  * @see config_set()
  * @see config_get()
  *
@@ -1248,8 +1248,8 @@ class ConfigFileStorage implements ConfigStorageInterface {
    */
   public function initializeStorage() {
     if (!file_prepare_directory($this->directory, FILE_CREATE_DIRECTORY | FILE_MODIFY_PERMISSIONS)) {
-      throw new ConfigStorageException(format_string("The directory '%directory' could not be created or could not be made writable.", array(
-        '%directory' => $this->directory,
+      throw new ConfigStorageException(format_string('The directory <code>@directory</code> could not be created or could not be made writable.', array(
+        '@directory' => $this->directory,
       )));
     }
   }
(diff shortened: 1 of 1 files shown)


core/includes/config.inc
```diff
@@ -1248,9 +1248,9 @@ class ConfigFileStorage implements ConfigStorageInterface {
    */
   public function initializeStorage() {
     if (!file_prepare_directory($this->directory, FILE_CREATE_DIRECTORY | FILE_MODIFY_PERMISSIONS)) {
-      throw new ConfigStorageException(format_string('The directory %directory could not be created or could not be made writable. ', array(
-      '%directory' => $config_directory,
-    )));
+      throw new ConfigStorageException(format_string("The directory '%directory' could not be created or could not be made writable.", array(
```
docwilmot: The "%" already displays as `<em>` tags so I think the quotes are unnecessary. Otherwise good.

klonos: We use italics (via `%`) when referring to names of things. When referring to paths and code, we use regular text (`@`) and wrap it around `<code>` tags:

```suggestion
      throw new ConfigStorageException(format_string('The directory <code>@directory</code> could not be created or could not be made writable.', array(
```

...and then the next line should be:

`'@directory' => $this->directory,`

docwilmot: > We use italics (via %) when referring to names of things. When referring to paths and code, we use regular text (@) and wrap it around <code> tags:

Really? Where is that documented?

ghost: I did try just italics first, but it didn't really stand out that much from the rest of the text and so wasn't obvious. I'll use `<code>` instead.

ghost:
Merged into 1.x and 1.16.x.

PR #3238, opened 2020-08-12
Issue #706: Protect config directories with htaccess files.

### LINKED ISSUE #706: Use `.htaccess` files to secure configuration directories
Currently, the `config` directory is 'secured' by appending a hash string to the dir path. This has caused some issues when changing database credentials, etc., and could be improved.

I suggest instead adding a .htaccess file to the directory upon creation (or when there isn't one, see below) that denies access in the same way that the .htaccess file in the `private` files directory works.

Also, when you delete the .htaccess file from the files directory and re-submit the File System form (admin/config/media/file-system), it re-creates the .htaccess file automatically. The same should be true of the `config` directory, and maybe even add a field to the File System page where you can specify the `config` directory (rather than in 'settings.php').

Advocate: @BWPanda

---

~Original PR: https://github.com/backdrop/backdrop/pull/3238~
New PR to address feedback form code review https://github.com/backdrop/backdrop/pull/3279

ghost:
Fixes https://github.com/backdrop/backdrop-issues/issues/706

### DIFF
--- a/core/includes/config.inc
+++ b/core/includes/config.inc
@@ -63,7 +63,7 @@ function config($config_file, $type = 'active') {
  * @param string $option
  *   The name of the config option within the file to delete. The config option
  *   may contain periods to indicate levels within the config file.
- * 
+ *
  * @see config_set()
  * @see config_get()
  *
@@ -1239,7 +1239,7 @@ class ConfigFileStorage implements ConfigStorageInterface {
 
   /**
    * Create a configuration directory, if it does not already exist, and ensure
-   * it is writable by the site.
+   * it is writable by the site. Additionally, protect it with a .htaccess file.
    *
    * @return bool
    *   TRUE on success, FALSE in case of an error.
@@ -1251,6 +1251,9 @@ class ConfigFileStorage implements ConfigStorageInterface {
       throw new ConfigStorageException(format_string('The directory %directory could not be created or could not be made writable. ', array(
       '%directory' => $config_directory,
     )));
+    }
+    else {
+      file_save_htaccess($this->directory);
     }
   }
 
--- a/core/modules/config/tests/config.test
+++ b/core/modules/config/tests/config.test
@@ -158,6 +158,23 @@ class ConfigurationTest extends BackdropWebTestCase {
     $this->assertEqual(FALSE, $config_data->isOverridden('foo.key1'), 'Original value no longer overridden.');
     $this->assertEqual($config_data->get('foo.key1'), 'chocolate', 'Original value has been updated to include overridden value.');
   }
+
+  /**
+   * Tests that config directories are protected by .htaccess files.
+   */
+  public function testConfigHtaccess() {
+    $dir = config_get_config_directory();
+
+    // Remove .htaccess file to then test that it gets re-created.
+    @backdrop_unlink(config_get_config_directory() . '/.htaccess');
+    $this->assertFalse(is_file(config_get_config_directory() . '/.htaccess'), "Successfully removed the .htaccess file in the 'active' config directory.");
+    file_save_htaccess($dir);
+    $this->assertTrue(is_file(config_get_config_directory() . '/.htaccess'), "Successfully re-created the .htaccess file in the 'active' config directory.");
+
+    // Verify contents of .htaccess file.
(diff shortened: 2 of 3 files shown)


quicksketch:
The current test failure seems related to these changes:
```
    fail The .htaccess file contains the proper content. config.test:174
```

core/includes/config.inc
```diff
@@ -1252,6 +1252,9 @@ class ConfigFileStorage implements ConfigStorageInterface {
       '%directory' => $config_directory,
     )));
     }
+    else {
+      file_save_htaccess($this->directory);
```
quicksketch, line 1256: Shouldn't this be called right after the directory is created as well? That is, move it outside of the `else {}` statement.

core/modules/config/tests/config.test
```diff
@@ -158,6 +158,21 @@ class ConfigurationTest extends BackdropWebTestCase {
...
   }
+
+  /**
+   * Tests that config directories are protected by .htaccess files.
+   */
+  public function testConfigHtaccess() {
+    // Remove .htaccess file to then test that it gets re-created.
+    @backdrop_unlink(config_get_config_directory() . '/.htaccess');
+    $this->assertFalse(is_file(config_get_config_directory() . '/.htaccess'), "Successfully removed the .htaccess file in the 'active' config directory.");
+    file_save_htaccess(config_get_config_directory());
```
quicksketch, line 169: Instead of calling `file_save_htaccess()`, could you do a normal web request and confirm that the .htaccess file is remade as needed?

core/modules/system/system.install
```diff
@@ -235,6 +235,17 @@ function system_requirements($phase) {
...
     // (for example) the /tmp directory was wiped.
     file_ensure_htaccess();
+
+    foreach (array('active', 'staging') as $type) {
+      $dir = config_get_config_directory($type);
+      file_save_htaccess($dir);
+
+      $htaccess_files[$dir . '/.htaccess'] = array(
+        'title' => $t('%type config directory', array('%type' => backdrop_ucfirst($type))),
+        'directory' => $dir,
```
quicksketch: Can you expand these `$dir` variables to the full spelling `$config_directory`?

core/includes/config.inc
ghost: My understanding is that `if (!file_prepare_directory(` means the directory wasn't created or isn't writable. Which means the `else {` is when the directory is available and writable. Moving the code out of the `else` risks running it when there's no writable directory, right?

core/modules/config/tests/config.test
ghost: Not sure what you mean, sorry...

core/modules/system/system.install
ghost: No problem, done.

core/modules/config/tests/config.test
ghost: Oh, you mean access the Status Report page to see if that triggers it?

quicksketch: Yeah that would be perfect.

core/includes/config.inc
quicksketch: > Moving the code out of the else risks running it when there's no writable directory, right?

Not in this case, because if the directory doesn't exist we `throw new ConfigStorageException`, which immediately exits the function, so nothing below it will get called.

jenlampton:
Replaced by https://github.com/backdrop/backdrop/pull/3279
addresses feedback from this code review.

core/modules/config/tests/config.test
jenlampton: resolved in new PR https://github.com/backdrop/backdrop/pull/3279

core/includes/config.inc
jenlampton: resolved in new PR https://github.com/backdrop/backdrop/pull/3279

core/modules/config/tests/config.test
jenlampton: Hm. the tests are still failing. I've confirmed the file is there on disk (and recreates successfully when visiting the status report) but these tests can't see it either before it's deleted, or after it's been recreated.

PR #3249, opened 2020-08-16
Issue #4545: Explain why the ellipsis in truncate_utf8() is wrapped in t()

### LINKED ISSUE #4545: [DX] Add inline comment to explain why the ellipsis in truncate_utf8() is wrapped in t()
https://github.com/backdrop/backdrop/blob/1.x/core/includes/unicode.inc#L338

```php
  if ($add_ellipsis) {
    // Truncate ellipsis in case $max_length is small.
    $ellipsis = backdrop_substr(t('…'), 0, $max_length);
    $max_length -= backdrop_strlen($ellipsis);
    $max_length = max($max_length, 0);
  }
```

Full story behind that here: https://www.drupal.org/project/drupal/issues/86461#comment-2829792

> According to the [Wikipedia article](http://en.wikipedia.org/wiki/Ellipsis), ellipses are used in many languages, including Chinese, though in that case, it is usually represented as 6 dots arranged vertically. It seems to me that the purpose of the `$dots` parameter is to "indicate an intentional omission," and that any logic about how (and whether) to represent it ought to be handled at the translation layer, since it's language specific.
>
>While `truncate_utf8()` just adds `' ...'`, perhaps it should add `t('&hellip;')`, `t('&#8230;')`, or something similar instead. I don't know if passing HMTL entities or numeric character references is valid in `t()`, but you get the point. A translator with the appropriate knowledge could then provide the proper character(s) for his or her language.
>
>Either way, I don't think it's something specific to truncating paths, and perhaps ought to be addressed in how `truncate_utf8()` handles it.

klonos:
Fixes https://github.com/backdrop/backdrop-issues/issues/4545


quicksketch:

PR #3250, opened 2020-08-16
Issue #4546: Odd text in the $break_points array in text_summary()

### LINKED ISSUE #4546: Odd text in the $break_points array in text_summary()
@BWPanda brought this up in Zulip:

> What is going on with the text at the end of this line?https://github.com/backdrop/backdrop/blob/1.x/core/modules/field/modules/text/text.module#L415

```php
  // If the first paragraph is too long, split at the end of a sentence.
  $break_points[] = array('. ' => 1, '! ' => 1, '? ' => 1, '。' => 0, '؟ ' => 1);
```

@klonos
>Good catch @BWPanda ...seems like the result of a search/replace or something like that. ...it's about line breaks. ...the equivalent line in D7 is exactly the same: https://git.drupalcode.org/project/drupal/-/blob/7.x/modules/field/modules/text/text.module#L420
>...same thing in D8: https://git.drupalcode.org/project/drupal/-/blob/8.9.x/core/modules/text/text.module#L134 (and assuming in D9)

@BWPanda
> It's weird though, because if I move my cursor through the line left-to-right, it jumps around towards the end...

Peek-2020-08-16-21-56

@klonos
>Yeah, I noticed that too ...seems like some odd tab character or something.
>...reminds me of the same odd thing that's happening when you try to select text split in <span>s in RTL languages.
>...there you go, that character is the Arabic question mark: https://en.wiktionary.org/wiki/؟

@BWPanda
>Ok, so that character is making that section of text be displayed RTL...

@klonos
>...not sure how it works TBH ...that array element declaration seems wrong to me.

klonos:
Fixes https://github.com/backdrop/backdrop-issues/issues/4546

### DIFF
--- a/core/modules/field/modules/text/text.module
+++ b/core/modules/field/modules/text/text.module
@@ -411,12 +411,11 @@ function text_summary($text, $format_id = NULL, $size = 600) {
   }
   $break_points[] = $line_breaks;
 
-  // Typing a number after the ؟ character causes text to be reversed mid-line.
-  // Using this variable here as a workaround, to stop that from happening. See
-  // https://github.com/backdrop/backdrop-issues/issues/4546
-  $arabic_questionmark_offset = 1;
-  // If the first paragraph is too long, split at the end of a sentence.
-  $break_points[] = array('. ' => 1, '! ' => 1, '? ' => 1, '。' => 0, '؟ ' => $arabic_questionmark_offset);
+  // Typing a number after the ؟ character (Arabic question mark) causes text to
+  // be reversed mid-line. That's the reason why the last element in this array
+  // may seem odd. It actually produces valid code.
+  // See https://github.com/backdrop/backdrop-issues/issues/4546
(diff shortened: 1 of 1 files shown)


core/modules/field/modules/text/text.module
```diff
@@ -411,6 +411,10 @@ function text_summary($text, $format_id = NULL, $size = 600) {
   }
   $break_points[] = $line_breaks;
 
+  // Typing a number after the ؟ character (Arabic question mark) causes text to
+  // be reversed mid-line. That's the reason why the last element in this array
+  // may seem odd. It actually produces valid code.
+  // See https://github.com/backdrop/backdrop-issues/issues/4546
   // If the first paragraph is too long, split at the end of a sentence.
```
ghost: I think this comment line should appear first, then your addition.

klonos: Done 👍

ghost:
Merged into 1.x.

PR #3251, opened 2020-08-17
Issue #4547: Better alignment of 'Add new field' row on 'Manage fields' page.

### LINKED ISSUE #4547: Better alignment of 'Add new field' row on 'Manage fields' page
The 'Add new field' row on the 'Manage fields' page is currently not aligned very well. The proposed PR fixes this as shown:

**Before:**
before1

**After:**
after1

And also when the drag handle sits above the label:

**Before:**
before2

**After:**
after2

---

**Original post**

Here's how that row currently looks normally:
image

And here's how it looks if the cells are wider and the draggable handle sit above the field:
image

ghost:
Fixes https://github.com/backdrop/backdrop-issues/issues/4547

### DIFF
--- a/core/modules/field_ui/css/field_ui.css
+++ b/core/modules/field_ui/css/field_ui.css
@@ -37,7 +37,11 @@
   display: none;
 }
 .field-ui-overview tr.add-new td {
+  vertical-align: bottom;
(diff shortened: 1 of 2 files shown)


quicksketch:

PR #3252, opened 2020-08-17
Issue #4548: Remove check for 'php_code' filter from text_summary().

### LINKED ISSUE #4548: Remove check for 'php_code' filter
Since we removed the PHP Filter module from core, we should remove the check for that filter from [`text_summary()`](https://api.backdropcms.org/api/backdrop/core%21modules%21field%21modules%21text%21text.module/function/text_summary/1):

```php
// We check for the presence of the PHP evaluator filter in the current
// format. If the body contains PHP code, we do not split it up to prevent
// parse errors.
if (isset($format->filters['php_code']) && $format->filters['php_code']->status && strpos($text, '<?') !== FALSE) {
  return $text;
}
```

ghost:
Fixes https://github.com/backdrop/backdrop-issues/issues/4548


klonos (approved):
LGTM 👍

ghost:
Merged into 1.x and 1.17.x.

PR #3254, opened 2020-08-17
Issue #1001: Increase width of admin bar search results drop-down.

### LINKED ISSUE #1001: [UX] Increase the width of the admin bar search results list (so that less results wrap in two lines).
This is why:

admin_bar-search_results-narrow

I was thinking perhaps increase the width to 300px? (this will BTW also accommodate some of the changes I proposed [here](https://github.com/backdrop/backdrop-issues/issues/525#issuecomment-111666887)):

admin_bar-search_results-wide

There should be no side-effect because there's room to the right due to the `.admin-bar-users`/`.admin-bar-account`/`.admin-bar-logout` items next to the search field.

ghost:
Fixes https://github.com/backdrop/backdrop-issues/issues/1001


quicksketch:

PR #3261, opened 2020-08-21
Issue #1435: Undefined index: quantity in views_plugin_pager_full

### LINKED ISSUE #1435: Notice: Undefined index: quantity in views_plugin_pager_full->options_form() (line 90 of views_plugin_pager_full.inc).
I got this while playing with the available options when filing #1430:

`Notice: Undefined index: quantity in views_plugin_pager_full->options_form() (line 90 of \core\modules\views\plugins\views_plugin_pager_full.inc).`

Happened when I clicked on the link with the options part **while the pager was set to mini** (notice how the error points to `views_plugin_pager_full`), after the "Mini |" part (`?q=admin/structure/views/nojs/display/my_view_name/page/pager_options`)

klonos:
Fixes https://github.com/backdrop/backdrop-issues/issues/1435


core/modules/views/plugins/views_plugin_pager_mini.inc
```diff
@@ -26,7 +26,7 @@ class views_plugin_pager_mini extends views_plugin_pager_full {
   function option_definition() {
     $options = parent::option_definition();
 
-    unset($options['quantity']);
+    $options['quantity'] = NULL;
     unset($options['tags']['first']);
     unset($options['tags']['last']);
```
ghost, line 33: What's wrong with just leaving the parent form's default values in place? This'll prevent this issue of missing values and possible future relayed issues with the other unneeded values.

```suggestion
```

indigoxela: > What's wrong with just leaving the parent form's default values in place?

Then the form items for first/last/quantity would show up in the settings form - they shouldn't, as the mini pager doesn't use these settings. It would be OK to keep the "unset($form['quantity']);" because that simply removes the form item, which shouldn't cause any problems. Setting it to NULL doesn't do any harm, either.

The `$options['quantity'] = NULL;` in function option_definition() is the essential change to fix the notice.

indigoxela:
> This branch is 1 commit ahead, 268 commits behind backdrop:1.x. 

@klonos can you please rebase your PR? Maybe then we get some movement here.

klonos:
> @klonos can you please rebase your PR?

Done @indigoxela 👍🏼 

Do you think that it's worth adding an inline comment here, to mention that we are intentionally setting to `NULL` instead of `unset()`ing? I'm afraid that in the future someone might revert this in an effort to bring us in parity with Views 7.x.

indigoxela:
> Do you think that it's worth adding an inline comment here, to mention that we are intentionally setting to `NULL`

Yes, that makes sense to me. It's needed to have the array key, but that's not self-explanatory.

One thing: IMO the change in function options_form() isn't necessary, I think, the change in function option_definition() should be sufficient. Did you already try that out?

core/modules/views/plugins/views_plugin_pager_mini.inc
```diff
@@ -42,7 +42,8 @@ class views_plugin_pager_mini extends views_plugin_pager_full {
    */
   function options_form(&$form, &$form_state) {
     parent::options_form($form, $form_state);
-    unset($form['quantity']);
+
+    $options['quantity'] = NULL;
```
indigoxela: Oops, that should be `$form['quantity'] = NULL;` - different variable than in function option_definition().

Yes, this change is necessary, too. I confirmed locally.

indigoxela: Update: no it's not necessary (blame caching). So better revert that change.

klonos:
Random failure: https://github.com/backdrop/backdrop-issues/issues/5226

core/modules/views/plugins/views_plugin_pager_mini.inc
```diff
@@ -26,7 +26,10 @@ class views_plugin_pager_mini extends views_plugin_pager_full {
   function option_definition() {
     $options = parent::option_definition();
 
-    unset($options['quantity']);
+    // Intentionally setting $options['quantity'] to NULL here instead of
+    // unset()'ing it, otherwise it throws "undefined index" notices.
```
laryn: @klonos  Is `unset()'ing it` the correct/intended spelling here? Looks like a typo to me.

laryn: I get why it's spelled that way, it just looks strange to me. I don't feel strongly about it.

core/modules/views/plugins/views_plugin_pager_mini.inc
klonos: Done with [`1d3e9a0` (#3261)](https://github.com/backdrop/backdrop/pull/3261/commits/1d3e9a0a89f7e75d5007dfb133f1d65f5ccf75ec)

core/modules/views/plugins/views_plugin_pager_mini.inc
klonos: I'll leave things as they are for now 😅

core/modules/views/plugins/views_plugin_pager_mini.inc
klonos: @laryn I've changed this to "... instead of using unset() ..." 😉 

See: [`a928f70` (#3261)](https://github.com/backdrop/backdrop/pull/3261/commits/a928f70dea988585cbf5f52ffa4227c7ba5dcb9d)

core/modules/views/plugins/views_plugin_pager_mini.inc
```diff
@@ -26,7 +26,10 @@ class views_plugin_pager_mini extends views_plugin_pager_full {
   function option_definition() {
     $options = parent::option_definition();
 
-    unset($options['quantity']);
+    // Intentionally setting $options['quantity'] to NULL here instead of using
+    // unset(), otherwise it throws "undefined index" notices.
+    // See: https://github.com/backdrop/backdrop-issues/issues/1435
```
ghost: Not sure we need to link to issues each time we make a change in core...

If we do decide this link is required here, I'd suggest the following syntax instead:
```suggestion
    // @see https://github.com/backdrop/backdrop-issues/issues/1435
```

ghost (changes requested):
IMO we should either remove the link, or re-format it.

core/modules/views/plugins/views_plugin_pager_mini.inc
klonos: I thought that `@see` was reserved for pointing to code, but you are absolutely right @BWPanda 👍🏼 

@see (pun 😅 ): https://www.drupal.org/docs/develop/standards/api-documentation-and-comment-standards#see

klonos: ...as for adding issue links, I think that this is only needed if there is more to it (i.e. a few words in the docblock are not enough to explain what's happening). Happy to remove it entirely (I believe that the core committer that merges this can take care of that - if not, then let me know, and I'll do it).

ghost (approved):

quicksketch:

PR #3263, opened 2020-08-24
Fixing problem with label for comment title field

### LINKED ISSUE #4413: [UX] Better label for "allow comment title" option
We are working on a Backdrop project which intensively deals with comments, so expect lot's of comment-related issues from AltaGrade these days.

**Description of the need**

Ensure the comment titles are not allowed in content type settings:

no-comment-titles

Add new comments on a content page and you will see the `(No subject)` comment titles everywhere like so:

comment-subjects

**Proposed solution**

I believe the whole point of not allowing comment titles in the settings is about not seeing any kind of titles. The `(No subject)` comment titles would make sense if the comment titles were allowed, but entered empty when comments submitted. Therefore, I suggest to completely remove comment titles in the comments if they were not allowed in the settings.

stpaultim:
Fixes backdrop/backdrop-issues#4413

### DIFF
--- a/core/modules/comment/comment.module
+++ b/core/modules/comment/comment.module
@@ -1193,11 +1193,11 @@ function comment_form_node_type_form_alter(&$form, $form_state) {
       ),
       '#access' => user_access('post comments', backdrop_anonymous_user()),
     );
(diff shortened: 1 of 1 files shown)


ghost:

PR #3268, opened 2020-08-26
Issue #4573: Fix URL alias 'maxlength' attribute on content type forms.

### LINKED ISSUE #4573: URL alias pattern 'maxlength's don't match
The 'Default URL alias pattern' field on `/admin/config/urls/path/patterns` has a `maxlength` attribute set to `1280`. But the same field on `/admin/structure/types/manage/[CONTENT_TYPE]` has a `maxlength` attribute set to `128`. We should increase the latter to match the former.

ghost:
Fixes https://github.com/backdrop/backdrop-issues/issues/4573


ghost:
Merged into 1.x and 1.16.x.

PR #3272, opened 2020-08-29
Issue #1259: CKEditor in source mode does not save when in dialogs.

### LINKED ISSUE #1259: When editing in source mode (inside a dialogue window) changes are not saved unless you switch back out of source mode
**Issue Summary (UPDATED July 31, 2020):**

If you edit a text field inside of a dialogue block (is that the right term) using the source mode and do not turn the source mode off before saving, your changes are lost. 

It had been thought that this problem was specific to custom blocks, but during a recent dev meeting we speculated (and later confirmed) that the problem is related to dialogue windows and NOT custom blocks. The same behavior happens when editing a view header with CKEditor in source mode (which happens in a dialogue window). 

The problem does NOT exist when creating a custom block on the custom block page (which does not use a dialogue window). It does happen when adding a custom block in directly to the layout (which does happen in a dialogue window). 

**Steps to reproduce:**

- Download and install Backdrop CMS (happening as recent as 1.16.2)
- Create a custom block. (/admin/structure/block/add)
- Add a layout (/admin/structure/layouts/add)
- Add the custom block to the layout.
- Click the "Configure" button to the right of the custom block.
- In the "Block content" editor input, click the "Source" button.
- Edit the block content.
- Click "Update block"
- Click "Save layout"

**Proposed Solution**

Look at core/misc/dialog.ajax.js to see if we are using the correct JQuery handler/event or series of handlers/events to process this action. 

(SEE comments on July 31, 2020)

**Original Issue:**

If we try to change image URL (created before in our block) then it would not be saved if it is done in Source mode. You need to switch to normal mode after the change then it will be saved.

quicksketch:
Fixes https://github.com/backdrop/backdrop-issues/issues/1259

### DIFF
--- a/core/modules/ckeditor/js/ckeditor.js
+++ b/core/modules/ckeditor/js/ckeditor.js
@@ -32,7 +32,13 @@
     },
 
     detach: function (element, format, trigger) {
-      var editor = CKEDITOR.dom.element.get(element).getEditor();
+      // CKEditor's getEditor() method takes a single parameter for "optimized"
+      // that defaults to true. This makes it so that only unmodified DOM
+      // textarea elements will qualify for locating the CKEditor instance.
+      // During detachment, other behaviors may also modify the source textarea,
+      // causing CKEditor to lose track of the editor. Therefore pass "false"
+      // to use the more aggressive attempt to find the editor instance.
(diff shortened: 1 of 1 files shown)


ghost:
Thanks @quicksketch! Merged into 1.x and 1.16.x.

PR #3274, opened 2020-08-30
Issue #4555: Use 'placeholder' when we mean %, and 'wildcard' when we…

### LINKED ISSUE #4555: [UX] Better help text for layout path field
**Description of the need**

I just spend a long time trying to figure out the correct path to create a custom layout for taxonomy terms. 
https://forum.backdropcms.org/forum/how-do-i-configure-custom-layout-taxonomy-term-page

In retrospect, it shouldn't have been that difficult. I think that the 3 most common layouts of this type would node/%, taxonomy/term/%, and user/%. Why not include all three of these examples in the help text?

Not even sure that is the right path for user, wouldn't it be nice if there was a clue in the UI. :-)

**Proposed solution**

Change the following:
OLD: "Use the "%" (percent) symbol to indicate a wildcard in the path. i.e. "node/%" or "node/%/edit"."
NEW"Use the "%" (percent) symbol to indicate a wildcard in the path. i.e. "node/%", "taxonomy/term/%", "user/%", "node/%/edit", etc.

1. Better documentation for this in the user manual. 

**Additional information**

Add_layout___Backdrop_CMS_Live

---

PR: https://github.com/backdrop/backdrop/pull/3274

jenlampton:
Fixes https://github.com/backdrop/backdrop-issues/issues/4555

### DIFF
--- a/core/modules/layout/layout.admin.inc
+++ b/core/modules/layout/layout.admin.inc
@@ -272,12 +272,21 @@ function layout_settings_form($form, &$form_state, Layout $layout) {
     }
   }
 
+  $items = array(
+    t('Content: <code>node/%</code>'),
+    t('User accounts: <code>user/%</code>'),
+    t('Taxonomy terms: <code>taxonomy/term/%</code>'),
+    t('Comments: <code>comment/%</code>'),
+  );
+  $description_list = theme('item_list', array('items' => $items));
+
   $form['path'] = array(
     '#title' => t('Path'),
+    '#field_prefix' => url('', array('absolute' => TRUE)),
     '#type' => 'textfield',
     '#default_value' => $layout->getPath(),
     '#required' => TRUE,
-    '#description' => t('Use the "%" (percent) symbol to indicate a wildcard in the path. i.e. "node/%" or "node/%/edit".'),
+    '#description' => '<p>' . t('Use the percent symbol ("%") to indicate a placeholder in the path. Some common examples include:') . '</p>' . $description_list,
     '#access' => !$layout->isDefault(),
   );
   $form['path_update'] = array(
@@ -308,14 +317,13 @@ function layout_settings_form($form, &$form_state, Layout $layout) {
     '#title' => t('Contexts'),
     '#type' => 'item',
     '#id' => 'context-wrapper',
-    '#description' => t('(Optional) Contexts define what blocks are available within this layout. Most wildcards in the path will be associated with a context automatically.'),
+    '#description' => t('(Optional) Contexts define what blocks are available within this layout. Most placeholders in the path will be associated with a context automatically.'),
     // Current user context is always present.
   );
   $form['context_wrapper']['#access'] = !$layout->isDefault();
   $form['context_wrapper']['#prefix'] = '<div id="layout-contexts">';
   $form['context_wrapper']['#suffix'] = '</div>';
   $form['context_wrapper']['context'] = array(
-    '#title' => t('Contexts'),
     '#type' => 'container',
     '#parents' => array('context'),
   );
@@ -419,7 +427,7 @@ function layout_settings_form($form, &$form_state, Layout $layout) {
   }
 
   // Display a notice if overriding a system path that normally contains
-  // wildcards e.g. node/1 instead of node/%. This is usually (but not always)
+  // placeholders e.g. node/1 instead of node/%. This is usually (but not always)
   // a misconfiguration of the layout.
   $layout_path = $layout->getPath();
   $router_item = menu_get_item($layout_path);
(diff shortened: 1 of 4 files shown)


core/modules/layout/layout.admin.inc
```diff
@@ -271,12 +271,21 @@ function layout_settings_form($form, &$form_state, Layout $layout) {
     }
   }
 
+  $items = array(
+    t('Content: &nbsp; <code>node/%</code>'),
+    t('User accounts: &nbsp; <code>user/%</code>'),
+    t('Taxonomy terms: &nbsp; <code>taxonomy/term/%</code>'),
+    t('Comments: &nbsp; <code>comment/%</code>'),
```
klonos: These `&nbsp;`s are not required here (the less HTML we show to translators, the better):

```suggestion
    t('Content: <code>node/%</code>'),
    t('User accounts: <code>user/%</code>'),
    t('Taxonomy terms: <code>taxonomy/term/%</code>'),
    t('Comments: <code>comment/%</code>'),
```

If there needs to be spacing between colon and the code tag (which I don't think it does), then let's add it via CSS.

core/modules/layout/layout.admin.inc
```diff
@@ -271,12 +271,21 @@ function layout_settings_form($form, &$form_state, Layout $layout) {
...
+  $description_list = theme('item_list', array('items' => $items));
+
   $form['path'] = array(
     '#title' => t('Path'),
+    '#field_prefix' => url('', array('absolute' => TRUE)),
     '#type' => 'textfield',
     '#default_value' => $layout->getPath(),
     '#required' => TRUE,
-    '#description' => t('Use the "%" (percent) symbol to indicate a wildcard in the path. i.e. "node/%" or "node/%/edit".'),
+    '#description' => '<p>' . t('Use the percent symbol ("%") to indicate a placeholder in the path. Some common examples include:') . '</p>' . $description_list,
```
klonos, line 289: No need for `<p>` tags here either. This works just fine:

```suggestion
    '#description' => t('Use the percent symbol ("%") to indicate a placeholder in the path. Some common examples include:') . ' ' . $description_list,
```

core/modules/layout/layout.admin.inc
jenlampton: It was really hard to read without the extra space. We could add it via CSS, but on this page only? Or for all code tags in Seven?

core/modules/layout/layout.admin.inc
jenlampton: Again, I found this hard to read when all squished together :) But I'm going to merge this because it's done and I like the code better this way. DX over UX?

core/modules/layout/layout.admin.inc
klonos: I think that this would be a general issue, so should be fixed in Seven. I agree that this can be done in a follow-up 👍

docwilmot:
Did some stuff creep in here that arent this issue?

core/modules/layout/layout.module
```diff
@@ -582,23 +582,23 @@ function layout_route_handler($router_item) {
...
-  // include a permissions check on the admin theme.
-  if ($selected_layout->name === 'admin_default' && !user_access('view the administration theme')) {
-    $selected_layout = layout_load('default');
-  }
-
   // Special handling for 404 and 403 pages to render in admin theme. The
   // current path will be the 404/403 system path, and we need to check the
   // original path, which is stored in "destination" by
   // backdrop_deliver_html_page().
-  if (in_array(backdrop_get_http_header('Status'), array('404 Not Found', '403 Forbidden')) && user_access('view the administration theme')) {
```
docwilmot: Is this another issue?

core/modules/layout/layout.module
```diff
@@ -582,23 +582,23 @@ function layout_route_handler($router_item) {
...
+  if (in_array(backdrop_get_http_header('Status'), array('404 Not Found', '403 Forbidden'))) {
     if (isset($_GET['destination']) && path_is_admin($_GET['destination'])) {
       $selected_layout = layout_load('admin_default');
     }
   }
 
+  // Safety check that we don't show an admin layout to anyone who cannot access
+  // administrative pages. This may happen because the cached layout list does
+  // not include a permissions check on the admin theme.
+  if ($selected_layout->name === 'admin_default' && !user_access('access administration pages')) {
```
docwilmot: Is this another issue?

core/modules/layout/layout.module
jenlampton: yep! Just removed these changes :)

core/modules/layout/layout.module
jenlampton: yep :)

core/modules/layout/layout.admin.inc
jenlampton: fixed in latest pr

klonos (approved):
LGTM 👍

jenlampton:
closing temporarliy...

jenlampton:
new sandboxes?

ghost:

PR #3276, opened 2020-08-30
Issue #4584: Add a :link sub-token for user tokens

### LINKED ISSUE #4584: Add a :link sub-token for user tokens
This is basically the same as the following Token 7.x issues:

- https://www.drupal.org/project/token/issues/1176592

> I'm working on D7 Submitted By module and found that there is no way to have a `$submitted` with a linked user name, like we normally have for content and comment.
>
> Suggesting adding a fully themed username, `[user:name_themed]` via `theme('username')`

- https://www.drupal.org/project/token/issues/698242

> `[user]` is replaced with `$account->name`. This makes it difficult when trying to output all user names the same way by overriding `theme_username()`.
> 
> Is there a reason that this line:
> 
> ```php
> $values['user'] = $account->uid ? check_plain($account->name) : variable_get('anonymous', 'Anonymous');
> ```
>
> Cannot be replaced with this line?:
>
> ```php
> $values['user'] = theme('username', $account);
> ```
>
>This would be a lot more consistent with how user names are output. If this change is not acceptable, then perhaps we could introduce a new token that does call the proper theme function.

klonos:
Fixes https://github.com/backdrop/backdrop-issues/issues/4584


core/modules/user/user.tokens.inc
```diff
@@ -40,6 +40,10 @@ function user_token_info() {
     'description' => t('The user roles associated with the user account.'),
     'type' => 'array',
   );
+  $user['link'] = array(
+    'name' => t("Profile link"),
+    'description' => t("Link to account profile page based on the <em>View user profiles</em> permission."),
```
ghost: After reading the names and descriptions of the other tokens in this function, I think the following wording would be better to use:

```suggestion
    'name' => t('Name (linked)'),
    'description' => t("The login name of the user account, displayed as a link to the user's profile page."),
```

klonos: I've updated these according to additional feedback in the issue.

ghost:
@klonos I'm having trouble with this PR in Tugboat. I think maybe you started this branch before the Tugboat files were in core... So could you please try updating your branch with the latest core `1.x`?

klonos:
OK, lets try a close/reopen first...

klonos:
...OK, updating my branch did it @BWPanda 👍

ghost:
Thanks! Merged into 1.x.

PR #3280, opened 2020-08-31
Issue #4569: Allow admin to hide the title on comments

VasasA:
Allow admin to hide the title on comments: https://github.com/backdrop/backdrop-issues/issues/4569


VasasA:
I will create a new updated PR.

PR #3283, opened 2020-09-02
Issue #4592: Use Filter library to fix file loading order.

### LINKED ISSUE #4592: Regression: CKEditor broken after #4527
**Description of the bug**

On nodes with CKEditor enabled fields and image fields the editor is broken.

js error:
`Uncaught TypeError: Backdrop.editors is undefined`

The problem seems to be the order of js files.
If fillter.js runs after ckeditor.js Backdrop.editors is undefined in ckeditor.js

I can fix the problem locally, if instead of adding the js file via `#attached` adding it via `backdrop_add_js()`, I don't even need to set a weight. ckeditor.js gets loaded after filter.js then (as the js dependencies are properly defined).

It seems that when using `#attached` the js file "slips" into the wrong js group.

**Steps To Reproduce**

1. Get a version of Backdrop including [this recent change](https://github.com/backdrop/backdrop/commit/9e72b28d570d585cdb68d87e203d96278a152a2f)
2. Go to node/add/post (post has a textarea with CKEditor enabled and an image field)
3. Editor won't load

**Additional information**

**Proposed solution**

~Use `backdrop_add_js()` instead of '#attached' in file_managed_file_process(). Or are there better ideas?~

Use a `#attached` library that handles the dependency instead of adding the CSS and JS individually.

PR: https://github.com/backdrop/backdrop/pull/3283/files

quicksketch:
Fixes https://github.com/backdrop/backdrop-issues/issues/4592

### DIFF
--- a/core/modules/file/file.module
+++ b/core/modules/file/file.module
@@ -1264,7 +1264,6 @@ function file_admin_paths() {
  */
 function file_managed_file_process($element, &$form_state, $form) {
   $fid = isset($element['#value']['fid']) ? $element['#value']['fid'] : 0;
-  $filter_path = backdrop_get_path('module', 'filter');
 
   // Set some default element properties.
   $element['#progress_indicator'] = empty($element['#progress_indicator']) ? 'none' : $element['#progress_indicator'];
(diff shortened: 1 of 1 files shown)


quicksketch:
Now RTBC in the issue, merged into 1.x and 1.16.x.

PR #3285, opened 2020-09-02
Issue #4591: Update system.core config name.

### LINKED ISSUE #4591: Mismatch in config file name and _config_name value in system.core.json
**Description of the bug**

I'm not 100% sure if this is a bug (it didn't seem to cause problems so far), but it's a bit weird:

The system module ships with default config in file system.core.json (core/modules/system/config), but the "_config_name" inside that file is `"system.performance"`.

Is this by intention or an oversight?

Note: the actual config file (files/config_thelonghashhere/active/system.core.json) does contain `"_config_name": "system.core"`

**Steps To Reproduce**

There's nothing to reproduce, just json content: https://github.com/backdrop/backdrop/blob/1.x/core/modules/system/config/system.core.json

**Expected behavior**

Usually the filename and the value of "_config_name" correspond.

ghost:
Fixes https://github.com/backdrop/backdrop-issues/issues/4591

### DIFF
--- a/core/modules/system/config/system.core.json
+++ b/core/modules/system/config/system.core.json
@@ -1,5 +1,5 @@
 {
-    "_config_name": "system.performance",
+    "_config_name": "system.core",
     "_config_static": true,
     "_config_translatables": [
         "anonymous",


klonos (approved):
👍

ghost:
Merged into 1.x and 1.16.x.

PR #3288, opened 2020-09-03
Issue #4454: Fix typos in layout storage definition

### LINKED ISSUE #4454: Typos in Layout storage definition
**Description of the bug**

This is the comment on the `$storage` property in `layout.class.inc`:

>    The storage state of the layout.
   
>  This is the machine-version of the $type variable, which represents whether
    this view is has a default, user-created, or overridden configuration.
    Possible values for this variable include the constants

This may be copied from Views and needs fixing:
- I dont think there is a `$type` variable
- It should be "this layout" not "this view"
- And "is has" is not grammar.

docwilmot:
Fixes https://github.com/backdrop/backdrop-issues/issues/4454

### DIFF
--- a/core/modules/layout/includes/layout.class.inc
+++ b/core/modules/layout/includes/layout.class.inc
@@ -53,10 +53,10 @@ class Layout {
   /**
    * The storage state of the layout.
    *
-   * This is the machine-version of the $type variable, which represents whether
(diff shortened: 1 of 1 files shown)


ghost:
Merged into 1.x and 1.17.x.

PR #3290, opened 2020-09-03
Issue #4383: Make default layout path null

### LINKED ISSUE #4383: Why do the Default and Admin Default layouts have different paths?
Default layout `path = ""` whereas Admin Default `path = NULL`.
Minor, but if you call `menu_get_item()` on the first path, it returns `FALSE` but on the second it returns a menu item.
Bug?

docwilmot:
Fixes https://github.com/backdrop/backdrop-issues/issues/4383


docwilmot:
Reopening. Sandbox was busted.

quicksketch:

PR #3296, opened 2020-09-07
Issue #4604: Prevent PHP fatal in upgrades, layout contexts should al…

### LINKED ISSUE #4604: Layout context regression: Updating to 1.17.0 throws PHP fatal "Unsupported operand types" error
**Description of the bug**

I am attempting to update a site from backdrop 1.15.0 to 1.17.0-preview, and I get the following PHP fatals when running update.php

It appears as though core assumes `$this-contexts` and `$this->getContexts()` will always return an array. It does not check to see if the value is an array in many places. It immediately runs a `foreach()` on the result in some places, and in others does `$this->contexts += $contexts` assuming both are arrays. Neither of these will work if contexts is not an array. 

```
Error
Error message

    Warning: Invalid argument supplied for foreach() in Layout->getContexts() (line 773 of backdrop/core/modules/layout/includes/layout.class.inc).
    Warning: Invalid argument supplied for foreach() in LayoutMenuItem->resetContexts() (line 265 of backdrop/core/modules/layout/includes/layout_menu_item.class.inc).
    Error: Unsupported operand types in LayoutMenuItem->getContexts() (line 253 of backdrop/core/modules/layout/includes/layout_menu_item.class.inc).

Unsupported operand types

```

**Steps To Reproduce**

To reproduce the behavior:
1. Install backdrop 1.15.0, 1.15.1, or 1.16.3
2. Replace code with 1.17.0-preivew
3. Visit `/core/update.php` and run the updates using the UI
4. Update fails due to fatal error:

Screen Shot 2020-09-10 at 4 32 10 PM

**Actual behavior**

PHP fatal prevents updates from running. 

**Expected behavior**

Update as usual

**Additional information**

We made some changes to `core/modules/layout/includes/layout.class.inc` in https://github.com/backdrop/backdrop-issues/issues/3750 that may have accidentally removed [the code that handled the case when layout contexts are null](https://github.com/backdrop/backdrop/pull/3271/files#diff-f9a46040a1020182a9f049bb8c6864e8L760). 

I've provided a PR that adds this back.

---

Sites tested:
* Site A: 1.16.2 - PHP 7.2.19-0ubuntu0.19.04.1
* Site B: 1.15.0 - PHP 5.5.9-1ubuntu4.11
* Site E: 1.14.1 - PHP 5.4.45
* Site F: 1.16.2 - PHP 7.3
* Site J: 1.16.2 - PHP 7.3
* Site O: 1.15.1 - PHP 7.2.19-0ubuntu0.19.04.1

---

PR https://github.com/backdrop/backdrop/pull/3296

jenlampton:
Fixes https://github.com/backdrop/backdrop-issues/issues/4604

### DIFF
--- a/core/modules/layout/includes/layout.class.inc
+++ b/core/modules/layout/includes/layout.class.inc
@@ -250,6 +250,11 @@ class Layout {
         $this->{$property_key}[$plugin_type_key] = layout_create_handler($plugin_type, $plugin_data['plugin'], $plugin_data['data']);
       }
     }
+
+    // Ensure contexts is always set to an array.
+    if (is_null($this->contexts)) {
+      $this->contexts = array();
+    }
   }
 
   /**
@@ -769,6 +774,10 @@ class Layout {
    * return LayoutContext[]
    */
   function getContexts() {
+    if (is_null($this->contexts)) {
+      $this->contexts = array();
+    }
+
     // Load objects for stored contexts.
     foreach ($this->contexts as $key => $context) {
       if (!isset($context->position) && !is_object($context->data)) {
@@ -893,4 +902,3 @@ class Layout {
     return TRUE;
   }
 }
-
--- a/core/modules/layout/includes/layout_menu_item.class.inc
+++ b/core/modules/layout/includes/layout_menu_item.class.inc
@@ -137,6 +137,11 @@ class LayoutMenuItem {
     foreach ($this->conditions as $key => $condition) {
       $this->conditions[$key] = layout_create_handler('layout_access', $condition['plugin'], $condition['data']);
     }
+
+    // Ensure contexts is always set to an array.
+    if (is_null($this->contexts)) {
+      $this->contexts = array();
+    }
   }
 
   /**
@@ -229,7 +234,12 @@ class LayoutMenuItem {
    * return LayoutContext[]
    */
   function getContexts() {
+    if (is_null($this->contexts)) {
+      $this->contexts = array();
+    }
+
     $this->resetContexts();
+
     // Required contexts are generated on-demand.
     $contexts = layout_context_required_by_path($this->path);
     $path_parts = explode('/', $this->path);


jenlampton:
My first PR only added 3 lines to `layout.class.inc`

```php
if (is_null($this->contexts)) {
  $this->contexts = array();
}
```

But that didn't seem to be enough. I still got the fatal until I added the following to `layout_menu_item.class.inc`

```php
    if (!is_null($this->contexts)) {
      $this->contexts += $contexts;
    }
    else {
      $this->contexts = $contexts;
    }
```

Could that be because of a menu cache, maybe?

jenlampton:
In the latest PR I tried setting `$contexts` specifically in the constructor, and that also seemed to work. This seems cleaner. PR ready for code review.

core/modules/layout/includes/layout.class.inc
```diff
@@ -250,6 +250,11 @@ class Layout {
         $this->{$property_key}[$plugin_type_key] = layout_create_handler($plugin_type, $plugin_data['plugin'], $plugin_data['data']);
       }
     }
+
+    // Ensure contexts is set to an array, even if empty.
+    if (is_null($this->contexts)) {
```
klonos, line 255: Should this better be:

```suggestion
    if (!is_array($this->contexts)) {
```

core/modules/layout/includes/layout.class.inc
```diff
@@ -769,6 +774,10 @@ class Layout {
    * return LayoutContext[]
    */
   function getContexts() {
+    if (is_null($this->contexts)) {
```
klonos, line 777: Same as previous comment:

```suggestion
    if (!is_array($this->contexts)) {
```

core/modules/layout/includes/layout_menu_item.class.inc
```diff
@@ -137,6 +137,11 @@ class LayoutMenuItem {
     foreach ($this->conditions as $key => $condition) {
       $this->conditions[$key] = layout_create_handler('layout_access', $condition['plugin'], $condition['data']);
     }
+
+    // Ensure contexts is set to an array, even if empty.
+    if (is_null($this->contexts)) {
```
klonos, line 142: ...same her:

```suggestion
    if (!is_array($this->contexts)) {
```

core/modules/layout/includes/layout.class.inc
jenlampton: What was it on your site @klonos? On mine it was NULL.

core/modules/layout/includes/layout.class.inc
jenlampton: I've committed these changes, but I'm surprized the `is_null()` check didn't work for you @klonos. What was `$this->contexts` on your site? Was it an empty string? `''`

klonos: Hmm 🤔 ...I still see `is_null()`, but anyway.

I've made these suggestions before actually testing. I simply went by what was expected by the offending line in the errors (`$this->contexts += $contexts;`) ...so if `$this->contexts` was expected to be an array, I thought that checking if it is not an array would suffice. Besides, `NULL` is not an array, and neither is `''`, so `!is_array()` would've been true in those cases too.

klonos: ...I should have said that when I suggested the `!is_array()` I was coming directly from https://github.com/backdrop/backdrop-issues/issues/4611#issuecomment-689549746

core/modules/layout/includes/layout.class.inc
klonos: It happened on a PR sandbox, while I was testing https://github.com/backdrop/backdrop/pull/3274, so I couldn't check 🤷

quicksketch:
Thanks! Merged into 1.x for 1.17.0.

PR #3300, opened 2020-09-10
Issue #4614: Revert the change that aded a date-format for dblog.

### LINKED ISSUE #4614: Follow-ups for new dblog date format (PHP fatal: undefined function date_format_type_options(), update problems, config value name change)

TO address whenever:

**Description of the bug**

**Steps To Reproduce**

To reproduce the behavior:

**Actual behavior**

Format corrupted.

**Expected behavior**

Previous format remains

**Additional information**

---

jenlampton:


jenlampton:
Replaced by https://github.com/backdrop/backdrop/pull/3302

PR #3301, opened 2020-09-11
Issue #4614: Fix issue with dblog date format.

### LINKED ISSUE #4614: Follow-ups for new dblog date format (PHP fatal: undefined function date_format_type_options(), update problems, config value name change)

ghost:

### DIFF
--- a/core/modules/dblog/dblog.module
+++ b/core/modules/dblog/dblog.module
@@ -151,6 +151,12 @@ function dblog_watchdog(array $log_entry) {
  * Implements hook_form_FORM_ID_alter() for system_logging_settings().
  */
 function dblog_form_system_logging_settings_alter(&$form, $form_state) {
+  // Make sure the Date module file is available so we can call
+  // date_format_type_options().
+  if (!module_exists('date')) {
+    form_load_include($form_state, 'module', 'date');
+  }
+
   $form['dblog_row_limit'] = array(
     '#type' => 'select',
     '#title' => t('Database log messages to keep'),
@@ -158,7 +164,6 @@ function dblog_form_system_logging_settings_alter(&$form, $form_state) {
     '#options' => array(0 => t('All')) + backdrop_map_assoc(array(100, 1000, 10000, 100000, 1000000)),
     '#description' => t('The maximum number of messages to keep in the database log. Requires a <a href="@cron">cron maintenance task</a>.', array('@cron' => url('admin/reports/status')))
   );
-
   $form['dblog_date_format'] = array(
     '#type' => 'select',
     '#title' => t('Date format for log messages'),
--- a/core/modules/dblog/dblog.install
+++ b/core/modules/dblog/dblog.install
@@ -128,12 +128,12 @@ function dblog_update_1002() {
   // No matching date format found, so create a custom one.
   if ($config->get('log_date_format') == $log_format) {
     system_date_format_save(array(
-      'name' => 'dblog_date_format',
+      'name' => 'dblog',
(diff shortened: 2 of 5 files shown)


core/modules/dblog/dblog.module
```diff
@@ -151,14 +151,19 @@ function dblog_watchdog(array $log_entry) {
  * Implements hook_form_FORM_ID_alter() for system_logging_settings().
  */
 function dblog_form_system_logging_settings_alter(&$form, $form_state) {
+  // Make sure the Date module file is available so we can call
+  // date_format_type_options().
+  if (!module_exists('date')) {
+    form_load_include($form_state, 'module', 'date');
```
jenlampton, line 157: I don't think we should use any code from date module here, this work-around feels like a bit of a hack. Are there any other places in core (other than date module) that use date formats? How do they do it?

If this is the first one, then it would probably be better to move this function into `date.inc` so we don't need to add the work-around.

quicksketch: Agreed. We shouldn't be making dependencies between optional core modules.

Because `date_format_type_options()` calls `system_get_date_formats()` (in system.module), I think the most appropriate place would be making a new function in system module (e.g. `system_date_format_options()`) and then deprecating `date_format_type_options()`. Note that `date_example_date()` from date.module is also called, that can be moved to `date.inc`.

ghost: FTR, I don't see this as a hack. I see this as using the `form_load_include()` function precisely for what it was created for.

Here's a similar, but different, example: https://api.backdropcms.org/api/backdrop/core%21modules%21locale%21locale.admin.inc/function/locale_date_format_form/1
At the top of that file you'll see it includes the `system.admin.inc` file in order to call `system_date_time_lookup()`.

But, if you're really set on removing this call to the core Date module, here's an example from Views which instead calls `system_get_date_formats()` and then generates the options array from that manually: https://api.backdropcms.org/api/backdrop/core%21modules%21views%21handlers%21views_handler_field_date.inc/function/views_handler_field_date%3A%3Aoptions_form/1
I guess we could do something similar here, even though it's more code...

jenlampton: > I see this as using the form_load_include() function precisely for what it was created for.

This function wasn't created so that we could ignore a module's enabled state, it is intended to load files that might not be loaded due to a form cache. 

> At the top of that file you'll see it includes the system.admin.inc file in order to call system_date_time_lookup().

System is not an optional module, so this isn't an apples-to-apples comparison. 

> here's an example from Views...

If we move `date_format_type_options()` (or a comparable function with a different name) to a location where other modules could use it safely, we could continue to use it in `dblog`, and `views` could use it also. Not only do we fix our problem now, but we add a useful tool for other modules (even contrib!) to use when they need to present a list of date formats in an options form. :)

jenlampton:
I got excited and posted [a proof-of-concept PR](https://github.com/backdrop/backdrop/pull/3302) that moves `date_format_type_options()` and `date_example_date()` to `includes/date.inc`. I don't think it would be too much more work to deprecate `date_format_type_options()` in favor of a comparable function in `system.module` if people feel that approach is preferred -- let me know!

quicksketch:
This will need refactoring after reverting the original feature. We have a new issue at https://github.com/backdrop/backdrop-issues/issues/4621 to re-introduce this.

PR #3302, opened 2020-09-11
4614 dblog date formats

### LINKED ISSUE #4614: Follow-ups for new dblog date format (PHP fatal: undefined function date_format_type_options(), update problems, config value name change)

jenlampton:

### DIFF
--- a/core/includes/date.inc
+++ b/core/includes/date.inc
@@ -451,6 +451,44 @@ function date_format_patterns($strict = FALSE) {
 }
 
 /**
+ * Creates an array of date format types for use as an options list.
+ *
+ * @param bool $html
+ *   Whether to include HTML date options or not.
+ *
+ * @return array
+ *   All date format options, keyed by machine name.
+ */
+function date_format_type_options($html = TRUE) {
+  $options = array();
+  $format_types = system_get_date_formats();
+  if (!empty($format_types)) {
+    foreach ($format_types as $type => $type_info) {
+      if (isset($type_info['label'])) {
+        // If html dates are not requested, don't include them.
+        if (!$html && substr($type, 0, 5) == 'html_') {
+          continue;
+        }
+        $options[$type] = $type_info['label'] . ' (' . date_format_date(date_example_date(), $type) . ')';
+      }
+    }
+  }
+  return $options;
+}
+
+/**
+ * Creates an example date that shows a clear difference between month and day,
+ * and between 12 and 24 hours.
+ *
+ * @return Date
+ *   An example date object.
+ */
+function date_example_date() {
+  // use a fixed unambiguous date and time.
+  return new BackdropDateTime('2099-12-31 23:59:00');
+}
+
+/**
  * Constructs an array of granularity options and their labels.
  *
  * @return array
--- a/core/modules/date/date.admin.inc
+++ b/core/modules/date/date.admin.inc
@@ -4,6 +4,8 @@
  * Date administration code.
  */
 
+require_once(BACKDROP_ROOT . '/core/includes/date.inc');
+
 /**
  * Settings for the default formatter.
  */
--- a/core/modules/dblog/dblog.module
+++ b/core/modules/dblog/dblog.module
@@ -151,6 +151,12 @@ function dblog_watchdog(array $log_entry) {
  * Implements hook_form_FORM_ID_alter() for system_logging_settings().
  */
 function dblog_form_system_logging_settings_alter(&$form, $form_state) {
+  // Make sure the Date module file is available so we can call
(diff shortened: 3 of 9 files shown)


core/includes/date.inc
```diff
@@ -450,6 +450,44 @@ function date_format_patterns($strict = FALSE) {
...
+
+/**
+ * Creates an example date that shows a clear difference between month and day,
+ * and between 12 and 24 hours.
+ *
+ * @return Date
+ *   An example date object.
+ */
+function date_example_date() {
+  // use a fixed unambiguous date and time.
```
klonos, line 487:
```suggestion
  // Use a fixed, unambiguous date and time.
```

core/modules/views/handlers/views_handler_field_date.inc
```diff
@@ -26,11 +26,9 @@ class views_handler_field_date extends views_handler_field {
...
   function options_form(&$form, &$form_state) {
 
-    $date_formats = array();
-    $date_types = system_get_date_formats();
-    foreach ($date_types as $machine_name => $date_type) {
-      $date_formats[$machine_name] = $date_type['label'] . ': ' . format_date(REQUEST_TIME, $machine_name);
-    }
+    require_once(BACKDROP_ROOT . '/core/includes/date.inc');
+    $date_formats = date_format_type_options(FALSE);
+    dpm($date_formats);
```
klonos, line 31: This needs to be removed.

core/includes/date.inc
```diff
@@ -450,6 +450,44 @@ function date_format_patterns($strict = FALSE) {
...
+ * @return array
+ *   All date format options, keyed by machine name.
+ */
+function date_format_type_options($html = TRUE) {
+  $options = array();
+  $format_types = system_get_date_formats();
+  if (!empty($format_types)) {
+    foreach ($format_types as $type => $type_info) {
+      if (isset($type_info['label'])) {
+        // If html dates are not requested, don't include them.
```
klonos, line 468:
```suggestion
        // If HTML dates are not requested, don't include them.
```

core/includes/date.inc
indigoxela: Why has function date_example_date() been moved here anyway?

Never mind, got it, because it's used here.

core/modules/date/date.admin.inc
```diff
@@ -4,6 +4,8 @@
  * Date administration code.
  */
 
+require_once(BACKDROP_ROOT . '/core/includes/date.inc');
+
```
indigoxela, line 8: Actually that shouldn't be necessary. I'd assume that core/includes/date.inc should already be included anyway.

core/includes/date.inc
```diff
@@ -450,6 +450,44 @@ function date_format_patterns($strict = FALSE) {
...
+/**
+ * Creates an array of date format types for use as an options list.
+ *
+ * @param bool $html
+ *   Whether to include HTML date options or not.
+ *
+ * @return array
+ *   All date format options, keyed by machine name.
+ */
+function date_format_type_options($html = TRUE) {
```
indigoxela, line 462: Hm. Is that API change really necessary? Function date_format_type_options() currently has no params.

core/modules/dblog/dblog.module
```diff
@@ -151,19 +151,19 @@ function dblog_watchdog(array $log_entry) {
  * Implements hook_form_FORM_ID_alter() for system_logging_settings().
  */
 function dblog_form_system_logging_settings_alter(&$form, $form_state) {
+  require_once(BACKDROP_ROOT . '/core/includes/date.inc');
```
indigoxela, line 154: Same as above: core/includes/date.inc should already be included in function _backdrop_bootstrap_full().

core/modules/date/date.admin.inc
quicksketch: Yep, `date.inc` is always included on full bootstrap. No need to include it here.

quicksketch (commented):
This approach seems fine to me with a few issues:

- Incorporate the feedback from @indigoxela and @klonos, minor cleanup.
- I think we need an update hook if we're going to change `dblog_date_format` to `dblog`. So rather than modifying `dblog_update_1002()`, make a `dblog_update_1003()` and rename the format there.

quicksketch:
This will need refactoring after reverting the original feature. We have a new issue at https://github.com/backdrop/backdrop-issues/issues/4621 to re-introduce this.

PR #3305, opened 2020-09-13
Issue #4618: Show human-readable project name on 'Install dependencies'

### LINKED ISSUE #4618: [UX] Project Installer: Machine name being used instead of human-readable name in the "Install Dependencies" step
I accidentally noticed this while replying to a comment in https://github.com/backdrop/backdrop-issues/issues/2200#issuecomment-691591318

Take installing webform_validate for example... After adding the module to the installation queue and clicking "Install", you're taken to the page where you are given the chance to select a version other than the latest. Notice that the module is listed using its human-readable name 👍 :

image

After that step, you are taken to the "Install Dependencies" step. Notice that the Webform module is listed with its machine name instead 👎 :

image

PS: while at it, the word "dependencies" in the page title should not be capitalized 👎

klonos:
Fixes https://github.com/backdrop/backdrop-issues/issues/4618

### DIFF
--- a/core/modules/installer/installer.pages.inc
+++ b/core/modules/installer/installer.pages.inc
@@ -175,7 +175,7 @@ function installer_browser_installation_page($op) {
       $content = installer_browser_installation_select_versions_page();
       break;
     case 'install_dependencies':
-      backdrop_set_title(t("Install Dependencies"));
+      backdrop_set_title(t("Install dependencies"));
       $content = installer_browser_installation_install_dependencies_page();
       break;
     case 'theme':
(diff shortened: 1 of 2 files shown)


core/modules/installer/installer.pages.inc
```diff
@@ -171,19 +171,19 @@ function installer_browser_installation_page($op) {
...
       $content = installer_browser_installation_install_dependencies_page();
       break;
     case 'theme':
-      backdrop_set_title(t("Set themes"));
+      backdrop_set_title(t('Set themes'));
       $content = installer_browser_installation_enable_page('theme');
       break;
     case 'enable':
-      backdrop_set_title(t("Enable modules"));
+      backdrop_set_title(t('Enable modules'));
```
klonos, line 186: ...and before anyone asks, https://api.backdropcms.org/php-standards#quotes

> ... single quote strings are known to be faster because the parser doesn't have to look for in-line variables. Their use is recommended except in two cases: ...

PR #3314, opened 2020-09-18
Issue #2398: Use the Layout context info to load a complete User entity for the current user context.

### LINKED ISSUE #2398: The 'current_user' context in a Layout does not contain a User entity

```
```

\Layout::getContexts
```
```
-------

bradbulger:
Fixes backdrop/backdrop-issues#2398  - check that the current_user context entity implements EntityInterface, and create a complete user entity for that context.

### DIFF
--- a/core/modules/layout/layout.module
+++ b/core/modules/layout/layout.module
@@ -1837,7 +1837,6 @@ function layout_current_user_context() {
   }
 
   $current_user = $load_function($user->uid);
-
   $current_user_context->setData($current_user);
 
   return $current_user_context;
--- a/core/modules/layout/layout.module
+++ b/core/modules/layout/layout.module
@@ -1824,18 +1824,19 @@ function layout_current_user_context() {
     'locked' => TRUE,
   ));
 
-  // Get a complete user entity
+  // Get a complete user entity.
+
   $context_info = layout_get_context_info('user');
+
+  // Check if a load function has been set by some other module. If not, default to user_load().
   if (isset($context_info['load callback']) && function_exists($context_info['load callback'])) {
-    // in case this has been set by other modules
     $load_function = $context_info['load callback'];
   }
   else {
-    // fall back to default
     $load_function = 'user_load';
   }
-  $current_user = $load_function($user->uid);
 
+  $current_user = $load_function($user->uid);
   $current_user_context->setData($current_user);
 
   return $current_user_context;
--- a/core/modules/layout/layout.module
+++ b/core/modules/layout/layout.module
@@ -1813,6 +1813,35 @@ function layout_context_required_by_path($path) {
 }
 
 /**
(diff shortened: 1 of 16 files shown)


core/modules/layout/layout.module
```diff
@@ -1812,6 +1812,35 @@ function layout_context_required_by_path($path) {
...
+function layout_current_user_context() {
+  global $user;
+
+  $current_user_context = layout_create_context('user', array(
+    'name' => 'current_user',
+    'label' => t('Current user'),
+    'locked' => TRUE,
+  ));
+
+  // Get a complete user entity
```
klonos:
```suggestion
  // Get a complete user entity.
```

core/modules/layout/layout.module
```diff
@@ -1812,6 +1812,35 @@ function layout_context_required_by_path($path) {
...
+  $context_info = layout_get_context_info('user');
+  if (isset($context_info['load callback']) && function_exists($context_info['load callback'])) {
+    // in case this has been set by other modules
+    $load_function = $context_info['load callback'];
+  }
+  else {
+    // fall back to default
```
klonos, line 1834:
```suggestion
    // Fall back to default.
```

core/modules/layout/layout.module
```diff
@@ -1812,6 +1812,35 @@ function layout_context_required_by_path($path) {
...
+  $current_user_context = layout_create_context('user', array(
+    'name' => 'current_user',
+    'label' => t('Current user'),
+    'locked' => TRUE,
+  ));
+
+  // Get a complete user entity
+  $context_info = layout_get_context_info('user');
+  if (isset($context_info['load callback']) && function_exists($context_info['load callback'])) {
+    // in case this has been set by other modules
```
klonos, line 1830:
```suggestion
    // In case this has been set by other modules.
```

klonos: Coding standards (https://api.backdropcms.org/php-standards#comments) + move the comment outside the conditional:

```suggestion
  // In case this has been set by other modules.
  if (isset($context_info['load callback']) && function_exists($context_info['load callback'])) {
```

core/modules/layout/layout.module
klonos: Same as previous comment:

```suggestion
  // Fall back to default.
  else {
```

core/modules/layout/layout.module
bradbulger: Done.

core/modules/layout/layout.module
bradbulger: Rewrote comment for both conditions.

core/modules/layout/layout.module
bradbulger: Rewrote comment for both conditions.

PR #3317, opened 2020-09-21
Issue #4632: date fields should respect localized display formats

### LINKED ISSUE #4632: Date fields never show in localized display formats, always in default formats
**Description of the bug**

It is possible to localize date formats so they show up according to the current language (admin/config/regional/date-time).
That works fine for the content overview page (for instance), but this is currently broken for date fields.

**Steps To Reproduce**

1. Enable locale and language module
2. Add a language (no translations needed), set up detection and selection to use path prefix (admin/config/regional/language/detection)
3. Add a date field to a content type
4. Localize your date formats (admin/config/regional/date-time)
5. Create content of that type with a date
6. Visit that node
7. Switch between languages

**Actual behavior**

The configured date display formats per language are ignored. You always only see the default (fallback) format.

**Expected behavior**

Switching between languages should also change the date display format - just like it does in D7.

**Additional information**

This happens in date_format_type_format(), which changed a bit.
Compare the [Backdrop version](https://api.backdropcms.org/api/backdrop/core%21modules%21date%21date.module/function/date_format_type_format/1) with the [Drupal 7 version](https://git.drupalcode.org/project/date/-/blob/7.x-2.x/date.module#L473).

Fun fact: the D7 version seems to be a bit awkward anyway, but works... coincidentally.

If it's unclear what I mean with "switching date display formats", visit the content overview page (admin/content) after localizing your formats - there it works correctly.

indigoxela:
Fixes https://github.com/backdrop/backdrop-issues/issues/4632

### DIFF
--- a/core/modules/date/date.module
+++ b/core/modules/date/date.module
@@ -529,22 +529,30 @@ function date_formatter_format($formatter, $settings, $granularity = array(), $l
 }
 
 /**
- * Helper function to get the right format for a format type.
+ * Helper function to get the right date format for a format type.
  *
- * Checks for locale-based format first.
+ * @param string $format_type
+ *   Name of the date format, for instance "short", "medium" or "long".
+ * @param string $langcode
+ *   Language code provided, for instance "en".
+ *
+ * @return string
+ *   PHP date format pattern for selected or current language, default pattern
+ *   if no localized variant exists.
  */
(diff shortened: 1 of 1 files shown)


indigoxela:
Requesting fresh sandbox.

ghost:

PR #3318, opened 2020-09-22
Issue #4633: Remove duplicated documentation from hook_element_info().

### LINKED ISSUE #4633: hook_element_info() has duplicated documentation
The documentation for `hook_element_info()` starts like this:

> Allows modules to declare their own Form API element types and specify their default values.
>
> This hook allows modules to declare their own form element types and to specify their default values. The values returned by this hook will be merged with the elements returned by hook_form() implementations and so can return defaults for any Form APIs keys in addition to those explicitly mentioned below.

The first two sentances are basically duplicates. We should remove one.

ghost:
Fixes https://github.com/backdrop/backdrop-issues/issues/4633

### DIFF
--- a/core/modules/system/system.api.php
+++ b/core/modules/system/system.api.php
@@ -203,11 +203,9 @@ function callback_queue_worker($queue_item_data) {
  * Allows modules to declare their own Form API element types and specify their
  * default values.
  *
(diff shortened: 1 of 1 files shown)


ghost:
Merged into 1.x and 1.17.x.

PR #3320, opened 2020-09-23
Issue #4207: Add classes to menu local tasks.

### LINKED ISSUE #4207: Can we add a class to the admin tabs?
**Description of the need**
I was looking at styling the tabs slightly differently and realized they don't have any identifying classes (e.g. the "Edit" tab does not have an "edit" class) so are hard to pinpoint precisely with CSS. Can we add a simple class to each tab?

jenlampton:
Fixes https://github.com/backdrop/backdrop-issues/issues/4207

### DIFF
--- a/core/includes/menu.inc
+++ b/core/includes/menu.inc
@@ -2116,6 +2116,15 @@ function menu_local_tasks($level = 0) {
         }
         if ($item['access']) {
           $link = $item;
+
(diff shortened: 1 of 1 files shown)


laryn:
@jenlampton Can you close and reopen to try to get a sandbox that is working? Then I'll do some testing.

PR #3327, opened 2020-09-24
Issue #4639: Fix PHP file comments longer than 80 characters.

### LINKED ISSUE #4639: Fix comments longer than 80 characters
As per our [documentation standards](https://api.backdropcms.org/doc-standards#general), comments should wrap at or before 80 characters. However there are quite a few places in core where comments run over the 80-character limit. This issue aims to fix that.

ghost:
Fixes https://github.com/backdrop/backdrop-issues/issues/4639


core/modules/simpletest/backdrop_web_test_case.php
```diff
@@ -3886,7 +3912,8 @@ protected function assertNoResponse($code, $message = '') {
    * The field in $name must have the content described in $value.
    *
    * @param $name
-   *   Name of field or message property to assert. Examples: subject, body, id, ...
+   *   Name of field or message property to assert. Examples: subject, body, id,
+   *   ...
```
klonos: We could either remove the `...`, or replace it with `etc.` here.

core/modules/system/system.api.php
```diff
@@ -580,7 +587,8 @@ function hook_menu_get_item_alter(&$router_item, $path, $original_map) {
  * load and return an "abc" object with internal id 123:
  * @code
  *   function mymodule_abc_load($abc_id) {
- *     return db_query("SELECT * FROM {mymodule_abc} WHERE abc_id = :abc_id", array(':abc_id' => $abc_id))->fetchObject();
+ *     return db_query("SELECT * FROM {mymodule_abc} WHERE abc_id = :abc_id",
```
klonos: This should either be one line, or something like:

```
 *   function mymodule_abc_load($abc_id) {
 *     return db_query(
 *       "SELECT * FROM {mymodule_abc} WHERE abc_id = :abc_id",
 *       array(':abc_id' => $abc_id)
 *     )->fetchObject();
 *   }
```

ghost: Thanks, I remember doing that before I discovered other places with `@code` blocks (and realising they shouldn't be wrapped), but then couldn't remember where this was to go back and undo it.

gitressa:
Thanks! Maybe update `For backdrop/backdrop-issues/issues/4639` in first comment to `Fixes *issue_URL*` to link them up?

PR #3328, opened 2020-09-24
Issue #4313: Display message when dashboard block shows no 'available updates'.

### LINKED ISSUE #4313: Make the "Available updates" in Dashboard little bit more verbose when there are no updates
**Description of the need**

When there are no updates the "Available updates" block on Dashboard  looks like:

Screen Shot 2020-02-18 at 8 11 35 PM

which new users could find confusing.

**Proposed solution**

I believe when there are no updates the "Available updates" block on Dashboard should either be automatically hide itself or contain some information. For example, depending on what kind of  information user choose to display on the block's settings, it could display something like:

> Congratulations! Your website is up to date as no more updates are found for Backdrop core, modules, themes or layouts.

This feature request is a small one, that could improve user interface's usability a bit.

ghost:
Fixes https://github.com/backdrop/backdrop-issues/issues/4313

### DIFF
--- a/core/modules/dashboard/includes/block.update.inc
+++ b/core/modules/dashboard/includes/block.update.inc
@@ -93,7 +93,7 @@ class DashboardUpdateBlock extends Block {
         }
         else {
           // No updates available.
-          $item .= ' - ' . t('No recommended updates');
+          continue;
         }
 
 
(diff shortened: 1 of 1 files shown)


core/modules/dashboard/includes/block.update.inc
```diff
@@ -104,10 +104,17 @@ class DashboardUpdateBlock extends Block {
...
-    );
+    if (!empty($items)) {
+      $panel['list'] = array(
+        '#theme' => 'item_list',
+        '#items' => $items,
+      );
+    }
+    else {
+      $panel['empty'] = array(
+        '#markup' => '<p class="dashboard-no-updates">' . t('There are no available updates. Your Backdrop site is up-to-date!') . '</p>',
```
klonos, line 115: Why do we need a class added here? ...in fact, why a `<p>`?

```suggestion
        '#markup' => t('There are no available updates. Your Backdrop site is up-to-date!'),
```

ghost: I don't know, I just thought better to have something rather than nothing, and a class for styling. But I'm not a themer so maybe not...

PR #3334, opened 2020-09-29
Issue #4659: Fix format_date() test using incorrect language.

bradbulger:
Fixes issue #4659 - replace "Monday" with "lundi"


bradbulger:
Unable to reproduce original problem.

PR #3336, opened 2020-09-30
Fixes https://github.com/backdrop/backdrop-issues/issues/4664 and htt…

### LINKED ISSUE #4664: Mixed use of $config->get and config_get in the same functions
**Description of the bug**

From https://api.backdropcms.org/change-records/variableget-variableset-and-variabledel-deprecated-cmi we know that:

> Every time you call config_get() or config_set(), it hits the disk to read/write the file. If you're doing a lot of reads or writes, it's better to do all the reads at once and all the writes at once like this:
> 
> ```
> $config = config('mymodule.settings');
> $value = $config->get('foo');
> ```

Unfortunately, core still have some functions which have multiple occurrences of `config_get` and sometimes mixed use of both `$config->get` and `config_get`. Take, for example, the [system_site_information_settings()](https://github.com/backdrop/backdrop/blob/f12660fd62fbdbd507bdf8b89deec086d4db9b6d/core/modules/system/system.admin.inc#L1200) function. 

It declares `$site_config = config('system.core');` in [the very beginning of the function](https://github.com/backdrop/backdrop/blob/f12660fd62fbdbd507bdf8b89deec086d4db9b6d/core/modules/system/system.admin.inc#L1201), so the content of `$site_config` could be re-used as expected in such lines as `'#default_value' => $site_config->get('site_name')`, `'#default_value' => $site_config->get('site_logo_theme')`, etc. 

However, the same function unnecessarily calls `config_get` multiple times at lines:

https://github.com/backdrop/backdrop/blob/f12660fd62fbdbd507bdf8b89deec086d4db9b6d/core/modules/system/system.admin.inc#L1306
https://github.com/backdrop/backdrop/blob/f12660fd62fbdbd507bdf8b89deec086d4db9b6d/core/modules/system/system.admin.inc#L1326
https://github.com/backdrop/backdrop/blob/f12660fd62fbdbd507bdf8b89deec086d4db9b6d/core/modules/system/system.admin.inc#L1335
https://github.com/backdrop/backdrop/blob/f12660fd62fbdbd507bdf8b89deec086d4db9b6d/core/modules/system/system.admin.inc#L1358

**Expected behavior**

To reduce the overhead, we need to use one read per function instead of multiple reads. 

I haven't analyzed other files, so there could be some other cases of mixed use of two methods in the same core functions. The above case is just what I've noticed in the `system.admin.inc` while working on something else, so I've decided to create a standalone issue for this particular file (system.admin.inc) only.

### LINKED ISSUE #4665: Core contains link to drupal.org whereas there is a respective documentation page for Backdrop
**Description of the bug**

If you go to `/admin/config/media/file-system` of your Backdrop site and click on the `more information about securing private files.` link then you'll be re-directed to https://www.drupal.org/docs/7/core/modules/file/overview. That's because  the [line 1807](https://github.com/backdrop/backdrop/blob/f12660fd62fbdbd507bdf8b89deec086d4db9b6d/core/modules/system/system.admin.inc#L1807) of system.admin.inc looks like:

`    '#description' => t('An existing local file system path for storing private files. It should be writable by Backdrop and not accessible over the web. See the online handbook for <a href="@handbook">more information about securing private files</a>.', array('@handbook' => 'https://www.drupal.org/docs/7/core/modules/file/overview')),`

However, we already have the same documentation page for Backdrop at https://backdropcms.org/user-guide/file-system, so it would be really nice to cleaning up such links to Drupal.org if we do have the respective documentation pages for Backdrop.

alanmels:
Fixes backdrop/backdrop-issues#4664 
Fixes backdrop/backdrop-issues#4665

I've replaced multiple occurrences of `config_get` in the same functions to `$config->get` leaving intact the single uses per function. Also I've replaced the link to drupal.org to respective documentation page on backdropcms.org. So this PR should take care of both https://github.com/backdrop/backdrop-issues/issues/4664 and https://github.com/backdrop/backdrop-issues/issues/4665

### DIFF
--- a/core/modules/system/system.admin.inc
+++ b/core/modules/system/system.admin.inc
@@ -1303,7 +1303,7 @@ function system_site_information_settings($form, &$form_state) {
     '#default_value' => ($site_config->get('site_frontpage') != 'user' ? backdrop_get_path_alias($site_config->get('site_frontpage')) : ''),
     '#size' => 40,
     '#description' => t('Optionally, specify a relative URL to display as the home page. Leave this blank to display the default content feed.'),
-    '#field_prefix' => url(NULL, array('absolute' => TRUE)) . (config_get('system.core', 'clean_url') ? '' : '?q='),
+    '#field_prefix' => url(NULL, array('absolute' => TRUE)) . ($site_config->get('clean_url') ? '' : '?q='),
     '#autocomplete_path' => 'path-autocomplete',
   );
   $form['front_page']['default_nodes_main'] = array(
@@ -1311,7 +1311,7 @@ function system_site_information_settings($form, &$form_state) {
     '#default_value' => $site_config->get('default_nodes_main'),
     '#options' => backdrop_map_assoc(array(1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 15, 20, 25, 30)),
     '#description' => t('The maximum number of posts displayed on overview pages such as the home page.'),
-    '#access' => (config_get('system.core', 'site_frontpage') == 'node'),
+    '#access' => ($site_config->get('site_frontpage') == 'node'),
   );
   $form['error_page'] = array(
     '#type' => 'fieldset',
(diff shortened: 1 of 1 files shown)


klonos (approved):
LGTM 👍

ghost:

PR #3341, opened 2020-10-01
Issue #3806: Fix docs in user_cancel_user_action()

### LINKED ISSUE #3806: Fix inline docs in `user_cancel_user_action()`
The inline docs in `user_cancel_user_action()` refer to nodes instead of users. These need to be updated to reflect what actually happens to user accounts in this function.

```php
/**
 * Queues a specific user account for cancellation.
 *
 * @param User $account
 *   The user entity whose account will be blocked.
 *
 * @ingroup actions
 */
function user_cancel_user_action($account, &$context) {
  // Save the list of nodes to be deleted in the session. Append to the existing
  // list if within the last minute, otherwise start a new list of nodes.

```

- - - - - - 

PR by @chuyich https://github.com/backdrop/backdrop/pull/3341

chuyich:
Fixes https://github.com/backdrop/backdrop-issues/issues/3806

### DIFF
--- a/core/modules/user/user.actions.inc
+++ b/core/modules/user/user.actions.inc
@@ -55,8 +55,9 @@ function user_unblock_user_action($account, &$context) {
  * @ingroup actions
  */
 function user_cancel_user_action($account, &$context) {
-  // Save the list of user accounts to be deleted in the session. Append to the existing
-  // list if within the last minute, otherwise start a new list of user accounts.
+  // Save the list of user accounts to be deleted in the session. Append to 
+  // the existing list if within the last minute, otherwise start a new list
+  // of user accounts.
   $last_action_time = 0;
   if (isset($_SESSION['user_cancel_action'])) {
     $last_action_time = $_SESSION['user_cancel_action']['timestamp'];
--- a/core/modules/user/user.actions.inc
+++ b/core/modules/user/user.actions.inc
@@ -55,8 +55,8 @@ function user_unblock_user_action($account, &$context) {
  * @ingroup actions
  */
 function user_cancel_user_action($account, &$context) {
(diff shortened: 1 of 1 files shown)


jenlampton:
This is a great start!

> // Save the list of users to be deleted

I think this should be `// Save the list of user accounts to be deleted` since we aren't actually deleting the people :)  The same goes for `otherwise start a new list of users`

core/modules/user/user.actions.inc
```diff
@@ -55,8 +55,8 @@ function user_unblock_user_action($account, &$context) {
  * @ingroup actions
  */
 function user_cancel_user_action($account, &$context) {
-  // Save the list of nodes to be deleted in the session. Append to the existing
-  // list if within the last minute, otherwise start a new list of nodes.
+  // Save the list of user accounts to be deleted in the session. Append to the existing
+  // list if within the last minute, otherwise start a new list of user accounts.
```
ghost: This needs to adhere to the 80-character limit for comments (https://api.backdropcms.org/doc-standards#general):

```suggestion
  // Save the list of user accounts to be deleted in the session. Append to the
  // existing list if within the last minute, otherwise start a new list of user
  // accounts.
```

ghost (changes requested):

core/modules/user/user.actions.inc
chuyich: Fixed on next commit.

ghost: Thanks!

ghost (approved):

ghost:
Thanks @chuyich! Merged into 1.x and 1.17.x.

PR #3347, opened 2020-10-03
Issue #1132: Don't process projects in the wrong directory.

### LINKED ISSUE #1132: Only process project (module/theme/layout) if the type in its .info file matches the folder it's placed in.
So I wanted to test a few PRs that have been made lately against the layout UI. Wanted to install Radix in order to get more layouts than the few we ship with core (especially in order to test wrapping in #988). So I googled "backdrop radix" and downloaded what I thought was [Radix the layout](https://github.com/backdrop-contrib/radix_layouts) when in reality I had just downloaded [Radix the theme](https://github.com/backdrop-contrib/radix).

...here's what I saw in the layout UI:

backdrop-contrib_layouts-radix-no_screenshot-name_issue

I was about to file an issue in the respective contrib project queue, when I realized what was wrong with the whole thing. So, can we please have it so that if say like in this case, a theme is accidentally placed under `/layouts`, then we do not parse the thing beyond the .info if we detect that `type = theme` or `type = module`?

...of course we should have similar checks in place so that modules or layouts accidentally placed under `/themes` are not parsed and so that layouts and themes placed under `/modules` are neither.

ghost:
Fixes https://github.com/backdrop/backdrop-issues/issues/1132


quicksketch:
Nice fix! I merged this into 1.x and 1.26.x. Thanks @BWPanda!

PR #3355, opened 2020-10-05
Update bg.php

amilenkov:
The transliteration in Bulgarian has been updated according to the official transliteration in Bulgaria, as shown here:
https://en.wikipedia.org/wiki/Bulgarian_alphabet

and in Bulgarian here:
https://www.lex.bg/bg/laws/ldoc/2135623667

Only for one letter (uppercase and lowercase) - lines 29 and 59 in this code

0x0426 => 'C',
0x0446 => 'c',

transliteration was made according to an older standard from 1995 ISO9

https://en.wikipedia.org/wiki/ISO_9

because it is practically more convenient (one letter C instead of two TS in latest legislation), this way is widely used in this by Bulgarians elsewhere and is easily recognized by search engines. 

The main concern of the legislator in creating the last transliteration TS for Ц was the transliteration of street signs and geographical names, not search engine optimized addresses on web pages - they were looking for convenience for foreigners when reading transliterated Bulgarian titles of geographical objects and there really TS for Bulgarian Ц is more appropriate because many foreigners would read C as K as in CocaCola, for example. For Bulgarians, C is naturally recognizable as Ц for historical reasons - many borrowings from old Latin language, for example, the English word Center is pronounced in Bulgarian as Център and the first three letters of the Latin alphabet ABC are pronounced in Bulgarian as a, be Це.

Finally, when reviewing the code more carefully, I found two more lines that transliterate letters from the Russian language, but are not used in Bulgarian - these are the symbols for a capital letter and a regular letter for ы - there is no such letter in the Bulgarian alphabet. 

The reason was probably that the Bulgarian and Russian alphabets largely coincide with only three differences - in the Bulgarian language there are no letters Э, э and Ы, ы while in the Russian language there is no letters Ъ, ъ. And probably Russian transliteration was originally used in the creation of the Bulgarian one.


quicksketch:
Thanks @amilenkov! Merged into 1.x and 1.17.x.

PR #3366, opened 2020-10-11
Issue #4694: Re-set 'Continue & configure' as default Views UI button.

### LINKED ISSUE #4694: Views UI: "Save & exit" button accidentally made primary instead of the "Continue & configure" button
@BWPanda [asked in Zulip](https://backdrop.zulipchat.com/#narrow/stream/218635-Backdrop/topic/Views/near/212958631):

> Can anyone point me to the issue where it was decided to make the Views' 'Save & Exit' button the default, instead of 'Save & Continue'? It changed sometime since 2015: https://github.com/backdrop/backdrop-issues/issues/999#issue-87900725

@BWPanda that seems like a bug to me. With the change that came with the PR for #882, what happens in a series of `'#type' => 'actions'` buttons is that the submit button that comes first in-code always gets the `.button-primary` class. The PR for that issue specifically moved the "Save & exit" button to come after the "Continue & configure" (see [here](https://github.com/backdrop/backdrop/pull/875/files?file-filters%5B%5D=.css&file-filters%5B%5D=.inc#diff-8e6c268efd723011d5ecd609622469ecL431). ...the order of the 2 buttons in-code [remains unchanged till today](https://github.com/backdrop/backdrop/blob/1.x/core/modules/views_ui/views_ui.admin.inc#L469), so **in theory** the "Continue & configure" button should still be the primary one.

I think that the culprit is [this line](https://github.com/backdrop/backdrop/commit/bae764ee21840cc510db68e3bd5c4cd3fe970ff7#diff-8e6c268efd723011d5ecd609622469ecR437), which came as a result of https://github.com/backdrop/backdrop/commit/2823cc0e6ca96df6d72939f8587d9b0fd1f2f7d3 (back in 2011), but the solution implemented for that issue might not be accounting for the re-ordering of the 2 buttons (which for us has happened in #882).

ghost:
Fixes https://github.com/backdrop/backdrop-issues/issues/4694

### DIFF
--- a/core/modules/views_ui/views_ui.admin.inc
+++ b/core/modules/views_ui/views_ui.admin.inc
@@ -472,6 +472,7 @@ function views_ui_add_form($form, &$form_state) {
     '#validate' => array('views_ui_wizard_form_validate'),
     '#submit' => array('views_ui_add_form_store_edit_submit'),
     '#process' => array_merge(array('views_ui_default_button'), element_info_property('submit', '#process', array())),
+    '#weight' => -1,
   );
   $form['actions']['save'] = array(
     '#type' => 'submit',


quicksketch:

PR #3367, opened 2020-10-12
Issue #4695: Make pager ellipsis visible in Basis.

### LINKED ISSUE #4695: Basis pager ellipsis colour
**Description of the bug**

The colour of the ellipsis in Basis' pager is the same as the pager background colour and so isn't visible:

image

**Steps To Reproduce**

To reproduce the behavior:
1. Use Devel Generate to create lots of Post nodes
2. Scroll to the bottom of the homepage to see the pager

**Actual behavior**

Ellipsis is present, but the same colour as the background and so not actually visible.

**Expected behavior**

Ellipsis is coloured white, like the other pager links.

ghost:
Fixes https://github.com/backdrop/backdrop-issues/issues/4695

### DIFF
--- a/core/themes/basis/css/skin.css
+++ b/core/themes/basis/css/skin.css
@@ -756,10 +756,6 @@ ul.action-links a {
   color: #ffffff;
 }
 
-.pager-ellipsis {
-  color: #586172;
-}
-
 .pager-first a,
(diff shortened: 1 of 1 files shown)


quicksketch:

PR #3375, opened 2020-10-14
Issue #4286: Remove 'Spock' from Backdrop.

### LINKED ISSUE #4286: Remove 'Spock' from Backdrop.
I'm looking at the Dashboard welcome block properly for the first time and I have some thoughts:

- Change the block icon
  Don't know why we have the Star Trek/Spock/Live-long-and-prosper icon there (apart from maybe we were feeling geeky at the time), but I think this should be changed to look more professional; something like https://fontawesome.com/icons/comment?style=solid, https://fontawesome.com/icons/handshake?style=solid, https://fontawesome.com/icons/smile?style=solid, or just the Backdrop mark logo.
- ~~Reword 'modules' to 'functionality'~~ Others don't seem to mind this wording, so leaving for now...
  'Functionality' is the term we use in the top-level of the admin bar (I know the term 'module' is used in the sub-links, but that's not what people see initially, and it's a bit of a Drupal-ism). We should use something like this wording instead for the first two items in the 'More actions' column respectively: 'Toggle existing functionality' (or 'Turn existing functionality on or off' if we don't like the word 'toggle'), 'Add new functionality'.

Current block:

Dashboard___Luke

ghost:
Fixes https://github.com/backdrop/backdrop-issues/issues/4286


klonos (approved):
LGTM, and the test failures are the known random ones.

ghost:
Merged into 1.x & 1.17.x.

PR #285, opened 2014-05-27
Issue #1634190 by LoMo: Fixed Long month names in locale.datepicker.js not using context properly.

Twinmat:
PR for https://drupal.org/node/1634190

PR #3379, opened 2020-10-19
Issue #285: Introduce Telemetry module for tracking usage information.

quicksketch:
Work in progress PR to introduce Telemetry support in core.

### DIFF
--- a/core/modules/telemetry/telemetry.admin.inc
+++ b/core/modules/telemetry/telemetry.admin.inc
@@ -1,0 +1,55 @@
+<?php
+/**
+ * Administrative pages for Telemetry module.
+ */
+
+function telemetry_report() {
+  $telemetry_info = telemetry_get_info();
+  $telemetry_data = telemetry_get_data();
+
+  $page = array();
+
+  $help = '';
+  $help .= '<p>' . t('Help us improve Backdrop CMS by sending anonymous, untracked information about what features your site uses.') . '</p>';
+  $help .= '<p>' . t('Below is a list of all data collected by the enabled modules on your site. Only projects that are publicly available on BackdropCMS.org will store usage information.') . '</p>';
+  $help .= '<p>' . t('If you prefer to not send usage information, you can disable the Telemetry module from the <a href="!modules">modules page</a>.', array('!modules' => url('admin/modules'))) . '</p>';
+
+  $page['help'] = array(
+    '#type' => 'help',
+    '#markup' => $help
+  );
+
+  $header = array(
+    t('Module'),
+    t('Description'),
+    t('Value'),
+  );
+
+  $rows = array();
+  foreach ($telemetry_info as $project => $project_data) {
+    foreach ($project_data as $key => $info) {
+      $row = [];
+      if (!isset($telemetry_data[$project][$key])) {
+        $value = '<span class="error">Missing value</span>';
+      }
+      else {
+        $value = check_plain($telemetry_data[$project][$key]);
+      }
+      $module_info = system_get_info('module', $info['module']);
+      $project_name = $info['project'] === 'backdrop' ? 'core' : check_plain($info['project']);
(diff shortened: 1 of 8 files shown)


core/modules/telemetry/telemetry.info
```diff
@@ -0,0 +1,8 @@
+name = Telemetry
+description = Collects anonymous usage information to help improve Backdrop CMS and contributed modules.
+package = System
+version = BACKDROP_VERSION
+type = module
+backdrop = 1.x
+
+configure = admin/config/system/telemetry
```
klonos, line 8: This links to `admin/config/system` in the PR sandbox.

core/modules/telemetry/telemetry.module
```diff
@@ -0,0 +1,139 @@
...
+/**
+ * @file
+ * Collects usage information to help improve Backdrop CMS.
+ */
+
+/**
+ * Implements hook_menu().
+ */
+function telemetry_menu() {
+  $items['admin/reports/telemetry'] = array(
```
klonos, line 11: This 404's in the PR sandbox.

core/modules/telemetry/telemetry.admin.inc
```diff
@@ -0,0 +1,63 @@
...
+      $row[] = array(
+        'data' => $value,
+        'class' => array('telemetry-item-value'),
+      );
+      $rows[] = $row;
+    }
+  }
+
+  $page['items'] = array(
+    '#theme' => 'table',
```
docwilmot, line 56: Should we make this tableselect, and allow users to choose? Once contrib starts implementing this, could be users may prefer a choice.

core/modules/telemetry/telemetry.module
herbdool: This works at this point (tested in Tugboat):

Screenshot from 2021-08-25 09-37-48

core/modules/telemetry/telemetry.module
```diff
@@ -0,0 +1,140 @@
...
+ */
+function _telemetry_destination_server_name() {
+  return parse_url(config_get('telemetry.settings', 'destination'), PHP_URL_HOST);
+}
+
+/**
+ * Get an anonymous, unique identifier for this site.
+ */
+function _telemetry_site_key() {
+  // The $base_url is not used as the $data here intentionally, as different
```
herbdool, line 136: Can this be clarified? `$data` here is 'telemetry_site_key' correct?

core/modules/telemetry/telemetry.admin.inc
herbdool: @docwilmot I think that would be a good idea for a follow-up. For now with only three items showing, we probably are okay.

core/modules/telemetry/telemetry.info
herbdool: I recommend removing the `configure =` line since there is no configuration at this point.

herbdool (changes requested):
Mostly looks great. Only a couple small changes, plus remove some debugging lines.

core/modules/telemetry/telemetry.admin.inc
quicksketch: My original implementation actually did have this as tableselect so that you could turn on and off specific keys. However the discussion in the meta issue took the direction of having all telemetry data be all-or-nothing. In truth that made things _much_ simpler, because determining how to add/request approval for new keys as modules update adds a whole new problem. Also the ability to turn off every key was a bit overwhelming. Most other platforms that collect telemetry data seem to be all-or-nothing.

core/modules/telemetry/telemetry.info
quicksketch: Yep, sorry this was from a previous iteration where you could select what telemetry data to send.

quicksketch:
Replaced by https://github.com/backdrop/backdrop/pull/3704.

PR #3388, opened 2020-11-04
Issue #4735: Impossible to edit automatically generated comment titles

### LINKED ISSUE #4735: Impossible to edit automatically generated comment titles
**Description of the bug**

Admins can not edit comment titles in case if `Allow a custom title to be set for each comment` is not checked. This becomes problematic when you intentionally do not want users to enter titles, but lot's of misspelled titles started to show up in the submitted comments or in the _Recent comments_ block as shown below:

Screen Shot 2020-11-03 at 8 34 55 PM

**Steps To Reproduce**

To reproduce the behavior:
1. Go to `/admin/structure/types/manage/post` and make sure the `Allow a custom title to be set for each comment` is not checked.
2. Go to any `post` content, for example, `/posts/your-first-post`.
3. Add a comment which starts with misspelled words and the system automatically generates the comment title, which you can not edit even if you are admin.

**Workaround**

You can always turn on the `Allow a custom title to be set for each comment` option, go back to the post page, find the comment and edit it's title.

**Expected behavior**

Workaround above is not the ideal solution to the problem, because if intention is not to allow users to enter comment titles, then every time you notice incorrect comment title you have to go to `/admin/structure/types/manage/post` switch the `Allow a custom title to be set for each comment`  option on, edit comment title, go back to the setting page and turn the option off. 

I believe admins should ALWAYS be able to edit comment titles regardless if users are allowed or not to enter comment titles.

alanmels:
Fixes https://github.com/backdrop/backdrop-issues/issues/4735

### DIFF
--- a/core/modules/comment/comment.module
+++ b/core/modules/comment/comment.module
@@ -1935,14 +1935,25 @@ function comment_form($form, &$form_state, $comment) {
     '#access' => $is_admin,
   );
 
-  $form['subject'] = array(
+  $form['author']['subject'] = array(
     '#type' => 'textfield',
     '#title' => t('Title'),
     '#maxlength' => 64,
     '#default_value' => $comment->subject,
-    '#access' => $node_type->settings['comment_subject_field'],
-    '#weight' => -1,
+    '#access' => $is_admin,
+    '#weight' => 0,
   );
+  
+  if (!$is_admin && $node_type->settings['comment_subject_field']) {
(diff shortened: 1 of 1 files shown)


ghost:
**PR sandbox**

URL: https://pr3388-x9jbsfj0zzttofuwhtesejkk2ratkgyz.tugboat.qa/
Username: admin
Password: 8391eaa09db3

PR #3392, opened 2020-11-07
Issue #3792: Workaround for php floating point precision problem

### LINKED ISSUE #3792: Simple decimals fail to pass validation
**Description of the bug**
I am running into a problem where a number/decimal field is not allowing values that should pass the validation.

It is possibly related to [this Drupal 8 issue](https://www.drupal.org/project/drupal/issues/2230909)? Not certain but sounds similar, and the steps to reproduce listed here also reproduce the error on a Backdrop install: 

> **Problem/Motivation**
> 
> - Create decimal field, set precision to 10 (minimum in the UI and scale to 4
> - Saving new node with value 19999.0000 succeeds (precision is 5+4 = 9).
> - Saving new node with value 99999.0000 fails (same precision as above).
> 
> or
> 
> - Create decimal field, set scale to anything over 6 (need 8 to store bitcoin values for example). I used 16 as precission.
> - Saving new node with value 20.12345678 fails validation while 0.1234567 succeeds.
> 
> This is because Drupal\Component\Utility\Number::validStep() returns false. Detailed investigation on this function (which is only used once in Drupal), reveals that the first argument ($value) is received as a string, but $step and $offset are floats. PHP seems to slightly mangle the $step value on the first case above from 0.0001 to 0.00010000000000000001438. Passing the step parameter as a string in the case of decimal numbers maintains the correct precision, and allows a correct approximation calculation.
>
> **Proposed resolution**
> 
> Bypass weak PHP floating-point handling by passing the step as a string.
> 

**Update**

After some experiments with a much simpler, then the D8 approach, it turned out that neither of these works properly.

It needs a small change in function number_field_widget_form() and a complete rewrite of function valid_number_step().

Also note that several of the existing tests were wrong and needed switching, see [this comment](https://github.com/backdrop/backdrop-issues/issues/3792#issuecomment-724819508) for details and explanation.

Several additional automated tests have been added to verify the new approach.

---
There are currently two PR, which might be confusing.

- by @indigoxela https://github.com/backdrop/backdrop/pull/3392
- by @BWPanda outdated, just fixes a few false negatives

indigoxela:
Fixes https://github.com/backdrop/backdrop-issues/issues/3792

### DIFF
--- a/core/includes/common.inc
+++ b/core/includes/common.inc
@@ -1489,10 +1489,12 @@ function valid_number_step($value, $step, $offset = 0.0) {
   $step_scale = strpos(strrev($step_string), '.');
   $value_scale = strpos(strrev($value_string), '.');
   if ($value_scale > $step_scale) {
-    if ($value_scale > 10 && strlen($value_string) - 1 > ini_get('precision')) {
-      return TRUE;
+    if ($value_scale > 10 && $step_scale > 0 && strlen($value_string) - 1 > ini_get('precision')) {
+      $value = round($value, $step_scale);
     }
-    return FALSE;
+    else {
+      return FALSE;
+    }
   }
 
   // Because of the floating point precision problem, fmod() has a high error
--- a/core/includes/common.inc
+++ b/core/includes/common.inc
@@ -1450,32 +1450,45 @@ function valid_url($url, $absolute = FALSE) {
  *
  * @return bool
  *   TRUE if no step mismatch has occured, or FALSE otherwise.
- *
- * @see http://opensource.apple.com/source/WebCore/WebCore-1298/html/NumberInputType.cpp
  */
 function valid_number_step($value, $step, $offset = 0.0) {
-  $double_value = (double) abs($value - $offset);
+  // Consider offset and get the unsigned values.
+  $value = (float) abs($value - $offset);
+  $step = abs($step);
 
-  // The fractional part of a double has 53 bits. The greatest number that could
-  // be represented with that is 2^53. If the given value is even bigger than
-  // $step * 2^53, then dividing by $step will result in a very small remainder.
-  // Since that remainder can't even be represented with a single precision
-  // float the following computation of the remainder makes no sense and we can
-  // safely ignore it instead.
-  if ($double_value / pow(2.0, 53) > $step) {
-    return TRUE;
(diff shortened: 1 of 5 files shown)


core/includes/common.inc
```diff
@@ -1450,32 +1450,59 @@ function valid_url($url, $absolute = FALSE) {
...
+
+  // Both numbers are integers - use modulo right away.
+  if ((int) $value == $value && (int) $step == $step && $value < PHP_INT_MAX) {
+    return $value % $step == 0;
+  }
+
+  // Convert from possible exponent notation to decimal strings to get the scale
+  // values. Note: sprintf also rounds values but seems to give more reliable
+  // results than number_format.
+  // "10" is the maximum value allowed by the number widget, so higher scale
```
klonos, line 1475: > "10" is the maximum value allowed by the number widget,

I wonder why this is the case here 🤔

core/includes/common.inc
```diff
@@ -1450,32 +1450,59 @@ function valid_url($url, $absolute = FALSE) {
...
+  $max_scale = 13;
+  $step_string = rtrim(sprintf("%.{$max_scale}F", $step), '0');
+  $value_string = rtrim(sprintf("%.{$max_scale}F", $value), '0');
+
+  // If the actual value has more significant decimals than expected, it has a
+  // higher precision than desired and is not divisible by the step.
+  // But if the scale value high, and the length of string (minus comma) is
+  // beyond the php setting for precision, it's more likely that the conversion
+  // to string caused a problem with rounding. It's a compromise to allow it,
+  // because there's no reliably way for php to handle it.
```
klonos: Should either be: `no reliable way` or `no way for php to handle it reliably.`

core/includes/common.inc
indigoxela: It's in the [settings form](https://github.com/backdrop/backdrop/blob/c6a99af4723ec52cfb35151af762c3efab85f7b8/core/modules/field/modules/number/number.module#L59)

I couldn't tell why :wink: , but 10 is the max for this settings form item.

core/includes/common.inc
indigoxela: :+1: and I also saw that an "is" is missing above.

quicksketch:

PR #3394, opened 2020-11-08
Issue #2158: Deprecate cache_clear_all() function.

### LINKED ISSUE #2158: [DX] Mark cache_clear_all() as deprecated
We should use the new `backdrop_flush_all_caches()` wherever possible to utilize our shiny new cache system.  Let's replace all core usage of `cache_clear_all` and mark the function deprecated.  This will require a chance notice.

ghost:
Fixes https://github.com/backdrop/backdrop-issues/issues/2158


core/modules/entity/entity.controller.inc
```diff
@@ -219,7 +219,7 @@ class DefaultEntityController implements EntityControllerInterface {
     // Reset the persistent cache.
     if ($this->persistentCache) {
       if (!empty($ids)) {
-        cache_clear_all($ids, 'cache_entity_' . $this->entityType);
+        cache('cache_entity_' . $this->entityType)->delete($ids);
```
quicksketch: Although this works with the DatabaseCache backend, it's sort of a fluke of the database abstraction layer. Because `$ids` is an array, it should be:

```suggestion
        cache('cache_entity_' . $this->entityType)->deleteMultiple($ids);
```

I just ran into this line causing an error in the Backdrop port of memcache, which was expecting a single item and was getting passed an array.

core/includes/cache.inc
```diff
@@ -165,8 +165,12 @@ function cache_flush($bin) {
  *   functionality.
  *
  * @see cache_flush()
+ *
+ * @deprecated since 1.22
```
quicksketch: Maybe we'll get this in next release? :smile: 
```suggestion
 * @deprecated since 1.24.0
```

core/includes/menu.inc
```diff
@@ -3495,13 +3495,19 @@ function _menu_clear_page_cache() {
   // Clear the page and block caches, but at most twice, including at
   //  the end of the page load when there are multiple links saved or deleted.
   if ($cache_cleared == 0) {
-    cache_clear_all();
+    if (module_exists('layout')) {
```
indigoxela, line 3498: Layout is a required module (`required = TRUE` in info file), so why checking that it's on, anyway? It can't be off.

That would halve the size of this PR... :grin:

PR #3396, opened 2020-11-11
Issue #3792: Fix step parameter type in valid_number_step().

### LINKED ISSUE #3792: Simple decimals fail to pass validation

ghost:
Fixes https://github.com/backdrop/backdrop-issues/issues/3792

### DIFF
--- a/core/includes/common.inc
+++ b/core/includes/common.inc
@@ -1454,6 +1454,7 @@ function valid_url($url, $absolute = FALSE) {
  * @see http://opensource.apple.com/source/WebCore/WebCore-1298/html/NumberInputType.cpp
  */
 function valid_number_step($value, $step, $offset = 0.0) {
+  $step = (string) $step;
   $double_value = (double) abs($value - $offset);
 
   // The fractional part of a double has 53 bits. The greatest number that could


ghost:
Closing in favour of https://github.com/backdrop/backdrop/pull/3392

PR #3407, opened 2020-11-22
Issue #4702: Change 'Execute' to 'Apply to selected items'

### LINKED ISSUE #4702: [UX][D8] VBO: Change "Execute" to "Apply to selected items"
This is related to #4701 

image

Long-time Drupal/Backdrop users may be familiar with how bulk operations work, but for new users it is not immediately apparent that the "select action" select menu and the "Execute" button are to be used with the checkboxes to the left of each result in the listing (and the fact that there is no header in the checkbox column to explain what the checkboxes are for makes that worse).

I realize that "Apply to selected items" is way more wordy than "Execute", but it clearly conveys the message of "first select which items the action is to be performed on". Compare the above screenshot from Backdrop, to this from D8:

image

klonos:
Fixes https://github.com/backdrop/backdrop-issues/issues/4702

### DIFF
--- a/core/misc/tableselect.js
+++ b/core/misc/tableselect.js
@@ -25,7 +25,7 @@ Backdrop.tableSelect = function () {
   };
 
   // Find all <th> with class select-all, and insert the check all checkbox.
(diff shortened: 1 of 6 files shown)


klonos:
Closing in favor of https://github.com/backdrop/backdrop/pull/3413

PR #3413, opened 2020-11-25
Issue #4702: Change 'Execute' to 'Apply to selected items'

### LINKED ISSUE #4702: [UX][D8] VBO: Change "Execute" to "Apply to selected items"

klonos:
Fixes https://github.com/backdrop/backdrop-issues/issues/4702

Alternative to https://github.com/backdrop/backdrop/pull/3407

### DIFF
--- a/core/modules/views/js/views_bulk_actions.js
+++ b/core/modules/views/js/views_bulk_actions.js
@@ -1,0 +1,45 @@
+/**
+ * @file
+ * This behavior is required when the Views UI module is disabled.
+ *
+ * @see views_form()
+ */
+(function ($) {
+
+"use strict";
+
+Backdrop.behaviors.viewsBulkFormApplyToggle = {
+  attach: function (context) {
+    var $context = $(context);
+    var $applyButton = $('#views-bulk-form-apply');
+    var buttonText = $applyButton.val();
+
+    // Check if at least one checkbox has been ticked, and also if an action has
+    // been selected. Update the disabled state of the "Apply" button, and its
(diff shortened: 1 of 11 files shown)


klonos:
ZenCI tripping again 😓

core/modules/views/js/views_bulk_actions.js
```diff
@@ -0,0 +1,45 @@
...
+      else {
+        var newButtonText = $rowChecked ? Backdrop.t('No action selected') : Backdrop.t('No item selected');
+        $applyButton.attr('disabled', 'disabled').addClass('no-js-hide form-button-disabled').prop('value', newButtonText);
+      }
+    }
+
+    // Initialize the "Apply" button once when the page loads.
+    toggleExecuteButton();
+    // Repeat each time any checkbox is ticked/unticked.
+    $(context).find('.form-checkbox').on('change', function () {
```
docwilmot, line 35: Why not `$('.form-checkbox').on('change', ...`? One less function call.

PR #3416, opened 2020-11-26
Issue #4777: PHP 7.4 test fixes.

### LINKED ISSUE #4777: Complete PHP 7.4 Test Fixes
**Description of the bug**

I tried running Backdrop's full test suite against PHP 7.4 and found a number of PHP notices and warnings.

Some errors that are thrown:

```
Exception Notice     bootstrap.test     613 BootstrapOverrideServerVariablesTes
    unserialize(): Error at offset 0 of 152 bytes

Exception Warning    session.inc        240 backdrop_session_initialize()      
    session_set_save_handler(): Cannot change save handler when session is
    active

Exception Notice     menu.inc          1125 menu_tree_output()                 
    Trying to access array offset on value of type bool
```

When doing a config sync on php 7.2 (or above) the following error is repeated:
```
Deprecated function: assert(): Calling assert() with a string argument is deprecated in HWLDFWordAccumulator->addWords() (line 885 of backdrop/core/includes/diff.inc).
```

It seems most or all of the errors thrown are caused when testing edge-case scenarios and may not occur in the normal usage of Backdrop, however if we are going to upgrade our testing to PHP 7.4+, we'll need to get all tests passing.

**Additional information**

- Backdrop CMS version: 1.18.x
- Web server and its version: PHP-FPM + nginx (Lando)

quicksketch:
Fixes https://github.com/backdrop/backdrop-issues/issues/4777

### DIFF
--- a/core/includes/diff.inc
+++ b/core/includes/diff.inc
@@ -196,8 +196,8 @@ class DiffEngine {
     $edits = array();
     $xi = $yi = 0;
     while ($xi < $n_from || $yi < $n_to) {
-      assert('$yi < $n_to || $this->xchanged[$xi]');
-      assert('$xi < $n_from || $this->ychanged[$yi]');
+      assert($yi < $n_to || $this->xchanged[$xi]);
+      assert($xi < $n_from || $this->ychanged[$yi]);
 
       // Skip matching "snake".
       $copy = array();
@@ -384,7 +384,7 @@ class DiffEngine {
           next($matches);
           if (empty($this->in_seq[$y])) {
             $k = $this->lcsPos($y);
-            assert('$k > 0');
+            assert($k > 0);
             $ymids[$k] = $ymids[$k - 1];
             break;
           }
@@ -393,7 +393,7 @@ class DiffEngine {
         while ($y = current($matches)) {
           next($matches);
           if ($y > $this->seq[$k - 1]) {
-            assert('$y < $this->seq[$k]');
+            assert($y < $this->seq[$k]);
             // Optimization: this is a common case:
             //	next match is just replacing previous match.
             $this->in_seq[$this->seq[$k]] = FALSE;
@@ -402,7 +402,7 @@ class DiffEngine {
           }
           elseif (empty($this->in_seq[$y])) {
             $k = $this->lcsPos($y);
-            assert('$k > 0');
+            assert($k > 0);
             $ymids[$k] = $ymids[$k - 1];
           }
         }
@@ -446,7 +446,7 @@ class DiffEngine {
       }
     }
 
-    assert('$ypos != $this->seq[$end]');
+    assert($ypos != $this->seq[$end]);
 
     $this->in_seq[$this->seq[$end]] = FALSE;
     $this->seq[$end] = $ypos;
@@ -534,7 +534,7 @@ class DiffEngine {
     $i = 0;
     $j = 0;
 
-    assert('count($lines) == count($changed)');
+    assert(count($lines) == count($changed));
     $len = count($lines);
     $other_len = count($other_changed);
 
@@ -555,7 +555,7 @@ class DiffEngine {
       }
 
       while ($i < $len && !$changed[$i]) {
-        assert('$j < $other_len && ! $other_changed[$j]');
+        assert($j < $other_len && !$other_changed[$j]);
         $i++;
         $j++;
         while ($j < $other_len && $other_changed[$j]) {
@@ -592,11 +592,11 @@ class DiffEngine {
           while ($start > 0 && $changed[$start - 1]) {
             $start--;
           }
-          assert('$j > 0');
+          assert($j > 0);
           while ($other_changed[--$j]) {
             continue;
           }
-          assert('$j >= 0 && !$other_changed[$j]');
+          assert($j >= 0 && !$other_changed[$j]);
         }
 
         /*
@@ -620,7 +620,7 @@ class DiffEngine {
             $i++;
           }
 
-          assert('$j < $other_len && ! $other_changed[$j]');
+          assert($j < $other_len && !$other_changed[$j]);
           $j++;
           if ($j < $other_len && $other_changed[$j]) {
             $corresponding = $i;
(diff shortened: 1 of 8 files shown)


core/includes/diff.inc
```diff
@@ -196,8 +196,8 @@ class DiffEngine {
     $edits = array();
     $xi = $yi = 0;
     while ($xi < $n_from || $yi < $n_to) {
-      assert('$yi < $n_to || $this->xchanged[$xi]');
-      assert('$xi < $n_from || $this->ychanged[$yi]');
+      assert($yi < $n_to || $this->xchanged[$xi]);
+      assert($xi < $n_from || $this->ychanged[$yi]);
```
quicksketch, line 200: Using strings for assertions is deprecated as of PHP 7.2+

https://www.php.net/assert#refsect1-function.assert-parameters

core/includes/menu.inc
```diff
@@ -1122,7 +1122,7 @@ function menu_tree_output($tree, $_current_depth = 0) {
     // the active class accordingly. But local tasks do not appear in menu
     // trees, so if the current path is a local task, and this link is its
     // tab root, then we have to set the class manually.
-    if ($data['link']['href'] == $router_item['tab_root_href'] && $data['link']['href'] != $_GET['q']) {
+    if ($router_item && $data['link']['href'] == $router_item['tab_root_href'] && $data['link']['href'] != $_GET['q']) {
```
quicksketch, line 1125: `$router_item` may be NULL, attempting to access it as an array throws a warning.

core/includes/menu.inc
```diff
@@ -2601,12 +2601,11 @@ function menu_link_get_preferred($path = NULL, $selected_menu = NULL) {
     $selected_menu = MENU_PREFERRED_LINK;
   }
 
-  if (!isset($preferred_links[$path])) {
+  if (!isset($preferred_links[$path]) && ($item = menu_get_item($path))) {
```
quicksketch, line 2604: `$item` may be NULL, attempting to access it as an array throws a warning, so we check that it is loaded before proceeding.

core/modules/dblog/tests/dblog.test
```diff
@@ -258,18 +258,21 @@ class DBLogTestCase extends BackdropWebTestCase {
...
-    $this->assertText(format_date($new_user_time, 'custom', 'M d, Y - j:ia'));
+    $this->assertText(format_date($new_user_time, 'custom', 'M d, Y - j:ia', $admin_timezone));
     // Login user.
     $this->assertLogMessage(t('Session opened for %name.', array('%name' => $name)), 'DBLog event was recorded: [login user]');
-    $this->assertText(format_date($login_time, 'custom', 'M d, Y - j:ia'));
+    $this->assertText(format_date($login_time, 'custom', 'M d, Y - j:ia', $admin_timezone));
     // Logout user.
     $this->assertLogMessage(t('Session closed for %name.', array('%name' => $name)), 'DBLog event was recorded: [logout user]');
-    $this->assertText(format_date($logout_time, 'custom', 'M d, Y - j:ia'));
+    $this->assertText(format_date($logout_time, 'custom', 'M d, Y - j:ia', $admin_timezone));
```
quicksketch, line 275: These fixes are needed simply when the web server timezone differs from the administrator timezone.

core/modules/simpletest/tests/bootstrap.test
```diff
@@ -605,9 +605,14 @@ class BootstrapOverrideServerVariablesTestCase extends BackdropUnitTestCase {
...
-    // This test only works on Travis CI through the SetEnv call in
-    // core/misc/travis-ci/vhost.conf. See backdrop_settings_initialize().
+    // This test only runs when the "BACKDROP_SETTINGS" environment variable is
+    // set. This usually is done through a Apache VirtualHost setting.
+    // See backdrop_settings_initialize().
+    $raw_settings = array();
     if (isset($_SERVER['BACKDROP_SETTINGS'])) {
+      $raw_settings = json_decode($_SERVER['BACKDROP_SETTINGS'], TRUE);
+    }
+    if (isset($raw_settings['test_setting'])) {
```
quicksketch, line 615: Because I was running in Lando, the `BACKDROP_SETTINGS` value was being set on all requests. This makes it so that tests can be run on Lando, DDEV, or Pantheon (which all use this ability) without this test failing.

core/modules/system/tests/system.test
```diff
@@ -2497,7 +2503,6 @@ class SystemValidTokenTest extends BackdropWebTestCase {
     // The following checks will throw PHP notices, so we disable error
     // assertions.
     $this->assertErrors = FALSE;
-    $this->assertFalse(backdrop_valid_token(NULL, new stdClass()), 'Token NULL, value object returns FALSE.');
```
quicksketch, line 2500: The second parameter of `backdrop_valid_token()` expects a string. Casting an object to a string now throws a notice, but as this situation should not occur (and PHP now throws the notice if it does occur), removing the check seemed appropriate.

core/modules/system/tests/system.test
```diff
@@ -129,10 +129,16 @@ class EnableDisableTestCase extends ModuleTestCase {
...
    */
   function testEnableDisable() {
-    // Try to enable, disable and uninstall all core modules, unless they are
-    // hidden or required, or testing-only modules.
+    // Try to enable, disable and uninstall all core modules.
     $modules = system_rebuild_module_data();
     foreach ($modules as $name => $module) {
+      // Check the "version" key matches the Backdrop version to restrict to
+      // core modules only.
+      if ($module->info['version'] !== BACKDROP_VERSION) {
```
klonos: This will fail if a contrib or custom module happens to use the same version as core. I think that checking the path for `/core` instead may be a better option here. Something like this (untested):

```suggestion
      if (!strstr($module->uri, 'core/modules')) {
```

quicksketch: Thanks good suggestion! I'll test it and update.

quicksketch:
Merged into 1.x and 1.17.x.

PR #3418, opened 2020-11-30
Issue #4779: Change mention of system_elements() to system_element_info()

### LINKED ISSUE #4779: Non-existing function system_elements() mentioned in template_preprocess_page()
See: https://github.com/backdrop/backdrop/blob/1.x/core/includes/theme.inc#L2688

I could not find any reference of that function, but in D6: https://api.drupal.org/api/drupal/modules%21system%21system.module/function/system_elements/6.x

Is that meant to be https://api.backdropcms.org/api/backdrop/1/search/system_element_info now?

klonos:
Fixes https://github.com/backdrop/backdrop-issues/issues/4779

### DIFF
--- a/core/includes/theme.inc
+++ b/core/includes/theme.inc
@@ -2685,7 +2685,7 @@ function template_preprocess_header(&$variables) {
 /**
  * Preprocess variables for page.tpl.php
  *
- * @see system_elements()
+ * @see system_element_info()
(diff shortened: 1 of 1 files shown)


laryn:
Thanks @klonos, @indigoxela and @BWPanda. I've merged this into 1.x and 1.22.x.

PR #3419, opened 2020-12-03
Issue 3813: Allow cloning user roles

### LINKED ISSUE #3813: [UX] Allow cloning a user role (along with its permissions)
I was tasked to clone a user role for a site the other day. Here are the options one has:

- Execute a db query -> not user-friendly 👎 
- Create a new role, then manually set the same permissions as another role -> endless scrolling and clicking on checkboxes 👎 ; becomes worse the more contrib with additional permissions a site has (or many content types, since they also add multiple permissions); becomes even worse if the role you want to clone has many permissions assigned to it (might also become a victim of https://www.drupal.org/project/drupal/issues/1203766 if the site has many roles)
- https://www.drupal.org/project/duplicate_role (have not actually tested it) -> not ported to Backdrop

We should have a "Clone role" action.

klonos:
Fixes https://github.com/backdrop/backdrop-issues/issues/3813


core/modules/user/user.admin.inc
```diff
@@ -710,6 +728,84 @@ function user_admin_role_submit($form, &$form_state) {
...
 }
 
+/**
+ * Form to clone a role.
+ *
+ * @ingroup forms
+ * @see user_admin_role_clone_submit()
+ */
+function user_admin_role_clone($form, &$form_state, $role) {
+  backdrop_set_title(t('Clone %source_role role', array('%source_role' => $role->label)), PASS_THROUGH);
```
klonos, line 746: I use `PASS_THROUGH` here, in order to allow the role name to be rendered in italics in the page title. Noting that I've added a `check_plain()` in `user_role_save()` for the role label.

core/modules/user/user.admin.inc
```diff
@@ -500,6 +509,13 @@ function user_admin_permissions_submit($form, &$form_state) {
 function user_admin_roles($form, $form_state) {
   $system_config = config('system.core');
 
+  $options = array(0 => t('disabled'));
+  $options += user_roles();
+  // Do not allow the anonymous or authenticated roles to be set as the
+  // administrator.
+  unset($options[BACKDROP_ANONYMOUS_ROLE]);
+  unset($options[BACKDROP_AUTHENTICATED_ROLE]);
+
```
klonos: Fix the select, so that the "disabled" option is at the top of the list. See next comment...

core/modules/user/user.admin.inc
```diff
@@ -508,16 +524,11 @@ function user_admin_roles($form, $form_state) {
...
-  unset($roles[BACKDROP_ANONYMOUS_ROLE]);
-  unset($roles[BACKDROP_AUTHENTICATED_ROLE]);
-  $roles[0] = t('disabled');
   $form['admin_role']['user_admin_role'] = array(
     '#type' => 'select',
     '#title' => t('The role that will be automatically assigned new permissions whenever a module is enabled:'),
-    '#default_value' => $system_config->get('user_admin_role'),
-    '#options' => $roles,
+    '#options' => $options,
+    '#default_value' => $system_config->get('user_admin_role') == NULL ? $options[0] : $system_config->get('user_admin_role'),
```
klonos: ...there's a slight (pre-existing) bug here, in that if no role is set as admin, the first available role in the select is selected as the default value. This in itself is harmless, but if you submit that form for some other reason (reorder roles for example), then the first role in the select is set as admin (whereas the intention could have been to keep the setting disabled).

klonos:
...ZenCI is acting up 😓

PR #3423, opened 2020-12-05
Issue #4783: Prevent accidentally setting admin role

### LINKED ISSUE #4783: An arbitrary role can accidentally be set as admin when the role that was previously set as admin is deleted
I've noticed this while working on #3813

## Steps to reproduce:

1. Depending on what you have access to, either:
   1. Go to `admin/config/development/configuration/single/export` -> Select "Configuration" from the group dropdown -> then select "System core settings"
   1. Open your `system.core.json` config file
1. Inspect the `user_admin_role` setting and confirm that it is set to: `"user_admin_role": "administrator"` 👍 
1. Go to `admin/config/people/roles`
1. Expand the "Administrator role" fieldset -> confirm that the "Administrator" role is selected as admin 👍 
1. Add a new role
1. Expand the "Administrator role" fieldset -> set the newly-added role from the previous step as admin -> save
1. Expand the "Administrator role" fieldset -> confirm that the new role is now configured as admin 👍
1. Once again, either check your `system.core.json` config file, or in `admin/config/development/configuration/single/export` -> confirm that the `user_admin_role` setting is set to the machine name of the new role, instead of `"administrator"` 👍 
1. Back in `admin/config/people/roles` -> delete the role you have created (which is also set as admin).
1. Yet once again, either check your `system.core.json` config file, or in `admin/config/development/configuration/single/export` -> confirm that the `user_admin_role` setting is missing/deleted 👍 (perhaps that shouldn't happen and this is the actual culprit? ...but that's the way things work currently)
1. Back in `admin/config/people/roles` -> Expand the "Administrator role" fieldset -> ...

## Actual behavior

A random role is selected in the dropdown menu (whichever role is at the top of the list; usually "Editor") 👎 ...this means that if people delete a role that was previously set as admin, and then they reorder the roles and save the form (without checking the "Administrator role" fieldset, which is collapsed by default), they may set a random role as admin ...which means that this role will be granted ALL permissions from there on 👎 👎 👎 👎 

## Expected behavior

Things we could do:

1. Make sure that the "disabled" option is the first option in the dropdown menu, so that that is the one that gets selected by default.
1. When deleting the role that's set as admin:
   - throw a warning
   - set the "Administrator role" fieldset to be expanded instead of collapsed (although people may still miss this and accidentally set an arbitrary role as admin).
1. Make sure that after a role that was previously set as admin is deleted:
   - the admin dropdown is set to the "disabled" option (so that no random role gets accidentally set as admin)
   - instead of removing the `"user_admin_role"` entry in `system.core.json`, explicitly set it to `"user_admin_role": 0`
1. In the "Delete role" confirmation dialog:
   - show a warning if the to-be-deleted role is set as admin
   - provide a dropdown, to prompt selecting a new role to be set as admin ...or select the "disabled" option

klonos:
Fixes https://github.com/backdrop/backdrop-issues/issues/4783

### DIFF
--- a/core/modules/user/user.install
+++ b/core/modules/user/user.install
@@ -1230,8 +1230,7 @@ function user_update_1018() {
 }
 
 /**
- * Explicitly set the "user_admin_role" attribute to "0" (disabled) if it's
- * missing.
+ * Set the "user_admin_role" attribute to "0" (disabled) if it's missing.
  */
 function user_update_1019() {
   $config = config('system.core');
--- a/core/modules/user/user.install
+++ b/core/modules/user/user.install
@@ -1230,7 +1230,8 @@ function user_update_1018() {
 }
 
 /**
- * Explicitly set the "user_admin_role" attribute to "0" if it's missing.
+ * Explicitly set the "user_admin_role" attribute to "0" (disabled) if it's
+ * missing.
  */
 function user_update_1019() {
   $config = config('system.core');
--- a/core/modules/user/user.install
+++ b/core/modules/user/user.install
@@ -1230,5 +1230,16 @@ function user_update_1018() {
 }
 
 /**
+ * Explicitly set the "user_admin_role" attribute to "0" if it's missing.
+ */
+function user_update_1018() {
+  $config = config('system.core');
+  if ($config->get('user_admin_role') ==  NULL) {
+    $config->set('user_admin_role', 0);
+    $config->save();
+  }
+}
+
+/**
  * @} End of "addtogroup updates-7.x-to-1.x"
  */
--- a/core/modules/user/user.install
+++ b/core/modules/user/user.install
@@ -1234,7 +1234,7 @@ function user_update_1018() {
  */
 function user_update_1018() {
   $config = config('system.core');
-  if ($config->get('user_admin_role') ==  NULL) {
+  if ($config->get('user_admin_role') == NULL) {
     $config->set('user_admin_role', 0);
     $config->save();
   }
--- a/core/modules/user/user.install
+++ b/core/modules/user/user.install
@@ -1232,7 +1232,7 @@ function user_update_1018() {
 /**
  * Explicitly set the "user_admin_role" attribute to "0" if it's missing.
  */
-function user_update_1018() {
+function user_update_1019() {
   $config = config('system.core');
   if ($config->get('user_admin_role') == NULL) {
     $config->set('user_admin_role', 0);
--- a/core/modules/user/user.module
+++ b/core/modules/user/user.module
@@ -2528,23 +2528,23 @@ function user_role_save($role) {
  *   The name of the role to be deleted.
  */
 function user_role_delete($role_name) {
-  $role = user_role_load($role_name);
-
-  // If this is the administrator role, delete the user_admin_role variable.
-  if ($role_name == config_get('system.core', 'user_admin_role')) {
-    $system_config = config('system.core');
-    $system_config->clear('user_admin_role');
+  // If this is the administrator role, explicitly set user_admin_role attribute
+  // to 0 ("disabled").
+  $system_config = config('system.core');
+  if ($role_name == $system_config->get('user_admin_role')) {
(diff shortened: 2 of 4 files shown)


klonos:
...refreshing PR sandbox

klonos:
...again 😓

core/modules/user/user.module
```diff
@@ -2528,23 +2528,23 @@ function user_role_save($role) {
...
-  $config->delete();
+  $user_config = config('user.role.' . $role_name);
+  $user_config->delete();
 
   // Update the users who have this role set:
   db_delete('users_roles')
     ->condition('role', $role_name)
     ->execute();
 
+  $role = user_role_load($role_name);
```
indigoxela: This confuses me.
Above you're using the "shortcut" to load the config directly instead of using user_role_load() - `config('user.role.' . $role_name);`

But here you're using user_role_load() again. On the role that has just been deleted. Mind to explain a bit?

core/modules/user/user.install
```diff
@@ -1229,6 +1229,17 @@ function user_update_1018() {
   $config->save();
 }
 
+/**
+ * Explicitly set the "user_admin_role" attribute to "0" if it's missing.
```
indigoxela: Maybe not important, but this is the text the admin user sees after running the update, right?
Maybe use "disabled" here instead of the ambiguous "0"? Just a suggestion.

core/modules/user/user.module
klonos: This is not actually me doing this - I'm just rearranging preexisting code. This line here was initially in line 2531, right after the function declaration.

core/modules/user/user.install
klonos: This is exactly what we're doing though 🤷 ...perhaps we could be more descriptive(?) ...perhaps something like this instead?:

```suggestion
 * Ensure that the admin role setting is disabled, by explicitly setting the
 * value of the "user_admin_role" attribute to "0" if it's missing.
```

core/modules/user/user.module
indigoxela: Right, but your rearrangement means that you load the role *after* it has been deleted. Before your change the role has been loaded *before* it got deleted.

This very likely breaks hook_user_role_delete (below).

core/modules/user/user.install
indigoxela: > This is exactly what we're doing though

That's correct. Your alternate suggestions seems a bit wordy to me. So maybe leave it as-is? Whatever you decide.

klonos: Thanks @indigoxela 🙏 ...I'll wait for a few more people to review, and see what they think. I'll update this accordingly then.

core/modules/user/user.module
klonos: 🤔

klonos: ...makes me think if `module_invoke_all()` below should be called earlier (before the role is deleted), or even if we should be passing the role name instead of the role object to the hook.

Anyway, it seems that this change should not be part of this PR. I'm usually good with adding inline comments when making changes that are required, and since I don't see any comments here, I don't think that this was required/intentional.

core/modules/user/user.install
klonos: I've reverted this change now.

core/modules/user/user.module
klonos: ...I've reverted that specific change now.

core/modules/user/user.install
```diff
@@ -1229,6 +1229,18 @@ function user_update_1018() {
   $config->save();
 }
 
+/**
+ * Explicitly set the "user_admin_role" attribute to "0" (disabled) if it's
+ * missing.
```
quicksketch: The first line of any function docblock should be a stand-alone sentence (followed by a blank line if there is more documentation). Let's shorten this up just by removing "Explicitly".

```suggestion
 * Set the "user_admin_role" attribute to "0" (disabled) if it's missing.
```

quicksketch:

PR #3429, opened 2020-12-08
Issue #4790: Use human name in message when deleting roles

### LINKED ISSUE #4790: [UX] Use the human-readable name of the role in the "role deleted" message
While working on #3813, I've noticed that the machine name of the role is shown in the message:

[screenshot coming soon]

klonos:
Fixes https://github.com/backdrop/backdrop-issues/issues/4790

### DIFF
--- a/core/modules/user/tests/user.test
+++ b/core/modules/user/tests/user.test
@@ -2190,7 +2190,7 @@ class UserRoleAdminTestCase extends BackdropWebTestCase {
     $role_name = 'administrator';
     $role = user_role_load($role_name);
     $this->backdropPost("admin/config/people/roles/delete/$role_name", array(), t('Delete'));
-    $this->assertText(t('The %role role has been deleted.', array('%role' => $role->label)), 'The role has been deleted');
+    $this->assertText(strip_tags(t('The %role role has been deleted.', array('%role' => $role->label))), 'The role has been deleted');
     $this->assertNoLinkByHref("admin/config/people/roles/edit/$role_name", 'Role edit link removed.');
     backdrop_static_reset('user_roles');
     $this->assertFalse(user_role_load($role_name), 'A deleted role can no longer be loaded.');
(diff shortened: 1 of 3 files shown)


klonos:
...ZenCI 😓

klonos:
🤦

klonos:
Finally!!! 😌

core/modules/user/tests/user.test
```diff
@@ -2190,7 +2190,7 @@ class UserRoleAdminTestCase extends BackdropWebTestCase {
     $role_name = 'administrator';
     $role = user_role_load($role_name);
     $this->backdropPost("admin/config/people/roles/delete/$role_name", array(), t('Delete'));
-    $this->assertText(t('The administrator role has been deleted.'), 'The role has been deleted');
+    $this->assertText(strip_tags(t('The %role role has been deleted.', array('%role' => $role->label))), 'The role has been deleted');
```
quicksketch, line 2193: If checking with HTML tags present, you can use `assertRaw` instead of `assertText`:

```suggestion
    $this->assertRaw(t('The %role role has been deleted.', array('%role' => $role->label)), 'The role has been deleted');
```

core/modules/user/user.module
```diff
@@ -2482,7 +2482,7 @@ function user_role_load_by_name($role_name) {
 function user_role_save($role) {
   // Trim spaces from names.
   $role->name = trim($role->name);
-  $role->label = trim($role->label);
+  $role->label = check_plain(trim($role->label));
```
quicksketch, line 2485: This change looks incorrect. If you save a role that includes an HTML entity like `&` then when you re-edit the role it will show the encoded version instead, saving again would then double-encode, triple-encode, etc with each save. I don't think this change is necessary to accomplish the objective, it should work fine without it.

quicksketch (changes requested):

PR #3431, opened 2020-12-08
Issue #4792: Add a "Cancel" link in the role edit form

### LINKED ISSUE #4792: [UX] Add a "Cancel" link in the role edit form
I noticed that there is no "Cancel" link in that form while working on #3813.

We should add one.

klonos:
Fixes https://github.com/backdrop/backdrop-issues/issues/4792


quicksketch:

PR #3436, opened 2020-12-11
Issue #4799: Check 'distribution_name' exists before returning it.

### LINKED ISSUE #4799: "Undefined index: distribution_name" for Drupal migrations
If a Drupal site is upgraded to Backdrop then the installation doesn't run and 'distribution_name' doesn't get set. However when we get that information from the profile info file, we assume it's there and don't allow for instances where it doesn't exist. Then users get the following notice:

> Notice: Undefined index: distribution_name in backdrop_install_profile_distribution_name() (line 222 of [SNIP]\core\includes\install.inc).

We should add a check to backdrop_install_profile_distribution_name() before attempting to return it.

ghost:
Fixes https://github.com/backdrop/backdrop-issues/issues/4799

### DIFF
--- a/core/includes/install.inc
+++ b/core/includes/install.inc
@@ -219,7 +219,12 @@ function backdrop_install_profile_distribution_name() {
   else {
     $profile = backdrop_get_profile();
     $info = system_get_info('module', $profile);
-    return $info['distribution_name'];
+    if (!empty($info['distribution_name'])) {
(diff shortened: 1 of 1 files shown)


ghost:
Since this PR branch was added to the core repo, and not to my personal fork, I'm going to close this PR until I can do it properly in future.

PR #3437, opened 2020-12-11
Issue #4800: Update the docblock for theme_username() since 'account'…

### LINKED ISSUE #4800: Update the documentation on theme_username(): `account` is not a user account.
This is a follow-up to https://github.com/backdrop/backdrop-issues/issues/4796

The function `theme_username()` has a variable with the key `account` that is not always a user account, but the docblock states that it is. This should be updated.

jenlampton:
Fixes https://github.com/backdrop/backdrop-issues/issues/4800

### DIFF
--- a/core/modules/user/user.theme.inc
+++ b/core/modules/user/user.theme.inc
@@ -56,7 +56,7 @@ function theme_user_signature($variables) {
  *
  * @param $variables
  *   An associative array containing:
- *   - account: The user object to format.
+ *   - account: Any object with 'name', 'uid', and 'picture' proprties.
  *   - name: The user's name, sanitized.
  *   - extra: Additional text to append to the user's name, sanitized.
  *   - link_path: The path or URL of the user's profile page, home page, or
@@ -201,11 +201,11 @@ function theme_user_admin_roles($variables) {
 }
 
 /**
- * Process variables for user-picture.tpl.php.
+ * Prepares variables for user_picture templates.
  *
- * The $variables array contains the following arguments:
- * - $account: A user, node or comment object with 'name', 'uid' and 'picture'
- *   fields.
+ * @param array $variables
+ *   An associative array containing:
+ *   - account: Any object with 'name', 'uid', and 'picture' proprties.
  *
  * @see user-picture.tpl.php
  */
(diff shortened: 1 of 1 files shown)


core/modules/user/user.theme.inc
```diff
@@ -56,7 +56,7 @@ function theme_user_signature($variables) {
  *
  * @param $variables
  *   An associative array containing:
- *   - account: The user object to format.
+ *   - account: Any object with 'name', 'uid', and 'picture' proprties.
```
klonos, line 59: ...perhaps add "such as a user object"?

avpaderno: I would use _An object with the 'name', 'uid', and 'picture' properties._ _properties_ has been written _proprties_.

core/modules/user/user.theme.inc
```diff
@@ -201,11 +201,11 @@ function theme_user_admin_roles($variables) {
...
 /**
- * Process variables for user-picture.tpl.php.
+ * Prepares variables for user_picture templates.
  *
- * The $variables array contains the following arguments:
- * - $account: A user, node or comment object with 'name', 'uid' and 'picture'
- *   fields.
+ * @param array $variables
+ *   An associative array containing:
+ *   - account: Any object with 'name', 'uid', and 'picture' proprties.
```
avpaderno, line 208: There is a typo: _proprties_ instead of _properties_.

core/modules/user/user.theme.inc
ghost: I agree that it'd be nice to still mention user objects as an example, since they make up the majority use cases.

avpaderno: Would *An object with the 'name', 'uid', and 'picture' properties, for example a user object.* be fine?

ghost: @kiamlaluno Yea, though I'd prefer to remove the word 'the' in your example.

avpaderno: Yes, since *properties* is plural, removing the article makes sense. (I still need to change from *Italian mind* when writing in English. 😅)

izmeez:
The typo: `proprties` is repeated several times.

PR #3438, opened 2020-12-11
Issue #4796: Rename the 'account' key in theme_username() since it ca…

### LINKED ISSUE #4796: [DX] Rename `account` variable in theme_username() and theme_user_picture() to `object` because they are not limited to User accounts
I am trying to theme the username, but my theme function is not working on the search results page, because instead of a user account I seem to have a node object in this location.

update: My `theme_usrename()` function is replacing the username with values from fields on the `User` account object.  When searching for those fields on `$variables['account']` I was surprised to find this item was not a `User` account at all, but a `Node` instead.

Node module seems to pass in the node `'user' => theme('username', array('account' => $node)),`.  Is this intentional?

edit: I just found the same problem in `theme_user_picture()`.

```
$results[] = array(
      'link' => url($uri['path'], array_merge($uri['options'], array('absolute' => TRUE))),
      'type' => check_plain(node_type_get_name($node)),
      'title' => $node->title,
      'user' => theme('username', array('account' => $node)),
      'date' => $node->changed,
      'node' => $node,
      'extra' => $extra,
      'score' => $item->calculated_score,
      'snippet' => search_excerpt($keys, $node->rendered),
      'langcode' => $node->langcode,
    );
```

Switching this line as follows fixes my problem:
```
$results[] = array(
      'link' => url($uri['path'], array_merge($uri['options'], array('absolute' => TRUE))),
      'type' => check_plain(node_type_get_name($node)),
      'title' => $node->title,
      'user' => theme('username', array('account' => user_load($node->uid))),
      'date' => $node->changed,
      'node' => $node,
      'extra' => $extra,
      'score' => $item->calculated_score,
      'snippet' => search_excerpt($keys, $node->rendered),
      'langcode' => $node->langcode,
    );
```

jenlampton:
Fixes https://github.com/backdrop/backdrop-issues/issues/4796

### DIFF
--- a/core/modules/comment/comment.admin.inc
+++ b/core/modules/comment/comment.admin.inc
@@ -122,7 +122,7 @@ function comment_admin_overview($form, &$form_state, $arg) {
           '#options' => array('attributes' => array('title' => $title), 'fragment' => 'comment-' . $comment->cid),
         ),
       ),
-      'author' => theme('username', array('account' => $comment)),
+      'author' => theme('username', array('object' => $comment)),
       'posted_in' => array(
         'data' => array(
           '#type' => 'link',
--- a/core/modules/comment/comment.module
+++ b/core/modules/comment/comment.module
@@ -1877,7 +1877,7 @@ function comment_form($form, &$form_state, $comment) {
     $form['author']['_author'] = array(
       '#type' => 'item',
       '#title' => t('Your name'),
-      '#markup' => theme('username', array('account' => $user)),
+      '#markup' => theme('username', array('object' => $user)),
     );
     $form['author']['name'] = array(
       '#type' => 'value',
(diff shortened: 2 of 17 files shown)


avpaderno:
It doesn't seem the tests are failing for the changes introduced in this MR. For example, the *Comment author successfully changed to an unverified user* test uses the following code.

```php
// Test changing the comment author to an unverified user.
$random_name = $this->randomName();
$this->backdropGet('comment/' . $comment->id . '/edit');
$comment = $this->postComment(NULL, $comment->comment, $comment->subject, array('name' => $random_name));
$this->backdropGet('node/' . $this->node->nid);
$this->assertText($random_name . ' (' . t('not verified') . ')', 'Comment author successfully changed to an unverified user.');
```

It fails because the *(not verified)* string isn't found in the page.

PR #3444, opened 2020-12-13
Issue #4804: Make message about update_free_access a proper warning

### LINKED ISSUE #4804: D7 upgrades: Make the reminder message about update_free_access a proper warning
When you upgrade a site from D7 to Backdrop, after running the pending updates, you get this message:

image

That warning in bold text should be a proper warning message, using `backdrop_set_meassge()`

klonos:
Fixes https://github.com/backdrop/backdrop-issues/issues/4804


core/update.php
```diff
@@ -228,7 +228,7 @@ function update_results_page() {
   }
 
   if (settings_get('update_free_access')) {
-    $output .= "<p><strong>Reminder: Don't forget to set the <code>\$settings[&#39;update_free_access&#39;]</code> value in your <code>settings.php</code> file back to <code>FALSE</code>.</strong></p>";
+    backdrop_set_message(t("Reminder: Don't forget to set the <code>\$settings['update_free_access']</code> value in your <code>settings.php</code> file back to <code>FALSE</code>."), 'warning');
```
indigoxela: I wonder if it makes sense here to add the code snippets as parameters to t().

`$setting = '<code>$settings[\'update_free_access\']</code>';`

`t('Reminder ... !setting ... !filename')`; (Placeholders with `!` are inserted as-is.

That would move all markup out of the translatable string. And neither setting nor filename need translation.
Totally optional, though...

docwilmot: This function is called as a case in [the switch statement below](https://github.com/backdrop/backdrop/blob/b2690394c575ffe659186a6a818dba93da418f6e/core/update.php#L538), and it seems neither this nor any of the other functions called return any translated code (there is no `t()` in either of them). I cannot say why, but for consistency we perhaps shouldnt be using `t()` here at all?

klonos: @indigoxela, @docwilmot I've now removed `t()` and filed this as a follow-up, to figure out why strings are not being translated in update.php: https://github.com/backdrop/backdrop-issues/issues/5394

klonos:
Got 2 (different) random failures for php7 and php5. Restarting...

core/update.php
```diff
@@ -228,7 +228,7 @@ function update_results_page() {
   }
 
   if (settings_get('update_free_access')) {
-    $output .= "<p><strong>Reminder: Don't forget to set the <code>\$settings[&#39;update_free_access&#39;]</code> value in your <code>settings.php</code> file back to <code>FALSE</code>.</strong></p>";
+    backdrop_set_message("Reminder: Don't forget to set the <code>\$settings['update_free_access']</code> value in your <code>settings.php</code> file back to <code>FALSE</code>.", 'warning');
```
ghost, line 231: Rather than escaping the variable, shouldn't this instead be:

```suggestion
    backdrop_set_message("Reminder: Don't forget to set the <code>" . $settings['update_free_access'] . "</code> value in your <code>settings.php</code> file back to <code>FALSE</code>.", 'warning');
```

klonos: @BWPanda your code change would actually try to concatenate the value of the `$settings['update_free_access']` variable, which would throw "undefined variable" errors etc.

What we are trying to do here is to output the actual string `$settings['update_free_access']`. Since this string already contains single quotes, we cannot use single quotes in `backdrop_set_message()`, but using double quotes would try to interpret the string as code (a variable). Hence the escaping of the `$`.

I hope that this makes sense now.

The alternative would be something like this:

```suggestion
    $free_access_warning = format_string("Reminder: Don't forget to set the <code>@free_access</code> value in your <code>settings.php</code> file back to <code>FALSE</code>.", array('@free_access' => "$settings['update_free_access']"));
    backdrop_set_message($free_access_warning, 'warning');
```

...but too much code IYAM.

ghost: Ah, ok, I get it now. Sorry, my bad.

ghost (approved):
Looks good to me.

quicksketch:

PR #3445, opened 2020-12-13
Issue #4748: fix mod_brotli double compression

### LINKED ISSUE #4748: Disable Brotli for pre-compressed CSS and JS
**Description of the need**

When Brotli is in use on a web server that serves pages through https and "aggregate css and(or) aggregate js" are enabled, the CSS and JS files do not "decode" properly, resulting in console errors for those files.

```
net::ERR_CONTENT_DECODING_FAILED 200
```

For additional context and patch see: https://www.drupal.org/project/drupal/issues/2960808

**Proposed solution**

Adding E=no-brotli:1 to the following lines (included) removes these errors and allows the pages to load properly.

```
    # Serve correct content types, and prevent mod_deflate double gzip.
    RewriteRule \.css\.gz$ - [T=text/css,E=no-gzip:1,E=no-brotli:1]
    RewriteRule \.js\.gz$ - [T=text/javascript,E=no-gzip:1,E=no-brotli:1]

```

**Additional information**
Example:
image

philsward:
Fixed: https://github.com/backdrop/backdrop-issues/issues/4748

Added support for Apache mod_brotli to prevent it from double compressing, compressed files such as .css.gzip or .js.gzip

Also updated the comment to reflect Brotli

### DIFF
--- a/.htaccess
+++ b/.htaccess
@@ -142,7 +142,7 @@ DirectoryIndex index.php index.html index.htm
     RewriteCond %{REQUEST_FILENAME}\.gz -s
     RewriteRule ^(.*)\.js $1\.js\.gz [QSA]
 
-    # Serve correct content types, and prevent double compression from mod_deflate (gzip) and mod_brotli (br).
+    # Serve correct content types, and prevent double compression.
     RewriteRule \.css\.gz$ - [T=text/css,E=no-gzip:1,E=no-brotli:1]
     RewriteRule \.js\.gz$ - [T=text/javascript,E=no-gzip:1,E=no-brotli:1]
 
--- a/.htaccess
+++ b/.htaccess
@@ -142,9 +142,9 @@ DirectoryIndex index.php index.html index.htm
     RewriteCond %{REQUEST_FILENAME}\.gz -s
     RewriteRule ^(.*)\.js $1\.js\.gz [QSA]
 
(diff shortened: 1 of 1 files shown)


.htaccess
```diff
@@ -142,9 +142,9 @@ DirectoryIndex index.php index.html index.htm
     RewriteCond %{REQUEST_FILENAME}\.gz -s
     RewriteRule ^(.*)\.js $1\.js\.gz [QSA]
 
-    # Serve correct content types, and prevent mod_deflate double gzip.
-    RewriteRule \.css\.gz$ - [T=text/css,E=no-gzip:1]
-    RewriteRule \.js\.gz$ - [T=text/javascript,E=no-gzip:1]
+    # Serve correct content types, and prevent double compression from mod_deflate (gzip) and mod_brotli (br).
```
quicksketch: This line is running over the 80 character wrapping limit, could we just match Drupal is using with the abbreviated:
```suggestion
    # Serve correct content types, and prevent double compression.
```

philsward: I think the proposed comment change is fine. If it becomes an issue of confusion, we can address it in more detail later.

One idea is to change it to say "Serve correct pre-compressed content". I won't split hairs on that part though.

quicksketch:
Thanks! Merged into 1.x and 1.19.x.

PR #3446, opened 2020-12-14
Issue #4802: Remove suggestions to 'contact an administrator' from core.

### LINKED ISSUE #4802: [UX] Remove all instances of "please contact an administrator" from messaging
I really hate it when Backdrop tells people to contact an administrator. 

For one, lots of people are the administrator when they see that message. Telling them to contact themselves is unhelpful. 

For two, lots of people have no idea who an administrator is, or how to contact them. Telling this person to do so is also unhelpful. 

For three, The people who can contact an administrator will usually do so anyway, when they get any kind of error - they don't need to be told to do it. 

## Proposal: 

I would like to remove all instances of 'please contact a site administrator' from core. 

Related: 
* see related conversations in #2229 
* Comment: https://github.com/backdrop/backdrop-issues/issues/3824#issuecomment-743503886

ghost:
FIxes https://github.com/backdrop/backdrop-issues/issues/4802


core/modules/system/system.install
```diff
@@ -67,7 +67,7 @@ function system_requirements($phase) {
     $requirements['php'] = array(
       'title' => $t('PHP'),
       'value' => $phpversion,
-      'description' => $t('The phpinfo() function has been disabled for security reasons. To see your server\'s phpinfo() information, change your PHP settings or contact your server administrator. For more information, <a href="@phpinfo">Enabling and disabling phpinfo()</a> handbook page.', array('@phpinfo' => 'http://drupal.org/node/243993')),
+      'description' => $t('The phpinfo() function has been disabled for security reasons. To see your server\'s phpinfo() information, change your PHP settings. For more information, <a href="@phpinfo">Enabling and disabling phpinfo()</a> handbook page.', array('@phpinfo' => 'http://drupal.org/node/243993')),
```
indigoxela: Two minor things:

`your server\'s` Life might be easier for translators, if the whole string was wrapped in `"`, because then the escaping isn't necessary.

The documentation link points to drupal.org/node/243993 - does an equivalent exist for Backdrop? (Can't seem to find one.)

core/modules/system/system.install
```diff
@@ -219,10 +219,7 @@ function system_requirements($phase) {
 
     if (!empty($description)) {
       if ($php_ini_path = get_cfg_var('cfg_file_path')) {
-        $description .= ' ' . $t('Increase the memory limit by editing the memory_limit parameter in the file %configuration-file and then restart your web server (or contact your system administrator or hosting provider for assistance).', array('%configuration-file' => $php_ini_path));
-      }
-      else {
-        $description .= ' ' . $t('Contact your system administrator or hosting provider for assistance with increasing your PHP memory limit.');
+        $description .= ' ' . $t('Increase the memory limit by editing the memory_limit parameter in the file %configuration-file and then restart your web server.', array('%configuration-file' => $php_ini_path));
```
indigoxela, line 224: What if they can't? I don't think many shared webhosting providers allow any user to restart the web server. :wink:

core/modules/system/system.install
ghost: If we wrapped in `"` then we wouldn't have to escape `server's`, but we would have to escape `<a href="@phpinfo">`. Because that's two instances instead of one, I'm guessing it's better to leave as-is.

As for documentation, if we don't have an equivalent page (which I assume we don't, then we should also leave as-is.

core/modules/system/system.install
ghost: I don't know. I just know that this text was always there, I just removed the 'contact an admin' parts...

jlfranklin (commented):
*moved inline*

core/includes/mail.inc
```diff
@@ -186,7 +186,7 @@ function backdrop_mail($module, $key, $to, $language, $params = array(), $reply
       // Log errors.
       if (!$message['result']) {
         watchdog('mail', 'Error sending e-mail (from %from to %to).', array('%from' => $message['from'], '%to' => $message['to']), WATCHDOG_ERROR);
-        backdrop_set_message(t('Unable to send e-mail. Contact the site administrator if the problem persists.'), 'error');
```
jlfranklin, line 189: While the "contact an administrator" isn't helpful (for reasons mentioned in the main issue), it's at least something actionable. There are a lot of reasons why mail may fail to send: invalid credentials, misconfigured local server, misconfigured remote server, no such domain or user, blocked for spam/virus/graylisting, message too long. Some of these can be addressed by the local site admin, others require a server admin or the receiver's admin to fix. Even if we don't put it in the on-screen message, we should add something to the watchdog logs. As the admin, please give me a clue.

core/modules/system/system.install
indigoxela: Nice catch. :+1: Feel free to mark as resolved (I can't)

core/modules/system/system.install
indigoxela: > I just know that this text was always there

Sure, but I see potential for improvement. The text has probably been untouched for a while, because hardly anyone ever sees it.
Anyway, your decision. :wink:

jlfranklin: There are a number of error messages where we could provide suggestions on how to resolve it.  Adding a link to a page on b.o that goes into detail about updating your php.ini and what to do for common hosting situations -- Ubuntu vs RHEL vs cheap shared hosting vs pro-grade shared hosting.  That said, such changes are well beyond the scope of this issue.

core/includes/mail.inc
indigoxela: I also hesitated a bit looking at that change, but if sending mail failed, there _will_ be a watchdog entry anyway with more information. To my understanding the end user (the one that tried to send a mail) won't need much more information than "it failed".

jlfranklin: The only watchdog message guaranteed here is "Error sending email (from %from to %to.)"  Does `$system` keep a last error message that we can include in the logs?  If not, that's something we should add to the mail handler API (and is out-of-scope for this issue.)

indigoxela: > The only watchdog message guaranteed here is "Error sending email (from %from to %to.)"

Depending on what's used for sending mails, there _will_ be additional ones. I don't think, there are more details available in function backdrop_mail(), which we could log. And getting more goes far beyond the scope of this PR - which tries to deal with wording.

But feel free to open an issue for whatever you're missing when sending mail fails.

core/modules/installer/installer.browser.inc
```diff
@@ -588,7 +588,7 @@ function installer_browser_add_remove_queue_dialog_link_submit($form, &$form_sta
 function installer_browser_check_zip_loaded() {
   $zip_loaded = array_key_exists('zip', archiver_get_info());
   if (!$zip_loaded) {
-    backdrop_set_message(t('Sorry, it seems that the PHP .zip extension is not loaded on your server. You will not be able to download any projects using Project Installer until this is fixed. Please contact your website administrator.'), 'warning', FALSE);
+    backdrop_set_message(t('Sorry, it seems that the PHP .zip extension is not loaded on your server. You will not be able to download any projects using Project Installer until this is fixed.'), 'warning', FALSE);
```
klonos: Might as well remove the "Sorry" and "it seems" bits here. Also a slight order change with the words "PHP" and ".zip" so it is more precise:

```suggestion
    backdrop_set_message(t('The .zip PHP extension is not loaded on your server. You will not be able to download any projects using Project Installer until this is fixed.'), 'warning', FALSE);
```

core/modules/system/system.install
klonos: >As for documentation, if we don't have an equivalent page (which I assume we don't, then we should also leave as-is.

Or, quickly whip up one and do it: https://backdropcms.org/user-guide/enabling-and-disabling-phpinfo 😉

core/modules/installer/installer.browser.inc
jlfranklin: Nit pick - it's not .zip, it's just zip.  (no leading dot.)

core/modules/system/system.install
ghost: Yes, I'd say a new issue should be opened for this.

core/includes/mail.inc
ghost: I'd say a new issue is needed for this.

core/modules/system/system.install
```diff
@@ -67,7 +67,7 @@ function system_requirements($phase) {
     $requirements['php'] = array(
       'title' => $t('PHP'),
       'value' => $phpversion,
-      'description' => $t('The phpinfo() function has been disabled for security reasons. To see your server\'s phpinfo() information, change your PHP settings or contact your server administrator. For more information, <a href="@phpinfo">Enabling and disabling phpinfo()</a> handbook page.', array('@phpinfo' => 'http://drupal.org/node/243993')),
+      'description' => $t('The phpinfo() function has been disabled for security reasons. To see your server\'s phpinfo() information, change your PHP settings. For more information, <a href="@phpinfo">Enabling and disabling phpinfo()</a> handbook page.', array('@phpinfo' => 'https://backdropcms.org/user-guide/enabling-and-disabling-phpinfo')),
       'severity' => REQUIREMENT_INFO,
```
indigoxela, line 73: One possibility to get rid of all escaping and remove all markup from the translatable string:

`$doclink = l($t('Enabling and disabling phpinfo()'), 'https://backdropcms.org/user-guide/enabling-and-disabling-phpinfo');`

and then:

```
      'description' => $t("The phpinfo() function has been disabled for security reasons. To see your server's phpinfo() information, change your PHP settings. For more information see the !doclink handbook page.", array(
        '!doclink' => $doclink,
      )),
```
This also includes a small wording change, BTW, "For more information see the...".

ghost: Are you sure putting `t()` inside `t()` works properly...?

indigoxela: I tested it locally, it works. At the time the !doclink placeholder is used, it's already populated with the full link, it's just markup.

klonos: @indigoxela what you are suggesting is against what is advised re links within `t()`. See: https://www.drupal.org/node/322774

> Bad examples:

```php
// DO NOT DO THESE THINGS
$BAD_EXTERNAL_LINK = t('Look at Drupal documentation at !handbook.', array(
  '!handbook' => '<a href="http://drupal.org/handbooks">'. t('the Drupal Handbooks') .'</a>',
));
$ANOTHER_BAD_EXTERNAL_LINK = t('Look at Drupal documentation at <a href="http://drupal.org/handbooks">the Drupal Handbooks</a>.');
$BAD_INTERNAL_LINK = t('To get an overview of your administration options, go to !administer in the main menu.', array(
  '!administer' => l(t('the Administer screen'), 'admin'),
);
```

> Good examples:

```php
// Do this instead.
$external_link = t('Look at Drupal documentation at <a href="@drupal-handbook">the Drupal Handbooks</a>.', array(
  '@drupal-handbook' => 'http://drupal.org/handbooks',
));
$internal_link = t('To get an overview of your administration options, go to <a href="@administer-page">the Administer screen</a> in the main menu.', array(
  '@administer-page' => url('admin'),
));
```

indigoxela: > what you are suggesting is against what is advised re links within t()

I disagree, because my suggestion obviously makes it easier for translators. If it's "against" something and may not be used... - it's only a suggestion. :wink:

BTW, my suggestion isn't actually covered by the "bad examples", as the variable containing the link is created beforehand and not within t().

ghost: Thanks for the link @klonos. I recalled seeing that somewhere before, but couldn't find it yesterday...

@indigoxela I feel your suggestion is close enough to that 'bad' example that we shouldn't do it. Especially if only to avoid escaping a single apostrophe. Happy to hear from other core committees though.

core/modules/system/system.install
```diff
@@ -69,7 +69,7 @@ function system_requirements($phase) {
     $requirements['php'] = array(
       'title' => $t('PHP'),
       'value' => $phpversion,
-      'description' => $t('The phpinfo() function has been disabled for security reasons. To see your server\'s phpinfo() information, change your PHP settings or contact your server administrator. For more information, <a href="@phpinfo">Enabling and disabling phpinfo()</a> handbook page.', array('@phpinfo' => 'http://drupal.org/node/243993')),
+      'description' => $t('The phpinfo() function has been disabled for security reasons. To see your server\'s phpinfo() information, change your PHP settings. For more information, <a href="@phpinfo">Enabling and disabling phpinfo()</a> handbook page.', array('@phpinfo' => 'https://backdropcms.org/user-guide/enabling-and-disabling-phpinfo')),
```
klonos: There's a verb missing there:

```suggestion
      'description' => $t('The phpinfo() function has been disabled for security reasons. To see your server\'s phpinfo() information, change your PHP settings. For more information, see the <a href="@phpinfo">Enabling and disabling phpinfo()</a> handbook page.', array('@phpinfo' => 'https://backdropcms.org/user-guide/enabling-and-disabling-phpinfo')),
```

klonos: ...either that ☝🏼 , or something like this:

```suggestion
      'description' => $t('The phpinfo() function has been disabled for security reasons. To see your server\'s phpinfo() information, change your PHP settings. More information here: <a href="@phpinfo">Enabling and disabling phpinfo()</a>', array('@phpinfo' => 'https://backdropcms.org/user-guide/enabling-and-disabling-phpinfo')),
```

quicksketch:
Thanks folks! Merged into 1.x.

PR #3293, opened 2020-09-05
Tugboat test

ghost:
THis PR changes CSS in Basis that makes all text red coloured and bold. This is to make sure the changes in the PR are being applied when you click the Tugboat preview link below...

PR #3447, opened 2020-12-14
#3293: Exposed filename filter for media library dialogue

laryn:
PR suggested by @elmuche.

### DIFF
--- a/core/modules/filter/filter.pages.inc
+++ b/core/modules/filter/filter.pages.inc
@@ -184,7 +184,7 @@ function filter_format_editor_image_form($form, &$form_state, $format) {
       $view .= ' ' . t('Check the <a href="!url">Views administration page</a> to enable or create the "@view_name" view as needed.', array('!url' => url('admin/structure/views'), '@view_name' => $view_name));
     }
   }
-
+  $view = str_replace(";page=1&amp;",";page=0&amp;",$view);
   $form['library_view'] = array(
     '#markup' => '<div class="library-view" data-editor-library-view="true">' . $view . '</div>',
   );
--- a/core/modules/filter/js/filter.js
+++ b/core/modules/filter/js/filter.js
@@ -192,6 +192,12 @@ Backdrop.behaviors.editorImageDialog = {
 
       var libraryShown = $('.editor-image-fields').find('[name="attributes[src]"]').is(':visible');
       if (libraryShown) {
+        if ($('ul').hasClass('pager')) {
+          $('li.pager-next a').click(function(){return true;}).click();
+        }
+        else {
+          $('div.view-filters').css({ 'display': 'none' });
+        }
         // Toggle state is set to show 'select an image'
         // so add library view to dialog display.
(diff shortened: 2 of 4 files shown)


core/modules/filter/filter.pages.inc
```diff
@@ -184,7 +184,7 @@ function filter_format_editor_image_form($form, &$form_state, $format) {
       $view .= ' ' . t('Check the <a href="!url">Views administration page</a> to enable or create the "@view_name" view as needed.', array('!url' => url('admin/structure/views'), '@view_name' => $view_name));
     }
   }
-
+  $view = str_replace(";page=1&amp;",";page=0&amp;",$view);
```
klonos, line 187: An inline comment here to explain why we are doing this?

elmuche: Change the link of the pager from page 1 to page 0 so that when clicking through javascript it shows the first page of the library

core/modules/views/js/ajax_view.js
```diff
@@ -57,7 +57,7 @@ Backdrop.views.ajaxView = function(settings) {
   this.settings = settings;
 
   // Add the ajax to exposed forms.
-  this.$exposed_form = $('form#views-exposed-form-'+ settings.view_name.replace(/_/g, '-') + '-' + settings.view_display_id.replace(/_/g, '-'));
+  this.$exposed_form = this.$view.children('.view-filters').children('form');
```
quicksketch, line 60: I believe this may cause Views that use a block for their exposed filters to break, because the ".view-filters" element would be outside of the view wrapper.

core/modules/filter/js/filter.js
```diff
@@ -211,7 +217,7 @@ Backdrop.behaviors.editorImageDialog = {
           $('.editor-image-fields').removeClass('editor-image-fields-full');
           $('.editor-image-library').css({ 'display': 'block' });
           $('.form-item-image-directory').css({ 'display': 'block' });
-
+
```
quicksketch, line 220: Looks like the addition of unintentional whitespace here.
```suggestion

```

core/modules/filter/js/filter.js
```diff
@@ -192,6 +192,12 @@ Backdrop.behaviors.editorImageDialog = {
 
       var libraryShown = $('.editor-image-fields').find('[name="attributes[src]"]').is(':visible');
       if (libraryShown) {
+        if ($('ul').hasClass('pager')) {
+          $('li.pager-next a').click(function(){return true;}).click();
```
quicksketch, line 196: I'm not sure what this click handler is doing. Seems like it's adding a callback that does nothing except for returning "true" and then it's immediately calling it?

If this is needed, this would be better written using `.on('click')` so that it could be removed if needed and prevent double-binding. But maybe we could come up with something else entirely, this seems strange.

quicksketch (commented):
I'm going to pull this down locally and see how it's working. Here are a few items in a read-through...

PR #3448, opened 2020-12-15
Issue #4808: Check theme_links() key before using as class.

### LINKED ISSUE #4808: Check theme_links() key before using as class
The [theme_links()](https://api.backdropcms.org/api/backdrop/core%21includes%21theme.inc/function/theme_links/1) function uses the `$key` of each link as its class. However if a simple array of links is passed in (i.e. if the key is just an integer), then the class is just `_`. We should check that first before outputting classes of `_`...

ghost:
FIxes https://github.com/backdrop/backdrop-issues/issues/4808


core/includes/theme.inc
```diff
@@ -1896,7 +1896,10 @@ function theme_links($variables) {
 
       $class = array();
       // Use the array key as class name.
-      $class[] = backdrop_html_class($key);
+      $key_class = backdrop_html_class($key);
+      if ($key_class != '_') {
```
indigoxela, line 1900: Isn't that very specific to the styleguide module?

That module's quite buggy, although usable (I use it regularly).

ghost: Not really. Any call to `theme_links()` that doesn't include array keys will cause this:

```php
dpm(theme_links(array('links' => array(
  array(
    'title' => 'one',
    'href' => '<front>',
  ),
  array(
    'title' => 'two',
    'href' => '<front>',
  ),
  array(
    'title' => 'three',
    'href' => '<front>',
  ),
))));
```

indigoxela: > Any call to theme_links() that doesn't include array keys...

That's a weird example, completely against the [official docs](https://docs.backdropcms.org/api/backdrop/core%21includes%21theme.inc/function/theme_links/1).

Your example throws two php Notices. Is the styleguide module doing it that way?

UPDATE: yes, apparently it does. :disappointed:

ghost: I don't know why the above throws those notices, I thought the missing things were optional...

PR #3449, opened 2020-12-16
Issue #4728: Add a 'Field has value' visibility condition for Layouts.

### LINKED ISSUE #4728: Expose Fields to Visibility Conditions
**Description of the need**

Long story short, I'm coming from D7 Panels to BackDrop Layouts. It's a bit of an adjustment, but I'm finding that "most" of the things that are done in Panels, can be done pretty easily in the BackDrop (BD here out) Layouts. Except for visibility conditions on Fields.

I started a discussion on Zulip to see if this was possible; I was given some direction and found that it is not possible to expose visibility conditions on blocks from individual fields. Full discussion can be found on Zulip: https://backdrop.zulipchat.com/#narrow/stream/218635-Backdrop/topic/Panels

If you're familiar with Drupal 7 Panels, you add a field as a "panel pane" into the panel. Once it has been placed, you can add what's called a "Visibility Rule" which is a lot like the BD "Visibility Conditions". The one difference is that the panels visibility rules will expose any field as a visibility condition, based on the criteria of any other field. This allows for some crazy powerful output of the node based on the criteria of the exposed fields data.

I have views added to the Panel variants where a visibility rule is set to show or hide that view based on the value (or empty/null value) of a particular field. I also have fields that are set to show or hide based on the value (or empty/null value) of another field.

For example, let's say you have an image field and a dedicated text field that is a short description of the image. If the image doesn't exist on that page, then why show the description? With panels, you simply set the visibility on the description field to "show if -> image field NOT empty". It's a loose example, but gives the point.

**Proposed solution**

What we need is a block visibility condition that is tied to the fields and their field data within the layout context.

This should probably mirror the way Panels does it:
1) A setting where the condition either matches to one condition or all conditions. (1 must pass or all must pass)
2) A Boolean value to check against true or false. Example: "Field data contains -> apples" vs "Field data NOT contains -> apples"
3) Needs to accept empty/null data values (very important for my workflow). Example: "Show if NOT Empty". I use this regularly for entity reference fields and paragraph reference fields where I hide the entire pane (block) if they are empty. This cuts down the output of the page by processing it, finding the stuff that doesn't need to be added to the ouput, then only showing the stuff that actually needs to be visible. Otherwise, I end up with a bunch of empty DIVS in the HTML. "If it doesn't have data, then don't show me in the html code where it's supposed to have data."

The best example as I have already described, is the Drupal Panels module.

Directly, unfortunately no.

**Additional information**

If screenshots are needed for further discussion, we'll go from there.

An additional thought is to have a way to maybe search for the fields on the page within the visibility condition. I would imagine that the dropdown would say "field" (where you have URL Path, Node: Nid etc). When selected and added, the next screen would show a list of the fields with a search box to filter them. Once choosing that field the next screen will show the data to choose from for the "criteria".

One thing panels didn't do that I always hated, was allow more than one value. For example, if I have a list field, I have to create a new rule for each and every key value in the list. I always wished that I could select more than one. The same should be applied to other fields like references, taxonomy etc. For example the condition I want is to show fruits. I would like to be able to select apples, oranges and grapes for that one condition even though carrots, potatoes and green beans are also within that fields data set.

---

PR: https://github.com/backdrop/backdrop/pull/3449

jenlampton:
Fixes https://github.com/backdrop/backdrop-issues/issues/4728
IAS

### DIFF
--- a/core/modules/layout/includes/layout.layout.inc
+++ b/core/modules/layout/includes/layout.layout.inc
@@ -100,13 +100,18 @@ function layout_layout_access_info() {
   // Provide type-specific plugins for each entity type.
   foreach ($entity_types as $entity_type => $entity_info) {
 
+    $entity_label = $entity_info['label'];
+    if ($entity_label == 'Node') {
+      $entity_label = 'Content';
+    }
+
     // Create a key such as "node_type" or "comment_node_type".
     // Skip users, which don't support multiple types.
     if (!empty($entity_info['entity keys']['bundle'])) {
-      $key = $entity_type . '_' . $entity_info['entity keys']['bundle'];
+      $bundle_key = $entity_type . '_' . $entity_info['entity keys']['bundle'];
       $bundle_label = array_key_exists('bundle label', $entity_info) ? $entity_info['bundle label'] : 'Type';
-      $info[$key] = array(
-        'title' => t('@entity: @bundle', array('@entity' => $entity_info['label'], '@bundle' => $bundle_label)),
+      $info[$bundle_key] = array(
+        'title' => t('@entity: @bundle', array('@entity' => $entity_label, '@bundle' => $bundle_label)),
         'description' => t('Control access by @entity_type @bundle_label.', array('@entity_type' => $entity_type, '@bundle_label' => backdrop_strtolower($bundle_label))),
         'entity_type' => $entity_type,
         'class' => 'EntityBundleLayoutAccess',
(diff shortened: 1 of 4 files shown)


core/modules/layout/includes/layout.layout.inc
```diff
@@ -85,13 +85,18 @@ function layout_layout_access_info() {
   // Provide type-specific plugins for each entity type.
   foreach ($entity_types as $entity_type => $entity_info) {
 
+    $entity_label = $entity_info['label'];
+    if ($entity_label == 'Node') {
+      $entity_label = 'Content';
```
klonos, line 105:
```suggestion
      $entity_label = t('Content');
```

...and not 100% sure if the translated labels are being stored for for entities in multilingual/translated site. If so, then the conditional should also be `== t('Node')` instead.

core/modules/layout/includes/layout.layout.inc
```diff
@@ -102,19 +107,37 @@ function layout_layout_access_info() {
     }
 
     // Create a entity id key such as "nid" or "cid".
```
klonos, line 124:
```suggestion
    // Create an entity id key such as "nid" or "cid".
```

core/modules/layout/includes/layout.layout.inc
klonos: ...actually this might be config translation territory here.

jenlampton: These are just values in the info at this point, so they shouldn't be translated yet. They'll be translated when they are rendered.

PR #3450, opened 2020-12-18
Issue #4810: Warning: Invalid argument supplied for foreach() in path…

### LINKED ISSUE #4810: Warning: Invalid argument supplied for foreach() in path_entity_load() (line 427 of ../path.module).
**Description of the bug**

A Backdrop website running on PHP 7.1.22 is giving the following warning:

`Warning: Invalid argument supplied for foreach() in path_entity_load() (line 427 of /var/www/docroot/core/modules/path/path.module).
`

Apparently, the warning triggered because the argument `$path_info['entity type']` in [line 427](https://github.com/backdrop/backdrop/blob/f3fa59b6cb8cb15e06e0ccf040fd5bd6a48d9690/core/modules/path/path.module#L427) doesn't exist.

**Additional information**

The offending block at whole looks like:

```
    foreach ($all_path_info as $path_info) {
      if (isset($path_info['entity type'])) {
        $path_entity_types[] = $path_info['entity type'];
      }
    }
```

Now, adding simple check if `$all_path_info` is empty before executing loop like this:

```
    if (!empty($all_path_info)) {
      foreach ($all_path_info as $path_info) {
        if (isset($path_info['entity type'])) {
          $path_entity_types[] = $path_info['entity type'];
        }
      }
    }
```

not only suppresses the warning messages, but saves useless loop from executing.

alanmels:
Resolves https://github.com/backdrop/backdrop-issues/issues/4810

### DIFF
--- a/core/modules/path/path.module
+++ b/core/modules/path/path.module
@@ -423,9 +423,11 @@ function path_entity_load($entities, $type) {
   if (!isset($path_entity_types)) {
     $path_entity_types = array();
     $all_path_info = path_get_info();
-    foreach ($all_path_info as $path_info) {
-      if (isset($path_info['entity type'])) {
-        $path_entity_types[] = $path_info['entity type'];
+    if (!empty($all_path_info)) {
+      foreach ($all_path_info as $path_info) {
+        if (isset($path_info['entity type'])) {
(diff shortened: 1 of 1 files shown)


klonos (approved):
LGTM 👍

PR #3451, opened 2020-12-19
Issue #2714: Merge all the 'Update' pages

### LINKED ISSUE #2714: [UX] Merge all the "Update core/module/theme/layout" pages into a single page.

Currently, in the "Available Updates report" page, the site admin can see which projects have updates, but the actions they can do are limited to a) downloading the .zip file of an update and b) visit the release notes page of that update.

In the meantime, we have the issue where the respective "Update xyz" pages for modules (`admin/modules/update`), themes (`admin/appearance/update`) and layouts (`admin/structure/layouts/update`) show all available updates of all project types **plus core**. So, I fail to see how they belong to the sections we currently have them in. Notice for example in the screenshot below how in the "Update themes" page, there is an update for a module as well as an update for core listed:

image

True, we could filter the results in those pages by project type, and the UX WTF will be gone. But I think that we have better options...

In the same screenshot, I like how we have the "Download these updates" button there, but we also have the installation queue in the Project Browser. Instead of this screen where the user needs to click the checkboxes of available updates, they could use the "Add to installation queue" links on the Project Browser and then go through the familiar process of the Project Installer. All we'd have to do is add "Up to date", "Security update required!" and "Update available" to the existing "Already enabled" and "Already downloaded" statuses of installed projects. Rough mockup of how this could look:

image

Adding to that, the "Includes: [enabled]/[disabled]" information that we have in the "Available updates" page for installed projects can be appended to the project descriptions in the Project Browser. We could finally have the entries in the Project Browser be color-coded (green/orange/red) based on update status.

### Steps to reproduce/Actual behavior

1. Intentionally install and enable an old version of a module.
2. Go to the "Available updates" page -> you can see that there are updates available, but you cannot install them from that page.
3. Go to `admin/appearance/update` -> the available update is listed in the "Update themes" page although it is a module update.

1. The user visits a single page where all projects are listed (both installed and new).
2. The user can filter by an "Updates available" filer, that is only made available if there are updates for installed projects.
3. The user can filter by a "Project type" filter.
4. The user can filter by installed/new projects.
5. The user can use the "Add to installation queue" link to add the module to the queue.
6. The old "Available updates" link in the admin bar redirects to the Project Browser page with the "Updates available" filter passed on as a parameter.

---

Advocate: @klonos 

---

PR: https://github.com/backdrop/backdrop/pull/3451

klonos:
Fixes https://github.com/backdrop/backdrop-issues/issues/2714


core/modules/update/update.compare.inc
```diff
@@ -194,24 +197,26 @@ function _update_process_info_list(&$projects, $list, $project_type, $status) {
     // or theme. If the project name is 'backdrop', we don't want it to show up
     // under the usual "Modules" section, we put it at a special "Core"
     // section at the top of the report.
-    if ($project_name == 'backdrop') {
-      $project_display_type = 'core';
-    }
-    else {
-      $project_display_type = $project_type;
-    }
+if ($project_name == 'backdrop') {
```
indigoxela: Not relevant for functionality, but I think this indentation change was unintentional. :wink:

klonos: It is intentional 🙂 ...so that I know to return to this bit of code for some additional cleanup. It will be fixed during the final polishing of this PR.

indigoxela: Ah, it was a "marker", got it. :+1:

core/misc/tableselect.js
```diff
@@ -20,6 +20,8 @@ Backdrop.tableSelect = function () {
     // Update table's select-all checkbox (and sticky header's if available).
     $(table).prev('table.sticky-header').addBack().find('th.select-all input[type="checkbox"]').each(function() {
       $(this).attr('title', state ? strings.selectNone : strings.selectAll);
+      $(this).toggleClass('select-none', state);
+      $(this).toggleClass('select-all', !state);
```
quicksketch: Is this a separate bug-fix?

core/modules/installer/installer.manager.inc
```diff
@@ -50,37 +50,42 @@
...
-      break;
-    case 'layout':
-      backdrop_set_title('Update layouts');
-      break;
-    case 'report':
-      backdrop_set_title('Update projects');
-      break;
-  }
-
+function installer_manager_update_form($form, $form_state, $context = NULL) {
```
quicksketch: I think you can just remove the `$context` parameter entirely. If someone passes a value anyway PHP will just ignore it. Plus I don't think a page callback like this is likely to be used by any outside modules.

klonos: Agreed 👍  ...and done.

core/misc/tableselect.js
klonos: Nope. Just adding a way to directly and explicitly select the "check all" checkbox in the js that I've added in installer.select_updates.js.

You see, the "select all" checkbox normally only has a `.form-checkbox` class, which also matches any other checkbox in the table. There is a `.select-all` class, but that is being added in the `<th>` that wraps the checkbox, instead of the checkbox itself.

Anyway, I've reverted this, and used `.select-all .form-checkbox:checked` as a selector in my js.

core/modules/dashboard/includes/block.update.inc
```diff
@@ -59,56 +59,123 @@ class DashboardUpdateBlock extends Block {
...
-    $items = array();
+    $header = array(
+      array('data' => t('Project')),
+      array('data' => t('Type')),
+      array('data' => t('Update')),
+    );
+    $updates = array();
+
     $available = update_get_available();
     module_load_include('inc', 'update', 'update.compare');
```
indigoxela, line 90: `$available` can be empty for several reasons, in that case most of the following code is futile (and causes "undefined something" notices).

indigoxela: You could pull the $panel up here and return it early (with appropriate content like "dunno" or "nothing found", possibly a "refresh info" button).

```
    $panel = array(
      '#theme' => 'dashboard_panel__update',
    );
    if (empty($available)) {
      // Useful content for $panel...
      return $panel;
    }
```
Before you're doing the more expensive stuff below.

core/modules/update/update.compare.inc
klonos: This one was reverted. I'll implement the clean-up I meant to do in a follow-up.

core/modules/dashboard/includes/block.update.inc
klonos: I cannot reproduce the issue any more. I'm suspecting that this has fixed it: [`b40cb52` (#3451)](https://github.com/backdrop/backdrop/pull/3451/commits/b40cb52ee48943df35379edb4b0e5df22f91a9e8)

core/modules/installer/installer.manager.inc
quicksketch: Looks like `$context = NULL` is still there in this latest version?
```suggestion
function installer_manager_update_form($form, $form_state) {
```

PR #3454, opened 2020-12-21
Issue #4821: API consistency with config_*() functions.

jlfranklin:
Still needs some test coverage.


gitressa:
Thanks @jlfranklin, you can link an issue with a PR via the first comment with `Fixes *ISSUE_URL*`, so if you update the syntax in this PR, your issue will probably also be linked.

PR #3455, opened 2020-12-23
Issue #4826: Fix RTL CSS for autocomplete throbber icons.

### LINKED ISSUE #4826: Autocomplete throbber icon doesn't support RTL languages
In https://github.com/backdrop/backdrop-issues/issues/3421 it was pointed out that the autocomplete throbber icon is right-alighed in RTL languages when it should be left-aligned. The solution used back then was to add new CSS to the Seven theme to fix this.

However it seems the root cause is actually incorrect CSS in the System module's CSS (https://github.com/backdrop/backdrop/blob/26101dbdc91f15b1f6b432f1428732bcec77f899/core/modules/system/css/system.css#L37):

```css
[dir="rtl"] .js input.form-autocomplete {
```

The `[dir="rtl"]` and `.js` are both on the same element, yet there's a space between the selectors, so this doesn't work.

ghost:
Fixes https://github.com/backdrop/backdrop-issues/issues/4826

### DIFF
--- a/core/modules/system/css/system.css
+++ b/core/modules/system/css/system.css
@@ -34,14 +34,14 @@
   background-position: 100% center; /* LTR */
   background-repeat: no-repeat;
 }
-[dir="rtl"] .js input.form-autocomplete {
+[dir="rtl"].js input.form-autocomplete {
   background-position: 1% center;
 }
 .js input.throbbing {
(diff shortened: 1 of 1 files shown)


quicksketch:

PR #3456, opened 2020-12-23
Issue #4827: Update Expires header date to use Backdrop's birthday.

### LINKED ISSUE #4827: Change the expires header (remove Dries birthday from Backdrop)
> Should we re-purpose this issue to just change the expires header, or should we open a new issue for that?

I think a new issue might make sense. The issue description contains a lot of info we don't need for the small expires change and we shouldn't lose it either.

_Originally posted by @indigoxela in https://github.com/backdrop/backdrop-issues/issues/4817#issuecomment-750366824_

ghost:
Fixes https://github.com/backdrop/backdrop-issues/issues/4827

### DIFF
--- a/core/modules/node/tests/node.test
+++ b/core/modules/node/tests/node.test
@@ -230,7 +230,7 @@ class NodeRevisionsTestCase extends BackdropWebTestCase {
     foreach ($logs as $log) {
       $this->assertText($log, t('Log message found.'));
     }
- 
+
     // Confirm that this is the active revision.
     $this->assertTrue($node->isActiveRevision(), 'Third node revision is the current one.');
 
@@ -243,7 +243,7 @@ class NodeRevisionsTestCase extends BackdropWebTestCase {
     )), 'Revision reverted.');
     $reverted_node = node_load($node->nid);
     $this->assertTrue(($nodes[1]->body[LANGUAGE_NONE][0]['value'] == $reverted_node->body[LANGUAGE_NONE][0]['value']), 'Node reverted correctly.');
- 
+
     // Confirm that this is not the current version.
     $node = node_load($node->nid, $node->vid);
     $this->assertFalse($node->isActiveRevision(), 'Third node revision is not the current one.');
@@ -723,7 +723,7 @@ class PagePreviewTestCase extends BackdropWebTestCase {
 
     $this->assertNotEqual($node->scheduled, 0, 'The node has a non zero published date.');
 
-    // Test that a user with permissions to create, but not edit, can still 
+    // Test that a user with permissions to create, but not edit, can still
     // access preview.
     $no_edit_user = $this->backdropCreateUser(array(
       'create page content',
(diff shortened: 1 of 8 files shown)


alanmels (commented):
I see you replaced instances of `dries` with `v1.0.0`. Will this be ok, when we move to v2 and further? Otherwise LGTM.

ghost:
@alanmels Yeah, 'v1.0.0' is just the label for the date (I.e. when version 1 was released).

klonos (approved):
LGTM 👍

ghost:
Gonna try re-running tests...

core/modules/node/tests/node.test
```diff
@@ -1558,31 +1558,31 @@ class NodeSaveTestCase extends BackdropWebTestCase {
...
-    $this->assertEqual($node->created, 280299600, 'Creating a node uses user-set "created" timestamp.');
-    $this->assertNotEqual($node->changed, 979534800, 'Creating a node doesn\'t use user-set "changed" timestamp.');
+    $this->assertEqual($node->created, 1421394600, 'Creating a node uses user-set "created" timestamp.');
+    $this->assertNotEqual($node->changed, 1474011660, 'Creating a node doesn\'t use user-set "changed" timestamp.');
 
     // Update the timestamps.
-    $node->created = 979534800;
-    $node->changed = 280299600;
+    $node->created = 1474011660;
+    $node->changed = 1421394600;
```
indigoxela, line 1580: Oops... changed is before created. :wink: 

```
> date -d @1474011660
Fr 16. Sep 09:41:00 CEST 2016
> date -d @1421394600
Fr 16. Jan 08:50:00 CET 2015
```
Check the others, too...

ghost: I don't see how the changes are different from the original...

created: 979534800 = 01/15/2001
changed: 280299600 = 11/19/1978

created: 1474011660 = 09/16/2016
changed: 1421394600 = 01/16/2015

indigoxela: I'm just trying to be helpful, and this timestamp thing was the first that caught my eye.

core/modules/simpletest/tests/form_test.module
```diff
@@ -1789,14 +1789,14 @@ function _form_test_disabled_elements($form, &$form_state) {
     '#title' => 'date',
     '#disabled' => TRUE,
     '#default_value' => array(
-      'day' => 19,
-      'month' => 11,
-      'year' => 1978,
+      'day' => 16,
+      'month' => 01,
```
indigoxela: Ah sorry for my previous comment, the actual problem is the number with starting "0"...

Numbers starting with 0 is considered octal values, either wrap it in `' '` or remove the leading zero. I'd suggest to use `'`.

core/modules/node/tests/node.test
indigoxela: You're right, :+1:  this is unrelated. Mark this comment as resolved (I can't).

core/modules/simpletest/tests/form_test.module
klonos: Good catch @indigoxela 👍 ...why use `'` though? (which converts the number to string AFAIK)

indigoxela: > why use ' though?

If values are provided via form elements, they're usually strings (in `$form_state['values']`).

ghost: Thanks @indigoxela! Fixed.

quicksketch:
Merged into 1.x for 1.18.0. Thanks!

PR #3458, opened 2020-12-24
Issue #4814: Save default theme settings to config when read from info files.

### LINKED ISSUE #4814: [DX] Automate saving of default theme settings values to config.
I'm reading through [Adding theme settings (advanced)](https://api.backdropcms.org/documentation/adding-theme-settings-advanced) in order to better understand theme settings (something which I rarely use/need), and am confused by this part near the end of that page:

> **Setting default values**
> In order to set the default value for any form element you add, you’ll need to add a simple line to your .info file: `settings[SETTING_NAME] = DEFAULT_VALUE`.

> **Initializing the default values**
> Add the following code near the top of your `template.php` file:
> ```php
> if (is_null(theme_get_setting('menu_tab_color'))) {
>   config_set('foo.settings', 'menu_tab_color', 'blue');
> }
> ```

It seems that we need to set the default value of our theme settings twice - once in the info file, and once in `template.php`. Why is this necessary?

If both places require a default value (i.e. if this isn't just outdated information), then can't we automate the latter? Can't we make it so that when a theme is installed/enabled, the info file is read, a config file is created, and default values are automatically populated?

It just seems unnecessary having to specify default values twice...

ghost:
Fixes https://github.com/backdrop/backdrop-issues/issues/4814

### DIFF
--- a/core/includes/theme.inc
+++ b/core/includes/theme.inc
@@ -1466,7 +1466,14 @@ function theme_get_setting($setting_name, $theme = NULL) {
 
   // Loop up through any base themes.
   if (is_null($value) && isset($theme_info->info['base theme'])) {
-    $value = theme_get_setting($setting_name, $theme_info->info['base theme']);
+    $theme = $theme_info->info['base theme'];
+    $value = theme_get_setting($setting_name, $theme);
+  }
+
+  // Save the setting to config.
+  if (!is_null($value)) {
+    $configs[$theme]->set($setting_name, $value);
+    $configs[$theme]->save();
   }
 
   return $value;


core/includes/theme.inc
```diff
@@ -1466,7 +1466,14 @@ function theme_get_setting($setting_name, $theme = NULL) {
...
   // Loop up through any base themes.
   if (is_null($value) && isset($theme_info->info['base theme'])) {
-    $value = theme_get_setting($setting_name, $theme_info->info['base theme']);
+    $theme = $theme_info->info['base theme'];
+    $value = theme_get_setting($setting_name, $theme);
+  }
+
+  // Save the setting to config.
+  if (!is_null($value)) {
+    $configs[$theme]->set($setting_name, $value);
```
jenlampton, line 1475: This would copy a setting from the base-theme to the sub-theme, but it wouldn't change the behavior at all because the base-theme already provides this value if it's not in the sub-theme. 

If we do want the settings from a base-theme to be copied to the config file for the sub-theme then we should do them all at once when the sub-theme is enabled. It should't happen one setting at a time, and certainly not during a read.

ghost:
This turned out to be unnecessary.

PR #3461, opened 2020-12-27
Issue #4730: Document States API OR/XOR conditions.

### LINKED ISSUE #4730: Document ability to use OR conditional in #states
When using Form API `#states` to make fields visible/hidden, etc. based on the value(s) of other fields, you can use an `OR` conditional to, say, show this field when this other field's value is 'x' OR that other field's value is 'y'. However this ability to use `OR` is un-documented in Backdrop. Drupal has a change record for it though: https://www.drupal.org/node/1464758 (which works in Backdrop too).

Rather than having to consult the Drupal CR each time I implement this (because I can never remember the right syntax), let's document it in [backdrop_process_states()](https://github.com/backdrop/backdrop/blob/1.x/core/includes/common.inc#L5199)

ghost:
Fixes https://github.com/backdrop/backdrop-issues/issues/4730

### DIFF
--- a/core/includes/common.inc
+++ b/core/includes/common.inc
@@ -5078,8 +5078,8 @@ function backdrop_process_attached($elements, $group = JS_DEFAULT, $dependency_c
  *
  * A "state" means a certain property on a DOM element, such as "visible" or
  * "checked". A state can be applied to an element, depending on the state of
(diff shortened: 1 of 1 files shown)


quicksketch:

PR #3463, opened 2020-12-29
Issue #4511: Language names not translated to their native names

### LINKED ISSUE #4511: Language names not translated to their native names
One thing baffles me: the language names themselves are not translated.

With one notable exception: the installer uses the native name.
Everywhere else the English names appear (language select list, language switcher block, language overview list, translation page...)

<strike>I wonder what the reason for this is.</strike>

The reason for this is a breaking change in early Drupal 8:

- Breaking commit 879b5357cd6c8cb9a62ee63e5b06e72b82fa05ae
- Drupal Issue: https://www.drupal.org/project/drupal/issues/1301148
- Drupal patch: https://www.drupal.org/files/kill-fake-locale-config-translation-13.patch
- Drupal changelog: https://www.drupal.org/node/1323146

----
Will also address this bug (moved from https://github.com/backdrop/backdrop-issues/issues/4838):

1. Enable locale, language modules
2. Create a view of type "Locale source"
3. Add a field "Locale target: Language", set display to "Native language", save and refresh preview

You get a bunch of:

> Notice: Undefined property: stdClass::$native in views_language_list() (line 752 of .../core/modules/views/views.module).

Reason: The "native" property has been removed from the $language object

indigoxela:
Fixes https://github.com/backdrop/backdrop-issues/issues/4511

Restores the "native" property for language objects.

### DIFF
--- a/core/includes/bootstrap.inc
+++ b/core/includes/bootstrap.inc
@@ -3527,7 +3527,7 @@ function language_multilingual() {
  * @return (stdClass|string)[]
  *   An associative array of languages, keyed by the language code.
  */
-function language_list($only_enabled = FALSE, $option_list = FALSE) {
+function language_list($only_enabled = FALSE, $option_list = FALSE, $native_names = FALSE) {
   $languages = &backdrop_static(__FUNCTION__);
   // Initialize master language list.
   if (!isset($languages)) {
@@ -3560,6 +3560,7 @@ function language_list($only_enabled = FALSE, $option_list = FALSE) {
       $languages['all']['en'] = (object) array(
         'langcode' => 'en',
         'name' => 'English',
+        'native' => 'English',
         'direction' => 0,
         'enabled' => TRUE,
         'weight' => 0,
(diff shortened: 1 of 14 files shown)


core/includes/install.core.inc
```diff
@@ -1589,6 +1589,7 @@ function install_import_translations(&$install_state) {
     $language = (object) array(
       'langcode' => $langcode,
       'name' => $langcode,
+      'native' => $langcode,
```
klonos, line 1592: While at it, would you please change the `The user will be able to edit afterwards.` bit in the inline comment above to something like this instead?: `If required, this can be edited post-installation.`

core/includes/bootstrap.inc
```diff
@@ -3527,7 +3527,7 @@ function language_multilingual() {
  * @return (stdClass|string)[]
  *   An associative array of languages, keyed by the language code.
  */
-function language_list($only_enabled = FALSE, $option_list = FALSE) {
+function language_list($only_enabled = FALSE, $option_list = FALSE, $native_names = FALSE) {
```
klonos, line 3533: The new parameter needs to be added to the `@param` list in the function docblock.

indigoxela:
Requesting fresh sandbox (unrelated test failures).

PR #3464, opened 2020-12-31
Issue #4842: Fix error re. colour theme settings.

### LINKED ISSUE #4842: Prevent undefined index 'settings' in theme_has_color_support()
**Description of the bug**

The error `Undefined index: settings in theme_has_color_support()` is displayed for default themes without any settings.

**Steps To Reproduce**

To reproduce the behavior:
1. Find a theme that doesn't have any settings (colour or otherwise) - e.g. Seven
2. Make it the default theme for the site
3. Go to the Dashboard

**Actual behavior**

`Undefined index: settings in theme_has_color_support()` error displayed.

image

**Expected behavior**

No error displayed.

ghost:
Fixes https://github.com/backdrop/backdrop-issues/issues/4842

### DIFF
--- a/core/includes/theme.inc
+++ b/core/includes/theme.inc
@@ -1515,7 +1515,7 @@ function theme_has_settings($theme) {
 function theme_has_color_support($theme) {
   $themes = list_themes();
   $theme_info = $themes[$theme];
-  if ($theme_info->info['settings']['color'] && file_exists(backdrop_get_path('theme', $theme_info->name) . '/color/color.inc')) {
+  if (!empty($theme_info->info['settings']['color']) && file_exists(backdrop_get_path('theme', $theme_info->name) . '/color/color.inc')) {
     return TRUE;
   }
   if (isset($theme_info->info['base theme'])) {


core/includes/theme.inc
```diff
@@ -1515,7 +1515,7 @@ function theme_has_settings($theme) {
 function theme_has_color_support($theme) {
   $themes = list_themes();
   $theme_info = $themes[$theme];
-  if ($theme_info->info['settings']['color'] && file_exists(backdrop_get_path('theme', $theme_info->name) . '/color/color.inc')) {
+  if (!empty($theme_info->info['settings']['color']) && file_exists(backdrop_get_path('theme', $theme_info->name) . '/color/color.inc')) {
```
klonos, line 1518: Shouldn't we also be checking whether color setting is set to `TRUE`? ...something like this perhaps:

```suggestion
  $theme_has_color_settings = isset($theme_info->info['settings']['color']) ? $theme_info->info['settings']['color'] : FALSE;
  $theme_has_color_inc = file_exists(backdrop_get_path('theme', $theme_info->name) . '/color/color.inc');
  if ($theme_has_color_settings && $theme_has_color_inc) {
```

ghost: > Shouldn't we also be checking whether color setting is set to TRUE?

That's what `!empty($theme_info->info['settings']['color'])` does. If it's set to TRUE/1, `!empty()` returns TRUE. If it's set to FALSE/0/not set at all, `!empty()` returns FALSE.

indigoxela: I agree with @BWPanda that `!empty()` does the right check here in function theme_has_color_support().

klonos: 👍

quicksketch:

PR #3467, opened 2021-01-01
Issue #4849: Making changes to default content on front page of site.

### LINKED ISSUE #4849: [UX] Add a new pageless node above the "Promoted Content" block on Home page
**Description of the need**

We've had a number of frustrated users report problems understanding how to use the home page layout in Backdrop CMS. Coming from Drupal, many users assume the need to change the default home page path to a node and then are confused when the home page layout is not working, because the new home page is a node and uses the node layout.

In todays [Design meeting](https://youtu.be/Fc-oQ55LQi0) we talked about this problem. The discussion spilled over after the dev meeting (after recording stopped). During the discussion, we discussed the possibility of creating a "home" or "welcome" page and then using this as part of the default content on the home page. 

This idea is based on a couple of assumptions:

1) That very seldom do users really need to change the default path for the home page. But, that there are things we could do to help new users understand this. Users coming from Drupal 7 are especially vulnerable to the assumption that they need to change the home page path to other content.
2) That most people do not want a feed of promoted content on the front page (well, this is my assumption, not sure how others feel about this). 

By putting an "existing content" block on the front page by default, we make it much easier for them to understand and proceed with customizing the home page layout, rather than changing the default path for the home page to a node. 

**Proposed solution**

I decided to give this whole thing a try to see what folks think. 

1) I created a new node of type page with the title of "Home page" and some helpful content as part of the install profile.
2) I replaced the "Promoted content block" in the default home page layout with an "Existing content" block that refers to the new "Home page" node. The node title is NOT used on the home page layout.
3) As per a suggestion by @philsward - I made sure that the new "Home page" node has NID 1 (I don't know if this is important and might not be a good idea).

Putting a new "existing content" block on the front page with the "promoted content block" seemed a bit cluttered. I am certainly open to other options. 

All of the changes in this PR effect the install profile and will only impact new sites. They will have no impact on any existing sites.

I went ahead and added a little text to the home page designed to introduce the idea of using the home page layout to new users. I'm not sure if folks will like this idea, but I wanted to give it a try and see what the feedback was - and listen for alternative ideas.

We could also improve this situation with some help text. 
https://github.com/backdrop/backdrop-issues/issues/4839

**Additional information**

home-page-new

stpaultim:
Fixes backdrop/backdrop-issues#4849


core/profiles/standard/config/layout.layout.home.json
```diff
@@ -114,19 +115,27 @@
...
                 "settings": {
-                    "title_display": "default",
+                    "title_display": "none",
                     "title": "",
                     "style": "default",
                     "block_settings": [],
-                    "contexts": []
+                    "contexts": [],
+                    "admin_label": "Home Page Content",
+                    "admin_description": "A node with content for the home page.",
```
klonos, line 130: We should not be using the word "node" in user-facing text. How about this instead?:

```suggestion
                    "admin_description": "A piece of content used as the home page.",
```

PR #3469, opened 2021-01-02
Issue #4850: Clean up unused db tables, if failed field creation

### LINKED ISSUE #4850: Properly clean up unused db tables, resulting from failed field creation
This is related to #2585 and the respective issue as https://www.drupal.org/project/drupal/issues/1343330 (which is slated for backporting to D7):

> ## Problem/Motivation
> When field creation via GUI is failed, you may have in your DB unused table(s).
>
> ## How to reproduce
> The simplest way to imitate/reproduce the case is:
> 
> - in your DB create a dummy table with name `field_revision_field_sample`;
> - open `admin/structure/types/manage/page/fields`, try to add field `field_sample` with text type;
> 
> As a result you will have newly created table `field_data_field_sample` in your database which not belongs to any field.

> ## Proposed resolution
> Similar to action of updating field, we enclose tables creation in transaction, either manual or provided by DB. See `field_sql_storage_field_storage_update_field()`.
> 
> ## User interface changes
> Also patch removes incorrectly displaying `'success'` message `'Your settings have been saved.'` when fields cannot be created.

klonos:
Fixes https://github.com/backdrop/backdrop-issues/issues/4850

Adapted from the patch in https://www.drupal.org/files/drupalcore-create-field-cleanup-1343330-5.diff

### DIFF
--- a/core/modules/field_ui/tests/field_ui.test
+++ b/core/modules/field_ui/tests/field_ui.test
@@ -805,9 +805,9 @@ class FieldUIAlterTestCase extends BackdropWebTestCase {
       ),
     ));
 
-    // Test that field_test_field_widget_properties_alter() sets the size to
-    // 42 and that field_test_field_widget_form_alter() reports the correct
-    // size when the form is displayed.
+    // Test that field_test_field_widget_properties_alter() sets the size to 42,
+    // and that field_test_field_widget_form_alter() reports the correct size
+    // when the form is displayed.
     $this->backdropGet('admin/structure/types/manage/post/fields/alter_test_text');
     $this->assertText('Field size: 42', 'Altered field size is found in hook_field_widget_form_alter().');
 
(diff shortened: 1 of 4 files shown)


core/modules/field_ui/tests/field_ui.test
```diff
@@ -846,6 +846,67 @@ class FieldUIAlterTestCase extends BackdropWebTestCase {
   }
 }
 
+/**
+ * Tests for correct messaging, and DB cleanup when field creation fails.
+ *
+ * @see based on FieldUIManageFields class.
+ */
+class FieldUIManageFieldsFailTestCase extends BAckdropWebTestCase {
```
indigoxela: There's a typo: this should be `extends BackdropWebTestCase`

core/modules/field_ui/tests/field_ui.test
```diff
@@ -846,6 +846,67 @@ class FieldUIAlterTestCase extends BackdropWebTestCase {
...
+  function setUp() {
+    parent::setUp();
+
+    $admin_user = $this->backdropCreateUser(array('access content', 'administer content types', 'administer users'));
```
indigoxela: Tests are failing because your `$admin_user` also needs the `'administer fields'` permission.

core/modules/field_ui/tests/field_ui.test
klonos: Thanks 🙏 ...fixed.

core/modules/field_ui/tests/field_ui.test
klonos: I rely on ZenCI for tests, and you know how it has a mind of its own 😓 ...thanks for the recommendation; I've added that permission.

klonos:
ZenCI acting up again...

core/modules/field_ui/tests/field_ui.test
```diff
@@ -846,6 +846,67 @@ class FieldUIAlterTestCase extends BackdropWebTestCase {
   }
 }
 
+/**
+ * Tests for correct messaging, and DB cleanup when field creation fails.
+ *
+ * @see based on FieldUIManageFields class.
+ */
+class FieldUIManageFieldsFailTestCase extends BackdropWebTestCase {
```
indigoxela, line 854: This is the complete output of phpcs for this section:

```
  855 | ERROR   | Missing function doc comment
  855 | ERROR   | Visibility must be declared on method "setUp"
  858 | ERROR   | If the line declaring an array spans longer than 80
      |         | characters, each element should be broken into its own line
  870 | ERROR   | Expected 1 space after "="; 2 found
  889 | ERROR   | Visibility must be declared on method
      |         | "testCreateFieldFailureCleanup"
  900 | ERROR   | Expected 1 space after comma in function call; 2 found
```

core/modules/field_ui/tests/field_ui.test
```diff
@@ -846,6 +846,67 @@ class FieldUIAlterTestCase extends BackdropWebTestCase {
...
+  function setUp() {
+    parent::setUp();
+
+    $admin_user = $this->backdropCreateUser(array('access content', 'administer content types', 'administer fields', 'administer users'));
```
indigoxela, line 858: It turned out that the admin user needs only two permissions. As this still would lead to a line wider than 80 chars:

```
    $admin_user = $this->backdropCreateUser(array(
      'administer content types',
      'administer fields',
    ));
```
(That's what "each element should be broken into its own line" means, but I actually think you know that anyway. :wink: )

PR #3473, opened 2021-01-03
Issue #3402: Standardise the display of labels and machine names.

### LINKED ISSUE #3402: [UX] Standardize Label and `machine name` usage in admin listings

ghost:
Fixes https://github.com/backdrop/backdrop-issues/issues/3402
Replaces https://github.com/backdrop/backdrop/pull/3128

### DIFF
--- a/core/modules/field_ui/field_ui.theme.inc
+++ b/core/modules/field_ui/field_ui.theme.inc
@@ -100,16 +100,17 @@ function theme_field_ui_view_modes($variables) {
   $element = $variables['element'];
   $view_modes = $element['#view_modes'];
   $entity_type = $element['#entity_type'];
+  $entity_bundle = $element['#bundle'];
   $path = $element['#admin_path'] . '/display';
 
-  // Get the node type machine name.
-  $node_type = '';
-  foreach (node_type_get_types() as $node_type_name => $node_type) {
-    if ($node_type->name == $element['#bundle']) {
-      $node_type = $node_type_name;
-      break;
-    }
+  // Get the machine name of the entity bundle.
+  $entity_info = entity_get_info($entity_type);
+  $bundle_machine_name = '';
+  $bundles_by_label = array();
+  foreach ($entity_info['bundles'] as $bundle => $bundle_info) {
+    $bundles_by_label[$bundle_info['label']] = $bundle;
   }
+  $bundle_machine_name = $bundles_by_label[$entity_bundle];
 
   $rows = array();
 
@@ -124,7 +125,7 @@ function theme_field_ui_view_modes($variables) {
 
   // Add the row for the default display mode.
   $row = array();
-  $row['label'] = theme('label_machine_name__node_display__' . $node_type, array(
+  $row['label'] = theme('label_machine_name__node_display__' . $bundle_machine_name, array(
     'label' => $label,
     'machine_name' => t('default'),
   ));
@@ -151,7 +152,7 @@ function theme_field_ui_view_modes($variables) {
     $row_classes = array('view-mode--' . str_replace('_', '-', $view_mode_name));
     $row = array();
 
-    $row['label'] = theme('label_machine_name__node_display__' . $node_type, array(
+    $row['label'] = theme('label_machine_name__node_display__' . $bundle_machine_name, array(
       'label' => $view_mode_info['label'],
       'machine_name' => $view_mode_name,
     ));
--- a/core/modules/field_ui/field_ui.theme.inc
+++ b/core/modules/field_ui/field_ui.theme.inc
@@ -102,6 +102,15 @@ function theme_field_ui_view_modes($variables) {
   $entity_type = $element['#entity_type'];
   $path = $element['#admin_path'] . '/display';
 
+  // Get the node type machine name.
+  $node_type = '';
+  foreach (node_type_get_types() as $node_type_name => $node_type) {
+    if ($node_type->name == $element['#bundle']) {
+      $node_type = $node_type_name;
+      break;
+    }
+  }
+
   $rows = array();
 
   $header = array(
@@ -115,7 +124,7 @@ function theme_field_ui_view_modes($variables) {
 
   // Add the row for the default display mode.
   $row = array();
-  $row['label'] = theme('label_machine_name__node_display__' . backdrop_strtolower($element['#bundle']), array(
+  $row['label'] = theme('label_machine_name__node_display__' . $node_type, array(
     'label' => $label,
     'machine_name' => t('default'),
   ));
@@ -142,7 +151,7 @@ function theme_field_ui_view_modes($variables) {
     $row_classes = array('view-mode--' . str_replace('_', '-', $view_mode_name));
     $row = array();
 
-    $row['label'] = theme('label_machine_name__node_display__' . backdrop_strtolower($element['#bundle']), array(
+    $row['label'] = theme('label_machine_name__node_display__' . $node_type, array(
       'label' => $view_mode_info['label'],
       'machine_name' => $view_mode_name,
     ));
--- a/core/modules/field_ui/field_ui.theme.inc
+++ b/core/modules/field_ui/field_ui.theme.inc
@@ -106,7 +106,7 @@ function theme_field_ui_view_modes($variables) {
 
   $header = array(
     array('data' => t('Display modes')),
-    array('data' => t('Machine name'), 'class' => array('priority-low')),
+    array('data' => t('Description'), 'class' => array('priority-low')),
     array('data' => t('Operations'), 'class' => array('operations')),
   );
 
@@ -115,8 +115,11 @@ function theme_field_ui_view_modes($variables) {
 
   // Add the row for the default display mode.
   $row = array();
-  $row['label'] = $label . '<div class="description">' . $label_desc . '</div>';
-  $row['machine'] = t('default');
+  $row['label'] = theme('label_machine_name__node_display__' . backdrop_strtolower($element['#bundle']), array(
+    'label' => $label,
+    'machine_name' => t('default'),
+  ));
+  $row['description'] = $label_desc;
   $row['operations'] = array(
     'data' => array(
       '#type' => 'dropbutton',
@@ -139,11 +142,14 @@ function theme_field_ui_view_modes($variables) {
     $row_classes = array('view-mode--' . str_replace('_', '-', $view_mode_name));
     $row = array();
 
-    $row['label'] = check_plain(t($view_mode_info['label']));
-    $row['machine'] = $view_mode_name;
+    $row['label'] = theme('label_machine_name__node_display__' . backdrop_strtolower($element['#bundle']), array(
+      'label' => $view_mode_info['label'],
+      'machine_name' => $view_mode_name,
+    ));
 
     $operations = array();
     if ($view_modes[$view_mode_name]['custom settings']) {
+      $row[] = '';
       $customized_count++;
       $operations['manage'] = array(
         'title' => t('Manage display'),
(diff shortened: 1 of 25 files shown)


core/modules/layout/layout.admin.inc
```diff
@@ -65,8 +67,13 @@ function layout_list_page() {
         $storage = t('Custom');
       }
 
+      // Create a link to the settings page.
+      $info = layout_get_layout_template_info($layout->layout_template);
```
klonos: Instead of repeating this call to the function for each specific layout, can we please move this in a `$all_layout_templates` variable (or similar name), and move it outside the `foreach`? We can then use the array to grab the info for each individual layout/template.

core/modules/layout/layout.admin.inc
```diff
@@ -3052,14 +3060,18 @@ function layout_block_list() {
           }
         }
         $usage_list = $usages ? theme('item_list', array('items' => $usages)) : t('Not in use');
-        $block_name = check_plain($block['info']) . ' (' . $delta . ')';
+        // $block_name = check_plain($block['info']) . ' (' . $delta . ')';
```
klonos: I guess this needs to go.

core/modules/system/system.theme.inc
```diff
@@ -798,3 +798,35 @@ function system_preprocess_page_components(&$variables) {
     'page_components__' . $child_delta,
   );
 }
+
+/**
+ * Returns HTML for the display of a label with both human and machine names.
```
```suggestion
 * Returns HTML for the display of an item, with both the human-readable label
 * and its machine name.
```

core/modules/system/system.theme.inc
```diff
@@ -798,3 +798,35 @@ function system_preprocess_page_components(&$variables) {
...
+ *
+ * @param array $variables
+ *   An associative array containing:
+ *   - label: The human-readable name or label.
+ *   - link: (optional) An associative array containing:
+ *     - path: The path to link the label to.
+ *     - options: (optional) An associative array of additional options used by
+ *       the l() function.
+ *   - machine_name: (optional) The machine-safe name or label.
```
klonos: We shouldn't be confusing the use of word "label", which we usually reserve for non-machine names.

```suggestion
 *   - machine_name: (optional) The machine-safe name.
```

core/modules/system/system.theme.inc
```diff
@@ -798,3 +798,35 @@ function system_preprocess_page_components(&$variables) {
...
+    $options = isset($variables['link']['options']) ? $variables['link']['options'] : array();
+    $label = l($variables['label'], $variables['link']['path'], $options);
+  }
+  else {
+    $label = check_plain($variables['label']);
+  }
+  $output = '<div class="label">' . $label . '</div>';
+
+  if (!empty($variables['machine_name'])) {
+    $machine_name = t('Machine name: @machine_name', array('@machine_name' => $variables['machine_name']));
```
klonos, line 816: I'd prefer if the machine name "prefix" was optional. Perhaps another `machine_name_prefix` variable, with a default value of `t('Machine name')`? Then we could do something like:

```suggestion
    $machine_name = isset($variables['machine_name_prefix']) ? $variables['machine_name_prefix'] . ': ' : t('Machine name') . ': ';
    $machine_name .= $variables['machine_name'];
```

Perhaps even allow `NULL`, in which case the prefix would be omitted entirely, and only the actual machine name would be printed.

core/modules/user/user.theme.inc
```diff
@@ -172,15 +172,18 @@ function theme_user_permission_description($variables) {
 function theme_user_admin_roles($variables) {
   $form = $variables['form'];
 
-  $header = array(t('Name'), t('Weight'), t('Operations'));
+  $header = array(t('Name'), t('Status'), t('Weight'), t('Operations'));
```
klonos, line 175: Is this a "status" or a "type"? 🤔

core/modules/field_ui/field_ui.theme.inc
```diff
@@ -158,7 +164,7 @@ function theme_field_ui_view_modes($variables) {
     else {
       $default_count++;
       $row_classes[] = 'uses-default';
-      $row['label'] .= '&nbsp;<small>(<em>' . t('Uses the Default display mode') . '</em>)</small>';
+      $row['description'] = '<em>' . t('Uses the Default display mode') . '</em>';
```
klonos, line 177: The "Default" bit of this text is what should be italicized - not the entire description. Also, if we wanted to be _technically_ accurate, we should say:

```suggestion
      $row['description'] = t('Uses the <em>Default</em> display mode settings');
```

klonos: ...or:

```suggestion
      $row['description'] = t('Uses the settings of the <em>Default</em> display mode');
```

core/modules/system/css/system.theme.css
```diff
@@ -551,3 +551,10 @@ img.align-right {
   display: table;
   clear: both;
 }
+
+/**
+ * Labels & machine names.
+ */
+.machine-name {
+  font-size: smaller;
```
jenlampton, line 559: I don't think we do this anywhere else in core. We should specify an actual size here.

core/modules/field_ui/field_ui.theme.inc
ghost: I feel this would be better addressed as part of https://github.com/backdrop/backdrop-issues/issues/4891#issuecomment-764165596 (comment) (comment added).

core/modules/layout/layout.admin.inc
ghost: Ooh, nice! Good call 😄

core/modules/layout/layout.admin.inc
ghost: Oops, thanks!

core/modules/system/css/system.theme.css
ghost: @jenlampton We do actually:
- https://github.com/backdrop/backdrop/blob/11619724ae5f80a1edaf957d367ce24c24aca4e4/core/themes/seven/css/seven.base.css#L93
- https://github.com/backdrop/backdrop/blob/11619724ae5f80a1edaf957d367ce24c24aca4e4/core/modules/system/css/system.maintenance.css#L9

core/modules/system/system.theme.inc
ghost: Let's discuss in the issue.

core/modules/user/user.theme.inc
ghost: I believe 'status' (like nodes can be published or unpublished, roles can be required or not required). Whether or not a role is required isn't defined by the 'type' of role. Unless we use the term 'type' here, and *also* change the 'Required' text to be 'Default' instead. *Then* it would make more sense to use 'type', but I think that's best left for a separate issue (which you can open if you feel so inclined @klonos ).

core/modules/field_ui/field_ui.theme.inc
```diff
@@ -115,8 +115,11 @@ function theme_field_ui_view_modes($variables) {
 
   // Add the row for the default display mode.
   $row = array();
-  $row['label'] = $label . '<div class="description">' . $label_desc . '</div>';
-  $row['machine'] = t('default');
+  $row['label'] = theme('label_machine_name__node_display__' . backdrop_strtolower($element['#bundle']), array(
```
jenlampton: why is the `strtolower` necessary here? Isn't `$element['#bundle']` the machine name of the bundle? If not, shouldn't we be using that instead of the bundle label?

jenlampton: Oh oh, never mind, this is just the default row. Marking as resolved!

core/modules/field_ui/field_ui.theme.inc
```diff
@@ -139,11 +142,14 @@ function theme_field_ui_view_modes($variables) {
     $row_classes = array('view-mode--' . str_replace('_', '-', $view_mode_name));
     $row = array();
 
-    $row['label'] = check_plain(t($view_mode_info['label']));
-    $row['machine'] = $view_mode_name;
+    $row['label'] = theme('label_machine_name__node_display__' . backdrop_strtolower($element['#bundle']), array(
```
jenlampton: okay, what about this one? ... Isn't `$element['#bundle']` the machine name of the bundle? If not, shouldn't we be using that instead of the bundle label?

core/modules/system/css/system.theme.css
jenlampton: well alright then! :)

core/modules/field_ui/field_ui.theme.inc
ghost: @jenlampton Ok, yes. Took some doing, but I managed to work out how to get the node type machine name (it wasn't in the data provided to this function). PR updated.

core/modules/user/user.theme.inc
```diff
@@ -172,15 +172,18 @@ function theme_user_permission_description($variables) {
...
   foreach (element_children($form['roles']) as $role_name) {
-    $label = $form['roles'][$role_name]['#role']->label;
     $row = array();
+    $row[] = theme('label_machine_name__role', array(
+      'label' => $form['roles'][$role_name]['#role']->label,
+      'machine_name' => $form['roles'][$role_name]['#role']->name,
+    ));
     if (in_array($role_name, array(BACKDROP_ANONYMOUS_ROLE, BACKDROP_AUTHENTICATED_ROLE))) {
-      $row[] = t('@name <em>(required)</em>', array('@name' => $label));
+      $row[] = '<em>' . t('Required') . '</em>';
```
klonos, line 183: Lets not emphasize this:

```suggestion
      $row[] = '<em>' . t('required') . '</em>';
```

core/modules/field_ui/field_ui.theme.inc
```diff
@@ -102,11 +102,20 @@ function theme_field_ui_view_modes($variables) {
...
 
+  // Get the node type machine name.
+  $node_type = '';
+  foreach (node_type_get_types() as $node_type_name => $node_type) {
+    if ($node_type->name == $element['#bundle']) {
+      $node_type = $node_type_name;
+      break;
+    }
+  }
+
```
klonos: This will only work for nodes. Please consider my PR here: https://github.com/BWPanda/backdrop/pull/15

core/modules/user/user.theme.inc
ghost: Do you mean 'emphasize' or 'capitalise'...?

jenlampton:
Replaced by https://github.com/backdrop/backdrop/pull/3710

PR #3474, opened 2021-01-03
Issue #4844: Layout and views menu items undeletable

### LINKED ISSUE #4844: Fix layout and views menu items get undeletable but stay visible once created
### Problem
When creating a menu item through a layout "menu settings" on a single URL path, it creates a "hard coded" link that persists even after removing that link from the menu setting of the layout. This link becomes impossible to remove.

**Reproduce**
1) Navigate to the layout list page: /admin/structure/layouts
2) Edit the "Menu Settings" of a single url layout group (Home is fine, adding and testing other groups produce the same result)
3) Set the "Type" to **Normal Menu Entry**, add a title called TEST and save
4) Visit the menu settings page: /admin/structure/menu/manage/main-menu
5) Notice the new "TEST" menu that can only be edited. There is no option to delete it.
6) Go back to the "Menu Settings" of the layout and change the "Type" back to **No menu entry**
7) Again visit the menu settings page: /admin/structure/menu/manage/main-menu and notice the link is still there with no way to delete it.

**Update:** This also spills over to Views. If a view menu has been created and is later set to "no menu entry", the menu is not removed and there is no way to delete it.

### Archived Description
**Description of the need**

I see this as more of a bug than a feature request, however unless I'm doing something wrong, it appears that the home menu link is referencing /home instead of /

**Proposed solution**

The hard coded menu should always point to the domain root / <front> (unless there are translation?)

The workaround is to create a new menu entry, use <front> as the path and disable the OOTB hard coded Home menu item.

**Additional information**

image
image

Workaround
image

indigoxela:
Fixes https://github.com/backdrop/backdrop-issues/issues/4844

Based on the [Drupal patch](https://www.drupal.org/project/drupal/issues/1079628) provided by joegraduate

### DIFF
--- a/core/includes/menu.inc
+++ b/core/includes/menu.inc
@@ -2968,9 +2968,11 @@ function _menu_link_build($item) {
 function _menu_navigation_links_rebuild($menu) {
   // Add normal and suggested items as links.
   $menu_links = array();
+  $default_menu_link_paths = array();
   foreach ($menu as $path => $item) {
     if ($item['_visible']) {
       $menu_links[$path] = $item;
+      $default_menu_link_paths[] = $path;
       $sort[$path] = $item['_number_parts'];
     }
   }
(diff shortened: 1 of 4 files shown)


ghost:

PR #3486, opened 2021-01-08
Issue #4856: Don't show preview warnings on saved content page.

### LINKED ISSUE #4856: [UX] The "this is a preview" message appears after content has saved
**Description of the bug**

When you preview a node, you see a warning message stating that the node you are viewing has not yet been saved. This is good.

However, when you click the "save" button on that page, you also see the same preview message stating that the node you are viewing has not yet been saved -- even though it has. Additionally, you also see a contradictory message stating that the node was just saved.

preview

**Steps To Reproduce**

To reproduce the behavior:
1. Create a node, save it.
2. Edit the node, hit preview. see message. :)
3. Scroll down, hit "Save". See two contradictory messages :(

**Actual behavior**

two contradictory messages :(

**Expected behavior**

one accurate message :)

klonos:
Fixes https://github.com/backdrop/backdrop-issues/issues/4856

### DIFF
--- a/core/modules/node/node.module
+++ b/core/modules/node/node.module
@@ -1118,6 +1118,11 @@ function node_view(Node $node, $view_mode = 'full', $langcode = NULL) {
     $langcode = $GLOBALS['language_content']->langcode;
   }
 
+  if (isset($node->preview) && ($node->preview == 'Preview')) {
+    backdrop_set_message(t('This is a preview. Links within the page are disabled.'), 'warning', FALSE);
+    backdrop_set_message(t('<strong>Changes are stored temporarily</strong>. Click <em>Save</em> to make your changes permanent, or click <em>Back to content editing</em> to make additional changes.'), 'warning', FALSE);
+  }
+
   // Populate $node->content with a render() array.
   node_build_content($node, $view_mode, $langcode);
 
--- a/core/modules/node/node.pages.inc
+++ b/core/modules/node/node.pages.inc
@@ -729,6 +729,11 @@ function node_preview_banner_form(array $form, array $form_state, Node $node, $n
  * @since 1.11.0
  */
 function node_preview_banner_form_node_submit(array $form, array &$form_state) {
+  // Prevent the "This is a preview." and "Changes are stored temporarily."
+  // warnings (previously set in node_preview_banner_form()) from bleeding onto
+  // the saved node page, after saving content changes from the preview mode.
+  backdrop_get_messages();
+
   form_load_include($form_state, 'inc', 'node', 'node.pages');
   $node = $form_state['values']['node'];
   node_object_prepare($node);
--- a/core/modules/node/node.pages.inc
+++ b/core/modules/node/node.pages.inc
@@ -560,7 +560,7 @@ function node_form_submit($form, &$form_state) {
  * @return
  *   An HTML-formatted string of a node preview.
  *
- * @see node_form_build_preview()
+ * @see node_form_preview()
  *
  * @since 1.11.0
  */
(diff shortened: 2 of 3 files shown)


core/modules/node/node.pages.inc
```diff
@@ -729,6 +729,11 @@ function node_preview_banner_form(array $form, array $form_state, Node $node, $n
  * @since 1.11.0
  */
 function node_preview_banner_form_node_submit(array $form, array &$form_state) {
+  // Prevent the "This is a preview." and "Changes are stored temporarily."
+  // warnings (previously set in node_preview_banner_form()) from bleeding onto
+  // the saved node page, after saving content changes from the preview mode.
+  backdrop_get_messages();
```
ghost: I wonder if it's worth editing the comment here to describe *how* this works... Inline comments are for the benefit of developers, but as a developer I'm not sure how or why this line fixes the problem of duplicated messages...

ghost (changes requested):

core/modules/node/node.pages.inc
klonos: @BWPanda if any developer has a quick look at https://api.backdropcms.org/api/backdrop/1/search/backdrop_get_messages they will see that the default values of the two parameters of that function are basically clearing the message queue. I understand that the name of that function may be misleading, but the help for the second parameter does a pretty good job at explaining things:

> **bool $clear_queue**: (optional) If this is TRUE, the queue will be cleared of messages of the type specified in the $type parameter. Otherwise the queue will be left intact. Defaults to TRUE.

...combined with the help for the first parameter, I believe that it's easy to put one and one together:

> **string $type**: (optional) Limit the messages returned by type. Defaults to NULL, meaning all types.

I may be wrong re thins being obvious, and I shouldn't be making assumptions. I have no idea how to explain this though. Care to provide a suggestion please?

ghost:
```suggestion
  // Clear the warning messages displayed in the preview form
  // (node_preview_banner_form()) to prevent confusion after saving.
  backdrop_get_messages('warning', TRUE);
```

Ok, so this change (hopefully) makes it more obvious that we're only clearing the 'warning' messages (setting the 2nd parameter to `TRUE` explicitly helps bring attention to this possibly-overlooked detail).

klonos: Although I agree with this @BWPanda, this will only work so long as the only type of messages we show in preview mode is warnings. If we happen to add another type on message in the future, then these will bleed onto the save page (which was the whole point of fixing the bug to begin with).

How about using `backdrop_get_messages(NULL, TRUE);` then, which clears all types of messages? That will be more future-proof I believe.

Alternatively, we could do something like this, which explicitly explains what `backdrop_get_messages()` does w/o any parameters passed to it:

```suggestion
  // Use backdrop_get_messages() to clear all messages displayed in the preview
  // form (see node_preview_banner_form()), and prevent them from being
  // displayed in the saved node page.
  backdrop_get_messages();
```

ghost: I was thinking that to be safer we should just clear 'warning' messages. If we ever change the type of messages above, we can change this too as needed. But I was worried about clearing other types of messages that might be set elsewhere on the site and displayed here (for whatever reason), so figured just clearing warnings was better than clearing everything.

Re. the parameters, I feel that explicitly defining them makes the code easier to read/understand. If `backdrop_get_messages()` was better named for the purpose it's being used for here (e.g. `backdrop_clear_messages()`), then that'd be fine to use as-is. But since it's not obvious how it's working in this context without further investigation, then I think the parameters are useful.

klonos: Thanks @BWPanda 🙏 ...I've moved the messages to `node_view()` instead + added some logic to only show them when this is a node preview, so all this is not required any more

klonos:
ZenCI again 😓

ghost (approved):

core/modules/node/node.module
```diff
@@ -1118,6 +1118,11 @@ function node_view(Node $node, $view_mode = 'full', $langcode = NULL) {
     $langcode = $GLOBALS['language_content']->langcode;
   }
 
+  if (isset($node->preview) && ($node->preview == 'Preview')) {
```
ghost, line 1121: Excellent! Much better 😃

klonos: 🙂

quicksketch:

PR #3494, opened 2021-01-14
Issue #4872: Better documentation for module enable/disable/uninstall functions.

### LINKED ISSUE #4872: Better documentation for dependencies when enabling, disabling or uninstalling modules
The [`module_enable()`](https://api.backdropcms.org/api/backdrop/core%21includes%21module.inc/function/module_enable/1), [`module_disable()`](https://api.backdropcms.org/api/backdrop/core%21includes%21module.inc/function/module_disable/1) and [`backdrop_uninstall_modules()`](https://api.backdropcms.org/api/backdrop/core%21includes%21install.inc/function/backdrop_uninstall_modules/1) functions all have a parameter that determines whether dependencies will be taken into account when performing these actions. However, it's not obvious to me whether setting this parameter to `FALSE` will bypass dependency checking altogether and just perform the action regardless of the implications, or whether it simply prevents dependencies being *automatically* added/removed as needed...

For example, `module_enable()` has an `$enable_dependencies` parameter set to `TRUE` by default. The documentation says:

> If TRUE, dependencies will automatically be added and enabled in the correct order. This incurs a significant performance cost, so use FALSE if you know $module_list is already complete and in the correct order.

It also says that the return value of the function is:

> FALSE if one or more dependencies are missing, TRUE otherwise.

So can the function only return FALSE if `$enable_dependencies` is `FALSE`? Will it always return `TRUE` otherwise, because it automatically adds dependencies? What if a dependency hasn't been downloaded and doesn't exist in the filesystem?
Or will it always return `FALSE` if there are dependency issues, and setting `$enable_dependencies` to `FALSE` only serves to make things quicker if it's known that all dependencies are already satisfied?

I'm similarly confused about the two other functions. Especially `backdrop_uninstall_modules()`:

> **$uninstall_dependents**: (optional) If TRUE, the function will check that all modules which depend on the passed-in module list either are already uninstalled or contained in the list, and it will ensure that the modules are uninstalled in the correct order. This incurs a significant performance cost, so use FALSE if you know $module_list is already complete and in the correct order. Defaults to TRUE.

> Returns TRUE if the operation succeeds or FALSE if it aborts due to an unsafe condition, namely, $uninstall_dependents is TRUE and a module in $module_list has dependents which are not already uninstalled and not also included in $module_list).

If I set `$uninstall_dependents` to `FALSE`, will it uninstall the module regardless of any dependencies (potentially breaking the site)?

I believe these three functions could use some better, less-ambiguous documentation...

ghost:
Fixes https://github.com/backdrop/backdrop-issues/issues/4872


core/includes/module.inc
```diff
@@ -379,15 +379,21 @@ function module_load_all_includes($type, $name = NULL) {
  * - Invoke hook_modules_installed().
  * - Invoke hook_modules_enabled().
  *
- * @param string[] $module_list
+ * @param string $module_list
  *   An array of module names.
```
klonos, line 383: Is it a string or an array?

core/includes/module.inc
```diff
@@ -379,15 +379,21 @@ function module_load_all_includes($type, $name = NULL) {
...
  * @param bool $enable_dependencies
- *   If TRUE, dependencies will automatically be added and enabled in the
- *   correct order. This incurs a significant performance cost, so use FALSE
+ *   If TRUE, dependency-checking is enabled. This means that if dependencies
+ *   are available, they will be automatically added and enabled in the correct
+ *   order. If they are unavailable, the module(s) will not be enabled.
+ *   Note that dependency-checking incurs a significant performance cost. If
+ *   FALSE, dependency-checking is bypassed and the module(s) will be enabled
+ *   regardless of whether or not their dependencies are available or listed in
+ *   the correct order. This could lead to problems, and so should only be used
```
klonos: I like this change 👍, but can we please split the "If TRUE..." and "If FALSE..." into two separate paragraphs, or bullet points? (for better readability)

core/includes/module.inc
ghost: Ha, good point!

core/includes/module.inc
ghost: I'm not sure you can have separate paragraphs within `@param` blocks, since a blank line indicates the end of the `@param` section... So I updated it to use bullets instead.

core/includes/module.inc
quicksketch: We don't strictly enforce this sort of type checking, but `string[]` means an array of strings, as is recognized by VSCode and PHPStorm.

core/includes/install.inc
```diff
@@ -992,18 +992,24 @@ function backdrop_install_system() {
...
- *   on the passed-in module list either are already uninstalled or contained in
- *   the list, and it will ensure that the modules are uninstalled in the
- *   correct order. This incurs a significant performance cost, so use FALSE if
- *   you know $module_list is already complete and in the correct order.
- *   Defaults to TRUE.
+ *   (optional)
+ *   - If TRUE, dependency-checking is enabled. This means that any modules
+ *     dependent on the passed-in module(s) are checked to see if they're
+ *     already uninstalled, or if they're included in $module_list. If the
+ *     latter, they will all be uninstalled in the correct order. If, however,
```
quicksketch: I know people that have trouble with the phrase "if the latter". And I have trouble remembering what the "latter" is sometimes when it's highly technical, so could we rephrase this from:

> If the latter, they will all be uninstalled in the correct order

Into:

> If all dependents are in the $module_list, they will all be uninstalled in the correct order

quicksketch:

PR #3499, opened 2021-01-17
Fix the links to non-existent pages in core

### LINKED ISSUE #4884: Fix the links to non-existent pages in core
It's not the bug per se, however not to confuse users we still have to address this one.

The `/core/modules/file/tests/file.test` contains links to non-existent pages as http://backdrop.org/node/1555862 and http://backdrop.org/node/1239376. There are no pages with such nids on http://backdropcms.org, however both links correspond to respective discussions on https://drupal.org/node/1555862 and https://drupal.org/node/1239376

alanmels:
Fixes https://github.com/backdrop/backdrop-issues/issues/4884


ghost:

PR #3500, opened 2021-01-18
Issue #4886: Notice: Undefined index: label in filter_admin_format_form()

### LINKED ISSUE #4886: PHP Notice: Undefined index: label in filter_admin_format_form()
**Description of the bug**

Implementation of [hook_editor_info()](https://api.backdropcms.org/api/backdrop/core%21modules%21filter%21filter.api.php/function/hook_editor_info/1) in a custom module's function produces the error:

```
Notice: Undefined index: label in filter_admin_format_form() (line 245 of /var/www/docroot/core/modules/filter/filter.admin.inc).
```
**Steps To Reproduce**

To reproduce the behavior:
1. Use code from [the hook's API page](https://api.backdropcms.org/api/backdrop/core%21modules%21filter%21filter.api.php/function/hook_editor_info/1) in a custom code. 
2. Enable the module and go to `admin/config/content/formats/filtered_html`, drop down the `Text editor` menu and note there is one empty option (next to None and CKEditor). 
3. Go to `admin/reports/dblog` to see the actual error.

**Additional information**

The API page for the hook is providing the following code:

```
function hook_editor_info() {
  $editors['myeditor'] = array(
    'title' => t('My Editor'),
    'settings callback' => '_myeditor_settings',
    'default settings' => array(
      'enable_toolbar' => TRUE,
      'toolbar_buttons' => array('bold', 'italic', 'underline', 'link', 'image'),
      'resizeable' => TRUE,
    ),
    'js settings callback' => '_myeditor_js_settings',
  );
  return $editors;
}
```
with the following explanation for the return value:

> **An associative array of editors, whose keys are internal editor names,**: which should be unique and therefore prefixed with the name of the module. Each value is an associative array describing the editor, with the following elements (all are optional except as noted):
> 
> - **title**: (required) A human readable name for the editor.
> - **settings** callback: The name of a function that returns configuration form elements for the editor. See [hook_editor_EDITOR_settings](https://api.backdropcms.org/api/backdrop/core%21modules%21filter%21filter.api.php/function/hook_editor_EDITOR_settings/1)() for details.
> - **default settings**: An associative array containing default settings for the editor, to be applied when the editor has not been configured yet.
> - **js settings callback**: The name of a function that returns configuration options that should be added to the page via JavaScript for use on the client side. See [hook_editor_EDITOR_js_settings](https://api.backdropcms.org/api/backdrop/core%21modules%21filter%21filter.api.php/function/hook_editor_EDITOR_js_settings/1)() for details.

However, in fact only [the following code](https://github.com/backdrop-contrib/bueditor/commit/08c2be1097257d6bff5a3ccad7fdbd42436a3c93) with minimal implementation worked correctly for BUEditor:

```
/**
 * Implements hook_editor_info().
 */
function bueditor_editor_info() {
  $editors['bueditor'] = array(
    'label' => t('BUEditor'),
    'library' => array('bueditor', 'backdrop.bueditor'),
  );
  return $editors;
}
```

So I believe API information for the hook needs to be re-worked.

alanmels:
Fixes https://github.com/backdrop/backdrop-issues/issues/4886

### DIFF
--- a/core/modules/filter/filter.api.php
+++ b/core/modules/filter/filter.api.php
@@ -141,7 +141,7 @@ function hook_filter_info_alter(&$info) {
  *   - default settings: An associative array containing default settings for
  *     the editor, to be applied when the editor has not been configured yet.
  *   - file: The name of a file containing the editor settings callback.
- *   - library: An associative array containing additional libraries.
+ *   - library: An associative array containing an optional library.
  *   - js settings callback: The name of a function that returns configuration
  *     options that should be added to the page via JavaScript for use on the
  *     client side. See hook_editor_EDITOR_js_settings() for details.
@@ -159,7 +159,7 @@ function hook_editor_info() {
       'resizeable' => TRUE,
     ),
     'file' => 'myeditor.admin.inc',
-    'library' => array('myeditor', 'backdrop.myeditor'),
+    'library' => array('mymodule', 'myeditor'),
     'js settings callback' => '_myeditor_js_settings',
   );
   return $editors;
--- a/core/modules/filter/filter.api.php
+++ b/core/modules/filter/filter.api.php
@@ -134,12 +134,14 @@ function hook_filter_info_alter(&$info) {
  *   which should be unique and therefore prefixed with the name of the module.
  *   Each value is an associative array describing the editor, with the
  *   following elements (all are optional except as noted):
- *   - title: (required) A human readable name for the editor.
+ *   - label: (required) A human readable name for the editor.
  *   - settings callback: The name of a function that returns configuration
  *     form elements for the editor. See hook_editor_EDITOR_settings() for
  *     details.
  *   - default settings: An associative array containing default settings for
  *     the editor, to be applied when the editor has not been configured yet.
+ *   - file: The name of a file containing the editor settings callback.
+ *   - library: An associative array containing additional libraries.
  *   - js settings callback: The name of a function that returns configuration
  *     options that should be added to the page via JavaScript for use on the
  *     client side. See hook_editor_EDITOR_js_settings() for details.
@@ -149,13 +151,15 @@ function hook_filter_info_alter(&$info) {
  */
 function hook_editor_info() {
   $editors['myeditor'] = array(
-    'title' => t('My Editor'),
+    'label' => t('My Editor'),
     'settings callback' => '_myeditor_settings',
     'default settings' => array(
       'enable_toolbar' => TRUE,
       'toolbar_buttons' => array('bold', 'italic', 'underline', 'link', 'image'),
       'resizeable' => TRUE,
     ),
+    'file' => 'myeditor.admin.inc',
+    'library' => array('myeditor', 'backdrop.myeditor'),
     'js settings callback' => '_myeditor_js_settings',
   );
   return $editors;
@@ -169,7 +173,7 @@ function hook_editor_info() {
  *   implementations.
  */
 function hook_editor_info_alter(&$editors) {
-  $editors['some_other_editor']['title'] = t('A different name');
+  $editors['some_other_editor']['label'] = t('A different name');
 }
 
 /**
--- a/core/modules/filter/filter.api.php
+++ b/core/modules/filter/filter.api.php
@@ -146,8 +146,8 @@ function hook_filter_info_alter(&$info) {
  *     options that should be added to the page via JavaScript for use on the
  *     client side. See hook_editor_EDITOR_js_settings() for details.
  *
- * @see filter_example.module
- * @see hook_filter_info_alter()
+ * @see ckeditor.module
+ * @see hook_editor_info_alter()
  */
 function hook_editor_info() {
   $editors['myeditor'] = array(
--- a/core/modules/filter/filter.module
+++ b/core/modules/filter/filter.module
@@ -1410,17 +1410,11 @@ function filter_process_format($element) {
  */
 function filter_get_attached($formats) {
   $attached = array();
-  $attached['library'][] = array('filter', 'filter');
 
-  foreach ($formats as $format_name => $format_info) {
-    if (isset($added[$format_name])) {
-      unset($formats[$format_name]);
-    }
-    else {
-      // Add the library associated with a format's editor if needed.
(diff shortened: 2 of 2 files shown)


core/modules/filter/filter.api.php
```diff
@@ -134,28 +134,32 @@ function hook_filter_info_alter(&$info) {
...
  *   Each value is an associative array describing the editor, with the
  *   following elements (all are optional except as noted):
- *   - title: (required) A human readable name for the editor.
+ *   - label: (required) A human readable name for the editor.
  *   - settings callback: The name of a function that returns configuration
  *     form elements for the editor. See hook_editor_EDITOR_settings() for
  *     details.
  *   - default settings: An associative array containing default settings for
  *     the editor, to be applied when the editor has not been configured yet.
+ *   - file: The name of a file containing the editor settings callback.
```
alanmels, line 143: See CKEditor's implementation of the hook.

core/modules/filter/filter.api.php
```diff
@@ -134,28 +134,32 @@ function hook_filter_info_alter(&$info) {
...
+ *   - library: An associative array containing additional libraries.
  *   - js settings callback: The name of a function that returns configuration
  *     options that should be added to the page via JavaScript for use on the
  *     client side. See hook_editor_EDITOR_js_settings() for details.
  *
- * @see filter_example.module
```
alanmels, line 147: The `filter_example.module` uses the `hook_filter_info()` which is totally different than `hook_editor_info()`.

core/modules/filter/filter.api.php
```diff
@@ -134,28 +134,32 @@ function hook_filter_info_alter(&$info) {
...
- * @see hook_filter_info_alter()
```
alanmels, line 148: Of course, this should be `hook_editor_info_alter()`.

core/modules/filter/filter.module
```diff
@@ -1410,17 +1410,11 @@ function filter_process_format($element) {
  */
 function filter_get_attached($formats) {
   $attached = array();
-  $attached['library'][] = array('filter', 'filter');
```
alanmels, line 1413: Not necessary.

core/modules/filter/filter.module
```diff
@@ -1410,17 +1410,11 @@ function filter_process_format($element) {
...
 
-  foreach ($formats as $format_name => $format_info) {
```
alanmels, line 1415: `$format_info` is not used anywhere in the function, so no need.

core/modules/filter/filter.module
```diff
@@ -1410,17 +1410,11 @@ function filter_process_format($element) {
...
-    if (isset($added[$format_name])) {
```
alanmels, line 1416: The `$added` variable is not set anywhere in the function, so this `if condition` is useless.

core/modules/filter/filter.module
indigoxela, line 1413: This library is the *filter* library, removing it is probably not intended, please restore that single line.

Regarding the removal of that non-existent "$added" variable: :+1:

core/modules/filter/filter.api.php
```diff
@@ -134,28 +134,32 @@ function hook_filter_info_alter(&$info) {
...
-    'title' => t('My Editor'),
+    'label' => t('My Editor'),
     'settings callback' => '_myeditor_settings',
     'default settings' => array(
       'enable_toolbar' => TRUE,
       'toolbar_buttons' => array('bold', 'italic', 'underline', 'link', 'image'),
       'resizeable' => TRUE,
     ),
+    'file' => 'myeditor.admin.inc',
+    'library' => array('myeditor', 'backdrop.myeditor'),
```
indigoxela: That might be a little less misleading, if...

`'library' => array('mymodule', 'myeditor'),`

core/modules/filter/filter.module
alanmels: Are you sure the line was intended as I've run all kind of tests without this line hitting no issues, no errors?

core/modules/filter/filter.api.php
alanmels: I can change that, though I was taking example of CKEditor module, which is in core. Please compare.

core/modules/filter/filter.api.php
```diff
@@ -134,28 +134,32 @@ function hook_filter_info_alter(&$info) {
...
- *   - title: (required) A human readable name for the editor.
+ *   - label: (required) A human readable name for the editor.
  *   - settings callback: The name of a function that returns configuration
  *     form elements for the editor. See hook_editor_EDITOR_settings() for
  *     details.
  *   - default settings: An associative array containing default settings for
  *     the editor, to be applied when the editor has not been configured yet.
+ *   - file: The name of a file containing the editor settings callback.
+ *   - library: An associative array containing additional libraries.
  *   - js settings callback: The name of a function that returns configuration
```
indigoxela, line 145: Not sure, but I think it can only be one library, keyed by the module name.

core/modules/filter/filter.api.php
indigoxela: Yes, that's like `'library' => array('ckeditor', 'backdrop.ckeditor')`

1. Module name
2. Library name

core/modules/filter/filter.module
indigoxela: Hm, I'll give it a try in the sandbox, stay tuned. :wink:

indigoxela: It still seems to get attached, but... I remember, that we had a related issue in the past when the filter library was missing.

For my peace of mind - please add it back. :smile:

alanmels: I've reverted the changed, though that line does nothing from my perspective. The library you see gets attached elsewhere and has something to do with filter, but editor. Lot's of things got mixed between the two in this function.

core/modules/filter/filter.api.php
alanmels: Ok, I will change the wording. However, looking at CKEditor's implementation I believe this should still be an array.

quicksketch:
Thanks! Merged this into 1.x and 1.18.x.

PR #3501, opened 2021-01-19
Issue #1794: Add styling to Seven theme for code blocks.

### LINKED ISSUE #1794: [UX] Add distinct `<code>` tag display in Seven theme
I doubt this is a problem anywhere else, but in the Installer Browser PR, the project details dialogs displaying any `<code>` or `<pre>` tags, the only markup is a monospaced font; it is difficult to see any difference. Should we have some other styling? Background color? Padding? Border? I could make this specific to that dialog, but thought I should mention in case a broader solution would be best.

Now as is:
codetag

See with light gray and a padding:
codegray

ghost:
Fixes https://github.com/backdrop/backdrop-issues/issues/1794

### DIFF
--- a/core/themes/seven/css/seven.base.css
+++ b/core/themes/seven/css/seven.base.css
@@ -149,9 +149,17 @@ code {
 code,
 pre,
 kbd {
-  font-size: 1em;
+  font-size: 0.9em;
+  background: #e9e9e9;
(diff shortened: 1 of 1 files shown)


klonos (approved):
LGTM 👍

quicksketch:

PR #3505, opened 2021-01-20
Issue #3036: Change the wording for display mode resets.

### LINKED ISSUE #3036: [UX] Remove the scary confirmation form for 'Reset display mode'
1. Create a new content type (preferably on a site with other existing content types)
2. Go to the 'Manage Display's tab/page
3. Note the presence of a Teaser (which we have on other content types, but don't want on this one)
4. Use the drop-down menu next to Teaser to select 'Reset to default' (to remove it)
5. Read the text on the conformation page:

> The Teaser display mode has been customized in the following places:
>
> - Page
> - Post
> - Other content type
> - This content type
> - Etc.
>
> A reset will cause all customizations to be lost. The Default display mode will be used instead.

6. Freak out because you don't want to remove the Teaser (or your customisations to it) from your other content types.
7. Hesitantly click 'Reset' and cross your fingers

This, in my opinion, is very confusing and unnecessary. Why list other content types at all if they're not affected? Or are they affected!?

ghost:
Fixes https://github.com/backdrop/backdrop-issues/issues/3036

### DIFF
--- a/core/modules/field_ui/field_ui.module
+++ b/core/modules/field_ui/field_ui.module
@@ -191,9 +191,9 @@ function field_ui_menu() {
             'file' => 'field_ui.admin.inc',
           );
           $items["$path/display/%/reset"] = array(
-            'title' => 'Reset to default',
-            'page callback' => 'backdrop_get_form',
-            'page arguments' => array('field_ui_view_mode_reset_form', $entity_type, $bundle_arg, $view_mode_position),
+            'title' => 'Reset Display mode',
+            'page callback' => 'field_ui_view_mode_reset',
+            'page arguments' => array($entity_type, $bundle_arg, $view_mode_position),
             'access arguments' => array('administer view modes'),
             'file' => 'field_ui.admin.inc',
           );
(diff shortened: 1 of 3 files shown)


core/modules/field_ui/field_ui.module
```diff
@@ -191,9 +191,9 @@ function field_ui_menu() {
             'file' => 'field_ui.admin.inc',
           );
           $items["$path/display/%/reset"] = array(
-            'title' => 'Reset to default',
-            'page callback' => 'backdrop_get_form',
-            'page arguments' => array('field_ui_view_mode_reset_form', $entity_type, $bundle_arg, $view_mode_position),
+            'title' => 'Reset Display mode',
```
klonos: @BWPanda I realize you are going for consistency here, but I think that it's the other items that should be changed here, to not capitalize the "D" in "display mode".

ghost: @klonos I think that's probably better fixed as part of https://github.com/backdrop/backdrop-issues/issues/4891. I've added a comment there about it.

jenlampton: If we're going to resolve this as part of https://github.com/backdrop/backdrop-issues/issues/4891 then let's not change it here too.  (Especially since this will require a translation update)

ghost:
Merged into 1.x and 1.18.x.

PR #3508, opened 2021-01-20
Issue #4880: Prevent older versions in update list for pre-releases

### LINKED ISSUE #4880: Backdrop shows a security update available if current version is a preview release
<strike>Not sure if this is a bug, so I start it as a question.</strike> This is a bug, function update_calculate_project_update_status() does exact string comparison, which never matches for a pre-release version.

On sites that had 1.18.0-preview installed (for testing), 1.18.0 shows as "security update".

The "Available update" page looks weird, I'm pretty sure, that's not as intended:

every-sec-update-ever-since

I'm almost sure for the lengthy sec update list there's already a bugreport somewhere.

indigoxela:
Fixes https://github.com/backdrop/backdrop-issues/issues/4880

Do not only rely on exact version string comparison.

### DIFF
--- a/core/modules/update/update.compare.inc
+++ b/core/modules/update/update.compare.inc
@@ -685,6 +685,14 @@ function update_calculate_project_update_status(&$project_data, $available) {
       break;
     }
 
+    // Pre-releases and the like need version_compare() to prevent older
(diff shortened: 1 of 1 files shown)


klonos (approved):
LGTM 👍

PR #3511, opened 2021-01-23
Issue #4901: Fix field UI titles for user and comment display forms

### LINKED ISSUE #4901: Field UI: (Almost) empty page titles on display settings pages for user and comment
**Description of the bug**

It starts in [field_ui_menu()](https://github.com/backdrop/backdrop/blob/0362219161a9b42275fdfbe54ae36d857c2af672/core/modules/field_ui/field_ui.module#L59), where comment and user entities have no bundle argument, so only the string (bundle name) can get handed over to function field_ui_display_form().

Function [field_ui_display_form()](https://github.com/backdrop/backdrop/blob/1.x/core/modules/field_ui/field_ui.admin.inc#L933) relies on the $bundle variable to be an object, to set the page title to "bundlename: displaymode".
So that won't work for comment or user, "bundlename" stays empty.

Which results in page titles like that:

empty-field-ui-title

It's a minor UI bug.

**Steps To Reproduce**

To reproduce the behavior:
1. Go to admin/config/people/manage/display/default
2. Or go to admin/structure/types/manage/post/comment/display/default
3. Inspect the page heading and meta title

indigoxela:
Fixes https://github.com/backdrop/backdrop-issues/issues/4901

### DIFF
--- a/core/modules/field_ui/field_ui.admin.inc
+++ b/core/modules/field_ui/field_ui.admin.inc
@@ -938,6 +938,13 @@ function field_ui_display_form($form, &$form_state, $entity_type, $bundle, $view
   elseif (is_object($bundle) && property_exists($bundle, 'label')) {
     $bundle_name = check_plain($bundle->label);
   }
+  elseif (is_string($bundle)) {
+    $entity_info = entity_get_info($entity_type);
+    if (isset($entity_info['label'])) {
+      $bundle_name = $entity_info['label'];
+    }
+  }
+
   $page_title = field_ui_manage_display_title($entity_type, $view_mode);
   backdrop_set_title(t('!page: !teaser display', array(
     '!page' => $bundle_name,
--- a/core/modules/field_ui/field_ui.admin.inc
+++ b/core/modules/field_ui/field_ui.admin.inc
@@ -933,10 +933,10 @@ function field_ui_display_overview($entity_type, $bundle) {
 function field_ui_display_form($form, &$form_state, $entity_type, $bundle, $view_mode) {
   $bundle_name = '';
   if (is_object($bundle) && property_exists($bundle, 'name')) {
-    $bundle_name = check_plain($bundle->name);
+    $bundle_name = $bundle->name;
   }
   elseif (is_object($bundle) && property_exists($bundle, 'label')) {
-    $bundle_name = check_plain($bundle->label);
+    $bundle_name = $bundle->label;
   }
   elseif (is_string($bundle)) {
     $entity_info = entity_get_info($entity_type);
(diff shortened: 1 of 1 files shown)


klonos (approved):
LGTM 👍

core/modules/field_ui/field_ui.admin.inc
```diff
@@ -938,6 +938,13 @@ function field_ui_display_form($form, &$form_state, $entity_type, $bundle, $view
   elseif (is_object($bundle) && property_exists($bundle, 'label')) {
     $bundle_name = check_plain($bundle->label);
   }
+  elseif (is_string($bundle)) {
+    $entity_info = entity_get_info($entity_type);
+    if (isset($entity_info['label'])) {
+      $bundle_name = $entity_info['label'];
```
quicksketch, line 944: The bundle name above is sanitized, but this one is not. I believe we should run this through check_plain():
```suggestion
      $bundle_name = check_plain($entity_info['label']);
```

quicksketch: However, down below the only place this is used is to set the title in `backdrop_set_title()`, which takes the 2nd argument to filter the string. We could either:

A) Remove check_plain() from the label in both places.
B) Set the 2nd argument in backdrop_set_title() to `PASS_THROUGH` to prevent double-escaping.

quicksketch (commented):
There's a little mixup in the filtering here that may result in double-encoding in some cases.

core/modules/field_ui/field_ui.admin.inc
indigoxela: Option B (PASS_THROUGH) seems problematic to me, because field_ui_manage_display_title() returns the raw `$view_mode['label']` - if the view mode isn't default.

I'd prefer to remove the redundant check_plain() from the $bundle->name and $bundle->label and let backdrop_set_title() do the sanitize task.

klonos: I would expect that the ideal thing would be that anything that comes out of `$bundle->label` is already passed through `check_plain()` ...but my knowledge of how the internals of the entity system works is very limited.

laryn:
I closed and reopened to trigger the Github functional tests. There was one fail that seems to be an [unrelated zombie](https://github.com/backdrop/backdrop-issues/issues/4180):

> Fail      Other      date_themes.test   150 DateThemeTestCase->testDateDisplayC
> Start and end date on same day with timezone, with remaining days is  rendered correctly.

Thanks @indigoxela for filing and fixing, and to @quicksketch, @klonos, and @docwilmot for code review and testing! I've merged into 1.x and 1.19.x.

PR #3512, opened 2021-01-24
Issue #206: Add new function for loading files of any type.

### LINKED ISSUE #206: [DX][BC] Replace module_load_include() with new, more generic function
The argument order in module_load_incude is frustratingly stupid. 

```
module_load_include('inc', 'node', 'node.admin');
```

It requires devs (even those of us who use this thing all the time) to continue to look up the order because it's so non-intuitive.  Ideally, this function would not be specific to modules, and would allow you to specify the entire file name as a single argument.  something like...

```
backdrop_load_include('module', 'node.admin.inc');
```

or 
```
backdrop_include_file('theme', 'basis', 'template.php');
```

We could even go one step further and always limit the search for .tpl.php files to the modulename/templates or themename/templates directory, .css files to modulename/css or themename/css and .js files to modulename/js or themename/js.  This pattern can be extended to anything else we decide to neatly place into subdirectories (.test files in modulename/tests, etc).

This issue will depend on the addition of the BC layer, and the dupal.inc file.

ghost:
Fixes https://github.com/backdrop/backdrop-issues/issues/206


core/includes/common.inc
```diff
@@ -8930,3 +8930,58 @@ function backdrop_get_filetransfer_info() {
...
+ */
+function backdrop_load_file($project_type, $project_name, $filename) {
+  static $files = array();
+
+  if (isset($files[$project_name][$filename])) {
+    return $files[$project_name][$filename];
+  }
+
+  if (function_exists('backdrop_get_path')) {
+    $file = BACKDROP_ROOT . '/' . backdrop_get_path($project_type, $project_name) . "/$filename";
```
klonos: Something about php micro-performance and parsing strings containing variables. This works better I've been told:

```suggestion
    $file = BACKDROP_ROOT . '/' . backdrop_get_path($project_type, $project_name) . '/' . $filename;
```

core/includes/module.inc
```diff
@@ -321,31 +321,19 @@ function module_load_install($module) {
  *
  * @return
  *   The name of the included file, if successful; FALSE otherwise.
+ *
+ * @deprecated since 1.22.0
```
klonos, line 325: Nice 👍🏼  ...we also need a matching `@since` in the docblock of `backdrop_load_file()`.

klonos (approved):
LGTM 👍🏼

core/includes/module.inc
ghost: Thanks, didn't know about that: https://github.com/backdrop-ops/docs.backdropcms.org/issues/200

bugfolder:
Closing in favor of https://github.com/backdrop/backdrop/pull/4520.

PR #3518, opened 2021-01-26
Issue #788: Make the Admin Bar 'sticky' by default.

### LINKED ISSUE #788: [UX] Keep admin_bar sticky at the top of the page by default.
I think that this is the most common setting but could be wrong. It has no effect in short pages, but reduces the scrolling to get to other admin sections in long pages so UX+. Until we get metrics in place (#285), I'd like to hear what others say:

screen shot 2018-12-11 at 1 30 59 am

Might not be ideal for mobile, so we can have a conditional. Not sure if we want to expose that as another setting (as a "only on wide screens" sub-checkbox).

ghost:
Fixes https://github.com/backdrop/backdrop-issues/issues/788

### DIFF
--- a/core/modules/admin_bar/config/admin_bar.settings.json
+++ b/core/modules/admin_bar/config/admin_bar.settings.json
@@ -1,7 +1,7 @@
 {
     "_config_name": "admin_bar.settings",
     "margin_top": 1,
-    "position_fixed": 0,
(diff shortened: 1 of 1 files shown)


klonos (approved):
LGTM 👍

ghost:
Merged into 1.x.

PR #3522, opened 2021-02-04
Issue #4936: Replace multiple occurences of `config_get()` in the same core functions with `$config->get`

### LINKED ISSUE #4936: Replace multiple occurences of `config_get()` in the same core functions with `$config->get` implementation
**Description of the bug**

As I mentioned in https://github.com/backdrop/backdrop-issues/issues/4664:

> I haven't analyzed other files, so there could be some other cases of mixed use of two methods in the same core functions. The above case is just what I've noticed in the system.admin.inc while working on something else, so I've decided to create a standalone issue for this particular file (system.admin.inc) only.

Checking other core files revealed plenty use of `config_get()` within the same functions, that is not ideal, because per https://docs.backdropcms.org/api/backdrop/core%21includes%21config.inc/function/config_get/1 the function:

```
function config_get($config_file, $option = NULL) {
  $config = config($config_file);
  return $config->get($option);
}
```

already reads the configuration file with `config()`, so there is no need to do it again.

**Additional information**

I've left all the instances of `config_get()` in mutually exclusive condition checks like:

```
function date_default_timezone($check_user = TRUE) {
  global $user;
  if ($check_user && config_get('system.date','user_configurable_timezones') && !empty($user->timezone)) {
    return $user->timezone;
  }
  else {
    $default = config_get('system.date','default_timezone');
    return empty($default) ? 'UTC' : $default;
  }
}
```

because thanks to `if else` it's always one configuration read. 

At the same time, wherever there is potentially two ore more reads, I've replaced with the recommended on https://docs.backdropcms.org/change-records/variableget-variableset-and-variabledel-deprecated-cmi method:

> Every time you call config_get() or config_set(), it hits the disk to read/write the file. If you're doing a lot of reads or writes, it's better to do all the reads at once and all the writes at once like this:
> 
> ```
> $config = config('mymodule.settings');
> $value = $config->get('foo');
> ```

alanmels:
Fixes https://github.com/backdrop/backdrop-issues/issues/4936

### DIFF
--- a/core/includes/common.inc
+++ b/core/includes/common.inc
@@ -3003,7 +3003,6 @@ function backdrop_deliver_page($page_callback_result, $default_delivery_callback
  * @see backdrop_deliver_page()
  */
 function backdrop_deliver_html_page($page_callback_result) {
-  $site_config = config('system.core');
   // Emit the correct charset HTTP header, but not if the page callback
   // result is NULL, since that likely indicates that it printed something
   // in which case, no further headers may be sent, and not if code running
@@ -3020,7 +3019,7 @@ function backdrop_deliver_html_page($page_callback_result) {
   // domain, but provide a variable to override this. If the code running for
   // this page request already set the X-Frame-Options header earlier, don't
   // overwrite it here.
-  $frame_options = $site_config->get('x_frame_options');
+  $frame_options = config_get('system.core','x_frame_options');
   if ($frame_options && is_null(backdrop_get_http_header('X-Frame-Options'))) {
     backdrop_add_http_header('X-Frame-Options', $frame_options);
   }
--- a/core/includes/common.inc
+++ b/core/includes/common.inc
@@ -3003,6 +3003,8 @@ function backdrop_deliver_page($page_callback_result, $default_delivery_callback
  * @see backdrop_deliver_page()
  */
 function backdrop_deliver_html_page($page_callback_result) {
+  $site_config = config('system.core');
+  
   // Emit the correct charset HTTP header, but not if the page callback
   // result is NULL, since that likely indicates that it printed something
   // in which case, no further headers may be sent, and not if code running
--- a/core/includes/common.inc
+++ b/core/includes/common.inc
@@ -3003,6 +3003,7 @@ function backdrop_deliver_page($page_callback_result, $default_delivery_callback
  * @see backdrop_deliver_page()
  */
 function backdrop_deliver_html_page($page_callback_result) {
+  $site_config = config('system.core');
   // Emit the correct charset HTTP header, but not if the page callback
   // result is NULL, since that likely indicates that it printed something
   // in which case, no further headers may be sent, and not if code running
@@ -3019,14 +3020,13 @@ function backdrop_deliver_html_page($page_callback_result) {
   // domain, but provide a variable to override this. If the code running for
   // this page request already set the X-Frame-Options header earlier, don't
   // overwrite it here.
-  $frame_options = config_get('system.core', 'x_frame_options');
+  $frame_options = $site_config->get('x_frame_options');
   if ($frame_options && is_null(backdrop_get_http_header('X-Frame-Options'))) {
     backdrop_add_http_header('X-Frame-Options', $frame_options);
   }
 
   // Menu status constants are integers; page content is a string or array.
   if (is_int($page_callback_result)) {
-    $site_config = config('system.core');
     // @todo: Break these up into separate functions?
     switch ($page_callback_result) {
       case MENU_NOT_FOUND:
--- a/core/includes/common.inc
+++ b/core/includes/common.inc
@@ -3021,7 +3021,7 @@ function backdrop_deliver_html_page($page_callback_result) {
   // domain, but provide a variable to override this. If the code running for
   // this page request already set the X-Frame-Options header earlier, don't
   // overwrite it here.
-  $frame_options = config_get('system.core','x_frame_options');
+  $frame_options = $site_config->get('x_frame_options');
   if ($frame_options && is_null(backdrop_get_http_header('X-Frame-Options'))) {
     backdrop_add_http_header('X-Frame-Options', $frame_options);
   }
--- a/core/includes/menu.inc
+++ b/core/includes/menu.inc
@@ -511,6 +511,8 @@ function menu_get_item($path = NULL, $router_item = NULL) {
  *   it will be printed directly to the page.
  */
 function menu_execute_active_handler($path = NULL, $deliver = TRUE, $route_handler = NULL) {
+  $site_config = config('system.core');
+  
   // Check if site is offline.
   $site_status = _menu_site_status($path, TRUE);
 
@@ -521,7 +523,7 @@ function menu_execute_active_handler($path = NULL, $deliver = TRUE, $route_handl
         // In most cases, Layout module will provide the route handler if
         // it is enabled. See layout_route_handler().
         if (empty($route_handler)) {
-          $route_handler = config_get('system.core', 'menu_route_handler');
+          $route_handler = $site_config->get('menu_route_handler');
         }
         if (empty($route_handler) || !function_exists($route_handler)) {
           $route_handler = 'menu_default_route_handler';
@@ -539,7 +541,6 @@ function menu_execute_active_handler($path = NULL, $deliver = TRUE, $route_handl
   // Offline pages are not cached in the database but have a Cache-Control
   // header to allow reverse proxies to serve offline hits instead of Backdrop.
   elseif ($site_status == MENU_SITE_OFFLINE) {
-    $site_config = config('system.core');
     $page_cache_enabled = (bool) $site_config->get('page_cache_maximum_age');
     $maintenance_max_age = $site_config->get('maintenance_page_maximum_age');
     if ($page_cache_enabled && $maintenance_max_age) {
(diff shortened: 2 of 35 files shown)


core/includes/common.inc
```diff
@@ -3019,14 +3021,13 @@ function backdrop_deliver_html_page($page_callback_result) {
   // domain, but provide a variable to override this. If the code running for
   // this page request already set the X-Frame-Options header earlier, don't
   // overwrite it here.
-  $frame_options = config_get('system.core', 'x_frame_options');
+  $frame_options = config_get('system.core','x_frame_options');
```
klonos: This change here should be reverted:

```suggestion
  $frame_options = config_get('system.core', 'x_frame_options');
```

klonos: ...also, I believe that the intention was to use:

```suggestion
  $frame_options = $site_config->get('x_frame_options');
```

core/modules/path/path.admin.inc
```diff
@@ -123,14 +123,15 @@ function path_admin_edit($path = array()) {
...
+  $site_config = config('system.core');
   $form['source'] = array(
     '#type' => 'textfield',
     '#title' => t('Existing system path'),
     '#default_value' => $path['source'],
     '#maxlength' => 255,
     '#size' => 45,
     '#description' => t('Specify the existing path you wish to alias. For example: <code>node/28</code> or <code>taxonomy/term/1</code>.'),
-    '#field_prefix' => url(NULL, array('absolute' => TRUE)) . (config_get('system.core', 'clean_url') ? '' : '?q='),
+    '#field_prefix' => url(NULL, array('absolute' => TRUE)) . ($site_config->get('clean_url') ? '' : '?q='),
```
klonos: Here, we could do something like:

```php
  $field_prefix = url(NULL, array('absolute' => TRUE)) . ($site_config->get('clean_url') ? '' : '?q=');
```

...and then reuse:

```php
  '#field_prefix' => $field_prefix,
```

It would save us 2 calls to `url()` 😉 ...but perhaps scope creep for this PR.

alanmels: Nice catch, Greg. How about if we leave this to a standalone PR, because if the review will be elevated to someone else, I don't want to confuse this PR with any code change other then just replacing `config_get`s with `site_config->get`s.

alanmels: But the other excellent catch of yours I have already committed. Thank you for noticing it!

core/includes/menu.inc
```diff
@@ -539,8 +539,9 @@ function menu_execute_active_handler($path = NULL, $deliver = TRUE, $route_handl
   // Offline pages are not cached in the database but have a Cache-Control
   // header to allow reverse proxies to serve offline hits instead of Backdrop.
   elseif ($site_status == MENU_SITE_OFFLINE) {
-    $page_cache_enabled = (bool) config_get('system.core', 'page_cache_maximum_age');
-    $maintenance_max_age = config_get('system.core', 'maintenance_page_maximum_age');
+    $site_config = config('system.core');
```
indigoxela: You could even move that up to the beginning of function menu_execute_active_handler(). - In line 524 there's already a call to `config_get('system.core', 'menu_route_handler');`

core/includes/theme.maintenance.inc
```diff
@@ -15,6 +15,7 @@
  */
 function _backdrop_maintenance_theme() {
   global $theme, $theme_key;
+  $site_config = config('system.core');
```
indigoxela: In this function it might make sense to move that config() down into the try{} - as that's the only place where this config is used.

There are two situations, where this config isn't needed at all, so it could be better to call it only if it's actually used.

core/modules/installer/installer.manager.inc
```diff
@@ -59,6 +59,7 @@
  * @ingroup forms
  */
 function installer_manager_update_form($form, $form_state, $context) {
+  $installer_settings_config = config('installer.settings');
```
indigoxela: Actually, in this loooong function :smile:  it's not only the same config, but also the same config setting: `config_get('installer.settings', 'core_update')`

It this case it would be possible to set a variable and use that twice, for instance:
```
$core_update_enabled = config_get('installer.settings', 'core_update');
```
It's up to you, whatever you think is best.

core/modules/path/path.admin.inc
```diff
@@ -123,14 +123,15 @@ function path_admin_edit($path = array()) {
  * @see path_admin_form_delete_submit()
  */
 function path_admin_form($form, &$form_state, $path = array('source' => '', 'alias' => '', 'langcode' => LANGUAGE_NONE, 'pid' => NULL)) {
+  $site_config = config('system.core');
```
indigoxela: Both occurrences of config_get() actually use the same config item, so it would be possible to:
```
$clean_urls = config_get('system.core', 'clean_url');
```
And use this variable in both places. There's no other config item used in this function.

core/modules/system/system.install
```diff
@@ -9,6 +9,7 @@
  */
 function system_requirements($phase) {
   global $base_url;
+  $site_config = config('system.core');
```
indigoxela, line 12: It might make sense to move this down to the phase and condition where it's actually needed - `if ($phase == 'runtime' && backdrop_is_apache()) `

Up here it's called, although it might not be necessary anyway, and is more than 200 lines away from its actual usage (line 254, 256).

indigoxela: There are several more places where it's in use, but again so far away, that it could be more readable, if it got set down there in these multiple `if ($phase == 'runtime') {`

My concern here is readability. In line 606 `$site_config` comes "out of nothing"... Per se your change is correct, though, so I'll leave it to you to decide.

core/modules/translation/tests/translation.test
```diff
@@ -417,6 +417,7 @@ class TranslationTestCase extends BackdropWebTestCase {
    *   TRUE if the language switch links are found, FALSE if not.
    */
   function assertLanguageSwitchLinks($node, $translation, $find = TRUE, $types = NULL) {
+    $site_config = config('system.core');
```
indigoxela: Again... :wink:  ... or you set a variable like:
```
$language_default = config_get('system.core', 'language_default');
```
As this language_default is the only config item needed here.

core/modules/user/user.module
```diff
@@ -3186,9 +3189,10 @@ function user_register_validate($form, &$form_state) {
  * @see user_register_form()
  */
 function user_register_submit($form, &$form_state) {
+  $site_config = config('system.core');
```
indigoxela: Hm... it seems, I repeat myself... :laughing: 

As 'user_email_verification' is the only thing needed in this function, you could:
```
$user_email_verification = config_get('system.core', 'user_email_verification');
```
And use that variable later on.

core/modules/user/user.tokens.inc
```diff
@@ -96,6 +96,7 @@ function user_token_info() {
  * Implements hook_tokens().
  */
 function user_tokens($type, $tokens, array $data = array(), array $options = array()) {
+  $site_config = config('system.core');
```
indigoxela: Oh, I really repeat myself often... I hope, it doesn't get boring.
```
$user_pictures = config_get('system.core', 'user_pictures');
```
Only "user_pictures" is needed in this function.

core/modules/views/includes/view.inc
```diff
@@ -1224,6 +1224,7 @@ class view {
    */
   function render($display_id = NULL) {
     $this->execute($display_id);
+    $views_settings_config = config('views.settings');
```
indigoxela: ... ehm ... guess what:
```
$show_additional_queries = config_get('views.settings', 'show_additional_queries');
```
Only "show_additional_queries" is needed in this function.

core/includes/menu.inc
alanmels: Nice catch! Thanks for the suggestion. Committed.

core/includes/theme.maintenance.inc
alanmels: As always you are right, @indigoxela! Learning from you!

core/modules/installer/installer.manager.inc
alanmels: Followed your recommendation and changed the file.

core/modules/system/system.install
alanmels: I had already started to make the recommended changes, when I noticed there are multiple instances of `if ($phase == 'runtime') {` condition. Which means if we move config read inside of `if` condition, then we would have multiple reads. So let's leave the config read in the very beginning of the function and re-use $config object anywhere we need.

core/modules/translation/tests/translation.test
alanmels: Thanks for noting this one too. Changed.

core/modules/user/user.module
alanmels: Done!

core/modules/user/user.tokens.inc
alanmels: Not boring at all, it's enlightening, @indigoxela! Thanks a lot!

core/modules/views/includes/view.inc
alanmels: Absolutely right! Corrected.

alanmels (commented):
I think I've done all the recommended changes. Let me know if you think something else should be changed too.

core/modules/path/path.admin.inc
alanmels: Done!

alanmels (commented):
I hope this is all.

core/includes/common.inc
alanmels: Thanks for the catch, @klonos. Committed.

quicksketch:

PR #3524, opened 2021-02-04
Issue #4924: Language switcher block: allow basic customization via the UI

### LINKED ISSUE #4924: [UX] Language switcher block: allow some basic customization via the block configuration form
This was brought up as a forum question: https://forum.backdropcms.org/forum/language-switcher-how-change-looks-looking-short-tutorial-tips

> My new Backdrop website will be bilingual (Dutch/English). The default language switch block is huge and doesn't look nice:
>
> image
>
> It is not so obvious to me how to tweak the appearance. I guess I have to use the configuration dialogue box for the Language switcher
> ...
> How do I get a simple NL / EN switch like this one:
>
> image
>
> I guess the answer will be that I need to add some CSS style definitions, but how?

I would like to propose adding a couple of checkboxes to the block, to allow for:

- displaying language links as bullet points, or as a delimited list (using a pipe `|`)
- rendering the full name of the languages, or the respective language codes

klonos:
Fixes https://github.com/backdrop/backdrop-issues/issues/4924

### DIFF
--- a/core/modules/locale/locale.theme.inc
+++ b/core/modules/locale/locale.theme.inc
@@ -101,3 +101,54 @@ function theme_locale_date_format_form($variables) {
   return $output;
 }
 
+/**
+ * Returns themed HTML for the language switcher block.
+ *
+ * @param $variables
+ *   An associative array containing:
+ *   - links: (object) The output of language_negotiation_get_switch_links():
+ *     - links: An associativearray of language links to be rendered.
+ *     - provider: ???
+ *   - settings: An associative array with the display settings for the language
+ *     switcher block:
+ *     - language_list_display: either "vertical" or "inline".
+ *     - language_link_label_display: either "name" or "code".
+ *     - custom_delimiter: either completely omitted (falls back to "|"), or a
+ *     single character, to be used as a delimiter when the language links are
+ *     to be rendered inline. This is added to the list items as a
+ *     data-list-delimiter attribute, and can be used in CSS via a 'content'
+ *     property. For example:
+ *     @code
+ *     .language-switcher.links.inline li:after {
+ *       content: attr(data-list-delimiter);
+ *       padding: 0 0.6em;
+ *       color: gray;
+ *     }
+ *     @endcode
+ *
+ * @ingroup themeable
+ */
+function theme_locale_block($variables) {
+  backdrop_add_css(backdrop_get_path('module', 'locale') . '/css/locale.css');
+  $settings = $variables['settings'];
+  $links = $variables['links'];
+  $links_variables = array();
+
+  if ($settings['language_link_label_display'] == 'code') {
+    foreach ($links->links as $code => $link) {
+      $links->links[$code]['title'] = $code;
+    }
+  }
+  $links_variables['links'] = $links->links;
+
+  $classes = array('language-switcher', "language-switcher-{$links->provider}");
+  if ($settings['language_list_display'] == 'inline') {
+    $classes[] = 'links';
+    $classes[] = 'inline';
+
+    $links_variables['li_attributes']['data-list-delimiter'] = isset($settings['custom_delimiter']) ? $settings['custom_delimiter'] : '|';
+  }
+  $links_variables['attributes']['class'] = $classes;
+
+  return theme('links', $links_variables);
+}
--- a/core/includes/theme.inc
+++ b/core/includes/theme.inc
@@ -1830,30 +1830,32 @@ function theme_link($variables) {
  *     is used as its CSS class. Each link should be itself an array, with the
  *     following elements:
  *     - title: The link text.
- *     - href: The link URL. If omitted, the 'title' is shown as a plain text
- *       item in the links list.
- *     - html: (optional) Whether or not 'title' is HTML. If set, the title
- *       will not be passed through check_plain().
- *     - attributes: (optional) Attributes for the anchor, or for the <span>
- *       tag used in its place if no 'href' is supplied. If element 'class' is
+ *     - href: The link URL.
+ *       - If supplied, the entire link array is passed to l() as its $options
+ *       parameter.
+ *       - If omitted, the 'title' is shown as a plain text item in the links
+ *       list.
+ *     - html: (optional) Whether or not 'title' is HTML. If set, the title will
+ *       not be passed through check_plain().
+ *     - attributes: (optional) Attributes for the anchor, or for the <span> tag
+ *       used in its place if no 'href' is supplied. If element 'class' is
  *       included, it must be an array of one or more class names.
- *     If the 'href' element is supplied, the entire link array is passed to
- *     l() as its $options parameter.
- *   - attributes: A keyed array of attributes for the UL containing the
- *     list of links.
- *   - heading: (optional) A heading to precede the links. May be an
- *     associative array or a string. If it's an array, it can have the
- *     following elements:
- *     - text: The heading text.
- *     - level: The heading level (e.g. 'h2', 'h3').
- *     - class: (optional) An array of the CSS classes for the heading.
- *     When using a string it will be used as the text of the heading and the
- *     level will default to 'h2'. Headings should be used on navigation menus
+ *   - attributes: A keyed array of attributes for the UL containing the list of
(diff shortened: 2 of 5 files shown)


core/modules/locale/locale.theme.inc
```diff
@@ -101,3 +101,54 @@ function theme_locale_date_format_form($variables) {
...
+ *     .language-switcher.links.inline li:after {
+ *       content: attr(data-list-delimiter);
+ *       padding: 0 0.6em;
+ *       color: gray;
+ *     }
+ *     @endcode
+ *
+ * @ingroup themeable
+ */
+function theme_locale_block($variables) {
```
jenlampton, line 131: I don't think we should use the word "block" in this theme function, since it is not a block. This theme function could be called from anywhere, and it would still produce a list of links. 

Since it all this function does is prepare variables that are passed into `theme_links()`, can we move this code to a preprocess function for `theme_links()`? If we used the theme hook suggestion `links__locale` we could do something like `template_preprocess_links__locale()`

docwilmot:
I like the idea of a configurable language list.

I suspect though, TBH, this PR does a bit more than is really needed here:
- Core already provided a hook suggestion for theming the contents of the locale block, via `$block['content'] = theme('links__locale_block', $variables);` If we just created a `links--locale_block.tpl.php`, it should override the display of the locale block on new sites, and old site which already created one would override the new core one.
- I'm not understanding the need for a new `li_attributes` key in `theme_links()`. In either case, it would be unnecessary if we did as above and to change such an important function just for one block is a bit overkill if alternatives exist.
- We shouldnt need the warning `$form['messages']['#markup'] .= '<div class="messages warning">' . t('This layout contains at least one language switcher block, but...` in `layout_content_form()`. We dont warn people anywhere else when they place blocks that (yet) dont make sense. I can place a Devel switcher block even if I have one user, I can place a node list even if I have no nodes, Etc.
- In either case the same warning is also duplicated on the block configure page (also not necessary there IMO, but not as concerning).

Finally, this could easily be contrib, no?

klonos:
- > Core already provided a hook suggestion...

  Agreed. I'm not good at this, so I didn't know what would be the best way to go about it. Once we have consensus, I'd be happy to revise according to what everyone agrees to be best practice, and backwards compatible 👍 

- > I'm not understanding the need for a new `li_attributes` key in `theme_links()`...

  The key point here is that the current delimiter for inline link lists is [a hardcoded comma in CSS](https://github.com/backdrop/backdrop/blob/1.x/core/themes/basis/css/component/small-text-components.css#L179). This is added in the `:after` of each `<li>`, and the only way I have found to  dynamically change that to something that is saved in config, without having to rely on JS is by adding the delimiter to a `data-*` attribute of each `<li>`.

  Unfortunately, this is a bigger problem (that's why I mentioned in the issue that perhaps it will need to be split to a separate issue). And that problem is that there is no way to alter the `<li>`s generated by `theme_links()`. See: https://api.drupal.org/comment/50053#comment-50053

  > There is no way to set attributes to the li-tag. It is not flexible.
  > 
  > In my case i need to create programmatically a menu block, looks like the standard menu in Drupal. And the difference is in
  > 
  > `<li class="leaf">`
  > You have to dance with a tambourine to get results.

  ...

  > I have used theme_item_list in order to alter the `<li>` classes.

  I could also have used `theme_item_list()` + individual `theme_link()`'s for each item, instead of `theme_links()`. That felt like re-implementing the entire `theme_links()` though, with only minor modifications 🤔 ...which has lead me down other rabbit holes like https://www.drupal.org/project/drupal/issues/2032645 which I wasn't ready for (which I'm sure will send @jenlampton down memory lane 😄 ):

  > We have two very similar functions in Durpal `theme_links` and `theme_item_list`. Both of these functions loop through a set of items, and create a HTML list. This creates twice as much code to maintain, and two separate, nearly identical, things to keep up to date.

- > We dont warn people anywhere else when they place blocks that (yet) dont make sense. ...

  That assumes that all people using the blocks and layouts UI know what they are doing, or how things work internally. All I was trying to do was to expose in the UI something that we have already added in a code comment. I don't understand why we thought this information is important enough so to add [a note there for developers](https://github.com/backdrop/backdrop/blob/1.x/core/modules/locale/locale.module#L859), but we don't think that it is equally important to let the people that use the UI be aware of.

  Anyway, my point is that in general I'm trying to avoid situations where there may be potential or obvious UX WTFs. In this case specifically, it would would go like this: "_I've placed a block and saved the layout, ...I've also cleared caches, but that block is not shown_" ...we make 2 assumptions here: a) that people will realize what's wrong, and b) that they'll know how to fix it, and where to go. Instead of making the assumptions, I thought that I'd warn people + provide a link to the page that allows them to fix things. ...besides, I would argue that if we aren't warning people anywhere else, then we should do just that, instead of continuing a tradition that hurts UX.

- > ... the same warning is also duplicated on the block configure page ...

  Good point. I'm not married to the idea of these messages, and they may be considered scope creep for the issue. I can move them to a follow-up issue, as I wouldn't want to block the main UX feature here.

  In any case, let me explain:

  - One warning is shown when adding a new language block, within the block config dialog. This covers the situation when no other language blocks have been added yet to the layout, so the other message (from `layout_content_form()`) won't be triggered. What we'd be doing there is to be letting the user know that "hey, this won't work until you add more languages" (people that already know this can ignore it, but reminding them won't hurt).
  - The other message is shown in any layout that has at least one language block when only one language is available. The use case here is that one of the site admins may have either removed or temporarily disabled languages to the point where there is only one active left on the site. When someone tries to troubleshoot why no language switcher block is shown, the first thing they are likely to do is to confirm that the block(s) are there in the layout(s). Without the message, they'd see that the block's there, and start wondering why it's not shown. With the message in place, they don't even have to check the block configuration. ...they'd either click the link and go to the language list to enable/add languages, or they'd remove the block altogether if no longer needed. Win-win.

docwilmot:
> The key point here is that the current delimiter for inline link lists is a hardcoded comma in CSS. This is added in the :after of each , and the only way I have found to dynamically change that to something that is saved in config, without having to rely on JS is by adding the delimiter to a data-* attribute of each.

But you already have a custom theme function, you could just build the inline version with HTML rather than calling `theme('links'...`. For example the code in `theme_locale_block()` (pseudocodified):
```

  ...

  if ($settings['language_link_label_display'] == 'code') {
    foreach ($links->links as $code => $link) {
      $links->links[$code]['title'] = $code;
    }
  }
  $links_variables['links'] = $links->links;

  $classes = array('language-switcher', "language-switcher-{$links->provider}");

  if ($settings['language_list_display'] == 'inline') {
    $output = '<span class="inline-links $classes">'
    foreach ($links_variables['links'] as $inline_link) {
      $output .= $inline_link . ' ' . $variables['settings']['delimiter'];
      $output .= $this_is_the_last_link ? '' : ' ';
    }
    $output .= '</span>'
    return $output;
  }
  else {
    $links_variables['attributes']['class'] = $classes;
    return theme('links', $links_variables);
  }

**Comment by docwilmot** [comment]
> The key point here is that the current delimiter for inline link lists is a hardcoded comma in CSS. This is added in the :after of each , and the only way I have found to dynamically change that to something that is saved in config, without having to rely on JS is by adding the delimiter to a data-* attribute of each.

But you already have a custom theme function, you could just build the inline version with HTML rather than calling `theme('links'...`. For example the code in `theme_locale_block()` (pseudocodified):
```

  ...

  if ($settings['language_link_label_display'] == 'code') {
    foreach ($links->links as $code => $link) {
      $links->links[$code]['title'] = $code;
    }
  }
  $links_variables['links'] = $links->links;

  $classes = array('language-switcher', "language-switcher-{$links->provider}");

  if ($settings['language_list_display'] == 'inline') {
    $output = '<span class="inline-links $classes">'
    foreach ($links_variables['links'] as $inline_link) {
      $output .= $inline_link . ' ' . $variables['settings']['delimiter'];
      $output .= $this_is_the_last_link ? '' : ' ';
    }
    $output .= '</span>'
    return $output;
  }
  else {
    $links_variables['attributes']['class'] = $classes;
    return theme('links', $links_variables);
  }

PR #3525, opened 2021-02-04
Issue #4935: Incorrect use of `config_get` in `node.api.php` file

### LINKED ISSUE #4935: Incorrect use of `config_get` in `node.api.php` file
**Description of the bug**

[Line 862](https://github.com/backdrop/backdrop/blob/ac819913fdb7aadcec9c7579fd33920327526792/core/modules/node/node.api.php#L862) in `node.api.php` file should be changed to `$config = config('my_module.settings');`, because further use of `$config->get` in the function expects `$config` to be an object containing all $option settings in the same configuration file.

At the present, the whole functions looks like:

```php
function hook_ranking() {
  // If voting is disabled, we can avoid returning the array, no hard feelings.
  $config = config_get('my_module.settings');
  if ($config->get('vote_node_enabled')) {
    return array(
      'vote_average' => array(
        'title' => t('Average vote'),
        // Note that we use i.sid, the search index's search item id, rather than
        // n.nid.
        'join' => array(
          'type' => 'LEFT',
          'table' => 'vote_node_data',
          'alias' => 'vote_node_data',
          'on' => 'vote_node_data.nid = i.sid',
        ),
        // The highest possible score should be 1, and the lowest possible score,
        // always 0, should be 0.
        'score' => 'vote_node_data.average / CAST(%f AS DECIMAL)',
        // Pass in the highest possible voting score as a decimal argument.
        'arguments' => array($config->get('vote_score_max')),
      ),
    );
  }
}
```

I will be fixing this minor bug as part a larger PR that addresses multiple other issues soon.

alanmels:
Fixes https://github.com/backdrop/backdrop-issues/issues/4935

### DIFF
--- a/core/modules/node/node.api.php
+++ b/core/modules/node/node.api.php
@@ -859,7 +859,7 @@ function hook_node_view_alter(&$build) {
  */
 function hook_ranking() {
   // If voting is disabled, we can avoid returning the array, no hard feelings.
-  $config = config_get('my_module.settings');
+  $config = config('my_module.settings');
   if ($config->get('vote_node_enabled')) {
     return array(
       'vote_average' => array(


klonos (approved):
LGTM 👍

ghost:

PR #3530, opened 2021-02-19
Issue #4954: Remove overridden backdropGetToken() function in file.test.

### LINKED ISSUE #4954: Remove backdropGetToken() from file.test
[`file.test`](https://github.com/backdrop/backdrop/blob/1.x/core/modules/file/tests/file.test) overrides the [`backdropGetToken()`](https://docs.backdropcms.org/api/backdrop/core%21modules%21simpletest%21backdrop_web_test_case.php/function/BackdropWebTestCase%3A%3AbackdropGetToken/1) function with its own version because, in Drupal, the core function didn't allow for hash salts (see https://www.drupal.org/node/1555862). However in Backdrop our core function does allow for that, so we should remove the overridden function now.

```php
/**
 * Overrides BackdropWebTestCase::backdropGetToken() to support the hash salt.
 *
 * @todo Remove when http://backdrop.org/node/1555862 is fixed in core.
 */
protected function backdropGetToken($value = '') {
  $private_key = backdrop_get_private_key();
  return backdrop_hmac_base64($value, $this->session_id . $private_key . backdrop_get_hash_salt());
}
```

ghost:
Fixes https://github.com/backdrop/backdrop-issues/issues/4954

### DIFF
--- a/core/includes/bootstrap.inc
+++ b/core/includes/bootstrap.inc
@@ -3027,6 +3027,7 @@ function _backdrop_bootstrap_configuration() {
  */
 function _backdrop_bootstrap_page_cache() {
   global $user;
+  $site_config = config('system.core');
 
   // Allow specifying special cache handlers in settings.php, like
   // using memcached or files for storing cache information.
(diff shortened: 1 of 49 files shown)


quicksketch:
Looks like we had a merge conflict. I updated via the GitHub UI to merge with `1.x`. Let's have tests run again to be sure.

PR #3531, opened 2021-02-19
Add keyboard move for editor blocks

docwilmot:
Partially solves https://github.com/backdrop/backdrop-issues/issues/4941#issuecomment-781436615

### DIFF
--- a/core/modules/layout/js/layout.admin.js
+++ b/core/modules/layout/js/layout.admin.js
@@ -90,6 +90,34 @@ Backdrop.behaviors.layoutDisplayEditor = {
         forcePlaceholderSize: true
       });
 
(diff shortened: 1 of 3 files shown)


core/modules/layout/js/layout.admin.js
```diff
@@ -90,6 +90,34 @@ Backdrop.behaviors.layoutDisplayEditor = {
         forcePlaceholderSize: true
       });
 
+    $('.layout-editor-block').bind('keydown', function(event) {
+        if(event.which == 38) {
```
klonos, line 94: @docwilmot these "magic" numbers are corresponding to keyboard keys (tab/up/down). Right? Can we either assign them to variables with meaningful names or add inline comments to explain?

PR #3533, opened 2021-02-22
Issue #4963: Slightly optimize path_admin_form()

### LINKED ISSUE #4963: Optimize `path_admin_form()`: Avoid calling `url()` multiple times.
This came up while reviewing https://github.com/backdrop/backdrop/pull/3522

See https://github.com/backdrop/backdrop/pull/3522#discussion_r578281484

> Here, we could do something like:
>
> ```php
>   $field_prefix = url(NULL, array('absolute' => TRUE)) . ($site_config->get('clean_url') ? '' : '?q=');
> ```
> ...and then reuse:
>
> ```php
>   '#field_prefix' => $field_prefix,
> ```
> 
> It would save us 2 calls to url() 😉 ...but perhaps scope creep for this PR.

klonos:
Fixes https://github.com/backdrop/backdrop-issues/issues/4963


quicksketch:

PR #3534, opened 2021-02-22
Issue #4962: Optimization of `cache_clear_all()`

### LINKED ISSUE #4962: Optimization of `cache_clear_all()`
The [cache_clear_all()](https://docs.backdropcms.org/api/backdrop/core%21includes%21cache.inc/function/cache_clear_all/1) function looks as:

```
function cache_clear_all($cid = NULL, $bin = NULL, $wildcard = FALSE) {
  // Default value cache flush.
  if (!isset($cid) && !isset($bin)) {
    if (module_exists('layout')) {
      cache('layout_path')->flush();
    }
    cache('page')->flush();
    return;
  }

  if (!$wildcard) {
    if (is_null($cid)) {
      cache($bin)->garbageCollection();
    }
    else {
      cache($bin)->delete($cid);
    }
  }
  else {
    if ($cid === '*') {
      cache($bin)->flush();
    }
    else {
      cache($bin)->deletePrefix($cid);
    }
  }
}
```

and I wonder do we really have to use the `if (module_exists('layout')) {` condition, when in fact the Layout module is part of core and it is required, so always be there.

Otherwise, every time we use `cache_clear_all()` the following chain of calls is invoked:

Now let's see how the [module_exists()](https://docs.backdropcms.org/api/backdrop/core%21includes%21module.inc/function/module_exists/1) function looks like:

```
function module_exists($module) {
  $list = module_list();
  return isset($list[$module]);
}
```

Not terribly long. But let's see the called [module_list()](https://docs.backdropcms.org/api/backdrop/core%21includes%21module.inc/function/module_list/1) function:

```
function module_list($refresh = FALSE, $bootstrap_refresh = FALSE, $sort = FALSE, $fixed_list = NULL) {
  static $list = array(), $sorted_list;

  if (empty($list) || $refresh || $fixed_list) {
    $list = array();
    $sorted_list = NULL;
    if ($fixed_list) {
      foreach ($fixed_list as $name => $module) {
        backdrop_get_filename('module', $name, $module['filename']);
        $list[$name] = $name;
      }
    }
    else {
      if ($refresh) {
        // For the $refresh case, make sure that system_list() returns fresh
        // data.
        backdrop_static_reset('system_list');
      }
      if ($bootstrap_refresh) {
        $list = system_list('bootstrap');
      }
      else {
        // Not using backdrop_map_assoc() here as that requires common.inc.
        $list = array_keys(system_list('module_enabled'));
        $list = (!empty($list) ? array_combine($list, $list) : array());
        // Remove modules that have been moved into core. This prevents loading
        // these .module files, which may otherwise cause duplicate function
        // fatal errors.
        $list = array_diff($list, backdrop_merged_modules());
      }
    }
  }
  if ($sort) {
    if (!isset($sorted_list)) {
      $sorted_list = $list;
      ksort($sorted_list);
    }
    return $sorted_list;
  }
  return $list;
}
```

So performance-wise, I believe this part:
```
  // Default value cache flush.
  if (!isset($cid) && !isset($bin)) {
    if (module_exists('layout')) {
      cache('layout_path')->flush();
    }
    cache('page')->flush();
    return;
  }
```

could be simplified just as:

```
  // Default value cache flush.
  if (!isset($cid) && !isset($bin)) {
    cache('layout_path')->flush();
    cache('page')->flush();
    return;
  }
```

Should I offer PR for this or there are cases when the Layout module is disabled?

alanmels:
Resolves https://github.com/backdrop/backdrop-issues/issues/4962

### DIFF
--- a/core/includes/cache.inc
+++ b/core/includes/cache.inc
@@ -169,9 +169,7 @@ function cache_flush($bin) {
 function cache_clear_all($cid = NULL, $bin = NULL, $wildcard = FALSE) {
   // Default value cache flush.
   if (!isset($cid) && !isset($bin)) {
-    if (module_exists('layout')) {
-      cache('layout_path')->flush();
-    }
+    cache('layout_path')->flush();
     cache('page')->flush();
     return;
   }


klonos (approved):
LGTM 👍

quicksketch:

PR #3537, opened 2021-02-23
Issue #4966: Allow disabling to overriding of settings.php values.

### LINKED ISSUE #4966: Allow server settings to be overridden
**UPDATE:** This issue started as a feature request to add the ability to override the server variable to Backdrop. However it was later pointed out that this variable can already be overridden by [adding some lines to `settings.php`](https://github.com/backdrop/backdrop-issues/issues/4966#issuecomment-786248854). This issue now serves as a reference to anyone needing to do this same thing in future.

---

**Description of the need**

If any settings are set via the `$_SERVER['BACKDROP_SETTINGS']` variable, they take precedence over settings in the `settings.php` file(s).

This feature is used, for example, by Lando, which [sets the database connection information](https://docs.lando.dev/config/backdrop.html#connecting-to-your-database) via that variable. This makes it easier for users in that they don't need to setup the DB connection info themselves. However this also means that there's no way to specify the DB connection info manually via `settings.php`, (e.g. for debugging, testing, etc.), since the `settings.php` values are **always overridden by the server variable**.

It's this hard-coded overriding that I'd like to make optional, so that users can choose which settings will be used for any given site.

**Proposed solution**

I recommend adding a new setting to `settings.php` which determines whether the `settings.php` values are overridden by the server or not.

I've tried to [override Lando's default setup](https://github.com/lando/lando/issues/2727), but to no avail.

Backdrop now allows you to disable the automatic overriding of `settings.php` values by the server.

ghost:
Fixes https://github.com/backdrop/backdrop-issues/issues/4966

### DIFF
--- a/core/includes/bootstrap.inc
+++ b/core/includes/bootstrap.inc
@@ -860,8 +860,8 @@ function backdrop_settings_initialize() {
 
   $is_https = backdrop_is_https();
 
-  // Load settings specified by the server, if present
-  if (isset($_SERVER['BACKDROP_SETTINGS'])) {
+  // Load settings specified by the server, if present and not overridden.
+  if (isset($_SERVER['BACKDROP_SETTINGS']) && empty($settings['override_server_settings'])) {
     $server_settings = @json_decode($_SERVER['BACKDROP_SETTINGS'], TRUE);
     if (!is_array($server_settings)) {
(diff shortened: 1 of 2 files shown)


ghost:
No longer needed (see issue for details).

PR #3538, opened 2021-02-24
Issue #4950: File usage has bad link

### LINKED ISSUE #4950: File usage has bad link.
**Description of the bug**

After uploading a fie to a Backdrop site, the file shows up in admin/content/files with some links to manage it.  Clicking "Manage" gives the user three tabs, including "Manage" and "Usage".  On the "Usage" page is a link to the file itself, but the link is malformed.

**Steps To Reproduce**

To reproduce the behavior:
1. Go to admin/content/files, click "Add a file...", and add something.
2. Go back to admin/content/files and click "Manage" for the file you just uploaded.
3. Click the "Usage" tab
4. Inspect the link to the file.

**Actual behavior**

The link is to something like  "http://example.com/public://myfile.txt".  Note the "public://" in the middle of the URL.

**Expected behavior**

The link should be to "http://example.com/files/myfile.txt." 

In `file_usage_page()`, the path needs to be translated to a URL with `file_create_url()` before being passed to `l()`.

**Additional information**

- Backdrop 1.18.1

indigoxela:
Fixes https://github.com/backdrop/backdrop-issues/issues/4950

### DIFF
--- a/core/modules/file/file.pages.inc
+++ b/core/modules/file/file.pages.inc
@@ -1026,6 +1026,7 @@ function file_usage_page($file) {
           // Unlike Node or Term entities File entities have a stream wrapper
           // to the file set as their path, but the path to the entities' page
           // is "path/[fid]".
+          // @todo Fix in https://github.com/backdrop/backdrop-issues/issues/2300
           if ($entity_type == 'file') {
             $uri = 'file/' . $entity_id;
           }
--- a/core/modules/file/file.pages.inc
+++ b/core/modules/file/file.pages.inc
@@ -1023,7 +1023,15 @@ function file_usage_page($file) {
 
         // Link the label to the URI when possible.
         if (!empty($entity_uri['path'])) {
-          $entity_label = l($label, $entity_uri['path']);
+          // Unlike Node or Term entities File entities have a stream wrapper
+          // set as path, which needs to be converted to an URL.
+          if ($entity_type == 'file') {
+            $uri = file_create_url($entity_uri['path']);
+          }
+          else {
+            $uri = $entity_uri['path'];
+          }
+          $entity_label = l($label, $uri);
         }
         else {
           $entity_label = check_plain($label);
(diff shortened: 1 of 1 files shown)


core/modules/file/file.pages.inc
```diff
@@ -1023,7 +1023,16 @@ function file_usage_page($file) {
 
         // Link the label to the URI when possible.
         if (!empty($entity_uri['path'])) {
-          $entity_label = l($label, $entity_uri['path']);
+          // Unlike Node or Term entities File entities have a stream wrapper
+          // to the file set as their path, but the path to the entities' page
+          // is "path/[fid]".
+          if ($entity_type == 'file') {
+            $uri = "file/{$entity_id}";
```
klonos: Excuse my ignorance, but why the curly syntax here?

indigoxela: In fact, it's very unlikely here that php mixes up the string with the variable, so the curly braces probably aren't necessary.
But they don't do any harm, either.

Do you insist in removing them?

jlfranklin: The typical code style in Backdrop for something like this is concatenation, not interpolation

```php
$uri = 'file/' . $entity_id;
```

indigoxela: > The typical code style in Backdrop for something like this...

Not sure, but why not.

core/modules/file/file.pages.inc
```diff
@@ -1023,7 +1023,16 @@ function file_usage_page($file) {
 
         // Link the label to the URI when possible.
         if (!empty($entity_uri['path'])) {
-          $entity_label = l($label, $entity_uri['path']);
+          // Unlike Node or Term entities File entities have a stream wrapper
+          // to the file set as their path, but the path to the entities' page
+          // is "path/[fid]".
+          if ($entity_type == 'file') {
```
quicksketch:
```suggestion
          // is "path/[fid]".
          // @todo Fix in https://github.com/backdrop/backdrop-issues/issues/2300
          if ($entity_type == 'file') {
```

quicksketch:
I added a `@todo` line in the code to reference fixing this in https://github.com/backdrop/backdrop-issues/issues/2300. Merged into 1.x and 1.18.x. Thanks!

PR #3550, opened 2021-02-28
POC of keyboard drag n drop

tarekdj:
This is a POC based on @docwilmot PR #3531 
The main idea is when you set focus on a draggable element you can use tha keys "j" and "k" to move between regions.

**Notes:**
* This is not production code.
* It relies on javascript
* Tested only using my laptop keyboard

### DIFF
--- a/core/modules/layout/css/layout.admin.css
+++ b/core/modules/layout/css/layout.admin.css
@@ -392,3 +392,7 @@
 [dir="rtl"] .edit-layout-top .dropbutton-wrapper {
   float: left;
 }
+
+*:focus .layout-editor-block-title {
+  background-color: #e9eebc;
+}
--- a/core/modules/layout/js/layout.admin.js
+++ b/core/modules/layout/js/layout.admin.js
@@ -78,6 +78,7 @@ Backdrop.behaviors.layoutConfigure = {
  */
 Backdrop.behaviors.layoutDisplayEditor = {
   attach: function(context) {
+    var synth = window.speechSynthesis;
     // Apply drag and drop to regions.
     var $regions = $('.layout-editor-region-content').once('layout-sortable');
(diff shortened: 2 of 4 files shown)


docwilmot:
This is really good work @tarekdj. Since you seem to know what you're doing, unlike my fumbling around, I'll give you all my thoughts:
- Would be nice to have screenreader "announcements" when a user has tabbed to an item where actions are available, like when you tab to a breadcrumb block, say "movable block, breadcrumb", or something like that.
- I did not expect the block to immediately jump to the next region. I thought pressing the "K" key would highlight the regions in order, then pressing something else (I tried space bar) would *then* move the last selected block. But if the block was announced as movable first as in my previous point then that may be OK I guess?
- For both your solution and mine, would it be best if instead of `tab` selecting a block, maybe we add a keypress (space) that definitively adds a "selected" status to the block before it becomes movable? So tab to block, press space key if you want to move it either up down or cross-region, then press the up down K or J buttons.
- Similarly when a block is moved up or down, announce "moved up" etc
- I did not expect the audible announcements in Chrome!! Is that just for testing?
- However before testing your PR I actually turned on Googles built in screenreader. The "Moved to..." announcement does not happen if the screenreader is on. 
- My most recent thought actually was it would be simpler to use JS to add a "Move to region..." to the block contextual links. But I like your solution.
- Minor matter: I expected K to move up and L to move down, since L comes after K.

tarekdj:
@docwilmot Thanks for your feedback! I'm defenetly not expert in a11y this was a little hack during my weekend.
I found this: 
* https://dev.opera.com/articles/accessible-drag-and-drop/example.html
* https://dev.opera.com/articles/accessible-drag-and-drop/

Sounds like the right direction.

PR #3551, opened 2021-03-02
Issue #4973: Improve the documentation for user_save() wrapper

### LINKED ISSUE #4973: Improve the documentation for user_save() wrapper function
**Description of the bug**

Following the documentation on https://docs.backdropcms.org/api/backdrop/core%21modules%21user%21user.module/function/user_save/1, I have created a test user object like so:

```
$account = new stdClass();
$account->name = 'Test User';
$account->mail = 'test@example.com';
$account->is_new = TRUE;
```

However attempting to save it with `user_save($account);` produced the following error:

```
Error: Call to undefined method stdClass::save() in user_save() (line 336 of /var/www/docroot/core/modules/user/user.module). 
```

Only following the documentation on https://docs.backdropcms.org/converting-modules-from-drupal:

> **Classed entities**
> 
> Several entities in Backdrop are now classed objects which extend the Entity Class. Users, nodes, comments, files, taxonomy terms and vocabularies have been converted to extend the new base class. Therefore programmatic creation of these objects has changed in Backdrop. For example:
> 
> Old
> 
> ```
> $node = new stdClass();
> $node->type = 'blog';
> $node->title = 'New Title';
> ...
> node_save($node);
> ```
> 
> New
> 
> ```
> $node = entity_create('node', array ('type' => 'blog'));
> $node->uid = $user->uid;
> $node->title = 'New Title';
> $node->save();
> ```

in particular, using this code worked as expected:

```
$extra_edit = array();
$account = entity_create('user', array(
  'name' => 'Test3 User',
  'mail' => 'info3@example.com',
  'roles' => array(
    BACKDROP_AUTHENTICATED_ROLE => 'authenticated',
    3 => 'committer',
  ),
  'pass' => user_password(),
  'status' => 1,
) + $extra_edit);
$account->save();
```

Please someone test and confirm this and I will create PR. Thanks!

alanmels:
Resolves https://github.com/backdrop/backdrop-issues/issues/4973

### DIFF
--- a/core/modules/user/user.module
+++ b/core/modules/user/user.module
@@ -324,10 +324,9 @@ function user_load_by_name($name) {
  *
  * @param $account
  *   The user object to modify or add. If you want to modify an existing user
- *   account, you need to ensure that (a) $account is an object, and (b) you
+ *   account, you need to ensure that $account is an entity object, and you 
  *   have set $account->uid to the numeric user ID of the user account you wish
- *   to modify. If you want to create a new user account, you can set
- *   $account->is_new to TRUE or omit the $account->uid field.
+ *   to modify. To create a new account, use entity_create('user', array());
  *
  * @return int
  *   Either SAVED_NEW or SAVED_UPDATED depending on the operation performed.
--- a/core/modules/user/user.module
+++ b/core/modules/user/user.module
@@ -322,7 +322,7 @@ function user_load_by_name($name) {
 /**
  * Save a new or updated user.
  *
- * @param $account
+ * @param User $account
  *   The user entity object to add or modify. 
  * @return int
  *   Either SAVED_NEW or SAVED_UPDATED depending on the operation performed.
--- a/core/modules/user/user.module
+++ b/core/modules/user/user.module
@@ -322,12 +322,13 @@ function user_load_by_name($name) {
 /**
  * Save a new or updated user.
  *
- * @par am $account
+ * @param $account
  *   The user entity object to add or modify. 
  * @return int
  *   Either SAVED_NEW or SAVED_UPDATED depending on the operation performed.
  *
  * @see entity_create()
+ * @see user_load()
  */
 function user_save($account) {
   return $account->save();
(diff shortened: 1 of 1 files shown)


core/modules/user/user.module
```diff
@@ -322,15 +322,12 @@ function user_load_by_name($name) {
...
  * Save a new or updated user.
  *
- * @param $account
- *   The user object to modify or add. If you want to modify an existing user
- *   account, you need to ensure that (a) $account is an object, and (b) you
- *   have set $account->uid to the numeric user ID of the user account you wish
- *   to modify. If you want to create a new user account, you can set
- *   $account->is_new to TRUE or omit the $account->uid field.
- *
+ * @par am $account
```
indigoxela: Nice! Brief and clear, but I think it should have been `@param object $account`.

alanmels: Silly me leaving that empty space in `par am`. Thanks for noticing it. Fixed also per your recommendation I've added the reference to user-load() too.

core/modules/user/user.module
```diff
@@ -323,14 +323,12 @@ function user_load_by_name($name) {
  * Save a new or updated user.
  *
  * @param $account
```
indigoxela: Sorry for being so picky... The datatype should get added. Actually I had to look up the `@param` documentation, too. :wink: 

> `@param datatype $paramname`
> ... The datatype should be a valid PHP type (int, string, bool, etc), a class name for the type of object, or simply "mixed".

So in our case that should be like:

`@param User $account`

As "User" is the name of the class. I hope, I don't overstretch your patience. Your work is really appreciated. :+1:

alanmels: > I hope, I don't overstretch your patience. Your work is really appreciated. +1

@indigoxela On the contrary I do appreciate your pointers as I'm learning from you a lot! When it comes to making it right, I'm nothing but patience.

quicksketch:

PR #3553, opened 2021-03-03
Issue #4977: Suppress empty css class attribute

### LINKED ISSUE #4977: Empty class attributes added to block markup when setting custom elements but not specifying a class
I noticed this while testing #1818, and have confirmed that it happens on Vanilla Backdrop.

Steps to reproduce (I'm using specific a layout/block here, but same applies to any layout/block):

1. Edit the Home page layout
2. Edit the "Promoted content" block
3. Change the block title to something custom (to ensure that it's being rendered)
3. Expand the "Style settings" fieldset, and change:
   - the "Wrapper tag" to something like a SECTION
   - the "Heading level" to something like a H6
   - the "Content tag" to something like an ASIDE
4. Do NOT specify any classes
5. Save the block and save the layout.
6. Clear caches (just in case)
7. Inspect the generated HTML

## Actual result

The block wrapper already includes some default classes, so you get the following, which is fine 👍 :

```html
<section class="block block-views-promoted-block contextual-links-region">
  ...
</section>
```

Empty `class` attributes are getting added for the block heading and the block content elements though 👎 :

```html
<h6 class="">some title</h6>
<aside class="">
  ...
</aside>
```

## Expected result

No empty `class` attributes added if no classes have been specified, and no default classes exist:

```html
<h6>some title</h6>
<aside>
  ...
</aside>
```

indigoxela:
Fixes https://github.com/backdrop/backdrop-issues/issues/4977

Do not render empty class attribute.

### DIFF
--- a/core/modules/layout/layout.theme.inc
+++ b/core/modules/layout/layout.theme.inc
@@ -530,9 +530,10 @@ function template_preprocess_block_dynamic(&$variables) {
   }
   $title_classes = array_filter($title_classes);
   $variables['title_tag'] = $style->settings['title_tag'];
-  $variables['title_attributes'] = array(
-    'class' => $title_classes,
-  );
+  $variables['title_attributes'] = array();
+  if (!empty($title_classes)) {
+    $variables['title_attributes']['class'] = $title_classes;
+  }
 
   // Add content tag and classes.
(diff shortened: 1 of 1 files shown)


quicksketch:

PR #3556, opened 2021-03-05
Issue #4941: Use manage blocks interface with keyboard

### LINKED ISSUE #4941: [UX] Fix for Manage blocks interface: cannot be used with keyboard navigation
**Description of the bug**

We received an email today from someone who is struggling with the manage blocks interface because they can not see. (The following is from a non-English speaker)

> Hi. I can not see. I am using my computer with NVDA reading the screen. Drag
and drop is done with the mouse. I have to use my computer only with the
keyboard. When using Drupal, there were line weights in block scrolling. I
was making the editing I wanted easy with the keyboard. Backdrop caught my
attention. However, I could not do anything with the mouse in the block
(settlement) area. Will the developer see this post? If it comes a line, we
will be very happy web developers who cannot see it. Thanks in advance.

The only obstacle to moving the site is the necessity of using the mouse.
The story is that everything can be done with the keyboard as if a mouse is
not attached.

**Steps To Reproduce**

To reproduce the behavior:
1. Go to the "manage blocks" page for Layouts
2. Attempt to reposition a block without using the mouse

**Actual behavior**

Unable to position blocks.

**Expected behavior**

Able to position blocks.

---

PR by @docwilmot and @tarekdj  : https://github.com/backdrop/backdrop/pull/3561 (drag n drop blocks, with cross-region drag. JS only)
PR by @docwilmot : https://github.com/backdrop/backdrop/pull/3556 (dropbutton link with dialog with to move blocks incl NOJS) Not currently preferred option,

docwilmot:
Fixes https://github.com/backdrop/backdrop-issues/issues/4941

### DIFF
--- a/core/modules/layout/layout.admin.inc
+++ b/core/modules/layout/layout.admin.inc
@@ -1617,6 +1617,199 @@ function layout_block_configure_ajax_update($form, $form_state) {
 }
 
 /**
+ * Menu callback; Change a block's position in a region, or move to a new
+ * region.
+ */
+function layout_block_move_form($form, &$form_state, Layout $layout, Block $block, $renderer_name, $region_name = NULL) {
+  form_load_include($form_state, 'inc', 'layout', 'layout.admin');
+
+  $form_state['layout'] = $layout;
+  $form_state['block'] = $block;
+
+  if (!isset($form_state['renderer'])) {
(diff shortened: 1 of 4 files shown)


quicksketch:
Replaced by https://github.com/backdrop/backdrop/pull/3561

PR #3557, opened 2021-03-07
Issue #4985: Add an aria announce method

### LINKED ISSUE #4985: [A11Y] Provide an API for aria-live region update announcements for accessibility improvements
**Description of the need**
Sister issue on D7: https://www.drupal.org/project/drupal/issues/1915302
Triggering issue on Backdrop: https://github.com/backdrop/backdrop-issues/issues/4941
Read about aria-live here: https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/ARIA_Live_Regions

Specifically, we are looking for ways to make the Layout block add page accessible.

**Proposed solution**

An API to allow devs to trigger announcements on certain screen events and interactions
PR upcoming; ported from D7.

---
PR by @docwilmot https://github.com/backdrop/backdrop/pull/3557

docwilmot:
Fixes https://github.com/backdrop/backdrop-issues/issues/4985

Includes a rough implementation on tabledrag, for demo purposes.

To test go to the roles page (or any page with a draggable table) while using a screenreader. Tab to a row then use the arrow keys to move the row. The screen reader should announce the direction the row was moved. Not very helpful yet if you know what key you just pressed anyway, but, again, a demo.

### DIFF
--- a/core/misc/announce.js
+++ b/core/misc/announce.js
@@ -14,66 +14,51 @@
  *     ));
  *   });
  */
-(function (document, Backdrop) {
+(function (Backdrop, debounce) {
+  var liveElement;
+  var announcements = [];
+  Backdrop.behaviors.backdropAnnounce = {
+    attach: function attach(context) {
+      if (!liveElement) {
+        liveElement = document.createElement('div');
+        liveElement.id = 'backdrop-live-announce';
+        liveElement.className = 'visually-hidden';
+        liveElement.setAttribute('aria-live', 'polite');
+        liveElement.setAttribute('aria-busy', 'false');
+        document.body.appendChild(liveElement);
+      }
+    }
+  };
 
-var liveElement;
+  function announce() {
+    var text = [];
+    var priority = 'polite';
+    var announcement;
+    var il = announcements.length;
 
-/**
- * Builds a div element with the aria-live attribute and attaches it
- * to the DOM.
- */
-Backdrop.behaviors.backdropAnnounce = {
-  attach: function (settings, context) {
-    if (!liveElement) {
-      liveElement = document.createElement('div');
-      liveElement.id = 'backdrop-live-announce';
-      liveElement.className = 'element-invisible';
-      liveElement.setAttribute('aria-live', 'polite');
+    for (var i = 0; i < il; i++) {
+      announcement = announcements.pop();
+      text.unshift(announcement.text);
+
+      if (announcement.priority === 'assertive') {
+        priority = 'assertive';
+      }
+    }
+
+    if (text.length) {
+      liveElement.innerHTML = '';
+      liveElement.setAttribute('aria-busy', 'true');
+      liveElement.setAttribute('aria-live', priority);
+      liveElement.innerHTML = text.join('\n');
       liveElement.setAttribute('aria-busy', 'false');
-      document.body.appendChild(liveElement);
     }
   }
-};
 
-/**
- * Triggers audio UAs to read the supplied text.
- *
- * @param {String} text
- *   - A string to be read by the UA.
- *
- * @param {String} priority
- *   - A string to indicate the priority of the message. Can be either
- *   'polite' or 'assertive'. Polite is the default.
- *
- * Use Backdrop.announce to indicate to screen reader users that an element on
- * the page has changed state. For instance, if clicking a link loads 10 more
- * items into a list, one might announce the change like this.
- * $('#search-list')
- *   .on('itemInsert', function (event, data) {
- *     // Insert the new items.
- *     $(data.container.el).append(data.items.el);
- *     // Announce the change to the page contents.
- *     Backdrop.announce(Backdrop.t('@count items added to @container',
- *       {'@count': data.items.length, '@container': data.container.title}
- *     ));
- *   });
- *
- * @see http://www.w3.org/WAI/PF/aria-practices/#liveprops
- */
-Backdrop.announce = function (text, priority) {
-  if (typeof text === 'string') {
-    // Clear the liveElement so that repeated strings will be read.
-    liveElement.innerHTML = '';
-    // Set the busy state to true until the node changes are complete.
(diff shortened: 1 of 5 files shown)


core/misc/announce.js
```diff
@@ -0,0 +1,64 @@
...
+ */
+(function (Backdrop, debounce) {
+  var liveElement;
+  var announcements = [];
+  Backdrop.behaviors.backdropAnnounce = {
+    attach: function attach(context) {
+      if (!liveElement) {
+        liveElement = document.createElement('div');
+        liveElement.id = 'backdrop-live-announce';
+        liveElement.className = 'visually-hidden';
```
indigoxela: "visually-hidden" means that it's visible in Backdrop - which actually uses the class "element-invisible" for that. (The right class in the wrong CMS :wink: )

core/modules/system/system.module
```diff
@@ -2030,6 +2030,15 @@ function system_library_info() {
     ),
   );
 
+  // Announce.
+  $libraries['announce'] = array(
+    'title' => 'Announce',
+    'version' => '1.0.0',
+    'js' => array(
+      'core/misc/announce.js' => array(),
```
indigoxela: In my testing environment the library was loaded right after "jquery.once.js", even before "backdrop.js" - which caused trouble with "undefined" nagging.

Could `'core/misc/announce.js' => array('group' => JS_DEFAULT),` be a valid solution for that?

core/modules/system/system.module
```diff
@@ -2030,6 +2030,15 @@ function system_library_info() {
     ),
   );
 
+  // Announce.
+  $libraries['announce'] = array(
```
quicksketch: Because this is a Backdrop-custom library, I think we should prefix this as
```suggestion
  $libraries['backdrop.announce'] = array(
```
That would line up with the other Backdrop-provided JS libraries. We should also move this array up before all the jQuery libraries.

core/misc/announce.js
docwilmot: Thanks

core/modules/system/system.module
docwilmot: It was a good solution, thanks.

core/modules/system/system.module
docwilmot: Done

docwilmot:
@quicksketch I've renamed the library, but wondering if this should go into Backdrop.is to avoid needing to declare a library, since it's likely to be used everywhere?

quicksketch:
I think it's reasonable to separate into its own file as you have done. Modules that call `Backdrop.announce()` should specify `backdrop.announce` as a dependency in their `hook_library_info()`.

core/misc/announce.js
```diff
@@ -0,0 +1,64 @@
+/**
+ * Adds an HTML element and method to trigger audio UAs to read system messages.
+ *
+ * Use Backdrop.announce() to indicate to screen reader users that an element on
```
indigoxela, line 4: It might be helpful to also mention here, that the library needs to get included with something like `backdrop_add_library('system', 'backdrop.announce')` or `$form['#attached']['library'][] = array('system', 'backdrop.announce');`

Not sure, how detailed this should be.

core/misc/backdrop.js
```diff
@@ -827,4 +827,34 @@ Backdrop.optimizedResize = (function() {
   }
 }());
 
+Backdrop.debounce = function (func, wait, immediate) {
```
quicksketch, line 854:
```suggestion
/**
 * Limits the invocations of a function in a given time frame.
 *
 * This can be useful to respond to an event that fires very frequently,
 * such as a "keyup" event while a user is typing in a field.
 *
 * A common use of debouncing in other systems is window resizing,
 * however Backdrop provides the Backdrop.optimizedResize() method,
 * which should be used for that purpose.
 *
 * @param function func
 *   The function to be invoked.
 * @param number wait
 *   The time period within which the callback function should only be
 *   invoked once. For example if the wait period is 250ms, then the callback
 *   will only be called at most 4 times per second.
 * @param bool immediate
 *   Whether we wait at the beginning or end to execute the function.
 *
 * @return function
 *   The debounced function.
 *
 * @since 1.18.2 Method added.
 */
Backdrop.debounce = function (func, wait, immediate) {
```

core/misc/announce.js
```diff
@@ -0,0 +1,64 @@
...
+    if (text.length) {
+      liveElement.innerHTML = '';
+      liveElement.setAttribute('aria-busy', 'true');
+      liveElement.setAttribute('aria-live', priority);
+      liveElement.innerHTML = text.join('\n');
+      liveElement.setAttribute('aria-busy', 'false');
+    }
+  }
+
+  Backdrop.announce = function (text, priority) {
```
quicksketch:
```suggestion
  /**
   * Triggers audio UAs to read the supplied text.
   *
   * The aria-live region will only read the text that currently populates its
   * text node. Replacing text quickly in rapid calls to announce results in
   * only the text from the most recent call to Backdrop.announce() being
   * read. By wrapping the call to announce in a debounce function, we allow for
   * time for multiple calls to Backdrop.announce() to queue up their
   * messages. These messages are then joined and append to the aria-live region
   * as one text node.
   *
   * @param string text
   *   A string to be read by the UA.
   * @param string priority
   *   A string to indicate the priority of the message. Can be either
   *   'polite' or 'assertive'.
   *
   * @return function
   *   The return of the call to debounce.
   *
   * @see http://www.w3.org/WAI/PF/aria-practices/#liveprops
   * @since 1.18.2 Method added.
   */
    Backdrop.announce = function (text, priority) {
```

quicksketch:
I added docblocks to both `Backdrop.debounce()` and `Backdrop.announce()`, and added `@since 1.18.2` indicators to clarify when this API became available.

We're not yet using this functionality but it is now in place for further enhancements.

I have merged this into 1.x and 1.18.x. Thanks!

PR #3558, opened 2021-03-07
Issue #4987: Add aria-label attributes to permissions page for screen readers.

### LINKED ISSUE #4987: [A11Y] Permissions page checkboxes are not accessible
**Description of the bug**

This problem was raised by @MustafaDogan78 in issue #4985.

The checkboxes on the permissions page (`admin/config/people/permissions`) are not accessible via a screen reader. Instead of each role and permission being read by the screen reader, @MustafaDogan78 reports that the checkboxes must be counted to associate which checkbox matches each role.

The original report (as run through Google Translate):

> I don't know English. I'm building a dialect with translation. It's a Problem, but there's no other way. There is another problem with accessibility. Setting permissions. Problem reading roles first, then checkboxes. When using Drupal, the cue box and the roles would be read simultaneously. I figure out what role to allow by counting the tick boxes.

**Steps To Reproduce**

To reproduce the behavior:
1. Enable a screen reader such as NVDA on Windows or VoiceOver on Mac.
2. Go to 'admin/config/people/permissions'
3. Use the tab key to focus on individual checkboxes that grant a permission
4. Listen to the voice over.

**Actual behavior**

The voice over says the word "Checkbox" but does not say what the checkbox does.

**Expected behavior**

The voice over should announce the name of the role for each checkbox.

**Additional information**

@MustafaDogan78 reports that Drupal 7 has a more desireable behavior that we may compare against. However, I think we intentionally reworked this page to avoid "element-invisible" labels because it made using the browser search difficult, because the permission name was repeated in the hidden text for every role, resulting in a confusing and tedious search functionality.

With improved accessibility options since Drupal 7, we should investigate if using an `aria-` attribute can provide an alternative that is both accessible to screen readers but does not cause a usability problem.

quicksketch:
Fixes https://github.com/backdrop/backdrop-issues/issues/4987

### DIFF
--- a/core/modules/user/user.admin.inc
+++ b/core/modules/user/user.admin.inc
@@ -427,36 +427,38 @@ function user_admin_permissions($form, $form_state, $role_name = NULL) {
   }
   asort($modules);
 
+  $all_permissions = array();
   foreach ($modules as $module => $display_name) {
-    if ($permissions = module_invoke($module, 'permission')) {
+    if ($module_permissions = module_invoke($module, 'permission')) {
+      $all_permissions += $module_permissions;
       $form['permission'][] = array(
         '#markup' => $module_info[$module]['name'],
         '#id' => $module,
       );
-      foreach ($permissions as $perm => $perm_item) {
+      foreach ($module_permissions as $permission_key => $permission_item) {
         // Fill in default values for the permission.
-        $perm_item += array(
+        $permission_item += array(
           'description' => '',
           'restrict access' => FALSE,
-          'warning' => !empty($perm_item['restrict access']) ? t('Warning: Give to trusted roles only; this permission has security implications.') : '',
+          'warning' => !empty($permission_item['restrict access']) ? t('Warning: Give to trusted roles only; this permission has security implications.') : '',
         );
-        $options[$perm] = '';
-        $form['permission'][$perm] = array(
+        $options[$permission_key] = '';
+        $form['permission'][$permission_key] = array(
           '#type' => 'item',
-          '#markup' => $perm_item['title'],
-          '#description' => theme('user_permission_description', array('permission_item' => $perm_item)),
+          '#markup' => $permission_item['title'],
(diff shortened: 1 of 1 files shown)


core/modules/user/user.admin.inc
```diff
@@ -468,6 +470,14 @@ function user_admin_permissions($form, $form_state, $role_name = NULL) {
       '#markup' => check_plain($role->label),
       '#tree' => TRUE,
     );
+
+    // Amend the entire list of checkboxes with AIRA labels for screen readers.
+    foreach ($all_permissions as $permission_key => $permission_item) {
+      $form['checkboxes'][$role_name][$permission_key]['#attributes']['aria-label'] = t('!role permission: !permission', array(
+        '!role' => $role->label,
```
docwilmot, line 477: The NVDA screen reader reads out the column title first, then reads the ARIA label, so we end up hearing the role twice. But the Chrome reader only reads the ARIA label. So I suppose this will be necessary. FYI.

core/modules/user/user.admin.inc
```diff
@@ -468,6 +470,14 @@ function user_admin_permissions($form, $form_state, $role_name = NULL) {
...
+        '!permission' => $permission_item['title'],
```
docwilmot, line 478: Some permission titles have HTML, and the reader reads this out as well, for example the Files permissions for each file type are in the format "Audio: *Edit own files*" which HTML is: `<em class="placeholder">Audio</em>: Edit own files` and which reads out verbose as "less than slash em class equal quote placeholder...". 

Changing to `strip_tags($permission_item['title'])` fixes this.

core/modules/user/user.admin.inc
quicksketch: Hm, we could set `role="presentation"` on the table or `aria-hidden` on the column headers. It would be nice to avoid duplicating the same thing being read aloud.

PR #3559, opened 2021-03-08
Issue #4217 : Config import excludes last file for archive import fixed

### LINKED ISSUE #4217: Config import is deleting files that it shouldn't
**Description of the bug**
Under specific circumstances when importing config files the last file (alphabetically) is being deleted. 

This happens when files have been added to config directory. It does not happen if there are NO change to config. 

**Steps To Reproduce**
To reproduce the behavior:
1. Set up a new site and export the config files
2. ~~Download https://github.com/backdrop-contrib/config_recipes~~ This step is not necessary.
3. Decompress the config files ~~and add the following files from config_recipes~~
~~* faq/field.instance.node.faq.body.json~~
~~* faq/node.type.faq.json~~
~~* faq/views.view.faq.json~~
EDIT: I thought that adding files was part of problem, but I tested without adding any files, just decompressing and recompressing and the problem occurred.
4. Recompress files and import the batch 
(This was my process for compressing the files, maybe my process is the problem: https://forum.backdropcms.org/forum/how-add-bundle-or-group-config-files-site)
5. Go here: admin/config/development/configuration/sync and note that the last config file is listed as "removed" - even though at no point in this process did we remove any config files. 

**Actual behavior**

Configuration_management___Kitties

**Expected behavior**

I expected to upload the original batch of config with the addition of my config recipe without loosing any config files. 

**Additional information**

1. I recognize that this happens during an 'unconventional' use of config import and that this scenario might not have been accounted for. It was a work-a-round to solve this problem:
https://github.com/backdrop/backdrop-issues/issues/3763
https://github.com/backdrop-contrib/config_recipes/issues/2

2. My tests were using my Triplo distribution that included webform module. Initially it was the webform config file that was being deleted (last config file when sorted by apha). I thought the problem might be related to that specific module for config file. I repeated the process after the webform config file had already been deleted and on the second pass, the view_ui config was deleted (which had become the last config file when sorted alpha). 

3. I have yet to test this on a raw/plain backdrop install. 

4. As stated, this is NOT a problem if one exports and then re-imports the same batch of config without decompressing and recompressing them. 

5. UDPATE steps to reproduce. There is no need to add config files to cause this problem. Decompressing and then recompressing the same batch of files causes same effect. My process for compressing files is a work-a-round to this problem: https://github.com/backdrop/backdrop-issues/issues/661

6. I found this command to help me ONLY compress and bundle the config files without their parent directory. Is this the problem? (I don't this this is the problem, because when I reopen a compressed directory of config that used this method, all the config files are there.)
```cd bd-config/ && tar -zcvf ../bd-config.tgz . && cd -```

rajpaul:
preg_grep filtered only json file to import from archive but its not reindexed when iterating. So i have fixed this using array_values to reindex the file list array. 

Fixes backdrop/backdrop-issues#4217

### DIFF
--- a/core/includes/config.inc
+++ b/core/includes/config.inc
@@ -1433,7 +1433,7 @@ class ConfigFileStorage implements ConfigStorageInterface {
       // Only extract JSON files, ignoring anything else in the archive.
       $file_list = preg_grep('/.json$/', $archiver->listContents());
       if ($file_list) {
-        $archiver->extract($this->directory, $file_list);
+        $archiver->extract($this->directory, array_values($file_list));
       }
     }
     catch (\Exception $e) {


stpaultim:
@rajpaul - Please, add this line to the first comment for this PR (you can edit it). 

`Fixes backdrop/backdrop-issues#4217`

Putting this note in the initial comment will allow Github to create a link on backdrop/backdrop-issues#4981 to this PR (in the right sidebar). This is helpful to other folks who want to find the PR quickly.

stpaultim:
@rajpaul - Sometimes if the automated tests fail, you just need to close your PR and then reopen it again to force them to run again.

PR #3561, opened 2021-03-10
Issue #4941: Keyboard block move

### LINKED ISSUE #4941: [UX] Fix for Manage blocks interface: cannot be used with keyboard navigation

docwilmot:
Fixes https://github.com/backdrop/backdrop-issues/issues/4941

Replaces https://github.com/backdrop/backdrop/pull/3531 and https://github.com/backdrop/backdrop/pull/3550

### DIFF
--- a/core/modules/layout/css/layout.admin.css
+++ b/core/modules/layout/css/layout.admin.css
@@ -392,3 +392,7 @@
 [dir="rtl"] .edit-layout-top .dropbutton-wrapper {
   float: left;
 }
+
+*:focus .layout-editor-block-title {
+  background-color: #e9eebc;
+}
--- a/core/modules/layout/css/layout.admin.css
+++ b/core/modules/layout/css/layout.admin.css
@@ -392,7 +392,3 @@
 [dir="rtl"] .edit-layout-top .dropbutton-wrapper {
   float: left;
 }
-
-.layout-editor-block:focus {
-  box-shadow: 5px 5px 5px 0 #000;
-}
--- a/core/modules/layout/css/layout.admin.css
+++ b/core/modules/layout/css/layout.admin.css
@@ -393,6 +393,6 @@
   float: left;
 }
 
-*:focus .layout-editor-block-title {
-  background-color: #e9eebc;
+.layout-editor-block:focus {
+  box-shadow: 5px 5px 5px 0 #000;
 }
--- a/core/modules/layout/js/layout.admin.js
+++ b/core/modules/layout/js/layout.admin.js
@@ -78,6 +78,7 @@ Backdrop.behaviors.layoutConfigure = {
  */
 Backdrop.behaviors.layoutDisplayEditor = {
   attach: function(context) {
+    var synth = window.speechSynthesis;
     // Apply drag and drop to regions.
     var $regions = $('.layout-editor-region-content').once('layout-sortable');
     if ($regions.length) {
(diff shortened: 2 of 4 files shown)


core/modules/layout/js/layout.admin.js
```diff
@@ -78,6 +78,7 @@ Backdrop.behaviors.layoutConfigure = {
  */
 Backdrop.behaviors.layoutDisplayEditor = {
   attach: function(context) {
+    var synth = window.speechSynthesis;
```
indigoxela: Hm, what is this var synth supposed to do?

core/modules/layout/plugins/renderers/layout_renderer_editor.inc
```diff
@@ -27,6 +27,7 @@ class LayoutRendererEditor extends LayoutRendererStandard {
     parent::addMeta();
     backdrop_add_library('system', 'ui.sortable');
     backdrop_add_js(backdrop_get_path('module', 'layout') . '/js/layout.admin.js');
+    backdrop_add_js('https://rawgit.com/jquery/jquery-ui/1-11-stable/external/jquery-simulate/jquery.simulate.js', 'external');
```
indigoxela: This library isn't used anyway, or am I missing something? If it were, we shouldn't include it like that, I suppose.

core/modules/layout/js/layout.admin.js
docwilmot: Removed

core/modules/layout/plugins/renderers/layout_renderer_editor.inc
docwilmot: Removed

core/modules/layout/css/layout.admin.css
```diff
@@ -392,3 +392,7 @@
 [dir="rtl"] .edit-layout-top .dropbutton-wrapper {
   float: left;
 }
+
+.layout-editor-block:focus {
+  box-shadow: 5px 5px 5px 0 #000;
```
indigoxela: Just a question: is this the usual design pattern to highlight focused items? It might be invisible on dark admin themes.

alanmels:
From https://github.com/backdrop/backdrop-issues/issues/4941#issuecomment-799003800: 

> Luckily tagboat was there for me this time and I've quickly tested to confirm: moving blocks works perfectly well! Great job, guys! Looking forward for it to be committed.

core/modules/layout/css/layout.admin.css
quicksketch: This focus seems "heavy" to me compared with other focus states. I do think it would be a good idea to give Seven a more substantial focus state (such as a blue or other color highlight), but this is the generic styling (not specific to Seven), which tends to stay within gray colors.

core/modules/layout/js/layout.admin.js
```diff
@@ -90,6 +90,125 @@ Backdrop.behaviors.layoutDisplayEditor = {
...
+          $('html, body').animate({
+            scrollTop: _offset.top,
+            scrollLeft: _offset.left
+          });
+        }
+      }
+
+      $('#layout-edit-main').on('keydown', '.layout-editor-block', function(event) { 
+        // Press k to move block to next region.
+        if(event.which == 75) {
```
quicksketch:
```suggestion
        if (event.which == 75) {
```

core/modules/layout/js/layout.admin.js
```diff
@@ -90,6 +90,125 @@ Backdrop.behaviors.layoutDisplayEditor = {
...
+
+          scrollIfNotVisible(nextDroppableId);
+
+          var droppable = $('#' + nextDroppableId + ' .layout-editor-region-content');
+          droppable.append($(this));
+          $(this).focus();
+        }
+
+        // Press j to block to previous region.
+        if(event.which == 74) {
```
quicksketch:
```suggestion
        if (event.which == 74) {
```

core/modules/layout/js/layout.admin.js
```diff
@@ -90,6 +90,125 @@ Backdrop.behaviors.layoutDisplayEditor = {
...
+
+          scrollIfNotVisible(nextDroppableId);
+
+          var droppable = $('#' + nextDroppableId + ' .layout-editor-region-content');
+          droppable.append($(this));
+          $(this).focus();
+        }
+
+        // Press up to move block up by one position.
+        if(event.which == 38) {
```
quicksketch:
```suggestion
        if (event.which == 38) {
```

core/modules/layout/js/layout.admin.js
```diff
@@ -90,6 +90,125 @@ Backdrop.behaviors.layoutDisplayEditor = {
...
+          $(this).insertBefore($(this).prev());
+          $(this).focus();
+        }
+        // Press down to move block down by one position.
+        if(event.which == 40) {
```
quicksketch:
```suggestion
        if (event.which == 40) {
```

core/modules/layout/js/layout.admin.js
```diff
@@ -90,6 +90,125 @@ Backdrop.behaviors.layoutDisplayEditor = {
...
+      // Find the next region.
+      var findNextDroppable = function (current, arr) {
+        var index = arr.indexOf(current);
+        if(index >= 0 && index < arr.length - 1) {
+          return arr[index + 1]
+        }
+        return false;
+      }
+
+      // Find th previous region.
```
quicksketch:
```suggestion
      // Find the previous region.
```

quicksketch (commented):
This code looks great! No real problems with the implementation to my eyes.

core/modules/layout/css/layout.admin.css
docwilmot: This was I suppose a "placeholder" for a focus CSS. I'm open to ideas, or leaving this off if its unnecessary.

core/modules/layout/layout.admin.inc
```diff
@@ -1285,10 +1285,11 @@ function layout_block_configure_form($form, &$form_state, Layout $layout, Block
   $template_info = layout_get_layout_template_info($layout->layout_template);
   $form['region'] = array(
     '#type' => 'select',
+    '#prefix' => '<div class="element-invisible">',
```
indigoxela: Is this still relevant?

I can "unhide" the select list by removing the class, but selecting another region doesn't seem to do anything. After saving the block with another region selected, the block doesn't get moved.

docwilmot: Agreed, restored to original code. We can work on adding a method to move blocks via select lists some other issue.

quicksketch:
Merged into 1.x and 1.18.x. Woohoo!

PR #3564, opened 2021-03-14
Issue #4997: Improve admin_bar accessibility by using a label

### LINKED ISSUE #4997: [A11Y] Add screenreader label to Admin Bar Menu Search Input Field
**Description of the need**

The backdrop CMS admin bar has an input field for Menu Search. Currently, this input field has no label.

`<input placeholder="Menu search" autocomplete="off" autocorrect="off" autocapitalize="off" type="search" size="60" maxlength="128" class="form-search">`

**Proposed solution**

A label should be added HTML code. My suggestion is to use the 'label for' tag method. My understanding is that bootstrap 4 is under the hood already? If so, the css tag should hide the label from displaying but work correctly with screen readers. If not, it can be added, as it can be used in various places.

```
<label for="menuSearch" class="sr-only">Administration Bar Menu Search</label>
<input id="menuSearch" placeholder="Menu search" autocomplete="off" autocorrect="off" autocapitalize="off" type="search" size="60" maxlength="128" class="form-search">
```

css

```
.sr-only{
   position:absolute;
   left:-10000px;
   top:auto;
   width:1px;
   height:1px;
   overflow:hidden;
}
```

An alternative can be using the aria-label. However, I prefer using the CSS trick, as it doesnt require the screen reader to be able to understand the aria tags. If we have no other choice, then I think aria is a reasonable solution, but using CSS should support more screenreaders.

**Further Reading**
https://blog.pope.tech/2020/02/28/missing-form-label-search/
https://github.com/twbs/bootstrap/issues/10446

indigoxela:
Fixes https://github.com/backdrop/backdrop-issues/issues/4997

### DIFF
--- a/core/modules/admin_bar/admin_bar.inc
+++ b/core/modules/admin_bar/admin_bar.inc
@@ -644,7 +644,7 @@ function admin_bar_links_search() {
   $search['search']['search'] = array(
     '#type' => 'search',
     '#title' => t('Admin search'),
-    '#title_display' => 'attribute',
+    '#title_display' => 'invisible',
     '#attributes' => array(
       'placeholder' => t('Menu search'),
       'autocomplete' => 'off',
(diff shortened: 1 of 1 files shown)


klonos (approved):
LGTM 👍🏼

laryn:
Thank you @indigoxela, @ericvanderwal and @klonos! I've merged this into 1.x and 1.19.x.

PR #3567, opened 2021-03-16
Issue #5002: The Modules page should follow Semantic Versioning when…

### LINKED ISSUE #5002: [UX] Make sure that versions shown in the Modules page are as consistent as possible
All is good for the core modules on the Modules page (`admin/modules`) as their versions always contain the Backdrop's version first and the module version after the hyphen, for example `1.19.x-dev` if you pulled the development branch from github.

As for the contributed modules, most of them show versions like, for example:

```
Automated Logout  1.x-4.5.4 
API  1.x-1.10-beta2 
API Search Integration 1.x-1.10-beta2
Coder Review  1.x-1.0.1-beta 
```
where, as I understand, the first part - `1.x` - is the main Backdrop version and the module version after the hyphen.

But some module versions get really messy.

For example, if you enable the _Forum_ module, then it shows 7.x, and that's totally the module's fault and I reported it on https://github.com/backdrop-contrib/forum/issues/39. 

However, what about modules like, for example, _Better Exposed Filters_ that shows `3.x-dev` or _BUEditor_ that shows `2.x-dev`? The system seems to be taking the last part of the default branch and add `-dev` part. Is that really how should it be or the first part of any module before the hyphen sign should always indicate the main Backdrop version?

Interestingly, if you use a downloaded zip version, for example, `https://github.com/backdrop-contrib/better_exposed_filters/archive/1.x-3.x.zip` and extract, then the Modules page shows empty for the version.

alanmels:
… displaying module versions

Resolves https://github.com/backdrop/backdrop-issues/issues/5002

### DIFF
--- a/core/modules/system/system.module
+++ b/core/modules/system/system.module
@@ -2948,7 +2948,7 @@ function _system_rebuild_module_data() {
           }
         }
         if ($git_version) {
-          $modules[$key]->info['version'] = $git_version;
+          $modules[$key]->info['version'] = $git_version . '-dev';
         }
       }
       else {
--- a/core/modules/system/system.module
+++ b/core/modules/system/system.module
@@ -2952,7 +2952,8 @@ function _system_rebuild_module_data() {
         }
       }
       else {
-        $modules[$key]->info['version'] = 'unknown';
+        $t = get_t();
+        $modules[$key]->info['version'] = $t('unknown');
       }
     }
 
--- a/core/modules/system/system.module
+++ b/core/modules/system/system.module
@@ -2952,7 +2952,7 @@ function _system_rebuild_module_data() {
         }
       }
       else {
-        $modules[$key]->info['version'] = BACKDROP_CORE_COMPATIBILITY . '-unknown';
+        $modules[$key]->info['version'] = 'unknown';
       }
     }
 
--- a/core/modules/system/system.module
+++ b/core/modules/system/system.module
@@ -2948,9 +2948,11 @@ function _system_rebuild_module_data() {
           }
         }
         if ($git_version) {
-          $git_version = preg_replace('/^' . preg_quote(BACKDROP_CORE_COMPATIBILITY, '/') . '-/', '', $git_version);
-          $modules[$key]->info['version'] = $git_version . '-dev';
+          $modules[$key]->info['version'] = $git_version;
         }
+      }
+      else {
+        $modules[$key]->info['version'] = BACKDROP_CORE_COMPATIBILITY . '-unknown';
       }
     }
 
--- a/core/modules/system/system.theme.inc
+++ b/core/modules/system/system.theme.inc
@@ -143,11 +143,12 @@ function theme_system_themes_page($variables) {
  */
 function theme_system_modules_fieldset($variables) {
   $form = $variables['form'];
+  $mixed_major_versions = $form['#mixed_major_versions'];
 
   // Individual table headers.
   $rows = array();
-  // Iterate through all the modules, which are
-  // children of this fieldset.
+
+  // Iterate through all the modules, which are children of this fieldset.
   foreach (element_children($form) as $key) {
     // Stick the key into $module for easier access.
     $module = $form[$key];
@@ -168,6 +169,12 @@ function theme_system_modules_fieldset($variables) {
     // Add the second column: module label and expand/collapse functionalty.
     $row[] = array('class' => array('module'), 'data' => '<label id="module-' . $key . '" for="' . $module['enable']['#id'] . '" class="module-name table-filter-text-source">' . backdrop_render($module['name']) . '</label>');
 
+    // If the module list does not have mixed/incompatible major versions, then
+    // remove the major version prefix ("1.x-") from all module versions, in
+    // order to show simplified version numbers.
+    if (!$mixed_major_versions) {
+      $module['version'] = preg_replace('/^' . preg_quote(BACKDROP_CORE_COMPATIBILITY, '/') . '-/', '', $module['version']);
+    }
     // Add the third column: module version number.
     $row[] = array('class' => array('version'), 'data' => backdrop_render($module['version']));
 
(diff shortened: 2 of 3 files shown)


core/modules/system/system.module
```diff
@@ -2948,10 +2948,12 @@ function _system_rebuild_module_data() {
           }
         }
         if ($git_version) {
-          $git_version = preg_replace('/^' . preg_quote(BACKDROP_CORE_COMPATIBILITY, '/') . '-/', '', $git_version);
           $modules[$key]->info['version'] = $git_version . '-dev';
         }
       }
+      else {
+        $modules[$key]->info['version'] = BACKDROP_CORE_COMPATIBILITY . '-unknown';
```
indigoxela: Assuming BACKDROP_CORE_COMPATIBILITY seems is a bit problematic to me.

I tested locally with a dummy module, that didn't even have "backdrop = 1.x" set in its info file, and the version showed up as "1.x-unknown".

alanmels: As I know BACKDROP_CORE_COMPATIBILITY is set in the core, not by a contributed module. Also it effects only the first part, the "-unknown" part is from your own suggestion in https://github.com/backdrop/backdrop-issues/issues/5002#issuecomment-800103642 That simply means your dummy module was not downloaded from github.

alanmels: In other words, when Backdrop becomes 2.x all contrib module first part also become 2.x, so BACKDROP_CORE_COMPATIBILITY is exactly what we want there.

indigoxela: Sure, but...

If I adapt the dummy module info file like that: `backdrop = 2.x` I still end up with `1.x-unknown`. I do find this problematic.

alanmels: But then the modules should follow the core version. Remember per your suggestion a module version should have format `core-major.minor.patch`, so until Backdrop core goes 2.x all modules should ideally declare `backdrop = 1.x`. Otherwise - by following what modules declared per their liking - we will have a real mess on the Modules page.

alanmels: If you insist I can change this, so the core version is derived from a contributed module's info, but then don't we risk potentially ending up with lot's of core versions on Modules page, when in fact all modules should follow core's version.

indigoxela: You seem to miss my point. :wink: Your code *assumes* that this module is for 1.x without any evidence. So the 1.x may be contradictory to the actual info file - which I wouldn't recommend.

alanmels (commented):
I got that part, but I thought that modules can not be for 2.x or higher before the core goes beyond 1.x. So I thought that if a module declares 2.x, or higher, or anything else, for example some gibberish after `backdrop = `, then it's better to force the `core` part in `core-major.minor.patch` to be just `1.x`. Maybe I'm wrong in assuming that modules can not be 2.x or higher unless the core goes that way. Please enlighten me if modules really can be for 2.x whereas the core 1.x?

core/modules/system/system.module
indigoxela: > I got that part, but I thought that modules can not be for 2.x or higher before the core goes beyond 1.x....

It's not that I insist. But I suspect you did *not* yet get what I mean. :wink: I'll give it another try:

Yes, the module can't get installed, it shows a red crossed button instead of a checkbox in the listing. But that's not my point. If adding "1.x" to the version string, you claim that this module is for core 1.x - but that's only an assumption and can be wrong. You use the constant from core (BACKDROP_CORE_COMPATIBILITY), which has nothing to do with the actual value from the module's info file. Did I explain it better now?

The check for core compatibility happens somewhere else. Our goal is only to improve the version display.

alanmels: Ok, I see what you mean. So do you want me to change the code, so the `core` part is derived from the module's info? What should it be if a contrib module didn't provide `backdrop =` line at all?

indigoxela: > So do you want me to change the code, so the core part is derived from the module's info?

That's your decision. My request was to avoid misleading info. If it's easy to achieve - sure, why not. If it's complicated, you can stick with just the word "unknown" - IMO that's already an improvement to the current display (nothing at all).

alanmels: It's not difficult as info about a module is already sitting there in variables, however taking into account the contrib modules cab write after `backdrop =` anything, I think your idea to leave just 'unknown' will provide a cleaner look. I've changed the file accordingly. Please review and thanks for your guidance to the best possible resolution!

core/modules/system/system.module
```diff
@@ -2948,10 +2948,12 @@ function _system_rebuild_module_data() {
           }
         }
         if ($git_version) {
-          $git_version = preg_replace('/^' . preg_quote(BACKDROP_CORE_COMPATIBILITY, '/') . '-/', '', $git_version);
           $modules[$key]->info['version'] = $git_version . '-dev';
         }
       }
+      else {
+        $modules[$key]->info['version'] = 'unknown';
```
indigoxela: @alanmels thank you so much for your patience. :smiley:

Here's yet another request: as we have a plain string here, it could be translatable. But be warned, this function is also called from the installer, so you would have to do it like that:

```
$t = get_t();
$t('unknown');
```

The docs: https://docs.backdropcms.org/api/backdrop/core%21includes%21bootstrap.inc/function/get_t/1
What do you think?

alanmels: Ah, I always forget about making strings translatable. Corrected. Thanks for noting.

klonos (approved):
LGTM 👍🏼

core/modules/system/system.theme.inc
```diff
@@ -168,6 +169,12 @@ function theme_system_modules_fieldset($variables) {
     // Add the second column: module label and expand/collapse functionalty.
     $row[] = array('class' => array('module'), 'data' => '<label id="module-' . $key . '" for="' . $module['enable']['#id'] . '" class="module-name table-filter-text-source">' . backdrop_render($module['name']) . '</label>');
 
+    // If the module list does not have mixed/incompatible major versions, then
+    // remove the major version prefix ("1.x-") from all module versions, in
+    // order to show simplified version numbers.
+    if (!$mixed_major_versions) {
+      $module['version'] = preg_replace('/^' . preg_quote(BACKDROP_CORE_COMPATIBILITY, '/') . '-/', '', $module['version']);
```
klonos: Sorry @alanmels, my bad. This should be:

```suggestion
      $module['version']['#markup'] = preg_replace('/^' . preg_quote(BACKDROP_CORE_COMPATIBILITY, '/') . '-/', '', $module['version']['#markup']);
```

PR #3568, opened 2021-03-16
Adding a call to module_invoke_all() after including a new config file

### LINKED ISSUE #3224: Module config in /config is not synced on install

If you create a module and put config in its /config directory, the config is copied to active during module install. However, the sync step is not run.

This leads to an inconsistent state, if the config requires database tables. For example, if a module includes configuration for fields, database tables need to be created for the fields to actually work. These would normally be created during the sync step, but this step is not run, so the field data tables are not created and the fields do not work.

1. In one Backdrop installation, create a field for a content type.
2. Copy the resulting field bundle and field instance config to the /config directory of a custom module (you might also need the content type it is on).
3. In a second Backdrop installation, install this module.
4. Attempt to create content for the content type including this field.

The field will not work correctly. If you try to add content to the content type of that field, you will see the field on the content editing form, but if you save, you will get a database error, because the field data tables do not exist in the database.

The field should work correctly. I.e., if you put config in /config for a module, and it is copied to active config storage, the sync step should be run so that the configuration actually works.

### Next Steps

We have a PR that seems to work, but we need someone to write tests. 

---
PR by @docwilmot https://github.com/backdrop/backdrop/pull/3695

hosef:
Fixes https://github.com/backdrop/backdrop-issues/issues/3224

Adding a call to module_invoke_all() after including a new config file and then also adjusting filed hooks to account for this.

### DIFF
--- a/core/modules/field/field.crud.inc
+++ b/core/modules/field/field.crud.inc
@@ -261,15 +261,6 @@ function field_validate_field($field, $update = FALSE) {
       throw new FieldException(t('The field "@name" could not be created because a database table with the name "@table" already exists.', array('@name' => $field['field_name'], '@table' => 'field_revision_' . $field['field_name'])));
     }
 
-    // Ensure the field name is unique.
-    $prior_field = field_read_field($field['field_name'], array('include_inactive' => TRUE, 'include_deleted' => TRUE));
-    if (!empty($prior_field)) {
-      $message = $prior_field['active']?
-        t('The field "@name" already exists.', array('@name' => $field['field_name'])):
-        t('The field "@name" already exists, but it is inactive. To create a field with the same name, the configuration and data from the old field must be manually deleted.', array('@name' => $field['field_name']));
-      throw new FieldException($message);
-    }
-
     // Field name cannot be longer than 32 characters. We use backdrop_strlen()
     // because the DB layer assumes that column widths are given in characters,
     // not bytes.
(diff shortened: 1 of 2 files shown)


hosef:
Fixes https://github.com/backdrop/backdrop-issues/issues/3224

laryn:
> Fixes [backdrop/backdrop-issues#3224](https://github.com/backdrop/backdrop-issues/issues/3224)

@hosef I think this has to be added in the description/first comment to automatically make the link in the "Linked issue" section.

core/modules/field/field.crud.inc
```diff
@@ -261,15 +261,6 @@ function field_validate_field($field, $update = FALSE) {
       throw new FieldException(t('The field "@name" could not be created because a database table with the name "@table" already exists.', array('@name' => $field['field_name'], '@table' => 'field_revision_' . $field['field_name'])));
     }
 
-    // Ensure the field name is unique.
-    $prior_field = field_read_field($field['field_name'], array('include_inactive' => TRUE, 'include_deleted' => TRUE));
-    if (!empty($prior_field)) {
-      $message = $prior_field['active']?
-        t('The field "@name" already exists.', array('@name' => $field['field_name'])):
-        t('The field "@name" already exists, but it is inactive. To create a field with the same name, the configuration and data from the old field must be manually deleted.', array('@name' => $field['field_name']));
```
quicksketch, line 269: I don't think we should be entirely removing these checks. Is there a way we can keep them while still having the config import on install work properly?

oadaeh: Since this "Validates a field configuration," would it be reasonable to place a `$config->isNew()` check (and related code) at an appropriate place in the function?

jenlampton:
Extended by https://github.com/backdrop/backdrop/pull/3695

PR #3574, opened 2021-03-23
Issue 5018: Node previews, wrong layout

### LINKED ISSUE #5018: Node previews: wrong layout
**Description of the bug**

When previewing a new or edited node, the system is supposed to use the layout that will apply when the node is normally rendered. It's not currently doing that, typically using the default layout.

**Steps To Reproduce**

1. Create a new layout with path `node/%`.
2. Add a custom block to it with some distinctive content so you can tell when it's being applied.
3. Create a new node and select 'Preview'.

The previewed node will be displayed with the default layout, not the custom layout that you just set up.

**What's going on**

The code introduced in [PR 2290](https://github.com/backdrop/backdrop/pull/2290) (in response to [Issue 3062](https://github.com/backdrop/backdrop-issues/issues/3062)) in function `layout_get_layout_by_path()` no longer does what it's supposed to, because:

* The path for preview nodes is different from what is in the code
* the `$tempstore_id` that was being requested was not returning the proper value.

There's also a problem a bit further down in that function in the setting of context data. The path for a node preview (e.g., `node/page/preview/%`) has its context in a different position from the position of the placeholder in the relevant layout path (e.g., `node/%`), which needs to be accounted for when setting context based on information in the router item.

A PR that fixes this will be forthcoming shortly.

This seems like a situation that should have a test as part of the test suite, and I'm willing to take on adding that, too. A question on logistics though: should the test be added in the same PR as the fix, or should the test be in a separate PR? (Of course the test would fail without the fix.)

**Additional information**

- Backdrop CMS version 1.18

bugfolder:
Fixes [#5018](https://github.com/backdrop/backdrop-issues/issues/5018).

If a custom layout is defined for a node path, use that layout for node preview pages.

Current code uses the default layout for preview. This fix will use the correct custom layout if one is defined.

### Setup

1. Create a new layout with path `node/%`.
2. Add a custom block to it with some distinctive content so you can tell when it's being applied.

(optional)

3. Add a custom block that uses a context, e.g., "Field: Body (body)". This will let you verify that the context is being properly set.

### Test: create new node

1. Create a new node, e.g., at `node/add/post`.
2. Select 'Preview'.
3. The preview page should show the custom layout and if the block with context was added, the block contents should be shown.

### Test: edit existing node

1. Select the edit page for an existing node, e.g., `node/1/edit`.
2. Select 'Preview'.
3. The preview page should show the custom layout and if the block with context was added, the block contents should be shown.

### DIFF
--- a/core/modules/layout/layout.module
+++ b/core/modules/layout/layout.module
@@ -1224,7 +1224,7 @@ function layout_get_layout_by_path($path = NULL, $router_item = NULL) {
     $preview_node = NULL;
     $node_types = node_type_get_types();
     foreach ($node_types as $type) {
-      $preview_path = "node/preview/{$type->type}/%";
+      $preview_path = 'node/preview/' . $type->type . '/%';
       if ($router_item['path'] == $preview_path) {
         $tempstore_id = $router_item['map'][3];
         $preview_node = node_get_node_tempstore($tempstore_id);
--- a/core/modules/layout/layout.module
+++ b/core/modules/layout/layout.module
@@ -1221,18 +1221,18 @@ function layout_get_layout_by_path($path = NULL, $router_item = NULL) {
     // @ todo: This adds very specific code for node previews which too closely
     // couples Node and Layout modules. This needs to be reverted in a future
     // commit.
-    $data = NULL;
+    $preview_node = NULL;
     $node_types = node_type_get_types();
     foreach ($node_types as $type) {
-      $preview_path = 'node/' . $type->type . '/preview';
+      $preview_path = "node/preview/{$type->type}/%";
       if ($router_item['path'] == $preview_path) {
-        $tempstore_id = node_build_tempstore_id($type->type);
-        $data = node_get_node_tempstore($tempstore_id);
+        $tempstore_id = $router_item['map'][3];
+        $preview_node = node_get_node_tempstore($tempstore_id);
         break;
       }
     }
-    if ($data && $node_layouts = layout_load_multiple_by_path('node/%', TRUE)) {
-      $layouts = $node_layouts;
+    if ($preview_node) {
+      $layouts = layout_load_multiple_by_path('node/%', TRUE);
     }
     else {
       $layouts = layout_load_multiple_by_path($router_item['path'], TRUE);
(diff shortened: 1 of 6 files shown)


core/modules/layout/layout.module
```diff
@@ -1244,21 +1244,27 @@ function layout_get_layout_by_path($path = NULL, $router_item = NULL) {
       foreach ($contexts as $context) {
         if (isset($context->position)) {
 
-          // Check for an object loaded by the menu router. Use a preview from
-          // Tempstore if on a node preview page.
+          // If on a node preview page use the preview from tempstore.
```
klonos:
```suggestion
          // If on a node preview page, use the preview from tempstore.
```

klonos (approved):
LGTM 👍🏼

core/modules/layout/layout.module
```diff
@@ -1221,18 +1221,18 @@ function layout_get_layout_by_path($path = NULL, $router_item = NULL) {
     // @ todo: This adds very specific code for node previews which too closely
     // couples Node and Layout modules. This needs to be reverted in a future
     // commit.
-    $data = NULL;
+    $preview_node = NULL;
     $node_types = node_type_get_types();
     foreach ($node_types as $type) {
-      $preview_path = 'node/' . $type->type . '/preview';
+      $preview_path = "node/preview/{$type->type}/%";
```
docwilmot: We tend to opt for `$preview_path = "node/preview/' . $type->type . '/%"`

bugfolder: Changed in the PR. (Guessing you intended to use single quotes all the way there.)

jenlampton: Thanks, this is a micro-performance improvement we've been following throughout core because double quotes `"`" are slightly slower since PHP is parsing the string for those variable replacements. Single quotes `''` go a bit faster :)

bugfolder:
This now has the tests from https://github.com/backdrop/backdrop/pull/3575 merged in.

quicksketch:
Thanks @bugfolder! Merged into 1.x and 1.18.x.

PR #3577, opened 2021-03-28
Issue #5028: Moving field_read_fields() out of a foreach loop so it is only called once in field_sync_field_status().

### LINKED ISSUE #5028: [PS] Optimize field_sync_field_status()
There is a function in the field module called `field_sync_field status()`. This is run every time one of 4 events happens:

- When cron is run
- When caches are cleared
- When a module is enabled
- When a module is disabled

Inside this function we run a loop that calls `field_associate_fields()` per installed module, and in there we call `field_read_fields()`. During this process it does a scan of the config directory for every module installed on the site, which gets more expensive as you get more modules or more config files.

We should optimize this to reduce the number of times that we call `field_read_fields()`

hosef:
Fixes: https://github.com/backdrop/backdrop-issues/issues/5028

### DIFF
--- a/core/modules/field/field.module
+++ b/core/modules/field/field.module
@@ -626,6 +626,8 @@ function field_sync_field_status() {
  * @param $fields
  *   An optional array of field definitions to operate on. If left empty, all
  *   fields will be loaded.
+ *
+ * @since 1.18.4 Second parameter $fields added for performance optimization.
  */
 function field_associate_fields($module, &$fields = NULL) {
   // Associate field types.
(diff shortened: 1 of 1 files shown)


core/modules/field/field.module
```diff
@@ -622,24 +623,38 @@ function field_sync_field_status() {
  *
  * @param $module
  *   The name of the module to update on.
+ * @param $fields
+ *   An optional array of field definitions to operate on. If left empty, all
+ *   fields will be loaded.
  */
```
quicksketch:
```suggestion
 * @param $fields
 *   An optional array of field definitions to operate on. If left empty, all
 *   fields will be loaded.
 *
 * @since 1.18.4 Second parameter $fields added for performance optimization.
 */
```

PR #2689, opened 2019-05-19
Issue #3794: DateThemeTestCase fails depending on testing time

indigoxela:
Rewriting testDateDisplayCombination() to prevent tests failing depending on testing time.

See https://github.com/backdrop/backdrop-issues/issues/3794 for details.

PR #3578, opened 2021-03-29
Issue #2689: Simplifying Basis' tabledrag styles and fixing indent

wesruv:
Indentation's styles are very tempermental, seem to have found a balance that works, the float on div.indentation is an enormous headache


gitressa:
Thanks! Just a tip: You can link an issue with a PR via the first comment by including `Fixes *ISSUE_URL*`.

> **Development**
> bugfolder/backdrop

So if you update the syntax in this PR, your issue will probably also be linked in https://github.com/backdrop/backdrop-issues/issues/2689.

PR #3579, opened 2021-03-29
Issue #4890: Rename display mode 'Enable' buttons to 'Customize'.

### LINKED ISSUE #4890: Rename 'Enable' buttons for display modes
Display modes can be 'customised', or 'not customised'. As discovered in https://github.com/backdrop/backdrop-issues/issues/3036 (see that issue for more background on this one), the 'not customised' ones aren't 'disabled' per se since you can still select them, for example, in Views' settings - they just end up using the Default display mode's settings.

image

On the Manage Displays page, 'customised' display modes can be 'reset' (makes sense 👍🏻 )
However, 'not customised' ones can be 'enabled'... (huh? 👎🏻 )

Recommend changing the 'Enable' text to 'Customise' instead.

ghost:
Fixes https://github.com/backdrop/backdrop-issues/issues/4890


quicksketch:

PR #3583, opened 2021-04-07
Issue 5034: Call to undefined function module_implements() in backdro…

### LINKED ISSUE #5034: Fix error: Call to undefined function `module_implements()` in `backdrop_autoload()`
**Description of the bug**

As reported on the forum: 

 - https://forum.backdropcms.org/forum/installation-error-call-undefined-function-moduleimplements-backdropautoload

> Error: Call to undefined function module_implements() in backdrop_autoload() (line 3999 of mysite/webroot/core/includes/bootstrap.inc).

I ran into this, too, as did another responder in the forum -- unfortunately I haven't been able to figure out what's triggering it. The site in question is a CiviCRM site (as was @bugfolder's site) so maybe something there. (The line number has changed since the original report, of course. I think it was 4012 or so for me.)

The fix @bugfolder suggests does function as a workaround, I'm not sure if it seems like a good solution for core or not. The workaround is to add: 

> include_once BACKDROP_ROOT . '/core/includes/module.inc';

before

> foreach (module_implements('autoload_info') as $module) {

bugfolder:
…p_autoload() (D7 Upgrade)

Fixes an issue that occurs during update.php when upgrading a D7 site that has a classed variable serialized in the variables table.

Fixes https://github.com/backdrop/backdrop-issues/issues/5034.

### DIFF
--- a/core/includes/bootstrap.inc
+++ b/core/includes/bootstrap.inc
@@ -4009,6 +4009,7 @@ function backdrop_autoload($class) {
     // module_invoke_all() because it uses array_merge_recursive(), which is an
     // expensive O(n^2) operation. Instead we use module_implements() and merge
     // the array manually, which is O(n).
+    include_once BACKDROP_ROOT . '/core/includes/module.inc';
     foreach (module_implements('autoload_info') as $module) {
       $module_classes = module_invoke($module, 'autoload_info');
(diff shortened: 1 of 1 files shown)


klonos (approved):
LGTM 👍🏼

quicksketch:

PR #3586, opened 2021-04-12
Issue #5037: Trying to sort NULL array in field_ui_widget_type_options

### LINKED ISSUE #5037: Trying to sort NULL array in field_ui_widget_type_options()
**Description of the bug**

A module can decide to **support**  a field type, by providing a new widget, by listing this in `hook_field_widget_info()`:
```
function my_module_field_widget_info() {
  return array(
    'my_module' => array(
      'label' => t('My module'),
      'field types' => array(
        'a_field_provided_by_another_module',
      ),
    ),
  );
}
```
If the field type doesnt exist however, there is a `Warning: asort() expects parameter 1 to be array, null given in field_ui_widget_type_options() (line 1517` error in the logs. This is because Backdrop has added a `asort($options[$widget_field_type])` line in that function, when in fact for our `a_field_provided_by_another_module` no options are available because the module is not installed.

**Steps To Reproduce**
Implement `my_module_field_widget_info()` as above and add a random string. In D7 this would be fine, in Backdrop theres watchdog errors about `aosort` and missing indexes.

**Expected behavior**

In D7 you could do this so that your new module could provide a widget for this field type *if* the user ever installed it. 

**Additional information**

In core since https://github.com/backdrop/backdrop/commit/656c8945be0089de1d3bf394eacb82528bc35846

indigoxela:
Fixes https://github.com/backdrop/backdrop-issues/issues/5037

Make sure the array is not empty before trying to sort it.

### DIFF
--- a/core/modules/field_ui/field_ui.admin.inc
+++ b/core/modules/field_ui/field_ui.admin.inc
@@ -1514,7 +1514,7 @@ function field_ui_widget_type_options($field_type = NULL, $by_label = FALSE) {
           $options[$widget_field_type][$name] = $widget_type['label'];
         }
       }
-      if (!empty($options)) {
+      if (isset($options[$widget_field_type])) {
         asort($options[$widget_field_type]);
       }
     }
--- a/core/modules/field_ui/field_ui.admin.inc
+++ b/core/modules/field_ui/field_ui.admin.inc
@@ -1514,7 +1514,9 @@ function field_ui_widget_type_options($field_type = NULL, $by_label = FALSE) {
           $options[$widget_field_type][$name] = $widget_type['label'];
         }
       }
-      asort($options[$widget_field_type]);
+      if (!empty($options)) {
+        asort($options[$widget_field_type]);
+      }
     }
   }
 


klonos (approved):
LGTM 👍🏼

core/modules/field_ui/field_ui.admin.inc
```diff
@@ -1514,7 +1514,9 @@ function field_ui_widget_type_options($field_type = NULL, $by_label = FALSE) {
           $options[$widget_field_type][$name] = $widget_type['label'];
         }
       }
-      asort($options[$widget_field_type]);
+      if (!empty($options)) {
```
quicksketch: Should this be checking `$options[$widget_field_type]`? I could see `$options` being set for some but not all of the widget types in the `foreach()` loop we're in side of.

```suggestion
      if (isset($options[$widget_field_type])) {
```

indigoxela: Doesn't that duplicate the check from line 1513? The loop above only adds the item to the `$options`array 
`if (isset($field_types[$widget_field_type]))`. So at that point (line 1517) I'd assume that any entry in the array has both, name and label set properly. Or am I missing something?

indigoxela: @quicksketch any chance you find the time to look at my answer comment?

quicksketch: >  The loop above only adds the item to the `$options` array `if (isset($field_types[$widget_field_type]))`

Other values might already be in `$options` from a previous loop. There are two levels of looping here. It's possible that the first pass on `foreach (field_info_widget_types() as $name => $widget_type) {` added something into `$options` already, and the second pass may not have put something in `$options[$widget_field_type]`. Even if that might not occur in common scenarios, it seems like we'd be safer checking the exact key exists before attempting to sort its value.

quicksketch:
Merged into 1.x and 1.19.x. Thanks for your work and patience @indigoxela. :bow:

PR #3587, opened 2021-04-16
Issue #5041: Notice: Undefined index: form_build_id in ajax_get_form(…

### LINKED ISSUE #5041: Fix PHP Notice: Undefined index: form_build_id in ajax_get_form()
**Steps To Reproduce**

To reproduce the behavior just go to http://backdrop.site/system/ajax and the log will show the following notice:

> Notice: Undefined index: form_build_id in ajax_get_form() (line 323 of /var/www/docroot/core/includes/ajax.inc).

and the warning:

> Invalid form POST data.

**Additional information**

Normally, the `system/ajax` page should not be accessed directly. Regardless, an additional check could help to avoid the above warning and notice.

alanmels:
…) (line 323 of core/includes/ajax.inc).

Resolves https://github.com/backdrop/backdrop-issues/issues/5041

### DIFF
--- a/core/includes/ajax.inc
+++ b/core/includes/ajax.inc
@@ -319,9 +319,10 @@ function ajax_render($commands = array()) {
  */
 function ajax_get_form() {
   // Exit early if $_POST is empty.
-  if (empty($_POST)) {
+  if (empty($_POST) || !isset($_POST['form_build_id'])) {
     backdrop_exit();
   }
+
   $form_state = form_state_defaults();
 
   $form_build_id = $_POST['form_build_id'];
--- a/core/includes/ajax.inc
+++ b/core/includes/ajax.inc
@@ -318,7 +318,7 @@ function ajax_render($commands = array()) {
  *   @endcode
  */
 function ajax_get_form() {
-    // Exit early if $_POST is empty.
+  // Exit early if $_POST is empty.
   if (empty($_POST)) {
     backdrop_exit();
   }
--- a/core/includes/ajax.inc
+++ b/core/includes/ajax.inc
@@ -318,6 +318,10 @@ function ajax_render($commands = array()) {
  *   @endcode
  */
 function ajax_get_form() {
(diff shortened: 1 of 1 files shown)


core/includes/ajax.inc
```diff
@@ -318,6 +318,10 @@ function ajax_render($commands = array()) {
  *   @endcode
  */
 function ajax_get_form() {
+    // Exit early if $_POST is empty.
```
klonos:
```suggestion
  // Exit early if $_POST is empty.
```

core/includes/ajax.inc
```diff
@@ -318,6 +318,10 @@ function ajax_render($commands = array()) {
  *   @endcode
  */
 function ajax_get_form() {
+  // Exit early if $_POST is empty.
+  if (empty($_POST)) {
```
indigoxela: One thought: form_build_id is essential in this function.

What if something gets posted (garbage stuff), but the form_build_id is still missing. Shouldn't we also check for that?

```
  if (empty($_POST) || !isset($_POST['form_build_id'])) {
    backdrop_exit();
  }
```
@alanmels what do you think?

alanmels: I've run checks many times and it's either `$_POST` is empty or it always contains `form_build_id`, so I thought checking the `$_POST` was sufficient. Can you confirm my finding?

indigoxela: I'm not sure, it was merely an idea based on the fact that nothing can be done in that function without the "form_build_id".

I'll leave it to you, to check for it or not. Checking for an empty $_POST will for sure fix most problems anyway.

indigoxela: Consider this discussion as resolved. :smile:

alanmels: Made the change, thanks for reviewing!

quicksketch:
Thanks @alanmels! Merged into 1.x and 1.18.x.

PR #3593, opened 2021-04-17
Issue #5046: Fix caching for the Modules page

### LINKED ISSUE #5046: Fixed performance on the Modules listing page /admin/modules
**Description of the bug**

I've been always bothered by how quickly the Functionality (admin/modules) page gets really slow with larger number of contributed modules enabled. So I decided to find out the culprit. 

**Steps To Reproduce**

1. Enable the Devel module;
2. Open the `system.module`, find the `system_rebuild_module_data()` function and put `sdpm($modules_cache);` like so:

```
function system_rebuild_module_data() {
  $modules_cache = &backdrop_static(__FUNCTION__);
  sdpm($modules_cache);
```
Refresh the modules page and see `$modules_cache` always returns empty, so the if condition further below is completely useless. 

However, caching for the modules page starts working as expected if the `$modules_cache = &backdrop_static('__FUNCTION__');` is replaced with `$modules_cache = &backdrop_static('modules_cache');` like so:

```
function system_rebuild_module_data() {
  $modules_cache = &backdrop_static('modules_cache');
  // Only rebuild once per request. $modules and $modules_cache cannot be
  // combined into one variable, because the $modules_cache variable is reset by
  // reference from system_list_reset() during the rebuild.
  if (!isset($modules_cache)) {
    $modules = _system_rebuild_module_data();
    ksort($modules);
    system_get_files_database($modules, 'module');
    system_update_files_database($modules, 'module');
    $modules = _module_build_dependencies($modules);
    $modules_cache = $modules;
  }
  return $modules_cache;
}
```

Finally, I have really fast Modules page.

To describe the problem: Two modules try to reset the caches on /admin/modules.

- The system module (which provides the page)
- The update module (which has nothing to do with this page)

To my understanding this happened by mistake (many years back) and should get fixed, so that only the module that's _responsible_ for that page flushes caches on that page.

---
PR by @docwilmot https://github.com/backdrop/backdrop/pull/3601

alanmels:
Fixes https://github.com/backdrop/backdrop-issues/issues/5046

### DIFF
--- a/core/modules/system/system.module
+++ b/core/modules/system/system.module
@@ -2990,7 +2990,7 @@ function _system_rebuild_module_data() {
  *   Array of all available modules and their data.
  */
 function system_rebuild_module_data() {
-  $modules_cache = &backdrop_static(__FUNCTION__);
+  $modules_cache = &backdrop_static('modules_cache');
   // Only rebuild once per request. $modules and $modules_cache cannot be
   // combined into one variable, because the $modules_cache variable is reset by
   // reference from system_list_reset() during the rebuild.
--- a/core/modules/system/system.module
+++ b/core/modules/system/system.module
@@ -2990,19 +2990,15 @@ function _system_rebuild_module_data() {
  *   Array of all available modules and their data.
  */
 function system_rebuild_module_data() {
-  $modules_cache = &backdrop_static('modules_cache');
   // Only rebuild once per request. $modules and $modules_cache cannot be
   // combined into one variable, because the $modules_cache variable is reset by
   // reference from system_list_reset() during the rebuild.
-  if (!isset($modules_cache)) {
-    $modules = _system_rebuild_module_data();
-    ksort($modules);
-    system_get_files_database($modules, 'module');
-    system_update_files_database($modules, 'module');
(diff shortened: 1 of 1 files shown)


core/modules/system/system.module
```diff
@@ -2990,7 +2990,7 @@ function _system_rebuild_module_data() {
  *   Array of all available modules and their data.
  */
 function system_rebuild_module_data() {
-  $modules_cache = &backdrop_static(__FUNCTION__);
+  $modules_cache = &backdrop_static('modules_cache');
```
docwilmot: The problem with this fix is that all calls to `backdrop_static_reset('system_rebuild_module_data')` become ineffective.

alanmels: Thanks for the update, @docwilmot, and yes, let's find the right solution. However, leaving the function as is also doesn't' seem right, because `$modules_cache` never gets populated, it is always empty.

quicksketch:
Closing in favor of https://github.com/backdrop/backdrop/pull/3601 as mentioned in https://github.com/backdrop/backdrop-issues/issues/5046#issuecomment-835582145.

PR #3602, opened 2021-04-19
Issue #5049: Make modules clean up their configuration files

### LINKED ISSUE #5049: [DX] API documentation: mention that it's not necessary to delete own config on uninstall
We could add documentation stating why it's not necessary to delete config files in `hook_uninstall()`. I think @indigoxela said it best:

> As long as a module properly implements `hook_config_info()`, it's not necessary to delete this config on uninstall. If, however, a module writes to config files provided by other modules, it should remove its own values from those other files.

---

Original issue:

**Description of the bug**

The https://docs.backdropcms.org/api/backdrop/core%21modules%21system%21system.api.php/function/hook_uninstall/1 pages gives this example code:

```
function hook_uninstall() {
  state_del('my_module_last_cron');
}
```
However, while clearing specific states or variables is a good practice upon a module uninstall, the configuration files must also be deleted. While https://github.com/backdrop/backdrop-issues/issues/4189 is about enabling core to cleanup after lazy modules that didn't delete their configuration files, this one is different in that I suggest to slightly change the API doc adding one more line like:

```
function hook_uninstall() {
  state_del('my_module_last_cron');
  config('my_module.settings')->delete();
}
```

which would make it easier for all Backdrop developers, especially new ones, copying the function from the https://docs.backdropcms.org/api/backdrop/core%21modules%21system%21system.api.php/function/hook_uninstall/1 page to also grab the deletion line, (instead of additionally searching on how to automatically delete the configuration file - nobody does this). One extra line doesn't hurt, but it can help to keep lot's of Backdrop sites cleaner.

alanmels:
Fixes https://github.com/backdrop/backdrop-issues/issues/5049

### DIFF
--- a/core/modules/system/system.api.php
+++ b/core/modules/system/system.api.php
@@ -3001,6 +3001,7 @@ function hook_update_last_removed() {
  */
 function hook_uninstall() {
   state_del('my_module_last_cron');
+  config('my_module.settings')->delete();
 }
 
 /**


klonos (approved):
LGTM 👍🏼

PR #3608, opened 2021-04-26
Issue 5067: Move node preview layout code into node module via hooks

### LINKED ISSUE #5067: Move node preview layout code into node module via new hook in layout module
This is a follow-on to https://github.com/backdrop/backdrop-issues/issues/5018, to take on the longer-term plan outlined in [comment](https://github.com/backdrop/backdrop-issues/issues/5018#issuecomment-825210749), as determined during the development meeting on 4/22:

> Separately, I will proceed with long-term solution of adding hooks to layout module and moving node-specific code into node module implementation of the hooks.

This is under way, and I will post a PR when I have something ready to review.

Advocate: @bugfolder

bugfolder:
Fixes https://github.com/backdrop/backdrop-issues/issues/5067.


bugfolder:
Hmm, looks like some tests are failing. Will investigate further.

core/modules/layout/layout.module
```diff
@@ -1214,83 +1214,54 @@ function layout_get_layout_by_path($path = NULL, $router_item = NULL) {
...
-      foreach ($contexts as $context) {
-        if (isset($context->position)) {
-
-          // If on a node preview page, use the preview from tempstore.
-          // @ todo: This is also specific code for node previews which needs
-          // to be reverted in a future commit.
-          if ($preview_node) {
-            $context->setData($preview_node);
+  // Use the static cache, but based on the normal path (not system path).
+  $selected_layout = &backdrop_static(__FUNCTION__ . '/' . $router_item['href']);
```
docwilmot, line 1218: This would complicate resetting the static variable. To call `backdrop_static_reset('layout_get_layout_by_path/' . $href)` would be cumbersome for one, and you'd need to know what is `$href` while calling.

core/modules/layout/layout.module
```diff
@@ -1214,83 +1214,54 @@ function layout_get_layout_by_path($path = NULL, $router_item = NULL) {
...
+  if (isset($selected_layout)) {
+    return $selected_layout;
+  }
+
+  $layouts = layout_load_multiple_by_path($router_item['path'], TRUE);
+  backdrop_alter('layout_layouts', $layouts, $router_item);
```
docwilmot, line 1224: Would it be reasonable here to just add the alter invocation from within `layout_load_multiple_by_path()`? That way we avoid having to invoke after *every* call to that function?

Also, I think the alter would need a more descriptive name; how about `backdrop_alter('layout_load_multiple_by_path', $layouts, $router_item)`?

core/modules/layout/layout.api.php
```diff
@@ -335,6 +335,109 @@ function hook_layout_presave(Layout $layout) {
...
+ *   The router item for the path. In addition to getting the system path at
+ *   $router_item['path'], you can use other elements of $router_item. For
+ *   example, $router_item['original_map'] contains an exploded version of the
+ *   normal path.
+ *
+ * @see layout_get_layout_by_path()
+ * @see node_layout_contexts_alter()
+ */
+function hook_layout_contexts_alter(Layout $layout, $router_item) {
+  // Example implementation: just to confuse people, if we're looking at node 1,
```
docwilmot, line 424: I dont think we should add examples that confuse people. 🙂

bugfolder: Good point. I was just trying to come up with something non-trivial as a teaching example. Do you think the example is OK if I take out the "confusing" comment?

core/modules/layout/layout.module
bugfolder: > Would it be reasonable here to just add the alter invocation from within layout_load_multiple_by_path()? That way we avoid having to invoke after every call to that function?

I discussed that possibility in my (long) comment on the issue. Two reasons not to:
1. You need data from the normal path (which you can get from the router item), but `layout_load_multiple_by_path()` only receives the system path as an argument.
2. There's a few cases where you _don't_ want to invoke after every call to `layout_load_multiple_by_path()`. The case in function `dashboard_menu()` is an example of that.

core/modules/layout/layout.module
bugfolder: True, good point. I can go back to the previous approach that uses `&backdrop_static(__FUNCTION__)`, but still use `'href'` as the key.

core/modules/layout/layout.module
docwilmot: Ah, hadnt fully read your long comment. 🙂 

I dont think that either of those two situations would be a problem. If you invoke the hook, its up to the caller to do the checks to determine *if* they want to alter. So Node module would do something like:
```
function node_layout_layouts_alter(&$layouts, $path) {
  // Check very carefully if this is a preview path
  ...
  // Check carefully if there really is a menu item available on this path
  ...
  // If above OK, do the alter, if not, do nothing. 
}
```
Any other call to `layout_load_multiple_by_path()` would be perfectly safe then.

bugfolder: The first one is still a problem. The `$path` that is passed to `layout_load_multiple_by_path()` is a _system path_, like `node/%`. But an implementer may need access to the _normal path_, like `node/1`. That is not easily available inside of `layout_load_multiple_by_path()`.

bugfolder: And for the second case, it could result in some very hard-to-diagnose bugs down the road. Suppose we put the invocation inside of `layout_load_multiple_by_path()`. That means that in `dashboard_menu()`, the hook would be invoked. Now, if we're passing the router item to the hooks, we need to rebuild the router item inside of `layout_load_multiple_by_path()`, when all we have is the system path (e.g., `node/%`). There's really no good way to do that. We could call `get_menu_item()`, but if the menu happened not to be built at that invocation, that call would cause the menus to rebuild, resulting in the infinite recursion that I mentioned in the final edit to my long comment. (Said infinite recursion is only sporadic, which is why things worked on my local but failed in testing.)

Also, someone might call `layout_load_multiple_by_path()` for some path other than the current path, and then there's _really_ no way to know what the applicable normal path should be inside the call.

core/modules/layout/layout.api.php
bugfolder: Maybe an equivalent example would be "Always use node 1 as the context no matter what node is being viewed."

core/modules/layout/layout.module
bugfolder: Just noticed that I skipped over your comment:

>Also, I think the alter would need a more descriptive name; how about 'backdrop_alter('layout_load_multiple_by_path', $layouts, $router_item)'? 

That would make the hook `hook_layout_load_multiple_by_path_alter()`.

So, I guess that's a matter of personal style (which might be overridden by house style, of course). My rationale was to keep them both short (but not too short), so choosing the name to specify just what is being altered: `layouts` in one, `contexts` in the other.

Because the two hooks work together, I think there's value in having them stylistically similar in name.

What about `hook_layout_layouts_load_alter()` and `hook_layout_contexts_load_alter()`? There we've introduced the action verb into the name, but they're still not too long and are similar, reinforcing the notion that they work together.

bugfolder:
Closing in favor of https://github.com/backdrop/backdrop/pull/3618.

PR #3609, opened 2021-04-26
Adding a call to resetAll() during the test prefix setup.

hosef:


gitressa:
Thanks! Just a tip: Ideally you should create an issue in [the Backdrop issue tracker](https://github.com/backdrop/backdrop-issues/issues?state=open) for the changes you are making, and then submit a Pull Request (PR) back to the Backdrop project for review.

https://docs.backdropcms.org/documentation/contribute-to-backdrop-core

The syntax to link an issue with a PR via the first comment is `Fixes *ISSUE_URL*`.

> **Development**
> bugfolder/backdrop

So if you create an issue and add the line `Fixes *ISSUE_URL*` here, your issue and the PR can also be linked.

PR #3614, opened 2021-05-02
parent 23004ee7fdad5f673e70aba0ad2bfce601358f28 author Greg Netsas <gregory@userfriendly.tech> 1608339283 +0200 committer Jen Lampton <jen@jeneration.com> 1619997684 -0700

### LINKED ISSUE #2714: [UX] Merge all the "Update core/module/theme/layout" pages into a single page.

jenlampton:
Issue #2714: Merge all the 'Update' pages into one.

Fixes: https://github.com/backdrop/backdrop-issues/issues/2714

Replaces: https://github.com/backdrop/backdrop/pull/3451

### DIFF
--- a/core/modules/dashboard/includes/block.menus.inc
+++ b/core/modules/dashboard/includes/block.menus.inc
@@ -97,10 +97,10 @@ class DashboardMenuBlock extends Block {
     }
 
     if (empty($rows)) {
-      // If there are existing menuse, but user has no access to manage any
-      // of them, hide the block completely.
+      // If there are existing menus, but user has no access to manage any of
+      // them, hide the block completely.
       if (!empty($menus)) {
-        return array();
+        return;
       }
       $rows[] = array(
         array(
--- a/core/modules/dashboard/includes/block.update.inc
+++ b/core/modules/dashboard/includes/block.update.inc
@@ -1,7 +1,7 @@
 <?php
 /**
  * @file
- * "Update" dashboard block.
+ * Available Updates dashboard block.
  */
 
 class DashboardUpdateBlock extends Block {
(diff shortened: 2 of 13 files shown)


quicksketch:
@klonos has merged these changes into https://github.com/backdrop/backdrop/pull/3451.

PR #3615, opened 2021-05-03
Issue #5074: Create a theme_layout_action_link() function.

### LINKED ISSUE #5074: Fix layout action links so additional links can be added to the list.
**Description of the bug**

There seem to be two `UL` tags with the class `action-links` being added to the page, rather than two links in the same list: 

Screen Shot 2021-05-02 at 6 49 21 PM

This is happening because both links fake the action-links `UL` by adding it with prefix and suffix:
```
'#prefix' => '<ul class="action-links"><li>',
'#suffix' => '</li></ul>',
```

This makes it impossible to add a second item into the action-links list.

jenlampton:
Fixes https://github.com/backdrop/backdrop-issues/issues/5074

### DIFF
--- a/core/modules/layout/layout.admin.inc
+++ b/core/modules/layout/layout.admin.inc
@@ -329,20 +329,22 @@ function layout_settings_form($form, &$form_state, Layout $layout) {
     '#parents' => array('context'),
   );
 
-  $form['context_wrapper']['context']['add'] = array(
-    '#type' => 'submit',
-    '#value' => t('Add context'),
-    '#attributes' => array('class' => array('layout-link-button', 'layout-access-add')),
-    '#validate' => array(),
-    '#submit' => array(
-      'layout_settings_form_update_layout',
-      'layout_settings_form_context_add',
+  $form['context_wrapper']['context']['links'] = array(
+    '#theme' => 'layout_list',
+    '#attributes' => array('class' => array('action-links', 'layout-action-links')),
+    'add' => array(
+      '#type' => 'submit',
+      '#value' => t('Add context'),
+      '#attributes' => array('class' => array('layout-link-button', 'layout-access-add')),
+      '#validate' => array(),
+      '#submit' => array(
+        'layout_settings_form_update_layout',
+        'layout_settings_form_context_add',
+      ),
+      '#ajax' => array(
+        'callback' => 'layout_ajax_form_open_dialog',
+      ),
     ),
-    '#ajax' => array(
-      'callback' => 'layout_ajax_form_open_dialog',
-    ),
-    '#prefix' => '<ul class="action-links"><li>',
-    '#suffix' => '</li></ul>',
   );
 
   // Get contexts from relationships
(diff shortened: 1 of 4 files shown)


core/modules/layout/layout.admin.inc
```diff
@@ -2471,18 +2483,24 @@ function layout_menu_item_form($form, &$form_state, LayoutMenuItem $menu_item) {
...
-  $form['conditions']['add'] = array(
-    '#type' => 'submit',
-    '#value' => t('Add visibility condition'),
-    '#attributes' => array('class' => array('layout-link-button', 'layout-access-add')),
-    '#submit' => array(
-      'layout_menu_item_form_condition_add'
+  $form['conditions']['links'] = array(
+    '#theme' => 'layout_action_links',
+    '#items' => array(
+      'add' => array(
```
quicksketch: I think by adding these elements as a property (`#items`), this will stop the recursive scanning of the form for child elements, and behavior-properties like `#ajax` and `#submit` won't behave properly.

jenlampton: Okay, I'll try with them as top-level children then.

quicksketch: So perhaps add the children like this instead:

```
$form['conditions']['links']['add'] = array(
```

Then in your theme function loop over the form children keys and render each one.

jenlampton: Yep, that fixed it!

core/modules/layout/layout.admin.inc
```diff
@@ -328,20 +328,21 @@ function layout_settings_form($form, &$form_state, Layout $layout) {
...
-    '#validate' => array(),
-    '#submit' => array(
-      'layout_settings_form_update_layout',
-      'layout_settings_form_context_add',
-    ),
-    '#ajax' => array(
-      'callback' => 'layout_ajax_form_open_dialog',
+  $form['context_wrapper']['context']['links'] = array(
+    '#theme' => 'layout_action_links',
+    'add' => array(
```
quicksketch: Everywhere we have this pattern, could we write it this way instead?

```
  $form['context_wrapper']['context']['links'] = array(
    '#theme' => 'layout_action_links',
  );
  $form['context_wrapper']['context']['links']['add'] => array(
```

That would line up with the way we normally deal with nested elements.

quicksketch (commented):
Looks like this PR might need to be rebased against 1.x still? Changes to the "Add relationship" button don't seem to be included, but that was the main thing we were wanting to fix.

core/modules/layout/layout.admin.inc
jenlampton: Please check the PR again (I just had the same thought, and made this update)

core/modules/layout/layout.theme.inc
```diff
@@ -228,19 +228,35 @@ function theme_layout_settings_context_table($variables) {
...
   return $output;
 }
 
+
+/**
+ * Output the list of visibility conditions for layouts.
+ *
+ * @deprecated since 1.19.0.
+ */
+function theme_layout_conditions($variables) {
```
quicksketch: If we're deprecating this theme function, shouldn't we stop calling it from places in core? But then it wouldn't be used anywhere... so perhaps just remove it?

Alternatively, we could leave this function as-is and add `theme_layout_list()` (or something like `theme_layout_action_links()`) and put a `@todo` in there to merge the two functions in 2.x?

jenlampton: Okay, I changed the name to `theme_layout_action_links()` and left it separate from conditions. You're right that conditions are special, with their operation links in the `<LI>`. I agree that they should probably not share the same function now, just because the function seems the same, we might want to replace those with something different in the future.

core/modules/layout/layout.theme.inc
```diff
@@ -227,6 +227,22 @@ function theme_layout_settings_context_table($variables) {
   ));
 }
 
+/**
+ * Output a list of items for layouts.
```
klonos:
```suggestion
 * Output a list of action links for layouts.
```

quicksketch:
I rebased and updated the PR to get tests passing. Since this moves the array location of the "add_relationship" button we needed to change how the submit handler finds it.

PR #3617, opened 2021-05-04
Issue 5077: add formatting to paths on layout listing page.

### LINKED ISSUE #5077: Added: Distinguish layout-created/module-provided/missing-callback paths on the layout listing page.
This is a follow-on to https://github.com/backdrop/backdrop-issues/issues/4986, which noted that for pages that have both a module-defined page and a layout with the same path, the page callback that actually gets called depends on the chronological order of the module-enabling and layout-creation that happened in the past. In that issue, I question whether we should have that past-history dependence. In this issue, I accept that dependence, but propose some changes that will make it easier for users to figure out what is going on.

Specifically, I propose to add some formatting to the system paths shown in the first column on the layout listing page at admin/structure/layouts:

1. If the page will be rendered by a module, use the current formatting (bold) for the path.
2. If the page will be rendered by a layout, use bold+italic.
3. If the page won't be rendered at all because there is no menu router item for the page, use bold+red.

Currently, there is no way to tell which of these three situations applies, short of attempting to go to the page and seeing what renders (or 404s, for case 3), which can be more puzzling than illuminating. If you apply the PR below, it's a lot easier to distinguish the different scenarios in https://github.com/backdrop/backdrop-issues/issues/4986. With regard to the 5 scenarios in https://github.com/backdrop/backdrop-issues/issues/4986, you should see:

* Scenario 1: bold (module provides the page)
* Scenario 2: bold+italic (layout provides the page, hiding module page)
* Scenario 3: bold+italic (layout provides the page)
* Scenario 4: bold+red (no menu router item for page, so 404)
* Scenario 5: bold+italic (layout provides the page)

I might also suggest that on the configure layout page, after the user has typed in a path, the AJAX call displays whether that page will be layout-provided or is an existing module-provided page, but that behavior isn't part of the PR below.

-- 

PR by @jenlampton, @bugfolder https://github.com/backdrop/backdrop/pull/3652 separates layouts into two tables
PR by @bugfolder https://github.com/backdrop/backdrop/pull/3864 labels each layout in a single table

bugfolder:
Fixes https://github.com/backdrop/backdrop-issues/issues/5077.

### DIFF
--- a/core/modules/layout/css/layout.admin.css
+++ b/core/modules/layout/css/layout.admin.css
@@ -5,6 +5,12 @@
 .layout-list .layout-group td:first-child {
   font-weight: bold;
 }
+.layout-list .layout-group td.layout-page {
+  font-style: italic;
+}
+.layout-list .layout-group td.missing-path a {
+  color: red;
+}
 .layout-list .layout-row td:first-child {
   padding-left: 24px;
 }
--- a/core/modules/layout/layout.admin.inc
+++ b/core/modules/layout/layout.admin.inc
@@ -38,6 +38,18 @@ function layout_list_page() {
     else {
       $path_link = strpos($path, '%') === FALSE ? l($path, $path) : check_plain($path);
       $row[] = array('data' => $path_link);
+      $page_callback = db_query('
+        SELECT page_callback
+        FROM {menu_router}
+        WHERE path = :path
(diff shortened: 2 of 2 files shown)


bugfolder:
Closing in favor of PR https://github.com/backdrop/backdrop/pull/3652.

PR #3618, opened 2021-05-06
Issue 5067: Move node preview layout code into node module via new hook

### LINKED ISSUE #5067: Move node preview layout code into node module via new hook in layout module

bugfolder:
Fixes https://github.com/backdrop/backdrop-issues/issues/5067.

Replaces https://github.com/backdrop/backdrop/pull/3608 with simpler approach (only one new hook).


core/modules/layout/layout.api.php
```diff
@@ -335,6 +336,98 @@ function hook_layout_presave(Layout $layout) {
   }
 }
 
+/**
+ * Perform alterations to the list of layouts that match the path of a router
+ * item.
+ *
+ * Modules can implement this hook to allow a given layout to be used for
+ *  multiple or different paths than its defined path.
```
quicksketch:
```suggestion
 * multiple or different paths than its defined path.
```

core/modules/layout/layout.api.php
```diff
@@ -335,6 +336,98 @@ function hook_layout_presave(Layout $layout) {
...
+ *   The list of layouts that should be considered for the path
+ *   $router_item['path']. Note that the final selection of layout will be based
+ *   on contexts and visibility conditions.
+ * @param array $router_item
+ *   The router item for the path. In addition to obtaining the system path at
+ *   $router_item['path'], you can use other elements of $router_item. For
+ *   example, $router_item['original_map'] contains an exploded version of the
+ *   normal path.
+ *
+ * @since 1.20.0 (?)
```
quicksketch:
```suggestion
 * @since 1.20.0
```

Let's go ahead and assume this will make it.

core/modules/layout/layout.module
```diff
@@ -979,15 +978,53 @@ function layout_load_multiple($layout_names, $skip_menu_items = FALSE) {
...
+ * desired and/or the router item is not available. Examples where this might be
+ * needed:
+ * - Inside of a hook_menu() implementation (see dashboard_menu()).
+ * - Inside of a hook_layout_layouts_alter() implementation (see
+ *   node_layout_layouts_alter()).
+ *
  * @param string $path
  *   The menu routing path, with all placeholders represented by "%" symbols.
+ * @param bool $skip_menu_items
+ *   Flag to skip the attempted loading of menu items for the loaded layouts.
```
quicksketch, line 1021: Thanks for documenting this :+1:

core/modules/layout/layout.module
```diff
@@ -1221,90 +1258,78 @@ function layout_provides_path($path) {
...
-                // object is not already loaded into $router_item. The built-in
-                // contexts don't do that, but a contrib module that creates a
-                // LayoutContext can have its load_callback called at this
-                // point.
-                $context_info = layout_get_context_info($context->plugin);
-                if (isset($context_info['load callback'])) {
-                  $context_data = call_user_func_array($context_info['load callback'], array($router_item['original_map'][$context->position]));
-                  $context->setData($context_data);
-                }
+  // Use the static cache, but keyed on the normal path (not system path) since
```
quicksketch: Perhaps provide examples here. It's not clear what "normal path" means.

> Use the static cache, but keyed on the href (such as node/1) rather than the system path (such as node/%) since client modules might serve up different layouts for the same system path.

core/modules/node/tests/node.test
```diff
@@ -810,6 +810,83 @@ class PagePreviewTestCase extends BackdropWebTestCase {
 
 }
 
+/**
+ * Tests that the correct layout is used for node previews.
+ */
+class NodeLayoutPreviewTestCase extends BackdropWebTestCase {
```
quicksketch, line 816: Nice, thanks for moving this :+1:

core/modules/layout/layout.module
```diff
@@ -979,15 +978,53 @@ function layout_load_multiple($layout_names, $skip_menu_items = FALSE) {
...
+ *   An array of load Layout object instances.
+ */
+function layout_load_multiple_by_router_item($router_item, $skip_menu_items = NULL) {
+  $href = $router_item['href'];
+  $layouts_by_item = &backdrop_static(__FUNCTION__, array());
+  if (isset($layouts_by_item[$href])) {
+    return $layouts_by_item[$href];
+  }
+  $layouts = layout_load_multiple_by_path($router_item['path'], $skip_menu_items);
+  backdrop_alter('layout_layouts', $layouts, $router_item);
```
quicksketch: After reading through this PR, I think this hook should be renamed. This hook does not apply when using `layout_load()`, `layout_load_multiple()`, or `layout_load_by_path()`, so I worry that it is going to be confusing based on its name implying that it would be called in those situations. The documentation you added helps, but the name itself is misleading.

Instead, I would suggest:
```suggestion
  backdrop_alter('layout_load_by_router_item', $layouts, $router_item);
```

Which would make the hook name `hook_layout_load_by_router_item_alter()`. 

While it is a long hook name, it's not unusual. We have other long hook names like `hook_field_formatter_settings_summary_alter()`, `hook_block_view_MODULE_NAME_DELTA_alter()`, and `hook_form_FORM_ID_alter()`.

This also makes this name similar to `hook_redirect_load_by_source_alter()`, which has the same sort of situation where it's only called in certain loading situations.

core/modules/layout/layout.api.php
```diff
@@ -335,6 +336,98 @@ function hook_layout_presave(Layout $layout) {
...
+ *
+ * @see layout_get_layout_by_path()
+ * @see node_layout_layouts_alter()
+ */
+function hook_layout_layouts_alter(&$layouts, $router_item) {
+  // Example taken from node_layout_layouts_alter(). For create node preview
```
quicksketch: Can you think of a simpler example that would be more useful as documentation? Developers could easily jump over to `node_layout_layouts_alter()` to view this same code if they wanted a full implementation example.

quicksketch (commented):
This looks great @bugfolder! Feedback below.

core/modules/layout/layout.module
bugfolder: Backdrop seems to call that a "normal path" in https://docs.backdropcms.org/api/backdrop/core%21includes%21path.inc/function/backdrop_get_normal_path/1.

quicksketch: Hm, yeah I see. "normal" really means "unaliased path". I think the example would still help even if we use "normal path". Something like:

> Use the static cache, but keyed on the normal path (such as node/1) rather than the system path (such as node/%) since client modules might serve up different layouts for the same system path.

klonos: Since we're tweaking, lets also add a comma before the "since" (too long sentences w/o any punctuation make them harder to read).

core/modules/node/node.module
```diff
@@ -1741,6 +1741,47 @@ function node_layout_context_info() {
   return $info;
 }
 
+/**
+ * Implements hook_layout_layouts_alter().
+ *
+ * For create node preview pages, the path will be of the form
```
klonos: "For create" ???

core/modules/layout/layout.api.php
klonos: ...also, "For create" ???

bugfolder: Part of the reason I used the code from `node_layout_load_by_router_item_alter()` as the example was that I could extensively comment it here in the API to explain what's going on. The `node_layout_load_by_router_item_alter()` is not extensively commented (in keeping with the general practice of keeping working code comments fairly terse). But since the code in API files is there explicitly for pedagogical purposes, this example illustrates several of the possible complexities and has a lot of explanation.

core/modules/layout/layout.api.php
bugfolder: Fixed in an updated PR.

core/modules/layout/layout.api.php
bugfolder: OK. Fixed in an updated PR.

core/modules/layout/layout.module
bugfolder: Examples (and comma — two, in fact, since the "rather" clause is an appositive) added in an update to the PR.

core/modules/layout/layout.module
bugfolder: Hook named change to your suggestion in update to the PR.

core/modules/layout/layout.api.php
bugfolder: "For create" is now fixed.

core/modules/node/node.module
bugfolder: Now fixed. "When creating".

quicksketch:
Merged into 1.x for 1.20. Thanks @bugfolder!

PR #3621, opened 2021-05-10
Issue #5085: [UX] Updates: Merge the manual and via-the-admin-UI tables

### LINKED ISSUE #5085: [UX] System updates form: Merge the manual updates section into the main updates table
The form under `admin/config/system/updates` seems too busy when the "Manual updates" option is disabled, having two sections/tables, and some help text in between:

image

How about something like this instead?:

issue-5085-merge-manual-updates-section

So basically:
- merge the manual updates section (only core really) with the main tableselect
- have the checkbox(es) for projects that require manual updates either be disabled/locked, or remove them completely
- indicate which projects require manual update via a suffix in their title
- add a more/less toggle link (same as those used in the modules listing and the status report page), which reveals manual update info

klonos:
Fixes https://github.com/backdrop/backdrop-issues/issues/5085


core/modules/installer/installer.manager.inc
```diff
@@ -298,13 +295,7 @@ function installer_manager_update_form($form, $form_state) {
...
  */
 function installer_manager_update_form_validate($form, &$form_state) {
-  if (!empty($form_state['values']['projects'])) {
-    $enabled = array_filter($form_state['values']['projects']);
-  }
-  if (!empty($form_state['values']['disabled_projects'])) {
-    $disabled = array_filter($form_state['values']['disabled_projects']);
-  }
-  if (empty($enabled) && empty($disabled)) {
+  if (empty(array_filter($form_state['values']['projects']))) {
```
indigoxela: Before php 5.5 empty() only takes *variables* as parameter. Try to move that array_filter() out of it.

klonos: So obvious 🤦🏼 ...thank you @indigoxela ❤️

core/modules/system/js/status.js
```diff
@@ -1,13 +1,11 @@
...
- * Filters the module list table by a text input search string.
- *
- * Additionally accounts for multiple tables being wrapped in "package" fieldset
- * elements.
+ * Toggles the more/less links that show/hide details in the "Status report"
+ * page.
  */
 Backdrop.behaviors.statusReport = {
-  attach: function(context, settings) {
+  attach: function() {
```
jenlampton, line 8: This changes the JS on the status report page -- is that intentional? The rest of this PR is for the system updates form, so I suspect these changes might be from another issue?

klonos: Yes, it is intentional. `context` and `settings` are never used anywhere in that function in status.js - they must have been introduced as a copy-paste from other code. 

...I actually copy-pasted this function into `installer.select_updates.js` since all it does is to provide the more/less toggle, and that's what I needed in the updates table. That's where I noticed that `context` and `settings` were not in use in either place. We could leave that out, but does this tiny thing really need a separate issue and PR? 🤷🏼‍♂️ (in principle, I understand that the answer to that is "yes", but in this case, really?)

klonos: PS: you will notice that the docblock in the `status.js` file mentions the modules list table (which I also fix in this PR), so there was definitely some copy-paste involved when this was introduced.

klonos: PPS: once this PR gets merged, we'll have a total of 3 places where we are using custom JS to provide this more/less functionality, and all that JS is ~90% similar. I've filed https://github.com/backdrop/backdrop-issues/issues/5090 to consider standardizing that UI pattern, and use `<details>` instead.

izmeez:
The PR has expired. Can it be refreshed to test?

klonos:
@izmeez done. Please test when you have time/energy/capacity to do so, and thanks 🙏🏼

izmeez:
I like this idea and the PR is working in that I can launch the tugboat instance and sign in. However, since it is using the development branch on the update report there is no update needed and the manual update required with drop down more text is not displayed. I will have to apply a patch to to test against an older version. If I press on within tugboat to install system updates, it doesn't tell me I can't instead it goes ahead and then hits an error:

bd-tugbot-pr-available-updates-install-system-updates-on-1-27-x-error

Plus, it invites me to continue and we all know this is going to lead nowhere, so why are we going down this path. I thought before with PR on the dev branch I thought I had seen a message that updates were disabled and had to be done manually. I'll have to sort that out in my mind.

PR #3624, opened 2021-05-16
Issue #2902430: [PHP 7.1] A non-numeric value encountered in theme_pager()

quicksketch:
Part of Drupal 7 cross-ports issue https://github.com/backdrop/backdrop-issues/issues/5042

Replaces https://github.com/backdrop/backdrop/pull/3599

---

See:
- http://drupal.org/node/2902430
- https://cgit.drupalcode.org/drupal/commit/?id=63473379cda3b1d0b08bff61017d13376ae19342

It turns out that we were missing the entire `pager.test` file in Backdrop (how come?), so I've added it in this PR.

One key difference with the D7 code is that I've moved the `assertClass()` and `assertNoClass()` assertions to `backdrop_web_test_case.php` instead, so that all tests can start using them, and benefit.

Questions:
- Is moving those assertions a good thing as I thought, or not and should be put back in `pager.test`?
- Should we call these assertions `assertCssClass` and `assertNoCssClass` instead, or keep as is for D7 backwards compatibility?
- Should we also add `assertCssId` and `assertNoCssId` assertions?

### DIFF
--- a/core/includes/pager.inc
+++ b/core/includes/pager.inc
@@ -159,7 +159,7 @@ function theme_pager($variables) {
   $tags = $variables['tags'];
   $element = $variables['element'];
   $parameters = $variables['parameters'];
-  $quantity = (int) $variables['quantity'];
+  $quantity = empty($variables['quantity']) ? 0 : (int) $variables['quantity'];
   global $pager_page_array, $pager_total;
 
   // Return if there is no pager to be rendered.
--- a/core/modules/simpletest/backdrop_web_test_case.php
+++ b/core/modules/simpletest/backdrop_web_test_case.php
@@ -484,6 +484,40 @@ abstract class BackdropTestCase {
   }
 
   /**
+   * Asserts that an element has a given class.
+   *
+   * @param SimpleXMLElement $element
+   *   The element to test.
+   * @param string $class
+   *   The class to assert.
+   * @param string $message
+   *   (optional) A verbose message to output.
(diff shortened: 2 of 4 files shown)


quicksketch:
Answering these questions from the original PR by @klonos:

> Is moving those assertions a good thing as I thought, or not and should be put back in pager.test?

The original PR put these assertions in `BackdropTestCase`, they should at least be in sub-class `BackdropWebTestCase`, since `BackdropTestCase` might be used on non-HTML based tests. However, looking at other `BackdropWebTestCase` assertion methods, they usually have a different parameter list. Most of them check the entire current page for an assertion, rather than taking a first parameter for `SimpleXMLElement $element`.

This inconsistency with other `assert*` methods makes me think we should instead not elevate these methods into the `BackdropWebTestCase` and leave them in `PagerFunctionalWebTestCase` (like D7) instead.

> Should we call these assertions assertCssClass and assertNoCssClass instead, or keep as is for D7 backwards compatibility?
> Should we also add assertCssId and assertNoCssId assertions?

These two questions are pretty moot if we end up not adding the methods to `BackdropWebTestCase`, since the methods would only apply to the pager tests.

I'll push up my PR with the changes that leave these methods in the pager.test file to get this to completion and taking the simplest approach of simply matching the D7 code.

quicksketch:
Merged into 1.x and 1.18.x.

PR #3625, opened 2021-05-16
Issue #2472025 by stupiddingo, stefanos.petrakis: [D7] Hide toolbar w…

### LINKED ISSUE #5048: Hide admin bar when printing
This is the respective of https://www.drupal.org/node/2472025, which is part of #5042

Do we need the same in Backdrop? If so, then this is the code change: https://git.drupalcode.org/project/drupal/commit/88de1737a6dac893d5f2e480dbf5b17263705628

jenlampton:
Fixes https://github.com/backdrop/backdrop-issues/issues/5048

### DIFF
--- a/core/modules/admin_bar/css/admin_bar-print.css
+++ b/core/modules/admin_bar/css/admin_bar-print.css
@@ -4,3 +4,6 @@
 #admin-bar {
   display: none;
 }
+.admin-bar body {
+  border-top: none !important;
+}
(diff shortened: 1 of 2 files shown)


core/modules/admin_bar/css/admin_bar-print.css
```diff
@@ -0,0 +1,6 @@
+/**
+ * Hide the admin bar when printing.
+ */
+#admin-bar {
+  display: none;
+}
```
quicksketch, line 6: We should also remove the border added to the `body` tag if the admin bar is present. When printing I get a big border at the top of the page where the admin bar would normally be present:

image

```suggestion
}
.admin-bar body {
  border-top: none !important;
}

```

quicksketch:

PR #3629, opened 2021-05-20
Issue #5090: UX/DX: Replace the custom more/less toggles with details

### LINKED ISSUE #5090: [UX] [DX] Consider replacing the more/less toggle links in the admin UI with "details" HTML elements
We currently use these more/less toggles in:

- the modules listing page at `admin/modules` (to show/hide dependency info):
  modules-more-less-toggle
- the site status report (to show/hide additional info for errors/warnings):
  status-page-more-less-toggle
- the "System updates" page, when there are manual updates (to be introduced with #5085):
  issue-5085-merge-manual-updates-section

These currently are implemented as 90% similar, yet separate pieces of JS, spread across the files of the respective modules providing these admin pages. We should consider using `<details>` for these, plus some `theme_more_less_toggle()` function perhaps. That would make the code reusable, consistent, and allow it to be updated in a single place. We'd also inherit any usability/accessibility improvements of the standard HTML element.

For older browsers, we'd need to use a polyfill. Drupal 8+ core currently uses 2 polyfills for this:
- [collapse.js](https://git.drupalcode.org/project/drupal/-/blob/9.2.x/core/misc/collapse.js) is a polyfill for the open/close behaviour from HTML5.1+ recommendation.
- [details-aria.js](https://git.drupalcode.org/project/drupal/-/blob/9.2.x/core/misc/details-aria.js) is a polyfill for the HTML-AAM recommendation.

klonos:
Fixes https://github.com/backdrop/backdrop-issues/issues/5090


core/includes/form.inc
```diff
@@ -4000,6 +4039,45 @@ function form_process_fieldset(&$element, &$form_state) {
...
+  // Contains form element summary functionalities.
+  $element['#attached']['library'][] = array('system', 'backdrop.form');
+
+  // The .form-wrapper class is required for #states to treat DETAILS elements
+  // like containers.
+  if (!isset($element['#attributes']['class'])) {
+    $element['#attributes']['class'] = array();
+  }
+
+  // DETAILS elements do not need 'collapsible'/'collapsed' classes, since they
```
jenlampton: Here's one that is MIT license: https://www.npmjs.com/package/details-polyfill

core/modules/system/css/system.css
```diff
@@ -371,3 +371,24 @@ section,
...
+  color: #0074bd;
+  cursor: pointer;
+  pointer-events: auto;
+}
+details[open] summary::after {
+  content: ' ' attr(data-close);
+}
+details:not([open]) summary::after {
+  content: ' ' attr(data-open);
+}
```
quicksketch: This styling seems really specific to the specific use we want on the modules page, but we may want to use `<details>` elements in other places (such as in Form API) in the future. Can we make these selectors specific to the modules use-case and only provide generic details styling directly on the `<details>` element itself.

core/includes/form.inc
klonos: Thanks @jenlampton 👍🏼  ...added with [`d8e51d6` (#3629)](https://github.com/backdrop/backdrop/pull/3629/commits/d8e51d6b2b82ce2310677de01766a3536c9ea449)

core/modules/system/css/system.css
klonos: All these selectors are `details.more-less-toggle` now. It won't only be used in the modules listing, but also in the status report page, the placeholder examples in the layout configuration pages, the consolidated updates page, and anywhere else we need a similar UI.

Does that address your concerns @quicksketch ?

core/themes/seven/css/style.css
```diff
@@ -799,7 +799,7 @@ details {
   background-color: #ffffff;
   display: table-cell;
 }
-details summary span {
+details:not(.more-less-toggle) summary span {
```
klonos: I anticipate people asking why this change here. If we don't exclude the "special" more/less details elements from this, then we have to specifically override this font color/size each time we use more/less `<details>` in:
- help text (which needs to be light gray)
- the modules listing (which needs to be black)
- the status page (needs to be black)
- ...any other place where blue text for the details summary is not required

core/modules/system/css/system.css
```diff
@@ -370,3 +370,86 @@ section,
...
+  font-size: 100%;
+  transform: none;
+}
+[dir="rtl"] details.more-less-toggle[open] summary .more-less-toggle-label::before {
+  transform: rotate(-90deg);
+}
+
+details.more-less-toggle .details-content-wrapper {
+  margin-top: 0.5em;
+}
```
bugfolder, line 487:
```suggestion
}

/* Don't show the outline for details elements that have custom more/less labels. */
details summary.more-less-toggle {
  outline: none;
}
```

core/includes/form.inc
```diff
@@ -3716,8 +3724,17 @@ function theme_details($variables) {
...
+  // Custom labels for the open/close toggle.
+  $more_less_toggle = '';
+  if (isset($element['#open_label']) && isset($element['#close_label'])) {
+    $more_less_toggle = '<span class="more-less-toggle-label" data-open-label="' . $element['#open_label'] . '" data-close-label="' . $element['#close_label'] . '"></span>';
+    $element['#attributes']['class'][] = 'more-less-toggle';
+  }
+
   $output = '<details' . backdrop_attributes($element['#attributes']) . '>';
-  $output .= '<summary><span>' . $element['#summary'] . '</span></summary>';
+  $output .= '<summary>' . $summary_content . $more_less_toggle . '</summary>';
```
bugfolder:
```suggestion
  if (empty($more_less_toggle)) {
    $output .= '<summary>';
  }
  else {
    $output .= '<summary class="more-less-toggle">';
  }
  $output .= $summary_content . $more_less_toggle . '</summary>';
```

bugfolder: Just realized that the appending of `$more_less_toggle` and closing tag might as well go inside the `if` and `else`.

klonos: I initially though to do something like this instead:
```suggestion
  $summary_class = !empty($more_less_toggle) ? ' class="more-less-toggle"' : '';
  $output .= '<summary' . $summary_class . '>' . $summary_content . $more_less_toggle . '</summary>';
```

I then thought that it would in general be useful to be able to specify attributes for the `<summary>` element besides the `<details>` element (which is done via the `#attributes` property), so I've introduced a `'#summary_attributes'` property for that, and set the `more-less-toggle` class that way.

klonos: ... @bugfolder since the `<summary>` element is always nested within the `<details>` element, and the `more-less-toggle` class is already being added to the `<details>` element, we can target the summary in CSS via `details.more-less-toggle summary` (or just `.more-less-toggle summary`). Do you still think that we really need to be repeating the same class in both elements?

core/themes/seven/css/style.css
klonos: I've added an inline comment to explain that. Resolving this comment here.

core/includes/form.inc
bugfolder: No, as long as it's targetable, that's all we need.

core/modules/system/css/system.css
klonos: @bugfolder I'm not sure about hiding the outline. It seems to be an accessibility feature, especially when using keyboard navigation, in order to indicate the currently focused element.

klonos: @bugfolder I ended up fixing this with some `tabindex` properties set. See [`18c4d90` (#3629)](https://github.com/backdrop/backdrop/pull/3629/commits/18c4d90547a3322ed3d6fb1221dcd32e8e76c1ab)

Things seem to only work properly in Chrome though. See my recent comment in the issue.

klonos:
...random failure on php8 only. Kicking things again.

klonos:
Hmm 🤔 ...php8 keeps failing with this:
```
Status    Group      Filename          Line    Function                            
------------------------------------------------------------------------------------------------------------------------
Exception Error      dashboard.module  14                                                                              
    Call to undefined function layout_load_multiple_by_path()
Fail      Other      upgrade.test      187     performUpgrade()                                                        
    "Updates were attempted" found
Fail      Other      upgrade_filled.te 23      testFilledUpgrade()                                                     
    The upgrade was completed successfully.
Error: Process completed with exit code 1.
```

Kicking things again ...

klonos:
That's annoying ...either a random failure, or Tugboat sandbox failing 😞 ...trying again.

klonos:
...different random failure in php5 this time 😞

klonos:
FINALLY!!! 🎉 🎉 😌 ...I'm not touching this PR/branch again! 😅  ...not until https://github.com/backdrop/backdrop/pull/4544 gets merged at least, at which point I'll rebase in order to get all tests green.

klonos:
...base Tugboat image was rebuilt, which destroyed sandboxes. Kicking things again 🤞🏼

core/includes/stream_wrappers.inc
```diff
@@ -147,8 +147,8 @@ interface BackdropStreamWrapperInterface extends StreamWrapperInterface {
    *
    * This function should return a URL that can be embedded in a web page
    * and accessed from a browser. For example, the external URL of
-   * "youtube://xIpLd0WQKCY" might be
-   * "http://www.youtube.com/watch?v=xIpLd0WQKCY".
+   * "youtube://123aBc456xYz" might be
+   * "http://www.youtube.com/watch?v=123aBc456xYz".
```
jenlampton, line 151: this video doesn't exist, but I prefer that to what's there now :)

core/modules/system/css/system.admin.css
```diff
@@ -221,11 +199,13 @@ span.admin-missing {
   display: block;
   font-weight: normal;
 }
-.system-status-report .status-description {
+.system-status-report .status-summary .details-content-wrapper {
```
jenlampton, line 202: Can we leave the `status-description` class? I wouldn't want themes that targeted that class specifically to stop working. We can remove our CSS but let's not change the markup.

core/modules/system/system.module
```diff
@@ -1448,6 +1448,20 @@ function system_library_info() {
...
+  // details-polyfill.
+  $libraries['details-polyfill'] = array(
+    'title' => 'Polyfill for the HTML5 `<details>` element',
+    'website' => 'https://github.com/rstacruz/details-polyfill',
+    'version' => '1.2.0',
+    'js' => array(
+      'core/misc/html5-details.js' => array(
+        'group' => JS_LIBRARY,
+        'weight' => -22,
+        'browsers' => array('IE' => 'lte IE 8', '!IE' => FALSE),
```
jenlampton: Do we even need this? I think we dropped support for IE.

core/modules/user/user.theme.inc
```diff
@@ -177,26 +177,46 @@ function theme_user_admin_permissions($variables) {
...
   if (!empty($permission_item['description'])) {
     $description[] = '<div class="permission-description description table-filter-text-source">' . $permission_item['description'] . '</div>';
   }
-  if (!empty($permission_item['restrict access'])) {
-    $description[] = '<em class="permission-warning">' . t('Warning: Give to trusted roles only; this permission has security implications.') . '</em>';
-  }
-  if (!empty($permission_item['warning'])) {
-    $description[] = '<a class="warning-toggle" href="#">more</a>';
-    $description[] = '<div class="permission-warning-description">' . $permission_item['warning'] . '</div>';
+  if ($permission_item['restrict access']) {
```
jenlampton, line 197: changing the logic here makes me nervous. If the result is the same let's not mess with it.

core/modules/simpletest/tests/theme.test
```diff
@@ -225,7 +225,7 @@ class ThemeTableUnitTest extends BackdropWebTestCase {
   /**
    * Tests that the 'footer' option works correctly.
    */
-  function testThemeTableFooter() {
+  public function testThemeTableFooter() {
```
jenlampton, line 228: This change needs to happen in a separate commit (can be a task)

core/includes/form.inc
```diff
@@ -3756,10 +3794,33 @@ function theme_details($variables) {
     $element['#attributes']['open'] = 'open';
   }
 
+  $summary_content = '<span>' . $element['#summary'] . '</span>';
+
+  // Custom labels for the open/close toggle.
+  $more_less_toggle = '';
+  if (isset($element['#open_label']) && isset($element['#close_label'])) {
```
jenlampton: I don't think the labels should be used to determine if the details has an open/close toggle.

core/includes/stream_wrappers.inc
klonos: This was tripping up cspell, so I wanted to make it happy but without having to add `xIpLd0WQKCY` to the list of known words in the dictionary. We could wrap this line or just the URL in a `cspell:disable`/`cspell:enable` pair of keywords, but then those keywords would end up in the function description in the API site. Or we could wrap the entire docblock in the cspell keywords, which then has the risk of introduction of typos for the entire docblock in the future.

core/modules/system/system.module
klonos: Removed with [`fb39c4e` (#3629)](https://github.com/backdrop/backdrop/pull/3629/commits/fb39c4ee918d30de6082dc10d57eeb1aee917842)

core/includes/form.inc
klonos: Fixed with [`f9b1a26` (#3629)](https://github.com/backdrop/backdrop/pull/3629/commits/f9b1a26b9dbc37ac2ae17b50ffdfd375cffb4771)

core/modules/user/user.theme.inc
klonos: OK, context: this is a before/after of the inconsistency I am trying to fix:
image

So with the changes that went in with https://github.com/backdrop/backdrop-issues/issues/5536, we standardized the warning that is being added to permissions that have either certain or potential security implications. As you can see in the "before", the warnings for the various "Use the `Filter name` text format" permissions are added as a `description` help text, unlike all other permissions that have the equivalent warnings consistently styled (and may or may not have a `description` help text in addition to and independently from that). In other words, we are building the contents of the permission column for all other permissions (those that adhere to the standardization we introduced) by adding the following components:
1. a permission label (all permissions have this)
2. a permission description/help (optional)
3. a permission warning (only if there are security implications)
4. a description for the reason of the warning (optional - hidden in a more/less toggle)

The "Use the `Filter name` text format" permissions though are breaking that consistency/standardization, and adding the warning as help text instead.

So if you take a look at the change in the `core/modules/filter/filter.module` file in this PR, I am doing the following two things for these "rogue" permissions:
1. adding a `'restrict access' => TRUE` (to trigger the consistent warning)
2. changing what was previously a `'description' =>` text to a `'warning' =>` text instead

The above ^^ is simply something that we missed in the PR for https://github.com/backdrop/backdrop-issues/issues/5536. We implemented the change for the `Administer text formats and filters` permission, but we failed to do the same for the various filter permissions (~20 lines below the code we changed in that issue/PR).

Now, if you look in `core/modules/user/user.admin.inc` in the `user_admin_permissions()` function we are doing the following **_for all permissions_**:
```php
      foreach ($permissions as $perm => $perm_item) {
        // Fill in default values for the permission.
        $perm_item += array(
          'description' => '',
          'restrict access' => FALSE, <--- this
          'warning' => '',
        );
```
So `$permission_item['restrict access']` will **_never_** be empty here. It will always default to `FALSE`, unless it has explicitly been set to `TRUE` for a permission. Either way, never empty - either `TRUE` or `FALSE`.

I hope that this makes more sense now. Let me know if you are still concerned/nervous about this change.

core/includes/form.inc
```diff
@@ -3725,27 +3725,70 @@ function form_process_details($element, &$form_state) {
...
+ *     - #open_label: Customize the default "more" label of the toggle. Only
+ *       applicable if the #toggle property has been set to TRUE.
+ *     - #close_label: Customize the default "less" label of the toggle. Only
+ *       applicable if the #toggle property has been set to TRUE. Setting this
+ *       to NULL will use the same label for both open/closed states.
  *
  * @return string
  *   The rendered HTML output.
  *
+ * @since 1.26.2:
```
quicksketch, line 3768: Let's try to target 1.27.0
```suggestion
 * @since 1.27.0:
```

PR #3631, opened 2021-05-22
#5102: Image autocomplete

laryn:
Fixes #5102


klonos (approved):
LGTM 👍🏼

core/modules/filter/filter.pages.inc
```diff
@@ -202,6 +189,30 @@ function filter_format_editor_image_form($form, &$form_state, $format) {
...
+      $view .= ' ' . t('Check the <a href="!url">Views administration page</a> to enable or create the "@view_name" view as needed.', array('!url' => url('admin/structure/views'), '@view_name' => $view_name));
+    }
+  }
+  $rendered_library = '<div class="library-view" data-editor-library-view="true">' . $view . '</div>';
+
+  return $rendered_library;
+}
+
+/**
+ * Render the image library.
```
quicksketch, line 210:
```suggestion
 * Menu delivery callback to print the image library page without any wrappers.
```

core/modules/filter/js/filter.js
```diff
@@ -209,8 +208,9 @@ Backdrop.behaviors.editorImageDialog = {
 
           // Display the library view.
           $('.editor-image-fields').removeClass('editor-image-fields-full');
-          $('.editor-image-library').css({ 'display': 'block' });
-          $('.form-item-image-directory').css({ 'display': 'block' });
+          $('.editor-image-library').load(Backdrop.settings.basePath + 'filter_image_library_ajax');
+          $('.editor-image-library').addClass('shown');
+          $('.form-item-image-directory').addClass('shown');
```
quicksketch, line 213: Not sure about this `shown` class instead of toggling the `display` property. There's some chance of class name conflicts.

core/modules/filter/js/filter.js
```diff
@@ -209,8 +208,9 @@ Backdrop.behaviors.editorImageDialog = {
 
           // Display the library view.
           $('.editor-image-fields').removeClass('editor-image-fields-full');
-          $('.editor-image-library').css({ 'display': 'block' });
-          $('.form-item-image-directory').css({ 'display': 'block' });
+          $('.editor-image-library').load(Backdrop.settings.basePath + 'filter_image_library_ajax');
```
quicksketch, line 211: I think this might not work with clean URLs disabled. You should perhaps use this instead:

```suggestion
          $('.editor-image-library').load(Backdrop.settings.basePath + '?q=filter_image_library_ajax');
```

I couldn't find a way to convert a menu path to a full URL on the JS side. Sometimes we'll get around this by passing the URL as a setting or as a data attribute. For example you might set a data attribute `data-editor-image-library-url` and then pull it off with `$('.editor-image-library').data('editor-image-library-url')`.

core/modules/filter/js/filter.js
```diff
@@ -238,12 +238,18 @@ Backdrop.behaviors.editorImageDialog = {
               var $submit = $form.find('.form-actions input[type=submit]:first');
               $submit.trigger('mousedown').trigger('click').trigger('mouseup');
             });
+          // Use autocomplete field as exposed filter for image library.
+          $('#edit-attributes-src').keyup(function () {
```
quicksketch, line 242: It would be good to use `Backdrop.debounce` on this to prevent race conditions. Right now an ajax request is fired for every keyup, which might be unnecessary if you're a fast typer.

quicksketch (commented):
This works great @laryn! Feedback below.

PR #3633, opened 2021-05-23
Issue #5098: Indicate new dashboard notifications

### LINKED ISSUE #5098: Highlight "New" Dashboard Notifications
**Description of the need**

In my opinion, Dashboard notifications are a pretty powerful and potentially useful tool for communicating with the community. However, it's so very easy to overlook them,especially given that they don't change very often. 

Could we do something to draw attention to NEW notifications when they first arrive?

**Proposed solution**

If a user views the dashboard and there are any new notifications. A "NEW" label in a different color is visible to draw their attention to this new notification. I think we have something like this for new content, but I don't remember which issue or how to trigger it. 

Ideally, the "New" label would disappear once a user have viewed the dashboard notification.

This is another good idea: https://github.com/backdrop/backdrop-issues/issues/5097
The above issue triggered me to post this idea as an alternative suggestion, but I'm generally in favor of either of them - or if there should be some combination. 

**Additional information**

This issue was inspired by this problem: https://github.com/backdrop/backdrop-issues/issues/5094
We are looking for ways to make sure that users get this notification about the need to upgrade their Backup and Migrate module and some way to get a user attention while they are in the process of updating or visiting their site would be good. 

Backdrop now includes...

klonos:
Fixes https://github.com/backdrop/backdrop-issues/issues/5098


core/modules/dashboard/includes/block.news.inc
```diff
@@ -40,13 +40,30 @@ class DashboardNewsBlock extends Block {
    */
   public function getContent() {
     $news = $this->getNews();
+    $days_before_news_is_old = 14;
```
jenlampton: hm, I'm not sure I like this variable name. How about `$new_period` or `$new_duration`?

core/modules/dashboard/includes/block.news.inc
```diff
@@ -40,13 +40,30 @@ class DashboardNewsBlock extends Block {
...
       foreach ($news as $key => $notifications) {
-        $links[] = '<p>' . l(
-            $notifications['title'],
-            $notifications['url']
-          ) . ' - ' . filter_xss($notifications['summary']) . '</p>';
+        $notification_date = format_date($notifications['timestamp'], 'custom', 'M j, Y');
+        $date_classes = array('class' => array('notification-date'));
+        $is_new = (time() - $notifications['timestamp']) < $days_before_news_is_old * 24 * 60 * 60;
+        if ($is_new) {
+          $date_classes['class'][] = 'marker';
```
jenlampton, line 53: I don't think we should use the `marker` class here. That has a specific meaning in other places, and is maybe not specific enough for this use. How about `recent` or even `notification-recent` ?

core/modules/dashboard/includes/block.news.inc
```diff
@@ -40,13 +40,30 @@ class DashboardNewsBlock extends Block {
...
+        $notification_date = '<span ' . backdrop_attributes($date_classes) . '>' . $notification_date . '</span>';
+        $notification_link = l($notifications['title'], $notifications['url'], array('attributes' => array('target'=>'_blank')));
+        $notification_summary = filter_xss($notifications['summary']);
+
+        $notification = array(
+          'data' => '<p>' . $notification_date . ': ' . $notification_link . ' - ' . $notification_summary . '</p>',
+        );
+
+        if ($is_new) {
+          $notification['class'] = array('is-new');
```
jenlampton, line 64: ah! Maybe this is all we need?

core/modules/dashboard/includes/block.news.inc
klonos: @jenlampton I'm happy to make this change, but we should then add an inline comment to explain what "new period" or "new duration" is, since both of these names you are suggestion seem too vague to me, and it won't be immediately apparent to everyone reading the code ("was there an old period? ...and what is it used for?").

If we are to add inline comments, and since that variable is not being used anywhere else, I'd be happy to remove it completely, and then change the `$days_before_news_is_old * 24 * 60 * 60` thing a few lines below to a hard-coded `1209600` and add a comment there instead...

core/modules/dashboard/includes/block.news.inc
```diff
@@ -40,13 +40,30 @@ class DashboardNewsBlock extends Block {
...
     $links = array();
     if ($news) {
       foreach ($news as $key => $notifications) {
-        $links[] = '<p>' . l(
-            $notifications['title'],
-            $notifications['url']
-          ) . ' - ' . filter_xss($notifications['summary']) . '</p>';
+        $notification_date = format_date($notifications['timestamp'], 'custom', 'M j, Y');
+        $date_classes = array('class' => array('notification-date'));
+        $is_new = (time() - $notifications['timestamp']) < $days_before_news_is_old * 24 * 60 * 60;
```
klonos: @jenlampton @stpaultim how about this?:

```suggestion
        // A notification is considered new, if no more that 14 days (1209600
        // seconds) have lapsed.
        $is_new = (time() - $notifications['timestamp']) < 1209600;
```

core/modules/dashboard/includes/block.news.inc
jenlampton: > If we are to add inline comments, and since that variable is not being used anywhere else, I'd be happy to remove it completely, and then change the $days_before_news_is_old * 24 * 60 * 60 thing a few lines below to a hard-coded 1209600 and add a comment there instead...

I like this idea :)

core/modules/dashboard/includes/block.news.inc
jenlampton: I love it!

stpaultim: Looks good

olafgrabienski: Suggestion for better readability:

```
        // A notification is considered new, if no more that 14 days
        // (1209600 seconds) have lapsed.
```

klonos: Done: [`f57e8a0` (#3633)](https://github.com/backdrop/backdrop/pull/3633/commits/f57e8a0be11a952070cb749eac46a8d30e621d6f)

@olafgrabienski please see our [coding standards re inline comments](https://docs.backdropcms.org/doc-standards#general):
> - Lines containing comments (including docblocks) must wrap **as close to 80 characters as possible** without going over...

core/modules/dashboard/includes/block.news.inc
klonos: Done: [`f57e8a0` (#3633)](https://github.com/backdrop/backdrop/pull/3633/commits/f57e8a0be11a952070cb749eac46a8d30e621d6f)

core/modules/dashboard/includes/block.news.inc
olafgrabienski: Fair enough! In my opinion it's good to interpret the standard formally less strict in such a case, for the sake of readability. But that discussion should probably happen in another place.

quicksketch:

PR #3644, opened 2021-06-14
Update comment.module

### LINKED ISSUE #4429: Clarify validation message when posting comments with existing username
**Description of the need**

If an anonymous user attempts to post a comment putting under the _Your name_ field his name, which to his/her "luck" coincides with a registered username, then all s/he gets is the _The name you used belongs to a registered user._ message, because of the following [condition check](https://github.com/backdrop/backdrop/blob/497c43b0b7a99b63ca1db81683b9ff59aa99ffac/core/modules/comment/comment.module#L2089):

```
elseif ($form_state['values']['is_anonymous']) {
    // Validate anonymous comment author fields (if given). If the (original)
    // author of this comment was an anonymous user, verify that no registered
    // user with this name exists.
    if ($form_state['values']['name']) {
      $query = db_select('users', 'u');
      $query->addField('u', 'uid', 'uid');
      $taken = $query
        ->condition('name', db_like($form_state['values']['name']), 'LIKE')
        ->countQuery()
        ->execute()
        ->fetchField();
      if ($taken) {
        form_set_error('name', t('The name you used belongs to a registered user.'));
      }
    }
  }
```

Git blame shows the check was introduced to resolve the problem for _Cannot override comment author to or from Anonymous_ reported on https://www.drupal.org/node/936844. Good for the problem was addressed, but at the same time the solution introduced a new problem: now users with popular names (e.g. Sarah, Tom, John, Alan, etc.) putting their real names under the _Your name_ field are punished, because the filed is validated against existing usernames in the database. 

To me it's a regression, because the UI does not explicitly explain one real Tom why he can not use his own real name. However, even if properly reflected (in unlikely scenario the  _Your name_ should become  _Your username_) it is not welcoming at all as it makes new users feel like we said _Sorry Tom, we have another Tom, therefore you can not be Tom_.

**Proposed solution**

Ideally, names and usernames must be explicitly separated and differed. And not only for comment forms, but everywhere in the system. While usernames must stay unique for each user, multiple occurrences of the same name should be allowed. 

I realize it's a bold proposal about probably massive code change in the core. Therefore, in case if this will be met by rather conservative take, I have alternative just a minor change proposal: at least, the message should be changed to something like:

> The name you used is a registered username. If it belongs to you then login to post. If not then use another name.

Please polish the text to make it more appealing to a confused user as otherwise the original text  _The name you used belongs to a registered user._ is not fool-proof as it does not suggest to take any further steps for those users who do not realize their names had already been taken and who just need to use another name.

shruti1421:
Fixes :  [#4429](https://github.com/backdrop/backdrop-issues/issues/4429) 
Changed the message inside comment.module as suggested by @alanmels .

### DIFF
--- a/core/modules/comment/comment.module
+++ b/core/modules/comment/comment.module
@@ -2126,7 +2126,7 @@ function comment_form_validate($form, &$form_state) {
         ->execute()
         ->fetchField();
       if ($taken) {
-        form_set_error('name', t('The name you used belongs to a registered user.'));
+        form_set_error('name', t('The name you used is a registered username. If it belongs to you then login to post. If not then use another name.'));
       }
     }
   }
--- a/core/modules/comment/comment.module
+++ b/core/modules/comment/comment.module
@@ -2126,7 +2126,7 @@ function comment_form_validate($form, &$form_state) {
         ->execute()
         ->fetchField();
       if ($taken) {
-        form_set_error('name', t('The name you used is a registered username. If it belongs to you then login to post. If not then use another name.'));
+        form_set_error('name', t('The name you used is a registered username. If it belongs to you, then login to post. If not, then use another name.'));
       }
     }
   }
--- a/core/modules/comment/tests/comment.test
+++ b/core/modules/comment/tests/comment.test
@@ -1187,7 +1187,7 @@ class CommentAnonymous extends CommentHelperCase {
     $this->setCommentSubject(TRUE);
 
     $this->backdropPost('comment/reply/' . $this->node->nid, $edit, t('Save'));
-    $this->assertText(t('The name you used belongs to a registered user.'));
+    $this->assertText(t('The name you used is a registered username. If it belongs to you, then login to post. If not, then use another name.'));
 
     // Require contact info.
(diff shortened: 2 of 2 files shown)


core/modules/comment/comment.module
```diff
@@ -2126,7 +2126,7 @@ function comment_form_validate($form, &$form_state) {
         ->execute()
         ->fetchField();
       if ($taken) {
-        form_set_error('name', t('The name you used belongs to a registered user.'));
+        form_set_error('name', t('The name you used is a registered username. If it belongs to you then login to post. If not then use another name.'));
```
klonos: Thanks for working on this @shruti1421 🙏🏼 ...please add commas to the if clauses, like so:

```suggestion
        form_set_error('name', t('The name you used is a registered username. If it belongs to you, then login to post. If not, then use another name.'));
```

See https://www.enago.com/academy/ifthen-using-conditional-sentences-in-academic-writing

> Use a comma to separate the if clause from the main clause when the if clause comes first.

quicksketch:
Merged into 1.x and 1.19.x. Thanks @shruti1421!

PR #3646, opened 2021-06-17
Issue #5094: Provide a pass-through of the PDO prepare method.

### LINKED ISSUE #5094: Potential regression for contrib (backup_migrate) caused by initial php 8 support
**Description of the bug**

Several users reported a problem with backup_migrate after updating to Backdrop 1.19.

https://github.com/backdrop-contrib/backup_migrate/issues/63

`Error : Cannot access protected property DatabaseConnection_mysql::$pdo`

`Error : Call to undefined method DatabaseConnection_mysql::exec()`

**Steps to reproduce:**
* Install backdrop 1.19.0
* Install and enable backup and migrate 1.x-1.0.16
* Attempt to create a backup
* Attempt to restore from this backup

---

I think, the first thing people need is instructions on how to fix this on their sites:

To avoid the problem, update Backup Migrate to 1.x-1.0.17 *before* you update core to 1.19.0

OR

If you're already in trouble, comment out one line in Backup Migrate as described [in this issue comment](https://github.com/backdrop-contrib/backup_migrate/issues/63#issuecomment-842397222) (then update the module)

**Additional information**

- Backdrop CMS version: 1.19.0
- Backup Migrate versions before 1.x-1.0.17

Probably caused by https://github.com/backdrop/backdrop/commit/4102cd21190e75e35e4274472565691f7e4fb6a9#diff-a30038705d970e43b622562028439144c59afc2abc9fb53048aff64286fbcfc3

quicksketch:
Fixes https://github.com/backdrop/backdrop-issues/issues/5094.

This provides a wrapper around the PDO::prepare() method so that it is available for Backup and Migrate. It does not restore complete compatibility with the previous behavior of Backdrop prior to 1.19.0, but it fixes the specific problem with Backup and Migrate module in https://github.com/backdrop/backdrop-issues/issues/5094

### DIFF
--- a/core/includes/database/database.inc
+++ b/core/includes/database/database.inc
@@ -480,6 +480,25 @@ abstract class DatabaseConnection {
   }
 
   /**
+   * Provides a pass-through to the PDO prepare method.
+   *
+   * In adding support for PHP 8, the Backdrop DatabaseConnection no longer
+   * directly extends the PHP PDO object directly. A side-effect is that calls
+   * directly to the PDO object no longer worked. This provides a pass-through
+   * to this method specifically for compatibility with previous versions of
+   * Backdrop. See https://github.com/backdrop/backdrop-issues/issues/5094.
+   *
+   * @param $query
+   *
+   * @return bool|PDOStatement
+   *
+   * @since 1.19.2
+   */
+  public function prepare($query) {
+    return $this->pdo->prepare($query);
+  }
+
+  /**
    * Prepares a query string and returns the prepared statement.
    *
    * This method caches prepared statements, reusing them when
--- a/core/includes/database/database.inc
+++ b/core/includes/database/database.inc
@@ -484,7 +484,7 @@ abstract class DatabaseConnection {
    *
    * In adding support for PHP 8, the Backdrop DatabaseConnection no longer
    * directly extends the PHP PDO object. A side-effect is that calls
-   * directly to the PDO object no longer worked. This provides a pass-through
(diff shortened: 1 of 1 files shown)


core/includes/database/database.inc
```diff
@@ -479,6 +479,25 @@ abstract class DatabaseConnection {
     }
   }
 
+  /**
+   * Provides a pass-through to the PDO prepare method.
+   *
+   * In adding support for PHP 8, the Backdrop DatabaseConnection no longer
+   * directly extends the PHP PDO object directly. A side-effect is that calls
```
indigoxela: > directly extends the PHP PDO object directly

Once would be enough ("directly"). :smile:

quicksketch:
```suggestion
   * directly extends the PHP PDO object. A side-effect is that calls
```

core/includes/database/database.inc
```diff
@@ -479,6 +479,25 @@ abstract class DatabaseConnection {
     }
   }
 
+  /**
+   * Provides a pass-through to the PDO prepare method.
+   *
+   * In adding support for PHP 8, the Backdrop DatabaseConnection no longer
+   * directly extends the PHP PDO object. A side-effect is that calls
+   * directly to the PDO object no longer worked. This provides a pass-through
```
quicksketch:
```suggestion
   * to PDO methods no longer worked. This provides a pass-through
```

quicksketch:
Merged into 1.x and 1.19.x.

PR #3647, opened 2021-06-18
Issue #5118: Dropbutton text on admin/people, also fixes translation …

### LINKED ISSUE #5118: Change `admin/people` dropbutton operation from "Cancel" to "Cancel account"
**Description of the bug**

The dropbutton operation for cancelling a user account on the `admin/people` page is called "Cancel". While the single term "Cancel" in this context means "Disable or delete", it means "Stop or abort an action" in most (I guess all) other places of the Backdrop interface.

This isn't a (big) problem in an English interface: readers are usually able to get the meaning of the term in context. But in German (and maybe other languages) you have to decide between two translations of "Cancel", depending on the context. The translation "Abbrechen" is fine for most places but not correct in the user account dropbutton.

**Actual behavior**

It's not possible to translate the term "Cancel" correctly for all places in Backdrop.

**Expected behavior**

Being able to translate the term "Cancel" without issues.

indigoxela:
…issue

Fixes https://github.com/backdrop/backdrop-issues/issues/5118

### DIFF
--- a/core/modules/user/views/views_handler_field_user_link_cancel.inc
+++ b/core/modules/user/views/views_handler_field_user_link_cancel.inc
@@ -22,7 +22,7 @@ class views_handler_field_user_link_cancel extends views_handler_field_user_link
     if ($uid && user_cancel_access($account)) {
       $this->options['alter']['make_link'] = TRUE;
 
-      $text = !empty($this->options['text']) ? $this->options['text'] : t('Cancel');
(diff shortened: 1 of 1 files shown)


indigoxela:
Request fresh sandbox (Server error in deploy)...

klonos (approved):
LGTM 👍🏼

quicksketch:

PR #3650, opened 2021-06-19
Issue 4947: Update Backdrop with better greek transliteration

### LINKED ISSUE #4947: Update Backdrop with better greek transliteration as drupal already did
**Description of the bug**
Hi, I'm greek and I can tell you with certainty that the transliteration used in backdrop for greek letters in url is not what greeks expect. I absolutely agree with the changes that took place in Drupal 7,8 and 9 core regarding the transliteration. The following links show the changes that need to take place and I agree with them:
https://www.drupal.org/project/transliteration/issues/1215872
https://www.drupal.org/project/drupal/issues/2926187

I also have installed Drupal 9 recently and the above transliterations are already in core! We should update Backdrop to include them too!
Actually I don't understand why Backdrop didn't include these changes already.

From the first link the patch works also for backdrop for the file /core/includes/transliteration/x03.php

cptX:
Fixes https://github.com/backdrop/backdrop-issues/issues/4947


cptX:
Hi, is this pull request merged to the core yet? What we need to do next?

PR #3651, opened 2021-06-20
Fixed document for adding keys with fields

### LINKED ISSUE #3759: Documentation for adding keys along with fields is confusing.
The **$keys_new** parameter for `db_add_field()` is not explained well in the docblock in `core/includes/database/database.inc` or [our API](https://api.backdropcms.org/api/backdrop/core%21includes%21database%21database.inc/function/db_add_field/1)

I spent forever trying to wrap my head around what this meant:

> Optional keys and indexes specification to be created on the table along with adding the field. The format is the same as a table specification, but without the 'fields' element. If you are adding a type 'serial' field, you MUST specify at least one key or index including it in this array. See db_change_field() for more explanation why.

I recommend we reword this as follows:

> Optional. An array containing the specification for keys and indexes to be created on the table along with the field. The format is the same as a schema definition, but without the 'columns' element.  If you are adding a 'serial' type field, you MUST specify at least one key or index in this array. See db_change_field().

Of note: 
* we do not have a `fields` element, we do ave a `columns` element.
* `table specification` is ambiguous, `schema definition` has a concrete meaning.

nzare:
Fixes https://github.com/backdrop/backdrop-issues/issues/3759

### DIFF
--- a/core/includes/database/database.inc
+++ b/core/includes/database/database.inc
@@ -2927,10 +2927,10 @@ function db_drop_table($table) {
  *   creating NOT NULL columns with no default value in existing tables.
  * @param $keys_new
  *   Optional. An array containing the specification for keys and indexes to be 
- *   created on the table along with the field. The format is the same as a schema 
- *   definition, but without the 'columns' element. If you are adding a 'serial' 
- *   type field, you MUST specify at least one key or index in this array. See
- *   db_change_field() for more explanation why.
+ *   created on the table along with the field. The format is the same as a
+ *   schema definition, but without the 'columns' element. If you are adding a
+ *   'serial' type field, you MUST specify at least one key or index in this
+ *   array. See db_change_field() for more explanation why.
  *
  * @see db_change_field()
  */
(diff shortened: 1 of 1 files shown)


core/includes/database/database.inc
```diff
@@ -2926,10 +2926,10 @@ function db_drop_table($table) {
...
  * @param $keys_new
- *   Optional keys and indexes specification to be created on the table along
- *   with adding the field. The format is the same as a table specification, but
- *   without the 'fields' element. If you are adding a type 'serial' field, you
- *   MUST specify at least one key or index including it in this array. See
+ *   Optional. An array containing the specification for keys and indexes to be 
+ *   created on the table along with the field. The format is the same as a schema 
+ *   definition, but without the 'columns' element. If you are adding a 'serial' 
+ *   type field, you MUST specify at least one key or index in this array. See
  *   db_change_field() for more explanation why.
```
ghost: We should make sure this adheres to the 80-character limit...

```suggestion
 *   Optional. An array containing the specification for keys and indexes to be 
 *   created on the table along with the field. The format is the same as a
 *   schema definition, but without the 'columns' element. If you are adding a
 *   'serial' type field, you MUST specify at least one key or index in this
 *   array. See db_change_field() for more explanation why.
```

ghost (changes requested):

ghost (approved):

quicksketch:
Thanks for the PR @nzare! I have merged this into 1.x and 1.19.x.

PR #3652, opened 2021-06-26
Issue #5077: Add formatting to paths on layout listing page (alt)

### LINKED ISSUE #5077: Added: Distinguish layout-created/module-provided/missing-callback paths on the layout listing page.

bugfolder:
Fixes https://github.com/backdrop/backdrop-issues/issues/5077.

An alternate approach from PR https://github.com/backdrop/backdrop/pull/3617 to adding formatting to distinguish module-provided from layout-provided paths, also paths whose menu router is missing.

### DIFF
--- a/core/modules/layout/css/layout.admin.css
+++ b/core/modules/layout/css/layout.admin.css
@@ -5,6 +5,13 @@
 .layout-list .layout-group td:first-child {
   font-weight: bold;
 }
+.layout-list .layout-group td.missing-path,
+.layout-list .layout-group td.missing-path a {
+  color: red;
+}
+.layout-list .layout-group td .path-parenthetical {
+  font-weight: normal;
+}
 .layout-list .layout-row td:first-child {
   padding-left: 24px;
 }
--- a/core/modules/layout/css/layout.admin.css
+++ b/core/modules/layout/css/layout.admin.css
@@ -7,7 +7,7 @@
 }
 .layout-list .layout-group td.missing-path,
 .layout-list .layout-group td.missing-path a {
-  color: #ff1100;
+  color: #c40d00;
 }
 .layout-list .layout-group td .path-parenthetical {
   font-weight: normal;
--- a/core/modules/layout/css/layout.admin.css
+++ b/core/modules/layout/css/layout.admin.css
@@ -7,7 +7,7 @@
 }
 .layout-list .layout-group td.missing-path,
 .layout-list .layout-group td.missing-path a {
-  color: red;
+  color: #ff1100;
 }
 .layout-list .layout-group td .path-parenthetical {
   font-weight: normal;
--- a/core/modules/layout/layout.admin.inc
+++ b/core/modules/layout/layout.admin.inc
@@ -29,7 +29,7 @@ function layout_list_page() {
     $row = array();
     // Default layouts have a null path.
     if ($path === '') {
-      $row[] = array('data' => t('Default layouts') . ' ' . '<span class="path-parenthetical">' . t('(used on all paths without a specific layout)') . '</span>');
+      $row[] = array('data' => t('Default layouts') . ' <span class="path-parenthetical">(' . t('used on all paths without a specific layout') . ')</span>');
       // Empty columns are intentional so that they may be hidden with
       // priority-low class on the $header array.
       $row[] = '';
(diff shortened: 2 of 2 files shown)


core/modules/layout/layout.admin.inc
```diff
@@ -29,15 +29,34 @@ function layout_list_page() {
...
+      $cell = array();
+      $page_callback = db_query('
+        SELECT page_callback
+        FROM {menu_router}
+        WHERE path = :path
+        ', array(':path' => $path))
+        ->fetchField();
+      if (!$page_callback) {
+        $cell['class'] = 'missing-path';
+        $path_link .= ' ' . '<span class="path-parenthetical">' .t('(no callback)'). '</span>';
```
indigoxela: Have you considered to put the parenthesis outside the translatable string? They are important and always should be there.

Another thought on the css class: How about "callback-missing" or "missing-callback" - IMO parenthetical is a bit misleading here. On the other hand, the parent item class "missing-path" could already be sufficient, to use `.missing-path span` as CSS selector.

@bugfolder what do you think?

core/modules/layout/css/layout.admin.css
```diff
@@ -5,6 +5,13 @@
 .layout-list .layout-group td:first-child {
   font-weight: bold;
 }
+.layout-list .layout-group td.missing-path,
+.layout-list .layout-group td.missing-path a {
+  color: red;
```
indigoxela: To my knowledge it's required to use the format `#ff0000` (lowercase hex codes) to conform to the coding standards. But using color is a bit tricky here, anyway, because of contrast.

core/modules/layout/layout.admin.inc
bugfolder: Thanks for pointing out putting parens outside. I'll make that change.

On the class, my rationale was to use the common class "path-parenthetical" for all parentheticals (which is what allows the parentheticals to be formatted as not-bold); so that includes this one, the (existing) parenthetical on Default Layouts, and possibly any future parenthetical that might be added for reasons we haven't thought of yet. 

And I thought that using the class, rather than the span element, as the CSS selector would then allow the possibility for future changes that might introduce other spans inside of that td (maybe some future tooltip-like thing).

core/modules/layout/css/layout.admin.css
bugfolder: Thanks, I'll make that change. Maybe try a slightly brighter red to improve the contrast.

bugfolder:
Closing in favor of PR https://github.com/backdrop/backdrop/pull/3732.

bugfolder:
Closing in favor of https://github.com/backdrop/backdrop/pull/3864.

PR #3659, opened 2021-07-09
Optimize menu_rebuild()

### LINKED ISSUE #5389: [PS] Optimize menu_navigation_links_rebuild()
## Description of the need

The function menu_navigation_links_rebuild() uses a large number of database queries when rebuilding the system menu links that are created based on hook_menu. This process uses nearly 300 database queries without any contributed modules, and gets more expensive the more modules you have.

## Proposed solution

These queries can be statically cached during the process and then all written back to the database at the end. Based on my initial testing, this reduces test execution time by 10-15%

## Additional information

I had to modify one test that deletes menu items without using menu_link_delete to simulate a database crash. The behavior in this case is changed because the links are stored using a static cache.

PR: https://github.com/backdrop/backdrop/pull/3659

Advocate: @hosef

hosef:
Reducing the number of database queries used when a menu link is saved or when the menus are rebuilt.

Fixes backdrop/backdrop-issues#5389


indigoxela:
@hosef which issue does this PR refer to?

hosef:
@indigoxela there isn't actually an issue. I initially wasn't even sure this would work. I just made the PR to see how the test bot would handle it.

docwilmot:
This PR says it optimizes `menu_rebuild()` but the issue says `menu_navigation_links_rebuild()`.

hosef:
@docwilmot
When I initially created the PR, I was trying to make `menu_rebuild()` faster, but I didn't create an issue because I wasn't sure it would work or if anything would come of it. After finally getting it working, it turned out that most of my changes where in `menu_navigation_links_rebuild()`, so that is the function name I put in the issue title.

jenlampton:
@hosef great work, I'm impressed you were able to remove so many database queries (and surprised there were so many to remove!) 

I have one question, why were these tests removed?
```php
   // Start over, forcefully delete child-1 from the database, simulating a
    // database crash. Check that the children of child-1 have been reassigned
    // to the parent, going up on the old path hierarchy stored in each of the
    // links.
```

hosef:
@jenlampton Those tests were checking the behavior of the menu re-parenting when a different page uses a manual database query to remove a menu link in the middle of your page load.
I don't think this should be supported behavior because we have provided menu_delete as a supported method of removing a menu link. We also don't support that in any other context, such as checking that a node wasn't deleted in the middle of generating a view.

jenlampton:
@hosef Does this change cause those tests to fail? 

If not, I'd prefer that we remove those tests in a separate issue where we can discuss why it was supported in the first place, just in case there's something else we're missing. Including that decision in an optimization issue might be setting us up for unforeseen consequences and it makes me nervous :)

hosef:
Those tests can't pass with the new code because they bypass the built in menu API functions, so they don't update the static cache used internally in menu_rebuild().

jenlampton:
Ah ok! Let's see if other committers have any concerns. If not it's
probably worth the risk :)

core/includes/menu.inc
```diff
@@ -3041,54 +3044,68 @@ function _menu_navigation_links_rebuild($menu) {
...
+      && $menu_link->customized == 0
+      && !in_array($menu_link->router_path, $default_menu_link_paths)) {
+      _menu_delete_item($menu_link, TRUE);
+    }
   }
 }
 
+/**
+ * Stores a copy of menu_links in memory for when lots of functions need to
+ * query from the table.
```
quicksketch: Elaborating the docblock here a little bit and fitting within 80 chars for the first line:


```suggestion
 * Stores a copy of menu_links in memory for menu load and save functions.
 *
 * Rather than querying the database for the menu links, menu operations are
 * expected to utilize this cache. Modification of links directly in the
 * database may cause this cache to become inconsistent, in which case it
 * should be rebuilt via menu_rebuild() if needed within the same request.
 *
 * This single cache is utilized by the following functions:
 *   - _menu_navigation_links_rebuild()
 *   - _menu_link_find_parent()
 *   - menu_link_children_relative_depth()
 *   - menu_link_save()
 *
 * Though this cache may be large (several megabytes for large menues), the
 * cache is left intact for the entire request in the event that multiple
 * menu links are modified in a single request, such as when updating the
 * weights or parents of several links in a single form.
 *
 * @return array
 *   An array of all rows within the menu_links table. Passed by reference.
```

quicksketch (approved):

quicksketch:
I've merged this into 1.x for inclusion in 1.21.0. Great work @hosef!

PR #3662, opened 2021-07-10
Update README.md

Ayushdubey86:
This will help new user to easily understand, as I faced too many problem understanding it

### DIFF
--- a/README.md
+++ b/README.md
@@ -48,9 +48,7 @@ contact the security team before filing an issue.
 
 Developers
 ----------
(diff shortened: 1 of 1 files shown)


indigoxela:
@Ayushdubey86 which issue in the core issue queue does this relate to? I can't find any link in the description.

To explain:
- The first thing is always an issue in our [issue queue](https://github.com/backdrop/backdrop-issues/issues).
- Every pull request is supposed to link to the issue in its description like `Fixes https://...full..link...here`

Otherwise it's hard to connect PR and Issue with each other.

PR #3668, opened 2021-07-15
Fix colour consistency for the buttons

### LINKED ISSUE #5130: [UX] Button colors inconsistent when Basis used as admin theme
**Description of the bug**

We should have four distinct button colors:
- primary (blue for basis)
- secondary (white for basis)
- default (gray for basis)
- dangerous (red for basis)

Every time there is a button in the admin theme, it should be gray, unless it also has a class for primary or secondary, in which case it should take on the specific color.

**Steps To Reproduce**

To reproduce the behavior:
1. Switch your site to use Basis for admin theme
2. Visit the content listing
3. Edit an Article

**Actual behavior**

Many buttons are blue, when they should be white or gray.

Screen Shot 2021-07-15 at 12 35 19 PM

**Expected behavior**

Only the "Save" button should be blue.

thivyaa-web-dev:
Fixes https://github.com/backdrop/backdrop-issues/issues/5130

### DIFF
--- a/core/themes/basis/css/skin.css
+++ b/core/themes/basis/css/skin.css
@@ -534,7 +534,7 @@ input.form-submit:active {
 }
 
 .button,
-input.form-submit{
+input.form-submit {
   color: #444b53;
   background: #e5e5e6;
 }
--- a/core/themes/basis/css/skin.css
+++ b/core/themes/basis/css/skin.css
@@ -534,7 +534,10 @@ input.form-submit:active {
 }
 
 .button,
-input.form-submit,
+input.form-submit{
+  color: #444b53;
+  background: #e5e5e6;
+}
 input.button-primary,
 button.button-primary,
 .ui-state-default,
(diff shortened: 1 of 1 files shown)


core/themes/basis/css/skin.css
```diff
@@ -534,7 +534,10 @@ input.form-submit:active {
 }
 
 .button,
-input.form-submit,
+input.form-submit{
```
jenlampton: we need to include a space in front of the `{`

himanshu007-creator: @thivyaa-web-dev  will make the changes

thivyaa-web-dev: @jenlampton  I made the changes as suggested by you.

laryn:
Thank you @thivyaa-web-dev -- much appreciated! I've merged this into 1.x and 1.19.x.

PR #3670, opened 2021-07-15
Issue #5021: use announce for block page

### LINKED ISSUE #5021: [A11Y] Use Backdrop.announce() when using Block drag and drop rearrangement
**Description of the need**

In https://github.com/backdrop/backdrop-issues/issues/4985 we added the new Backdrop.announce() method, but we are not yet using it. It was added to make the drag and drop of blocks accessible via screenreader, now that it is accessible via keyboard (see https://github.com/backdrop/backdrop-issues/issues/4941).

**Proposed solution**

Backdrop.announce() should be used to describe the new position of a block and what region it is in after a block has changed position (whether by mouse or keyboard).

As an accessibility improvement, I this can qualify for any bugfix release.

docwilmot:
Fixes https://github.com/backdrop/backdrop-issues/issues/5021

### DIFF
--- a/core/modules/layout/plugins/renderers/layout_renderer_editor.inc
+++ b/core/modules/layout/plugins/renderers/layout_renderer_editor.inc
@@ -138,8 +138,8 @@ class LayoutRendererEditor extends LayoutRendererStandard {
   function getAriaLabel($block) {
     $content = $block->getAdminPreview();
 
-    $label = t('Draggable block');
-    $label .= $block->getAdminTitle();
+    $label = t('Draggable block ');
+    $label .= strip_tags($block->getAdminTitle());
 
     return $label;
   }
--- a/core/modules/layout/plugins/renderers/layout_renderer_editor.inc
+++ b/core/modules/layout/plugins/renderers/layout_renderer_editor.inc
@@ -103,17 +103,18 @@ class LayoutRendererEditor extends LayoutRendererStandard {
    * {@inheritdoc}
    */
   function renderBlock($block) {
+    $aria_label = $this->getAriaLabel($block);
     $buttons = $this->getBlockLinks($block);
     $content = $block->getAdminPreview();
 
-    $output = '<div tabindex="0" class="layout-editor-block clearfix" id="layout-editor-block-' . $block->uuid . '" data-block-id="' . $block->uuid . '">';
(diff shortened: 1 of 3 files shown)


docwilmot:
Why fail?

docwilmot:
The sandbox didnt delete itself? Why not?

core/modules/layout/plugins/renderers/layout_renderer_editor.inc
```diff
@@ -103,17 +103,18 @@ class LayoutRendererEditor extends LayoutRendererStandard {
    * {@inheritdoc}
    */
   function renderBlock($block) {
+    $aria_label = $this->getAriaLabel($block);
     $buttons = $this->getBlockLinks($block);
     $content = $block->getAdminPreview();
 
-    $output = '<div tabindex="0" class="layout-editor-block clearfix" id="layout-editor-block-' . $block->uuid . '" data-block-id="' . $block->uuid . '">';
+    $output = '<div tabindex="0" aria-label="' . $aria_label . '" class="layout-editor-block clearfix" id="layout-editor-block-' . $block->uuid . '" data-block-id="' . $block->uuid . '">';
```
indigoxela: @docwilmot I wonder if it might make sense (for readability) to create an array of attributes and run it through backdrop_attributes() then.
```
$attributes = array(
  'tabindex' => '0',
  'class' => array(
    ...
  ),
  ...
);

$output = '<div ' . backdrop_attributes($attributes) . '>';
```
That's really only for readability - I find that looong line hard to read. :slightly_smiling_face: 
That function would also do sanitization (with check_plain, though).

docwilmot: > create an array of attributes

This makes sense, I agree.
>That function would also do sanitization (with check_plain, though).

I thought since `getAriaLabel()` runs everything through `strip_tags()`, and  `$block->getAdminTitle()` already runs its output through `check_plain()` we'd be OK with that?

indigoxela: @docwilmot whatever you choose, it's only a suggestion. :smile:

docwilmot: Ahh I misunderstood. You mean `backdrop_attributes()` does sanitization. I thought you were saying that `getAriaLabel()` needs sanitization. Cool, even better.

docwilmot: Done

quicksketch:

PR #3671, opened 2021-07-18
Issue #5131: Changed views wizard button text

### LINKED ISSUE #5131: [UX] Fixed: Views wizard button text says 'edit' and not 'configure'
**Description of the bug**

From @klonos: In the Views UI we have this for the add view form:

create_view_form_buttons

I don't like the "Save & exit" - particularly the "exit" thing does not sit well with me. How about we change these buttons to "Save and edit" and "Save view" instead? The default does not need to be changed I think since the save and edit seems to be the most common action.

This was split-off from https://github.com/backdrop/backdrop-issues/issues/999 so we could separate that issue into two.

**Actual behavior**

Button text is currently "Continue and Configure" and "Save & exit"

**Expected behavior**

Button text should read "Continue and Configure" (TBD?) and "Save view"

**Steps To Reproduce**

To reproduce the behavior:
1. Go to `admin/structure/views/add`

Aryamanz29:
-  UX : Changed button text ( save view, continue and configure ).

Closes backdrop/backdrop-issues#5131

### DIFF
--- a/core/modules/views/tests/views_ui.test
+++ b/core/modules/views/tests/views_ui.test
@@ -35,7 +35,7 @@ class ViewsUIWizardBasicTestCase extends ViewsUIWizardHelper {
     $view1['name'] = strtolower($this->randomName(16));
     $view1['description'] = $this->randomName(16);
     $view1['page[create]'] = FALSE;
-    $this->backdropPost('admin/structure/views/add', $view1, t('Save & exit'));
+    $this->backdropPost('admin/structure/views/add', $view1, t('Save view'));
     $this->assertText(t('Your view was saved. You may configure it from the list below.'));
     $this->assertText($view1['human_name']);
     $this->assertText($view1['description']);
@@ -62,7 +62,7 @@ class ViewsUIWizardBasicTestCase extends ViewsUIWizardHelper {
     $view2['page[path]'] = $this->randomName(16);
     $view2['page[feed]'] = 1;
     $view2['page[feed_properties][path]'] = $this->randomName(16);
-    $this->backdropPost('admin/structure/views/add', $view2, t('Save & exit'));
+    $this->backdropPost('admin/structure/views/add', $view2, t('Save view'));
 
     // Since the view has a page, we expect to be automatically redirected to
     // it.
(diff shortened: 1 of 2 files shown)


Aryamanz29:
@BWPanda @indigoxela  Need your review

jenlampton:
This PR looks great @Aryamanz29, but one thing we should do is change all our buttons to Sentence case (where only the first letter of the button is capitalized). It's hard to tell we do this from looking at the buttons since they appear to be ALL CAPS but we do that transform with CSS. (The only transform we can't do with CSS is to make them all Sentence case, so that's why we do it that way in the original text.)

Aryamanz29:
> This PR looks great @Aryamanz29, but one thing we should do is change all our buttons to Sentence case (where only the first letter of the button is capitalized). It's hard to tell we do this from looking at the buttons since they appear to be ALL CAPS but we do that transform with CSS. (The only transform we can't do with CSS is to make them all Sentence case, so that's why we do it that way in the original text.)

Actually, there is no direct way to do `Sentence case` with CSS but one thing we can do that -
```
p {
    text-transform: lowercase;
}

p:first-letter {
    text-transform: uppercase;
}
``` 
So, should I try this method ? @jenlampton 
ref : https://stackoverflow.com/questions/4322314/convert-uppercase-letter-to-lowercase-and-first-uppercase-in-sentence-using-css

ghost:
@Aryamanz29 I think @jenlampton meant remove the capital letters from your PR...

Aryamanz29:
> @Aryamanz29 I think @jenlampton meant remove the capital letters from your PR...

Sorry my bad 😅

Aryamanz29:
> @Aryamanz29 I think @jenlampton meant to remove the capital letters from your PR...

Done :heavy_check_mark: @BWPanda  @jenlampton

jenlampton:
Yes, perfect! Thank you both :) RTBC!

quicksketch:
Thanks @Aryamanz29! I merged this change into 1.x and 1.19.x. Thanks @jenlampton and @BWPanda for reviewing!

PR #3674, opened 2021-07-29
Issue #5138: Fix icon repetition on admin/config page.

### LINKED ISSUE #5138: Fixed repeating icons on main Configuration page
**Description of the bug**

The section icons on the *Configuration* page (`admin/config`) are displayed in a strange way. The most striking issue: each icon is displayed repeatedly.

**Steps To Reproduce**

To reproduce the behavior:
1. Install a fresh Backdrop site, version 1.19.2, or update your site to 1.19.2.
2. Go to the main *Configuration* page (`admin/config`).
3. Have a look at the sections, especially the icons.

**Actual behavior**

Each icon is displayed repeatedly, the headings are hidden by the icons, and they look generally not correct.

**Expected behavior**

Icons are displayed once, headings displaying correctly.

**Screenshots**

Backdrop 1.19.2:
backdrop-admin-config-issue-icons

Backdrop 1.19.1 / expected behavior:
backdrop-admin-config-ok

**Additional information**

The problem was probably introduced by #5083, see https://github.com/backdrop/backdrop-issues/issues/5038#issuecomment-886196936.

jenlampton:
Fixes https://github.com/backdrop/backdrop-issues/issues/5138

### DIFF
--- a/core/modules/system/system.theme.inc
+++ b/core/modules/system/system.theme.inc
@@ -462,7 +462,7 @@ function theme_admin_block($variables) {
 
   $output .= '<div class="' . implode(' ', $classes) . '">';
   if (!empty($block['title'])) {
-    $output .= '<h3>' . $block['title'] . '</h3>';
+    $output .= '<h3 class="has-icon">' . $block['title'] . '</h3>';
   }
   if (!empty($block['content'])) {
     $output .= '<div class="body">' . $block['content'] . '</div>';
--- a/core/themes/seven/css/style.css
+++ b/core/themes/seven/css/style.css
@@ -1189,7 +1189,8 @@ div.filter-options select {
 .admin-panel .block-content {
   padding: 0 9px 9px;
 }
-.admin-panel .block-title {
+.admin-panel .block-title,
+.admin-panel h3.has-icon {
   margin: 9px 9px 0;
   padding: 6px 0 6px 35px; /* LTR */
   font-size: 0.923em;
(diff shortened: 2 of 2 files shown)


core/themes/seven/css/style.css
```diff
@@ -1189,7 +1189,8 @@ div.filter-options select {
 .admin-panel .block-content {
   padding: 0 9px 9px;
 }
-.admin-panel .block-title {
+.admin-panel .block-title,
+.admin-panel h3.has-icon {
```
indigoxela, line 1195: Hint: the `[dir="rtl"]` also needs that selector to also fix rtl display.

core/modules/system/system.theme.inc
```diff
@@ -462,7 +462,7 @@ function theme_admin_block($variables) {
 
   $output .= '<div class="' . implode(' ', $classes) . '">';
   if (!empty($block['title'])) {
-    $output .= '<h3>' . $block['title'] . '</h3>';
+    $output .= '<h3 class="has-icon">' . $block['title'] . '</h3>';
```
quicksketch, line 465: My understanding of front-end practices is that we should be giving this a class that can be used alone without the tag name, such as `admin-panel-heading` or `admin-panel-title`. Then when we have selectors, instead of `.admin-panel h3.has-icon`, we would use the single class `.admin-panel-title`. That makes it so that you have a low level of specificity for themes to override.

We already had 2 levels of specificity with `.admin-panel h3`, changing it to `.admin-panel h3.has icon` is now 3 levels.

jenlampton: I was trying to keep it as close to 1.19.1 as possible, I'm worried that by changing the level of level of specificity in addition to adding a class is going to cause a lot of problems with admin themes.  

I like this approach for 2.x, but since we are already trying to fix a regression, I think we should change as little as possible.

I'm going to close this PR in favor of @olafgrabienski's, which keeps things closer to how they were in 1.19.1 before the visual break.

jenlampton:
Closing in favor of https://github.com/backdrop/backdrop/pull/3675

PR #3676, opened 2021-07-30
Issue #5141: FieldException on content type creation

### LINKED ISSUE #5141: FieldException on content type creation if the Body field is still getting purged
I found an old D7 bug that also exists in Backdrop. Steps to reproduce (in strict order):

Install backdrop
delete both Articles and Posts content types
create a new custom content type, hit save
The following error occurs when saving any/all new content types:

```
FieldException: Attempted to create an instance of field body, but that field does not exist.
in field_create_instance() (line 515 of /home/mysite/public_html/core/modules/field/field.crud.inc).
```

This is a common series of events that results in a damaged site. It also chimes into the discussion over "Config defaults" and why less is more in terms of default settings. In this case the Body field default setting actually breaks the site. I feel a simple fix would be to not add a Body field by default and not many people will miss it.

indigoxela:
Fixes https://github.com/backdrop/backdrop-issues/issues/5141

### DIFF
--- a/core/modules/node/node.types.inc
+++ b/core/modules/node/node.types.inc
@@ -93,7 +93,7 @@ function node_type_form($form, &$form_state, $type = NULL) {
   // Create the body field by default on form submission. This may be altered by
   // other modules.
   $body_config = config_get('field.field.body');
-  if (isset($body_config) && $body_config['deleted'] == TRUE) {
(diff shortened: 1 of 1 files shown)


indigoxela:
Closing in favor of https://github.com/backdrop/backdrop/pull/3681

PR #3678, opened 2021-08-01
Issue 5142: Add missing & to calls to backdrop_static().

### LINKED ISSUE #5142: '&' missing from calls to backdrop_static()
The design pattern for using [`backdrop_static()`](https://docs.backdropcms.org/api/backdrop/core%21includes%21bootstrap.inc/function/backdrop_static/1)` calls for a reference assignment, like this:
```
function example_list($field = 'default') {
  $examples = &backdrop_static(__FUNCTION__);
  ...
```

However, there are several places in core where the '&' is missing; easily found by a search for the text `= backdrop_static(`.

Omission of the '&' means that subsequent calls won't get the static variable created inside the function, thereby losing the value of this construction. To see an example of the difference, run this code somewhere (with Devel module enabled):

```
$foo1 = backdrop_static('FOO1', array());
$foo1['foo1'] = 'FOO1';
$foo1a = backdrop_static('FOO1', array());
dpm($foo1a,'$foo1a');

$foo2 = &backdrop_static('FOO2', array());
$foo2['foo2'] = 'FOO2';
$foo2a = &backdrop_static('FOO2', array());
dpm($foo2a,'$foo2a');
```

In the first case (incorrect), `$foo1a` will be a new empty array. In the second (correct), `$foo2a` is a reference to the same array as `$foo2`, as intended.

Solution is straightforward: just add an `&` in all the places that they're missing (I count 7 occurrences).

Effect should be some speedups since the values weren't previously getting the benefit of static caching.

See https://github.com/backdrop/backdrop/pull/3673/#discussion_r680400810.

bugfolder:
Fixes https://github.com/backdrop/backdrop-issues/issues/5142.

Add missing & to several calls to `backdrop_static()`. This doesn't include a change needed in `path.module`, which should be addressed by the PR connected to https://github.com/backdrop/backdrop-issues/issues/5137.

### DIFF
--- a/core/modules/field/field.info.inc
+++ b/core/modules/field/field.info.inc
@@ -670,7 +670,7 @@ function field_info_bundles($entity_type = NULL) {
  */
 function field_info_field_map() {
   // Read from the "static" cache.
-  $field_info_cache = backdrop_static(__FUNCTION__, array());
+  $field_info_cache = &backdrop_static(__FUNCTION__, array());
 
   if (!empty($field_info_cache)) {
     return $field_info_cache;
--- a/core/modules/file/file.module
+++ b/core/modules/file/file.module
@@ -2083,7 +2083,7 @@ function file_get_download_token($file) {
  * @see file_file_predelete()
  */
 function file_get_file_references(File $file, $field = NULL, $age = FIELD_LOAD_REVISION, $field_type = 'file', $check_access = TRUE) {
-  $references = backdrop_static(__FUNCTION__, array());
+  $references = &backdrop_static(__FUNCTION__, array());
   $fields = isset($field) ? array($field['field_name'] => $field) : field_info_fields();
 
   foreach ($fields as $field_name => $file_field) {
--- a/core/modules/redirect/redirect.module
+++ b/core/modules/redirect/redirect.module
@@ -223,7 +223,7 @@ function redirect_set_current_redirect($redirect) {
  *   The redirect object (if any) for the current page. FALSE if not found.
  */
 function redirect_get_current_redirect() {
-  $redirect = backdrop_static('redirect_set_current_redirect', NULL);
+  $redirect = &backdrop_static('redirect_set_current_redirect', NULL);
 
   if (!isset($redirect)) {
     $redirect = redirect_load_by_source(current_path(), $GLOBALS['language']->langcode, backdrop_get_query_parameters());
--- a/core/modules/simpletest/tests/common.test
+++ b/core/modules/simpletest/tests/common.test
@@ -2782,6 +2782,10 @@ class CommonFormatDateTestCase extends BackdropWebTestCase {
     $this->backdropPost('admin/config/regional/date-time/formats/add', $edit, t('Add format'));
     $this->assertText(t('Date format updated.'));
 
+    // Reset the statically cached date formats so that these calls to
+    // format_date() will pick up the newly created date formats from config.
+    backdrop_static_reset('system_get_date_formats');
+
     $timestamp = strtotime('2007-03-10T00:00:00+00:00');
     $this->assertIdentical(format_date($timestamp, 'example_style', '', 'America/Los_Angeles'), '9 Mar 07', 'Test format_date() using an admin-defined date type.');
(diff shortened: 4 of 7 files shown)


laryn:
@bugfolder I am not sure why, but I notice there are two `format_date` related tests failing:

 -  fail Test format_date() using an admin-defined date type. common.test:2786
 -  fail Test format_date() using an admin-defined date type with different case. common.test:2787

The PR does adjust `system_get_date_formats` -- could it be related?

bugfolder:
> The PR does adjust system_get_date_formats -- could it be related?

Yes, they're definitely related. And it's kind of interesting why the test is failing. 

Let's start with the change in the PR, which affects `system_get_date_formats()`, which, pre-PR, looks like this:
```
function system_get_date_formats($date_format_name = NULL) {
  $date_formats = backdrop_static(__FUNCTION__, array());
  if (empty($date_formats)) {
    $date_formats = config_get('system.date', 'formats');
    // ... do some stuff
  }
  // more stuff
}
```

The PR changes one line to:
```
  $date_formats = &backdrop_static(__FUNCTION__, array());
```

With the `&` missing, the return value from the function is always an empty array, and so `system_get_date_formats()` has to fetch the system date formats from config every time, which sorta loses the benefit of using a static variable here. So I think that including the `&` is the right thing to do.

But then there's the matter of the failing test. The relevant bit of the test  looks like this:
```
    $this->admin_user = $this->backdropCreateUser(array('administer site configuration'));
    $this->backdropLogin($this->admin_user);

    // Add new date format.
    $edit = array(
      'label' => 'Example Style',
      'name' => 'example_style',
      'pattern' => 'j M y',
    );
    $this->backdropPost('admin/config/regional/date-time/formats/add', $edit, t('Add format'));

   // Add a second date format with a different case than the first.
    $edit = array(
      'label' => 'Example Style Uppercase',
      'name' => 'example_style_uppercase',
      'pattern' => 'j M Y',
    );
    $this->backdropPost('admin/config/regional/date-time/formats/add', $edit, t('Add format'));
    $this->assertText(t('Date format updated.'));

    $timestamp = strtotime('2007-03-10T00:00:00+00:00');
    $this->assertIdentical(format_date($timestamp, 'example_style', '', 'America/Los_Angeles'), '9 Mar 07', 'Test format_date() using an admin-defined date type.');
```

The test creates two new date formats, then compares that they give the same result. Each call to `backdropPost()` creates the new date form and stores it in config. But it does NOT clear the statically cached value (if any) from `system_get_date_formats()`.

So, with the `&` missing, there's no static cached date formats, so the call to `format_date()` has to get a fresh copy of the system formats from config, and the newly-created format is there, so the test passes.

But with the `&` present, then `system_get_date_formats()` uses the statically cached value, and since we created those two new formats as part of the same function call to `testAdminDefinedFormatDate()`, they're not present in the static cache. So `format_date()` doesn't know about them, and the test fails.

Obviously, we could just revert to the old way (no `&`), but that loses the benefit of static caching everywhere.

Although I'm generally nervous about the logic of "the test has started to fail, let's change the test so it doesn't fail," but I think the right thing to do here is _in the test_, clear the statically cached system date formats.

Since I'm nervous, though, I solicit further input.

bugfolder:
And on further pondering, I don't think the solution is to change the test; rather, the code that follows when one adds a new date format should clear the static cache at the time it is storing it in config.

laryn:
That makes sense to me.

bugfolder:
Made sense to me, too. But I was wrong! Turns out that the submission function for adding a date format already clears the static cache (as it should). But that has no effect within this test function. That's because those calls to `backdropPost()` end up making a cURL request to submit the edit form, which initiates a new, independent page load. Within that page load processing, yes, the statically cached date format is reset; but that static cache lives within a different scope from the current call to `testAdminDefinedFormatDate()`, whose static copy is (a) stale, (b) not reset. Hence the failure.

So, on third pondering, the clearing of the static cache belongs in the test. I'll update the PR. And the funny thing is, that clearing of the static cache should have been in the test all along, but its omission didn't matter because of the missing `&` that effectively disabled static caching of `system_get_date_formats`.

laryn:
Thanks @bugfolder for the fix, including the extra effort to get the tests working, and for the review @indigoxela. I've merged into 1.x and 1.19.x.

PR #3679, opened 2021-08-03
Issue #5092: Wrong url to update.php if language prefix in use

### LINKED ISSUE #5092: Fixed: Wrong url to update.php if language prefix in use
**Description of the bug**

On multilingual sites the first link on the status page to run update.php is wrong, as it includes the language prefix - it shouldn't.

**Steps To Reproduce**

On a site that uses language prefixes:

1. Install a module update manually (on the file system)
2. Go to admin/reports/status
3. Click on *the first link*

status-page-right-and-wrong-url

**Actual behavior**

Land on a 404 page ` The requested page "/en/core/update.php" could not be found. ` (note the language prefix)

**Expected behavior**

The link should *not* contain the language prefix, as the page /core/update.php is a real file.

**Additional information**

Here's the part of code in core/modules/system/system.install:
```
        if ($current_schema < $latest_schema) {
          $requirements['update']['severity'] = REQUIREMENT_ERROR;
          $requirements['update']['value'] = $t('Out of date') . ' (' . l(t('Run database updates'), 'core/update.php'). ')';
          $requirements['update']['description'] = $t('Some modules have database schema updates to install. You should run the <a href="@update">database update script</a> immediately.', array('@update' => base_path() . 'core/update.php'));
          break;
        }
```
Two variants in use:

` l(t('Run database updates'), 'core/update.php')` // wrong
` base_path() . 'core/update.php'` // right

An additional problem with the description:

`Some modules have database schema updates to install` ... in 90% it's not a db update, but a config update. But that might make sense as an extra issue (wording usually needs more discussion and the term "database updates" is used all over the place).

jayelless:
Fixes https://github.com/backdrop/backdrop-issues/issues/5092

### DIFF
--- a/core/modules/system/system.install
+++ b/core/modules/system/system.install
@@ -403,9 +403,10 @@ function system_requirements($phase) {
         $current_schema = backdrop_get_installed_schema_version($module);
         $latest_schema = end($updates);
         if ($current_schema < $latest_schema) {
+          $update_link = base_path() . 'core/update.php';
           $requirements['update']['severity'] = REQUIREMENT_ERROR;
-          $requirements['update']['value'] = $t('Out of date') . ' (' . l(t('Run database updates'), 'core/update.php'). ')';
-          $requirements['update']['description'] = $t('Some modules have database schema updates to install. You should run the <a href="@update">database update script</a> immediately.', array('@update' => base_path() . 'core/update.php'));
+          $requirements['update']['value'] = $t('Out of date (<a href="@update">Run database updates</a>)', array('@update' => $update_link));
+          $requirements['update']['description'] = $t('Some modules have database schema updates to install. You should run the <a href="@update">database update script</a> immediately.', array('@update' => $update_link));
           break;
         }
       }


core/modules/system/system.install
```diff
@@ -403,9 +403,10 @@ function system_requirements($phase) {
         $current_schema = backdrop_get_installed_schema_version($module);
         $latest_schema = end($updates);
         if ($current_schema < $latest_schema) {
+          $update_link = base_path() . 'core/update.php';
           $requirements['update']['severity'] = REQUIREMENT_ERROR;
-          $requirements['update']['value'] = $t('Out of date') . ' (' . l(t('Run database updates'), 'core/update.php'). ')';
-          $requirements['update']['description'] = $t('Some modules have database schema updates to install. You should run the <a href="@update">database update script</a> immediately.', array('@update' => base_path() . 'core/update.php'));
+          $requirements['update']['value'] = $t('Out of date') . ' (<a href="' . $update_link . '">' . t('Run database updates') . '</a>)';
```
indigoxela: We're not supposed to use a variable in a string like that.

Instead we do variable substitution via t / format_string. There are some examples in the [docs](https://docs.backdropcms.org/api/backdrop/core%21includes%21bootstrap.inc/function/t/1)

See also right below `<a href="@update">` for an example.

jayelless: @indigoxela I did consider a structure like `
$t('Out of date') . ' (' . t('<a href="@updates">Run database updates</a>', array('@updates' => $update_link)) . ')'
`
but I decided against that because it makes the call to "t" overly complex.  However, if that is what is required ...

klonos (approved):
LGTM 👍🏼

quicksketch:

PR #3681, opened 2021-08-11
Issue #5141: FieldException on content type creation

### LINKED ISSUE #5141: FieldException on content type creation if the Body field is still getting purged

jenlampton:
Fixes https://github.com/backdrop/backdrop-issues/issues/5141

### DIFF
--- a/core/modules/node/node.types.inc
+++ b/core/modules/node/node.types.inc
@@ -93,7 +93,7 @@ function node_type_form($form, &$form_state, $type = NULL) {
   // Create the body field by default on form submission. This may be altered by
   // other modules.
   $body_config = config_get('field.field.body');
-  if (isset($body_config) && $body_config['deleted'] == TRUE) {
+  if (!empty($body_config['deleted'])) {
     // Avoid FieldException on submit, if the body field is still getting
     // purged. That happens when all content types got deleted.
     $hint = t('The body field is marked for deletion, but the purge job did not complete yet. Not adding a body field now.');
--- a/core/modules/node/node.types.inc
+++ b/core/modules/node/node.types.inc
@@ -96,8 +96,31 @@ function node_type_form($form, &$form_state, $type = NULL) {
   if (!empty($body_config['deleted'])) {
     // Avoid FieldException on submit, if the body field is still getting
     // purged. That happens when all content types got deleted.
-    $hint = t("The Body field has been marked for deletion so it's not added.");
-    backdrop_set_message($hint, 'warning', FALSE);
+    $hint = t('The Body field was previously deleted.');
+
+    $cron = config_get('system.core', 'cron_safe_threshold');
+    $time = format_interval($cron);
+    if (user_access('administer site configuration')) {
+      $url = url('admin/config/system/cron');
+      $hint .= ' ' . t('<a href=":url">Run cron</a> to restore it, or try again in :time.', array(':url' => $url, ':time' => $time));
+    }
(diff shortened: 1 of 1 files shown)


core/modules/node/node.types.inc
```diff
@@ -92,10 +92,42 @@ function node_type_form($form, &$form_state, $type = NULL) {
...
+      $url = url('admin/config/system/cron');
+      $hint .= ' ' . t('<a href=":url">Run cron</a> to restore it, or try again in :time.', array(':url' => $url, ':time' => $time));
+    }
+    else {
+      $hint .= ' ' . t('You may try again in :time.', array(':time' => $time));
+    }
+
+
+
+    $hint = t();
```
docwilmot, line 113: Stray? Plus extra empty lines.

core/modules/node/node.types.inc
```diff
@@ -92,10 +92,42 @@ function node_type_form($form, &$form_state, $type = NULL) {
...
   // Create the body field by default on form submission. This may be altered by
   // other modules.
-  $form['body'] = array(
-    '#type' => 'value',
-    '#value' => TRUE,
-  );
+  $body_config = config_get('field.field.body');
+  if (!empty($body_config['deleted'])) {
+    // Avoid FieldException on submit, if the body field is still getting
+    // purged. That happens when all content types got deleted.
```
docwilmot, line 98: Maybe "...when all content types were deleted..."

core/modules/node/node.types.inc
```diff
@@ -92,10 +92,42 @@ function node_type_form($form, &$form_state, $type = NULL) {
...
+      $hint .= ' ' . t('You may try again in :time.', array(':time' => $time));
+    }
+
+
+
+    $hint = t();
+    $form['messages'] = array(
+      '#theme' => 'status_messages',
+      '#messages' => array(
+        'warning' => array(t($hint)),
```
docwilmot, line 117: `$hint` is already translated though.

core/modules/node/node.types.inc
```diff
@@ -92,10 +92,42 @@ function node_type_form($form, &$form_state, $type = NULL) {
...
+  if (!empty($body_config['deleted'])) {
+    // Avoid FieldException on submit, if the body field is still getting
+    // purged. That happens when all content types got deleted.
+    $hint = t('The Body field was previously deleted.');
+
+    $cron = config_get('system.core', 'cron_safe_threshold');
+    $time = format_interval($cron);
+    if (user_access('administer site configuration')) {
+      $url = url('admin/config/system/cron');
+      $hint .= ' ' . t('<a href=":url">Run cron</a> to restore it, or try again in :time.', array(':url' => $url, ':time' => $time));
```
docwilmot, line 105: Colon is not a valid `t()` placeholder is it?

PR #3682, opened 2021-08-11
Issue #3008 - handle paths when docroot is a symlink

### LINKED ISSUE #3008: Fixed: Symlink to document root causing several update and install issues

---

- core updates
- theme updates
- layout updates

---

bradbulger:
Fixes backdrop/backdrop-issues#3008 by comparing realpaths of the existing portions of the requested path and the system docroot.


bradbulger:
well, i guess it's not the branch...

quicksketch:
Sorry that Zen.CI isn't allowing access to the results page for some reason. As a core maintainer I can log in and see the results, which are related to the Installer module:

```
---- Installer (module): Installer UI tests (InstallerBrowserAdministrationTestCase) ----


Fail      Other      installer.test     135 InstallerBrowserAdministrationTestC
Fail      Other      installer.test     138 InstallerBrowserAdministrationTestC
    The Install Dependencies page was loaded
    "eee_installer_test 1.1.1" found
    DDD Installer test exists in the modules directory
Fail      Other      installer.test     141 InstallerBrowserAdministrationTestC
Fail      Other      installer.test     141 InstallerBrowserAdministrationTestC
    Failed to set field maintenance_mode to
Fail      Other      installer.test     141 InstallerBrowserAdministrationTestC
    Found the Install button
Fail      Other      installer.test     145 InstallerBrowserAdministrationTestC
    Found the requested form fields at
    The Enable modules page was loaded
Fail      Other      installer.test     148 InstallerBrowserAdministrationTestC
    EEE Installer test exists in the modules directory
Fail      Other      installer.test     155 InstallerBrowserAdministrationTestC
    Failed to set field modules[ddd_installer_test] to 1
Fail      Other      installer.test     155 InstallerBrowserAdministrationTestC
    Failed to set field modules[eee_installer_test] to 1
Fail      Other      installer.test     155 InstallerBrowserAdministrationTestC
    Found the Enable modules button
Fail      Other      installer.test     155 InstallerBrowserAdministrationTestC
    Found the requested form fields at
Fail      Other      installer.test     157 InstallerBrowserAdministrationTestC
    DDD Installer test was enabled.
Fail      Other      installer.test     158 InstallerBrowserAdministrationTestC
    EEE Installer test was enabled.
```

quicksketch:
Running the tests locally, they pass there also. 

```
Backdrop test run
---------------

Tests to be run:
 - Installer UI tests (InstallerBrowserAdministrationTestCase)

Test run started:
 Thursday, August 12, 2021 - 4:27pm

Test summary
------------

Installer (module): Installer UI tests (InstallerBrowserAdministrationTestCase) 152 passes, 0 fails, 0 exceptions, and 47 debug messages. [7.338s]

Test run duration: 7 sec
```

I tried restarting the Zen.CI tests several times but it is failing every time there.

quicksketch:
My guess is that this might be caused by Zen.CI itself using a symlink for the document root. When Zen.CI initializes, it calls this set up script: https://github.com/backdrop/backdrop/blob/1.x/core/misc/zen-ci/init_test.sh

Which includes:
```
# Link Backdrop files
ln -s $GITLC_DEPLOY_DIR/* ./
ln -s $GITLC_DEPLOY_DIR/.htaccess ./
```

So in a way this is most puzzling. The tests never should have been passing if the document root being a symlink is a problem. But with this PR, changing the handling to support symlinks actually breaks the testing environments.

quicksketch:
I reproduced this failure locally by setting up symlinks in the same way Zen.CI does.

Zen.CI does not use a symlink for the entire root. Instead it uses a symlink _per directory_. Looking at my local where I reproduced this, the directory structure looks like this:

```
ls -la
total 29
lrwxrwxrwx  1 nate nate    18 Aug 12 16:44 .htaccess -> backdrop/.htaccess
drwxrwxr-x 11 nate nate  4096 Aug 12 11:41 backdrop/
lrwxrwxrwx  1 nate nate    13 Aug 12 16:44 core -> backdrop/core/
drwxrwxrwx  7 nate nate  4096 Aug 12 16:45 files/
lrwxrwxrwx  1 nate nate    18 Aug 12 16:44 index.php -> backdrop/index.php*
lrwxrwxrwx  1 nate nate    16 Aug 12 16:44 layouts -> backdrop/layouts/
lrwxrwxrwx  1 nate nate    20 Aug 12 16:44 LICENSE.txt -> backdrop/LICENSE.txt
lrwxrwxrwx  1 nate nate    16 Aug 12 16:44 modules -> backdrop/modules/
lrwxrwxrwx  1 nate nate    18 Aug 12 16:44 README.md -> backdrop/README.md
lrwxrwxrwx  1 nate nate    19 Aug 12 16:44 robots.txt -> backdrop/robots.txt
lrwxrwxrwx  1 nate nate    27 Aug 12 16:44 settings.local.php -> backdrop/settings.local.php
-rw-rw-r--  1 nate nate 19308 Aug 12 11:41 settings.php
lrwxrwxrwx  1 nate nate    14 Aug 12 16:44 sites -> backdrop/sites/
lrwxrwxrwx  1 nate nate    15 Aug 12 16:44 themes -> backdrop/themes/
``` 

Running the tests in this configuration fails with the following fatal error:

```
Argument 1 passed to installer_browser_get_project_release_data() must be of the type array, null given, called in /app/backdrop/core/modules/installer/tests/installer.test on line 186
```

I think overall we might have a problem where Zen.CI might not be using Clean URLs, perhaps because it's symlinking the .htaccess file, which might disable it from being read by Apache because symlinks have a 777 permission. I'm still investigating.

bradbulger:
What is the "Internal Server Error" from the zenci/deploy/pr-3682 step? This last set of check looks like it is never going to finish, but in prior cases it seemed as if there were uncommitted changes at some step.

Does the installer work in theory with Clean URLs disabled? I doubt that I tried that case.

quicksketch:
> Does the installer work in theory with Clean URLs disabled? I doubt that I tried that case.

I don't believe it does. I'm trying various testing over in my own PR at https://github.com/backdrop/backdrop/pull/3685, but I haven't had any luck yet getting things any further.

> What is the "Internal Server Error" from the zenci/deploy/pr-3682 step? 

I don't think that's related to anything we're working on here. Zen.CI does a request over to backdropcms.org and the server returned a 503 for the response. We might have been doing a deployment at that time. I restarted that test.

bradbulger:
I tried setting up an installation in the way outlined above, using ddev configured to run Apache. It seems to be working OK with Clean URLs enabled. Once I fixed the file system permissions, I was able to install Devel through the UI, and all the Installer UI aka Installer (module) tests worked. With Clean URLs disabled, 20 of the Installer UI tests fail. This is what I saw in a normal installation as well. All of this is the same on the base 1.x code as on this issue3008v2 branch.

quicksketch:
I found that I can get all tests passing if we prepare the installation on Zen.CI by symlinking the entire web root instead of symlinking each of the subdirectories individually. I filed a PR against this branch at https://github.com/bradbulger/backdrop/pull/1.

I think this might actually point out a new (but even less common) scenario. Symlinking individual directories is indeed broken by these changes. Consider the Zen.CI setup:

Git Checkout: `/home/test/github/pr/3685`
www root: `/home/test/www`

And then symlinks for each directory from the web root to the git checkout, such as:

`/home/test/www/modules` -> `/home/test/github/pr/3685/modules` 

So when checking the jail directory versus the install directory, they end up in different locations:

Jail: `/home/test/www`
Install path: `/home/test/github/pr/3685/modules`

If the _entire_ directory is symlinked, then the Jail and install path can be calculated to be within each other:

Jail: `/home/test/github/pr/3685`
Install path: `/home/test/github/pr/3685/modules`

Note in my own testing, I nested the symlinked copy within the same directory, so that was considered okay.

I think the Clean URLs thing turns out to be completely unrelated. We should split that out into a new issue.

PR #3687, opened 2021-08-14
Prevent Gray Header bar from affecting header region of flexible layo…

### LINKED ISSUE #5149: Fixed style issues with Flexible layout template UI and make it more consistent with the layout/blocks UI
This is part of the #4046 meta.

During this week's design meeting, we discussed various styling issues with the flexible layout templates UI:

1. The "Header" region has a darker gray background than the rest of the regions, and when you hover over it, there is a dark gray bottom border shown:

    Screenshot 2021-08-12 at 23 52 21

   ...for context, hovering over the other regions makes them acquire a white background/container, with a bottom/right border of the same dark gray color as the one that appears when hover :

    image

   It seems that the header inherits the style of the header of the admin theme 🤷🏼 

2. The next issue with this UI is that there is no indication that the various rows can be dragged, unless you hover over the rows (so users without a pointing device won't know). The blocks in the layout admin UI on the other hand have a crosshair icon shown always at their top-left.

   Since rows in flexy templates can only be moved up/down, we should add an up/down arrow icon.

3. If you configure a row in a flexy template, and change the "Row width behavior" (under the "Style" fieldset) setting to "full width", then the region style changes to indicate that, but it may not be as clear that this is the case. Compare the widths of the containers of the "CONTENT" vs the "FOOTER" rows in the screenshot below:

   Screenshot 2021-08-13 at 00 20 16

jenlampton:
Fixes https://github.com/backdrop/backdrop-issues/issues/5149

### DIFF
--- a/core/modules/layout/templates/layout--flexible.tpl.php
+++ b/core/modules/layout/templates/layout--flexible.tpl.php
@@ -35,11 +35,15 @@
       ?>
       <<?php print $row['element']; ?> data-row-id="<?php print $name; ?>" class="<?php print implode(' ', $row_classes); ?>" <?php print $row['row_id']; ?>>
       <div class="<?php print $row['row_class']; ?>">
-        <?php if ($region_buttons): ?>
-          <div class="layout-flexible-region-buttons clearfix">
-            <?php print $region_buttons[$name]; ?>
-          </div>
-        <?php endif; ?>
+        <div class="layout-flexible-region-top clearfix">
+          <span class="handle"></span>
+          <h2 class="label">Row</h2>
+          <?php if ($region_buttons): ?>
+            <span class="buttons">
+              <?php print $region_buttons[$name]; ?>
+            </span>
+          <?php endif; ?>
+        </div>
         <div class="l-flexible-row row">
         <?php foreach ($row['regions'] as $region): ?>
           <div class="l-col col-md-<?php print $region['region_md']; ?>">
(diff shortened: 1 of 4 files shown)


core/modules/layout/templates/layout--flexible.tpl.php
```diff
@@ -35,11 +35,15 @@
...
       <<?php print $row['element']; ?> data-row-id="<?php print $name; ?>" class="<?php print implode(' ', $row_classes); ?>" <?php print $row['row_id']; ?>>
       <div class="<?php print $row['row_class']; ?>">
-        <?php if ($region_buttons): ?>
-          <div class="layout-flexible-region-buttons clearfix">
-            <?php print $region_buttons[$name]; ?>
-          </div>
-        <?php endif; ?>
+        <div class="layout-flexible-region-top clearfix">
+          <span class="handle"></span>
+          <h2 class="label">Row</h2>
```
klonos:
```suggestion
          <h2 class="label"><?php print t('Row'); ?></h2>
```

core/modules/layout/templates/layout--flexible.tpl.php
```diff
@@ -48,6 +52,14 @@
                 <div class="layout-editor-region-title clearfix">
                   <h2 class="label"><?php print $region['region_name']; ?></h2>
                 </div>
+                <div class="layout-editor-block">
+                  <div class="layout-editor-block-title">
+                    <span class="text">Example Block</span>
```
klonos:
```suggestion
                    <span class="text"><?php print t('Example block'); ?></span>
```

core/modules/layout/templates/layout--flexible.tpl.php
```diff
@@ -48,6 +52,14 @@
...
+                  </div>
+                  <div class="layout-editor-block-content">
+                    <p>Example block content</p>
```
klonos:
```suggestion
                    <p><?php print t('Example block content'); ?></p>
```

core/themes/seven/css/style.css
```diff
@@ -1310,7 +1310,8 @@ div.filter-options select {
   margin: 0;
   background: none;
 }
-#layout-edit-main .l-header:before {
+#layout-edit-main .l-header:before,
+#layout-flexible-edit-main .l-header:before {
   /* Make sure the page gray bg isn't showing inside layout edit area */
```
klonos, line 1315:
```suggestion
  /* Don't show the gray background of the page inside the layout edit area. */
```

jenlampton:
Closing in favor of https://github.com/backdrop/backdrop/pull/3691

PR #3688, opened 2021-08-15
Issue #4753: Add Github Actions

### LINKED ISSUE #4753: Added: GitHub Actions for automated tests
There has been discussion about using [GitHub Actions](https://docs.github.com/en/free-pro-team@latest/actions) for automated testing of PRs, instead of the current Zen.ci service provided by @Gormartsen.

Opening this issue for discussion and possible testing of the service, so we can make a decision if it's going to be suitable for us or not.

UPDATE: there seems to be consensus now, that switching to GHA is the way to go.

indigoxela:
Fixes https://github.com/backdrop/backdrop-issues/issues/4753

Complicated by https://github.com/backdrop/backdrop-issues/issues/1478

The concept:

- Whatever can run from shared memory gets moved there. The database and the files directory.
- Apache runs as www-data, but php runs as "runner", so we need no chown. They communicate over a local socket (that's a common setup)
- Three files:
  - Apache Vhost file
  - Bash script for webserver install and setup
  - The actual workflow file defining the test matrices

See also the [official docs](https://docs.github.com/en/actions/reference/workflow-syntax-for-github-actions)


.github/workflows/functional-tests.yml
```diff
@@ -0,0 +1,81 @@
+name: Functional tests
+on:
+  pull_request:
+    branches: [1.*]
+jobs:
+  simpletest:
+    name: Simpletest batches
+    runs-on: ubuntu-18.04
```
quicksketch, line 8: Can we bump this to 20.04, the most recent Ubuntu LTS release?

indigoxela: No problem, but then I'd have to install a different Database, because 20.04 ships with MySQL8 - which Backdrop doesn't fully support yet.

indigoxela: After two hours of wasted life time (MySQL 8 won't start), and a quick experiment with the MariaDB action - docker container won't work with our changes), I switched back to ubuntu-18.04.

quicksketch: Thanks @indigoxela, we can tackle this in a follow-up. I'm sorry my request turned out to be a lot more challenging that I had expected.

indigoxela: > we can tackle this in a follow-up

Phew, maybe it was just bad timing - being in a hurry is never helpful. :wink: I'm sure we can figure out what's so different in U20 or M8, that it doesn't work.

PR #3689, opened 2021-08-18
Issue #5157: Fix #states when adding #attributes to container FAPI elements

### LINKED ISSUE #5157: [DX] Form API: Adding `#attributes` to fields of type `container` breaks `#states`
This is a sibling issue to #3817...

It is a known issue that `'#markup'` form elements do not support `#states`. So this won't work:

```php
    $form['controller'] = array(
      '#type' => 'checkbox',
      '#title' => t('Show/hide markup element'),
    );

    $form['warning'] = array(
      '#markup' => t('blah blah...'),
      '#states' => array(
        'visible' => array(
          ':input[name="controller"]' => array('checked' => TRUE),
        ),
      ),
    );
```

The workaround is to add a `'#type' => 'container'` around them, and use `#states` on the container form element instead of directly on the markup element.

This works fine:

```php
    $form['controller'] = array(
      '#type' => 'checkbox',
      '#title' => t('Show/hide container'),
    );

    $form['some_warning'] = array(
      '#type' => 'container',
      '#states' => array(
        'visible' => array(
          ':input[name="controller"]' => array('checked' => TRUE),
        ),
      ),
    );
    $form['some_warning']['message'] = array(
      '#markup' => t('blah blah...'),
    );
```

**This issue here is not about solving that problem**...

Everything works as expected with the above-mentioned workaround, until you try adding any class to the container form element (via `#attributes`).

This does NOT work:

```php
    $form['controller'] = array(
      '#type' => 'checkbox',
      '#title' => t('Show/hide container'),
    );

    $form['some_warning'] = array(
      '#type' => 'container',
      '#attributes' => array('class' => array('messages', 'warning')), // Adding this breaks #states.
      '#states' => array(
        'visible' => array(
          ':input[name="controller"]' => array('checked' => TRUE),
        ),
      ),
    );
    $form['some_warning']['message'] = array(
      '#markup' => t('blah blah...'),
    );
```

To make it work again, one needs to manually add the `form-wrapper` class to the container element:

```php
    $form['controller'] = array(
      '#type' => 'checkbox',
      '#title' => t('Show/hide container'),
    );

    $form['some_warning'] = array(
      '#type' => 'container',
      // The "form-wrapper" class is required here, otherwise #states doesn't
      // work.
      '#attributes' => array('class' => array('form-wrapper', 'messages', 'warning')),
      '#states' => array(
        'visible' => array(
          ':input[name="controller"]' => array('checked' => TRUE),
        ),
      ),
    );
    $form['some_warning']['message'] = array(
      '#markup' => t('blah blah...'),
    );
```

klonos:
Fixes https://github.com/backdrop/backdrop-issues/issues/5157


klonos:
Zen CI acting up again 😓

core/includes/form.inc
```diff
@@ -3629,17 +3629,23 @@ function form_process_container($element, &$form_state) {
...
       $element['#attributes']['id'] = $element['#id'];
     }
-    // Add the form-wrapper class.
-    if (empty($element['#attributes']['class'])) {
+    // Ensure that the classes within #attributes have been initialized.
+    if (!array_key_exists('class', $element['#attributes'])) {
+      $element['#attributes']['class'] = array();
+    }
+    // Add the form-wrapper class if not already added.
+    if (!in_array('form-wrapper', $element['#attributes']['class'])) {
```
klonos: Not sure why we were previously adding this class only when there were no other classes. Doesn't make sense to me at all 🤷🏼

klonos: ...this is why: https://github.com/backdrop/backdrop-issues/issues/711

We need to test and make sure that that issue has not regressed with this change here

core/includes/form.inc
```diff
@@ -3629,17 +3629,23 @@ function form_process_container($element, &$form_state) {
  */
 function theme_container($variables) {
   $element = $variables['element'];
-  // Ensure #attributes is set.
-  $element += array('#attributes' => array());
+  // Ensure #attributes is set, but do not overwrite them if they have already
+  // been set.
+  $attributes = isset($element['#attributes']) ? $element['#attributes'] : array();
+  $element += array('#attributes' => $attributes);
```
indigoxela: This part of the change here seems redundant to me. The `+=` already takes care that the array is only initialized, if it doesn't exist yet.

An example to play:
```
$foo = array('attributes' => array('class' => 'foo'));
$foo += array('attributes' => array());
var_dump($foo);
```
^^ The foo class is preserved.

Maybe the [docs](https://www.php.net/manual/en/language.operators.array.php) refresh your memory after your long vacation. :smile:

klonos: You are absolutely right of course 😅 ...I don't know why this was initially breaking on my local.

core/includes/form.inc
```diff
@@ -3629,23 +3629,21 @@ function form_process_container($element, &$form_state) {
...
-    if (!array_key_exists('class', $element['#attributes'])) {
-      $element['#attributes']['class'] = array();
-    }
-    // Add the form-wrapper class if not already added.
-    if (!in_array('form-wrapper', $element['#attributes']['class'])) {
+    // Add the form-wrapper class if not already added, but only for 'container'
+    // elements, otherwise #states does not work properly for 'checkboxes' and
+    // 'radios' elements.
+    // See: https://github.com/backdrop/backdrop-issues/issues/711#issuecomment-84353916
+    if ($element['#type'] == 'container' && !in_array('form-wrapper', $element['#attributes']['class'])) {
```
indigoxela: As this happens in function theme_container, I don't think that the extra check `$element['#type'] == 'container'` is needed here. I breaks the tests anyway, as '#type' doesn't exist.

klonos: @indigoxela please see https://github.com/backdrop/backdrop-issues/issues/197 where with https://github.com/backdrop/backdrop/pull/324 we deprecated `theme_radios()` and `theme_checkboxes()`, and we now have `theme_container()` also handle radios and checkboxes. So `$element['#type']` hare can also be `radios` or `checkboxes`.

klonos: ...what I'm trying to achieve here is to add the `form-wrapper` class to containers, and not to radios/checkboxes.

core/includes/form.inc
```diff
@@ -3629,17 +3629,21 @@ function form_process_container($element, &$form_state) {
...
+    // Assign an HTML ID.
     if (!array_key_exists('id', $element['#attributes'])) {
       $element['#attributes']['id'] = $element['#id'];
     }
-    // Add the form-wrapper class.
-    if (empty($element['#attributes']['class'])) {
+    // Add the form-wrapper class if not already added, but only for 'container'
+    // elements, otherwise #states does not work properly for 'checkboxes' and
+    // 'radios' elements.
+    // See: https://github.com/backdrop/backdrop-issues/issues/711#issuecomment-84353916
```
jenlampton: I would remove the @see here, the explanation is enough :)

core/includes/form.inc
jenlampton: @klonos then we also need to add isset($element['#type'])` before `$element['#type'] == 'container'` to prevent the notices where `#type` doesn't exist.

core/includes/form.inc
klonos: Done 👍🏼 ...see: [`f398185` (#3689)](https://github.com/backdrop/backdrop/pull/3689/commits/f39818515d0fb73d9e9dfc3dba170318016e44d9)

core/includes/form.inc
klonos: @jenlampton done with [`4ca123c` (#3689)](https://github.com/backdrop/backdrop/pull/3689/commits/4ca123c4b12d3b8b1f8ae8fe472aff5756087fad)

quicksketch:

PR #3690, opened 2021-08-18
Issue #3008: handle docroots that are symlinks

### LINKED ISSUE #3008: Fixed: Symlink to document root causing several update and install issues

bradbulger:
Fixes backdrop/backdrop-issues#3008 - update FileTransfer::checkPath() to check real filesystem paths when normal path checking fails. Usually encountered when the web server is serving from a document root that is a symlink.

### DIFF
--- a/core/includes/filetransfer/filetransfer.inc
+++ b/core/includes/filetransfer/filetransfer.inc
@@ -168,8 +168,15 @@ abstract class FileTransfer {
    */
   protected final function checkPath($path) {
     $chroot_jail = $this->chroot . $this->jail;
-    $real_jail = backdrop_realpath($chroot_jail);
     $chroot_path = $this->chroot . $path;
+    $full_path = backdrop_realpath(substr($chroot_path, 0, strlen($chroot_jail)));
+    $full_path = $this->fixRemotePath($full_path, FALSE);
+    if ($chroot_jail === $full_path) {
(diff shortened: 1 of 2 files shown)


core/includes/filetransfer/filetransfer.inc
```diff
@@ -167,12 +167,32 @@ abstract class FileTransfer {
...
+      return;
     }
+    // look for a symlink
+    $parts = explode('/', $chroot_path);
+    // an empty string is the first entry
+    $front_path = array_shift($parts);
+    while ($parts) {
+      $front_path .= '/' . array_shift($parts);
+      if (is_link($front_path)) {
+        $real_path = realpath($front_path);
```
indigoxela: Switching from backdrop_realpath() to realpath() means, that stream wrapper support in that function gets dropped.

My concern is, that this might break other stuff, unrelated to updates. I'm not sure, if we have a test case for that (Zen.CI is broken today, so we're sort of "blind").

bradbulger: Then it shouldn't do the is_link() either - only test if it exists and if it does, try backdrop_realpath().

bradbulger:
Deferred to another PR to work around the issue by patching updater.php

PR #3691, opened 2021-08-20
Prevent Gray Header bar from affecting header region of flexible layout template.

### LINKED ISSUE #5149: Fixed style issues with Flexible layout template UI and make it more consistent with the layout/blocks UI

jenlampton:
Fixes https://github.com/backdrop/backdrop-issues/issues/5149

### DIFF
--- a/core/modules/layout/css/layout.admin.css
+++ b/core/modules/layout/css/layout.admin.css
@@ -1,3 +1,7 @@
+/**
+ * Layout admin css.
+ */
+
 /* Layout listing form. */
 .layout-list .layout-group td {
   background: #dddddd;
@@ -153,6 +157,9 @@
   -webkit-box-sizing: border-box;
   box-sizing: border-box;
 }
+.layout-editor-region-title {
+  padding: 0 0 10px 0;
+}
 .layout-editor-region-title .label {
   float: left; /* LTR */
   margin: 0;
@@ -160,7 +167,7 @@
   line-height: 1.2;
   text-transform: uppercase;
   font-weight: normal;
-  color: #908990;
+  color: #5b575b;
 }
 [dir="rtl"] .layout-editor-region-title .label {
   float: right;
@@ -173,7 +180,7 @@
 .layout-editor-region-title .dropbutton-wrapper {
   float: left; /* LTR */
   width: 140px;
-  margin: 4px 0 10px
+  margin: 4px 0 10px;
 }
 [dir="rtl"] .layout-editor-region-title .dropbutton-wrapper {
   float: right;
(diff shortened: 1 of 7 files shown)


core/modules/layout/css/layout.admin.css
```diff
@@ -173,7 +180,7 @@
 .layout-editor-region-title .dropbutton-wrapper {
   float: left; /* LTR */
   width: 140px;
-  margin: 4px 0 10px
+  margin: 4px 00;
```
klonos: ??

core/modules/layout/css/layout.flexible.admin.css
```diff
@@ -0,0 +1,82 @@
+/**
+ * Flexible template admin css.
+ */
+
+ .layout-flexible-region-top .label {
+   display: inline-block;
+   margin: 0;
+   padding: 4px 12px 4px 0; /* LTR */
```
klonos: There's no matching `[dir="rtl"]` for this

core/modules/layout/css/layout.admin.css
jenlampton: oops, fixed! (I was trying to add the `;`)

core/modules/layout/css/layout.flexible.admin.css
jenlampton: added

core/modules/layout/css/layout.admin.css
```diff
@@ -153,14 +157,17 @@
...
+}
 .layout-editor-region-title .label {
   float: left; /* LTR */
   margin: 0;
   padding: 4px 12px 4px 4px; /* LTR */
   line-height: 1.2;
   text-transform: uppercase;
   font-weight: normal;
-  color: #908990;
+  color: #5b575b; /* WCAG AAA on white bg */
```
klonos: As previously agreed, remove this, and add a GitHub comment.

```suggestion
  color: #5b575b;
```

core/modules/layout/css/layout.admin.css
```diff
@@ -173,7 +180,7 @@
 .layout-editor-region-title .dropbutton-wrapper {
   float: left; /* LTR */
   width: 140px;
-  margin: 4px 0 10px
+  margin: 4px 0 10px;
```
klonos, line 183: Good catch 👍🏼 ...how did we miss that? ...cant wait to have a linter as part of our PR tests.

core/modules/layout/layout.flexible.inc
```diff
@@ -115,6 +115,11 @@ function layout_flexible_template_configure_form($form, &$form_state, LayoutFlex
   form_load_include($form_state, 'inc', 'layout', 'layout.admin');
   form_load_include($form_state, 'inc', 'layout', 'layout.flexible');
 
+  $form['help'] = array(
+    '#type' => 'help',
+    '#markup' => t('This page allows you to add or remove rows, and change the row order by dragging them up and down. You can also change the number of regions within each row, and adjust the width of each region.'),
```
klonos:
```suggestion
    '#markup' => t('This page allows you to add or remove rows, and change the order of rows by dragging them up and down. You can also change the number of regions within each row, and adjust the width of each region.'),
```

klonos (approved):
LGTM 👍🏼

quicksketch:

PR #3695, opened 2021-08-22
Issue #3224: Sync config on install

### LINKED ISSUE #3224: Module config in /config is not synced on install

docwilmot:
Fixes https://github.com/backdrop/backdrop-issues/issues/3224

### DIFF
--- a/core/includes/config.inc
+++ b/core/includes/config.inc
@@ -297,6 +297,7 @@ function config_install_default_config($module, $config_name = NULL) {
         if ($config->isNew()) {
           $config->setData($data);
           $config->save();
+          module_invoke_all('config_create', $config);
         }
       }
     }
--- a/core/includes/config.inc
+++ b/core/includes/config.inc
@@ -301,6 +301,8 @@ function config_install_default_config($module, $config_name = NULL) {
         }
       }
     }
+
+    backdrop_flush_all_caches();
   }
 }
 
--- a/core/includes/config.inc
+++ b/core/includes/config.inc
@@ -296,8 +296,8 @@ function config_install_default_config($module, $config_name = NULL) {
         // We only create new configs, and do not overwrite existing ones.
         if ($config->isNew()) {
           $config->setData($data);
-          $config->save();
           module_invoke_all('config_create', $config);
+          $config->save();
         }
       }
     }
--- a/core/includes/config.inc
+++ b/core/includes/config.inc
@@ -301,8 +301,6 @@ function config_install_default_config($module, $config_name = NULL) {
         }
       }
     }
-
-    backdrop_flush_all_caches();
   }
 }
 
--- a/core/modules/config/tests/config_test_hooks/config_test_hooks.module
+++ b/core/modules/config/tests/config_test_hooks/config_test_hooks.module
@@ -45,6 +45,12 @@ function config_test_hooks_config_create(Config $staging_config) {
     // Set a value before the config is saved.
     $staging_config->set('favorite_color', 'red');
   }
+  if ($staging_config->getName() === 'config_test.simple') {
+    // Set a value on the config file in another module when it is enabled.
+    $config = config('config_test_hooks.data');
+    $config->set('favorite_cheese', 'cheddar');
+    $config->save();
+  }
 }
 
 /**
@@ -66,3 +72,16 @@ function config_test_hooks_config_delete(Config $active_config) {
     config_set('config_test.simple', 'deleted_animal', $active_config->get('favorite_animal'));
   }
 }
+
+/**
+ * Implements hook_config_info().
+ */
+function config_test_hooks_config_info() {
+  $prefixes['config_test_hooks.data'] = array(
+    'name' => t('Configuration test hooks data'),
+    'group' => t('Configuration test items'),
+  );
+
+  return $prefixes;
+}
+
--- a/core/modules/field/field.module
+++ b/core/modules/field/field.module
@@ -562,8 +562,21 @@ function field_config_delete_validate(Config $active_config, $all_changes) {
 function field_config_create(Config $staging_config) {
   $config_name = $staging_config->getName();
   if (strpos($config_name, 'field.field.') === 0) {
+    $field_data_table_exists = FALSE;
+
     $field = $staging_config->get();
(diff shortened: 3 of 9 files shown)


core/modules/field/field.module
```diff
@@ -562,8 +562,21 @@ function field_config_delete_validate(Config $active_config, $all_changes) {
 function field_config_create(Config $staging_config) {
   $config_name = $staging_config->getName();
   if (strpos($config_name, 'field.field.') === 0) {
+    $field_data_table_exists = FALSE;
+
     $field = $staging_config->get();
-    field_create_field($field);
+    $schema = _field_sql_storage_schema($field);
```
quicksketch, line 568: This makes me a little nervous because it's introducing hard-coupling between field_sql_storage and field module. We don't support other field storages however, so this might be acceptable. We already introduced something similar in `field_validate_field()`, where we also explicitly check if the DB tables exist.

I think handling this we should try either of the following:

1. Determine why `field_config_create()` is getting called twice and avoid doing it in the first place.
2. Create a follow-up issue to introduce `field_storage_exists()` and then put a TODO both here and in `field_validate_field()`.

docwilmot: > Determine why field_config_create() is getting called twice and avoid doing it in the first place.

It isn't called twice as far as I have noticed. Did you notice that happening?
> Create a follow-up issue to introduce field_storage_exists() and then put a TODO both here and in field_validate_field().

Sounds good. We could use `_field_info_collate_fields()` for this as it provides the field storage details.

quicksketch: > It isn't called twice as far as I have noticed. Did you notice that happening?

This may be an incorrect assumption on my part. I would have thought that it getting called twice would be the reason for why we need to check if the table exists, because it would have already been called once before and created them. If that's not the case, why do these tables already exist at the time this is called?

docwilmot: Ahh, I should have realised. The reason is here: https://github.com/backdrop/backdrop-issues/issues/3224#issuecomment-903198426

jenlampton: > We don't support other field storages however, so this might be acceptable. 

IMHO I think this is acceptable :)

docwilmot: > Create a follow-up issue to introduce field_storage_exists() and then put a TODO both here and in field_validate_field().

See https://github.com/backdrop/backdrop-issues/issues/5180

core/includes/config.inc
```diff
@@ -296,10 +296,13 @@ function config_install_default_config($module, $config_name = NULL) {
         // We only create new configs, and do not overwrite existing ones.
         if ($config->isNew()) {
           $config->setData($data);
+          module_invoke_all('config_create', $config);
           $config->save();
         }
       }
     }
+
+    backdrop_flush_all_caches();
```
quicksketch: The commit message for adding this line says "Fixing missing admin bar".

Is that a problem somehow created by the invoking of `hook_config_create()`, or was it an existing issue when enabling new modules?

`backdrop_flush_all_caches()` is pretty intensive and would happen for every individual module enabled (if multiple were enabled at once). `backdrop_flush_all_caches()` is already called after enabling and disabling the entire set of modules in `system_modules_submit()`.

Could this be made more targeted or elaborate on why it's needed?

quicksketch: I removed this in testing and could not find a consequence of it. I pushed up a commit for now without it for further testing.

docwilmot: We were having a weird thing where admin bar disappeared (consistently) after syncing a field and instance. Tried lots of things, thought that that cache clear fixed it, but forgot to ask @stpaultim to test that. Wasnt meant to be permanent, I suspect what we needed was an admin bar cache clear, but that error was a bit frustrating.

quicksketch: Okay, I hope it's fine that I removed it then. We have lots of places where the admin bar cache needs to be flushed via a hook, including language updates, menu links, vocabularies, content types... maybe we should add `admin_bar_config_create()` to flush the caches on new config creation, or `admin_bar_enable_modules()` to flush on module enable (though I think all caches are already flushed after module enable).

core/modules/config/tests/config_test_hooks/config_test_hooks.module
```diff
@@ -45,6 +45,12 @@ function config_test_hooks_config_create(Config $staging_config) {
     // Set a value before the config is saved.
     $staging_config->set('favorite_color', 'red');
   }
+  if ($staging_config->getName() === 'config_test.simple') {
+    // Set a value on the config file in another module when it is enabled.
+    $config = config('config_test_hooks.data');
+    $config->set('favorite_cheese', 'cheddar');
```
jenlampton, line 51: @docwilmot you are going to be responsible for making core cheesy :D

core/includes/config.inc
jenlampton: I tested with Tim's test recipe on my local site, and the admin bar appeared for me after installing the module on my local site.

laryn:
Merged into 1.x so this will show up in 1.20.0. Big thanks @docwilmot, @hosef, @jhodgdon-drp, @quicksketch, @jenlampton, and @stpaultim!

PR #3696, opened 2021-08-22
Issue #5082: Visibility condition options incomplete for layouts with non-string contexts

### LINKED ISSUE #5082: Visibility condition options incomplete for layouts with non-string contexts
On the "Configure Layout" page, when creating a layout with a path with a placeholder and a non-string context (like Node) for the placeholder, the visibility conditions options don't include all possible options.

To reproduce:

1. Go to admin/structure/layouts.
2. Select "Add layout".
3. Enter a layout name and select a template.
4. For the Path, enter "foobaz/%".
5. Under Contexts, select Type = Node.
6. Click "Add visibility condition".
7. Click the "Select" dropdown to see the available options.

Expected result: all possible conditions, including "Node: NID" and "Node: Type".

Actual result: a shorter list, shown here:

options-1

Now, we can get the longer list if we save the layout and then go back and configure it.

1. Cancel out of the "Add visibility condition" dialog.
2. Click "Create Layout" button.
3. Click "Configure Layout" tab.
4. Click "Add visibility condition".
5. Click the "Select" dropdown to see the available options.

This time, we see all possible conditions, as here:

options-2

In stepping through the code, I find that a place where things go awry is in layout.admin.inc, function `layout_condition_add_form()`, line 1817, which is:

```
$layout_has_context = $layout && $layout->hasContexts($required_contexts);
```

At this point, `$required_contexts` contains a `node` context, but the relevant context in `$layout` is a LayoutStringContext, rather than an EntityLayoutContext (which is what it should be if Type = Node). So I infer that the ajax called when we change the context Type is not updating the tempstore Layout as it should.

We can encounter the same issue by creating and saving the layout with a String pass-through context, then going back and editing the layout to change the context to a Node type.

So, the workaround is simple: just save the layout, then go back and re-edit it to get the full list of possible visibility conditions. But it would be nice to have them show up without having to save-and-re-edit.

This seems related to https://github.com/backdrop/backdrop-issues/issues/5065. Attempting to save the layout with that added visibility condition results in the problems of https://github.com/backdrop/backdrop-issues/issues/5065.

Doing testing on 1.x.

jayelless:
Fixes https://github.com/backdrop/backdrop-issues/issues/5082.

This also fixes issue #5145 which is a duplicate of #5082.

### DIFF
--- a/core/modules/layout/layout.admin.inc
+++ b/core/modules/layout/layout.admin.inc
@@ -649,6 +649,16 @@ function layout_settings_form_validate(&$form, &$form_state) {
   foreach ($contexts as $context) {
     $context_plugins[] = $context->plugin;
   }
+
+  // Add any plugin specified for an arbitrary wildcard path if required.
+  if (isset($form_state['input']['context']['required'])) {
+    $new_context_plugins = array();
+    foreach ($form_state['input']['context']['required'] as $required_context) {
+      $new_context_plugins[] = $required_context['plugin'];
+    }
+    $context_plugins = array_merge($context_plugins, $new_context_plugins);
+  }
+
   foreach ($layout->conditions as $key => $condition) {
     $required_contexts = $condition->getRequiredContexts();
     if (array_diff($required_contexts, $context_plugins)) {
@@ -1814,6 +1824,21 @@ function layout_condition_add_form($form, &$form_state, Layout $layout = NULL, B
       'callback' => 'layout_condition_ajax_style',
     ),
   );
+
+  if ($is_new_condition && isset($form_state['input']['context']['required'])) {
+    foreach ($form_state['input']['context']['required'] as $required_context) {
+      $new_context_plugin = $required_context['plugin'];
+      switch ($new_context_plugin) {
+        case 'node':
+        case 'file':
+        case 'taxonomy_term':
+          $new_context = new EntityLayoutContext($new_context_plugin);
+          $layout->setContexts($new_context_plugin, $new_context);
+          break;
+      }
+    }
+  }
+
   foreach ($all_access_info as $access_info) {
     $required_contexts = !empty($access_info['required contexts']) ? $access_info['required contexts'] : array();
     $layout_has_context = $layout && $layout->hasContexts($required_contexts);


core/modules/layout/layout.admin.inc
```diff
@@ -649,6 +649,16 @@ function layout_settings_form_validate(&$form, &$form_state) {
   foreach ($contexts as $context) {
     $context_plugins[] = $context->plugin;
   }
+
+  // Add any plugin specified for an arbitrary wildcard path if required.
+  if (isset($form_state['input']['context']['required'])) {
+    $new_context_plugins = array();
+    foreach ($form_state['input']['context']['required'] as $required_context) {
```
quicksketch, line 656: We should never be using `$form_state['input']` (which is essentially the raw $_POST data) to determine values. This looks like a potential security hole. This should be updated to `$form_state['values']`.

core/modules/layout/layout.admin.inc
```diff
@@ -1814,6 +1824,21 @@ function layout_condition_add_form($form, &$form_state, Layout $layout = NULL, B
       'callback' => 'layout_condition_ajax_style',
     ),
   );
+
+  if ($is_new_condition && isset($form_state['input']['context']['required'])) {
+    foreach ($form_state['input']['context']['required'] as $required_context) {
```
quicksketch, line 1829: Switch `$form_state['input']` to `$form_state['values']` again here.

core/modules/layout/layout.admin.inc
```diff
@@ -1814,6 +1824,21 @@ function layout_condition_add_form($form, &$form_state, Layout $layout = NULL, B
...
+      $new_context_plugin = $required_context['plugin'];
+      switch ($new_context_plugin) {
+        case 'node':
+        case 'file':
+        case 'taxonomy_term':
+          $new_context = new EntityLayoutContext($new_context_plugin);
```
quicksketch, line 1835: This seems weird to special case these particular context plugins. This should be implemented in a way in which it allows all context plugins or has a way of determining which context plugins will work with this approach.

quicksketch (changes requested):
I think this may have some security concerns. It also will need new test coverage.

core/modules/layout/layout.admin.inc
jayelless: The patch works fine with this suggested change as the info needed is present in $form_state['values'] as well as $form_stater['input'];

core/modules/layout/layout.admin.inc
jayelless: This change will not work. The problem is that while the required value of "['context']['required']" is present in the $form_state['input'] array, it is not present in the $form_state['values'] array.

core/modules/layout/layout.admin.inc
jayelless: It should be possible to check the class type on the context plugin and switch on it being an entity context type.

jayelless: I find that the recent changes to core now prevent the patch in its original form working, as it is coming up with an error of 
`Notice: Undefined index: id in EntityLayoutContext->getAdminSummary() (line 100 of /home/james/backdrop/web/core/modules/layout/plugins/context/entity_layout_context_handler.inc). `
I will need to go back to investigation of this and revise the patch.

PR #3697, opened 2021-08-22
Issue #3520: Move add another item and dec text into the field value table.

### LINKED ISSUE #3520: [UX] "Add another item" buttons are hard to distinguish from the field label
This was reported by @bradbulger in #3516 along with other, related issues. Separating it to its own issue here:

> These started as comments from a client, I largely agree though. The overall point is that once the content types get to a certain level of complexity, the node forms begin to look awkward in the default Seven styles. Buttons and other elements that are there for the user to perform some kind of action aren't always easy to see and tell what they are.
>
>For instance, on a text field with unlimited values, the "ADD ANOTHER ITEM" button is the same size, shape, and color as the field title. ... If you have several of these fields in a form, it gets confusing visually.

screen shot 2019-02-01 at 4 08 13 pm

D8 for comparison:
screen shot 2019-02-01 at 4 07 40 pm

jenlampton:
Fixes https://github.com/backdrop/backdrop-issues/issues/3520

### DIFF
--- a/core/modules/field/css/field.css
+++ b/core/modules/field/css/field.css
@@ -29,6 +29,6 @@ form .field-multiple-table .field-multiple-drag .tabledrag-handle {
 [dir="rtl"] form .field-multiple-table .field-multiple-drag .tabledrag-handle {
   padding-left: .5em;
 }
-form .field-add-more-submit {
-  margin: .5em 0 0;
+form input.field-add-more-submit {
+  margin: 0;
 }
--- a/core/themes/seven/css/style.css
+++ b/core/themes/seven/css/style.css
@@ -1032,7 +1032,7 @@ button.button-primary:focus {
 }
 a.button-secondary,
 input.button-secondary,
-button.button-secondary{
+button.button-secondary {
   color: #828188;
   background-color: white;
   -webkit-box-shadow: inset 0px 0px 0px 2px #97979e;
--- a/core/themes/seven/css/style.css
+++ b/core/themes/seven/css/style.css
@@ -1032,7 +1032,7 @@ button.button-primary:focus {
 }
 a.button-secondary,
 input.button-secondary,
-button.button-secondary {
+button.button-secondary{
   color: #828188;
   background-color: white;
   -webkit-box-shadow: inset 0px 0px 0px 2px #97979e;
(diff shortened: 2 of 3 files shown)


core/modules/field/css/field.css
```diff
@@ -29,6 +29,7 @@ form .field-multiple-table .field-multiple-drag .tabledrag-handle {
 [dir="rtl"] form .field-multiple-table .field-multiple-drag .tabledrag-handle {
   padding-left: .5em;
 }
-form .field-add-more-submit {
-  margin: .5em 0 0;
+form input.field-add-more-submit {
+  margin: 0;
+  margin-bottom: 0;
```
olafgrabienski: Stumbled by accident on this. In the sandbox, the `margin` is enough, `margin-bottom` doesn't seem necessary to me.

jenlampton: The `margin-bottom` was added because `input.form-submit` has a `margin-bottom: 1em;` that does not get overridden by the `margin: 0;`.

jenlampton: I just tested this in the sandbox, and removing the `margin-bottom` didn't seem to change anything -- even with CSS aggregation disabled. But I added it because I was still seeing the bottom margin after adding `margin: 0;`. Maybe something else changed since then, and it's not an issue anymore. (though in the inspector, it looks like it is?) I'll push an update to the PR and we can test again :)

olafgrabienski: Thanks for the update! Tested in Chrome, looks good to me: The `margin-bottom: 1em;` gets overridden by the `margin: 0;`, both in the inspector (see screenshot) and on the page.

backdrop-add-another-submit-margin

core/themes/seven/css/style.css
```diff
@@ -1032,7 +1032,7 @@ button.button-primary:focus {
 }
 a.button-secondary,
 input.button-secondary,
-button.button-secondary {
+button.button-secondary{
```
indigoxela: This appears to be an unintentional white space removal.

olafgrabienski: @jenlampton did you see this? Looks like we only need one whitespace fix to get a RTBC for this important PR.

klonos:
```suggestion
button.button-secondary {
```

Adding this as a code suggestion. A core committer should be able to add it as a commit before merging this PR.

klonos (approved):
LGTM 👍🏼 (apart from that whitespace)

laryn:
@jenlampton Did this PR get closed/deleted without merging?

jenlampton:
oh jeez, sorry! restoring branch now :)

laryn:
Looks great -- a nice visual improvement. Thanks @jenlampton for the fix, and all the teamwork here in the PR and in the issue itself by @klonos, @herbdool, @indigoxela, @olafgrabienski, @quicksketch, @stpaultim, and @oadaeh.

PR #3698, opened 2021-08-23
Issue #5065: Context validation error on Configure Layout page and WSOD on cancel

### LINKED ISSUE #5065: Context validation error on Configure Layout page and empty page output on cancel
A couple of errors can be encountered when adding/configuring layouts. 

1. Go to Add layout. Pick any name and template.
2. Set the path to `node/%`. Context appears and is set to type Node.
3. Change the path to `foobar/%`. Context switches to "String pass-through".
4. Select context type of Node again from the type selector.
5. Click "Create Layout" to save the layout.
6. Click "Configure Layout" tab to re-configure it. (Note that the context is still set to "Node".)
7. Click "Add visibility condition"
8. Select visibility condition of "Node: Type".
8. Select a node type, e.g., "Page".
9. Click "Add visibility condition".
10. Click "Save Layout".
11.  This produces an error `The condition 'Type is <em class="placeholder">Page</em>' does not have the required contexts at this path. Remove the condition to save the layout.`.
12. Hit the cancel button, and get WSOD at path /system/ajax. Watchdog error is `Notice: Undefined index: #ajax in ajax_form_callback() (line 393 of <snip>/core/includes/ajax.inc).`

The extra HTML in the form error is pretty clearly because the condition summary is getting double-sanitized, so that should be an easy fix.

Don't know why the WSOD is happening, but guessing that it might be related to the question of whether one should be able to put a Node context on a bare % placeholder (whose value could well be an invalid nid) and how rigorously that is enforced.

jayelless:
Fixes https://github.com/backdrop/backdrop-issues/issues/5065

### DIFF
--- a/core/modules/layout/plugins/access/layout_access.inc
+++ b/core/modules/layout/plugins/access/layout_access.inc
@@ -41,7 +41,7 @@ class LayoutAccess extends LayoutHandler {
    * @param $contexts
    *   An array containing available contexts.
    * @return string
-   *   The summary string for this access check.
+   *   The sanitized HTML summary string for this access check.
    */
   function summary() {
     return '';
--- a/core/modules/layout/layout.admin.inc
+++ b/core/modules/layout/layout.admin.inc
@@ -661,7 +661,7 @@ function layout_settings_form_validate(&$form, &$form_state) {
   foreach ($layout->conditions as $key => $condition) {
     $required_contexts = $condition->getRequiredContexts();
     if (array_diff($required_contexts, $context_plugins)) {
-      form_error($form['conditions']['active'][$key], t('The condition \'@condition\' does not have the required contexts at this path. Remove the condition to save the layout.', array('@condition' => $condition->summary())));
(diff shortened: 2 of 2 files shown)


core/modules/layout/plugins/access/layout_access.inc
```diff
@@ -41,10 +41,10 @@ class LayoutAccess extends LayoutHandler {
    * @param $contexts
    *   An array containing available contexts.
    * @return string
-   *   The summary string for this access check.
+   *   The sanitized HTML summary string for this access check.
    */
   function summary() {
-    return '';
+    return t('');
```
bugfolder: You don't need to translate empty strings. So `return '';` was correct.

jayelless: Strange however.  Removing the t() call results in one of the date-theme tests failing, but restoring the t() call allows all tests to pass???

klonos (approved):
LGTM 👍🏼

quicksketch:
Merged into 1.x and 1.20.x. Thanks @jayelless!

PR #3701, opened 2021-08-25
Issue #5168: Translated string breaks db update

### LINKED ISSUE #5168: Database update fails when Backdrop is set to another language than English and the string "Apply pending updates" has a translation
**Steps To Reproduce**

1. Install an outdated Backdrop (old modules should also do the trick) with pending db/config updates
2. Add another language (e.g. German)
3. Make that language the _default_ one, disable English
4. Translate "Apply pending updates" - anything that's different from the original string will do
5. Run updates

You'll end up with a white (empty) page in step "Run updates", but that's no WSOD - just a page with zero content.

<hr />
Original report:

**Description of the bug**

When you're using Backdrop CMS with another language than Englisch and you're trying to update the database (core tables) you'll get a white screen. The log files are empty.

**Steps To Reproduce**

To reproduce the behavior:
1. Add another language (e.g. German)
2. Set your language to default
3. Disable English language
4. Update core
5. Run update.php

**Actual behavior**

You'll get a white screen. Enable English language and try again update.php works.

- Backdrop CMS version: 1.19.3 

indigoxela:
Fixes https://github.com/backdrop/backdrop-issues/issues/5168

### DIFF
--- a/core/update.php
+++ b/core/update.php
@@ -544,7 +544,7 @@ if (update_access_allowed()) {
         break;
       }
 
-    case 'Apply pending updates':
+    case t('Apply pending updates'):
       if (isset($_GET['token']) && backdrop_valid_token($_GET['token'], 'update')) {
         // Generate absolute URLs for the batch processing (using $base_root),
         // since the batch API will pass them to url() which does not handle


indigoxela:
Requesting fresh sandbox (subtext oh-oh Zen.CI)...

laryn:
Thanks for the fix @indigoxela and @oadaeh -- merged into 1.x and 1.19.x.

PR #3707, opened 2021-08-26
Issue #5180: Create a field_storage_exists() function

### LINKED ISSUE #5180: Create a field_storage_exists() function
**Description of the need**

In https://github.com/backdrop/backdrop-issues/issues/5170 We described how Backdrop allows field configs to exist without a corresponding field data table. A consequence of this situation is also discovered in https://github.com/backdrop/backdrop-issues/issues/3224#issuecomment-903198426 where if we attempt to save an entity with an attached field and instance, we get an exception.

**Proposed solution**

This new function would allow Backdrop to check if any storage method described in a field has been set up for that field. Currently in core, we only have `field_sql_storage` as a method

We could also attempt to create storage automatically upon trying to save any attached field, in addition to this check

docwilmot:
Fixes https://github.com/backdrop/backdrop-issues/issues/5180


core/modules/field/field.api.php
```diff
@@ -1680,6 +1680,8 @@ function hook_field_attach_delete_bundle($entity_type, $bundle, $instances) {
  *   key/value pairs:
  *   - label: The human-readable name of the storage backend.
  *   - description: A short description for the storage backend.
+ *   - validate: The name of a function which returns true if the storage
```
klonos, line 1683: This seems like an API change, so can we please add an `@since` to this docblock?

core/modules/field/field.info.inc
```diff
@@ -957,6 +957,45 @@ function field_info_storage_settings($type) {
...
+ * Returns whether a field's storage exists.
+ *
+ * @param string $field_name
+ *   The field name.
+ *
+ * @return
+ *   TRUE if the storage exists and is writable or if a validate key is not
+ *   defined in the field storage type's hook_field_storage_info().
+ */
+function field_storage_exists($field_name) {
```
klonos, line 970: Same here, since this is an API addition, can we please add an `@since`?

core/modules/field/field.info.inc
```diff
@@ -957,6 +957,45 @@ function field_info_storage_settings($type) {
   return isset($info['settings']) ? $info['settings'] : array();
 }
 
+/**
+ * Returns whether a field's storage exists.
+ *
+ * @param string $field_name
+ *   The field name.
+ *
+ * @return
```
klonos, line 966:
```suggestion
 * @return bool
```

PR #3708, opened 2021-08-27
Issue #3008: Fix module updates with symlinks

### LINKED ISSUE #3008: Fixed: Symlink to document root causing several update and install issues

indigoxela:
Fixes https://github.com/backdrop/backdrop-issues/issues/3008

Alternate approach with less impact, as the change in CoreUpdater, ModuleUpdater, ThemeUpdater and LayoutUpdater does not interfere with other FileTransfers.

### DIFF
--- a/core/modules/system/system.updater.inc
+++ b/core/modules/system/system.updater.inc
@@ -80,7 +80,7 @@ class ModuleUpdater extends Updater {
     else {
       $relative_path = 'modules';
     }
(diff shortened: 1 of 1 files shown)


laryn (approved):
LGTM -- but didn't investigate all the failing tests completely yet. Are those all the ones being addressed in separate issues?

quicksketch:
Merged into 1.x and 1.19.x!

PR #3722, opened 2021-09-05
Issue #5198: Insufficient access to Taxonomy Block on Dashboard

### LINKED ISSUE #5198: Insufficient access to Taxonomy Block on Dashboard
**Description of the bug**

With the addition of #382, we need to refine access to the "Taxonomy" block on the dashboard for roles to create or delete terms. 

Currently, only those with permission to administer vocabularies or edit terms have access to this block, when this block is actually useful for anyone with permission to do anything with taxonomies. This was not the case prior to #382. 

ALSO - user with "Administer Terms and Vocabularies" permission had a link to "configure" but it took them to the "list terms" page instead of the "configure" page. 

**Steps To Reproduce**

To reproduce the behavior:
1. Login to a new site as site editor (with default permission "Tags: Create terms")
2. Visit the dashboard and see that the taxonomy block is not visible. 
3. Add "Tags: Edit terms" permission and view dashboard again and you will see that the block is available. 

**Actual behavior**

Even with edit and delete terms, these are the only options you get with the block. 

dashboard-terms

**Expected behavior**

Since the only two links that are provided by this block directly are "List Terms" and "Add New Terms", this block is useful to editors with any taxonomy permission. The ability to see available taxonomies and list the terms is valuable. 

@olafgrabienski and @herbdool

stpaultim:
Fixes backdrop/backdrop-issues#5198

### DIFF
--- a/core/modules/dashboard/includes/block.taxonomy.inc
+++ b/core/modules/dashboard/includes/block.taxonomy.inc
@@ -74,6 +74,7 @@ class DashboardTaxonomyBlock extends Block {
         $delete_terms = user_access('delete terms in ' . $vocabulary->machine_name);
         if ($administer_taxonomy || $create_terms || $edit_terms || $delete_terms) {
           $terms = format_plural($term_count, '1 term', '@count terms');
+          $links = array();
           $links['list'] = array(
             'title' => t('List terms'),
             'href' => 'admin/structure/taxonomy/' . $vocabulary->machine_name,
@@ -104,7 +105,6 @@ class DashboardTaxonomyBlock extends Block {
               backdrop_render($operations),
             ),
           );
-          unset($links);
         }
         else {
           $no_access = TRUE;
--- a/core/modules/dashboard/includes/block.taxonomy.inc
+++ b/core/modules/dashboard/includes/block.taxonomy.inc
@@ -68,19 +68,23 @@ class DashboardTaxonomyBlock extends Block {
     foreach ($vocabularies as $machine => $vocabulary) {
       if (in_array($machine, $this->settings['vocabularies']) || empty($this->settings['vocabularies'])) {
         $term_count = db_query("SELECT count(*) FROM {taxonomy_term_data} WHERE vocabulary = :name", array(':name' => $vocabulary->machine_name))->fetchField();
-        if (user_access('administer taxonomy') || user_access('create terms in ' . $vocabulary->machine_name) || user_access('edit terms in ' . $vocabulary->machine_name) || user_access('delete terms in ' . $vocabulary->machine_name)) {
+        if (user_access('administer taxonomy') || 
+          user_access('create terms in ' . $vocabulary->machine_name) || 
+          user_access('edit terms in ' . $vocabulary->machine_name) || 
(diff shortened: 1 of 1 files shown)


stpaultim:
Need feedback on 

1) Lines 71-74. I had one really long line, wasn't sure if I should focus max line length or shorter breaks for better readability. 

2) Are my nested "if" statements ok

3) I had to "unset($links)" (line 107) to avoid problems with inaccurate values carrying over from one  vocabulary to the next. This "unset" works, but I'm not sure if its the right way to do this.

klonos:
@stpaultim the php5.3 failures are unrelated to this change. See https://github.com/backdrop/backdrop-issues/issues/5200

An idea to tackle this in a slightly better way (less calls to the `user_access()` function), would be to assign those things to variables. For example:

```php
$administer_taxonomy = user_access('administer taxonomy');
$create_terms = user_access('create terms in ' . $vocabulary->machine_name);
$edit_terms = user_access('edit terms in ' . $vocabulary->machine_name);
$delete_terms = user_access('delete terms in ' . $vocabulary->machine_name);
```

...and then do something like this:

```php
if ($administer_taxonomy || $create_terms || $edit_terms || $delete_terms) {
  ...
}
if ($administer_taxonomy || $create_terms) {
  ...
}
if ($administer_taxonomy) {
  ...
}
```

Re the nested `if`'s, it should be OK, but from a quick look it seems that `user_access('administer taxonomy')` seems to be a common conditional for all of them. So perhaps group them all inside that, instead of repeatedly making that check. Combine that with the previous suggestion, and you'd get something like this instead (untested):

```php
if ($administer_taxonomy) {
  $terms = format_plural($term_count, '1 term', '@count terms');
  if ($create_terms) {
    if ($edit_terms || $delete_terms) {
      $links['list'] = array(
        ...
      );
    }
    $links['add'] = array(
      ...
    );
  }
  $links['configure'] = array(
    ...
  );
  $operations = array(
    ...
  );
  $rows[] = array(
    ...
  );
}
```

stpaultim:
@klonos - I like the variable idea and updated the PR with that change. How does the code look now?

core/modules/dashboard/includes/block.taxonomy.inc
```diff
@@ -98,6 +104,7 @@ class DashboardTaxonomyBlock extends Block {
               backdrop_render($operations),
             ),
           );
+          unset($links);
```
klonos: Instead of `unset()`ing `$links` here, can we add a `$links = array();` after `$terms = ...`?

stpaultim: Done!! @klonos

stpaultim:
> from a quick look it seems that user_access('administer taxonomy') seems to be a common conditional for all of them. So perhaps group them all inside that, instead of repeatedly making that check.

@klonos - For the record, I considered this, but I don't THINK that the logic works. I think we need to check administer taxonomy each time because they are "OR" conditionalites. I could easily be wrong and there might be a more effecient way to handle this. I look forward to your revised PR because I don't write a lot of code like this, so I may be missing some obvious approaches that are better.

PR #3724, opened 2021-09-05
Issue #4903: Alternative version with Cards content type

### LINKED ISSUE #4903: Provide a new hidden-path ("Card") content type in core
**Description of the need**

With 'existing content' blocks and  'hidden-path' content, Backdrop provides a powerful alternative to custom blocks: fieldable, translatable as any other content, and quite editor-friendly. I've got however a notion that many people aren't aware of its potential. Let's make the approach more visible and integrate it better into Backdrop by adding a new 'hidden-path' content type to Backdrop core.

**Proposed solution**

Add a third content type to Backdrop, and configure it for the use of existing content blocks. Display an example of such a content block, e.g. on the home page (see https://github.com/backdrop/backdrop-issues/issues/4849 and https://github.com/backdrop/backdrop-issues/issues/4867).

1. Create a custom content type on my own sites. Works fine *for me* but doesn't make the approach more visible.

2. Build a contributed module with such a content type. Should be possible and might make the approach more visible. It couldn't be used in a vanilla Backdrop installation, though.

Backdrop now includes a powerful alternative to custom blocks: the flexible and editor-friendly Promobox content type.

---

Advocate: @stpaultim 

--- 

PR: ~https://github.com/backdrop/backdrop/pull/3700~
PR: https://github.com/backdrop/backdrop/pull/3862

stpaultim:
Fixes backdrop/backdrop-issues#4903


core/profiles/standard/standard.install
```diff
@@ -171,14 +192,78 @@ function standard_install() {
     ->set('node_post_pattern', 'posts/[node:title]')
     ->save();
 
-  // Create an initial post to welcome users on install.
+  // Create three default Promobox nodes.
```
klonos: If we are to call these "Cards", then this inline comment should also be updated to reflect that:

```suggestion
  // Create three default Card nodes.
```

stpaultim:
OK, I'm stuck now trying to import images. I've done it in another module: https://github.com/backdrop-contrib/sample_animal_content

Starting on line 371 - I tried what I used in the Sample Animal Content, because I was not able to get the second option to work. The second option on line 410 is more how I think it SHOULD be done, but it's not working there either. 

My code DOES sucessfully make a copy of the file in the files/field/image directory. AND - clearly Backdrop thinks that BOTH nodes have images, but none of the data shows up in the database. I suspect that I'm creating a database record for the image, just not populating  it with the right info. 

Any help? Can anyone point me at some better sample code?

indigoxela:
>  Can anyone point me at some better sample code?

You're using [file_unmanaged_save_data()](https://docs.backdropcms.org/api/backdrop/core%21includes%21file.inc/function/file_unmanaged_save_data/1) which does not create a file record in table.
The docs say:
"This function is identical to file_save_data() except the file will not be saved to the {file_managed} table and none of the file_* hooks will be called."
Does that help?

klonos:
@stpaultim if using `file_save_data()` directly doesn't work after all, perhaps you could try using `system_retrieve_file()` instead (it also uses `file_save_data()` under the hood, but also seems to be doing some other "magic").

@bugfolder asked a similar question in Zulip re how to create a file entity from local file for a module, and he mentioned that that worked for him: https://backdrop.zulipchat.com/#narrow/stream/218635-Backdrop/topic/create.20File.20entity.20from.20local

bugfolder:
@stpaultim, here's the solution I used (based on @klonos's suggestion in the Zulip chat linked above) for the demo module for Views Bootstrap 1.x-3.5. See `views_bootstrap_demo_create_nodes()`, which creates some nodes with images, where the sample images are stored in the module. https://github.com/backdrop-contrib/views_bootstrap/blob/1.x-3.x/views_bootstrap_demo/views_bootstrap_demo.module

stpaultim:
@klonos and @bugfolder - For the record, it took a little experimenting, but I got the code from views_bootstrap to work. I have not pushed it up yet, because it's a mess. 

But, I think I'm "unblocked" for a few minutes. Got to sleep. Will try again to make progress tomorrow.

PR #3483, opened 2021-01-06
Issue #4862: Fix PathPatternBulkUpdateTestCase

klonos:
Fixes https://github.com/backdrop/backdrop-issues/issues/4862

PR #3733, opened 2021-09-11
Issue #4862: Fix PathPatternBulkUpdateTestCase.

### LINKED ISSUE #4862: Random test failures with PathPatternBulkUpdateTestCase
This is part of the #1478 meta.

The 2 failures that we are usually getting are:
- `"Generated 20 URL aliases." found path_pattern.test:929`
- `"Generated 1 URL alias." found path_pattern.test:943`

I've had enough with this 😅 ...so I took a look at the code, and have noticed some inconsistencies in it. Perhaps addressing these inconsistencies will also fix these failures.

quicksketch:
Fixes https://github.com/backdrop/backdrop-issues/issues/4862.

Replaces https://github.com/backdrop/backdrop/pull/3483

### DIFF
--- a/core/includes/batch.inc
+++ b/core/includes/batch.inc
@@ -447,17 +447,16 @@ function _batch_finished() {
         $queue->deleteQueue();
       }
     }
+    // Clean-up the session. Not needed for CLI updates.
+    if (isset($_SESSION)) {
+      unset($_SESSION['batches'][$batch['id']]);
+      if (empty($_SESSION['batches'])) {
+        unset($_SESSION['batches']);
+      }
+    }
   }
   $_batch = $batch;
   $batch = NULL;
-
-  // Clean-up the session. Not needed for CLI updates.
-  if (isset($_SESSION)) {
-    unset($_SESSION['batches'][$_batch['id']]);
-    if (empty($_SESSION['batches'])) {
-      unset($_SESSION['batches']);
-    }
-  }
 
   // Redirect if needed.
   if ($_batch['progressive']) {
--- a/core/modules/path/path.admin.inc
+++ b/core/modules/path/path.admin.inc
@@ -858,8 +858,7 @@ function path_bulk_update_batch_finished($success, $results, $operations) {
     }
   }
   else {
-    $error_operation = reset($operations);
-    backdrop_set_message(t('An error occurred while processing @operation with arguments : @args', array('@operation' => $error_operation[0], '@args' => print_r($error_operation[0], TRUE))));
+    backdrop_set_message(t('An error occurred while generating the URL aliases.'), 'error');
   }
 }
 
--- a/core/scripts/run-tests.sh
+++ b/core/scripts/run-tests.sh
@@ -970,7 +970,7 @@ function simpletest_script_print_error($message) {
  */
 function simpletest_script_print($message, $status) {
   global $args;
-  if ($args['color']) {
+  if (isset($args['color']) && $args['color']) {
     $color_code = simpletest_script_color_code($status);
     $message = "\033[" . $color_code . "m" . $message . "\033[0m";
   }
--- a/core/modules/path/tests/path_pattern.test
+++ b/core/modules/path/tests/path_pattern.test
@@ -900,66 +900,68 @@ class PathPatternLocaleTestCase extends PathPatternFunctionalTestHelper {
 class PathPatternBulkUpdateTestCase extends PathPatternFunctionalTestHelper {
   protected $profile = 'standard';
   private $nodes;
+  private $terms;
+  private $users;
 
   function testBulkUpdate() {
     $this->nodes = array();
-    // Add a path pattern on the post content type.
-    $edit = array(
-      'path_pattern' => 'post/[node:title]',
-    );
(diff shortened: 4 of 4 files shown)


core/modules/path/path.admin.inc
```diff
@@ -801,13 +801,7 @@ function path_bulk_update_form_generate_submit($form, &$form_state, $op = 'gener
       $path_info['choices'] = empty($choices['base']) ? $choices : array();
       $path_info['op'] = $op;
       $callback = $path_info['batch update callback'];
-
-      if (!empty($path_info['batch file'])) {
-        $batch['operations'][] = array('path_bulk_update_batch_process', array($callback, $path_info));
-      }
-      else {
```
indigoxela: Does this mean, that this `else` never strikes, because all path bulk jobs have a `'batch file'` set?

Function path_get_info() adds "module" and "batch file path" to all implementations of hook_path_info(), but no "batch file". Could this potentially break something, or is that check in function path_bulk_update_form_generate_submit() redundant and really only needed in path_bulk_update_batch_process() to include the right file?

(Obviously I'm no specialist regarding path pattern batches. :wink:)

core/scripts/run-tests.sh
```diff
@@ -970,7 +970,7 @@ function simpletest_script_print_error($message) {
  */
 function simpletest_script_print($message, $status) {
   global $args;
-  if ($args['color']) {
+  if (isset($args['color']) && $args['color']) {
```
indigoxela, line 973: I belief, this change is not necessary, because "color" get's a default value (FALSE) in function simpletest_script_parse_args(), anyway. To my understanding it's always set.

core/modules/path/path.admin.inc
```diff
@@ -858,8 +852,7 @@ function path_bulk_update_batch_finished($success, $results, $operations) {
     }
   }
   else {
-    $error_operation = reset($operations);
-    backdrop_set_message(t('An error occurred while processing @operation with arguments : @args', array('@operation' => $error_operation[0], '@args' => print_r($error_operation[0], TRUE))));
+    backdrop_set_message(t('An error occurred while generating the URL aliases.'), 'error');
```
indigoxela, line 861: IMO it totally makes sense to *not* assume that in case of failure $operations is a non-empty array.

But maybe some watchdog() could be helpful? Or is this overdone?

core/modules/path/path.admin.inc
quicksketch: I'm no expert either but by the looks of things `path_bulk_update_batch_process()` _also_ checks for the existent of a `'batch_file'` key. However, I was just trying things to see if I could "clean up" and get tests passing consistently. Now I'm sure this isn't related so I'll revert it back to what it was.

core/scripts/run-tests.sh
quicksketch: I found that `$args` is not populated with the defaults if you pass an argument that doesn't exist. I think it's because `simpletest_script_parse_args()` itself calls this function while it's doing the arg parsing. Then you get a PHP notice that `$args['color']` is not set. e.g. `./run-scripts.php --non-existent-setting`

core/includes/batch.inc
```diff
@@ -447,18 +447,17 @@ function _batch_finished() {
...
   $_batch = $batch;
   $batch = NULL;
 
-  // Clean-up the session. Not needed for CLI updates.
-  if (isset($_SESSION)) {
-    unset($_SESSION['batches'][$_batch['id']]);
-    if (empty($_SESSION['batches'])) {
-      unset($_SESSION['batches']);
-    }
-  }
```
quicksketch, line 460: This just brings our implementation inline with Drupal's approach from https://www.drupal.org/i/3084945

core/modules/path/path.admin.inc
quicksketch: I removed the printing of `$error_operation` because in our failures, that value was NULL and causing a PHP notice by trying to access it as an array.

My thinking is that if failures occurred, other PHP notices/warnings or watchdog calls would have happened where the database failed to write or an alias already existed, etc.

indigoxela: > My thinking is that if failures occurred, other PHP notices/warnings or watchdog calls would have happened...

Makes sense. So let's stick with that simple message here. :+1:

core/scripts/run-tests.sh
indigoxela: Odd.

simpletest_script_parse_args() does call simpletest_script_print_error() when an argument doesn't exists. But it does it after the defaults array was set. Possibly there's something wrong with argument parsing in general?

Anyway, that single isset() won't hurt. :+1:

quicksketch: > But it does it after the defaults array was set

`$args` is set as a root level variable, but only after `simpletest_script_parse_args()` finishes. Inside of the function it's only local. Admittedly the whole way arguments are handled in that script could be a lot better. Really this isn't related at all, it's just something I ran across testing all these random failures recently.

core/modules/path/path.admin.inc
quicksketch: This has been reverted.

quicksketch:
I merged this into 1.x and 1.20.x. Thanks @indigoxela for reviewing!

PR #3734, opened 2021-09-11
Issue #4862: Attempt to fix random failures in batch operations.

quicksketch:
A different approach at trying to fix some of our random test failures.

This tries something similar to https://www.drupal.org/project/drupal/issues/2828559, which discovered a weird issue with random failures caused by PHP's Garbage Collection occurring when running update.php. This may be similar to our problem with tests that use Batch API (like PathPatternBulkUpdateTestCase) failing intermittently.

Relates to https://github.com/backdrop/backdrop-issues/issues/4862


quicksketch:
This approach has not seemed to help with the random failures. Closing.

PR #3736, opened 2021-09-12
Issue #5098: Indicate new dashboard notifications based on timestamp

### LINKED ISSUE #5098: Highlight "New" Dashboard Notifications

klonos:
Fixes https://github.com/backdrop/backdrop-issues/issues/5098

Alternative to https://github.com/backdrop/backdrop/pull/3633, which uses a timestamp to track if the user has seen the notifications before.


klonos:
Tugboat sandbox failed to be created...

klonos:
Feedback was that this approach is not acceptable. Closing.

PR #3737, opened 2021-09-12
Issue #5198: Better Taxonomy Permissions for the Dashboard taxonomy block

### LINKED ISSUE #5198: Insufficient access to Taxonomy Block on Dashboard

klonos:
Fixes https://github.com/backdrop/backdrop-issues/issues/5198

Alternative to https://github.com/backdrop/backdrop/pull/3722

### DIFF
--- a/core/modules/dashboard/includes/block.taxonomy.inc
+++ b/core/modules/dashboard/includes/block.taxonomy.inc
@@ -54,69 +54,60 @@ class DashboardTaxonomyBlock extends Block {
       return;
     }
 
-    $vocabularies = taxonomy_get_vocabularies();
-    $current_path = current_path();
-    $options = array('destination' => $current_path);
+    $allowed_vocabularies = !empty($this->settings['vocabularies']) ? $this->settings['vocabularies'] : FALSE;
+    // Passing FALSE to taxonomy_vocabulary_load_multiple() loads all
+    // vocabulaies.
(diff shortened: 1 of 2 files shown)


laryn:
Thank you @stpaultim and @klonos -- this has been merged into 1.x for 1.20!

PR #3739, opened 2021-09-13
Issue #5215: Fix filter_dialog test syntax on older php versions

### LINKED ISSUE #5215: Fixed: Simpletest FilterEditorLinkValidateTestCase silently broken on PHP 5.3
It might have been broken for a while, but nobody realized because it doesn't produce an actual test failure.

Only on PHP 5.3:

```
PHP Parse error:  syntax error, unexpected ')' in /home/runner/work/backdrop/backdrop/core/modules/filter/tests/filter_dialog.test on line 26
ERROR: FATAL FilterEditorLinkValidateTestCase: test runner returned a non-zero error code (255).
```

When reading "syntax error" in 5.3 my first thought is the short array syntax, but there doesn't seem to be any (we fixed one recently).

Here's an example: https://github.com/backdrop/backdrop/runs/3585091100?check_suite_focus=true#step:9:287

Note: no test failure.

indigoxela:
Fixes https://github.com/backdrop/backdrop-issues/issues/5215

### DIFF
--- a/core/modules/filter/tests/filter_dialog.test
+++ b/core/modules/filter/tests/filter_dialog.test
@@ -22,7 +22,7 @@ class FilterEditorLinkValidateTestCase extends BackdropWebTestCase {
       'ckeditor',
       'filter_formtest',
       'language',
-      'locale',
+      'locale'
     );
 
     // Add a language and set it as default.


quicksketch:
Merged into 1.x and 1.19.x. Thanks @indigoxela!

PR #3746, opened 2021-09-17
Issue #5235: Change multiple value button text to 'Add another'.

### LINKED ISSUE #5235: The `Add Another Item` Button on multiple value field form is too specific
**Description of the bug**

I often use fields to describe things that are not items, so having the button read `Add another item` seems very strange. In today's example I am using this on a field for "Authors".

I would like to prose we change the button text to `Add another` which is likely to work for all fields, rather than only those containing items.

jenlampton:
Fixes https://github.com/backdrop/backdrop-issues/issues/5235


klonos (approved):
LGTM 👍🏼

quicksketch:
Merged into 1.x for 1.21.0.

PR #3492, opened 2021-01-12
Issue #4868: Add collapsible help in the layout listing

klonos:
Fixes https://github.com/backdrop/backdrop-issues/issues/4868

PR #3754, opened 2021-09-25
Issue #4868: Add collapsible help in the layout listing

### LINKED ISSUE #4868: [UX] Layout listing: Add help text, to explain how the order and grouping in that page works
This is related to #4841 and was brought up during the Dec. 31, 2020 weekly UX meeting, https://youtu.be/Fc-oQ55LQi0?t=2190 (~37min into the call)

@philsward (referring to the layout listing page) explained:

> ...just coming and looking at it, I feel that everything is jumbled together ...I don't know what I'm looking at. Nothing is organized, and that was kinda my big beaf with it.

@jenlampton said:

> ...everything _is_ organized, but it's not clear to anyone looking at the page that it is. ...it's confusing.

Then @stpaultim asked a question about the organization of things:

> ...I've always understood that the higher up a layout is on the list, the more priority it has.

...then @jenlampton went on, to explain how the grouping and priority works on the layout listing page.

@cellear:

> I just learned 30sec ago that the order in this page matters. I had no idea.

@stpaultim:

> It doesn't say that anywhere. ...the only reason I think I know that is I think I've heard @jenlampton say it, but I might have misunderstood.

@jenlampton 

> ...and if you have 2 things that _can_ be ordered, a button shows up that says "Reorder". But if you don't ever have them, you would never know that. ...if you never create anything that needs an order, you'll never even know that there _are_ things that can have an order.

This is what @jenlampton was referring to:

Screen Shot 2021-01-12 at 6 04 08 pm

---

## Proposed solution:

Screen Shot 2022-01-03 at 2 37 35 PM

bugfolder:
Fixes https://github.com/backdrop/backdrop-issues/issues/4868.

This is a possible alternative PR to PR https://github.com/backdrop/backdrop/pull/3492.

### DIFF
--- a/core/modules/layout/layout.admin.inc
+++ b/core/modules/layout/layout.admin.inc
@@ -94,22 +94,21 @@ function layout_list_page() {
     array('data' => t('Operations'), 'class' => array('layout-operations')),
   );
 
+  $help = '<p>' . t('As a page is rendered, its <em>Path</em> is compared to those listed here. If a match is found, <em>Visibility conditions</em> will be evaluated to determine which <em>Layout</em> is used. If no match is found, or if visbiility conditions are not met, the page will use one of the <em>Default layouts</em> shown at the bottom of this page.') . '</p>';
+
   $help_items = array();
-  $help_items[] = t('To determine which layout is to be used for a page, the path is compared to those shown here.');
-  $help_items[] = t('Paths used by the layout system are <em>system paths</em> (also called <em>normal paths</em>), not !link.', array('!link' => l(t('URL aliases'), 'admin/config/urls/path')));
-  $help_items[] = t('They are listed alphabetically below. All layouts that might apply to each path are listed below the path.');
-  $help_items[] = t('All pages without a layout explicitly defined for its path will use one of the <em>Default layouts</em> (shown at the bottom of this page).');
+  $help_items[] = t('Paths below are listed alphabetically.');
+  $help_items[] = t('All possible layouts for each path are listed beneath the path.');
   $help_items[] = array(
-    'data' => t('Each path listed here will have one or more layouts defined:'),
+    'data' => t('Each path may have one or more available layouts:'),
     'children' => array(
-      t('If any path has exactly one layout defined, then that layout will be used (if visibility conditions allow).'),
-      t('Otherwise, starting from top to bottom, all layouts for that path are checked for matching visibility conditions.'),
+      t('Starting from top to bottom, all layouts for that path are checked for matching visibility conditions.'),
       t('The first layout that matches all conditions will be the one that gets used.'),
       t('To adjust priority, you may reorder the layouts for a given path.'),
-      t('If no match is found, the page will use one of the Default layouts.'),
     ),
   );
-  $help = theme('item_list', array('items' => $help_items));
+  $help_items[] = t('Paths used by the layout system are <em>system paths</em> (example: <code>node/%</code>), not normal paths (example: <code>node/2</code>) or !link (example: <code>about</code>)', array('!link' => l(t('URL aliases'), 'admin/config/urls/path')));
+  $help .= theme('item_list', array('items' => $help_items));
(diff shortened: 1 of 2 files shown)


klonos (approved):
LGTM 👍🏼

core/modules/layout/layout.admin.inc
```diff
@@ -94,7 +94,34 @@ function layout_list_page() {
     array('data' => t('Operations'), 'class' => array('layout-operations')),
   );
 
+  $help_items = array();
+  $help_items[] = t('To determine which layout is to be used for a page, the path is compared to those shown here.');
+  $help_items[] = t('Paths used by the layout system are <em>system paths</em> (also called <em>normal paths</em>), not !link.', array('!link' => l(t('URL aliases'), 'admin/config/urls/path')));
+  $help_items[] = t('They are listed alphabetically below. All layouts that might apply to each path are listed below the path.');
```
klonos: "They"? ...which ones? ...the "system paths" or the "URL aliases"? ...or both? Let's please be explicit/specific here, to avoid confusion for people that are new to all of this.

core/modules/layout/layout.admin.inc
```diff
@@ -94,7 +94,34 @@ function layout_list_page() {
...
+  $help_items[] = t('All pages without a layout explicitly defined for its path will use one of the <em>Default layouts</em> (shown at the bottom of this page).');
```
klonos: - "pages" here is plural, but we later say "its path" (singular).
- Should the word "Default" be capitalized here? I think not.

core/modules/layout/layout.admin.inc
```diff
@@ -94,7 +94,34 @@ function layout_list_page() {
     array('data' => t('Operations'), 'class' => array('layout-operations')),
   );
 
+  $help_items = array();
+  $help_items[] = t('To determine which layout is to be used for a page, the path is compared to those shown here.');
```
jenlampton: > To determine which layout will be used for a given page, the path for that page is compared to those shown here.

We should put the point before the reason. `The path for the current page will be compared to those listed here to determine which layout should be used."

core/modules/layout/layout.admin.inc
jenlampton: I agree, `They` here should be `Paths`.  It would also be clearer if we put the disambiguation after the explanation so move `Paths used by the layout system are...` below `They are listed alphabetically...`.

core/modules/layout/layout.admin.inc
jenlampton: > "pages" here is plural, but we later say "its path" (singular).

I don't think we even need that part...  `All pages without a layout explicitly defined will use one of the ...` would be fine here.

> Should the word "Default" be capitalized here? I think not.

I don't mind the capital here, actually, because it matches the text in the interface.

core/modules/layout/layout.admin.inc
```diff
@@ -94,7 +94,34 @@ function layout_list_page() {
     array('data' => t('Operations'), 'class' => array('layout-operations')),
   );
 
+  $help = '<p>' . t('As a page is rendered, a <em>Layout</em> is selected as follows.') . '</p>';
+  $help_items = array();
+  $help_items[] = array(
+    'data' => 'The <em>Path</em> for the page is compared to those in the alphabetical list here, and <em>Visibility conditions</em> are evaluated.',
```
stpaultim: @jenlampton or @bugfolder - It seems like all of the text except for this is in a t(function). Is there a reason that this line is not?

bugfolder: Nope, good catch. It should be. Will make that change.

stpaultim:
@bugfolder - That was quick. And, I guess I did find something helpful. :-)

core/modules/layout/layout.admin.inc
bugfolder: I think this was addressed in a recent PR, so marking this as resolved.

quicksketch:

PR #3757, opened 2021-09-27
Issue #2626PR [UX] Add a contextual menu for quick access to editing things like the layout

### LINKED ISSUE #2626: [UX] Add links for quick access to editing things like the layout, theme, etc. of the current page
This thought just occurred to me as I glanced over a notification that came in on another thread ("...there does not seem to be a way of seeing exactly which layout is in action for a page, even by inspecting the code." [Source](https://github.com/backdrop/backdrop-issues/issues/2612#issuecomment-292567621))

What about a new little icon in the admin menu that either links to a separate page or has a little dropdown info pane with details about the current page you are viewing?

home___pr_1782_backdrop_backdrop_testing_site

For example, all in one place, info (and potentially links to edit/adjust) about the:

- Layout used on this page 
- Theme
- Content type
- etc.

That may be overkill but thought I'd throw it out there to see what people thought.

### LINKED ISSUE #3946: [UX] Add contextual link for editing the currently-used layout
**Description of the need**
As I am working on a page, I frequently find myself clicking on a "configure block" contextual link and then closing the block configuration to get to the layout itself. I can get directly to a node, view, or block by using contextual links, but I can not get to the layout directly. 

I think that this would be a nice improvement. 

If a site editor was ever confused about which layout was being used for a page (is that likely), they could use this to find the right layout. 

**Proposed solution**
Add a contextual menu in the upper right hand corner of every page that takes the site editor to the layout that is being used for that page. 

I'm usually looking for the "Manage Blocks" tab as opposed to the "Configure Layout" tab, so that would be my preference. 

@jenlampton said something similar here: https://github.com/backdrop/backdrop-issues/issues/1725#issuecomment-298144809
> There is one feature from Panels (Drupal 7) that we are missing in Backdrop that might serve the same goal, in a slightly less dangerous way: A contextual link on the layout itself.

bugfolder:
Revised PR that puts changes in `layout_preprocess_layout()`.

Fixes https://github.com/backdrop/backdrop-issues/issues/2626.
Fixes https://github.com/backdrop/backdrop-issues/issues/3946.

### DIFF
--- a/core/modules/contextual/css/contextual.css
+++ b/core/modules/contextual/css/contextual.css
@@ -101,14 +101,16 @@
 .contextual-links-wrapper a {
   text-decoration: none;
 }
-.contextual-links li a {
+.contextual-links li a,
+.contextual-links li .info {
   color: #333 !important;
   display: block;
   margin: 0.25em 0;
   padding: 0.25em 1em 0.25em 0.5em; /* LTR */
   text-indent: 0;
 }
-[dir="rtl"] .contextual-links li a {
+[dir="rtl"] .contextual-links li a,
+[dir="rtl"] .contextual-links li .info {
   padding: 0.25em 0.5em 0.25em 1em;
 }
 .contextual-links li a:hover {
--- a/core/modules/contextual/js/contextual.js
+++ b/core/modules/contextual/js/contextual.js
@@ -37,8 +37,12 @@ Backdrop.behaviors.contextualLinks = {
       // Prepend the trigger.
       $wrapper.prepend($trigger);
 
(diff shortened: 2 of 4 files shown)


bugfolder:
Closing this PR based on [decisions taken at 9/30 design meeting](https://github.com/backdrop/backdrop-issues/issues/2626#issuecomment-931743014). In brief:

* Near-term needs can be met by the [Layout Contextual Links contrib module](https://backdropcms.org/projects/layout_contextual_links)
* Eventual integration into core will be addressed by [Issue 2626](https://github.com/backdrop/backdrop-issues/issues/2626).

PR #3758, opened 2021-09-27
Issue #5260: Fixes block admin listing.

### LINKED ISSUE #5251: broken html at custom blocks admin page
Since I updated to 1.20.0, my custom blocks admin page is broken. It happens with all my updated sites which already had custom blocks in it, but NOT the ones without, even if I add them now:

Cursor_and_Custom_blocks___lluitanoviolenta_cat

This happens with seven but also with basis as admin theme.

jenlampton:
Fixes https://github.com/backdrop/backdrop-issues/issues/5251


bugfolder:
Reviewed and tested sandbox. LGTM.

PR #3759, opened 2021-09-29
Issue 5262: Visibility condition for node ID does not throw error when non-integer

### LINKED ISSUE #2668: [UX] Layout and Block visibility conditions: add validation when entering entity IDs to ensure that only numbers are allowed.
**Updated issue description**

---

1. Add a node, e.g. `node/7`.

### LINKED ISSUE #5262: Visibility condition for node ID does not throw error when non-integer
**Description of the bug**

I just watched someone enter node/4 into the ID box, and was able to save the block, and then save the layout. 

**Steps To Reproduce**

To reproduce the behavior:
1. Create a new layout at node/%
2. Add a new block to the right sidebar
3. Add a visibility condition 
4. Enter `node/1` into the ID box (instead of 1)
5. Save the block configuration
6. Save the layout configuration
7. Configure the block again -- the Node ID visibility condition has been changed to `0`

**Expected behavior**

The visibility conditions box should validate (or be a number field) so that non-integer values cannot be entered.

**Additional information**

bugfolder:
Fixes https://github.com/backdrop/backdrop-issues/issues/5262.
Fixes https://github.com/backdrop/backdrop-issues/issues/2668.

### DIFF
--- a/core/modules/layout/plugins/access/entity_id_layout_access.inc
+++ b/core/modules/layout/plugins/access/entity_id_layout_access.inc
@@ -27,7 +27,8 @@ class EntityIDLayoutAccess extends LayoutAccessNegatable {
 
     $form['entity_id'] = array(
       '#title' => t('@entity @entity_id_key is', array('@entity' => $this->entity_info['label'], '@entity_id_key' => strtoupper($this->entity_info['entity keys']['id']))),
-      '#type' => 'textfield',
+      '#type' => 'number',
+      '#min' => 1,
       '#default_value' => is_null($this->settings['entity_id']) ? NULL : (int) $this->settings['entity_id'],
       '#weight' => 100,
     );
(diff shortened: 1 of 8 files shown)


core/modules/layout/plugins/access/entity_id_layout_access.inc
```diff
@@ -33,6 +33,25 @@ class EntityIDLayoutAccess extends LayoutAccessNegatable {
     );
   }
 
+  /**
+   * {@inheritdoc}
+   */
+  function formValidate($form, &$form_state) {
```
jenlampton, line 40: I don't think we need any validation if we change the form element type to number. I wonder what that error messaging would look like though...

edit:
Screen Shot 2021-09-29 at 4 50 59 PM

bugfolder: I guess 'number' needs to be added as a type to the documentation: https://docs.backdropcms.org/form_api. 

Maybe doing explicit validation is worth getting more explicit and helpful validation messages. Looking through `form_validate_number()`, it looks like the error message for entering a non-integral value (after appropriately setting `#step`) would be `'%name is not a valid number'`.

Also, in the PR I'm validating for the entity ID to exist. If we drop `formValidate()`, then we lose that bit of validation. Seems like it would be useful to have.

klonos:
I prefer https://github.com/backdrop/backdrop/pull/3760 over this PR.

bugfolder:
Closing for replacement with PR https://github.com/backdrop/backdrop/pull/3771.

PR #3760, opened 2021-09-29
Issue #2668: Visibility condition for node ID does not throw error wh…

### LINKED ISSUE #2668: [UX] Layout and Block visibility conditions: add validation when entering entity IDs to ensure that only numbers are allowed.

jenlampton:
Fixes https://github.com/backdrop/backdrop-issues/issues/2668


klonos (approved):
LGTM 👍🏼 

I prefer this solution to the addition of a custom validation in https://github.com/backdrop/backdrop/pull/3759

quicksketch:
Thanks, merged into 1.x and 1.20.x!

PR #3761, opened 2021-09-30
Issue #5263: backdrop_is_front_page() fixed for path aliases

### LINKED ISSUE #5263: "Is the home page" visibility condition does not show blocks unless path is empty
**Description of the bug**

The "Is the home page" visibility condition can't tell it's on the home page when a node (specified as front page) is accessed via its alias, instead of via the bare domain.

**Steps To Reproduce**

To reproduce the behavior:
1. Set the site home page to `node/2`
2. Add a block to the default layout, add vis condition "Is the front page".
3. Visit `/about` -- the block is not shown.
5. Visit `/` -- the block is shown on the same page.

**Actual behavior**

Blocks do not appear when alias used.

**Expected behavior**

Blocks should appear when alias used.

**Additional information**

jenlampton:
Fixes https://github.com/backdrop/backdrop-issues/issues/5263

### DIFF
--- a/core/includes/path.inc
+++ b/core/includes/path.inc
@@ -276,7 +276,16 @@ function backdrop_is_front_page() {
   if (!isset($is_front_page)) {
     // As backdrop_path_initialize updates $_GET['q'] with the 'site_frontpage' path,
     // we can check it against the 'site_frontpage' variable.
-    $is_front_page = ($_GET['q'] == config_get('system.core', 'site_frontpage'));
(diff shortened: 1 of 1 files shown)


bugfolder:
@jenlampton, can you edit your initial comment to change the fixed issue number to 5263, not 5264.

jenlampton:
yes, sorry! (and thank you)

klonos (approved):
LGTM

PR #3763, opened 2021-09-30
Issue #3888: Remove menu item when Dashboard module is disabled

### LINKED ISSUE #3888: Remove menu item when Dashboard layout (or module) is disabled
**Description of the bug**
If a user disables the dashboard layout, the icon for the menu item disappears from the admin menu, but the dashboard menu remains without any logical destination.

**Steps To Reproduce**
To reproduce the behavior:
1. Go to '/admin/structure/layouts'
2. Click on 'Disable Layout' for the Dashboard
3. Note that the icon for the icon for the Dashboard in the admin menu disappears, but the menu item itself remains. 

Turns out, the same thing happens when you disable the Dashboard module. If you disable the Dashboard module, the icon disappears but the menu remains. 

**Actual behavior**

When you click on the Dashboard menu item AFTER disabling the layout, you get this:

Page_not_found___PR_2747_backdrop_backdrop_testing_site

**Expected behavior**
I guess I would expect that if I disabled the module or the layout for the Dashboard, that the menu would also disappear, since it is no longer relevant. 

I would believe that this might be a difficult problem to solve, but the fact that the menu icon is going away suggests it should also be possible to make the menu item go away as well. Could the the menu item be dependent upon the availability of the layout that it is pointed at?

klonos:
Fixes https://github.com/backdrop/backdrop-issues/issues/3888

### DIFF
--- a/core/modules/dashboard/dashboard.install
+++ b/core/modules/dashboard/dashboard.install
@@ -15,13 +15,13 @@ function dashboard_install() {
  * Implements hook_enable().
  */
 function dashboard_enable() {
-  // Enable the Dashboard menu item.
-  $link = db_query("SELECT * FROM {menu_links} WHERE router_path LIKE 'admin/dashboard' AND module = 'system'")->fetchAssoc();
+  // Enable the Dashboard menu item (if it exists and is disabled).
+  $link = db_query("SELECT * FROM {menu_links} WHERE router_path LIKE 'admin/dashboard' AND module = 'system' AND hidden = 1")->fetchAssoc();
   if ($link) {
     $link = menu_link_load($link['mlid']);
     $link['hidden'] = 0;
     menu_link_save($link);
-    // menu_rebuild();
+    menu_rebuild();
   }
 
   // Enable the Dashboard layout.
@@ -33,13 +33,13 @@ function dashboard_enable() {
  * Implements hook_disable().
  */
 function dashboard_disable() {
-  // Disable the Dashboard menu item.
-  $link = db_query("SELECT * FROM {menu_links} WHERE router_path LIKE 'admin/dashboard' AND module = 'system'")->fetchAssoc();
+  // Disable the Dashboard menu item (if it exists and is enabled).
+  $link = db_query("SELECT * FROM {menu_links} WHERE router_path LIKE 'admin/dashboard' AND module = 'system' AND hidden = 0")->fetchAssoc();
   if ($link) {
     $link = menu_link_load($link['mlid']);
     $link['hidden'] = 1;
     menu_link_save($link);
-    // menu_rebuild();
+    menu_rebuild();
   }
 
   // Disable the Dashboard layout.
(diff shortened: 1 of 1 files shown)


core/modules/dashboard/dashboard.install
```diff
@@ -12,10 +12,40 @@ function dashboard_install() {
 }
 
 /**
- * Implements hook_uninstall().
+ * Implements hook_enable().
+ */
+function dashboard_enable() {
+  // Enable the Dashboard menu item (if it exists and is disabled).
+  $link = db_query("SELECT * FROM {menu_links} WHERE router_path LIKE 'admin/dashboard' AND module = 'system' AND hidden = 1")->fetchAssoc();
```
ghost, line 19: It seems you're only using the mlid here, so probably best to just return that field rather than all of them...

core/modules/dashboard/dashboard.install
```diff
@@ -12,10 +12,40 @@ function dashboard_install() {
...
+  layout_reset_caches();
+}
+
+/**
+ * Implements hook_disable().
  */
 function dashboard_disable() {
-  // Disable the dashboard on module disable.
+  // Disable the Dashboard menu item (if it exists and is enabled).
+  $link = db_query("SELECT * FROM {menu_links} WHERE router_path LIKE 'admin/dashboard' AND module = 'system' AND hidden = 0")->fetchAssoc();
```
ghost, line 38: Same here.

ghost (changes requested):

core/modules/dashboard/dashboard.install
swampopus: They are using more than the mlid from the table, so this is fine. Resolving comment.

core/modules/dashboard/dashboard.install
swampopus: They are using more than the mlid from the table, so this is fine. Resolving comment.

swampopus (approved):
This code looks good to me.

yorkshire-pudding:
Closing and re-opening to initiate tests.

core/modules/dashboard/dashboard.install
yorkshire-pudding: Actually, `mlid` is the only thing used from the database. The other fields are from `menu_link_load`

core/modules/dashboard/dashboard.install
yorkshire-pudding: Actually, `mlid` is the only thing used from the database. The other fields are from `menu_link_load`

core/modules/dashboard/dashboard.install
swampopus: Ah, right.

swampopus:
Okay-- I created a new PR to make the code changes (I couldn't figure out how to suggest specific code changes here).
New PR: https://github.com/backdrop/backdrop/pull/5334

PR #3765, opened 2021-10-01
Issue #5273: Nested contextual links triggers sometimes unreachable

### LINKED ISSUE #5273: Nested contextual links triggers sometimes unreachable
**Description of the bug**

**Steps To Reproduce**

To reproduce the behavior:

**Actual behavior**

**Expected behavior**

**Additional information**

- Backdrop CMS version: 1.20.

bugfolder:

### DIFF
--- a/core/modules/contextual/js/contextual.js
+++ b/core/modules/contextual/js/contextual.js
@@ -37,8 +37,26 @@ Backdrop.behaviors.contextualLinks = {
       // Prepend the trigger.
       $wrapper.prepend($trigger);
 
-      // Prevent child contextual links overlapping their parent ones.
-      $wrapper.siblings().find('.contextual-links-wrapper').eq(0).css('margin-top', '20px');
+      // Prevent child contextual link triggers from overlapping their parent
+      // ones by shifting their position.
+      var $child = $wrapper.siblings().find('.contextual-links-wrapper').eq(0);
+      if ($child.length > 0) {
+        var height = $child.parent().height();
+        var parentOffset = $wrapper.offset();
+        var childOffset = $child.offset();
+        if (childOffset.top < parentOffset.top + 20 && childOffset.left > parentOffset.left  - 25) {
+          // There's a collision, so we need to shift the child.
+          if (height >= 40) {
+              // If there's enought vertical room in the child contextual links
+              // region, shift the child links down.
+            $child.css('margin-top', '20px');
+          }
+          else {
+            // Otherwise shift the child links to the left.
+            $child.css('margin-right', '25px');
+          }
+        }
+      }
     });
   }
 };
--- a/core/modules/contextual/js/contextual.js
+++ b/core/modules/contextual/js/contextual.js
@@ -50,8 +50,8 @@ Backdrop.behaviors.contextualLinks = {
             if (childOffset.left > parentOffset.left - 25) {
               // There's a collision at right, so we need to shift the child.
               if (parentHeight >= 40) {
-                  // If there's enought vertical room in the child contextual links
-                  // region, shift the child links down.
+                  // If there's enough vertical room in the child contextual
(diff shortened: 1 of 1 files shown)


core/modules/contextual/js/contextual.js
```diff
@@ -37,8 +37,26 @@ Backdrop.behaviors.contextualLinks = {
...
+        if (childOffset.top < parentOffset.top + 20 && childOffset.left > parentOffset.left  - 25) {
+          // There's a collision, so we need to shift the child.
+          if (height >= 40) {
+              // If there's enought vertical room in the child contextual links
+              // region, shift the child links down.
+            $child.css('margin-top', '20px');
+          }
+          else {
+            // Otherwise shift the child links to the left.
+            $child.css('margin-right', '25px');
```
klonos: @bugfolder I haven't tested this yet, but I imagine that this would need to account for RTL vs LTR, and shift either right or left accordingly.

bugfolder: Great point, thanks! I've updated the PR to do so. Details and new test code in the issue.

core/modules/contextual/js/contextual.js
```diff
@@ -37,8 +37,46 @@ Backdrop.behaviors.contextualLinks = {
...
+                $child.css('margin-top', '20px');
+              }
+              else {
+                // Otherwise shift the child links to the left.
+                $child.css('margin-right', '25px');
+              }
+            }
+          }
+          else {
+            // Handle RTL
```
klonos:
```suggestion
            // Handle RTL.
```

core/modules/contextual/js/contextual.js
```diff
@@ -37,8 +37,46 @@ Backdrop.behaviors.contextualLinks = {
...
+      // ones by shifting their position.
+      var $child = $wrapper.siblings().find('.contextual-links-wrapper').eq(0);
+      if ($child.length > 0) {
+        var parentHeight = $child.parent().height();
+        var parentOffset = $wrapper.offset();
+        var childOffset = $child.offset();
+        var dir = $('html').attr('dir');
+        if (childOffset.top < parentOffset.top + 20) {
+          if (dir == 'ltr') {
+            if (childOffset.left > parentOffset.left  - 25) {
```
klonos:
```suggestion
            if (childOffset.left > parentOffset.left - 25) {
```

core/modules/contextual/js/contextual.js
```diff
@@ -37,8 +37,46 @@ Backdrop.behaviors.contextualLinks = {
...
+              }
+              else {
+                // Otherwise shift the child links to the left.
+                $child.css('margin-right', '25px');
+              }
+            }
+          }
+          else {
+            // Handle RTL
+            if (childOffset.left < parentOffset.left  + 25) {
```
klonos:
```suggestion
            if (childOffset.left < parentOffset.left + 25) {
```

core/modules/contextual/js/contextual.js
```diff
@@ -37,8 +37,46 @@ Backdrop.behaviors.contextualLinks = {
...
+                $child.css('margin-right', '25px');
+              }
+            }
+          }
+          else {
+            // Handle RTL.
+            if (childOffset.left < parentOffset.left + 25) {
+              // There's a collision at left, so we need to shift the child.
+              if (parentHeight >= 40) {
+                  // If there's enought vertical room in the child contextual links
```
jenlampton: minor typo here. `enought`

bugfolder: Thanks, fixed (in two places). Also rewrapped the comments, which were over 80 char long.

bugfolder:
Closing in favor of https://github.com/backdrop/backdrop/pull/4300.

PR #3768, opened 2021-10-01
Issue-4677 Remove Search is not casesensitive text

### LINKED ISSUE #4677: The help text in "String contains" field in "User interface translation" says that search is case-sensitive, when it's not
Screen Shot 2020-10-02 at 1 38 00 pm

mmachinen:
Fixes backdrop/backdrop-issues#4677
The help text in "String contains" field in "User interface translation" says that search is case-sensitive when the search is case-insensitive. 
Fix: Removed the help text that says search is case-sensitive.


quicksketch:
Merged into 1.x and 1.20.x. Thanks and welcome @mmachinen!

PR #3771, opened 2021-10-01
Issue #2668: [UX] Layout and Block visibility conditions: add validation

### LINKED ISSUE #2668: [UX] Layout and Block visibility conditions: add validation when entering entity IDs to ensure that only numbers are allowed.

### LINKED ISSUE #5262: Visibility condition for node ID does not throw error when non-integer

bugfolder:
Fixes https://github.com/backdrop/backdrop-issues/issues/5262.
Fixes https://github.com/backdrop/backdrop-issues/issues/2668.

Replaces PR https://github.com/backdrop/backdrop/pull/3759.

### DIFF
--- a/core/modules/layout/plugins/access/entity_id_layout_access.inc
+++ b/core/modules/layout/plugins/access/entity_id_layout_access.inc
@@ -71,8 +71,8 @@ class EntityIDLayoutAccess extends LayoutAccessNegatable {
   function summary() {
     if ($this->settings['negate']) {
       return t('@entity ID is NOT: @entity_id', array(
-        '@entity'  => $this->entity_info['label'],
-        '@entity_id'  => $this->settings['entity_id'],
+        '@entity' => $this->entity_info['label'],
+        '@entity_id' => $this->settings['entity_id'],
       ));
     }
     else {
--- a/core/modules/layout/plugins/access/entity_id_layout_access.inc
+++ b/core/modules/layout/plugins/access/entity_id_layout_access.inc
@@ -28,6 +28,7 @@ class EntityIDLayoutAccess extends LayoutAccessNegatable {
     $form['entity_id'] = array(
       '#title' => t('@entity @entity_id_key is', array('@entity' => $this->entity_info['label'], '@entity_id_key' => strtoupper($this->entity_info['entity keys']['id']))),
       '#type' => 'number',
+      '#required' => TRUE,
       '#min' => 1,
       '#default_value' => is_null($this->settings['entity_id']) ? NULL : (int) $this->settings['entity_id'],
       '#weight' => 100,
--- a/core/modules/layout/plugins/access/entity_id_layout_access.inc
+++ b/core/modules/layout/plugins/access/entity_id_layout_access.inc
@@ -28,8 +28,8 @@ class EntityIDLayoutAccess extends LayoutAccessNegatable {
     $form['entity_id'] = array(
       '#title' => t('@entity @entity_id_key is', array(
         '@entity' => $this->entity_info['label'],
-        '@entity_id_key' => strtoupper($this->entity_info['entity keys']['id'],
-      ))),
+        '@entity_id_key' => strtoupper($this->entity_info['entity keys']['id']),
+      )),
       '#type' => 'number',
       '#required' => TRUE,
       '#min' => 1,
--- a/core/modules/layout/plugins/access/entity_id_layout_access.inc
+++ b/core/modules/layout/plugins/access/entity_id_layout_access.inc
@@ -37,6 +37,22 @@ class EntityIDLayoutAccess extends LayoutAccessNegatable {
   /**
    * {@inheritdoc}
    */
+  function formValidate($form, &$form_state) {
+    parent::formValidate($form, $form_state);
+
+    $entity_id = $form_state['values']['entity_id'];
+    if (is_numeric($entity_id) && round($entity_id) == $entity_id && $entity_id >= 1) {
+      $entity = entity_load($this->entity_type, $entity_id);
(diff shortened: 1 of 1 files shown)


bugfolder:
Per [this comment](https://github.com/backdrop/backdrop-issues/issues/5281#issuecomment-932991744), we can prevent non-numeric entries by setting the element `'#required' => TRUE`. The browser validation will submit an empty string, and then standard form validation will reject the empty submission.

PR updated.

core/modules/layout/plugins/access/entity_id_layout_access.inc
```diff
@@ -28,12 +28,29 @@ class EntityIDLayoutAccess extends LayoutAccessNegatable {
...
+   */
+  function formValidate($form, &$form_state) {
+    parent::formValidate($form, $form_state);
+
+    $entity_id = $form_state['values']['entity_id'];
+    if (is_numeric($entity_id) && round($entity_id) == $entity_id && $entity_id >= 1) {
+      $entity = entity_load($this->entity_type, $entity_id);
+      if (!$entity) {
+        $label = $this->entity_info['label'];
+        form_set_error('entity_id', t('There is no @label with ID @entity_id.', array('@label' => $label, '@entity_id' => $entity_id)));
```
indigoxela: LGTM, just a tiny coding standard request:

```
         form_set_error('entity_id', t('There is no @label with ID @entity_id.', array(
           '@label' => $label,
           '@entity_id' => $entity_id,
         )));
```
That's only for better readability.

bugfolder: Sure, done. Also for the similar constructions in function `form()` and `summary()` in that same file.

core/modules/layout/plugins/access/entity_id_layout_access.inc
```diff
@@ -47,10 +70,16 @@ class EntityIDLayoutAccess extends LayoutAccessNegatable {
    */
   function summary() {
     if ($this->settings['negate']) {
-      return t('@entity ID is NOT: @entity_id', array('@entity'  => $this->entity_info['label'], '@entity_id'  => $this->settings['entity_id']));
+      return t('@entity ID is NOT: @entity_id', array(
+        '@entity'  => $this->entity_info['label'],
+        '@entity_id'  => $this->settings['entity_id'],
```
klonos, line 75:
```suggestion
        '@entity_id' => $this->settings['entity_id'],
```

klonos:
```suggestion
        '@entity' => $this->entity_info['label'],
        '@entity_id' => $this->settings['entity_id'],
```

core/modules/layout/plugins/access/entity_id_layout_access.inc
```diff
@@ -47,10 +70,16 @@ class EntityIDLayoutAccess extends LayoutAccessNegatable {
...
+      ));
     }
     else {
-      return t('@entity ID is: @entity_id', array('@entity'  => $this->entity_info['label'], '@entity_id'  => $this->settings['entity_id']));
+      return t('@entity ID is: @entity_id', array(
+        '@entity'  => $this->entity_info['label'],
+        '@entity_id'  => $this->settings['entity_id'],
```
klonos:
```suggestion
        '@entity' => $this->entity_info['label'],
        '@entity_id' => $this->settings['entity_id'],
```

core/modules/layout/plugins/access/entity_id_layout_access.inc
```diff
@@ -26,14 +26,37 @@ class EntityIDLayoutAccess extends LayoutAccessNegatable {
...
   }
 
+  /**
+   * {@inheritdoc}
+   */
+  function formValidate($form, &$form_state) {
+    parent::formValidate($form, $form_state);
+
+    $entity_id = $form_state['values']['entity_id'];
+    if (is_numeric($entity_id) && round($entity_id) == $entity_id && $entity_id >= 1) {
```
docwilmot, line 48: Why this rounding?

bugfolder: To make sure it's an integer.

klonos: Wouldn't typecasting to `(int)` work?

klonos: ...or `intval()` perhaps?

bugfolder: `intval()` would work, too; but it's one character longer. Is there a reason why it would be better?

quicksketch: Either approach seems fine to me. :shrug: It's weird how hard it is to check if a string value is an integer number and all solutions I know of are a little weird.

quicksketch:
Merged into 1.x and 1.20.x. Thanks @bugfolder, @indigoxela, @docwilmot, and @klonos!

PR #3775, opened 2021-10-01
Issue #5279: Adding aria attributes to the pagination component

### LINKED ISSUE #5279: [A11Y] Pagination component is not as accessible as it could be
### 10. Pagination component is not as accessible as it could be

**Description of Issue**: Ideally, the pagination component would make use of a little more semantic HTML, such as the `<nav>` element and some ARIA attributes to aid a user in knowing which part of the pagination is currently acitve.

**URL**: https://pr3750-llnmetezlaix5oe10ddq5m3nya7zdz2b.tugboat.qa/

**Testing Environment/Browser**: Windows Lenovo PC, Windows 10 Pro, Chrome Browser v. 94.0.4606.61, JAWS 2020.12.13

**[Severity](https://github.com/backdrop/backdrop-issues/issues/5244#issuecomment-931739260)**: P1

**File to change**: `core/includes/pager.inc`, line 158

**Documentation**: https://docs.backdropcms.org/api/backdrop/core%21includes%21pager.inc/function/theme_pager/1

**Solution**: 

Suggested markup is as below, but significant changes are listed as follows. Some of these items can be achieved in Backdrop 1.x, but others will need to wait until 2.x, as they would be braking changes for existing sites.

1) Add `aria-current="page"` attribute to the active pager item.
1) Add `aria-label` attribute to all other pages, with page number.
1) Remove the h2 that is used nearby and says "pages"
1) Add `display:none` to specific elements to create desktop vs mobile 

1) 2.x change: Add a surrounding `nav` tag with `aria-label="Pagination"`
1) 2.x change: use button elements for First / Last / Next / Previous

## Mobile recommendation:
```
<nav aria-label="Pagination">
  <button>Previous</button><!-- can be disabled (not removed) if this is the first -->
  <p>Page 1 of 10</p>
  <button>Next</button><!-- can be disabled (not removed) if this is the last -->
</nav>
```

## desktop recommendation
```
<nav aria-label="Pagination">
  <button>Previous</button><!-- can be disabled (not removed) if this is the first -->
  <button>First</button><!-- can be disabled (not removed) if this is the first -->
  <ul>
    <li><a href="/1" aria-label="Page 1" aria-current="page">1</a></li>
    <li><a href="/2" aria-label="page 2">2</a></li>
  </ul>
    <button>Last</button><!-- can be disabled (not removed) if this is the last -->
    <button>Next</button><!-- can be disabled (not removed) if this is the last -->
</nav>


**Screenshot:**
![image](https://user-images.githubusercontent.com/12848168/135544627-17ba43ed-617a-48d8-b9d2-a8308f2ecc18.png)

### PR DESCRIPTION
**Pull request by katielo-0830**
Fixes: https://github.com/backdrop/backdrop-issues/issues/5279

### DIFF
--- a/core/includes/pager.inc
+++ b/core/includes/pager.inc
@@ -230,9 +230,11 @@ function theme_pager($variables) {
           );
         }
         if ($i == $pager_current) {
+          $pager_current_parameters = $parameters;
+          $pager_current_parameters['aria-current'] = "page";
           $items[] = array(
             'class' => array('pager-current'),
-            'data' => $i,
+            'data' => theme('pager_link', array('text' => $i, 'page_name' => pager_load_array($i - 1, $element, $pager_page_array), 'element' => $element, 'parameters' => $pager_current_parameters)),
           );
         }
         if ($i > $pager_current) {
@@ -262,10 +264,10 @@ function theme_pager($variables) {
         'data' => $li_last,
       );
     }
-    return '<h2 class="element-invisible">' . t('Pages') . '</h2>' . theme('item_list', array(
+    return '<nav aria-label="Pagination>' . theme('item_list', array(
       'items' => $items,
       'attributes' => array('class' => array('pager')),
-    ));
+    )) . '</nav>';
   }
 
   // Single page of contents, no pager needed.
@@ -303,6 +305,7 @@ function theme_pager_first($variables) {
 
   // If we are anywhere but the first page
   if ($pager_page_array[$element] > 0) {
+    $parameters['aria-label'] = "first page";
     $output = theme('pager_link', array('text' => $text, 'page_new' => pager_load_array(0, $element, $pager_page_array), 'element' => $element, 'parameters' => $parameters));
   }
 
@@ -341,6 +344,7 @@ function theme_pager_previous($variables) {
     }
     // The previous page is not the first page.
     else {
+      $parameters['aria-label'] = 'previous page';
       $output = theme('pager_link', array('text' => $text, 'page_new' => $page_new, 'element' => $element, 'parameters' => $parameters));
     }
   }
@@ -379,6 +383,7 @@ function theme_pager_next($variables) {
     }
     // The next page is not the last page.
     else {
+      $parameters['aria-label'] = "next page";
       $output = theme('pager_link', array('text' => $text, 'page_new' => $page_new, 'element' => $element, 'parameters' => $parameters));
     }
   }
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

File: core/includes/pager.inc [hunk 1]
```diff
@@ -262,10 +264,10 @@ function theme_pager($variables) {
         'data' => $li_last,
       );
     }
-    return '<h2 class="element-invisible">' . t('Pages') . '</h2>' . theme('item_list', array(
+    return '<nav aria-label="Pagination>' . theme('item_list', array(
```
jenlampton, line 265: I don't think we want an additional nav tag added here, or we'll get duplicated nav tags. All that needs to be changed on this line is removing the h2. 

edit: (I'm not sure about this actually... checking)

core/includes/pager.inc
```diff
@@ -262,10 +264,10 @@ function theme_pager($variables) {
...
       'items' => $items,
       'attributes' => array('class' => array('pager')),
-    ));
+    )) . '</nav>';
```
jenlampton, line 268: same thing here, we'll get nested nav tags if we include this.

core/includes/pager.inc
```diff
@@ -230,9 +230,11 @@ function theme_pager($variables) {
           );
         }
         if ($i == $pager_current) {
+          $pager_current_parameters = $parameters;
+          $pager_current_parameters['aria-current'] = "page";
           $items[] = array(
             'class' => array('pager-current'),
-            'data' => $i,
+            'data' => theme('pager_link', array('text' => $i, 'page_name' => pager_load_array($i - 1, $element, $pager_page_array), 'element' => $element, 'parameters' => $pager_current_parameters)),
```
jenlampton: I don't think this should be a link, since it's the current page marker (and doesn't link to any other page). Why not just use `'<span aria-current="page">' . $i . '</span>'`

core/includes/pager.inc
katielo-0830: It seems that there is no nav tag.
Here is the html I copied from the inspector from the build of the pull request (Tugboat)
```
<nav aria-label="Pagination><div class=" item-list">
  <ul class="pager">
    <li class="pager-current odd first"><span aria-current="page">1</span></li>
    <li class="pager-item even"><a aria-label="last page" title="Go to page 2" href="/home?page=1">2</a></li>
    <li class="pager-next odd"><a aria-label="last page" title="Go to next page" href="/home?page=1">next ›</a></li>
    <li class="pager-last even last"><a aria-label="last page" title="Go to last page" href="/home?page=1">last »</a>
    </li>
  </ul>
</nav>
```

core/includes/pager.inc
katielo-0830: Sounds good! Addressed in the new commit.

PR #3778, opened 2021-10-01
Issue #5280: Fix variables in user profile and comment theme functions

### LINKED ISSUE #5280: View mode not accessible from user or comment template files
**Description of the bug**

With nodes and terms you can access the `$view_mode` variable from the template file, but it does not seem to be available for user profiles. (Or potentially comments based on a little research I've done).

laryn:
Fixes https://github.com/backdrop/backdrop-issues/issues/5280


core/modules/comment/templates/comment.tpl.php
```diff
@@ -45,9 +46,10 @@
  *   modules, intended to be displayed after the main title tag that appears in
  *   the template.
  *
- * These two variables are provided for context:
+ * These variables are provided for context:
  * - $comment: Full comment object.
  * - $node: Node entity the comments are attached to.
+ * - $view_mode: Display mode, e.g. 'full', or 'default'.
```
indigoxela: @laryn I wonder, why you put the view_mode into the context section. Isn't it just a simple variable like the ones above?

IMO the context are the full objects, everything else are variables for direct output. What do you think?

laryn: @indigoxela Maybe I'm interpreting "context" in the wrong way. I was just thinking the `view_mode` isn't really data that someone has entered, but more like the context in which they are viewing the entity. 

But I see your point and I can move it up into the main `$variables` section.

quicksketch:
Merged into 1.x and 1.21.x. Thanks @laryn and @indigoxela!

PR #3781, opened 2021-10-03
Issue #5264: [UX] Clean up english on number form validation failure

### LINKED ISSUE #5264: Fixed: [UX] Clean up English on number form validation failure.
**Description of the bug**

When entering 0 into a field that requires a positive integer, the error messaging reads 
`Node NID is must be higher or equal to 1.`

There are several issues with this that should be resolved:
* 2 verbs, `is` and `must be`
* strange phrasing `higher or equal to`. 

**Steps To Reproduce**

To reproduce the behavior:
0. apply the patch at https://github.com/backdrop/backdrop/pull/3760
1. Create a layout at node/5
2. Add a block
3. add a node ID visibility condition to the block
3. enter 0 for your node ID

**Actual behavior**

the error messaging reads 
`Node NID is must be higher or equal to 1.`

**Expected behavior**

how about these:

* `%name: the value must be numeric.`
* `%name: the value must be greater than or equal to %min.`
* `%name: the value must be less than or equal to %max.`
* `%name: the value must be an integral number of steps %step from %offset.`

We could also add a separate test for integrality (probably the most common use case for `#step`). If both `#min` and `#step` are integers, then the error would be

* `%name: the value must be an integer.`

**Additional information**

- Backdrop CMS version: 1.20.0 + patch

---

This is a follow-up to https://github.com/backdrop/backdrop-issues/issues/2668

bugfolder:
Fixes https://github.com/backdrop/backdrop-issues/issues/5264.


core/includes/form.inc
```diff
@@ -4594,7 +4596,13 @@ function form_validate_number(&$element, &$form_state) {
     $offset = isset($element['#min']) ? $element['#min'] : 0.0;
 
     if (!valid_number_step($value, $element['#step'], $offset)) {
-      form_error($element, t('%name is not a valid number.', array('%name' => $name)));
+      // Integer steps are the most common use case, so give a specific message.
+      if ($element['#step'] == 1 && is_numeric($offset) && $offset == round($offset)) {
+        form_error($element, t('%name: the value must be an integer.', array('%name' => $name)));
+      }
+      else {
+        form_error($element, t('%name: the value must be an integral number of steps %step from %offset.', array('%name' => $name, '%step' => $element['#step'], '%offset' => $offset)));
```
docwilmot: I'm not so sure the average person would catch what "integral" means here? That word commonly used has a different meaning from integer. I had to Google it to confirm the meaning myself.

bugfolder: Good point. I'm so used to the mathematical interpretation that I forget about the common-language meaning. I'll update the wording.

bugfolder: "A whole number of steps."

klonos: Yup, I agree. "whole number" makes more sense than "integral number" if you're not a mathematician 👍🏼

core/includes/form.inc
```diff
@@ -4594,7 +4596,13 @@ function form_validate_number(&$element, &$form_state) {
     $offset = isset($element['#min']) ? $element['#min'] : 0.0;
 
     if (!valid_number_step($value, $element['#step'], $offset)) {
-      form_error($element, t('%name is not a valid number.', array('%name' => $name)));
+      // Integer steps are the most common use case, so give a specific message.
+      if ($element['#step'] == 1 && is_numeric($offset) && $offset == round($offset)) {
+        form_error($element, t('%name: the value must be an integer.', array('%name' => $name)));
+      }
+      else {
+        form_error($element, t('%name: the value must be a whole number of steps %step from %offset.', array('%name' => $name, '%step' => $element['#step'], '%offset' => $offset)));
```
klonos: Would something like this make better sense?:

```suggestion
        form_error($element, t('%name: the value must be a whole number of steps of %step, starting from %offset.', array('%name' => $name, '%step' => $element['#step'], '%offset' => $offset)));
```

The same goes for `'step_mismatch_general'` in core/modules/simpletest/tests/form.test

klonos (approved):
LGTM 👍🏼

core/includes/form.inc
bugfolder: That is indeed better; better still (and what I changed to) is:

> the value must be a whole number of steps of size %step, starting from %offset.

Also just noticed the "integer" in line 4601. For consistency, this, too, should be "a whole number," so also changed.

quicksketch:
Merged into 1.x and 1.21.x. Thanks @bugfolder!

PR #3783, opened 2021-10-04
Issue #5283: Update text on the account cancellation screen

### LINKED ISSUE #5283: [UX] Update Help text on the account cancellation screen
Related to: #5121

## Problem/Motivation
When a user visits the account cancellation form, We have four options:

- Disable the account and keep its content.
- Disable the account and unpublish its content.
- Delete the account and make its content belong to the _Anonymous_ user.
- Delete the account and its content.

And the description below states:

> Select the method to cancel the account above. This action cannot be undone.

The "This action cannot be undone" statement does not apply to the first two options though.

Our Drupal brethren in https://www.drupal.org/project/drupal/issues/3199972 have decided the following:

> We discussed this issue at [#3200598: Drupal Usability Meeting 2021-02-26](https://www.drupal.org/project/drupal/issues/3200598)
>
> The suggestion was to update the text of last two options to include word "Permanently".
> The last two options will be:
>
> - Permanently delete the account and make its content belong to the "anonymous user name" user.
> - Permanently delete the account and its content.
>
> And the other suggestion is to remove the following text from the description available below the options:
>
> > This action cannot be undone.

In our case, the "Cancel account" button is also a red `danger` button, which shouldn't be the case if one of the two first options are selected.

klonos:
Fixes https://github.com/backdrop/backdrop-issues/issues/5283

### DIFF
--- a/core/modules/user/user.admin.inc
+++ b/core/modules/user/user.admin.inc
@@ -50,11 +50,6 @@ function user_multiple_cancel_confirm($form, &$form_state) {
     }
   }
 
-  // Get the name of the current user account in the array. If only one account
-  // was selected, then this will happen to be the only account name.
-  $user = current($accounts);
-  $user_name = $user->name;
-
   $form_state['accounts'] = $accounts;
 
   module_load_include('inc', 'user', 'user.pages');
@@ -82,6 +77,10 @@ function user_multiple_cancel_confirm($form, &$form_state) {
     '#description' => t('When enabled, the user will receive an email notification after the account has been cancelled.'),
   );
 
+  // Get the name of the current user account in the array. If only one account
+  // was selected, then this will happen to be the only account name.
+  $user = current($accounts);
+  $user_name = $user->name;
   $confirm_question = format_plural(count($accounts), 'Are you sure you want to cancel the account %name?', 'Are you sure you want to cancel these user accounts?', array('%name' => $user_name));
   return confirm_form($form, $confirm_question, 'admin/people', '', t('Cancel accounts'), t('Cancel'));
 }
--- a/core/modules/user/user.admin.inc
+++ b/core/modules/user/user.admin.inc
@@ -157,7 +157,6 @@ function user_admin_settings($form, &$form_state) {
     // Disable all account cancellation methods that have #access defined, as
     // those cannot be configured as default method.
     if (isset($form['registration_cancellation']['user_cancel_method'][$element]['#access'])) {
-      // $form['registration_cancellation']['user_cancel_method'][$element]['#access'] = FALSE;
       $form['registration_cancellation']['user_cancel_method'][$element]['#disabled'] = TRUE;
       $form['registration_cancellation']['user_cancel_method'][$element]['#description'] = t('This cancellation method cannot be selected as the default, as it specifies #access.');
     }
--- a/core/modules/user/user.admin.inc
+++ b/core/modules/user/user.admin.inc
@@ -50,13 +50,10 @@ function user_multiple_cancel_confirm($form, &$form_state) {
   }
 
   module_load_include('inc', 'user', 'user.pages');
-  $form['user_cancel_method'] = array(
-    '#type' => 'item',
-    '#title' => t('When cancelling these accounts'),
-  );
-  $form['user_cancel_method'] += user_cancel_methods();
+  $form += user_cancel_methods();
 
-  // Remove method descriptions.
+  // Remove individual method descriptions (these are only displayed on the
+  // confirmation form for own account cancellation).
   foreach (element_children($form['user_cancel_method']) as $element) {
     unset($form['user_cancel_method'][$element]['#description']);
   }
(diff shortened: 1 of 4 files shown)


core/modules/user/user.pages.inc
```diff
@@ -485,29 +487,53 @@ function user_cancel_confirm_form_submit($form, &$form_state) {
...
+          t('Your account will be <u><em>permanently</em></u> removed, and all account information deleted.'),
+          t('All of your content will also be deleted.'),
+        )
+      )),
+      'danger' => TRUE,
+      // This was meant to be used in user_admin_settings() in order to remove
+      // cancellation methods that have #access defined, as those cannot be
+      // configured as the default method. However this property was never being
+      // returned by this function here before. ¯\_(ツ)_/¯
       'access' => user_access('administer users'),
```
klonos, line 537: This pre-exiting code was meant to be used in user_admin_settings() in order to remove cancellation methods that have `#access` defined, as those cannot be configured as the default method. However this property was never being returned by this function here before 🤷🏼 

I've raised https://github.com/backdrop/backdrop-issues/issues/5121 for that.

core/modules/user/user.pages.inc
```diff
@@ -516,16 +542,30 @@ function user_cancel_methods() {
...
+    // dangerous/irreversible.
+    if (isset($method['danger']) && $method['danger']) {
+      $additional_properties[$name]['#danger'] = $method['danger'];
+    }
+    $additional_properties[$name]['#description'] = $method['description'];
+    // Flag used to exclude cancellation methods from being set as the
+    // default.
+    if (isset($method['access']) && $method['access']) {
+      $additional_properties[$name]['#access'] = $method['access'];
+    }
```
klonos, line 558: Making sure that this property of each cancellation option is available for any function that happens to call `user_cancel_methods()`. What we actually do with it in `user_admin_settings()` (`admin/config/people/settings`) is to be decided in https://github.com/backdrop/backdrop-issues/issues/5121

core/modules/user/user.admin.inc
```diff
@@ -150,23 +147,24 @@ function user_admin_settings($form, &$form_state) {
...
+    unset($form['registration_cancellation']['user_cancel_method'][$element]['#description']);
+    // Disable all account cancellation methods that have #access defined, as
     // those cannot be configured as default method.
     if (isset($form['registration_cancellation']['user_cancel_method'][$element]['#access'])) {
-      $form['registration_cancellation']['user_cancel_method'][$element]['#access'] = FALSE;
-    }
-    // Remove the description (only displayed on the confirmation form).
-    else {
-      unset($form['registration_cancellation']['user_cancel_method'][$element]['#description']);
+      // $form['registration_cancellation']['user_cancel_method'][$element]['#access'] = FALSE;
```
klonos, line 160: This line was doing absolutely nothing before, as `'#access'` was not passed in `user_cancel_methods()` before (see https://github.com/backdrop/backdrop-issues/issues/5121). What I'm doing here should be moved to a separate PR after we've made a decision re whether to be hiding such options, or showing them with disabled/locked radios instead.

core/modules/user/user.admin.inc
```diff
@@ -150,23 +154,23 @@ function user_admin_settings($form, &$form_state) {
...
+    // for own account cancellation).
+    unset($form['registration_cancellation']['user_cancel_method'][$element]['#description']);
+    // Disable all account cancellation methods that have #access defined, as
     // those cannot be configured as default method.
     if (isset($form['registration_cancellation']['user_cancel_method'][$element]['#access'])) {
-      $form['registration_cancellation']['user_cancel_method'][$element]['#access'] = FALSE;
-    }
-    // Remove the description (only displayed on the confirmation form).
-    else {
-      unset($form['registration_cancellation']['user_cancel_method'][$element]['#description']);
```
klonos, line 167: This line was doing absolutely nothing before, as `'#access'` was not passed from `user_cancel_methods()` before (see https://github.com/backdrop/backdrop-issues/issues/5121). Perhaps what I'm doing here should be moved to a separate PR, after we've made a decision re whether to be hiding such options, or showing them with disabled/locked radios instead (which is what I think provides better UX).

...the code change is rather small though, so if we come to a consensus in https://github.com/backdrop/backdrop-issues/issues/5121, we might as well fix it here.

core/modules/user/user.admin.inc
```diff
@@ -150,23 +154,23 @@ function user_admin_settings($form, &$form_state) {
...
   foreach (element_children($form['registration_cancellation']['user_cancel_method']) as $element) {
-    // Remove all account cancellation methods that have #access defined, as
+    // Remove the individual options descriptions provided by
+    // user_cancel_methods() (these are only displayed on the confirmation form
+    // for own account cancellation).
+    unset($form['registration_cancellation']['user_cancel_method'][$element]['#description']);
+    // Disable all account cancellation methods that have #access defined, as
     // those cannot be configured as default method.
     if (isset($form['registration_cancellation']['user_cancel_method'][$element]['#access'])) {
-      $form['registration_cancellation']['user_cancel_method'][$element]['#access'] = FALSE;
```
klonos, line 163: This line here was doing absolutely nothing before, as `'#access'` was not passed from `user_cancel_methods()` before (see https://github.com/backdrop/backdrop-issues/issues/5286). Perhaps what I'm doing here should be moved to a separate PR, after we've made a decision re whether to be hiding such options, or showing them with disabled/locked radios instead (which is what I think provides better UX).

...the code change is rather small though, so if we come to a consensus in https://github.com/backdrop/backdrop-issues/issues/5286, we might as well fix it here.

core/modules/user/user.pages.inc
```diff
@@ -492,22 +494,41 @@ function user_cancel_confirm_form_submit($form, &$form_state) {
...
-      'title' => t('Delete the account and its content.'),
-      'description' => t('Your account will be removed and all account information deleted. All of your content will also be deleted.'),
+      'title' => t('Permanently delete the account and its content.'),
+      'description' => theme('item_list', array('items' => array(
+          t('Your account will be <u><em>permanently</em></u> removed, and all account information deleted.'),
+          t('All of your content will also be deleted.'),
+        )
+      )),
+      'danger' => TRUE,
       'access' => user_access('administer users'),
```
klonos, line 532: This pre-exiting code was meant to be used in `user_admin_settings()` in order to remove cancellation methods that have `#access` defined, as those cannot be configured as the default method. However this property was never being returned by this function here before 🤷🏼 

I've raised https://github.com/backdrop/backdrop-issues/issues/5286 for that.

core/modules/user/user.pages.inc
```diff
@@ -516,16 +537,30 @@ function user_cancel_methods() {
...
+    // dangerous/irreversible.
+    if (isset($method['danger']) && $method['danger']) {
+      $additional_properties[$name]['#danger'] = $method['danger'];
+    }
+    $additional_properties[$name]['#description'] = $method['description'];
+    // Flag used to exclude cancellation methods from being set as the
+    // default.
+    if (isset($method['access']) && $method['access']) {
+      $additional_properties[$name]['#access'] = $method['access'];
+    }
```
klonos, line 553: This is a preexisting property of each cancellation option. What I'm doing here is to just make sure that it is actually available for any function that happens to call `user_cancel_methods()`. What we actually do with it in `user_admin_settings()` (`admin/config/people/settings`) is to be decided in https://github.com/backdrop/backdrop-issues/issues/5286

PR #3786, opened 2021-10-06
Issue #5288: Prevent PHP warning on date formats during upgrade from D7.

### LINKED ISSUE #5288: PHP warning on upgrade from Drupal 7 in system_get_date_formats()
**Description of the bug**

I'm seeing the following PHP warning when I upgrade sites from Drupal 7:

```
Warning: Invalid argument supplied for foreach() in system_get_date_formats() (line 3911 of system.module 
```

I know this isn't an issue because if you push past it and complete the upgrade everything then works fine, but it would be nice if we could avoid any kind of scary warnings during the upgrade process, which can be stressful enough :)

**Steps To Reproduce**

To reproduce the behavior:
1. Load in Drupal 7 database
2. Visit `core/update.php`
3. Error appears on this page

**Additional information**

- Backdrop CMS version: 1.20.1
- Web server and its version:

jenlampton:
Fixes https://github.com/backdrop/backdrop-issues/issues/5288

### DIFF
--- a/core/modules/system/system.module
+++ b/core/modules/system/system.module
@@ -3908,8 +3908,10 @@ function system_get_date_formats($date_format_name = NULL) {
   $date_formats = &backdrop_static(__FUNCTION__, array());
   if (empty($date_formats)) {
     $date_formats = config_get('system.date', 'formats');
-    foreach ($date_formats as $name => $format) {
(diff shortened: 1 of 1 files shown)


quicksketch:

PR #3788, opened 2021-10-07
Issue #5291: Fixing a php error when a field formatter has no output.

### LINKED ISSUE #5291: Unsupported opperand types error when return value from a field formatter is empty
**Description of the bug**

When a field formatter returns an empty array instead of a render array, it causes the following PHP error:
```
Error: Unsupported operand types in node_build_content() (line 1211 of /srv/http/backdrop/core/modules/node/node.module)
``` 

**Steps To Reproduce**

To reproduce the behavior:
1. Install the Reference module with the changes from https://github.com/backdrop-contrib/reference/issues/5
2. Add a reference field to allow nodes to reference each other and that uses the rendered entity field formatter. 
3. Create a circular reference where 2 nodes reference each other.

hosef:
Fixes backdrop/backdrop-issues#5291

### DIFF
--- a/core/modules/node/node.module
+++ b/core/modules/node/node.module
@@ -1203,7 +1203,13 @@ function node_build_content(Node $node, $view_mode = 'full', $langcode = NULL) {
   // twice.
   field_attach_prepare_view('node', array($node->nid => $node), $view_mode, $langcode);
   entity_prepare_view('node', array($node->nid => $node));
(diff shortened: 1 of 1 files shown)


klonos:
Fixes https://github.com/backdrop/backdrop-issues/issues/5291

@hosef I tried, but that ^^ doesn't seem to work - it needs to be added to the first comment in the PR, by the PR owner.

klonos (approved):
LGTM 👍🏼

PR #3789, opened 2021-10-07
Issue #5232: Docs/usage inconsistency: should theme() be used?

### LINKED ISSUE #5232: Docs/usage inconsistency: should theme() be used?
The [documentation for function `theme()`](https://docs.backdropcms.org/api/backdrop/core%21includes%21theme.inc/function/theme/1) says:

> All requests for themed output must go through this function (however, calling the theme() function directly is strongly discouraged...Avoid calling this function directly. It is preferable to replace direct calls to the theme() function with calls to backdrop_render() by passing a render array with a #theme key to backdrop_render(), which in turn calls theme().

However, there are nearly 500 calls to `theme('...', ...)` in core. (In fact, there are only about 400 calls to `backdrop_render()`, so the discouraged usage is more common than the blessed usage.) Even recent additions like [`theme_label_machine_name()`](https://docs.backdropcms.org/change-records/theme_file_type_overview-deprecated-in-favor-of-theme_label_machine_name) show a direct call to `theme()` as the usage example.

So should that initial discouragement of using `theme()` be removed from the documentation?

bugfolder:
Fixes https://github.com/backdrop/backdrop-issues/issues/5232.

### DIFF
--- a/core/includes/theme.inc
+++ b/core/includes/theme.inc
@@ -879,16 +879,14 @@ function backdrop_find_base_themes($themes, $key, $used_keys = array()) {
 /**
  * Generates themed output.
  *
- * All requests for themed output must go through this function (however,
- * calling the theme() function directly is strongly discouraged - see next
- * paragraph). It examines the request and routes it to the appropriate
- * @link themeable theme function or template @endlink, by checking the theme
- * registry.
+ * All requests for themed output must go through this function. It examines the
+ * request and routes it to the appropriate @link themeable theme function or
+ * template @endlink, by checking the theme registry.
  *
- * Avoid calling this function directly. It is preferable to replace direct
- * calls to the theme() function with calls to backdrop_render() by passing a
- * render array with a #theme key to backdrop_render(), which in turn calls
- * theme().
+ * You can call this function directly or make calls to backdrop_render() by
+ * passing a render array with a #theme key, which in turn calls theme(). Either
+ * is acceptable. In some situations a render array will be preferable when the
+ * ability for other modules to alter the render array is provided.
  *
  * @section sec_theme_hooks Theme Hooks
  * Most commonly, the first argument to this function is the name of the theme
--- a/core/includes/theme.inc
+++ b/core/includes/theme.inc
@@ -883,10 +883,11 @@ function backdrop_find_base_themes($themes, $key, $used_keys = array()) {
  * request and routes it to the appropriate @link themeable theme function or
  * template @endlink, by checking the theme registry.
  *
- * You can call this function directly or make calls to backdrop_render() by
- * passing a render array with a #theme key, which in turn calls theme(). Either
- * is acceptable. In some situations a render array will be preferable when the
- * ability for other modules to alter the render array is provided.
+ * You can call this function directly, or you can make a call to
+ * backdrop_render() by passing a render array with a #theme key, which in turn
+ * calls theme(). Either is acceptable. In some situations a render array will
+ * be preferable when the ability for other modules to alter the render array is
+ * provided.
  *
  * @section sec_theme_hooks Theme Hooks
  * Most commonly, the first argument to this function is the name of the theme


core/includes/theme.inc
```diff
@@ -879,16 +879,14 @@ function backdrop_find_base_themes($themes, $key, $used_keys = array()) {
...
- * paragraph). It examines the request and routes it to the appropriate
- * @link themeable theme function or template @endlink, by checking the theme
- * registry.
- *
- * Avoid calling this function directly. It is preferable to replace direct
- * calls to the theme() function with calls to backdrop_render() by passing a
- * render array with a #theme key to backdrop_render(), which in turn calls
- * theme().
+ * All requests for themed output must go through this function. It examines the
+ * request and routes it to the appropriate @link themeable theme function or
```
klonos, line 883: Lets please use Oxford commas in longer sentences:

```suggestion
 * request, and routes it to the appropriate @link themeable theme function or
```

core/includes/theme.inc
```diff
@@ -879,16 +879,14 @@ function backdrop_find_base_themes($themes, $key, $used_keys = array()) {
...
+ * template @endlink, by checking the theme registry.
+ *
+ * You can call this function directly or make calls to backdrop_render() by
```
klonos:
```suggestion
 * You can call this function directly, or make calls to backdrop_render() by
```

klonos (approved):
LGTM 👍🏼

core/includes/theme.inc
bugfolder: That's not an Oxford comma. Oxford commas occur in lists of 3 or more items. See https://www.grammar-monster.com/lessons/commas_the_Oxford_comma.htm for details.

core/includes/theme.inc
bugfolder: Again, only two items, so no comma.

core/includes/theme.inc
klonos: Yeah, perhaps "oxford comma" was not the right term here, but I have to admit that I wasn't aware of the 3 items rule either 🤔 ...I have been living wrong all this time 🤦🏼 

I just went and (re)read a bunch of articles on the matter:
- https://webapps.towson.edu/ows/modulecomma.htm
- https://owl.purdue.edu/owl/general_writing/punctuation/commas/extended_rules_for_commas.html
- https://www.businessinsider.com/commas-in-compound-predicates-2014-5
- https://english.stackexchange.com/questions/318958/comma-usage-with-one-subject-multiple-verbs-and-multiple-objects
- https://www.grammarly.com/blog/comma

...yet, I still feel somewhat confused. Some things make sense, while others still don't 🤷🏼 

Some articles mention that the use of comma as a "soft pause" is OK/correct (especially in long sentences with multiple verbs), others that it depends on the writing style/preference (so long as no grammatical rules are being violated, and the meaning of the text doesn't change), others that it can be used in cases where there's a chance that the reader may interpret the text differently, while other articles say that we should stick to the grammatical rules. ...I give up 😅

core/includes/theme.inc
klonos: Hmm 🤔 ...doesn't that way make it so that the "by passing a render array" bit applies to both? Whereas in reality it only applies to "or make calls to backdrop_render()". So:

- with the comma:
  - You can (either) call this function directly
  - or make calls to backdrop_render() by passing a render array...
- w/o the comma:
  - You can call this function directly [by passing a render array...]
  - or make calls to backdrop_render() by passing a render array...

So in the first case the sentence may be interpreted as: "the passing of a render array can be done either when you call this function directly or (also) when you make calls to backdrop_render()". Whereas by using a comma you are making it clear that "calling the function directly" is one thing, and "making calls to backdrop_render() by passing a render array..." is another.

...or do I have my grammar and punctuation all wrong? 🤔

bugfolder: IMHO (also [Grammarly's](https://www.grammarly.com/blog/comma-before-or/)) your grammar/punctuation is wrong, but your sense that a pause is needed is correct. We can satisfy both by making the dependent clause independent, which now requires a comma:

"You can call this function directly, or you can make a call to backdrop_render()..."

Just noticed that for parallel construction, it should be "a call," not "calls".

core/includes/theme.inc
quicksketch: The original structure reads correctly by my understanding. In any case the documentation fix is definitely an improvement. Let's proceed with it as-is.

quicksketch (approved):

quicksketch:

PR #3790, opened 2021-10-09
Issue #2626: [UX] Add a contextual menu for quick access to editing things like the layout, theme, etc. of the current page

### LINKED ISSUE #2626: [UX] Add links for quick access to editing things like the layout, theme, etc. of the current page

bugfolder:
Fixes https://github.com/backdrop/backdrop-issues/issues/2626.


core/modules/layout/layout.module
```diff
@@ -2243,6 +2243,94 @@ function layout_layout_presave($layout) {
...
+      'visible' => array(
+        ':input[name="use_layout_contextual_links"]' => array('checked' => TRUE),
+      ),
+    ),
+  );
+  $theme_default = config_get('system.core', 'theme_default');
+  $form['layout']['shifts']['default_theme_shift'] = array(
+    '#type' => 'checkbox',
+    '#title' => t('Shift for default theme'),
+    '#description' => t('Shift contextual links downward for all pages that display the admin bar in the default theme (%theme_default).', array('%theme_default' => $theme_default)),
```
klonos, line 2287: This gets us the machine name of the theme. Can we please have the theme title instead?

core/modules/layout/layout.module
```diff
@@ -2243,6 +2243,94 @@ function layout_layout_presave($layout) {
...
+    '#default_value' => $config->get('default_theme_shift'),
+  );
+  $admin_theme = config_get('system.core', 'admin_theme');
+  $form['layout']['shifts']['admin_theme_shift'] = array(
+    '#type' => 'checkbox',
+    '#title' => t('Shift for admin theme'),
+    '#description' => t('Shift contextual links downward for all pages that display the admin bar in the admin theme (%admin_theme).', array('%admin_theme' => $admin_theme)),
```
klonos, line 2294: Same as the previous comment, can we please have the theme title instead of its machine name?

core/modules/layout/layout.module
```diff
@@ -2243,6 +2243,94 @@ function layout_layout_presave($layout) {
...
+    '#default_value' => $config->get('include_theme_contextual_links'),
+    '#states' => array(
+      'visible' => array(
+        ':input[name="use_layout_contextual_links"]' => array('checked' => TRUE),
+      ),
+    ),
+  );
+  $form['layout']['shifts'] = array(
+    '#type' => 'fieldset',
+    '#title' => t('Shifts'),
```
klonos: If we know this to be an issue, shouldn't we be shifting the contextual links by default in the first place? (and get rid of this entire fieldset here)

Otherwise, can we please consider having this fieldset collapsed by default? (and perhaps come up with a better label for it too?)

core/modules/layout/layout.module
bugfolder: Done in updated PR.

core/modules/layout/layout.module
bugfolder: Done in updated PR.

core/modules/layout/layout.module
bugfolder: This gets a bit complicated, so let me address this in the issue.

core/modules/layout/layout.module
```diff
@@ -2243,6 +2243,97 @@ function layout_layout_presave($layout) {
...
+    '#states' => array(
+      'visible' => array(
+        ':input[name="use_layout_contextual_links"]' => array('checked' => TRUE),
+      ),
+    ),
+    '#collapsible' => TRUE,
+    '#collapsed' => TRUE,
+  );
+  $themes = list_themes();
+  $theme_default = $themes[config_get('system.core', 'theme_default')]->info['name'];
```
klonos, line 2283: I'm surprised this didn't break php5 tests. ...perhaps we don't have tests around this feature? 🤷🏼  ...perhaps I'm wrong though 🤔

core/modules/layout/layout.module
```diff
@@ -2243,6 +2243,97 @@ function layout_layout_presave($layout) {
...
+  $form['layout']['contextual_links_position']['default_theme_shift'] = array(
+    '#type' => 'checkbox',
+    '#title' => t('Shift for default theme'),
+    '#description' => t('Shift contextual links downward for all pages that display the admin bar in the default theme (%theme_default).', array('%theme_default' => $theme_default)),
+    '#default_value' => $config->get('default_theme_shift'),
+  );
+  $admin_theme = $themes[config_get('system.core', 'admin_theme')]->info['name'];
```
klonos, line 2290: Same as my previous comment re php5. I'll need to test this on my local with php5.

core/modules/system/system.install
```diff
@@ -3221,6 +3221,44 @@ function system_update_1077() {
...
+ */
+function system_update_1078() {
+  config_set('layout.settings', 'use_layout_contextual_links', TRUE);
+  config_set('layout.settings', 'include_theme_contextual_links', TRUE);
+
+  if (module_exists('layout_contextual_links')) {
+    config_set('layout.settings', 'default_theme_shift', config_get('layout_contextual_links.settings', 'default_theme_shift'));
+    config_set('layout.settings', 'admin_theme_shift', config_get('layout_contextual_links.settings', 'admin_theme_shift'));
+    module_disable(array('layout_contextual_links'));
+    backdrop_uninstall_modules(array('layout_contextual_links'));
```
klonos, line 3237: It's been a while since we've done this, but I believe the right way to go about ensuring that there are no conflicts with contrib modules that get merged into core is to update the array returned by [`backdrop_merged_modules()`](https://docs.backdropcms.org/api/backdrop/1/search/backdrop_merged_modules).

core/modules/layout/layout.module
bugfolder: Now that you mention it, since this appears on a newly created page, there should probably be a test created as part of this PR that loads the page.

klonos: Scratch that. I just tested with php 5.6.40 and this works fine, w/o any errors.

core/modules/layout/layout.module
klonos: Nope. Works fine in php 5.6.40.

core/modules/system/system.install
klonos: ...I'm pretty sure that this will cause the contrib module to not load if it exists in the codebase, but you should still be able to check via `$contrib_config = config('layout_contextual_links.settings');`, and only grab its config with `config_get()` if `$contrib_config` exists.

klonos: ...I just checked https://github.com/backdrop-contrib/layout_contextual_links/blob/1.x-1.x/layout_contextual_links.install and there are no install/uninstall hooks, so we should be good 👍🏼

bugfolder: The `$contrib_config` check works. We still need to uninstall the module in the update to not leave residue in the "Uninstall modules" tab on the modules page.

core/modules/layout/layout.module
bugfolder: Test that exercises this code now added, and it's passing 5.3 tests (just to confirm).

bugfolder:
Closing in favor of PR https://github.com/backdrop/backdrop/pull/3850.

PR #3793, opened 2021-10-18
Issue 5303: Date Theme automated test fails from 23:00 to 23:59 UTC

### LINKED ISSUE #5303: Date Theme automated test fails from 23:00 to 23:59 UTC
**Description of the bug**

Any PR that runs the automated tests between 23:00 and 23:59 UTC (i.e., between 4:00pm and 4:59pm in PDT) will fail the "Date Theme" test `DateThemeTestCase->testDateDisplayCombination()` at line 150.

**Steps To Reproduce**

To reproduce the behavior:

1. Enable Simpletest.
2. Wait until the current time is between 23:00 and 23:59 UTC (i.e., between 4:00pm and 4:59pm in PDT).
3. Run the "Date Theme" test.

Note that outside of this time window, the test will pass.

**Actual behavior**

The test fails. Specifically, the expected test will contain the string "7 days remaining" but the output value will be "6 days remaining".

**Expected behavior**

During this time interval, the correct value should be "6 days remaining."

**Additional information**

In the test code, the event start time is set 7 days out, but if the start time is >23:00, then the start time is pushed earlier by 1 hour (to keep the end time in the same day). But that means that there will only be 6 days until the event, since days remaining is rounded down.

It's an easy fix, but I question whether it's good practice to have an automated test in which the code branch executed depends on the time of day that the automated test happens to run.

We discovered this in https://github.com/backdrop/backdrop-issues/issues/5065.

bugfolder:
Fixes https://github.com/backdrop/backdrop-issues/issues/5303

To fully test this change, need to run tests between 23:00 and 23:59 UTC, and then sometime outside this window. (The current PR tests are running outside the window.)

### DIFF
--- a/core/modules/date/tests/date_themes.test
+++ b/core/modules/date/tests/date_themes.test
@@ -127,6 +127,10 @@ class DateThemeTestCase extends BackdropWebTestCase {
     if ($currentHour > 22) {
       // Make sure start and end are the same day.
       $dateTimeStart->modify('- 1 hour');
+      $days_remaining = 6; // Because it's one week less one hour.
+    }
+    else {
+      $days_remaining = 7;
     }
     $dateTimeStart->modify('+ 1 week');
     $dateTimeEnd = clone $dateTimeStart;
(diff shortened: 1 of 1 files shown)


klonos (approved):
LGTM 👍🏼

quicksketch:
Merged into 1.x and 1.20.x. Woohoo!

PR #3794, opened 2021-10-18
Issue #5305: Fix views exposed filter checkbox value handling

### LINKED ISSUE #5305: Fixed: PHP 8: Get views exposed filters with checkbox working
**Description of the bug**

Currently the default admin content view doesn't work on PHP 8.

**Steps To Reproduce**

1. Set your web to use PHP 8
2. Go to /admin/content
3. If you still see nodes (with a "new" marker) go to these node's pages and then come back to the admin overview
4. Add new nodes, one by one
5. Go back to /admin/content

**Actual behavior**

The result is always empty, if there are no "new" nodes, no matter which filters are set, or if filters are set at all. Only nodes with the "new" marker appear. And they vanish, as soon as you visited their pages.

**Expected behavior**

Correct list of content.

**Additional information**

TLDR: `var_dump('All' == 0)` returns true in PHP prior to version 8, false since 8.

So this check in function accept_exposed_input() worked in previous PHP versions, but is now broken, which is why the filter always gets added to the query.

indigoxela:
Fixes https://github.com/backdrop/backdrop-issues/issues/5305: Fix views exposed filter checkbox value handling

Don't expect all checks on PHP 8 to pass - this is just another baby-step.

<s>Todo: revert the temporary switch to php 8.0 in GHA.</s> (done)

### DIFF
--- a/.github/workflows/functional-tests.yml
+++ b/.github/workflows/functional-tests.yml
@@ -10,7 +10,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        php-versions: ['5.3', '8.0']
+        php-versions: ['5.3', '7.4']
         fraction: ['1/3', '2/3', '3/3']
         database-versions: ['mariadb-10.3']
 
--- a/.github/workflows/functional-tests.yml
+++ b/.github/workflows/functional-tests.yml
@@ -10,7 +10,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        php-versions: ['5.3', '7.4']
+        php-versions: ['5.3', '8.0']
         fraction: ['1/3', '2/3', '3/3']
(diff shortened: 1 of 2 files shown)


albanycomputers:
I've just tested and can confirm the problem... The content disappears after it has been viewed.

albanycomputers:
Just tested and it works for me :-)

indigoxela:
@albanycomputers comments like these are more useful directly on the issue - they might get missed, if added to the PR. Comments on PRs are usually coding specific. Anyway, many thanks for taking a look and testing. :+1:

klonos (approved):
LGTM 👍🏼

quicksketch:
Merged into 1.x and 1.20.x. Thanks @indigoxela, @klonos, and @albanycomputers!

PR #3796, opened 2021-10-18
Issue #5304: Finalize incomplete port of form cache test.

### LINKED ISSUE #5304: Form cache test not properly ported from Drupal
This is a follow up to Issue #3375: Make cache_form expiration configurable, to mitigate runaway cache_form tables.
    
    By @klonos and @herbdool.
    
`FormsFormCacheTestCase` was added to the codebase, but never adapted from Drupal. It extends `DrupalWebTestCase` instead of `BackdropWebTestCase` and still has a `getInfo()` method. It was never added into `simpletest.tests.info` so has likely never been run. 

I stumbled across this while working on another issue that adds a test to this file. When I did a find/replace for DrupalWebTestCase/BackdropWebTestCase I caught an extra one I wasn't expecting!

indigoxela:
Fixes https://github.com/backdrop/backdrop-issues/issues/5304

### DIFF
--- a/core/modules/simpletest/tests/form.test
+++ b/core/modules/simpletest/tests/form.test
@@ -1445,23 +1445,20 @@ class FormsFormStoragePageCacheTestCase extends BackdropWebTestCase {
 /**
  * Test cache_form.
  */
-class FormsFormCacheTestCase extends DrupalWebTestCase {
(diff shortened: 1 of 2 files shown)


quicksketch:

PR #3801, opened 2021-10-20
Issue #5315: Fix function description for system_theme_settings()

### LINKED ISSUE #5315: [DX] Fix function description for system_theme_settings()
It currently says this (see https://github.com/backdrop/backdrop/blob/1.x/core/modules/system/system.admin.inc#L400):

```php
/**
 * Form builder; display theme configuration for entire site and individual themes.
 *
```

(btw, this line also exceeds the 80 characters limit)

...but in #130 we've removed the global theme settings page at `admin/appearance/settings`, so the description should be changed to:

`Form builder; display configuration for individual themes.`

klonos:
Fixes https://github.com/backdrop/backdrop-issues/issues/5315

### DIFF
--- a/core/modules/system/system.admin.inc
+++ b/core/modules/system/system.admin.inc
@@ -397,7 +397,7 @@ function system_theme_default() {
 }
 
 /**
- * Form builder; display theme configuration for entire site and individual themes.
+ * Form builder; display configuration for individual themes.
  *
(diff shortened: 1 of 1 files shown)


quicksketch:
Nice. Merged into 1.x and 1.22.x. Thanks!

PR #3807, opened 2021-10-22
Issue #5326: Update cache flush doesnt work

### LINKED ISSUE #5326: Fixed: Update cache flush doesnt work
**Description of the bug**

Clicking "Update data" on admin bar should call `_update_cache_clear()` which should truncate the `cache_update` table and clear the update cache. It doesn't. This is because the menu system passes the string `"update" ` as the `$cid` to `_update_cache_clear()`, and since that cache id doesnt exist, nothing happens.
```
function _update_cache_clear($cid = NULL, $wildcard = FALSE) {      <---$cid is set to "update"
  if (empty($cid)) {                 <-- so skipped.
    db_delete('cache_update')
      ->execute();
  }
  else {
    $query = db_delete('cache_update');
      $query->condition('cid', $cid);      <---there is no cache named "update" so also skipped.
    $query->execute();
  }
}
``` 

**Steps To Reproduce**

Check the `cache_update` table is populated, then clear update cache; nothing happens. It should empty.

**Actual behavior**

Nothing happens

**Expected behavior**

The table should empty.

docwilmot:
Fixes https://github.com/backdrop/backdrop-issues/issues/5326

### DIFF
--- a/core/modules/update/update.module
+++ b/core/modules/update/update.module
@@ -692,7 +692,9 @@ function _update_get_cache_multiple($cid_prefix) {
  *   the exact cache ID specified by $cid. Defaults to FALSE.
  */
 function _update_cache_clear($cid = NULL, $wildcard = FALSE) {
-  if (empty($cid)) {
+  // When clearing caches via Admin Bar links, the string "update" is passed to
+  // this function as $cid by the menu system.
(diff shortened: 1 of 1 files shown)


quicksketch:

PR #3816, opened 2021-11-08
Issue #3583: Set autofocus on title only when creating nodes

### LINKED ISSUE #3583: Node's `Title` field gets focus regardless of weight
**Description of the bug**
I have a content type with the Title field (renamed to 'Description') sitting below other fields (ordered on the `/admin/structure/types/manage/[TYPE]/fields` page). When I go to create content of that type, the title/description field has the focus, even though it's the last/bottom field on the form. This seems weird and undesirable (for my situation).

**Steps To Reproduce**
To reproduce the behavior:
1. Create a new content type
2. Add a few custom fields
3. Arrange the fields so that Title is the last one (i.e. has the highest weight)
4. Create a node of that type

**Actual behavior**
The Title field (at the bottom of the form) has the focus.

**Expected behavior**
The top/first field gets the focus.

**Additional information**
Possible related issue: https://github.com/backdrop/backdrop-issues/issues/2625

indigoxela:
Fixes https://github.com/backdrop/backdrop-issues/issues/3583

### DIFF
--- a/core/modules/node/node.module
+++ b/core/modules/node/node.module
@@ -3498,10 +3498,11 @@ function node_content_form(Node $node, $form_state) {
       '#default_value' => $node->title,
       '#maxlength' => 255,
       '#weight' => -5,
-      '#attributes' => array(
-        'autofocus' => TRUE,
-      ),
     );
+    // Set the autofocus attribute for new nodes.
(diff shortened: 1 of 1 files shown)


indigoxela:
Not enough consensus.

PR #3821, opened 2021-11-12
Issue #5350: Clean up permissions when deleting content type

### LINKED ISSUE #5350: Make sure that any content type permissions are removed from roles when the respective content type is deleted
## Steps To Reproduce
1. create a new content type
2. on the "Permissions" vertical tab, assign some permissions to the "Editor" role
3. save the content type
4. navigate to `admin/config/development/configuration/single/export`
5. select "User roles" in the "Configuration group" drop-down
6. select "Administrator" in the "Configuration name" drop-down
7. scroll to the end of the permissions list shown in the "Exported configuration" field -> verify that the permissions for the new content type are there 👍🏼 
8. switch to the "Editor" role via the "Configuration name" drop-down
9. scroll to the end of the permissions list shown in the "Exported configuration" field -> verify that the permissions for the new content type are there 👍🏼 
10. delete the content type you have created for the purposes of this test
11. navigate to `admin/config/development/configuration/single/export` and repeat steps 5 through 9 above...

## Actual behavior

The permissions for the content type that was just deleted are still there for all the roles they were configured for.

## Expected behavior

The permissions for the content type that was just deleted should be removed from all the roles they were configured for.

## Additional information

- If we decide that this is a valid request, then we should make sure that permissions are removed from roles from other places, such as:
  - when deleting text formats
  - when deleting vocabularies
  - ???

  These should be separate issues though.

- In Drupal 9.3.x with the introduction of https://www.drupal.org/node/3055548 such permissions are being cleaned up automatically, because permissions are allowed to specify dependencies.

klonos:
Fixes https://github.com/backdrop/backdrop-issues/issues/5350


core/modules/simpletest/backdrop_web_test_case.php
```diff
@@ -1354,7 +1354,7 @@ protected function backdropCreateRole(array $permissions, $name = NULL) {
       $name = $this->randomName();
     }
 
-    // Check the all the permissions strings are valid.
+    // Check that all the permission strings are valid.
     if (!$this->checkPermissions($permissions)) {
```
klonos, line 1358: If I change this bit here to this, those nasty "Invalid permission" failures go away and everything is green 🤔 

```suggestion
    if (!$this->checkPermissions($permissions, TRUE)) {
```

I've raised https://github.com/backdrop/backdrop-issues/issues/5354 to discuss this further.

core/modules/simpletest/backdrop_web_test_case.php
```diff
@@ -1347,8 +1347,9 @@ protected function backdropCreateRole(array $permissions, $name = NULL) {
       $name = $this->randomName();
     }
 
-    // Check the all the permissions strings are valid.
-    if (!$this->checkPermissions($permissions)) {
+    // Check that all the permission strings are valid.
+    // @todo: Remove this change when we sort out https://github.com/backdrop/backdrop-issues/issues/5354
```
jenlampton, line 1351: Let's be more specific so we don't accidentally delete the whole check...
`// @todo: Remove the TRUE parameter`

PR #3823, opened 2021-11-13
Issue #5354: Allow permission refresh when calling backdropCreateRole()

### LINKED ISSUE #5354: [DX] Tests: always force permission refresh when calling backdropCreateRole()
I was working on adding some nice tests in #5350, where I am:
1. creating a new content type.
2. calling `node_type_cache_reset();` to make sure that the new content type is available for the rest of the test.
3. attempting to gather the permissions relevant to the new content type, by calling `node_list_permissions($node_type)` - adding a `dpm()` at this point reveals that the array of all 5 node type permissions is successfully assigned to the variable I am using 👍🏼 
4. attempting to create a new user role, and assign it those 5 node permissions from the previous step - at this point I am getting 5 test failures of "Invalid permission" 😞 

After (A LOT of) local debugging, I realized that any remaining failures would pass if I change this bit in `backdropCreateRole()`:

```php
    // Check that all the permission strings are valid.
    if (!$this->checkPermissions($permissions)) {
      return FALSE;
    }
```

...to this:

```php
    // Check that all the permission strings are valid.
    if (!$this->checkPermissions($permissions, TRUE)) {
      return FALSE;
    }
```

The addition of that second `TRUE` parameter there forces `checkPermissions()` to refresh available permissions (by calling `module_invoke_all('permission')`), which solved all the "Invalid permission ..." failures I was getting 🎉 

Unfortunately, there is no way to pass that parameter to `checkPermissions()` via `backdropCreateRole()`, and calling `backdrop_static_reset('user_roles');` and/or `backdrop_static_reset('user_access');` before `backdropCreateRole()` does not help either.

Would it be reasonable to request that we add another `$reset = FALSE` parameter to `backdropCreateRole()`, which is passed on to `checkPermissions()`, to cover for these cases?

klonos:
Fixes https://github.com/backdrop/backdrop-issues/issues/5354


core/modules/simpletest/backdrop_web_test_case.php
```diff
@@ -1377,14 +1381,16 @@ protected function backdropCreateRole(array $permissions, $name = NULL) {
...
-   * @return
-   *   TRUE or FALSE depending on whether the permissions are valid.
+   * @param array $permissions
+   *   Array of permission names to validate.
+   * @param bool $reset
+   *   (optional) Flag to force a reset of cached available permissions.
+   *   Defaults to FALSE.
+   *
+   * @return bool
+   *   TRUE or FALSE, depending on whether the permissions are valid ot not.
```
klonos:
```suggestion
   *   TRUE or FALSE, depending on whether the permissions are valid or not.
```

PR #3831, opened 2021-11-22
Issue #5290 PHP7.4 and empty entity_keys for taxonomy term result in notices

### LINKED ISSUE #5290: PHP7.4 and empty entity_keys for taxonomy term result in notices
**Description of the bug**

Errors when navigating taxonomy entities during special scenarios when using PHP 7.4

```
Notice: Trying to access array offset on value of type null in entity_extract_ids()
(line 353 of /core/modules/entity/entity.module).
Notice: Trying to access array offset on value of type null in entity_extract_ids()
(line 353 of /core/modules/entity/entity.module).
Notice: Trying to access array offset on value of type null in entity_extract_ids()
(line 354 of /core/modules/entity/entity.module).
Notice: Trying to access array offset on value of type null in entity_extract_ids()
(line 354 of /core/modules/entity/entity.module).
```

**The Fix**

Apply the same fix as https://www.drupal.org/project/drupal/issues/3166668

**Steps To Reproduce**

To reproduce the behavior:
1) Ensure PHP 7.4 is used
2) Install display_suite and display_suite_ui from: https://github.com/backdrop-contrib/ds
3) Add a term to a vocabulary
4) Navigate to **Manage Display** and set the "select layout" to something other than **- None -** and save the display
5) Navigate to the term page and witness the errors

**Additional information**

See issue https://github.com/backdrop-contrib/ds/issues/5 for additional details where this problem was found.

Excerpt

> It seems that whenever a taxonomy term is rendered in a custom view mode it comes with empty entity_keys and this causes the lines 8000 and 8001 in includes/common.inc to through the above notices. The same notices are produced in the "Manege Display" page of the same content type.
This might have to do with Display Suite through which the custom view modes are created, but the issue would remain that the entity_extract_ids function can not cope with empty entity keys under PHP 7.4.

**Personal reference**
Backdrop location: /core/modules/entity/entity.module
Line: 352

philsward:
Fixes: https://github.com/backdrop/backdrop-issues/issues/5290

Proposed fix from Drupal issue: https://www.drupal.org/project/drupal/issues/3166668

### DIFF
--- a/core/modules/entity/entity.module
+++ b/core/modules/entity/entity.module
@@ -351,8 +351,8 @@ function entity_extract_ids($entity_type, $entity) {
 
   // Objects being created might not have id/vid yet.
   if (!empty($info['entity keys'])) {
-  $id = isset($entity->{$info['entity keys']['id']}) ? $entity->{$info['entity keys']['id']} : NULL;
-  $vid = ($info['entity keys']['revision'] && isset($entity->{$info['entity keys']['revision']})) ? $entity->{$info['entity keys']['revision']} : NULL;
+    $id = isset($entity->{$info['entity keys']['id']}) ? $entity->{$info['entity keys']['id']} : NULL;
+    $vid = ($info['entity keys']['revision'] && isset($entity->{$info['entity keys']['revision']})) ? $entity->{$info['entity keys']['revision']} : NULL;
   }
   else {
     $id = NULL;
--- a/core/modules/entity/entity.module
+++ b/core/modules/entity/entity.module
@@ -350,8 +350,14 @@ function entity_extract_ids($entity_type, $entity) {
   $info = entity_get_info($entity_type);
 
   // Objects being created might not have id/vid yet.
+  if (!empty($info['entity keys'])) {
   $id = isset($entity->{$info['entity keys']['id']}) ? $entity->{$info['entity keys']['id']} : NULL;
   $vid = ($info['entity keys']['revision'] && isset($entity->{$info['entity keys']['revision']})) ? $entity->{$info['entity keys']['revision']} : NULL;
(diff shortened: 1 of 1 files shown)


core/modules/entity/entity.module
```diff
@@ -350,8 +350,14 @@ function entity_extract_ids($entity_type, $entity) {
   $info = entity_get_info($entity_type);
 
   // Objects being created might not have id/vid yet.
+  if (!empty($info['entity keys'])) {
   $id = isset($entity->{$info['entity keys']['id']}) ? $entity->{$info['entity keys']['id']} : NULL;
   $vid = ($info['entity keys']['revision'] && isset($entity->{$info['entity keys']['revision']})) ? $entity->{$info['entity keys']['revision']} : NULL;
```
klonos: This code looks good @philsward, but you will need to indent these 2 lines by 2 spaces, to adhere to our coding standards.
```suggestion
    $id = isset($entity->{$info['entity keys']['id']}) ? $entity->{$info['entity keys']['id']} : NULL;
    $vid = ($info['entity keys']['revision'] && isset($entity->{$info['entity keys']['revision']})) ? $entity->{$info['entity keys']['revision']} : NULL;
```

klonos (approved):
LGTM 👍🏼

klonos:
I've tested things on my local (our PR sandboxes run on php 7.2 - not 7.4), and the 4 errors that are related to core code are gone with the changes applied. The other 2 errors are specific to Display Suite code, and will be fixed in https://github.com/backdrop-contrib/ds/issues/5

quicksketch:
Merged into 1.x and 1.20.x. Thanks @philsward!

PR #3838, opened 2021-11-30
Issue #3879: Status report page: Show jQuery and jQuery UI as info instead of status

klonos:


klonos:
...something terribly messed with this branch. Closing PR.

PR #3839, opened 2021-11-30
Issue #4342: Opening modal popups should not scroll the page

### LINKED ISSUE #4342: [UX] Modal popups jump to the top of the page when opened
**Description of the bug**
When you open a modal popup (e.g. the Image Library browser), the page behind the modal jumps up to the top of the page and, when you close the modal, you have to scroll back down to where you were previously. This is very frustrating.

**Steps To Reproduce**
To reproduce the behavior:
1. On a new sandbox site, edit the 'Your first post' node: `/node/1/edit`
2. Scroll down to the Image field and click 'Select existing file'
3. Note the page behind has jumped back to the top
4. Close the modal popup and groan in frustration as you scroll back down to the image field

Screencast:
Peek 2020-03-14 14-18

indigoxela:
Fixes https://github.com/backdrop/backdrop-issues/issues/4342

### DIFF
--- a/core/misc/dialog.theme.css
+++ b/core/misc/dialog.theme.css
@@ -2,16 +2,6 @@
  * Presentational styles for Backdrop dialogs.
  */
 
-/* Lock scrolling when an auto-resize dialog is open. */
-html.dialog-auto-resize {
-  overflow: hidden;
-  height: 100%;
-  left: 0;
-  position: fixed;
-  top: 0;
-  width: 100%;
-}
-
 .ui-dialog {
   position: absolute;
   z-index: 1260;
--- a/core/misc/collapse.js
+++ b/core/misc/collapse.js
@@ -7,6 +7,7 @@ Backdrop.toggleFieldset = function (fieldset) {
   var $fieldset = $(fieldset);
   if ($fieldset.is('.collapsed')) {
     var $content = $('> .fieldset-wrapper', fieldset).hide();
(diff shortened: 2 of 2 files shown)


core/misc/dialog.theme.css
```diff
@@ -2,16 +2,6 @@
  * Presentational styles for Backdrop dialogs.
  */
 
-/* Lock scrolling when an auto-resize dialog is open. */
-html.dialog-auto-resize {
```
docwilmot, line 6: Why delete this? If deleting is required though, this class is only added in `dialog.js`, so should we remove that code too? 

As far as I can see, the only effect of this CSS is that without it, if you try to scroll while the dialog is resizing, the dialog achieves its final size off center, then animates to center, whereas if scrolling is locked it achieves its final size dead center. You have to try hard to get this effect, so we can live without it if this CSS causes other problems.

indigoxela: > Why delete this?

Because this is the change that causes the jumping to top of page. So reverting this is essential.

> the only effect of this CSS is that without it, if you try to scroll while the dialog is resizing,

That's not correct. This "lock scrolling" causes the regression. I wouldn't remove the class, as it potentially is in use in some custom or contrib styles. And the class per se does no harm.

docwilmot: In that case this is ready then.

quicksketch:
Merged into 1.x and 1.20.x. Woo!

PR #3842, opened 2021-12-01
Issue #5380: Add 'empty' option to theme_item_list()

### LINKED ISSUE #5380: [DX][D8] Add 'empty' option to theme_item_list()
This is the same as https://www.drupal.org/project/drupal/issues/2120807, and it is going to be an API addition for Backdrop.

Change record: https://www.drupal.org/node/2124021

> `theme('item_list')` (or the `theme_item_list()` function, as it is currently implemented) has a new feature: You can pass in variable `'empty'` with a string or render array value, and if there are no list items, this value will be output. This is similar to the `'empty'` variable in `theme('table')`.
> 
> Example of using this from a render array:
> 
> ```php
> $build['foo'] = array(
>   '#theme' => 'item_list',
>   // This is the list of items to output.
>   '#items' => $results,
>   // This can be a string or a render array.
>   '#empty' => array(
>       '#markup' => t('There were no results. Try a less restrictive filter.'),
>   ),
>   '#title' => t('Search Results'),
>   '#list_type' => 'ol',
> );
> ```

klonos:
Fixes https://github.com/backdrop/backdrop-issues/issues/5380

### DIFF
--- a/core/includes/common.inc
+++ b/core/includes/common.inc
@@ -7814,7 +7814,7 @@ function backdrop_common_theme() {
       'variables' => array('type' => MARK_NEW),
     ),
     'item_list' => array(
-      'variables' => array('items' => array(), 'title' => '', 'type' => 'ul', 'attributes' => array()),
+      'variables' => array('items' => array(), 'title' => '', 'type' => 'ul', 'attributes' => array(), 'empty' => NULL),
     ),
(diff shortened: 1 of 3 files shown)


quicksketch:
Merged into 1.x for 1.21.0. Thanks!

PR #3837, opened 2021-11-30
Fixes field name not being shown in 'required' validation for multiple value fields

argiepiano:
Fixes https://github.com/backdrop/backdrop-issues/issues/5377

This code is taken from a Drupal patch found [here](https://www.drupal.org/files/issues/980144-114.patch)

PR #3843, opened 2021-12-02
Improve validation error for decimal multi-value fields

### LINKED ISSUE #5378: Confusing validation error for multi-value decimal field
## Description of the bug

An unclear validation error "_field_x_ is not a valid number" is thrown when entering values with too many decimals in a decimal field with multiple cardinality. Two issues: 

- instead of the "human" field name, the machine name of the field is shown (this is related to #5377), 
-  the "is not a valid number" error could be more clear. The Decimal field was configured to accept only two decimals, and the error is shown if the user enters more than 2 decimals.  

This error only happens ONLY if you click "Add another item" after entering a faulty number with more than 2 decimals.
Screen Shot 2021-11-29 at 6 09 12 PM

## Steps To Reproduce

1. Add a field of type Number (Decimal) to a node
2. Configure the field with a "Scale" value of 2 (the default) 
3. Configure the field for Unlimited cardinality
4. Create a new node and enter a number with more than two decimals
5. Click "Add another item" and watch the error appear

## Actual behavior

The form shows the error in the image above.

## Expected behavior

The form should show a validation error that at least contains the name of the field it's referring to (in the attached image, it should refer to field "Decimal Multiple"), and perhaps more clarity with the problem with the entered number.   

## Additional information

- Backdrop CMS version: 1.20.2

argiepiano:
Fixes [#5378](https://github.com/backdrop/backdrop-issues/issues/5378)

Please check if this makes sense - I'm not a native speaker!

### DIFF
--- a/core/includes/form.inc
+++ b/core/includes/form.inc
@@ -4594,7 +4594,12 @@ function form_validate_number(&$element, &$form_state) {
     $offset = isset($element['#min']) ? $element['#min'] : 0.0;
 
     if (!valid_number_step($value, $element['#step'], $offset)) {
-      form_error($element, t('%name is not a valid number.', array('%name' => $name)));
+      $number_of_digits = strlen(substr($element['#step'], strpos($element['#step'], '.') + 1));
+      if ($number_of_digits == 0) {
+        form_error($element, t('%name is configured to accept no decimals.', array('%name' => $name)));
+      } else {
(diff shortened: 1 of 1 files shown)


argiepiano:
Oops. It looks like there are some tests that actually check for the old validation error language, which just failed. Oh well... 

If there is a strong feeling about changing the validation language I can rewrite the test. Otherwise, I think our energy is better spent somewhere else. At least we fixed the most pressing issue in PR #3837

PR #3845, opened 2021-12-02
Issue #5381: Form API: Automatically trim text input by default

### LINKED ISSUE #5381: [DX][UX][D8] Form API: Automatically trim leading/trailing whitespace from all text input fields by default
This is a follow-up to #5348 and the respective of https://www.drupal.org/project/drupal/issues/67769

> ## Problem
> - Form handlers _and_ API code manually calls `trim()` on values too many times.
> - It's unclear who is responsible for trimming text values / user input.
> - The lack of custom `trim()` calls can trigger error messages that are hard to recognize and resolve by a user, in case a submitted form value contains leading or trailing white-space (which is not visible/apparent).
> 
> ## Proposed solution
> 1. Introduce a generic `#trim` property for all form elements that accept `#input`.
> 1. Make `#trim` default to `TRUE` for all form element types that accept text as user input.
> 1. Process `#trim` directly in `form_handle_input_element()`. If the resulting `#value` is an empty string, trigger the regular processing flow that would be triggered for no user input - including `#required` validation.
> 1. Allow code that does not want this behavior to opt-out by setting `#trim` to `FALSE`. (There's no use-case for that in core, but still...)
> 1. Remove all manual calls of `trim()` from API code. Whoever is calling into APIs is expected to provide proper (trimmed) values.

---

> ## Original summary
> If the user registers and by mistake includes and extra space before or after their email address, the system will reject the email address with an error message that is difficult to understand (as there is nothing obviously wrong with the email address). I have seen this several times when I paste my email address into the email address box and mistakenly include a trailing space. I just tried it on the Drupal.org site and got the error:
> 
> > The e-mail address testaccount@test.org is not valid.
> 
> My guess is that this problem might affect 2-3% of all users who ever try to set up Drupal accounts and leave them a bit frustrated.
> 
> It would be very easy to have the form on the page user/register trim the spaces off of the email address, to solve this problem.

klonos:
Fixes https://github.com/backdrop/backdrop-issues/issues/5381


core/includes/form.inc
```diff
@@ -2133,6 +2133,18 @@ function _form_builder_handle_input_element($form_id, &$element, &$form_state) {
       // be explicitly distinguished from missing input. (see below)
       $input_exists = NULL;
       $input = backdrop_array_get_nested_value($form_state['input'], $element['#parents'], $input_exists);
+      // Trim any leading/trailing whitespace from text input fields, unless
+      // their default '#trim' property has been overridden. To see which types
+      // of input elements are being trimmed by default, see
+      // system_element_info().
+      if (isset($element['#trim']) && $element['#trim'] != FALSE) {
```
quicksketch, line 2145: Might also be good to check that `$input` is actually a string here.

quicksketch (commented):
Why no `#trim` on `$types['password']`?

core/includes/form.inc
klonos: A bit superfluous, since the `#trim` property is by default only added to form elements that have text input, but you can't be too sure I guess - lets not allow people to shoot themselves on the foot.

klonos: ...done with [`928b7b6` (#3845)](https://github.com/backdrop/backdrop/pull/3845/commits/928b7b60ed022d28f7477452b8b5c5d52ce3db1c)

PR #3852, opened 2021-12-05
Issue 5384: Permissions matrix cut off on Permissions tab

### LINKED ISSUE #5384: [UX] Permissions matrix cut off on Permissions tab of Create Content page
## Description of the bug

This was [first reported in the Field Permissions issue queue](https://github.com/backdrop-contrib/field_permissions/issues/6), but it's actually a core issue.

If there are many roles, then on the Permissions tab of the Add Content Type page (admin/structure/types/add), some of the permissions are cut off on the right side with no way to reach them.

## Steps To Reproduce

To reproduce the behavior:
1. Create one or more roles with really long single-word names, e.g., "Supercalifragilisticexpialodocious."
2. Go to admin/structure/types/add.
3. Click on the Permissions tab.

## Actual behavior

Rightmost permissions are cut off.

sample

## Expected behavior

The permissions matrix should scroll as needed, so a little CSS tweak should fix it.

bugfolder:
Fixes https://github.com/backdrop/backdrop-issues/issues/5384.

### DIFF
--- a/core/modules/node/css/node.admin.css
+++ b/core/modules/node/css/node.admin.css
@@ -1,4 +1,3 @@
-
 /**
  * @file
  * Styles for administration pages.
@@ -9,4 +8,11 @@
  */
 .revision-current {
   background: #ffc;
+}
+
+/**
+ * Allow permissions matrix to scroll on "Add Content Type" form.
+ */
+#edit-permissions > .fieldset-wrapper > .form-item {
+  overflow-x: scroll;
 }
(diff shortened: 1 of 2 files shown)


core/modules/node/css/node.admin.css
```diff
@@ -10,3 +9,10 @@
 .revision-current {
   background: #ffc;
 }
+
+/**
+ * Allow permissions matrix to scroll on "Add Content Type" form.
+ */
+#edit-permissions > .fieldset-wrapper > .form-item {
+  overflow-x: scroll;
```
indigoxela: If setting overflow-x to scroll, the scrollbar below will always be there, even if not necessary.
I belief, this should be "auto", so the scrollbar only appears when needed.

bugfolder: Indeed, thanks! Now fixed.

PR #3857, opened 2021-12-11
Issue 4689: Negative visibility condition for roles

### LINKED ISSUE #4689: Negative visibility condition for roles
**Description of the need**

I would like to be able to set a visibility rule for conditions in which a user does not have a specific role. You can set this kind of role in Rules with the "negate" option. 

**Proposed solution**

Add a negate option to visibility conditions. 

**Additional information**

Something like this (this is a "condition" from the rules module):

Add_a_new_condition___Simplo

Found out that we have the "negate" option on UID and NID visibility condition. Would it be difficult to add it to Roles?

Sites___Simplo1

This also came up in the forum (OK, I posted it in both places). 
https://forum.backdropcms.org/forum/visibility-rule-user-does-not-have-role

bugfolder:
Fixes https://github.com/backdrop/backdrop-issues/issues/4689.

### DIFF
--- a/core/modules/layout/plugins/access/user_role_layout_access.inc
+++ b/core/modules/layout/plugins/access/user_role_layout_access.inc
@@ -3,7 +3,7 @@
  * @file
  * Plugin to provide access control based upon role membership.
  */
-class UserRoleLayoutAccess extends LayoutAccess {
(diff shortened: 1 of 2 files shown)


quicksketch:
Merged into 1.x for 1.21.0. Thanks!

PR #3861, opened 2021-12-12
Issue #5365: Improve system_transliterate_machine_name

### LINKED ISSUE #5365: [DX]: Fixed: Add default value for options in `system_transliterate_machine_name()`.
**Description of the bug**

Two issues:
- In `system_transliterate_machine_name($string, $options)` the `$options` parameter needs a default value, since its not required. In other words, calling `system_transliterate_machine_name('string')` causes a fatal; you have to explicitly set an empty options array: `system_transliterate_machine_name('string', array())`. The funtion should be defined as `system_transliterate_machine_name($string, $options = array())` .
- Not really a bug, but this function is really useful to create machine names from entered strings; the `machine_name` element does this with JS (`core/misc/machine-name.js`) and there is no other PHP function that does this, AFAIK. So we should move it to `system.module` not `system.admin.inc` so other modules can use it without including that inc file.

ghost:
Fixes https://github.com/backdrop/backdrop-issues/issues/5365


core/modules/system/system.module
```diff
@@ -4561,3 +4561,59 @@ function system_theme_debug_enabled_warning() {
     backdrop_set_message(t('Theme debugging is enabled. See the <a href="!url">Status report</a> page for more info.', array('!url' => url('admin/reports/status'))), 'warning', FALSE);
   }
 }
+
+/**
+ * Wrapper around transliteration_get() that handles common clean up on strings.
+ *
+ * @param string $string
+ *   The string to be transliterated.
+ * @param $options
```
klonos: Let's please add the variable type here:
```suggestion
 * @param array $options
```

core/modules/system/system.module
```diff
@@ -4561,3 +4561,59 @@ function system_theme_debug_enabled_warning() {
...
+ *   An array of options for further cleanup. This array includes:
+ *   - langcode: The language code of the source string.
+ *   - replace: The character with which to replace unknown characters in the
+ *     string. Defaults to '_'.
+ *   - replace_pattern: A pattern for replacing unknown characters with the
+ *     "replace" value.
+ *   - maxlength: The maximum allowed length of the returned string.
+ * @return string
```
klonos, line 4579: Can we please add a blank line to separate parameters from the return value?:
```suggestion
 *
 * @return string
```

docwilmot: Those were preexisting. If @najmieh has time to do so, otherwise this can be a followup issue to work on core docblocks generally.

core/modules/system/system.module
docwilmot: Those were preexisting. If @najmieh has time to do so, otherwise this can be a followup issue to work on core docblocks generally.

core/modules/system/system.module
ghost: Yes, I can do it.

> Can we please add a blank line to separate parameters from the return value?:

I need to work on this?

quicksketch:
Merged into 1.x and 1.20.x. Thanks @najmieh!

PR #3862, opened 2021-12-13
Issue #4903 - New Content Type with Content

### LINKED ISSUE #4903: Provide a new hidden-path ("Card") content type in core

stpaultim:
Fixes backdrop/backdrop-issues#4903


core/themes/basis/css/skin.css
```diff
@@ -1226,3 +1226,24 @@ div.error:before {
   /* background-image: -webkit-linear-gradient(transparent, transparent), url(../../../misc/message-error.svg); */
   /* background-image: linear-gradient(transparent, transparent), url(../../../misc/message-error.svg); */
 }
+
+/* Promobox view on default home page */
```
olafgrabienski: The term "Promobox" seems to be a leftover from an earlier state. Better change to "Card".

olafgrabienski (commented):
Was there a particular reason changing to NID? The creation date has one advantage, it can be changed by editors, so they have a workaround if they want to change the order of cards in the grid.

stpaultim:
@olafgrabienski - I switched to using NID, because they just weren't displaying in the right order in Tugboat when I sorted by time. They displayed in the right order locally, but not on Tugboat. I assume that maybe they are all getting created at exactly the same time on Tugboat and the date/time sort isn't working?

I did a little experimenting, but I can try some more later. @docwilmot or @BWPanda - any thoughts?

ghost:
> I assume that maybe they are all getting created at exactly the same time on Tugboat and the date/time sort isn't working?

Sounds likely...

docwilmot:
Agreed, but neither are very good ways to sort, so I don't think we're missing much. 

Thought: is it too much to add a "Front grid sort" number field to the card content type? We could then sort by the field and filter by field has value. It would also solve the problem that the card cannot be used elsewhere without showing up in the View.

stpaultim:
I'm not eager to add another field for this purpose. But, it's worth considering. It might be the best solution for end users, as @docwilmot points out, neither date nor NID are very logical ways of sorting.

olafgrabienski:
> is it too much to add a "Front grid sort" number field to the card content type? We could then sort by the field and filter by field has value

Interesting, but I *think* it works only at the beginning. When people add more cards it doesn't work until they remove the number field value of older ones. Or did I miss something?

docwilmot:
To be honest I'm thinking more and more the front page layout should just have 3 columns. People are going to want to change things. I really think a person trying out Backdrop for the first time and having to tangle with Views Module right away is dangerous UX.

olafgrabienski:
> To be honest I'm thinking more and more the front page layout should just have 3 columns

Do you want to bring this up in the issue thread (again)? I'm open for a more simple approach, and I had the impression others like the more advanced one, but I don't think they are absolutely against something more simple.

docwilmot:
Lets see what @stpaultim thinks.

stpaultim:
> To be honest I'm thinking more and more the front page layout should just have 3 columns.

This was my first idea and I agree that it would be the easiest for users to understand and use. I did not go this route #/1 Because, for this to really be easy for users, we should have a layout row with three regions (columns), one for each block. I was skeptical that I would get support for changing the home page layout. I'm nervous about unexpected complications from a new home page layout. I suppose we could do this using css to manage the blocks. But, this is less easy for end users to understand, but maybe still better than using views. Three separate regions, one for each block means that users can drag and drop to reposition (nice).

#/2 - The second reason for using views instead of three columns is that some people were requesting that we better showcase how to use views. I tried to add a second post on the front page to better showcase views (a view with one node does not showcase how to use views), but I  got push back saying that was too much content and asking me to remove the second post. So, I thought that using views for the Cards would combine both goals. 

We could drop the idea of showcasing how to use views (for now, save this for future) and move towards three columns. Then, I just need suggestions - does this mean a new front page layout that has a row with three regions (for columns) OR do we drop three existing content blocks into an existing region and use css to make them look like columns?

core/themes/basis/css/skin.css
```diff
@@ -1226,3 +1226,48 @@ div.error:before {
   /* background-image: -webkit-linear-gradient(transparent, transparent), url(../../../misc/message-error.svg); */
   /* background-image: linear-gradient(transparent, transparent), url(../../../misc/message-error.svg); */
 }
+
+/* Card view on default home page */
```
indigoxela: @stpaultim you may not be aware, but this section in skin.css is inside a `/* Color Module: Don't touch */` section, which means that color changes will not get processed here.

If that's by intention, it's OK, but if it just was an oversight, you might want to move that CSS above the "Don't touch" in line 1162.

[Core code that handles this behavior](https://github.com/backdrop/backdrop/blob/1.x/core/modules/color/color.module#L669)

core/themes/basis/css/skin.css
```diff
@@ -1226,3 +1226,48 @@ div.error:before {
...
+
+.block-views-card-grid-block {
+  margin-top: -2em;
+  padding-top: 2rem;
+  padding-bottom: 2rem;
+  background-color: #f2f2f2;
```
indigoxela: This color isn't used in Basis so far. Even if this section would get processed by the Color module, there won't be any replacement for `#f2f2f2`.

core/profiles/standard/config/layout.layout.home.json
```diff
@@ -114,6 +117,47 @@
...
+                "style": {
+                    "plugin": "dynamic",
+                    "data": {
+                        "settings": {
+                            "classes": "",
+                            "wrapper_tag": "div",
+                            "title_tag": "h2",
+                            "title_classes": "block-title",
+                            "content_tag": "div",
+                            "content_classes": "container card-design"
```
indigoxela: Out of curiosity: why the "container" class? Isn't that view supposed to be inside a container, anyway?

olafgrabienski: I think the reason is that the `l-top` region has no container class on a inner wrapper, at least in the Home layout.

olafgrabienski:
> We could drop the idea of showcasing how to use views (for now, save this for future) and move towards three columns. Then, I just need suggestions - does this mean a new front page layout that has a row with three regions (for columns) OR do we drop three existing content blocks into an existing region and use css to make them look like columns?

I guess, we'd need a new Home page layout. There is no wrapper for the Top region on the home page. At first sight, I don't see how to get the grey full-width background with three single blocks in a row using CSS only.

If we need addtions to the Home layout – be it a new region below or a row within Top – I'd suggest to implement the rest using our Bootstrap classes: `col-md-4` for each of the three blocks. This way, people can change the behavior, e.g. to `col-lg-4` and so on, even switch to displaying two blocks and use `col-md-6`.

olafgrabienski:
> be it a new region below or a row within Top

Not sure if a single row makes sense. Is it possible to render a row only if there are blocks? Is a single row visible in the user interface?

olafgrabienski:
> I switched to using NID, because they just weren't displaying in the right order in Tugboat when I sorted by time. (...) I assume that maybe they are all getting created at exactly the same time on Tugboat and the date/time sort isn't working?

Coming back to this, in case we stick with the Views grid. Can we sort by date/time as first criteria and NID as a second fallback criteria for Tugboat et al? So, if the creation time is different, everything should work as expected, and people can even change the dates, if they really want to. Only if the creation time is the same for all cards, NID order steps in.

olafgrabienski:
Hm, the last sandbox https://github.com/backdrop/backdrop/pull/3862#issuecomment-1053366684 isn't working for me - "Preview Not Found".

stpaultim:
I closed the PR and reopened it. The sandbox is working now. @olafgrabienski

core/profiles/standard/standard.install
```diff
@@ -339,6 +317,170 @@ function standard_install() {
...
+  file_prepare_directory($field_image_dir, FILE_CREATE_DIRECTORY);
+
+  $cards = array(
+    'card1-organize.png' => array(
+      'title' => st('Organize your content'),
+      'body' => st('<p>Backdrop CMS includes three pre-defined types of content: Pages, Posts, and Cards. Custom content types can also be added.</p><p><a href="https://docs.backdropcms.org/documentation/content-types">Learn more about content types</a></p>'),
+    ),
+    'card2-card.png' => array(
+      'title' => st('What is a Card?'),
+      'body' => st('<p>A Card is a flexible content item. Cards can be put on any page by placing them into a layout as "Existing content" blocks.</p><p><a href="https://docs.backdropcms.org/documentation/views">Learn more about Cards</a></p>'),
```
indigoxela:
```
<a href="https://docs.backdropcms.org/documentation/views">Learn more about Cards</a>
```
I guess, that's the wrong link.

stpaultim:
@indigoxela - Yes, good catch - we should have pointed that out in the code comments. There is no documentation page for "Cards" yet and I'm reluctant to create documentation for Cards, until we're sure that we're actually going to create them. So for now, we have no place to link to. 

Maybe the code should include a @TODO items or something.

jenlampton:
@stpaultim we can create an unpublished node for docs, and include the link to that node. It will be easier to make changes to the node than to core, so let's get the PR as close to finalized as possible :)

core/themes/basis/css/skin.css
```diff
@@ -400,6 +400,9 @@ fieldset .fieldset-legend {
   }
 }
 
+.l-top {
```
olafgrabienski: The scope of this definition (gray background for the top region) is too broad. It should be restricted to the home page.

core/modules/field/tests/field_views.test
```diff
@@ -34,7 +34,7 @@ class FieldViewsTestHelper extends ViewsSqlTest {
    */
   public $instances;
 
-  protected function CreateUser($extra_edit = array()) {
```
jenlampton: This change is unrelated to this PR and should probably be removed. (Changing class names is likely to have repercussions, so should probably wait until 2.x)

core/modules/field/tests/field_views.test
```diff
@@ -158,7 +156,7 @@ class FieldViewsDataTest extends FieldViewsTestHelper {
         'field_name_1' => array($langcode => array((array('value' => $this->randomName())))),
         'field_name_2' => array($langcode => array((array('value' => $this->randomName())))),
       );
-      $this->users[] = $this->CreateUser($edit);
```
jenlampton: This change is unrelated to this PR and should probably be removed.

core/profiles/standard/config/image.style.large_cropped.json
```diff
@@ -0,0 +1,18 @@
+{
+    "_config_name": "image.style.large_cropped",
```
jenlampton: Remove this image style.

core/profiles/standard/config/views.view.promoted_cards.json
```diff
@@ -0,0 +1,444 @@
...
+                        "element_wrapper_class": "",
+                        "element_default_classes": 1,
+                        "empty": "",
+                        "hide_empty": 0,
+                        "empty_zero": 0,
+                        "hide_alter_empty": 1,
+                        "click_sort_column": "fid",
+                        "type": "image",
+                        "settings": {
+                            "image_style": "large_cropped",
```
jenlampton: Switch to large image style

core/profiles/standard/standard.install
```diff
@@ -63,7 +63,7 @@ function standard_install() {
   $filtered_html_format = (object) $filtered_html_format;
   filter_format_save($filtered_html_format);
 
-  // Changed display name from Full HTML to Raw HTML but keeping machine name 
+  // Changed display name from Full HTML to Raw HTML but keeping machine name
```
jenlampton, line 66: Tense here should be present: `// Change display...`

core/profiles/standard/standard.install
```diff
@@ -157,6 +157,28 @@ function standard_install() {
       ),
       'is_new' => TRUE,
     ),
+    array(
+      'type' => 'card',
+      'name' => st('Card'),
+      'description' => st('Add a piece of content to be used for blocks and views. A card does not have a public path (hide path display). Example use-case: Place card content as "Existing content" block in a Layout.'),
```
jenlampton: This is too much text for a description. Recommendation:
`Add a card for use on an existing page. Display as a block, or in a view.`

For reference, here it is along side the other two descriptions:
* `Add a page with static content, like the 'About' page.`
* `Add a blog post, news article, or press release.`
* `Add a card for use on an existing page. Display as a block, or in a view.`

jenlampton: I added `Display as a block, or in a view.` to meet the same needs as the original text, but now that I see it, I think we could exclude that too if people prefer.

core/profiles/standard/standard.install
```diff
@@ -157,6 +157,28 @@ function standard_install() {
...
+        'status_default' => NODE_PUBLISHED,
+        'node_preview' => TRUE,
+        'promote_enabled' => TRUE,
+        'promote_default' => FALSE,
+        'sticky_enabled' => FALSE,
+        'sticky_default' => FALSE,
+        'revision_enabled' => TRUE,
+        'revision_default' => FALSE,
+        'scheduling_enabled' => FALSE,
+        'show_submitted_info' => TRUE,
```
jenlampton: We shouldn't show the submitted info for cards, let's set this to false.

core/profiles/standard/standard.install
```diff
@@ -157,6 +157,28 @@ function standard_install() {
...
+        'comment_default' => COMMENT_NODE_CLOSED,
+        'node_submitted' => FALSE,
+        'node_user_picture' => TRUE,
```
jenlampton: same for user picture, this might set to FALSE automatically once `show_submitted_info` is false.

core/profiles/standard/standard.install
```diff
@@ -169,53 +191,10 @@ function standard_install() {
   config('path.settings')
     ->set('node_page_pattern', '[node:title]')
     ->set('node_post_pattern', 'posts/[node:title]')
+    ->set('node_card_pattern', 'cards/[node:title]')
```
jenlampton, line 194: Should we remove the path alias pattern? These nodes will never be seen at their paths.

core/profiles/standard/standard.install
```diff
@@ -339,6 +318,181 @@ function standard_install() {
   );
   field_create_instance($instance);
 
+  $instance = array(
+    'field_name' => 'field_image',
+    'entity_type' => 'node',
+    'label' => 'Image',
+    'bundle' => 'card',
+    'description' => st('Upload an image to go with this card.'),
```
jenlampton, line 326: This description should be removed. It doesn't add any value (only restates node type and field name, already on the page) and adds clutter to the node form.

core/profiles/standard/standard.install
```diff
@@ -339,6 +318,181 @@ function standard_install() {
...
+        'preview_image_style' => 'thumbnail',
+      ),
+      'weight' => -1,
+    ),
+
+    'display' => array(
+      'default' => array(
+        'label' => 'hidden',
+        'type' => 'image',
+        'settings' => array('image_style' => 'large_cropped', 'image_link' => ''),
```
jenlampton: set to large image style (large_cropped should be removed)

core/profiles/standard/standard.install
```diff
@@ -339,6 +318,181 @@ function standard_install() {
...
+    'link_path' => 'node/' . $page->nid,
+    'link_title' => 'About',
+    'weight' => 1,
+  );
+  menu_link_save($item);
+
+  // Creating three default Card nodes.
+
+  // Credits for images.
+  // --- @TODO
```
jenlampton: Image credits still need to be added here.

core/profiles/standard/standard.install
```diff
@@ -339,6 +318,181 @@ function standard_install() {
...
+  // --- @TODO
+  // --- @TODO
+
+  $field_image_dir = 'public://field/image';
```
jenlampton, line 421: Can we change this to `public://cards/images`? The directory `/field/` could be used for any image field on the site, it doesn't seem specific enough to warrant having it's own directory. (So we could also leave the images in the root `public://` -- but I always prefer having a directory to keep things tidy :)

core/profiles/standard/standard.install
```diff
@@ -339,6 +318,181 @@ function standard_install() {
...
+  file_prepare_directory($field_image_dir, FILE_CREATE_DIRECTORY);
+
+  $cards = array(
+    'card1-organize.png' => array(
+      'title' => st('Organize your content'),
+      'body' => st('<p>Backdrop CMS includes three pre-defined types of content: Pages, Posts, and Cards. Custom content types can also be added.</p><p><a href="https://docs.backdropcms.org/documentation/content-types">Learn more about content types</a></p>'),
+    ),
+    'card2-card.png' => array(
+      'title' => st('What is a Card?'),
+      'body' => st('<p>A Card is a flexible content item. Cards can be put on any page by placing them into a layout as "Existing content" blocks.</p><p><a href="https://docs.backdropcms.org/documentation/blocks">Learn more about blocks</a></p>'),
```
jenlampton: This link still points to the blocks page. Can we link it to the creating content page instead? 
https://docs.backdropcms.org/documentation/creating-and-editing-content
I have created [a new revision](https://docs.backdropcms.org/node/46899/revisions/48777/view) that includes information about Cards, we can update it when this gets merged.

core/profiles/standard/standard.install
```diff
@@ -339,6 +318,181 @@ function standard_install() {
...
+    ),
+    'card3-layout.png' => array(
+      'title' => st('The Home page layout'),
+      'body' => st('<p>These Cards on the Home page were added to its layout through a block display from the "Card grid" view.</p><p><a href="https://docs.backdropcms.org/documentation/the-home-page">Learn more about the Home page</a></p>'),
```
jenlampton: This information is probably too technical and specific for the first-time persona. They won't know what any of the following are yet: block, display, view. Since the title is about the layout, should we maybe describe only that?

Maybe something like... `The home page layout is only active on this page. These cards only appear here`.

core/profiles/standard/standard.install
```diff
@@ -339,6 +318,181 @@ function standard_install() {
...
+              'format' => 'filtered_html',
+            ),
+          ),
+        ),
+        'field_image' => array(
+          LANGUAGE_NONE => array(
+            array(
+              'fid' => $file->fid,
+              'filename' => $image_filename,
+              'uri' => 'public://field/image/' . $image_filename,
```
jenlampton, line 466: `public://cards/image/`

core/profiles/standard/standard.install
```diff
@@ -339,6 +318,181 @@ function standard_install() {
...
+        'type' => 'card',
+      )
+    );
+
+    $card->save();
+  }
+
+  // Modify the default promoted view to exclude cards.
+  $promoted_view = views_get_view('promoted');
+  $promoted_view->display['default']->display_options['filters']['type'] = array(
```
jenlampton, line 484: nice!

core/themes/basis/css/component/cards.css
```diff
@@ -0,0 +1,50 @@
+/**
+ * @file
+ * Card listing view on the default home page.
+ */
+
+.l-top .block-views-promoted-cards-block {
+  margin-top: -2rem;/* Negative top margin collapses spacing under header */
```
jenlampton: This fixes a bug in basis (that we may fix shortly) should we add a @todo comment?

core/themes/basis/css/component/cards.css
```diff
@@ -0,0 +1,50 @@
...
+
+.block-views-promoted-cards-block .views-view-grid-cols-3 {
+  grid-template-columns: repeat(1, 1fr);
+}
+.view-promoted-cards .views-grid-box {
+  border: none;
+}
+
+.view-promoted-cards .views-grid-box {
+  border-radius: 0.25rem;
```
jenlampton, line 23: I see this radius on the bottom of the cards, but not the top.  Is that intentional?

core/themes/basis/css/skin.css
jenlampton: This is fixed in the latest PR>

core/profiles/standard/standard.install
jenlampton: Oh, just seeing now that this line only had a whitespace change. Maybe ignore this.

core/profiles/standard/standard.install
jenlampton: This description doesn't seem like it answers the question "what is a card?" Attempting a reword...

Thought process: 
* I don't think we need to say a Card is content, since we said that on the `Organize` card. 
* Is specifically mentioning the "Existing content" block helpful here? Or is that too much detail?
* If we do mention the  block placement, we shouldn't exclude a view as another option. Stealing text from the node type description.. 

`A Card is intended to be viewed only as part of another page. Place them individually as Blocks, or grouped as in a View.`

jenlampton: Ugh, `is intended to be viewed only as` still sounds cumbersome. trying again...

`A Card can only be seen as part of another page. Place them individually as Blocks, or grouped as in a View.` This leaves room for the type of block... 

`A Card can only be seen as part of another page. Place them individually as "Existing content" Blocks, or grouped as in a View.`

jenlampton:
This code is looking fantastic :) 

I've done a very detailed code review. Notable recommendations are as follows.
* We should remove the new image style, and switch to using Large instead. This results in absolutely no change because the images we have provided will not get cropped in the Y direction. There is no need to add the additional cruft to core and need to worry about what to name it. (`Large, cropped` doesn't differentiate it from the other `Large`, which is also cropped.)
* Location of card images shouldn't be `public://field/image/....` because `field` isn't specific enough, it could be used for every image field on the whole site, which almost negates the reason for adding a directory. I recommended `public://cards/images/....`

Things we'll need to fix before the merge:
* Add credits for the card images
* The link on cards card points to blocks docs (that's a tongue twister!)
* The house image still has the three extra dots at the top left.

Text change recommendations (these changes can come after merge):
* Description text recommendation for the new node type: `Add a card for use on an existing page. Display as a block, or in a view.`
* Home page layout card text recommendation: `The home page layout is only active on this page. These cards only appear here.`
* Card card text recommendation: `A Card can only be seen as part of another page. Place them individually as "Existing content" Blocks, or grouped as in a View.`

Questions: 
* I see the CSS border-radius on the bottom of the cards (white), but not the top (blue).  Is that intentional?
* The -2rem top margin fixes a bug in basis (that we may fix shortly) should we add a @todo comment?
* Should we remove the path alias pattern? These nodes will never be seen at their paths.  (I don't add aliases for any content on my sites that won't be seen at the page, to remind my editors that they won't be able to email a link to the card to someone else for review)

olafgrabienski:
@jenlampton, many thanks for the review! I'll come back to the the questions later. Before, I'd like to comment the image style topic. In my opinion, we should leave the new image style. Its goal is to prevent the display of custom images in different heights.

With the default card images, there is indeed no difference between the image styles `Large (800x600)` and the new style `Large, cropped (800x533)`. However, when people upload custom images, the aspect ratio will often not be the same for all three images. In this case, the `Large` style will result in images displayed at different heights. Which isn't appropriate for a card. See a screenshot from my demo site as example:

cards-example-image-style-1-large

The cropped image style fixes the problem. Screenshot example for comparison:

cards-example-image-style-2-large-cropped

> `Large, cropped` doesn't differentiate it from the other `Large`, which is also cropped.

`Large` just scales an image to 800x600, it doesn't crop it.

jenlampton:
I thought I confirmed that Large also uses scale and crop. I could be
mistaken.

My concern is that we're now shipping core with two nearly identical image
styles, with nearly identical names. That begs the question, why not just
use one? If Large doesn't do what we need, let's just change it. :)

stpaultim:
> I thought I confirmed that Large also uses scale and crop. I could be
mistaken.

> My concern is that we're now shipping core with two nearly identical image
styles, with nearly identical names. That begs the question, why not just
use one? If Large doesn't do what we need, let's just change it. :)

1) I just confirmed that the current Large image format does not use scale and crop, which is why we did this. 
2) I would not object to adding scale and crop to default image formats. BUT, I think it would be weird to add it to just one of the current image formats. We could add it to all three, but that seems like a major change to this PR with changes that really don't fall in the scope of this issue. I'm willing to do it, but I'm afraid that it might cause concerns for others which end up blocking this issue. 
3) I don't think it's a problem to add a new image format with similar behavior and similar title, if there is a reason to. However, I think I would agree with @jenlampton that if we don't need two different image formats, why not change the existing one (unless anyone has a reason that this is a bad idea). 

MY PROPOSAL:

The current PR will work without the new image format. The new image format is only there for users that add their own images and they can fix the problem my modifying the current image format or creating a new one. I propose that we:

1) Drop the new image format. 
2) Create a new issue to add cropping to all existing image formats. I'd be willing to create a PR for that right now, so we might be able to get both in the next release. HOWEVER, if change in image formats becomes a problem, at least it won't hold up the rest of the PR. 
3) If it turns out that there are reasons that we should not modify the existing image formats, we can add the new one later (after more discussion). 

Thoughts?

(I thought I was commenting in the issue, not on the PR. Will leave this here for now.)

See: https://github.com/backdrop/backdrop-issues/issues/5590

olafgrabienski:
@stpaultim Thanks for your proposal. I guess, add cropping to existing images isn't an option, see https://github.com/backdrop/backdrop-issues/issues/5590#issuecomment-1102154705 and below.

I'd prefer to *not* drop the new image format, OR proceed directly with (3), i.e. add three new ones. In any case, let's try to ship cards with a convenient image style.

> I thought I was commenting in the issue, not on the PR. Will leave this here for now.

Not sure about the best place. I was replying to Jen's review, therefore here on the PR (and also because the issue is already so long). I'm fine with switching place, if we need more visibility.

olafgrabienski:
Answering questions by @jenlampton :

> I see the CSS border-radius on the bottom of the cards (white), but not the top (blue). Is that intentional?

Good question! Actually, I don't remember. In my opinion, border-radius only on the bottom looks slightly better, so I'd prefer the current look, but would also be fine with radius on all corners. In any case, the CSS definition should be more specific: only define bottom corners, or add top radiuses for the image element.

> The -2rem top margin fixes a bug in basis (that we may fix shortly) should we add a @todo comment?

Are you referring to https://github.com/backdrop/backdrop-issues/issues/4095? It has a 2.x milestone, but a `@todo` sounds good, even if we don't fix the issue shortly.

> Should we remove the path alias pattern? These nodes will never be seen at their paths. (I don't add aliases for any content on my sites that won't be seen at the page, to remind my editors that they won't be able to email a link to the card to someone else for review)

Interesting! The reminder use-case makes sense ... even if it works probably best on a individual basis. Anyway, Backdrop comes with the Default URL alias pattern `[node:content-type]/[node:title]` which applies to all content types without specific patterns. So, to remove the pattern for cards, we had to remove the default pattern, and that's doesn't sound like a good idea to me.

stpaultim:
> @stpaultim Thanks for your proposal. I guess, add cropping to existing images isn't an option

It is still not clear to me that this is true. I'm waiting for an explanation in the other issue as to why this would be true.
https://github.com/backdrop/backdrop-issues/issues/5590#issuecomment-1103161348

ghost:
FTR, I'm not in favour of changing core's default image styles just for this use case. Further discussion in the other issue Tim created, but I agree with using new image styles here.

stpaultim:
> FTR, I'm not in favour of changing core's default image styles just for this use case.

Well, I don't think we would be making this change JUST for this use case. If we make the change, it would be because this use case revealed something that possibly should have been addressed along time ago. A number of good arguments can be made why the current configuration is problematic, with or without this use case. 

If you oppose changing the current image styles, are you in favor of adding a new one to address this use case? Right now, I think the discussion is about either adding a new image style or fixing the existing ones.

NEVERMIND: I see that you do prefer creating a new image style.

jenlampton:
> Interesting! The reminder use-case makes sense ... even if it works probably best on a individual basis. Anyway, Backdrop comes with the Default URL alias pattern [node:content-type]/[node:title] which applies to all content types without specific patterns. So, to remove the pattern for cards, we had to remove the default pattern, and that's doesn't sound like a good idea to me.

Ah, I must have lost that battle (not to have a default path pattern). Nevermind this suggestion!

> Are you referring to https://github.com/backdrop/backdrop-issues/issues/4095? It has a 2.x milestone, but a @todo sounds good, even if we don't fix the issue shortly.

Yep :) Mostly so we don't forget to fix that CSS when we get to that issue.

> I'd prefer to not drop the new image format, OR proceed directly with (3), i.e. add three new ones.

We certainly shouldn't be adding 3 new image styles in this issue, so let's take that off the table (and leave it for the follow-up) 

Let's look into switching Large to "Scale and crop" (for new installs only). It should have very little effect since the only place that image style is used in core is for post full displays. (And some tests)

stpaultim:
> Let's look into switching Large to "Scale and crop" (for new installs only). It should have very little effect since the only place that image style is used in core is for post full displays. (And some tests)

It seems that any changes to existing image styles might generate discussion and cause delays. My proposal was to move this discussion to other issues, either the one I created or if someone has a different proposal they can open a new ticket. 

I'd be willing to deal with this here if I thought we could come to a quick consensus, but I'm skeptical that modifying existing image styles is going to be a quick discussion - given that we've already gotten some opposition to it.

Does this make sense @jenlampton - or do you think we should try to include this change as part of this PR?

olafgrabienski:
> I'd prefer to not drop the new image format, OR proceed directly with (3), i.e. add three new ones. In any case, let's try to ship cards with a convenient image style.

Considering to take back my statement. I'd still be happy to have a good image style for the cards grid, to allow custom images without the need to create a custom style and change the view. But https://github.com/backdrop/backdrop-issues/issues/5590 became quite complicated very fast. So, if we need more time, I'm fine with using `Large` for the moment and add a convenient style in a follow-up.

stpaultim:
Here is my response to all the issues raised by @jenlampton - including comments I made in other places. 

> We should remove the new image style, and switch to using Large instead. This results in absolutely no change because the images we have provided will not get cropped in the Y direction. There is no need to add the additional cruft to core and need to worry about what to name it. (`Large, cropped` doesn't differentiate it from the other `Large`, which is also cropped.)

As mentioned above, the current Large image style is not cropped and the new one does make it much more difficult for someone to break this view simply by uploading an image that does not have the same dimensions. BUT, we have removed this new image style for now and opened a new issue to deal with that problem: https://github.com/backdrop/backdrop-issues/issues/5590

> Location of card images shouldn't be `public://field/image/....` because `field` isn't specific enough, it could be used for every image field on the whole site, which almost negates the reason for adding a directory. I recommended `public://cards/images/....`

So in fact, this is already the directory that is currently used for the default image field provided by core and used in posts. If we would like to create a more specific directory for images by content type, I am inclined to do that in a followup issue which addresses the problem for both Cards and Posts. I don't want to treat the images in Cards differently than we already treat the images in Posts without a discussion. 

> Add credits for the card images

I've added a mention to @olafgrabienski in the code. I'm not sure if we need something better, but I'll need specific guidance from @olafgrabienski on what to add. 

> The link on cards card points to blocks docs (that's a tongue twister!)

Well, that was intentional, because we didn't have a better idea. However, I like your idea and your new documentation, so I've changed this link to the Creating and Editing content page. 

> The house image still has the three extra dots at the top left.

I must have missed any discussion about removing those dots. I don't recall that coming up before. I'll need a new image from @olafgrabienski if we want to change that. 

> Description text recommendation for the new node type: `Add a card for use on an existing page. Display as a block, or in a view.`

DONE. I'm good with this change, hopefully at @olafgrabienski is too. If not, should we merge this and open a new issue to fix language after feature freeze?

> Home page layout card text recommendation: `The home page layout is only active on this page. These cards only appear here.`

I implemented a version of this with my own changes. Again, we can tweak text further after feature freeze, right. 

> Card card text recommendation: `A Card can only be seen as part of another page. Place them individually as "Existing content" Blocks, or grouped as in a View.`

Done. Again, we can tweak text further after feature freeze, right. 

Questions: 
> I see the CSS border-radius on the bottom of the cards (white), but not the top (blue).  Is that intentional?

@olafgrabienski likes it this way. The CSS is actually changing the border on top and bottom, but the top is being hidden by the image. Again, since @olafgrabienski likes this and he did most of the design work on this PR, I'm inclined to leave things as they are for now and open a new issue if we want to discuss this further. 

> The -2rem top margin fixes a bug in basis (that we may fix shortly) should we add a @todo comment?

I added a TODO comment, but wasn't really sure what to say. Please, provide specific text to improve it. 

> Should we remove the path alias pattern? These nodes will never be seen at their paths.

REMOVED. But, as @olafgrabienski pointed out, there is a default that generates a custom path anyway. 

> This change is unrelated to this PR and should probably be removed. (Changing class names is likely to have repercussions, so should probably wait until 2.x)

This is related to changes added to the test files, which were added by @quicksketch (See comments by @jenlampton above). I am not inclined to remove these changers without a better understanding of why they were added in first place. Can we get a comment from a Core Committer?

core/modules/field/tests/field_views.test
quicksketch: Considering this is a protected method and specific a single field + views test, this is unlikely to be used anywhere else. This method should probably be removed entirely and replaced to use the generic `BackdropWebTestCase::backdropCreateUser()`, but that would have been an even larger change.

quicksketch:
I re-reviewed and I think this seems good to go. I would have preferred we kept the new image style, but I'm fine with separating that part of the issue into https://github.com/backdrop/backdrop-issues/issues/5590.

I think at this point we've haggled to the point we have at least separated out the contentious parts. We have follow-ups, but I think we should get this in.

docwilmot:
> I think at this point we've haggled to the point we have at least separated out the contentious parts. We have follow-ups, but I think we should get this in.

Agreed!
> This change is unrelated to this PR and should probably be removed. (Changing class names is likely to have repercussions, so should probably wait until 2.x)

Agreed! (Unlikely, but someone somewhere could be extending the class or calling `CreaateUser()` in a test, no?)

core/modules/field/tests/field_views.test
ghost: I agree that unrelated changes shouldn't be included here. Better to discuss separately and minimise the changes in this already large PR.

quicksketch: kk, I reverted this method name change.

olafgrabienski:
@stpaultim Thanks for adjusting the PR! Here my answers to topics where you asked for my feedback:

**1. Credits for the card images**

>  I've added a mention to olafgrabienski in the code.

The icons are by @indigoxela, I've only adapted the images a bit, so credits should go to indigoxela.

**2. Extra dots in the house icon**

> I must have missed any discussion about removing those dots. I don't recall that coming up before. I'll need a new image from olafgrabienski if we want to change that.

Wes mentioned the dots ('squares') in https://github.com/backdrop/backdrop-issues/issues/4903#issuecomment-1083556667:
> I think the house one is a bit "visually busy". I think removing a few 'areas of interest' would improve it, e.g.: Get rid tight negative space spots, the squares, the line for the roof, and maaaybe the door?

A few comments below, I suggested to discuss the look of the icons in a follow-up issue to not derail this one. Still true, in my opinion.

**3. Description text recommendation for the new node type**

> DONE. I'm good with this change, hopefully at olafgrabienski is too

Yes, I like it!

**4. Remove the path alias pattern**

> REMOVED. But, as olafgrabienski pointed out, there is a default that generates a custom path anyway

Thanks for mentioning my point. If you agree, it's however important to *not* remove the specific pattern for cards. Why remove something when the goal of the removal isn't achieved anyway? It may only lead to confusion.

stpaultim:
> Thanks for mentioning my point. If you agree, it's however important to not remove the specific pattern for cards. Why remove something when the goal of the removal isn't achieved anyway? It may only lead to confusion.

I don't feel strongly either way, but I see your point now (I didn't fully get it before). We already have patterns for pages and posts, why would we specifically remove (or not create) the one for cards if it doesn't really change anything anyway. My default on other questions has been to go for consistency. I suggest we follow that practice here for now.

If we don't like this behavior, let's open a follow-up issue. This PR adds the pattern for cards, the others were there already. Removing this does not change anything, because the pattern is provided by default anyway as it would also be if we removed the posts pattern. 

If we want to get rid of unnecessary code, I suggest a follow-up issue.

```  config('path.settings')
    ->set('node_page_pattern', '[node:title]')
    ->set('node_post_pattern', 'cards/[node:title]')
    ->save();
```

olafgrabienski (commented):
Should probably be "node_**card**_pattern" (instead of "node_post_pattern")?

stpaultim:
> Should probably be "node_card_pattern" (instead of "node_post_pattern")?

@olafgrabienski 

:-(

Of course it should. We need to get this merged before I start making new problems. FIXED.

```
  config('path.settings')
    ->set('node_page_pattern', '[node:title]')
    ->set('node_card_pattern', 'cards/[node:title]')
    ->save();
```

quicksketch:
!'ve merged this into 1.x for 1.22.0 :tada:

jenlampton:
This is going to be such a great feature!

PR #3865, opened 2021-12-15
Issue 5399: Improvements to the Configure Layout page

### LINKED ISSUE #5399: [UX][DX] Improve the description for the Path field on the configure layout page.
In tinkering with Layouts lately (and writing up some description for a blog post), I thought of a couple changes to the Configure Layout page (`admin/structure/layouts/manage/%/configure`) that might be improvements. Putting them out there in this issue to get feedback and further thoughts.

Here's the relevant section of that page and the issues:

layouts_before

**1. Long description for Path.** There are 5 lines of description text for paths, which takes up a good-sized chunk of vertical space. Surely most users don't need to see the full list of examples after the first or second time. This seems like "extra information" that is a good candidate for the recently-introduced `details` form element, so it can stay collapsed most of the time, but is expandable when needed (and has summary text that hints at what's inside).

**2. Layout-created page or overridden page?** This seems like it's important enough that it should be visible to the user as soon as they've finished typing in their path. So I'd like to see some immediate indication of whether the given path gives rise to a layout-created or overridden page.

**3. (Optional) contexts.** The description text for Contexts begins with "(Optional)", but at least some of the time, there's nothing optional about it, for example, for the path `node/%`, as in the example above. And when it is optional, it should be obvious from the existence of a select drop-down that there's some options going on. So I think we can drop that parenthetical.

After making changes in response to those, here's what that section looks like:

layouts_after

Expanding the "Paths and placeholders" details element gives this:

paths_placeholders

And for layout-created pages, we see this:

layout_created

I've made a PR so people can try it out in the Tugboat sandbox. I solicit feedback on any aspect, but two in particular:

1. The layout override message is "This layout will override the page(s) created by function _node_page_view_,"which is a bit unsatisfactory because most non-coding site builders neither know nor care about the name of the page generating function. I wanted to give a little more information than just "This page will override some page(s)," though. Any ideas on what a better message would be here?

2. The layout-created/override message doesn't update quite as quickly as it should (for example, not as quickly as the Context sections update—you need to tab out to get the former to update). If I'm missing something in the ajax implementation, I'd appreciate any pointers toward a better way of doing the updating. [Update: this is now working the way it should.]

bugfolder:
Fixes https://github.com/backdrop/backdrop-issues/issues/5399.

### DIFF
--- a/core/modules/layout/layout.admin.inc
+++ b/core/modules/layout/layout.admin.inc
@@ -296,10 +296,6 @@ function layout_settings_form($form, &$form_state, Layout $layout) {
     '#default_value' => $layout->getPath(),
     '#required' => TRUE,
     '#access' => !$layout->isDefault(),
-    '#ajax' => array(
-      'callback' => 'layout_path_matches_callback',
-      'wrapper' => 'path-matches',
-    ),
   );
   $path = isset($form_state['values']['path']) ? $form_state['values']['path'] : $layout->getPath();
   $path_matches = db_query('
@@ -643,13 +639,6 @@ function layout_settings_form($form, &$form_state, Layout $layout) {
 }
 
 /**
- * Callback to return the path matching information.
- */
-function layout_path_matches_callback($form, &$form_state) {
-  return $form['path_matches'];
-}
-
-/**
  * Validates layout_settings_form(), ensuring a valid path.
  */
 function layout_settings_form_validate(&$form, &$form_state) {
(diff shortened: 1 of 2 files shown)


core/modules/layout/layout.admin.inc
```diff
@@ -314,7 +314,7 @@ function layout_settings_form($form, &$form_state, Layout $layout) {
       ', array(':path' => $path))
       ->fetchAllKeyed();
     if (count($path_matches)) {
-      $path_match_text = t('This layout will override the default layout for existing page(s) (which are created by %page_callback).', array('%page_callback' => $path_matches[$path]));
+      $path_match_text = t('This layout will override the default layout for existing page(s).');
```
klonos: Would it make sense to add something like "...that match this path pattern." here?

core/modules/layout/layout.admin.inc
```diff
@@ -281,23 +281,57 @@ function layout_settings_form($form, &$form_state, Layout $layout) {
...
+    t('User accounts: <code>user/%</code>'),
+    t('Taxonomy terms: <code>taxonomy/term/%</code>'),
+    t('Comments: <code>comment/%</code>'),
+  );
+  $description_list = theme('item_list', array('items' => $description_items));
+  $form['path_details'] = array(
+    '#type' => 'details',
+    '#summary' => t('Placeholder examples'),
+    '#details' => t('Some common examples of placeholders include:') . $description_list,
+    '#attributes' => array('class' => array('description')),
```
klonos, line 305: We need to exclude this entire form element if this is a default layout:

```suggestion
    '#attributes' => array('class' => array('description')),
    '#access' => !$layout->isDefault(),
```

Otherwise you end up with this "rogue" fieldset:

image

core/modules/layout/layout.admin.inc
```diff
@@ -281,23 +281,57 @@ function layout_settings_form($form, &$form_state, Layout $layout) {
...
+      $path_match_text = t('This path will create a page.');
+    }
+  }
+  else {
+    $path_match_text = '';
+  }
+
+  $form['path_matches_wrapper'] = array(
+    '#type' => 'container',
+    '#id' => 'path-matches',
```
klonos, line 330: Better be safe here too:

```suggestion
    '#id' => 'path-matches',
    '#access' => !$layout->isDefault(),
```

core/modules/layout/layout.admin.inc
bugfolder: Something like that. Technically, "pages" don't match a pattern, their paths do, so:

"This layout will override the default layout for existing page(s) whose paths match this pattern."

bugfolder:
Closing in favor of https://github.com/backdrop/backdrop/pull/3890.

PR #3868, opened 2021-12-17
Issue #5401: Update jQuery UI to 1.13.0

### LINKED ISSUE #5401: Update jQuery UI to 1.13.0
There's already a [meta issue](https://github.com/backdrop/backdrop-issues/issues/2493) for updating both, jQuery and jQuery UI, but this one deserves its own. There are some problems to solve.

"Recently" jQuery UI released a new version:

https://github.com/jquery/jquery-ui/releases/tag/1.13.0

The problem is: how to get it into Backdrop?

Backdrop has individual minified js and css files under core/misc/ui. All filenames are prefixed with jquery.ui.

The [official downloads](https://jqueryui.com/download/) only provide a single minified js file (concatenated). The sources have a different structure and the files are not minified.

The only solution that comes to my mind: get the sources (not minified), rename and reorder all files manually.

Is that the right approach?

I also [tried to get info](https://github.com/backdrop/backdrop-issues/issues/2493#issuecomment-991912583) about how this has been done before, but unfortunately got no answer. Possibly nobody remembers, how this has to be done?

# jquery UI functions in Backdrop
Here's a checklist of functions in core and contrib that use jquery UI
### Blocks 
- [x]  Added a custom block
- [x]  Moved another block

### Field UI
- [x]  Changed the order of fields on one of the Manage Display pages

### Layout UI 
- [x]  for path node/% does context appear automatically?
- [x]  drag and drop blocks
- [x]  Added a custom block
- [x]  Moved another block

### Better Exposed Filters 
- [ ]  datepicker
- [ ]  slider

### CKEditor
- [x]  link dialog
- [x]  image dialog
- [x]  Configuration: Added buttons to the editor toolbar
- [x]  Sortable
- [x]  Draggable

### Coffee 
- [x]  autocomplete

### Draggable Views
- [x]  draggable
- [x]  droppable

### Elements 
- [ ]  autocomplete

### Rules 
- [x]  autocomplete 
- [x]  button

### Views UI

- [x]  Added a view
- [x]  Changed to the display of fields
- [x]  Modal windows
- [x]  Rearranged the fields

### Weight 
- [x]  draggable
- [x]  droppable

---

advocate: Jen Lampton.

indigoxela:
Fixes https://github.com/backdrop/backdrop-issues/issues/5401

This will be a nightmare to review...


core/modules/system/system.module
```diff
@@ -1847,6 +1847,7 @@ function system_library_info() {
     'website' => 'http://jqueryui.com/effect/',
     'version' => $libraries['ui']['version'],
     'js' => array(
+      'core/misc/ui/vendor/jquery-color/jquery.color.js'  => array('group' => JS_LIBRARY, 'weight' => -9),
```
klonos: What is the reasoning for adding this under `$libraries['effects']`?

indigoxela: This code used to be part of the effects, but now ship as an extra file.
Effects still depend on it, so it makes sense to keep it combined.
See also [this comment](https://github.com/backdrop/backdrop-issues/issues/5401#issuecomment-997176747)

core/misc/ui/jquery.ui.core.min.js
```diff
@@ -1,4 +1 @@
-/*! jQuery UI - v1.12.1 - 2017-03-31
-* http://jqueryui.com
-* Copyright jQuery Foundation and other contributors; Licensed  */
-!function(){define(["jquery","./data","./disable-selection","./focusable","./form","./ie","./keycode","./labels","./plugin","./safe-active-element","./safe-blur","./scroll-parent","./tabbable","./unique-id","./version"])}();
```
quicksketch, line 4: Other compressed JS files (other than translations) seem to have kept their docblock at the top. Should this be restored/updated to match the other files?

core/misc/ui/jquery.ui.jquery-var-for-color.min.js
```diff
@@ -0,0 +1 @@
+!function(e){"use strict";"function"==typeof define&&define.amd?define(["jquery","./version"],e):e(jQuery)}(function(e){});
```
quicksketch, line 1: This file doesn't seem to be referenced in the hook_libraries() list. Should it be a dependency somewhere?

core/modules/system/system.module
quicksketch: I think this should be moved to its own library under the name `jquery.color` and then be separately placed directly in `/misc/` like other jQuery libraries. jQuery Color is an entirely separate project with a completely different version number and possibly release cycle (see https://github.com/jquery/jquery-color).

Then this `effects` library should depend on the new `jquery.color` library.

core/misc/ui/jquery.ui.jquery-var-for-color.min.js
indigoxela: We could, but that file wouldn't do anything. See [this comment](https://github.com/backdrop/backdrop-issues/issues/5401#issuecomment-997176747) for an explanation.

core/misc/ui/jquery.ui.core.min.js
indigoxela: > Other compressed JS files (other than translations) seem to have kept their docblock at the top...

This file has no docblock in the sources, hence no comment to keep here.

See the original: https://github.com/jquery/jquery-ui/blob/1.13.0/ui/core.js
That file is now deprecated and will get removed with the 1.14 release.

core/modules/system/system.module
klonos: This was fixed with [`07c4693` (#3868)](https://github.com/backdrop/backdrop/pull/3868/commits/07c46932ca4b021b191ea088c841447cb7bfd4f0)

klonos: ...actually no. @indigoxela you seem to have missed this suggestion:

> ...be separately placed directly in `/misc/` like other jQuery libraries.

core/misc/ui/jquery.ui.jquery-var-for-color.min.js
quicksketch: Thanks, I saw that comment but I didn't fully understand it's meaning. :+1:

core/modules/system/system.module
```diff
@@ -1842,14 +1842,24 @@ function system_library_info() {
       array('system', 'ui'),
     ),
   );
+  $libraries['jquery.color'] = array(
+    'title' => 'jQuery Color',
+    'website' => 'https://github.com/jquery/jquery-color',
+    'version' => '2.2.0',
+    'js' => array(
+      'core/misc/ui/vendor/jquery-color/jquery.color.js'  => array('group' => JS_LIBRARY, 'weight' => -9),
```
quicksketch: Could you place this directly in `/misc/jquery.color.js` to match other jQuery libraries? Then we'll exclude the entire `vendor` directory from the `/ui/` directory.

laryn:
Big thanks to @indigoxela, @quicksketch, @klonos for the PR, reviews, and fixes in this thread. Full credits in the merge commit and issue itself. Merged into `1.x` for `1.21.0`!

PR #3870, opened 2021-12-19
Issue #4718: Use localStorage to store tableDrag.showWeight value

### LINKED ISSUE #4718: Fixed: Add sameSite attribute to Backdrop tableDrag cookie.

klonos:
Fixes https://github.com/backdrop/backdrop-issues/issues/4718

Alternative to https://github.com/backdrop/backdrop/pull/3869
Mimics: https://git.drupalcode.org/project/drupal/-/commit/148871a8c914c2db9c28ceac3ebf9763a0e83e7e + follow-up https://git.drupalcode.org/project/drupal/-/commit/96bca83bcc2eb969d1330f84092652bc2615faf8

### DIFF
--- a/core/misc/tabledrag.js
+++ b/core/misc/tabledrag.js
@@ -134,7 +134,7 @@ Backdrop.tableDrag = function (table, tableSettings) {
   $(document).on('mousemove pointermove', function (event) { return self.dragRow(event, self); });
   $(document).on('mouseup pointerup', function (event) { return self.dropRow(event, self); });
   // React to localStorage event showing or hiding weight columns.
-  $(window).bind('storage', $.proxy(function (e) {
+  $(window).on('storage', $.proxy(function (e) {
     // Only react to 'Backdrop.tableDrag.showWeight' value change.
     if (e.originalEvent.key === 'Backdrop.tableDrag.showWeight') {
       // This was changed in another window, get the new value for this window.
--- a/core/misc/tabledrag.js
+++ b/core/misc/tabledrag.js
@@ -137,7 +137,7 @@ Backdrop.tableDrag = function (table, tableSettings) {
   $(window).on('storage', function (e) {
     // Only react to 'Backdrop.tableDrag.showWeight' value change.
     if (e.originalEvent.key === 'Backdrop.tableDrag.showWeight') {
-      // This was changed in another window, get the new value for this window.
+      // This was changed in another window. Get the new value for this window.
       showWeight = JSON.parse(e.originalEvent.newValue);
       self.displayColumns(showWeight);
     }
--- a/core/misc/tabledrag.js
+++ b/core/misc/tabledrag.js
@@ -1,5 +1,20 @@
 (function ($) {
 
+"use strict";
+
+/**
+ * Store the state of weight columns display for all tables.
+ *
+ * Default value is to hide weight columns.
+ *
+ * We are using JSON.parse here, because true/false get stored as strings.
+ * Otherwise showWeight would be either 'true' or 'false', and they would both
+ * be true in the toggle condition.
+ */
+var showWeight = JSON.parse(
+  localStorage.getItem('Backdrop.tableDrag.showWeight')
+);
+
 /**
  * Drag and drop table rows with field manipulation.
  *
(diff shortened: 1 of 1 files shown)


core/misc/tabledrag.js
```diff
@@ -98,15 +113,10 @@ Backdrop.tableDrag = function (table, tableSettings) {
...
-    .click(function () {
-      if ($.cookie('Backdrop.tableDrag.showWeight') == 1) {
-        self.hideColumns();
-      }
-      else {
-        self.showColumns();
-      }
-      return false;
-    })
+    .click($.proxy(function (e) {
```
indigoxela: I see one possible problem with this: [jQuery.proxy() is deprecated](https://api.jquery.com/jQuery.proxy/)

core/misc/tabledrag.js
```diff
@@ -123,6 +133,15 @@ Backdrop.tableDrag = function (table, tableSettings) {
   $(document).on('touchend', function (event) { return self.dropRow(event.originalEvent.touches[0], self); });
   $(document).on('mousemove pointermove', function (event) { return self.dragRow(event, self); });
   $(document).on('mouseup pointerup', function (event) { return self.dropRow(event, self); });
+  // React to localStorage event showing or hiding weight columns.
+  $(window).bind('storage', $.proxy(function (e) {
```
indigoxela: Function jQuery bind is also [deprecated](https://api.jquery.com/bind/).

core/misc/tabledrag.js
klonos: @indigoxela it seems that `.proxy()` was deprecated in jQuery v3.3, whereas we are on 1.x in Backdrop. Besides, it is used in many other .js files in core. Should we leave that to be part of https://github.com/backdrop/backdrop-issues/issues/3106 instead?

core/misc/tabledrag.js
klonos: Fixed with [`c5723f7` (#3870)](https://github.com/backdrop/backdrop/pull/3870/commits/c5723f7f78571004fac900e7c5eb4ed63ee5a248)

core/misc/tabledrag.js
klonos: Fixed with [`fdd72a4` (#3870)](https://github.com/backdrop/backdrop/pull/3870/commits/fdd72a40bc1c4c4bbe1c080b3c35a73d85b09c62)

core/misc/tabledrag.js
```diff
@@ -1,5 +1,20 @@
...
+/**
+ * Store the state of weight columns display for all tables.
+ *
+ * Default value is to hide weight columns.
+ *
+ * We are using JSON.parse here, because true/false get stored as strings.
+ * Otherwise showWeight would be either 'true' or 'false', and they would both
+ * be true in the toggle condition.
+ */
+var showWeight = JSON.parse(
```
indigoxela, line 14: Any chance to avoid this global variable?

klonos: Not sure @indigoxela 🤷🏼 ...my goal was to simply backport https://git.drupalcode.org/project/drupal/-/commit/148871a8c914c2db9c28ceac3ebf9763a0e83e7e

If there's a way to achieve that, then please feel free to suggest via PR against mine, or a completely new PR. Would be great to also repay our Drupal brethren by contributing the same back to Drupal.

klonos:
Closing this in favor of https://github.com/backdrop/backdrop/pull/3869

PR #3871, opened 2021-12-19
Issue #3687: Add an upgrade path from Admin Menu to Admin bar

### LINKED ISSUE #3687: Add an upgrade path from Admin Menu to Admin bar
from @klonos in https://github.com/backdrop/backdrop-issues/issues/3680#issuecomment-484934403

> Re admin_menu, doesn't it provide (some of) the same options as the core admin_bar module ("sticky at top of pages" etc.)? We should be migrating these settings over ...if not already doing so.

Reply from me: 

> I think that's the only setting that's the same between contrib and core, though it wouldn't hurt to check. I'll create an issue for the checking!

TODO: 
* check all the variables set in admin_menu
* assure any that are the same are upgraded in admin_bar
* document those that are not provided in admin_bar
* update the documentation at https://backdropcms.org/project/admin_menu

docwilmot:
Fixes https://github.com/backdrop/backdrop-issues/issues/3687

### DIFF
--- a/core/modules/admin_bar/admin_bar.install
+++ b/core/modules/admin_bar/admin_bar.install
@@ -22,3 +22,33 @@ function admin_bar_install() {
   // other modules in hook_menu_alter().
   module_set_weight('admin_bar', 100);
 }
+
+/**
+ * Migrate Admin Bar variables to config.
+ */
+function admin_bar_update_1000() {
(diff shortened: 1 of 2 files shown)


docwilmot:
Forgot, D7 upgrades must be in System module.

PR #3873, opened 2021-12-19
Issue 5405: Change "path alias" to "URL alias"

### LINKED ISSUE #5405: Change "path alias" to "URL alias" throughout core code
Per https://github.com/backdrop/backdrop-issues/issues/4868#issuecomment-996279953, the term "path alias" has been replaced by "URL alias" in Backdrop. But the term "path alias" is used in many places in the code and, importantly, in the visible documentation:

There is a "path alias" here: https://docs.backdropcms.org/api/backdrop/core%21includes%21path.inc/function/backdrop_get_normal_path/1

This page uses both "URL alias" and "path alias":
https://docs.backdropcms.org/api/backdrop/core%21includes%21path.inc/1

A search through core code turns up 61 uses of "path alias" but 187 uses of "URL alias".

It will help consistency and documentation to use the same term everywhere.

Although there is https://github.com/backdrop/backdrop-issues/issues/2821, which suggests to rename the `url_alias` table to `path_alias`, which would seem to be counter to "URL alias" being the official term.

bugfolder:
Fixes https://github.com/backdrop/backdrop-issues/issues/5405.


core/includes/bootstrap.inc
```diff
@@ -3702,7 +3702,7 @@ function language_default() {
  * - http://example.com/node/306 returns "node/306".
  * - http://example.com/backdropfolder/node/306 returns "node/306" while
  *   base_path() returns "/backdropfolder/".
- * - http://example.com/path/alias (which is a path alias for node/306) returns
+ * - http://example.com/path/alias (which is a URL alias for node/306) returns
```
klonos: Wondering if we should also change the example to `http://example.com/url/alias`, and the "path/alias" in the following line to "url/alias" 🤔

core/includes/path.inc
```diff
@@ -322,8 +322,8 @@ function backdrop_match_path($path, $patterns) {
  * - http://example.com/node/306 returns "node/306".
  * - http://example.com/backdropfolder/node/306 returns "node/306" while
  *   base_path() returns "/backdropfolder/".
- * - http://example.com/path/alias (which is a path alias for node/306) returns
- *   "node/306" as opposed to the path alias.
+ * - http://example.com/path/alias (which is a URL alias for node/306) returns
```
klonos: Same as my previous comment.

core/includes/bootstrap.inc
bugfolder: That makes sense, but that also triggers the thought: `http://example.com/url-alias` seems like a more "alias-like" URL, since they are very often hyphenated versions of the node title. Would that be a better example? Or perhaps, `http://example.com/post/url-alias`?

core/includes/path.inc
bugfolder: Ditto.

core/includes/bootstrap.inc
quicksketch: I like your second suggestion @bugfolder but it doesn't fit on one line. So let's use the first one:

```suggestion
 * - http://example.com/url-alias (which is a URL alias for node/306) returns
```

core/includes/path.inc
quicksketch:
```suggestion
 * - http://example.com/url-alias (which is a URL alias for node/306) returns
```

quicksketch:
There are minor merge conflicts with 1.20.x. Since this involves string changes anyway, I've only merged it into 1.x for 1.21.0. Thanks @bugfolder!

PR #3876, opened 2021-12-24
Issue 5415: [DX] Simpletest doesn't handle empty or malformed .tests.info file well

### LINKED ISSUE #5415: [DX] Fixed: Simpletest doesn't handle empty or malformed `.tests.info` files well.
## Description of the bug

If Simpletest encounters an empty .tests.info file, one containing only the packaging information but nothing else, then when you go to the Tests list, admin/config/development/testing, it puts a slew of "Warning: Illegal string offset..." messages into watchdog. These clog watchdog and it's not clear which of the possibly many .tests.info files might be causing this.

## Steps To Reproduce

To reproduce the behavior:

1. Add a module with one or more empty .tests.info file to your installation. (The current version of Search API module is a good test case.)
2. Clear the watchdog log.
3. Go to admin/config/development/testing.
4. Clear the cache.
5. Go back to watchdog and see all the fireworks.

## Actual behavior

Lots of errors in watchdog, all showing up in simpletest.pages.inc, when the actual culprit is somewhere else (but no sign of where).

## Expected behavior

Some indication of where the problem lies.

## Additional information

The creation of a packaged but empty tests.info file is not as unusual as one might think, because Coder Upgrade sometimes generates these and they're easy to overlook since they're typically buried in subdirectories.

---

## Translation update:

New Strings:
* `File %uri contains no testing classes.`
* `File %uri contains a definition for %class that is not an array of testing class fields.`
* `File %uri contains a definition for %class that is missing field %keys.`
* `File %uri contains a definition for %class that is missing fields %keys.`
* `See the Testing Framework documentation.`

bugfolder:
Fixes https://github.com/backdrop/backdrop-issues/issues/5415.

### DIFF
--- a/core/modules/simpletest/simpletest.module
+++ b/core/modules/simpletest/simpletest.module
@@ -371,7 +371,7 @@ function simpletest_test_get_all() {
         // is loaded. So we'll check and put a more informative error here.
         $class_values = array_values($classes);
         if (empty($classes) || !is_array($class_values[0])) {
-          watchdog('simpletest', 'File %uri contains no testing classes.', array('%uri' => $file->uri), WATCHDOG_WARNING);
+          watchdog('simpletest', 'File %uri contains no testing classes. <a href="https://docs.backdropcms.org/documentation/testing-framework">See the Testing Framework documentation for details.</a>', array('%uri' => $file->uri), WATCHDOG_WARNING);
           continue;
         }
         foreach ($classes as $class => $info) {
--- a/core/modules/simpletest/simpletest.module
+++ b/core/modules/simpletest/simpletest.module
@@ -366,12 +366,12 @@ function simpletest_test_get_all() {
       $groups = array();
       foreach ($files as $file) {
         $classes = backdrop_parse_info_file($file->uri, TRUE);
+        // If the .tests.info file contained no classes, then it puts a slew of
+        // hard-to-diagnose errors into the watchdog log when the testing page
+        // is loaded. So we'll check and put a more informative error here.
         $class_values = array_values($classes);
         if (empty($classes) || !is_array($class_values[0])) {
-          // If the .tests.info file contained no classes, then it puts a slew
-          // of hard-to-diagnose errors into the watchdog log when the testing
-          // page is loaded. So we'll put a more informative error here.
-          watchdog('simpletest', 'There are no testing classes in file %uri', array('%uri' => $file->uri), WATCHDOG_WARNING);
+          watchdog('simpletest', 'File %uri contains no testing classes.', array('%uri' => $file->uri), WATCHDOG_WARNING);
           continue;
         }
         foreach ($classes as $class => $info) {
--- a/core/modules/simpletest/simpletest.module
+++ b/core/modules/simpletest/simpletest.module
@@ -371,7 +371,7 @@ function simpletest_test_get_all() {
         // is loaded. So we'll check and put a more informative error here.
         $class_values = array_values($classes);
         if (empty($classes) || !is_array($class_values[0])) {
-          watchdog('simpletest', 'File %uri contains no testing classes. <a href="https://docs.backdropcms.org/documentation/testing-framework">See the Testing Framework documentation for details.</a>', array('%uri' => $file->uri), WATCHDOG_WARNING);
+          watchdog('simpletest', 'File %uri contains no testing classes.', array('%uri' => $file->uri), WATCHDOG_WARNING, '<a href="https://docs.backdropcms.org/documentation/testing-framework">See the Testing Framework documentation.</a>');
           continue;
         }
         foreach ($classes as $class => $info) {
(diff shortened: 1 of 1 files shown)


klonos (approved):
LGTM 👍🏼

core/modules/simpletest/simpletest.module
```diff
@@ -366,6 +366,14 @@ function simpletest_test_get_all() {
...
       foreach ($files as $file) {
         $classes = backdrop_parse_info_file($file->uri, TRUE);
+        // If the .tests.info file contained no classes, then it puts a slew of
+        // hard-to-diagnose errors into the watchdog log when the testing page
+        // is loaded. So we'll check and put a more informative error here.
+        $class_values = array_values($classes);
+        if (empty($classes) || !is_array($class_values[0])) {
+          watchdog('simpletest', 'File %uri contains no testing classes.', array('%uri' => $file->uri), WATCHDOG_WARNING, l(t('See the Testing Framework documentation.'), 'https://docs.backdropcms.org/documentation/testing-framework'));
+          continue;
+        }
```
docwilmot, line 373: Nitpicking, but its probably more likely that youll bail out because of an empty info file so we should run that first then check if theres an empty class, so maybe
```
        $classes = backdrop_parse_info_file($file->uri, TRUE);
        // Ensure test info file is not empty and that each entry is properly
        // formatted.
        if (empty($classes)) {
          watchdog('simpletest', 'File %uri contains no testing classes.', array('%uri' => $file->uri), WATCHDOG_WARNING, l(t('See the Testing Framework documentation.'), 'https://docs.backdropcms.org/documentation/testing-framework'));
          continue;
        }
        foreach ($classes as $class_name => $class) {
          $class_values = array_values($classes);
          if (!is_array($class)) {
            watchdog('simpletest', 'File %uri contains  a malformed test info entry for %class_name.', array('%uri' => $file->uri, '%class_name' => $class_name), WATCHDOG_WARNING, l(t('See the Testing Framework documentation.'), 'https://docs.backdropcms.org/documentation/testing-framework'));
            unset($classes[$class_name]); // or maybe continue 2, depending on whats happening later in the code, I havent checked
          }
        }
```

bugfolder: Actually, in the situations I'd encountered, it was the case that the file contains no testing classes, but that would go through the second branch (wrongly) in this example code. (This is the case with Search API module, for example.) 

What happens (happened) is that Coder Upgrade creates a completely empty file, but when the person creates a release, the packager adds the timestamp, etc., so the file is no longer empty, and that leads to the second situation. So it's not that the file contains a malformed class; it's that it truly contains no classes at all, but `backdrop_parse_info()` is what creates the appearance of a malformed class.

The current (simpler) code covers both situations, and gives the right message: "go look at your .info file", so I suggest we leave this as it currently is.

docwilmot: The situation *should* be that contrib should be responsible to ensure that `test.info` files are properly formatted. But in this case we find that if contrib messes up, in any way, core doesnt properly warn us about which contrib is messing up, so we should fix core this time. So I feel the solution should seek to discover the possible ways contrib could mess up and mitigate that. The solution youve got is specifically for empty files. But we get errors in these situations:
- files with statements (xxx = yyy) only
- file with statements before test cases
- incomplete sections (missing either or all of name, description, group, or file)

In any of those situations the user would have no idea where this error is coming from, which is what I think we should try and solve, not just one situation.

FWIW my "solution" above was untested, and wouldnt actually sove any of what I just described.

bugfolder: Those are all good cases to trap. I'll put together a suite of files malformed in those (and any other ways I can think of) and revise the PR to handle them. Copying the relevant bits of your comment into the issue for visibility.

core/modules/simpletest/simpletest.module
```diff
@@ -364,28 +364,52 @@ function simpletest_test_get_all() {
...
+            // is loaded. So we'll check for various malformalities and put a more
+            // informative error.
+            if (!is_array($info) || empty($info)) {
+              $error = format_string('File %uri contains a definition for %class that is not an array of testing class fields.', array('%class' => $class));
+              continue;
+            }
+            else {
+              foreach ($check_keys as $key) {
+                if (!isset($info[$key])) {
+                  $error = format_string('File %uri contains a definition for %class that is missing field %key.', array('%class' => $class, '%key' => $key));
```
klonos: The logic seems to be bailing out soon as a single error is found for any of the `$check_keys`. I'd expect a `$keys` (plural) array, so that if a .tests.info file has multiple malformities, then we warn about all of them. So I'd also expect a `format_plural()` here, checking on the count of `$keys`, and if multiple, then say "...is missing the [implode()'d $keys] fields." for the plural variant.

Either that ^^, or alternatively multiple errors, one for each missing key.

bugfolder: Definitely the former. Putting multiple watchdog messages per file for each detected malformation would be pretty annoying. In a similar vein, if there are multiple malformed classes in the file, we don't put a message for each malformation detected (especially because once `backdrop_parse_info()` gets confused, all bets are off on what might show up in the parsed array). But reporting if multiple fields are missing is a nice courtesy, so added.

quicksketch:

PR #3878, opened 2021-12-25
Issue #4120: Strip core version prefix in _backdrop_version_compare_convert()

### LINKED ISSUE #4120: Nested dependencies causing UX issues in Project Installer
**Description of the bug**
When a module has a dependency and that dependency has a further dependency, it causes some UX confusion in the Project Installer.

**Steps To Reproduce**
To reproduce the behavior:
1. Try to install Ubercart.

**Actual behavior**
First it asks in an additional step to download Rules (which Ubercart depends on). Then when we do that, it asks in another separate step to download Entity Plus and Entity UI (which Rules depends on). Then it doesn't allow Ubercart to be enabled:
Screen Shot 2019-10-08 at 1 48 17 PM

**Expected behavior**
Ideally it would download all dependencies in a single step (if that is possible). Then it would be nice to be able to enable the main module (and any dependencies would automatically be enabled/checked when that module was checked).

Related for the last step: Would be nice to see submodules for enabling as well: https://github.com/backdrop/backdrop-issues/issues/2200

klonos:
Fixes https://github.com/backdrop/backdrop-issues/issues/4120

### DIFF
--- a/core/includes/common.inc
+++ b/core/includes/common.inc
@@ -8743,12 +8743,18 @@ function backdrop_check_incompatibility(array $dependency_info, $current_version
  * Converts a Backdrop version string into numeric-only version string.
  *
  * @param string $version_string
- *   A version string such as 1.10.0-beta4 or 1.4.x-dev.
+ *   A version string such as 1.x-1.2.3, or 1.10.0-beta4, or 1.4.x-dev.
  * @return string
  *   A converted string only containing numbers, for use in PHP's
  *   version_compare() function.
  */
 function _backdrop_version_compare_convert($version_string) {
+  // Remove the "1.x-" prefix (indicating Backdrop core version compatibility).
+  $core_prefix = BACKDROP_CORE_COMPATIBILITY . '-';
+  if (substr($version_string, 0, strlen($core_prefix)) == $core_prefix) {
+    $version_string = substr($version_string, strlen($core_prefix));
+  }
+
   // Convert "dev" releases to be the highest possible version number. For
   // example 1.5.x-dev should be considered higher than any other 1.5 release,
   // so we replace .x with 99999.
--- a/core/includes/common.inc
+++ b/core/includes/common.inc
@@ -8743,7 +8743,7 @@ function backdrop_check_incompatibility(array $dependency_info, $current_version
  * Converts a Backdrop version string into numeric-only version string.
  *
  * @param string $version_string
- *   A version string such as 1.x-1.2.3, or 1.10.0-beta4, or 1.4.x-dev.
+ *   A version string such as 1.x-1.2.3, 1.10.0-beta4, or 1.4.x-dev.
  * @return string
  *   A converted string only containing numbers, for use in PHP's
  *   version_compare() function.
--- a/core/includes/common.inc
+++ b/core/includes/common.inc
@@ -8751,7 +8751,7 @@ function backdrop_check_incompatibility(array $dependency_info, $current_version
 function _backdrop_version_compare_convert($version_string) {
   // Remove the "1.x-" prefix (indicating Backdrop core version compatibility).
   $core_prefix = BACKDROP_CORE_COMPATIBILITY . '-';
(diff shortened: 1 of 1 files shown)


argiepiano:
@klonos  rather than stripping this within `_backdrop_version_compare_convert()` I suggest we do it before invoking `backdrop_check_incompatibility()` inside `installer_browser_installation_enable_form()`

Please notice that this stripping happens in a similar situation in in system.admin.inc line [630](https://github.com/backdrop/backdrop/blob/ecf79edf4eedea2453448099a6c79bb9920bea3e/core/modules/system/system.admin.inc#L630), before invoking `backdrop_check_incompatibility()`. 

Doing the stripping before `backdrop_check_incompatibility()` will keep the same pattern in both places.

klonos:
@argiepiano I'm not 100% sure...

If we do it before invoking `backdrop_check_incompatibility()`, then we are making it the responsibility of developers to make sure that the versions passed as parameters to that function are "sanitized" (properly formatted).

`backdrop_check_incompatibility()` invokes `_backdrop_version_compare_convert()` twice:
- for `$current_version`
- for `$required_version['version']`

By making sure that `_backdrop_version_compare_convert()` properly formats versions, we lift the responsibility from developers. Moreover:
- As I have explained in the issue, the purpose of `_backdrop_version_compare_convert()` is exactly that: to convert version strings to numeric-only version strings. This is not happening with versions like `1.x-1.2.3`, and this is what this PR here is fixing.
- If we fix this in `installer_browser_installation_enable_form()`, then we're fixing things in a single place, for a single use case. Fixing it in `_backdrop_version_compare_convert()` feels like we're fixing the issue in its root (`backdrop_check_incompatibility()` will return proper/expected results regardless where it's being called from, and regardless of the format the versions being passed to it are in).

Happy to be convinced otherwise.

argiepiano:
That makes sense, and there is no harm with attempting to strip here. I tested your PR and it works for me

core/includes/common.inc
```diff
@@ -8743,12 +8743,18 @@ function backdrop_check_incompatibility(array $dependency_info, $current_version
  * Converts a Backdrop version string into numeric-only version string.
  *
  * @param string $version_string
- *   A version string such as 1.10.0-beta4 or 1.4.x-dev.
+ *   A version string such as 1.x-1.2.3, or 1.10.0-beta4, or 1.4.x-dev.
```
ghost:
```suggestion
 *   A version string such as 1.x-1.2.3, 1.10.0-beta4, or 1.4.x-dev.
```

core/includes/common.inc
```diff
@@ -8743,12 +8743,18 @@ function backdrop_check_incompatibility(array $dependency_info, $current_version
...
  * @return string
  *   A converted string only containing numbers, for use in PHP's
  *   version_compare() function.
  */
 function _backdrop_version_compare_convert($version_string) {
+  // Remove the "1.x-" prefix (indicating Backdrop core version compatibility).
+  $core_prefix = BACKDROP_CORE_COMPATIBILITY . '-';
+  if (substr($version_string, 0, strlen($core_prefix)) == $core_prefix) {
```
ghost: This seems simpler:

```suggestion
  if (strpos($version_string, $core_prefix) === 0) {
```

ghost (changes requested):

core/includes/common.inc
klonos: Committed with [`40e9095` (#3878)](https://github.com/backdrop/backdrop/pull/3878/commits/40e9095bf78be2f8274f9ea44d729ada63368a6f)

core/includes/common.inc
klonos: Indeed 😅 

Committed with [`a896e1e` (#3878)](https://github.com/backdrop/backdrop/pull/3878/commits/a896e1ea5e97e37cb5b29a44ba485d69e6e8bbf2)

ghost (approved):

quicksketch:
Merged into 1.x and 1.20.x. Thanks folks!

PR #3879, opened 2021-12-30
Issue #5398: Support php 8.1

### LINKED ISSUE #5398: Added PHP 8.1 support.
All my scripts default to using 7.4, but I just updated lando on my local to 3.6.0 and gave it a spin with php8.1/apache2:
- bee gave me multiple `Fatal error: During inheritance of ArrayAccess: Uncaught ArgumentCountError: Too few arguments to function bee_error_handler()` so installing/configuring automatically failed
- ```
  Deprecated: Return type of DatabaseStatementBase::execute($args = [], $options = [])
  should either be compatible with PDOStatement::execute(?array $params = null): bool,
  or the #[\ReturnTypeWillChange] attribute should be used to temporarily suppress the notice
  in /core/includes/database/database.inc on line 2266
  ```
- multiple "Deprecated function" errors during manual installation:

  - ```
    Deprecated function: Return type of DatabaseStatementEmpty::current()
    should either be compatible with Iterator::current(): mixed,
    or the #[\ReturnTypeWillChange] attribute should be used
    to temporarily suppress the notice in require_once()
    (line 2352 of /core/includes/database/database.inc).
    ```
  - ```
    Deprecated function: Return type of DatabaseStatementEmpty::next()
    should either be compatible with Iterator::next(): void,
    or the #[\ReturnTypeWillChange] attribute should be used
    to temporarily suppress the notice in require_once()
    (line 2352 of /core/includes/database/database.inc).
    ```

  - ```
    Deprecated function: Return type of DatabaseStatementEmpty::key()
    should either be compatible with Iterator::key(): mixed,
    or the #[\ReturnTypeWillChange] attribute should be used
    to temporarily suppress the notice
    in require_once() (line 2352 of /core/includes/database/database.inc).
    ```

  - ```
    Deprecated function: Return type of DatabaseStatementEmpty::valid()
    should either be compatible with Iterator::valid(): bool,
    or the #[\ReturnTypeWillChange] attribute should be used
    to temporarily suppress the notice
    in require_once() (line 2352 of /core/includes/database/database.inc).
    ```

  - ```
    Deprecated function: Return type of DatabaseStatementEmpty::rewind()
    should either be compatible with Iterator::rewind(): void,
    or the #[\ReturnTypeWillChange] attribute should be used
    to temporarily suppress the notice
    in require_once() (line 2352 of /core/includes/database/database.inc).
    ```

  - ```
    Deprecated function: strlen():
    Passing null to parameter #1 ($string) of type string is deprecated
    in backdrop_random_bytes() (line 2556 of /core/includes/bootstrap.inc).
    ```

  - ```
    Deprecated function: strlen():
    Passing null to parameter #1 ($string) of type string is deprecated
    in backdrop_validate_utf8() (line 2089 of /core/includes/bootstrap.inc).
    ```

  - ```
    Deprecated function: str_replace():
    Passing null to parameter #3 ($subject) of type array|string is deprecated
    in filter_xss() (line 1831 of /core/includes/common.inc).
    ```

  - ```
    Deprecated function: strlen():
    Passing null to parameter #1 ($string) of type string is deprecated
    in backdrop_validate_utf8() (line 2089 of /core/includes/bootstrap.inc).
    ```

  - ```
    Deprecated function: str_replace():
    Passing null to parameter #3 ($subject) of type array|string is deprecated
    in filter_xss() (line 1831 of /core/includes/common.inc).
    ```

  - ```
    Deprecated function: strpos():
    Passing null to parameter #1 ($haystack) of type string is deprecated
    in url_is_external() (line 2755 of /core/includes/common.inc).
    ```

  - ```
    Deprecated function: str_replace():
    Passing null to parameter #3 ($subject) of type array|string is deprecated
    in url_is_external() (line 2757 of /core/includes/common.inc).
    ```

  - ```
    Deprecated function: ltrim():
    Passing null to parameter #1 ($string) of type string is deprecated
    in url() (line 2652 of /core/includes/common.inc).
    ```

indigoxela:
Fixes https://github.com/backdrop/backdrop-issues/issues/5398

### DIFF
--- a/core/includes/common.inc
+++ b/core/includes/common.inc
@@ -1822,9 +1822,10 @@ function filter_xss($string, $allowed_tags = NULL) {
   }
   // Only operate on valid UTF-8 strings. This is necessary to prevent cross
   // site scripting issues on Internet Explorer 6.
-  if (!$string || !backdrop_validate_utf8($string)) {
+  if (!backdrop_validate_utf8($string)) {
     return '';
   }
+  $string = (string) $string;
   // Store the text format.
   _filter_xss_split($allowed_tags, TRUE);
   // Remove NULL characters (ignored by some browsers).
--- a/core/includes/common.inc
+++ b/core/includes/common.inc
@@ -5982,6 +5982,7 @@ function _backdrop_bootstrap_full() {
   require_once BACKDROP_ROOT . '/core/includes/actions.inc';
   require_once BACKDROP_ROOT . '/core/includes/ajax.inc';
   require_once BACKDROP_ROOT . '/core/includes/token.inc';
+  require_once BACKDROP_ROOT . '/core/includes/errors.inc';
 
   // Detect string handling method
   unicode_check();
--- a/core/includes/common.inc
+++ b/core/includes/common.inc
@@ -5978,7 +5978,6 @@ function _backdrop_bootstrap_full() {
   require_once BACKDROP_ROOT . '/core/includes/actions.inc';
   require_once BACKDROP_ROOT . '/core/includes/ajax.inc';
   require_once BACKDROP_ROOT . '/core/includes/token.inc';
-  require_once BACKDROP_ROOT . '/core/includes/errors.inc';
 
   // Detect string handling method
   unicode_check();
--- a/core/includes/common.inc
+++ b/core/includes/common.inc
@@ -770,7 +770,7 @@ function backdrop_parse_url($url) {
  *   The encoded path.
  */
 function backdrop_encode_path($path) {
-  return str_replace('%2F', '/', rawurlencode($path));
+  return str_replace('%2F', '/', rawurlencode((string) $path));
 }
 
 /**
@@ -1102,7 +1102,7 @@ function backdrop_http_request($url, array $options = array()) {
   // or PUT request. Some non-standard servers get confused by Content-Length in
   // at least HEAD/GET requests, and Squid always requires Content-Length in
   // POST/PUT requests.
-  $content_length = strlen($options['data']);
+  $content_length = strlen((string) $options['data']);
   if ($content_length > 0 || $options['method'] == 'POST' || $options['method'] == 'PUT') {
     $options['headers']['Content-Length'] = $content_length;
   }
@@ -1178,7 +1178,7 @@ function backdrop_http_request($url, array $options = array()) {
   $result->headers = array();
 
   // Parse the response headers.
-  while ($line = trim(array_shift($response))) {
+  while ($line = trim((string) array_shift($response))) {
     list($name, $value) = explode(':', $line, 2);
     $name = strtolower($name);
     if (isset($result->headers[$name]) && $name == 'set-cookie') {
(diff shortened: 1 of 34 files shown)


core/includes/common.inc
```diff
@@ -5977,7 +5982,6 @@ function _backdrop_bootstrap_full() {
   require_once BACKDROP_ROOT . '/core/includes/actions.inc';
   require_once BACKDROP_ROOT . '/core/includes/ajax.inc';
   require_once BACKDROP_ROOT . '/core/includes/token.inc';
-  require_once BACKDROP_ROOT . '/core/includes/errors.inc';
```
indigoxela: If I keep this line, Backdrop breaks completely. I don't understand why. Breaks means it breaks in real life when using Backdrop, not only the test runs.

core/modules/simpletest/backdrop_web_test_case.php
```diff
@@ -11,6 +11,12 @@
  */
 global $backdrop_test_info;
 
+/**
+ * Experimental!
+ * @todo Figure out what breaks require_once inside functions.
+ */
+require_once __DIR__ . '/../../../core/includes/errors.inc';
```
indigoxela: This made a whole bunch of tests work all of a sudden.

Do we have to think over where and when errors.inc gets included throughout core?

indigoxela: I still do have concerns re inclusion of errors.inc (as we do it now, it might be too late for some problem), but this one is obsolete now.

core/modules/path/tests/path_pattern.test
```diff
@@ -115,10 +115,10 @@ class PathPatternTestHelper extends BackdropWebTestCase {
     return $term;
   }
 
-  function assertEntityPattern($entity_type, $bundle, $language = LANGUAGE_NONE, $expected) {
+  function assertEntityPattern($entity_type, $bundle, $langcode, $expected) {
```
jenlampton, line 118: nice catch here!

core/includes/database/mysql/database.inc
```diff
@@ -68,6 +68,11 @@ class DatabaseConnection_mysql extends DatabaseConnection {
       PDO::MYSQL_ATTR_USE_BUFFERED_QUERY => TRUE,
       // Because MySQL's prepared statements skip the query cache, because it's dumb.
       PDO::ATTR_EMULATE_PREPARES => TRUE,
+      // Convert numeric values to strings when fetching. In PHP 8.1,
+      // PDO::ATTR_EMULATE_PREPARES now behaves the same way as non emulated
+      // prepares and returns integers. See https://externals.io/message/113294
+      // for further discussion.
+      \PDO::ATTR_STRINGIFY_FETCHES => TRUE,
```
quicksketch, line 75: I wasn't sure about including this change, since it may be hard to go back from, but my impression is that all integers are returned as strings currently, and this maintains that same behavior in PHP 8.1+. Additionally both D7 and D9 use this same change, so it will be expected with modules being ported. This seems like it's the best choice even if we don't particularly like it.

quicksketch:
Merged into 1.x and 1.21.x. Great work @indigoxela!

PR #1739, opened 2017-01-14
#2315 Link to updates page

laryn:
Serundeputy laryn update link v2
Fixes test error.

PR #3880, opened 2021-12-31
Issue #1739: Entity view() documentation follow-ups.

quicksketch:
Small follow-ups for https://github.com/backdrop/backdrop-issues/issues/1739 that add `@since` lines and update a few docblock parameters and return values.

### DIFF
--- a/core/modules/entity/entity.class.inc
+++ b/core/modules/entity/entity.class.inc
@@ -132,6 +132,7 @@ interface EntityInterface {
    *   (optional) A language code to use for rendering. Defaults to the global
    *   content language of the current request.
    *
+   * @since 1.21.0 Method added.
    */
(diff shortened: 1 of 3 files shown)


klonos (approved):
LGTM 👍🏼

argiepiano:
I suggest also adding the `@since` lines to:

- `EntityControllerInterface::buildContent` and `EntityControllerInterface::view`
- `DefaultEntityController::buildContent` and `DefaultEntityController::view` 
- `Entity::buildContent` and `Entity::view`

quicksketch:
> I suggest also adding the `@since` lines to:

In my first pass I was only adding it to the lowest level of the inheritance tree, but I have no problem being a little redundant since I think it will help, especially during the first few releases while contrib transitions to using the new APIs.

I added all these new `@since` lines in the latest commit.

PR #3883, opened 2022-01-01
Issue #5425: A11Y - Make required asterisks AA

### LINKED ISSUE #5425: [A11Y] Fixed: Increase contrast for "required" asterisk in all core themes.
This is a sibling issue to #5426, and part of the #5244 meta.

Current style is:
- Seven: `#FF0000` text color over `#FAFAFA` background, which is AA for large text, but fails for normal size text.
- Basis/Bartik: `#FF0000` text color over `#FFFFFF` background, which is AA for large text, but fails for normal size text.

Making the text color `#E60000` would pass AA for Seven/Basis/Bartik (would make it AAA for large size text, and AA for normal size text).

Making it `#AD0000` would pass AAA for Seven/Basis/Bartik, for both large and normal size text.

klonos:
Fixes https://github.com/backdrop/backdrop-issues/issues/5425

Alternative to https://github.com/backdrop/backdrop/pull/3884

### DIFF
--- a/core/modules/system/css/system.theme.css
+++ b/core/modules/system/css/system.theme.css
@@ -119,7 +119,7 @@ input.form-radio {
   vertical-align: middle;
 }
 .form-required {
-  color: #f00;
+  color: #e60000;
 }
 .marker{
   font-size: 0.9em;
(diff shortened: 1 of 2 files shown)


quicksketch:

PR #3886, opened 2022-01-01
Issue #5426: A11y - make secondary submit buttons AAA

### LINKED ISSUE #5426: [A11Y] Seven/Basis: secondary submit button text does not have enough contrast
This is a sibling issue to #5425 and part of the #5244 meta.

Current style is `#828188` text color over `#FFFFFF` background, which is AA for large text, but fails for normal size text.

Making the text color `#75747B` would pass AA for both Seven/Basis (would make it AAA for large size text, and AA for normal size text).

Making it `#57575C` would pass AAA for both Seven/Basis, for both large and normal size text.

PS: we need to also:
- match the button border color with the text color, and then make sure that that has enough contrast against the page background
- check contrast of these buttons on hover as well

klonos:
Fixes https://github.com/backdrop/backdrop-issues/issues/5426

Alternative to https://github.com/backdrop/backdrop/pull/3885

### DIFF
--- a/core/themes/basis/css/skin.css
+++ b/core/themes/basis/css/skin.css
@@ -570,23 +570,26 @@ button.button-primary:active {
 a.button-secondary,
 input.button-secondary,
 button.button-secondary {
-  color: #828188;
+  color: #57575c;
   background-color: white;
   box-shadow: inset 0 0 0 0.125em #97979e;
 }
 
+a.button-secondary:hover,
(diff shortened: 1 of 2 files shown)


klonos:
Random failure in one of the php8 tests. Restarting...

ghost:
Closing in favour of https://github.com/backdrop/backdrop/pull/3885 based on feedback in the issue.

PR #3887, opened 2022-01-01
Issue #5427: Allow modules to provide flexible template row styles

### LINKED ISSUE #5427: Allow modules to provide flexible template row styles
## Description of the need

Flexible templates have predetermined set of row styles provided by the function [layout_flexible_row_styles()](https://github.com/backdrop/backdrop/blob/1.x/core/modules/layout/layout.module#L2284). There is a respectable selection there but if a user wanted a row with more that four columns, this is not provided; neither is there any way to provide alternate column widths.

We need a hook to allow modules to offer additional styles.

## Proposed solution

A new `hook_layout_flexible_row_styles()` and `hook_layout_flexible_row_styles_alter()`

None

docwilmot:
Fixes https://github.com/backdrop/backdrop-issues/issues/5427

### DIFF
--- a/core/modules/layout/css/layout.flexible.admin.css
+++ b/core/modules/layout/css/layout.flexible.admin.css
@@ -41,13 +41,18 @@
 
 .layout-flexible-editor .flexible-row {
   padding: 0;
-  border: 1.25px dashed #0074bd;
   margin-bottom: 15px;
   border-radius: 4px;
   -webkit-box-shadow: 3px 3px 0 0 #c6c6c6;
   -moz-box-shadow: 3px 3px 0 0 #c6c6c6;
   box-shadow: 3px 3px 0 0 #c6c6c6;
 }
+
+.layout-flexible-editor .flexible-row.error {
+  border-color: #ff0000;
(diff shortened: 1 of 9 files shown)


stpaultim:
@docwilmot Any chance you would be able to rebase this PR to build a sandbox. I may be willing to work on this issue and/or advocate for moving it forward.

docwilmot:
I seem to remember this coming up on something difficult to solve, then deciding that no one else ever asked for this so probably not worth the time. You think its worth it? Never been even mentioned anywhere.

stpaultim:
@docwilmot You may be right. I'm not sure that this is an important need. I was thinking of this issue:
https://github.com/backdrop/backdrop-issues/issues/4052 

@olafgrabienski in particular is interested in that one.

olafgrabienski:
@stpaultim Thanks for mentioning https://github.com/backdrop/backdrop-issues/issues/4052. I'm still interested in custom IDs for rows of flexible layout templates. Not sure if this PR here would help to achieve custom row IDs. Maybe in an indirect way?

PR #3890, opened 2022-01-02
5399 configure layout page compact

### LINKED ISSUE #5399: [UX][DX] Improve the description for the Path field on the configure layout page.

bugfolder:
Fixes https://github.com/backdrop/backdrop-issues/issues/5399.

This version provides a more compact presentation of the additional information, putting it all in the description text under the field. <strike>It also leaves off the URL prefix.</strike>

### DIFF
--- a/core/modules/layout/css/layout.admin.css
+++ b/core/modules/layout/css/layout.admin.css
@@ -85,6 +85,34 @@
   margin-left: 0;
 }
 
+/* Inject a little space after the path matches text within the description. */
+#path-matches {
+  margin-bottom: 0.5em;
+}
+
+/* Toggle for the placeholder examples */
+.placeholder-examples-toggle {
+  white-space: nowrap;
+}
+.placeholder-examples-toggle .arrow {
+  border-bottom-color: transparent;
+  border-left-color: transparent;
+  border-right-color: transparent;
+  border-style: solid;
+  border-width: 0.3333em 0.3333em 0;
+  display: inline-block;
+  height: 0;
+  line-height: 0;
+  width: 0;
+  overflow: hidden;
+  margin: 0.15em 0.3333em;
+}
+.placeholder-examples-toggle .arrow.close {
+  border-top-color: transparent;
+  border-bottom-color: #0074bd;
+  border-width: 0 0.3333em 0.3333em;
+}
+
 /* Menu item argument settings. */
 #layout-arguments strong,
 #layout-context-table strong {
--- a/core/modules/layout/css/layout.admin.css
+++ b/core/modules/layout/css/layout.admin.css
@@ -86,15 +86,18 @@
 }
 
 /* Inject a little space after the path matches text within the description. */
-#path-matches {
+#layout-path-matches {
   margin-bottom: 0.5em;
+}
+.layout-placeholder-examples {
+  margin-top: 0.5em;
 }
 
 /* Toggle for the placeholder examples */
-.placeholder-examples-toggle {
+.layout-placeholder-examples-toggle {
   white-space: nowrap;
 }
-.placeholder-examples-toggle .arrow {
+.layout-placeholder-examples-toggle .arrow {
   border-bottom-color: transparent;
   border-left-color: transparent;
   border-right-color: transparent;
@@ -107,7 +110,7 @@
   overflow: hidden;
   margin: 0.15em 0.3333em;
 }
-.placeholder-examples-toggle .arrow.close {
+.layout-placeholder-examples-toggle .arrow.close {
   border-top-color: transparent;
   border-bottom-color: #0074bd;
   border-width: 0 0.3333em 0.3333em;
--- a/core/modules/layout/js/layout.admin.js
+++ b/core/modules/layout/js/layout.admin.js
@@ -40,10 +40,10 @@ Backdrop.behaviors.layoutConfigure = {
     function examples_toggle_handler(e) {
       var $examples = $(this).next().toggle();
       if ($examples.is(':visible')) {
-        $(this).text(Backdrop.t('less')).append('<span class="arrow close"></span>');
+        $(this).text(Backdrop.t('Hide examples')).append('<span class="arrow close"></span>');
       }
       else {
-        $(this).text(Backdrop.t('more')).append('<span class="arrow"></span>');
+        $(this).text(Backdrop.t('Show examples')).append('<span class="arrow"></span>');
       }
       e.preventDefault();
       e.stopPropagation();
--- a/core/modules/layout/js/layout.admin.js
+++ b/core/modules/layout/js/layout.admin.js
@@ -33,6 +33,22 @@ Backdrop.behaviors.layoutList = {
  */
 Backdrop.behaviors.layoutConfigure = {
   attach: function(context) {
+
+    /**
+     * Toggle handler for the placeholder examples.
+     */
+    function examples_toggle_handler(e) {
+      var $examples = $(this).next().toggle();
+      if ($examples.is(':visible')) {
+        $(this).text(Backdrop.t('less')).append('<span class="arrow close"></span>');
+      }
+      else {
+        $(this).text(Backdrop.t('more')).append('<span class="arrow"></span>');
+      }
+      e.preventDefault();
+      e.stopPropagation();
+    };
+
     var $form = $('.layout-settings-form').once('layout-settings');
     if ($form.length && Backdrop.ajax) {
       var ajax = Backdrop.ajax['edit-path-update'];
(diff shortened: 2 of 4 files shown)


core/modules/layout/js/layout.admin.js
```diff
@@ -52,6 +72,12 @@ Backdrop.behaviors.layoutConfigure = {
       });
     }
 
+    // Hide the placeholder examples.
+    $form.find('.placeholder-examples').hide();
+
+    // Handle toggling the placeholder examples.
+   $('a.placeholder-examples-toggle').click(examples_toggle_handler);
```
klonos:
```suggestion
    $('a.placeholder-examples-toggle').click(examples_toggle_handler);
```

core/themes/seven/css/style.css
```diff
@@ -808,6 +808,28 @@ details .details-child-wrapper {
   margin: 10px 0;
 }
 
+/**
+ * Description version of details element.
+ */
+details.description {
+  padding: 6px 15px;
+  border-radius: 0 0 4px 4px;
+  border: 2px solid #EAEAEA;
```
klonos: We always use lowercase in hex colors in CSS:

```suggestion
  border: 2px solid #eaeaea;
```

core/themes/seven/css/style.css
```diff
@@ -808,6 +808,28 @@ details .details-child-wrapper {
...
+  margin: 1em 0;
+  overflow: hidden;
+  background-color: transparent;
+  display: table-cell;
+}
+details.description summary span {
+  color: #000000;
```
klonos: We can use the short variation for the hex color here (consistency):

```suggestion
  color: #000;
```

klonos:
@bugfolder can you please close the other PR at https://github.com/backdrop/backdrop/pull/3865 to avoid confusion? Thanks.

core/themes/seven/css/style.css
bugfolder: Ah; that's residue from when we used the details element in this PR. We're no longer using it in the latest PR, so I will remove that styling. @jenlampton (who introduced it) might want to re-introduce it as a general styling issue for details element as a new issue.

core/themes/seven/css/style.css
bugfolder: Ditto.

core/modules/layout/layout.admin.inc
```diff
@@ -667,18 +743,21 @@ function layout_settings_form_validate(&$form, &$form_state) {
 }
 
 /**
- * Submit handler for layout_settings_form() that updates available contexts.
+ * Submit handler for layout_settings_form() that updates available contexts
+ * and information about path matching.
```
quicksketch: Function summary description should fit on a single 80 character line. You could put this onto it's own separate line:

```suggestion
 * Submit handler for layout_settings_form().
 *
 * Updates available contexts and information about path matching.
```

core/modules/layout/layout.admin.inc
```diff
@@ -667,18 +743,21 @@ function layout_settings_form_validate(&$form, &$form_state) {
...
  */
 function layout_settings_form_path_rebuild($form, &$form_state) {
   $form_state['rebuild'] = TRUE;
 }
 
 /**
- * AJAX handler for layout_settings_form() the returns updated contexts.
+ * AJAX handler for layout_settings_form() the returns updated contexts and
+ * information about path matching.
```
```suggestion
 * AJAX handler for layout_settings_form().
 *
 * Returns updated contexts and information about path matching.
```

core/modules/layout/layout.admin.inc
```diff
@@ -281,23 +281,89 @@ function layout_settings_form($form, &$form_state, Layout $layout) {
...
+  if (empty($path)) {
+    $path_matches_text = t('Please enter a path.');
+  }
+  else {
+    $path_matches = db_query('
+      SELECT path, page_callback
+      FROM {menu_router}
+      WHERE path = :path
+      ', array(':path' => $path))
+      ->fetchAllKeyed();
```
quicksketch: Could you used `menu_get_item()` here instead of a direct database query? Seems like we should have some function for loading menu items by router path. If not, perhaps put this in a private function?

core/modules/layout/layout.admin.inc
```diff
@@ -281,23 +281,89 @@ function layout_settings_form($form, &$form_state, Layout $layout) {
...
+  $form['path_matches_text'] = array(
+    '#type' => 'value',
+    '#value' => $path_matches_text,
+  );
+  $path_matches = array(
+    '#type' => 'container',
+    '#access' => !$layout->isDefault(),
+    '#attributes' => array(
+      'id' => array('path-matches'),
+    ),
```
quicksketch: I try to avoid IDs for most things these days. Could you make this a class an namespace it under `layout`, like `class="layout-path-matches"`

core/modules/layout/js/layout.admin.js
```diff
@@ -52,6 +72,12 @@ Backdrop.behaviors.layoutConfigure = {
       });
     }
 
+    // Hide the placeholder examples.
+    $form.find('.placeholder-examples').hide();
+
+    // Handle toggling the placeholder examples.
+    $('a.placeholder-examples-toggle').click(examples_toggle_handler);
```
quicksketch: Can we namespace all these classes to ensure no conflicts from other CSS/JS files? `.layout-placeholder-examples` and `a.layout-placeholder-examples-toggle`.

core/modules/layout/css/layout.admin.css
```diff
@@ -85,6 +85,34 @@
   margin-left: 0;
 }
 
+/* Inject a little space after the path matches text within the description. */
+#path-matches {
+  margin-bottom: 0.5em;
+}
```
quicksketch: Adding this same margin before the examples would be beneficial.

image

See my other comments about using a class and namespacing these selectors.

core/modules/layout/layout.admin.inc
bugfolder: Take a look at function `layout_settings_form_path_ajax()`. All three `ajax_command_html()` calls use IDs. Why should they not match?

core/modules/layout/css/layout.admin.css
bugfolder: Margin added.

core/modules/layout/js/layout.admin.js
bugfolder: Yes. done.

core/modules/layout/layout.admin.inc
bugfolder: `menu_get_item()` is not the right tool; it takes a normal path (`node/1`) and returns the menu router path (`node/%`). This function is checking for the existence of the menu router path itself. So a function it will be.

I created `_layout_menu_router_path_exists()` for this function. But it looks like something that might be useful elsewhere, so I raise the question of whether this should be a public function in `menu.inc`, perhaps called `menu_router_path_exists()`?

core/modules/layout/layout.admin.inc
quicksketch: Sorry I didn't see that this was part of an AJAX command, in which case I think this is fine :+1:

Might be good to namespace it still, i.e. `layout-path-matches`

bugfolder: OK. I did namespace it as you suggested in the latest PR update.

core/modules/layout/layout.admin.inc
quicksketch: I think leaving it private unless we need it in multiple places seems fine.

bugfolder: OK!

quicksketch:

PR #3904, opened 2022-01-12
Issue #5447: Report db type and version (2 digits)

### LINKED ISSUE #5447: Fix mysql/mariadb version for Telemetry and the Status Report
## Description of the need

We capture the database type and version, but the outcome doesn't seem correct. It's at least questionable, as I'd expect MariaDB to be used more often.

wrong-db-stats

## Proposed solution

Seems like we'd need to evaluate the version string more accuratly.

## Additional information

Some examples from actual strings on the status page, by OS:

- Debian Buster: `5.5.5-10.3.31-MariaDB-0+deb10u1`
- Debian Buster: `5.7.36-1` - hey that's actually MySQL
- Debian Bullseye: `5.5.5-10.5.12-MariaDB-0+deb11u1`
- AlmaLinux 8: `5.5.5-10.3.28-MariaDB`
- CentOS 7: `5.5.5-10.3.32-MariaDB`
- Ubuntu Focal Fossa: `5.5.5-10.5.11-MariaDB-1:10.5.11+maria~focal-log`

All MariaDB versions start with 5.5.5, so I think, all those get counted as MySQL 5.5. My suspicion: the majority in our stats are actually MariaDB.

I didn't find an existing issue for that, please point me to it if there is.

Currently the Telemetry module does a [super-simple check](https://github.com/backdrop/backdrop/blob/1.x/core/modules/telemetry/telemetry.telemetry.inc#L37).
We could either fix it there by enhancing the code, or do better evaluation server-side.

---

Later Note: Even Mariadb versions starting with 10 are sometimes reported as 5.5 (as on Pantheon). See https://github.com/backdrop-ops/backdrop-pantheon/issues/68#issuecomment-1433751043

klonos:
Fixes https://github.com/backdrop/backdrop-issues/issues/5447

Reports db type and version, so the version number decimals/accuracy cannot be handled by the current trimming/aggregation options provided by the `project_telemetry` module. That's why the reported value is trimmed down to only 2 digits in `core/modules/telemetry/telemetry.telemetry.inc` (before it gets sent to b.org).

image

The above would previously be reported as `5.5.5-10.3.27-MariaDB`, which would result in `5.5.5` in the telemetry stats on b.org

### DIFF
--- a/core/modules/telemetry/telemetry.telemetry.inc
+++ b/core/modules/telemetry/telemetry.telemetry.inc
@@ -9,13 +9,13 @@
  */
 function telemetry_telemetry_info() {
   $info['php_version'] = array(
-    'label' => t('PHP Version'),
-    'description' => t('The current version of PHP running on your server.'),
+    'label' => t('PHP version'),
+    'description' => t('The version of PHP running on your server.'),
     'project' => 'backdrop',
   );
   $info['mysql_version'] = array(
-    'label' => t('MySQL Version'),
-    'description' => t('The version number of your database (either MySQL or MariaDB).'),
+    'label' => t('Database version'),
+    'description' => t('The version of your database server (either MySQL or MariaDB).'),
     'project' => 'backdrop',
   );
   $info['server_os'] = array(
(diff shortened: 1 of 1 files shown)


jenlampton:
Replaced by https://github.com/backdrop/backdrop/pull/4346/

PR #3905, opened 2022-01-13
Issue #5450: Disable sending Telemetry data from GitHub Actions

### LINKED ISSUE #5450: Disable sending Telemetry data from GitHub Actions during test runs
This is a follow-up from #5444.

@quicksketch 
> >  Should we use the same setting for GHA? (Follow-up?)
>
>Note that indeed this only solves _PR Sandboxes_, it doesn't yet solve the problem with GitHub Action tests reporting Telemetry data, which is probably the biggest contributor to misreporting, since 9 test instances are spun up every PR.

indigoxela:
Fixes https://github.com/backdrop/backdrop-issues/issues/5450

### DIFF
--- a/.github/misc/settings.local.php
+++ b/.github/misc/settings.local.php
@@ -1,0 +1,8 @@
+<?php
+/**
+ * @file
+ * Disable sending of telemetry data from GitHub Action runners.
+ * @see .github/workflows/functional-tests.yml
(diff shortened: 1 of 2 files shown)


quicksketch:
Merged into 1.x for 1.21.0.

PR #3906, opened 2022-01-13
Issue #5451: Fix comment for system_update_1079

### LINKED ISSUE #5451: Incomplete comment opener for system_update_1079 in system.install
## Description of the bug

A tiny problem, actually no real problem, I just realized that the message extraction for bee's updb doesn't work as expected then.

[Source code](https://github.com/backdrop/backdrop/blob/50dad22dae9d615812197a6027236f872fc99950/core/modules/system/system.install#L3251)

That should be two asterisks. :smile:

indigoxela:
Fixes https://github.com/backdrop/backdrop-issues/issues/5451

My smallest PR ever... :grinning:

### DIFF
--- a/core/modules/system/system.install
+++ b/core/modules/system/system.install
@@ -3248,7 +3248,7 @@ function system_update_1078() {
   }
 }
 
-/*
+/**
  * Clear the cache to reset allowlist settings.
  */
 function system_update_1079() {


laryn:
Good catch, thanks for the fix (whatever the size) @indigoxela! Merged into `1.x`.

PR #3908, opened 2022-01-15
Issue #5459: People should not be referred to as users.

### LINKED ISSUE #5459: Don't call people "Users" in new token description text
Follow up to https://github.com/backdrop/backdrop-issues/issues/5445

Strings replaced:
* `...This token is only available when the system sends an Account Cancellation email to a user.`
* `...This token is only available when the system sends a Password Recovery email to a user.`

These new strings should be as follows:

Strings replaced:
* `...This token is only available when an Account Cancellation email is sent.`
* `...This token is only available when a Password Recovery email is sent.`

---

##Translation update:

**This assumes the translation update for https://github.com/backdrop/backdrop-issues/issues/5445 was skipped.**

Strings replaced:
* `The URL of the confirm delete page for the user account.` becomes `The URL of the confirm delete page for the user account. This token is only available when an Account Cancellation email is sent`
* `The URL of the one-time login page for the user account.` becomes `The URL of the one-time login page for the user account. This token is only available when a Password Recovery email is sent.`

jenlampton:
Fixes https://github.com/backdrop/backdrop-issues/issues/5459

### DIFF
--- a/core/modules/user/user.tokens.inc
+++ b/core/modules/user/user.tokens.inc
@@ -74,22 +74,20 @@ function user_token_info() {
     'type' => 'url',
   );
 
-  // This token is replaced only when the system sends an Account Cancellation
-  // email to a user.
+  // This token is replaced only when an Account Cancellation email is sent.
   // @see user_mail_tokens()
   $user['cancel-url'] = array(
     'name' => t('Account cancellation URL'),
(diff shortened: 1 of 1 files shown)


klonos (approved):
LGTM 👍🏼

quicksketch:

PR #3909, opened 2022-01-18
Issue #5462: Separate option descriptions for book navigation block.

### LINKED ISSUE #5462: [UX] Separate descriptions for book navigation block settings.
## Description of the bug

The description text for the `Book navigation block display` block is all lumped together into a very long paragraph, instead of being separated into a description for each item. 
Screen Shot 2022-01-17 at 5 17 51 PM

Current description text:
`If Show block on all pages is selected, the block will contain the automatically generated menus for all of the site's books. If Show block only on book pages is selected, the block will contain only the one menu corresponding to the current page's book. In this case, if the current page is not in a book, no block will be displayed. The Page specific visibility settings or other visibility settings can be used in addition to selectively display this block.`

Recommended solution:
**Show block on all pages**
`The block will contain the automatically generated menus for all of the site's books.`
** Show block only on book pages**
`The block will contain only the one menu corresponding to the current page's book. If the current page is not in a book, no block will be displayed.`

Overall description:
`Add additional visibility conditions to display this block more selectively.`

Screen Shot 2022-01-17 at 5 25 31 PM

## Steps To Reproduce

To reproduce the behavior:
1. Enable the book module
2. Place the `Book Navigation` block
3. Examine settings

## Actual behavior

Description text is confusing.

## Additional information

- Backdrop CMS version: 1.x

jenlampton:
Fixes https://github.com/backdrop/backdrop-issues/issues/5462

### DIFF
--- a/core/modules/book/book.module
+++ b/core/modules/book/book.module
@@ -335,8 +335,8 @@ function book_block_configure($delta = '', $settings = array()) {
     '#options' => $options,
     '#default_value' => $settings['book_mode'],
   );
-  $form['book_mode']['all pages']['#description'] = t('The block will contain the automatically generated menus for all of the site\'s books.');
-  $form['book_mode']['book pages']['#description'] = t('The block will contain only the one menu corresponding to the current page\'s book. If the current page is not in a book, no block will be displayed.');
+  $form['book_mode']['all pages']['#description'] = t("The block will contain the automatically generated menus for all of the site's books.");
+  $form['book_mode']['book pages']['#description'] = t("The block will contain only the one menu corresponding to the current page's book. If the current page is not in a book, no block will be displayed.");
 
   return $form;
 }
--- a/core/modules/book/book.module
+++ b/core/modules/book/book.module
@@ -334,8 +334,9 @@ function book_block_configure($delta = '', $settings = array()) {
     '#title' => t('Book navigation block display'),
     '#options' => $options,
     '#default_value' => $settings['book_mode'],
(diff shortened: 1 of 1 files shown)


core/modules/book/book.module
```diff
@@ -334,8 +334,10 @@ function book_block_configure($delta = '', $settings = array()) {
     '#title' => t('Book navigation block display'),
     '#options' => $options,
     '#default_value' => $settings['book_mode'],
-    '#description' => t("If <em>Show block on all pages</em> is selected, the block will contain the automatically generated menus for all of the site's books. If <em>Show block only on book pages</em> is selected, the block will contain only the one menu corresponding to the current page's book. In this case, if the current page is not in a book, no block will be displayed. The <em>Page specific visibility settings</em> or other visibility settings can be used in addition to selectively display this block."),
+    '#description' => t('Add additional visibility conditions to display this block more selectively.'),
   );
+  $form['book_mode']['all pages']['#description'] = t('The block will contain the automatically generated menus for all of the site\'s books.');
+  $form['book_mode']['book pages']['#description'] = t('The block will contain only the one menu corresponding to the current page\'s book. If the current page is not in a book, no block will be displayed.');
```
klonos: This is a nice use case where we'd use double quotes, in order to avoid having to escape the apostrophes:

```suggestion
  $form['book_mode']['all pages']['#description'] = t("The block will contain the automatically generated menus for all of the site's books.");
  $form['book_mode']['book pages']['#description'] = t("The block will contain only the one menu corresponding to the current page's book. If the current page is not in a book, no block will be displayed.");
```

jenlampton: Double quotes is slower in PHP because PHP is searching for variable replacements. Since there are no variables to replace in the strings here, I think escaping is the better option.

klonos: That's not what our coding standards say though: https://docs.backdropcms.org/php-standards#quotes

klonos (approved):
RTBC 👍🏼

quicksketch:

PR #3910, opened 2022-01-19
Issue #5464: Status Report should flag duplicate module installs

### LINKED ISSUE #5464: Status Report should flag duplicate module installs
## Description of the need

Modules can be installed in the root of the modules folder or in a subfolder.  This makes it possible to install multiple copies of the same module onto one Backdrop system.  Backdrop will choose one of these modules to make live, but it's (at best) not clear which one it has chosen, and a user can easily become confused, thinking a different module is loaded than the one that really is.

## Proposed solution

A test can be added to the status report which would alert the administrators that duplicate modules are installed.

Another option would be to allow duplicate modules to appear multiple times on the modules page.

## Additional information

The Status Report now flags duplicate modules.

bugfolder:
Fixes https://github.com/backdrop/backdrop-issues/issues/5464.

### DIFF
--- a/core/modules/system/css/system.admin.css
+++ b/core/modules/system/css/system.admin.css
@@ -234,9 +234,6 @@ span.admin-missing {
     white-space: nowrap;
   }
 }
-.system-status-report .duplicate-module-overriding {
-  font-weight: bold;
-}
 
 /**
  * Theme settings.
--- a/core/modules/system/css/system.admin.css
+++ b/core/modules/system/css/system.admin.css
@@ -234,7 +234,9 @@ span.admin-missing {
     white-space: nowrap;
   }
 }
-
+.system-status-report .duplicate-module-overriding {
+  font-weight: bold;
+}
 
 /**
  * Theme settings.
--- a/core/modules/system/system.install
+++ b/core/modules/system/system.install
@@ -530,7 +530,7 @@ function system_requirements($phase) {
     }
   }
 
-  if ($phase == 'runtime' && settings_get('multiple_module_version_warnings')) {
+  if ($phase == 'runtime' && !settings_get('disable_multiple_modules_warnings')) {
     // Check for multiple versions of the same module.
     $files = system_rebuild_module_data();
     foreach ($files as $filename => $file) {
--- a/core/modules/system/system.install
+++ b/core/modules/system/system.install
@@ -536,9 +536,9 @@ function system_requirements($phase) {
     if ($path && strpos($path, 'core/modules/ckeditor') === FALSE) {
       $requirements['ckeditor_replaced'] = array(
         'title' => $t('CKEditor'),
-        'value' => $t('Duplicate module detected'),
+        'value' => $t('Override of core module %name detected', array('%name' => 'ckeditor')),
         'severity' => REQUIREMENT_WARNING,
-        'description' => $t('Backdrop core now provides a bundled CKEditor module. A different copy of CKEditor module has been located at %path. To use the new CKEditor module, disable and then uninstall the current module, remove it from your installation, and then enable the new CKEditor module from the modules page. For more information see the <a href="https://api.backdropcms.org/node/42827" target="_blank">CKEditor change notice</a>.', array('%path' => BACKDROP_ROOT . '/' . $path)),
+        'description' => $t('Backdrop core provides a bundled CKEditor module. A different copy of CKEditor module is at %path. To use the new CKEditor module, disable and then uninstall the current module, remove it from your installation, and then enable the new CKEditor module from the modules page. For more information see the <a href="https://api.backdropcms.org/node/42827" target="_blank">CKEditor change notice</a>.', array('%path' => BACKDROP_ROOT . '/' . $path)),
       );
     }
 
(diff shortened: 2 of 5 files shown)


core/modules/system/css/system.admin.css
```diff
@@ -234,7 +234,9 @@ span.admin-missing {
     white-space: nowrap;
   }
 }
-
+.system-status-report .duplicate-module-overriding {
```
indigoxela: An alternative to styling it more prominently via font-weight would be to wrap the actual path in `<strong>`.
So it not only looks more important, but the markup also marks it as being more important (might be relevant for assistive technologies).

bugfolder: Makes sense. Done, thanks!

core/modules/system/system.install
```diff
@@ -531,59 +531,45 @@ function system_requirements($phase) {
...
+        elseif ($file->name == 'link') {
+          $description .= $t('Backdrop core provides a bundled Link module. A different copy of Link module is at %path. Remove this module from your installation to use the core Link module. For more information see the <a href="https://api.backdropcms.org/node/42930" target="_blank">Link change notice</a>.', array('%path' =>  $path));
+        }
+        elseif ($file->name == 'date') {
+          $description .= $t('Backdrop core provides a bundled Date module. A different copy of Date module is at %path. Remove this module from your installation to use the core Date module. For more information see the <a href="https://api.backdropcms.org/node/43066" target="_blank">Date change notice</a>.', array('%path' => $path));
+        }
+        elseif ($file->name == 'redirect') {
+          $description .= $t('Backdrop core provides a bundled Redirect module. A different copy of Redirect module is at %path. Remove this module from your installation to use the core Redirect module. For more information see the <a href="https://api.backdropcms.org/node/43994" target="_blank">Redirect change notice</a>.', array('%path' => $path));
+        }
```
quicksketch: This could be consolidated into a single string and used for all of these (and future ones):

```
           $description .= $t('Backdrop core provides a bundled @name module. A different copy of @name module is at %path. Remove this module from your installation to use the core @name module. For more information see the <a href="!url" target="_blank">@name change notice</a>.', array('%path' => $path, '@name' => t('Redirect'), '@core_name' => t('Redirect'), '!url' => 'https://api.backdropcms.org/node/43994'));
```

bugfolder: If we make that change, then all of the modules will be sent to the change record for Redirect. Note that the node ID is  different in each URL. They can still be consolidated, but we'll need a switch statement for the URL. (Also we can take the opportunity to use the docs.b.org URL, rather than api.b.org.)

settings.php
```diff
@@ -407,6 +407,19 @@
...
+ *
+ * If multiple versions of the same module are present (for example, a version
+ * of a module in the /modules directory that has the same name as a /core
+ * module), Backdrop will load only the last module encountered when Backdrop
+ * scans for module files and Backdrop will show a warning on the status report
+ * page. For some multisite configurations, one might wish to intentionally
+ * override a core or contrib module with a different version in the /sites
+ * folder. To disable the status report warning, set this value to FALSE.
+ */
+$settings['multiple_module_version_warnings'] = TRUE;
```
klonos: A few notes/suggestions:
- The first sentence is really long, and without punctuation it becomes difficult to read.
- We could split this huge paragraph into two smaller ones:
  - one to explain how module discovery works and when the warnings are thrown
  - another to explain the situations when it might be useful to disable the warnings.
- I think that it might be better to have the setting ready for silencing the warnings, but commented out.
- If the setting name was using a verb, then it'd be easier to scan and figure out what it does, and what TRUE/FALSE would do. Something like `disable_multiple_modules_warnings` perhaps.

How about something like this instead? (feel free to tweak):

```suggestion
/**
 * Suppress warnings of multiple instances of the same module being found.
 *
 * When scanning for module files, if Backdrop encounters multiple instances of
 * the same module (for example, a version of a module in the /modules directory
 * that has the same name as a module in /core), then only the last module will
 * be loaded. In such cases, Backdrop will show a warning on the status report
 * page.
 *
 * Having multiple versions of the same module may be intentional in certain use
 * cases though, such as in some multisite configurations, when there is need to
 * override a core or contrib module with a different version in the /sites
 * folder. In such cases, you may want to disable the status report warnings.
 *
 * Uncomment the line below to disable the warnings.
 */
// $settings['disable_multiple_modules_warnings'] = TRUE;
```

bugfolder: I like the description change!👍

That variable change makes the call at point of use

```
if (!settings_get('disable_multiple_modules_warnings')) {...}
```

I don't like the double-negative ("if not disable..."). There's a use case for `$settings['...'] = TRUE;`, and "To disable the status report warning, set this value to FALSE" seems pretty clear, too, and goes along with the general sense of "TRUE = enable, FALSE = disable."

bugfolder: After a good night's sleep, I've come around to this point of view (in part, because it makes the warning the default even for people who don't update their settings.php file). PR updated accordingly.

quicksketch:
Super! I've merge this into 1.x for 1.22.0.

PR #3514, opened 2021-01-25
Issue #4904: PDOException when uninstalling comment module

indigoxela:
Fixes https://github.com/backdrop/backdrop-issues/issues/4904

Prevent fatal error by flushing the cached information about entity types.

PR #3915, opened 2022-01-22
Issue #5470: Fatal error on node save

### LINKED ISSUE #5470: Fatal error on node save after disabling the comment module
## Description of the bug

When the comment module is not enabled, it's not possible to add new nodes.

## Steps To Reproduce

1. Disable and uninstall the comment module
2. Try to add a new node immediately after that

## Actual behavior

```
EntityStorageException: SQLSTATE[42S22]: Column not found: 1054 Unknown column 'comment_close_override' in 'field list' in NodeStorageController->save() (line 510 of ...core/modules/node/node.entity.inc). 
```
entity-storage-exception

## Expected behavior

Guess :wink: 

## Additional information

- Backdrop CMS version: 1.21.1
- PHP: 7.4

indigoxela:
Fixes https://github.com/backdrop/backdrop-issues/issues/5470

### DIFF
--- a/core/modules/comment/comment.install
+++ b/core/modules/comment/comment.install
@@ -66,6 +66,9 @@ function comment_uninstall() {
   if (db_field_exists('node', 'comment_close_override')) {
     db_drop_field('node', 'comment_close_override');
   }
+  // Flush outdated cache data.
+  // @todo Be more specific re cache prefix.
+  cache('cache')->flush();
 
   // Remove default, un-modified view.
   $config = config('views.view.comments_recent');
--- a/core/modules/comment/comment.install
+++ b/core/modules/comment/comment.install
@@ -67,8 +67,8 @@ function comment_uninstall() {
     db_drop_field('node', 'comment_close_override');
   }
   // Flush outdated cache data.
-  // @todo Be more specific re cache prefix.
-  cache('cache')->flush();
+  cache('cache')->deletePrefix('schema');
+  backdrop_static_reset('_node_types_build');
 
   // Remove default, un-modified view.
   $config = config('views.view.comments_recent');
--- a/core/modules/simpletest/tests/module.test
+++ b/core/modules/simpletest/tests/module.test
@@ -348,15 +348,16 @@ class ModuleUninstallTestCase extends BackdropWebTestCase {
     );
     $this->backdropPost('admin/modules/uninstall', $edit, t('Uninstall'));
     $this->backdropPost(NULL, array(), t('Uninstall'));
+
     $node_edit = array(
       'title' => 'This is a new node',
     );
     $this->backdropGet('node/add/post');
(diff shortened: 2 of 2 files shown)


core/modules/simpletest/tests/module.test
```diff
@@ -332,6 +332,32 @@ class ModuleUninstallTestCase extends BackdropWebTestCase {
...
+  /**
+   * Test that uninstalling Comment and then immediately adding a node works.
+   */
+  public function testUninstallCommentAddNode() {
+    $admin_user = $this->backdropCreateUser(array(
+      'administer modules',
+      'bypass node access',
+    ));
+    $this->backdropLogin($admin_user);
+    module_disable(array('comment'), FALSE);
```
klonos, line 345: This works, but should we be doing it via the UI instead (`$this->backdropPost('admin/modules', $edit, t('Save configuration'));`), same as below with the uninstallation of the module? ...if not (because the purpose of this test is not actually to test installation/disabling of modules), then should we be also doing the uninstallation manually via `backdrop_uninstall_modules()`?

core/modules/comment/comment.install
```diff
@@ -66,6 +66,9 @@ function comment_uninstall() {
   if (db_field_exists('node', 'comment_close_override')) {
     db_drop_field('node', 'comment_close_override');
   }
+  // Flush outdated cache data.
+  // @todo Be more specific re cache prefix.
+  cache('cache')->flush();
```
klonos: In https://github.com/backdrop/backdrop/pull/3514 (for https://github.com/backdrop/backdrop/issues/4904) we chose to clear caches via `backdrop_register_shutdown_function()` via an implementation of `hook_modules_uninstalled()` in the .module file instead of in the `hook_uninstall()` implementation in .install. Out of curiosity, why not do the same here?

Also, shouldn't we be making sure that this clearing of caches is done at the end of this function here? (after the deletion of the view config file)

core/modules/simpletest/tests/module.test
indigoxela: This test tries to capture problems when uninstalling the module, so disabling isn't actually relevant here. I chose module_disable for efficiency reasons (to not waste test run resources).

The uninstall, however, should be done in a realistic way - via UI. That's why I chose backdropPost here.

core/modules/comment/comment.install
indigoxela: The difference is the context in which the flushing is done.

The entity module acts on changes done by other modules, so it's important that flushing runs last (in the shutdown function). Otherwise it still might pick up outdated info.
The comment module acts on changes done in it's own uninstall hook.

indigoxela:
Turn it off and on again...

klonos (approved):
LGTM 👍🏼

quicksketch:

PR #3917, opened 2022-01-24
Issue #5472: Patch from Silkscreen - update SQL SELECT to work with MySQL and PostgreSQL.

jlfranklin:
Add the other columns from the SELECT to the GROUP BY clause to prevent an issue with PostgreSQL.

This is a patch added to Silkscreen CMS and submitted upstream to Backdrop to help reduce drift between the two projects.


quicksketch:
Thanks @jlfranklin!

PR #3920, opened 2022-01-27
Issue #5475. Adds `original` to vocabulary object sent to update and presave hooks

### LINKED ISSUE #5475: Provide `original` vocabulary to presave and update vocabulary hooks
## Description of the need

When vocabularies were entities in D7, whenever you updated a vocabulary, the presave and update hooks received the original (unchanged) vocabulary under $entity->original. In Backdrop, when vocabularies were demoted to config classes, this feature was removed.

It would be very helpful to bring back this feature. It's not hard to do. 

## Proposed solution

In the `save` method of TaxonomyVocabulary, we'd need to load the original vocabulary before it's saved, and add it to the vocabulary object under `original`. We would also need to create a `vocabulary_load_unchanged()` function to circumvent the cache and load the original vocabulary before it's saved.

I can probably provide a PR for this soon. But I'd like to ask if there is interest in this feature?

argiepiano:
Fixes https://github.com/backdrop/backdrop-issues/issues/5475

### DIFF
--- a/core/modules/taxonomy/taxonomy_vocabulary.class.inc
+++ b/core/modules/taxonomy/taxonomy_vocabulary.class.inc
@@ -69,10 +69,17 @@ class TaxonomyVocabulary {
    *   vocabulary was updated.
    */
   function save() {
-    module_invoke_all('taxonomy_vocabulary_presave', $this);
-
     $config = config('taxonomy.vocabulary.' . $this->machine_name);
     $is_new = $config->isNew();
+
+    if (!$is_new) {
+      $data = $config->get();
+      $original = new TaxonomyVocabulary($data);
+      $this->original = $original;
+    }
+
+    module_invoke_all('taxonomy_vocabulary_presave', $this);
+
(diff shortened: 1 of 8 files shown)


core/modules/taxonomy/taxonomy_vocabulary.class.inc
```diff
@@ -28,6 +28,15 @@ class TaxonomyVocabulary {
    */
   public $description;
 
+  /**
+   * Stores the original TaxonomyVocabulary object before changes during update operations.
+   *
+   * Outside of updates, this property is usually NULL.
+   *
+   * @var TaxonomyVocabulary
```
quicksketch: Nice, I think this is allowed to have the same property defined in child classes. And this override uses a more-specific declaration. Seems good to me!

quicksketch:
Merged into 1.x and 1.21.x. Thanks @argiepiano! Sorry for the run-around with the adding and removing of the property to the class objects.

PR #3922, opened 2022-01-28
Issue #3992: Increase minimum PHP version to 5.6.

### LINKED ISSUE #3992: Drop support for PHP versions less than 5.6.
Sibling issue to #214 (I thought we already had one for this, but I could not find anything).

None of the tools that I use to set up local environments (ddev, lando etc) support php versions below 5.6, which makes it hard for me to test things when there are failures in my PRs for php53, but not for php70.

Here's a few stats from around the net, to help us decide...

https://blog.packagist.com/php-versions-stats-2019-1-edition

image

>A few observations: PHP 7.3 is growing almost as well as 7.2 last year. With PHP 5.6 and 7.0 reaching end of life at the end of 2018, the number of people using maintained PHP versions dropped to 80% (from 84%), which is really not as bad as I expected as 31% of people had to upgrade. PHP 5 usage across all versions is now around 10%.

https://w3techs.com/technologies/details/pl-php/all/all

Screen Shot 2019-08-25 at 3 25 09 am

https://w3techs.com/technologies/details/pl-php/5/all

Screen Shot 2019-08-25 at 3 34 27 am

https://wordpress.org/about/stats

Screen Shot 2019-08-25 at 3 29 03 am

---

- [x] Initial warning PR: https://github.com/backdrop/backdrop/pull/3881
- [ ] Actual increase of minimum version in bootstrap.inc: https://github.com/backdrop/backdrop/pull/3922

quicksketch:
Fixes https://github.com/backdrop/backdrop-issues/issues/3992

### DIFF
--- a/.github/misc/setup-php-fpm.sh
+++ b/.github/misc/setup-php-fpm.sh
@@ -7,19 +7,6 @@ if [ $# -lt 1 ]; then
   echo 'Fatal: Parameter missing'
   exit 1
 fi
-# This script handles one required numeric parameter - the PHP version.
-OLDPHP=$(echo "$1 < 7" | bc -l)
-
-# Apache mpm_event refuses to work with older PHP versions, we have to switch to
-# mpm_prefork for 5.x. The paths to config files differ for 5.x as set up by
-# shivammathur/setup-php@v2
-if [ $OLDPHP -eq 1 ]; then
-  sudo a2dismod mpm_event
-  sudo a2enmod mpm_prefork
-  CONFPATHS=$(ls /usr/local/php/*/etc/php-fpm.conf)
-else
-  CONFPATHS=$(ls /etc/php/*/fpm/pool.d/www.conf)
-fi
 
 # Configure php-fpm to run as user "runner". That makes moving files around
 # obsolete. Additionally tweak it for better performance, start and allow more
@@ -30,7 +17,7 @@ sudo sed -i -e 's/user = www-data/user = runner/' \
   -e 's/pm.start_servers = 2/pm.start_servers = 4/' \
   -e 's/pm.min_spare_servers = 1/pm.min_spare_servers = 2/' \
   -e 's/pm.max_spare_servers = 3/pm.max_spare_servers = 4/' \
-  $CONFPATHS
+  ls /etc/php/*/fpm/pool.d/www.conf
 
 # Let above changes take effect and setup Apache to work with php-fpm.
 sudo systemctl restart php${1}-fpm.service
--- a/core/includes/bootstrap.inc
+++ b/core/includes/bootstrap.inc
@@ -17,7 +17,7 @@ define('BACKDROP_CORE_COMPATIBILITY', '1.x');
 /**
  * Minimum supported version of PHP.
  */
-define('BACKDROP_MINIMUM_PHP', '5.3.2');
+define('BACKDROP_MINIMUM_PHP', '5.6.0');
 
 /**
  * Minimum recommended value of PHP memory_limit.
--- a/core/includes/bootstrap.inc
+++ b/core/includes/bootstrap.inc
@@ -272,7 +272,7 @@ if (!defined('DATE_RFC7231')) {
 
 // Exit early if running an incompatible PHP version to avoid fatal errors.
 if (version_compare(PHP_VERSION, BACKDROP_MINIMUM_PHP) < 0) {
-  print 'Your PHP installation is too old. Backdrop CMS requires at least PHP 5.3.2. See the <a href="https://backdropcms.org/requirements">System Requirements</a> page for more information.';
+  print 'Your PHP installation is too old. Backdrop CMS requires at least PHP ' . BACKDROP_MINIMUM_PHP . '. See the <a href="https://backdropcms.org/requirements">System Requirements</a> page for more information.';
   exit;
 }
 
@@ -2055,8 +2055,7 @@ function format_string($string, array $args = array()) {
  *
  * @return string
  *   An HTML safe version of $text. If $text is not valid UTF-8, an empty string
- *   is returned and, on PHP < 5.4, a warning may be issued depending on server
- *   configuration (see @link https://bugs.php.net/bug.php?id=47494 @endlink).
+ *   is returned.
  *
  * @see backdrop_validate_utf8()
  * @ingroup sanitization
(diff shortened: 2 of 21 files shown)


core/modules/file/file.module
```diff
@@ -2015,8 +2015,11 @@ function file_match_mimetypes($needle, $haystack) {
...
- * We include this, because Windows servers do not implement fnmatch() until
- * PHP Version 5.3. See: http://php.net/manual/en/function.fnmatch.php
+ * Windows servers did not implement fnmatch() until PHP Version 5.3. Now that
+ * Backdrop requires newer PHP versions, the fnmatch() function should be
+ * called directly.
+ *
+ * @deprecated Since 1.22.0
  */
 function file_fnmatch($pattern, $string) {
   if (empty($pattern) || empty($string)) {
```
jenlampton, line 2029: should we add a watchdog deprecated here?

core/includes/bootstrap.inc
```diff
@@ -272,7 +272,7 @@ if (!defined('DATE_RFC7231')) {
 
 // Exit early if running an incompatible PHP version to avoid fatal errors.
 if (version_compare(PHP_VERSION, BACKDROP_MINIMUM_PHP) < 0) {
-  print 'Your PHP installation is too old. Backdrop CMS requires at least PHP 5.3.2. See the <a href="https://backdropcms.org/requirements">System Requirements</a> page for more information.';
+  print 'Your PHP installation is too old. Backdrop CMS requires at least PHP ' . BACKDROP_MINIMUM_PHP . '. See the <a href="https://backdropcms.org/requirements">System Requirements</a> page for more information.';
```
klonos, line 275: Shouldn't we be using `t()`/`st()` here? ...in which case the concatenation of `BACKDROP_MINIMUM_PHP` in the string should go in a replacements array (as should the URL in `href`).

core/modules/simpletest/tests/image.test
```diff
@@ -404,18 +404,6 @@ class ImageToolkitGdTestCase extends BackdropWebTestCase {
...
         $correct_colors = TRUE;
 
-        // PHP 5.5 GD bug: https://bugs.php.net/bug.php?id=65148. PHP 5.5 GD
-        // rotates differently then it did in PHP 5.4 resulting in different
-        // dimensions. For the test images, the dimensions will be 1 pixel
-        // smaller in both dimensions (though other tests have shown a
-        // difference of 0 to 3 pixels in both dimensions.
-        // @todo: The PHP bug was fixed in PHP 7.0.26 and 7.1.12. Change the code
-        //   below to reflect that in https://www.drupal.org/node/2918570.
-        if (version_compare(PHP_VERSION, '5.5', '>=') && $values['function'] === 'rotate' && $values['arguments'][0] % 90 != 0) {
```
klonos, line 414: If I'm reading the comment correctly, then this bug exists from php5.5 all through 7.0.26 and 7.1.12. If that's the case, then we should leave this in place.

core/includes/bootstrap.inc
klonos: ...also noting that we're using the very same string in 3 places:
- bootstrap.inc
- update.php
- install.php

Perhaps worth consolidating that in a single place instead.

quicksketch: These places don't use `t()` because it's so early in the bootstrap that multilingual isn't available. I think it's this way specifically because if other .inc files were included that used unsupported language features, it would immediately crash PHP rather than give a "friendly" error message.

core/modules/simpletest/tests/image.test
quicksketch: I don't think the code comment is accurate. Our tests for PHP 7+ have always passed with this in place. It's possible that early versions of PHP 7.0.x and 7.1.x had the "old" calculation, but the latest releases of those versions pass with the current code. I think change in calculation happened mid-cycle, so while it's possible that this might be a problem on some installations of PHP 7.0 and 7.1, I doubt this problem will affect us enough to try and pin down the exact version numbers; hence just removing the code comments (current code doesn't pin it down either).

core/modules/file/file.module
quicksketch: Added.

core/modules/simpletest/tests/image.test
quicksketch: It turns out I get different results between my local PHP 5.6.40 and the 5.6.40 used by GHA. My local works the same as PHP 7+, but GHA has a different result. I restored this work-around but loosened it further, and the work-around only applies to older versions, so in the future we'll just be able to delete this section.

.github/misc/setup-php-fpm.sh
```diff
@@ -7,19 +7,6 @@ if [ $# -lt 1 ]; then
...
-# Apache mpm_event refuses to work with older PHP versions, we have to switch to
-# mpm_prefork for 5.x. The paths to config files differ for 5.x as set up by
-# shivammathur/setup-php@v2
-if [ $OLDPHP -eq 1 ]; then
-  sudo a2dismod mpm_event
-  sudo a2enmod mpm_prefork
-  CONFPATHS=$(ls /usr/local/php/*/etc/php-fpm.conf)
-else
-  CONFPATHS=$(ls /etc/php/*/fpm/pool.d/www.conf)
-fi
```
quicksketch, line 22: @indigoxela I ended up just removing all of this, as PHP 5.6+ all seems to work with mpm_event and the PHP configuration is now always in `/etc/php`

core/modules/file/file.module
```diff
@@ -2015,10 +2018,14 @@ function file_match_mimetypes($needle, $haystack) {
 /**
  * A wrapper function for the PHP function fnmatch().
  *
- * We include this, because Windows servers do not implement fnmatch() until
- * PHP Version 5.3. See: http://php.net/manual/en/function.fnmatch.php
+ * Windows servers did not implement fnmatch() until PHP Version 5.3. Now that
```
```suggestion
 * Windows servers did not implement fnmatch() until PHP version 5.3. Now that
```

core/includes/file.inc
```diff
@@ -872,9 +872,6 @@ function file_valid_uri($uri) {
  * - If the $source and $destination are equal, the behavior depends on the
  *   $replace parameter. FILE_EXISTS_REPLACE will error out. FILE_EXISTS_RENAME
  *   will rename the file until the $destination is unique.
- * - Works around a PHP 5.3 bug where copy() does not properly support streams
- *   if safe_mode or open_basedir are enabled.
- *   @see https://bugs.php.net/bug.php?id=60456
```
indigoxela, line 877: How does it workaround that? Is it the `@copy()`? Could that mean that the error control operator is obsolete now?

core/includes/file.inc
```diff
@@ -1139,9 +1136,6 @@ function file_move(File $source, $destination = NULL, $replace = FILE_EXISTS_REN
  *   is reported.
  * - If file already exists in $destination either the call will error out,
  *   replace the file or rename the file based on the $replace parameter.
- * - Works around a PHP 5.3 bug where rename() does not properly support streams
- *   if safe_mode or open_basedir are enabled.
- *   @see https://bugs.php.net/bug.php?id=60456
```
indigoxela, line 1144: Similar question to above: what is that workaround? Is it the `@rename()`? Is it obsolete?

core/modules/simpletest/tests/image.test
indigoxela: Should we add an `@todo` comment to remove it once B requires 7.1+? With exact versions when that odd bug was fixed in php?

> My local works the same as PHP 7+, but GHA has a different result.

Interesting, the action uses php 5.6 from the Sury repo (AFAIK). Where does your local version come from?

klonos: > Should we add an `@todo` comment to remove it once B requires 7.1+?

I think that at the very least a `@todo` 👍🏼  ...however, since the minimum version is specified via `BACKDROP_MINIMUM_PHP` and we might miss the `@todo` when we update it, I'd suggest to add a conditional check here for `version_compare(BACKDROP_MINIMUM_PHP, '7.1.12', '>=')` and throw a `watchdog` message. That is more likely to be caught and/or reported by people, and that in turn will remind us to actually do the `@todo` 😉

ghost:
Thanks! Merged into 1.x.

PR #3925, opened 2022-01-28
Issue #3650: Update documentation re. adding CSS/JS files to all pages.

### LINKED ISSUE #3650: API documentation inconsistency re adding JS to every page
The [documentation](https://api.backdropcms.org/api/backdrop/core%21includes%21common.inc/function/backdrop_add_js/1) for `backdrop_add_js()` says:

> It is critical that all preprocessed files are added unconditionally on every page, even if the files are not needed on a page. This is normally done by calling backdrop_add_js() in a hook_init() implementation.

However, the [documentation](https://api.backdropcms.org/api/backdrop/core%21modules%21system%21system.api.php/function/hook_init/1) for `hook_init()` says:

> To add CSS or JS files that should be present on all pages, modules should not implement this hook, but declare these files in their .info file.

I'm guessing the latter is the correct one...

ghost:
Fixes https://github.com/backdrop/backdrop-issues/issues/3650

### DIFF
--- a/core/includes/common.inc
+++ b/core/includes/common.inc
@@ -3400,7 +3400,7 @@ function backdrop_css_defaults($data = NULL) {
  * all typical visitors and most pages of a site. It is critical that all
  * preprocessed files are added unconditionally on every page, even if the
  * files do not happen to be needed on a page. However, it is preferred that
- * modules do not implement this hook, but declare CSS files intended for all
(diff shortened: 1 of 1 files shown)


klonos (approved):
LGTM 👍🏼

quicksketch:
Thanks @BWPanda! Merged into 1.x and 1.21.x.

PR #3928, opened 2022-01-29
Issue #4013: Allow text formats to be deleted.

### LINKED ISSUE #4013: Allow fully deleting text formats
This is a follow-up to https://github.com/backdrop/backdrop-issues/issues/1575. That issue allowed disabled text formats to be re-enabled via the admin UI.

This issue is to discuss allowing text formats to be fully deleted/removed from a site (their config files deleted). Concerns revolve around whether this should be allowed at all, and if so, what should happen to content that uses a deleted format.

ghost:
Fixes https://github.com/backdrop/backdrop-issues/issues/4013


quicksketch:
Closing/reopening to get a new sandbox.

PR #3929, opened 2022-01-29
Issue #5479: Override Entity::createAccess() for File entities

### LINKED ISSUE #5479: Implement createAccess for File entities
All core entities except File implement a createAccess() override. As the default return value of Entity::access is TRUE, this leads to the problem that the value for File is always TRUE - ignoring "create files" permission.

Our assumption is currently, that not creating that override when adding entity_access in #3011 was an oversight.

I'm not sure if this is a bug or a feature request. Feedback is welcome.

Also related: #4975 - TRUE is a dangerous default value, anyway.

indigoxela:
Fixes https://github.com/backdrop/backdrop-issues/issues/5479

### DIFF
--- a/core/modules/file/file.entity.inc
+++ b/core/modules/file/file.entity.inc
@@ -142,6 +142,18 @@ class File extends Entity {
   }
 
   /**
+   * Overrides Entity::createAccess().
+   */
+  public static function createAccess($bundle = NULL, $account = NULL) {
+    // If no user object is supplied, the access check is for the current user.
+    if (empty($account)) {
+      $account = $GLOBALS['user'];
+    }
+
+    return user_access('create files', $account);
+  }
+
+  /**
    * Overrides Entity::access().
    *
    * The "view" operation is not supported on the File Entity itself because it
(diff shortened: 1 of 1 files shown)


core/modules/file/file.entity.inc
```diff
@@ -141,6 +141,18 @@ class File extends Entity {
     );
   }
 
+  /**
+   * Overrides Entity::createAccess().
+   */
+  public static function createAccess($bundle = NULL, $account = NULL) {
+    // If no user object is supplied, the access check is for the current user.
+    if (empty($account)) {
+      $account = $GLOBALS['user'];
```
klonos, line 150: @indigoxela is this check here really needed? `$account` is only ever used in `user_access()` below, and `user_access()` already checks for this.

indigoxela: I tried to stay consistent with the other createAccess implementations. They all do it that way (node, user, taxonomy, comment).

indigoxela:
Closing in favor of https://github.com/backdrop/backdrop/pull/3954

PR #3931, opened 2022-01-30
Issue #4014: Fix issues with File Type configuration form.

### LINKED ISSUE #4014: [UX] File Type configuration form changes
From https://github.com/backdrop/backdrop-issues/issues/2632#issuecomment-527664811:
> I do this all the time, I think that's bad advice. We should remove that text... where did you find it?

@jenlampton It comes from the description text on the 'Media types' field at `/admin/structure/file-types/manage/[TYPE]`: "You should not use the same media types in more than one file type."

image

ghost:
Fixes https://github.com/backdrop/backdrop-issues/issues/4014

### DIFF
--- a/core/modules/file/file.admin.inc
+++ b/core/modules/file/file.admin.inc
@@ -196,7 +196,7 @@ function file_type_form($form, &$form_state, $type = NULL) {
   );
 
   $mimetypes_description = t('The media (MIME) type(s) to be associated with this file type. When uploading a file, Backdrop automatically identifies the file type based on its media type. If multiple file types have the same media type, you will be prompted to manually select the file type.');
-  $mimetypes_description .= ' ' . l(t('More information about media types.'), 'https://en.wikipedia.org/wiki/Media_type');
+  $mimetypes_description .= ' ' . l(t('More information about media types'), 'https://en.wikipedia.org/wiki/Media_type') . '.';
   $mimetypes_description .= theme('item_list', array('items' => array(
     t('Enter one media type per line'),
     t('Asterisks can be used as wildcards (e.g.: !example)', array('!example' => '<code>image/*</code>')),
--- a/core/modules/file/file.admin.inc
+++ b/core/modules/file/file.admin.inc
@@ -173,6 +173,7 @@ function file_type_form($form, &$form_state, $type = NULL) {
     '#description' => t('This is the human readable name of the file type.'),
     '#required' => TRUE,
     '#default_value' => $type->name,
+    '#size' => 30,
   );
 
   $form['type'] = array(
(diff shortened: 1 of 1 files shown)


core/modules/file/file.admin.inc
```diff
@@ -197,7 +198,7 @@ function file_type_form($form, &$form_state, $type = NULL) {
   $form['mimetypes'] = array(
     '#type' => 'textarea',
     '#title' => t('Media types'),
-    '#description' => t("This is the list of media types (MIME types) which will be aassociated with this file type. When files with these media types are uploaded, Backdrop will identify the file type based on its media type. You should not use the same media types in more than one file type. Enter one media type per line. <br/><a href='@media_types_info'>More information about media types</a>.", array('@media_types_info' => 'https://en.wikipedia.org/wiki/Media_type')),
+    '#description' => t("The media (MIME) type(s) to be associated with this file type. When uploading a file, Backdrop automatically identifies the file type based on its media type. If multiple file types have the same media type, you will be prompted to manually select the file type. <a href='@media_types_info'>More information about media types</a>.<ul><li>Enter one media type per line</li><li>Asterisks can be used as wildcards (e.g. 'image/*')</li></ul>", array('@media_types_info' => 'https://en.wikipedia.org/wiki/Media_type')),
```
klonos, line 201: This looks great @BWPanda, thanks 🙏🏼 

Just a couple of things:
1. Lets switch the outer double-quotes in `t()` into single quotes, and the single quotes in `href=` into double ones. I am told that parsing is faster with single quotes, since php does not try to find any variables that need to be escaped that way.
2. Lets make life easier for translators by avoiding HTML markup in `t()`:
   - we can use `theme_item_list()` for the bullet list.
   - we can move the "More information about media types" link out of the main description and concatenate it with an `l()` instead.

core/modules/file/file.admin.inc
```diff
@@ -197,7 +198,7 @@ function file_type_form($form, &$form_state, $type = NULL) {
...
     '#default_value' => implode("\n", $type->mimetypes),
   );
```
klonos: The result looks great @BWPanda, thanks 🙏🏼 

Just a couple of things re implementation:
1. Lets switch the outer double-quotes in `t()` into single quotes, and the single quotes in `href=` into double ones. I am told that parsing is faster with single quotes, since php does not try to find any variables that need to be escaped that way.
2. Lets make life easier for translators by avoiding HTML markup in `t()`:
   - we can use `theme_item_list()` for the bullet list.
   - we can move the "More information about media types" link out of the main description and concatenate it with an `l()` instead (since it is a whole independent sentense, which is a link in its entirety).
   - as much as we should avoid HTML markup, if we add either double or single quotes around things that people can copy/paste or use as examples, there's always the risk of them also copy/pasting the quotes, when we only mean to have them use the value within them. To avoid that successfully, we can wrap `image/*` in a `<code>` tag.

So this would work better I think:

```php
  $description = t('The media (MIME) type(s) to be associated with this file type. When uploading a file, Backdrop automatically identifies the file type based on its media type. If multiple file types have the same media type, you will be prompted to manually select the file type.');
  $description .= ' ' . l(t('More information about media types'), 'https://en.wikipedia.org/wiki/Media_type', array('attributes' => array('target' => '_blank'))) . '.';
  $help = array();
  $help[] = t('Enter one media type per line');
  $help[] = t('Asterisks can be used as wildcards (e.g.: <code>image/*</code>)');
  $description .= theme('item_list', array('items' => $help));
  $form['mimetypes'] = array(
    '#type' => 'textarea',
    '#title' => t('Media types'),
    '#description' => $description,
    '#default_value' => implode("\n", $type->mimetypes),
  );
```

The result is the same, but better for translators:

image

ghost: Thanks @klonos. I implemented your suggestions, with a few changes:
- Added '.' to translated link as I suspect better for translators to have a full sentence to work with
- Simplified PHP code for adding list
- Moved code example outside of translation as I suspect this should never be translated

core/modules/file/file.admin.inc
```diff
@@ -194,10 +195,16 @@ function file_type_form($form, &$form_state, $type = NULL) {
     '#default_value' => $type->description,
   );
 
+  $mimetypes_description = t('The media (MIME) type(s) to be associated with this file type. When uploading a file, Backdrop automatically identifies the file type based on its media type. If multiple file types have the same media type, you will be prompted to manually select the file type.');
+  $mimetypes_description .= ' ' . l(t('More information about media types.'), 'https://en.wikipedia.org/wiki/Media_type');
```
quicksketch:
```suggestion
  $mimetypes_description .= ' ' . l(t('More information about media types'), 'https://en.wikipedia.org/wiki/Media_type') . '.';
```

quicksketch:

PR #3934, opened 2022-02-01
Issue #5483: Cache contexts

### LINKED ISSUE #5483: Fixed: The layout context system is called way too many times.
## Description of the bug

On a vanilla Backdrop, a debug/dpm in `Layout::getContexts()` shows that function is called 14-16 times with each page load. The function is not cached. 

Every block added to the layout increases this by 2-3 times.

Furthermore the function `Layout::getContextFromRelationship()` is called 14-16 times for each relationship added to the layout. 

## Steps To Reproduce

Just add your favorite debug call (`dpm('whatever')` for example) to `Layout::getContexts()` and visit any page.
Or worse, create a layout with path `node/%`, then add a relationship to the layout, then add this dpm to `Layout::getContextFromRelationship()` instead. Add it twice. (This is probably also a bug that you can add it twice). I'm seeing a minimum of 30 calls per page load with two relationships (45 calls with 3 relationships, etc)

## Actual behavior

OOTB a node layout results in too many calls, especially to `Layout::getContextFromRelationship()`

## Expected behavior

This should either be refined, or if unavoidable, we should try and cache this.

## Additional information

1.22.0

docwilmot:
Fixes https://github.com/backdrop/backdrop-issues/issues/5483

### DIFF
--- a/core/modules/layout/includes/layout.class.inc
+++ b/core/modules/layout/includes/layout.class.inc
@@ -789,8 +789,8 @@ class Layout {
    */
   function getContexts($include_types = LayoutContext::USAGE_TYPE_ALL) {
     $contexts_cache = &backdrop_static(__FUNCTION__, array());
-    if (!empty($contexts_cache)) {
-      return $this->getContextsByType($contexts_cache, $include_types);
+    if (!empty($contexts_cache[$this->name])) {
+      return $this->getContextsByType($contexts_cache[$this->name], $include_types);
     }
 
     // Initialize a list of contexts if not defined.
@@ -819,7 +819,7 @@ class Layout {
     }
 
     // Load contexts from relationships.
-    if (empty($contexts_cache) || $include_types & LayoutContext::USAGE_TYPE_RELATIONSHIP) {
+    if (empty($contexts_cache[$this->name]) || $include_types & LayoutContext::USAGE_TYPE_RELATIONSHIP) {
       $this->contexts += $this->getContextsFromRelationships();
     }
 
@@ -836,7 +836,7 @@ class Layout {
       ));
     }
 
-    $contexts_cache = $this->contexts;
+    $contexts_cache[$this->name] = $this->contexts;
 
     return $this->getContextsByType($this->contexts, $include_types);
   }
--- a/core/modules/layout/includes/layout.class.inc
+++ b/core/modules/layout/includes/layout.class.inc
@@ -789,7 +789,7 @@ class Layout {
    */
   function getContexts($include_types = LayoutContext::USAGE_TYPE_ALL) {
     $contexts_cache = &backdrop_static(__FUNCTION__, array());
-    if (!empty($contexts_cache[$this->name])) {
+    if (isset($contexts_cache[$this->name])) {
       return $this->getContextsByType($contexts_cache[$this->name], $include_types);
     }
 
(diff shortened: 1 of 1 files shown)


core/modules/layout/includes/layout.class.inc
```diff
@@ -831,14 +836,23 @@ class Layout {
...
 
+    $contexts_cache[$this->name] = $this->contexts;
+
+    return $this->getContextsByType($this->contexts, $include_types);
+  }
+
+  /**
+   * Get context by requested type.
+   */
+  function getContextsByType($contexts, $include_types = LayoutContext::USAGE_TYPE_ALL) {
```
indigoxela: Some coding standard improvement could be:

- declare visibility on the new function
- document the parameters and return

core/modules/layout/includes/layout.class.inc
```diff
@@ -788,6 +788,11 @@ class Layout {
    * @return LayoutContext[]
    */
   function getContexts($include_types = LayoutContext::USAGE_TYPE_ALL) {
+    $contexts_cache = &backdrop_static(__FUNCTION__, array());
+    if (!empty($contexts_cache[$this->name])) {
```
indigoxela: I'm no expert on backdrop_static, but wouldn't work the default value NULL und isset() instead of empty() better?

Can the array rightly be empty at some point? Not sure about that, either. :wink:

core/modules/layout/includes/layout.class.inc
docwilmot: > document the parameters and return

This is true, will do
> declare visibility on the new function

I dont understand this one though.

core/modules/layout/includes/layout.class.inc
docwilmot: `$this->contexts` could be not set or it could be array at that point, but in either of those two cases, we would want to proceed to build the context array, so I think `empty()` is better here.

core/modules/layout/includes/layout.class.inc
indigoxela: Re visibility: it could be protected or public in the Layout class, whatever fits best. If not declared, it defaults to "public", but it's usually best practice to actually declare it.

core/modules/layout/includes/layout.class.inc
argiepiano: Question: since this method is called many times, should we consider using the "advanced backdrop_static() pattern", rather than calling backdrop_static() on every call?

docwilmot: > Can the array rightly be empty at some point? Not sure about that, either

Came to answer @argiepiano's question and noticed I didnt quite pay attention to this point you made. You're quite right, there isn't any chance that `$contexts_cache[$this->name]` could be NULL, since its always set at least to `array()`, so `isset()` should be fine there. Is there though a performance advantage of just `isset()` vs `empty()` though? Curious.

> Question: since this method is called many times, should we consider using the "advanced backdrop_static() pattern", rather than calling backdrop_static() on every call?

I did consider, but the docs mention if the function is called "100s of times" which this is not. I dont quite know the disadvantages of the advanced pattern, so cant be sure if there is any harm in using it here.

core/modules/layout/includes/layout.class.inc
docwilmot: Ahh, visibility, I get it now. Hadnt realised this was what this was called. Will add. I think protected, since it wouldnt really make sense to be accessible publicly.

klonos: Yeah, I didn't know what it was called either - I always referred to it as "scope" 😅

docwilmot: Done

docwilmot:
Could this get another review please?

quicksketch:
Looks great! @bugfolder gave it another review and I checked it just now.

PR #3937, opened 2022-02-02
Issue #5482: Invalid relationship context data

### LINKED ISSUE #5482: Fixed: Incorrect context data passed by `LayoutRelationshipAuthorFromNode`.
## Description of the bug

## Steps To Reproduce

```
/**
 */
  $blocks = array();
    'info' => t('test'),
    'description' => t('test'),
    'required contexts' => array(
      'user' => 'user',
    ),
  );

  return $blocks;
}

/**
 */
  $block = array();
  dpm($contexts);
  switch ($delta) {
    case 'test_block':
      $block['content'] = 'test';
      break;
  }
  return $block;
}
```
Then navigate to any node.

## Actual behavior

## Expected behavior

## Additional information

1.22.0

docwilmot:

### DIFF
--- a/core/modules/layout/includes/layout.class.inc
+++ b/core/modules/layout/includes/layout.class.inc
@@ -940,7 +940,10 @@ class Layout {
             if ($plugin_array[1] != 'child') {
               $relationship_data->childDelta = $plugin_array[1];
             }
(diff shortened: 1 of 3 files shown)


docwilmot:
Replaced by https://github.com/backdrop/backdrop/pull/4046

PR #3940, opened 2022-02-03
Issue #3992: GHA file system permission problem

indigoxela:
Now patches the right fpm conf file (in /etc, not /usr/local).

### DIFF
--- a/.github/misc/setup-php-fpm.sh
+++ b/.github/misc/setup-php-fpm.sh
@@ -20,7 +20,13 @@ fi
 
 # Ini files for php 5.6+ are in /etc/php/, for php versions prior to 5.6 files
 # are in /usr/local/php/.
-CONFPATHS=$(ls /etc/php/*/fpm/pool.d/www.conf)
+INIPATH_ETC=$(echo "$1 > 5.3" | bc -l)
+
+if [ $INIPATH_ETC -eq 1 ]; then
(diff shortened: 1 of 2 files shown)


.github/misc/setup-php-fpm.sh
```diff
@@ -16,9 +16,16 @@ OLDPHP=$(echo "$1 < 7" | bc -l)
 if [ $OLDPHP -eq 1 ]; then
   sudo a2dismod mpm_event
   sudo a2enmod mpm_prefork
-  CONFPATHS=$(ls /usr/local/php/*/etc/php-fpm.conf)
-else
+fi
+
+# Ini files for php 5.6+ are in /etc/php/, for php versions prior to 5.6 files
+# are in /usr/local/php/.
+INIPATH_ETC=$(echo "$1 > 5.3" | bc -l)
```
indigoxela, line 23: Reminder to myself: this should be `5.5` instead of 5.3.

quicksketch:
I incorporated this PR into https://github.com/backdrop/backdrop/pull/3922.

I think we don't need to do any version checking any more. It seems that PHP 5.6 not only uses the `/etc/php/` location for PHP-FPM configuration, but it also works with MPM Event in Apache (no need to switch to MPM Prefork anymore).

Thanks this was a HUGE help! I had no idea why PHP 5.6 wasn't running.

indigoxela:
All done here, time to close. :wink:

PR #3941, opened 2022-02-03
Issue 5489: Project Installer: Clicable module labels

### LINKED ISSUE #5489: [UX] Project installer: tick/untick checkboxes when the project name is selected.
When you want to tick/untick boxes in the modules listing page at `/admin/modules`, you can either click the checkbox itself, or click on the respective module name. The same happens when selecting modules to uninstall at `/admin/modules/uninstall` 👍🏼 

When installing new modules via the project installer, during the "Enable modules" step, the module name does not trigger the ticking/unticking of the respective checkbox though. We should fix that for consistency in UX.

klonos:
Fixes https://github.com/backdrop/backdrop-issues/issues/5489


core/modules/installer/installer.pages.inc
```diff
@@ -701,17 +701,27 @@ function installer_browser_installation_enable_form($form, &$form_state, $projec
...
         if (empty($modules[$project['name']]->info['hidden'])) {
+          $module_label = $modules[$project['name']]->info['name'];
+          $module_version = $modules[$project['name']]->info['version'];
           if ($dependency_check) {
+            $module_label = array(
+              '#type' => 'head_tag',
+              '#tag' => 'label',
+              '#value' => $module_label,
+              '#attributes' => array(
+                'for' => 'edit-modules-' . backdrop_html_class($modules[$project['name']]->name),
```
jenlampton: we should probably put 'edit-modules-' inside `backdrop_html_class()`. If we ever change how classes are constructed, we'd want the whole class name to be changed, not just part of it. 
```php
backdrop_html_class('edit-modules-' . $modules[$project['name']]->name)
```

core/modules/installer/installer.pages.inc
```diff
@@ -735,7 +745,7 @@ function installer_browser_installation_enable_form($form, &$form_state, $projec
 
   if (!empty($all_options)) {
     $form['module_instructions'] = array(
-      '#type' => 'item',
+      '#type' => 'help',
```
jenlampton, line 743: ❤️

PR #3943, opened 2022-02-04
Issue #5491: Check for wrong file paths in .tests.info file

### LINKED ISSUE #5491: Simpletest semi-silently ignores wrong file paths in .tests.info file
As noted in https://github.com/backdrop/backdrop-issues/issues/5490, 

> it turned out that several tests actually never run:
> core/modules/system/tests/system.tests.info contains wrong file paths
> It's still not clear, why these tests ran in the past, but don't for several weeks. Eventually related: Issue https://github.com/backdrop/backdrop-issues/issues/5415: Gracefully handle empty or malformed .tests.info files.

In https://github.com/backdrop/backdrop-issues/issues/5415, we introduced code that checked for malformed .tests.info files and put errors into watchdog when they're detected. We can (should) add to this error-checking to verify that the file paths specified in the tests.info file are correct.

PR to follow.

bugfolder:
Fixes https://github.com/backdrop/backdrop-issues/issues/5491.

### DIFF
--- a/core/modules/simpletest/simpletest.module
+++ b/core/modules/simpletest/simpletest.module
@@ -372,11 +372,11 @@ function simpletest_test_get_all() {
           $error = 'File %uri contains no testing classes.';
         }
         else {
+          // If the .tests.info file is malformed, it puts a slew of
+          // hard-to-diagnose errors into the watchdog log when the testing page
+          // is loaded. So we'll check for various malformalities and put a more
+          // informative error.
           foreach ($classes as $class => $info) {
-            // If the .tests.info file is malformed, it puts a slew of
-            // hard-to-diagnose errors into the watchdog log when the testing page
-            // is loaded. So we'll check for various malformalities and put a more
-            // informative error.
             if (!is_array($info) || empty($info)) {
               $error = format_string('File %uri contains a definition for %class that is not an array of testing class fields.', array('%class' => $class));
               continue;
@@ -394,9 +394,18 @@ function simpletest_test_get_all() {
                 continue;
               }
             }
+            // Verify that the test file is present in the same directory as
+            // specified in the .tests.info file.
+            if (empty($error)) {
+              $info['file path'] = dirname($file->uri);
+              $test_file_full = BACKDROP_ROOT . '/' . $info['file path'] . '/' . $info['file'];
+              if (!file_exists($test_file_full)) {
+                $error = format_string('File %uri specifies a non-existent file %test_file_full. Please verify that the test file exists in the location specified in %uri.', array('%test_file_full' => $test_file_full));
+                continue;
+              }
+            }
             if (empty($error)) {
             // If the format passed, build the rest of the groups information.
-              $info['file path'] = dirname($file->uri);
               // If this test class requires a non-existing module, skip it.
               if (!empty($info['dependencies'])) {
                 foreach ($info['dependencies'] as $module) {
--- a/core/modules/simpletest/simpletest.module
+++ b/core/modules/simpletest/simpletest.module
@@ -369,7 +369,7 @@ function simpletest_test_get_all() {
         $classes = backdrop_parse_info_file($file->uri, TRUE);
         $error = '';
         if (empty($classes)) {
-          $error = 'File %uri contains no testing classes.';
+          $error = 'File @uri contains no testing classes.';
         }
         else {
           // If the .tests.info file is malformed, it puts a slew of
@@ -378,7 +378,7 @@ function simpletest_test_get_all() {
           // informative error.
           foreach ($classes as $class => $info) {
             if (!is_array($info) || empty($info)) {
-              $error = t('File %uri contains a definition for %class that is not an array of testing class fields.', array('%class' => $class));
+              $error = t('File @uri contains a definition for @class that is not an array of testing class fields.', array('@class' => $class));
               continue;
             }
             else {
@@ -390,7 +390,7 @@ function simpletest_test_get_all() {
               }
               if (!empty($missing_keys)) {
                 $keys = implode(', ', $missing_keys);
-                $error = format_plural(count($missing_keys), 'File %uri contains a definition for %class that is missing field %keys.','File %uri contains a definition for %class that is missing fields %keys.', array('%class' => $class, '%keys' => $keys));
+                $error = format_plural(count($missing_keys), 'File @uri contains a definition for @class that is missing field @keys.','File @uri contains a definition for @class that is missing fields @keys.', array('@class' => $class, '@keys' => $keys));
                 continue;
               }
             }
@@ -400,7 +400,7 @@ function simpletest_test_get_all() {
               $info['file path'] = dirname($file->uri);
               $test_file_full = BACKDROP_ROOT . '/' . $info['file path'] . '/' . $info['file'];
               if (!file_exists($test_file_full)) {
-                $error = t('File %uri specifies a non-existent file %test_file_full. Please verify that the test file exists in the location specified in %uri.', array('%test_file_full' => $test_file_full));
+                $error = t('File @uri specifies a non-existent file @test_file_full. Please verify that the test file exists in the location specified in @uri.', array('@test_file_full' => $test_file_full));
                 continue;
               }
             }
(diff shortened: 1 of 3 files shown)


core/modules/simpletest/simpletest.module
```diff
@@ -394,9 +394,18 @@ function simpletest_test_get_all() {
...
               }
             }
+            // Verify that the test file is present in the same directory as
+            // specified in the .tests.info file.
             if (empty($error)) {
-            // If the format passed, build the rest of the groups information.
               $info['file path'] = dirname($file->uri);
+              $test_file_full = BACKDROP_ROOT . '/' . $info['file path'] . '/' . $info['file'];
+              if (!file_exists($test_file_full)) {
+                $error = t('File %uri specifies a non-existent file %test_file_full. Please verify that the test file exists in the location specified in %uri.', array('%test_file_full' => $test_file_full));
```
indigoxela: I just realized, that this happens in a loop (`foreach ($classes as $class => $info)`). Doesn't that mean that if several paths are wrong, only the last file gets mentioned?

core/modules/simpletest/simpletest.module
```diff
@@ -372,13 +372,13 @@ function simpletest_test_get_all() {
...
+          // is loaded. So we'll check for various malformalities and put a more
+          // informative error.
           foreach ($classes as $class => $info) {
-            // If the .tests.info file is malformed, it puts a slew of
-            // hard-to-diagnose errors into the watchdog log when the testing page
-            // is loaded. So we'll check for various malformalities and put a more
-            // informative error.
             if (!is_array($info) || empty($info)) {
-              $error = format_string('File %uri contains a definition for %class that is not an array of testing class fields.', array('%class' => $class));
+              $error = t('File %uri contains a definition for %class that is not an array of testing class fields.', array('%class' => $class));
```
indigoxela: I'm curious: why switching from format_string to t()?

core/modules/simpletest/simpletest.module
bugfolder: In GHA tests, only the first .tests.info file will be mentioned. In the UI, each bad .tests.info file will be mentioned. However, only the first (not last) bad file path in a .tests.info file will be mentioned. Hopefully, this alerts the user to go examine their .tests.info file if they've done the same error in multiple places within it.

The reason for bailing out when an error is found, rather than, say, accumulating a list of errors, is that for some types of malformation, once the parser gets confused, it generates a whole slew of gibberish errors, so rather than stuffing watchdog (and the messages area) with all of that where only the first error in the .tests.info file was really relevant, we just report the first to let the person know they need to go take a look.

core/modules/simpletest/simpletest.module
bugfolder: I realized these error messages show up in the UI if one is running Simpletest from the UI, so seemed like they should have been translated all along.

core/modules/simpletest/simpletest.module
indigoxela: Makes sense, many thanks for the explanation. Consider this question as resolved.

core/modules/simpletest/simpletest.module
```diff
@@ -369,16 +369,16 @@ function simpletest_test_get_all() {
         $classes = backdrop_parse_info_file($file->uri, TRUE);
         $error = '';
         if (empty($classes)) {
-          $error = 'File %uri contains no testing classes.';
+          $error = t('File @uri contains no testing classes.');
```
indigoxela, line 372: Is that `@uri` supposed to get replaced by `$file->uri`? I'm asking because that replacement doesn't happen. I'm aware it also didn't happen before, but this might be the right time to fix it.

core/modules/simpletest/simpletest.module
```diff
@@ -369,16 +369,16 @@ function simpletest_test_get_all() {
...
+          // is loaded. So we'll check for various malformalities and put a more
+          // informative error.
           foreach ($classes as $class => $info) {
-            // If the .tests.info file is malformed, it puts a slew of
-            // hard-to-diagnose errors into the watchdog log when the testing page
-            // is loaded. So we'll check for various malformalities and put a more
-            // informative error.
             if (!is_array($info) || empty($info)) {
-              $error = format_string('File %uri contains a definition for %class that is not an array of testing class fields.', array('%class' => $class));
+              $error = t('File @uri contains a definition for @class that is not an array of testing class fields.', array('@class' => $class));
```
indigoxela, line 381: Same question as above - should that `@uri` get replaced?

core/modules/simpletest/simpletest.module
```diff
@@ -390,13 +390,22 @@ function simpletest_test_get_all() {
               }
               if (!empty($missing_keys)) {
                 $keys = implode(', ', $missing_keys);
-                $error = format_plural(count($missing_keys), 'File %uri contains a definition for %class that is missing field %keys.','File %uri contains a definition for %class that is missing fields %keys.', array('%class' => $class, '%keys' => $keys));
+                $error = format_plural(count($missing_keys), 'File @uri contains a definition for @class that is missing field @keys.','File @uri contains a definition for @class that is missing fields @keys.', array('@class' => $class, '@keys' => $keys));
```
indigoxela, line 393: And yet another time ... `@uri`?

core/modules/simpletest/simpletest.module
```diff
@@ -390,13 +390,22 @@ function simpletest_test_get_all() {
...
               }
             }
+            // Verify that the test file is present in the same directory as
+            // specified in the .tests.info file.
             if (empty($error)) {
-            // If the format passed, build the rest of the groups information.
               $info['file path'] = dirname($file->uri);
+              $test_file_full = BACKDROP_ROOT . '/' . $info['file path'] . '/' . $info['file'];
+              if (!file_exists($test_file_full)) {
+                $error = t('File @uri specifies a non-existent file @test_file_full. Please verify that the test file exists in the location specified in @uri.', array('@test_file_full' => $test_file_full));
```
indigoxela, line 403: Again... `@uri`...

core/modules/simpletest/simpletest.module
argiepiano: Alll `@uri` (or `%uri`) are replaced later on by `format_string`. But there is a bigger issue here. The [D7 documentation for tests](https://www.drupal.org/docs/7/testing/simpletest-testing-tutorial-drupal-7#when-to-use-t) lists very few cases when `t()` is supposed to be used. This is not one of those. My suggestion is to stick with `format_string` here.  (Although the case may be made that this is not a test, but the module that runs them)

argiepiano: In any case, it's strange to have t() for some of these and format_string later on. So, if we are going to use t(), then I agree that it's best to replace uri  each time it's used.

argiepiano: OK, I just realized some of this had been discussed in the main issue (and I had posted the same link 🙃 !) Feel free to disregard.

core/modules/simpletest/simpletest.module
bugfolder: > I'm asking because that replacement doesn't happen.

Oh yes it does. Try it out by creating a malformed test file, or just note this code at lines 429–430:

```
            backdrop_set_message(format_string($error, array('@uri' => $file->uri)), 'error');
            watchdog('simpletest', $error, array('@uri' => $file->uri), WATCHDOG_WARNING, l(t('See the Testing Framework documentation.'), 'https://docs.backdropcms.org/documentation/testing-framework'));
```

So, you might ask: what's going on here? We have several possible error strings, so they each get translated individually as we define `$error` in several different places. But that string gets fed to `backdrop_set_message()` and `watchdog()`, which want to see already-translated strings. `backdrop_set_message()` wants variables substituted in (so we use `format_string()`, not `t()` because the string was already translated). `watchdog()` wants its variables as a separate argument, so that's what we provide.

core/modules/simpletest/simpletest.module
bugfolder: OK.

core/modules/simpletest/simpletest.module
bugfolder: See above.

core/modules/simpletest/simpletest.module
bugfolder: See above.

core/modules/simpletest/simpletest.module
indigoxela: I still find it confusing that different replacements happen in different places, but if that's unavoidable ... :shrug:

indigoxela:
@bugfolder please rebase the PR - it's massively behind core. (That should fix the phpcs failure.)

bugfolder:
Closing in favor of rebased https://github.com/backdrop/backdrop/pull/4249.

core/modules/simpletest/simpletest.module
bugfolder: I tried something different in the new PR with all replacements in one place. https://github.com/backdrop/backdrop/pull/4249

PR #3945, opened 2022-02-05
Issue #5478: backdrop_add_css() and backdrop_add_js() follow-up

### LINKED ISSUE #5478: Allow attributes to be passed to drupal_add_[css|js] (SRI)
This is the same as https://www.drupal.org/project/drupal/issues/1664602

> ## Problem/Motivation
> Currently there is no way to pass `#attributes` to the `drupal_add_js()` function. This is not a feature that is widely used, however, there are unique cases where the `<script>` tag generated needs to have available an `id=""` or other unique attributes to that tag.
> 
> ## Proposed resolution
> By initializing the `#attributes` key of the `$js_element` array we can easily pass specific `#attribute` elements in our `drupal_add_js()` calls, and since `theme('html_tag', array('element' => $js_element));` handles `#attributes` passed to it, all additional support is inherited. Here is an example of the new attribute code working in the case 'external' type of `drupal_add_js()` is invoked. The patch of course accounts for all cases of `'type'`:

```php
case 'external':
  $js_element = $element;

  // Attach custom attributes to the element.
  $js_element['#attributes'] = !empty($item['attributes']) ? $item['attributes'] : array();

  // Preprocessing for external JavaScript files is ignored.
  if ($item['defer']) {
    $js_element['#attributes']['defer'] = 'defer';
  }

  $js_element['#attributes']['src'] = $item['data'];
  $processed[$index++] = theme('html_tag', array('element' => $js_element));
  break;
```

klonos:
Fixes https://github.com/backdrop/backdrop-issues/issues/5478


core/modules/simpletest/tests/common.test
```diff
@@ -1618,7 +1621,7 @@ class CommonJavaScriptTestCase extends BackdropWebTestCase {
   /**
    * Tests adding JavaScript files with additional attributes.
    */
-  public function testAttributes() {
+  public function testRenderScriptsAttributes() {
```
klonos, line 1624: Renamed to clarify and be more self-explanatory, and to also match `testRenderStylesAttributes()`

PR #3948, opened 2022-02-08
Issue #5497: Make sure the auto column already exists

### LINKED ISSUE #5497: Possible D7 upgrade regression caused by user/1 path alias update hook
## Description of the bug

When upgrading from D7, following exceptions pop up:

```
PDOException: SQLSTATE[42S22]: Column not found: 1054 Unknown column 'auto' in 'field list': INSERT INTO {url_alias} (`pid`, `source`, `alias`, `langcode`, `auto`) VALUES (default, :db_insert_placeholder_0, :db_insert_placeholder_1, :db_insert_placeholder_2, :db_insert_placeholder_3); Array ( [:db_insert_placeholder_0] => user/1 [:db_insert_placeholder_1] => accounts/superadmin [:db_insert_placeholder_2] => und [:db_insert_placeholder_3] => 1 ) in backdrop_write_record() (line 8288 of /.../core/includes/common.inc).

EntityStorageException: SQLSTATE[42S22]: Column not found: 1054 Unknown column 'auto' in 'field list' in EntityDatabaseStorageController->save() (line 786 of ...core/modules/entity/entity.controller.inc).
```
Possibly introduced with Issue #1417: User URL alias not applied to admin user out of the box.

## Steps To Reproduce

Upgrade from a Drupal 7

## Actual behavior

Exception. Note that the alias for user/1 does get created (in a second update run).

## Expected behavior

No exception

## Additional information

Comparing the D7 and B1 database structure, table url_alias has a column auto, but the Drupal table has not.
That column gets added in `function system_update_1039()`.
If we let function user_update_1021() depend on the system update hook, that should be fine, because then the column exists.

See also: [hook_update_dependencies](https://docs.backdropcms.org/api/backdrop/core%21modules%21system%21system.api.php/function/hook_update_dependencies/1)

indigoxela:
Fixes https://github.com/backdrop/backdrop-issues/issues/5497

### DIFF
--- a/core/modules/user/user.install
+++ b/core/modules/user/user.install
@@ -198,6 +198,10 @@ function system_update_dependencies() {
   $dependencies['user'][1007] = array(
     'system' => 1025,
   );
+  // Ensure the 'auto' column already exists in url_alias table.
+  $dependencies['user'][1021] = array(
+    'system' => 1039,
+  );
   return $dependencies;
 }
 
--- a/core/modules/user/user.install
+++ b/core/modules/user/user.install
@@ -193,7 +193,7 @@ function user_install() {
 /**
  * Implements hook_update_dependencies().
  */
-function system_update_dependencies() {
+function user_update_dependencies() {
   // Ensure layouts have been converted before renaming our user roles.
   $dependencies['user'][1007] = array(
     'system' => 1025,


indigoxela:
A random buddy (ddd_installer_test)...

quicksketch:
Merged into 1.x and 1.23.x. Thanks @indigoxela!

PR #3950, opened 2022-02-11
Issue #5499: Enable required modules that are unique to Backdrop on migration

### LINKED ISSUE #5499: Add the ability to enable Backdrop-only required modules during an upgrade from Drupal 7.
There are modules (e.g., Entity Plus, Entity UI) that didn't exist in D7 and are required by Backdrop versions of D7 modules. Examples of B modules that need Entity Plus (E+, for short), for example, are Rules, Ubercart, and [the modules in this list](https://github.com/backdrop-contrib/rules/issues/138#issuecomment-1032112517).

A problem arises when upgrading a D7 site that uses those modules. If the user merely adds the E+ module to the B codebase (the logical thing to do), when trying to upgrade the site, B sees that there was no enabled version of the D7 module, so it assumes the B version is not enabled, and therefore disables it and all of the downstream modules (like Rules, Ubercart, and their submodules), and so doesn't run their update hooks, and the user ends up with a site with only partially upgraded and a bunch of disabled modules. (This is particularly problematic with Ubercart, which has about 20 submodules, plus further contrib sub-submodules.)

The current solution, which has been recommended for Rules and Ubercart for a while, has been _stub modules_, basically fake/empty D7 modules that you enable before performing the upgrade. See [here](https://github.com/backdrop-contrib/rules/wiki/Upgrading-Rules-from-Drupal-7) and [here](https://github.com/backdrop-contrib/ubercart/wiki/Upgrading-Ubercart-from-Drupal-7).

More recently, [it was suggested](https://github.com/backdrop-contrib/rules/issues/138) to enable the B modules in an update hook. We can't actually do this during `hook_update_N()`, because by the time that would get called, the damage is done: the downstream modules are already disabled, so their update hooks don't get called. However, it can be done during `hook_requirements('update')` (suggested example [here](https://github.com/backdrop-contrib/manualcrop/commit/033d6c4cf80186292fbf6089e1f6b84c15ce4fbd), PR that does this for Rules module [here](https://github.com/backdrop-contrib/rules/pull/172)). So that works, but it's not ideal, for a couple of reasons:

* If this is the preferred solution, (nearly) the same chunk of code would need to be implemented by each contrib module that requires E+, E-UI, etc.
* More importantly, if we enable the required module by calling `module_enable()` as part of `hook_requirements()` (as in the PR), this results in calling `hook_enable()` for the module, and `hook_modules_enabled()` for _all other modules_, and this happens before the whole series of `hook_update_N()` calls has happened; so these hooks, which might be expecting fully-configured B modules, will be fired before the B modules are upgraded from D7.
* Also, if multiple modules are doing this, there's no particular control over the order in which the modules are enabled. (If that matters.)

So, it would be nice if Backdrop core was able to provide support for this situation: **when an enabled D7 module is being upgraded to B and depends on a B module that didn't exist in D7**.

Some possibilities:

* Add a `module_enable_lite()` function that lets the client module "enable" the dependency B-only module (i.e., create the appropriate entry in the `system` table) without firing the enable hooks, for use by the contrib modules in their `hook_requirements()` call.
* Add a new updating hook that lets client modules declare dependency B-only modules that ought to be treated as enabled if they exist in the codebase, and then core handles the "lite" enablement of the B-only modules.

We could also just put E+ (which is currently the most popular contrib module) and E-UI into core, so that they're already enabled in new installs. (But this wouldn't be a total solution; B Ubercart still has a B-only module that isn't in D7.)

This issue arose from discussions and issues with @argiepiano and @laryn; please add anything I've missed, and everyone else, please give your thoughts. (H/T also to @indigoxela, who raised [the issue for Rules](https://github.com/backdrop-contrib/rules/issues/138).)

---

Advocate: @bugfolder
PR: https://github.com/backdrop/backdrop/pull/3950

argiepiano:
Fixes https://github.com/backdrop/backdrop-issues/issues/5499


core/includes/update.inc
```diff
@@ -312,6 +312,25 @@ function update_fix_requirements() {
       update_variable_del('drupal_private_key');
     }
 
+    // Get list of enabled modules.
+    $result = db_query("SELECT name FROM {system} WHERE type = :type AND status = :status", array(':type' => 'module', ':status' => 1));
+    // Load their .install files.
+    $enabled_modules = array_keys($result->fetchAllAssoc('name'));
```
klonos: That comment describes what is happening in the next line:

```suggestion
    $enabled_modules = array_keys($result->fetchAllAssoc('name'));
    // Load their .install files.
```

core/includes/update.inc
```diff
@@ -294,6 +294,46 @@ function update_module_add_to_system($modules = array()) {
...
+    // Remove duplicates.
+    $modules_to_enable = array_unique($modules_to_enable);
+    if (!empty($modules_to_enable)) {
+      module_enable($modules_to_enable);
+      $modules_string = implode(', ', $modules_to_enable);
+
+      state_set('update_migration_dependencies', TRUE);
+      $requirements['enable_modules_on_migrate'] = array(
+        'title' => t('Modules enabled automatically for migration.'),
+        'description' => t('The followig Backdrop modules needed to be enabled in order for the site to be migrated successully: %modules.', array('%modules' => $modules_string)),
```
klonos: Lets do something like this here instead (some typos + render the modules as a proper list):

```suggestion
        'description' => t('The following Backdrop modules needed to be enabled in order for the site to be migrated successfully: %modules.', array('%modules' => theme('item_list', array('items' => $modules_string)))),
```

klonos: ...would be even better if we had the human-readable names of the modules instead of machine names.

core/includes/update.inc
```diff
@@ -294,6 +294,46 @@ function update_module_add_to_system($modules = array()) {
...
+    $modules_to_enable = array();
+    foreach ($enabled_modules as $enabled_module) {
+      if (!empty($files[$enabled_module]->requires)) {
+        $required_modules = array_keys($files[$enabled_module]->requires);
+        // Add required modules that are not enabled.
+        $modules_to_enable = array_merge($modules_to_enable, array_diff($required_modules, $enabled_modules));
+      }
+    }
+    // Remove duplicates.
+    $modules_to_enable = array_unique($modules_to_enable);
```
klonos: Can we move this call inside the `if` that follows? No point in doing this if there are no modules to enable, right?

core/includes/update.inc
```diff
@@ -294,6 +294,46 @@ function update_module_add_to_system($modules = array()) {
...
+    if (!empty($modules_to_enable)) {
+      module_enable($modules_to_enable);
+      $modules_string = implode(', ', $modules_to_enable);
+
+      state_set('update_migration_dependencies', TRUE);
+      $requirements['enable_modules_on_migrate'] = array(
+        'title' => t('Modules enabled automatically for migration.'),
```
bugfolder:
```suggestion
        'title' => t('Modules enabled automatically'),
```

Shorter title (leaves more room for the description column), also titles aren't punctuated.

klonos: 👍🏼  ...plus, this is not a "migration" anyway - we call these "upgrades", right?

core/update.php
```diff
@@ -323,6 +323,9 @@ function update_info_page() {
   $output .= "<li>Install your new files into the appropriate location, as described in <a href=\"https://backdropcms.org/upgrade\">the handbook</a>.</li>\n";
   $output .= "</ol>\n";
   $output .= "<p>After performing the above steps proceed using the continue button.</p>\n";
+  if ($modules_to_enable = update_upgrade_check_dependencies()) {
```
bugfolder:
```suggestion
  $modules_to_enable = update_upgrade_check_dependencies();
  if (!empty($modules_to_enable)) {
```
Suggestion to improve code readability. Putting an assignment inside a conditional looks like it might be a typo (`=` vs `==`). Splitting the assignment and test is easier on the human (and no worse for the computer).

Also, since elsewhere `$modules_to_enable` is an array, but here the result of `update_upgrade_check_dependencies()` is a string, I suggest you change the name of the variable here (perhaps `$status_report` to match the variable name above?).

core/includes/update.inc
```diff
@@ -294,6 +294,88 @@ function update_module_add_to_system($modules = array()) {
...
+ *    to be displayed in the 'info' page of the update process. 
+ */
+
+function update_upgrade_check_dependencies() {
+  $status_report = '';
+  if (backdrop_get_installed_schema_version('system') > 7000) {
+    $modules_to_enable = update_upgrade_modules_to_enable();
+    
+    if (empty($modules_to_enable)) {
+      return;
```
bugfolder:
```suggestion
      return '';
```
Since this function should return a string.

core/update.php
argiepiano: Thanks @bugfolder. I will include this in the next PR in a minute.

core/includes/update.inc
argiepiano: Thanks @bugfolder. I will include this in the next PR in a minute.

core/includes/update.inc
```diff
@@ -294,6 +294,88 @@ function update_module_add_to_system($modules = array()) {
...
+/**
+ * Enables needed dependencies.
+ */
+function update_upgrade_enable_dependencies() {
+  if (backdrop_get_installed_schema_version('system') > 7000 && !state_get('update_upgrade_enable_dependencies', FALSE)) {
+    // When upgrading a D7 site we need to enable entity.module at this point
+    // since the calls to 'module_enable()' below need some of entity's functions.
+    // In a non-upgrade situation, entity.module will be enabled later,
+    // inside the hook implementation system_update_1000(). 
+    module_enable(array('entity'));
```
quicksketch: Seems like this use `update_module_enable()` like `system_update_1000()` does, which skips `hook_module_enable()` hooks.

core/includes/update.inc
```diff
@@ -294,6 +294,88 @@ function update_module_add_to_system($modules = array()) {
   }
 }
 
+/** 
+ * Checks for disabled dependencies during a Drupal 7 upgrade.
+ * 
+ * @return string
+ *    A themed message that lists modules that will be enabled,
+ *    to be displayed in the 'info' page of the update process.
```
quicksketch: Extra space in this docblock.
```suggestion
 * @return string
 *   A themed message that lists modules that will be enabled, to be displayed
 *   in the 'info' page of the update process. 
```

core/includes/update.inc
```diff
@@ -294,6 +294,88 @@ function update_module_add_to_system($modules = array()) {
...
+    }
+  }
+}
+
+/** 
+ * Find out which modules need to be enabled.
+ * 
+ * @return array
+ *    An array keyed by the machine names of module names needed to be enabled,
+ *    whose values are arrays of the modules that depend on them.
```
quicksketch:
```suggestion
 * @return array
 *   An array keyed by the machine names of module names needed to be enabled,
 *   whose values are arrays of the modules that depend on them.
```

core/includes/update.inc
```diff
@@ -294,6 +294,88 @@ function update_module_add_to_system($modules = array()) {
...
+    }
+
+    $files = system_rebuild_module_data();
+    $module_list = '';
+    foreach ($modules_to_enable as $module_to_enable => $dependents) {
+      $module_list .= '<strong>' . $files[$module_to_enable]->info['name'] . '. ' . t('Required by:') . '</strong>';
+      $module_list .= theme('item_list', array('items' => $dependents));
+    }
+    $status_report = '<h4>' . t('Some modules need to be enabled') . '</h4>';
+    $status_report .= t('The following Backdrop modules will be enabled in order for the site to be upgraded successfully:') . '</p>';
```
quicksketch: I don't believe using `t()` is completely safe here. At the point we're upgrading from Drupal 7 to Backdrop, I don't think the translation system will be working yet. In English `t()` has no effect, but in other languages this will probably cause an error. This is called from `update_info_page()` which, just has all strings in English. I think all these `t()` calls should simply be removed.

core/includes/update.inc
```diff
@@ -294,6 +294,88 @@ function update_module_add_to_system($modules = array()) {
...
+ *    An array keyed by the machine names of module names needed to be enabled,
+ *    whose values are arrays of the modules that depend on them.  
+ */
+function update_upgrade_modules_to_enable() {
+  // Get list of enabled modules. 
+  $result = db_query("SELECT name FROM {system} WHERE type = :type AND status = :status", array(':type' => 'module', ':status' => 1));
+    
+  // Get list of modules to be enabled.
+  $enabled_modules = array_keys($result->fetchAllAssoc('name'));
+  // Get information about all modules available
```
quicksketch:
```suggestion
  // Get information about all modules available.
```

core/includes/update.inc
bugfolder: We had some discussion of this issue at https://github.com/backdrop/backdrop-issues/issues/5499#issuecomment-1052769082 and following comments. Looked like using `update_module_enable()` was going to be more complicated. Would you review those comments and comment further?

argiepiano: The issue with using `update_module_enable` to enable `entity` at this point is that this function will actually not load the code of the `entity` module - it only changes the `system` table in the database. So you still will get an `Call to undefined function entity_info_cache_clear()` when this process tries to enable Backdrop modules required by the Standard installation such as Redirect and others (those modules are not enabled by the current update.php code  - see the discussion in the issue associated to this PR).

So, the only way to use `update_module_enable(array('entity');` is to manually load the code with `backdrop_load('module', 'entity');` right after. I've tested this and it seems to work fine (and it avoids firing the enable hooks). 

Still, I'm not sure if this is preferable. Any thoughts?

argiepiano: By the way, `module_enable('entity')` is needed here due to the new code below, which enables Backdrop's core modules required by Standard installation.  

Ironically, enabling `entity` here is  ONLY needed if your D7 site did not use Entity API. If your D7 used Entity API (that is, D7's `entity` contrib module), the D7 database will already have `entity` as enabled, and the Backdrop `entity` code will be loaded right away at the start of  the update.php process, making this line unnecessary in those cases.

core/includes/update.inc
argiepiano: Got rid of t() in later commit. Thanks!

core/includes/update.inc
```diff
@@ -294,6 +294,84 @@ function update_module_add_to_system($modules = array()) {
...
+    $modules_to_enable = update_upgrade_modules_to_enable();
+    if (!empty($modules_to_enable)) {
+      update_module_enable(array_keys($modules_to_enable));
+      state_set('update_upgrade_enable_dependencies', TRUE);
+    }
+  }
+}
+
+/** 
+ * Find out which modules need to be enabled.
```
bugfolder:
```suggestion
 * Finds out which modules need to be enabled.
```

Really a tiny thing: consistent verb tense.

bugfolder:
```suggestion
 * Finds out which modules need to be enabled.
```
```suggestion
 * Find out which modules need to be enabled.
```

Really a tiny thing: consistent verb tense.

argiepiano: committed

core/update.php
```diff
@@ -545,6 +549,7 @@ function update_check_requirements($skip_warnings = FALSE) {
       }
 
     case t('Apply pending updates'):
+      update_upgrade_enable_dependencies();
```
quicksketch, line 552: This perhaps should be moved _before_ the "Apply pending updates" button is clicked? The `update_fix_compatibility()` function does the exact opposite of this function: it _disables_ old modules. This function enables new ones. Perhaps the two of them could be either named similar or be merged together some how?

By moving up to above, before the updates are applied, any installation requirements that are not met by the newly installed modules will be displayed by `update_check_requirements()`. For example if a newly installed module required a PHP extension that we not available, that should be installed before attempting to upgrade the site, as update hooks may depend on the extension.

argiepiano: @quicksketch ,  I see the logic of your comment. One of the reason I decided to place this call later in the upgrade process was that the user would have a chance to see what modules would be automatically enabled before proceeding with the updates, giving them a chance to cancel. This was suggested in a comment by @docwilmot. If I move `update_upgrade_enable_dependency()` before `update_fix_compatibility()` then this option is gone.

If you are OK with this, I'll move it as requested. Let me know.

quicksketch:
Per discussion in the main issue, we've decided this is ready to go as-is. I've merged this into 1.x for 1.22.0. Big thanks @argiepiano!

PR #3951, opened 2022-02-11
Issue #4302: Allow revision log messages to be made required

### LINKED ISSUE #4302: Required revision messages
**Description of the need**
When you have content that is quite important to keep track of changes to, you can enable revisions and turn them on by default for the content type. But there doesn't seem to be a way to 'force' editors to enter a revision message when updating content...

**Proposed solution**
I'd like to see:
- The ability to make the revision message a required field on nodes
- The ability to disable/hide the 'Create new revision' checkbox (so you can't disable it)

Drupal has an [Enforce revision log message](https://www.drupal.org/project/enforce_revlog) module that sounds like what I'm envisioning. Perhaps that can be ported an merged into core...?

ghost:
Fixes https://github.com/backdrop/backdrop-issues/issues/4302


core/modules/node/node.types.inc
```diff
@@ -299,6 +299,23 @@ function node_type_form($form, &$form_state, $type = NULL) {
     '#title' => t('Create new revision by default'),
     '#default_value' => $type->settings['revision_default'],
     '#parents' => array('revision_default'),
+    '#states' => array(
+      'visible' => array(
+        'input[name="revision_enabled"]' => array('checked' => TRUE),
+      ),
+    ),
+  );
+  $form['revision']['revision']['revision_required'] = array(
```
klonos, line 308: @BWPanda can we please remove the extraneous `['revision']` from this checkbox and the 2 checkboxes before that? It seems that they have been added accidentally.

Thanks.

PR #3952, opened 2022-02-11
Issue #5498: Fix database update of langcodes in hook

### LINKED ISSUE #5498: Fixed: The "langcode" column in the node table should not be emptied when upgrading from Drupal 7
## Description of the bug

A short description of what the problem is.
While ugrading from Drupal 7 the column `language` will be renamed to `langcode` but the values are gone. So this column is empty in my case (using German language).

## Steps To Reproduce

To reproduce the behavior:
1. Use a website with German language (maybe all languages instead of English?)
2. Upgrade your website to Backdrop CMS
3. Check you `node` table

## Actual behavior

Loading a node produces the notices
`Notice: Trying to get property 'direction' of non-object in locale_preprocess_node() line 1044 of locale.module`

`Notice: Trying to get property 'langcode' of non-object in locale_preprocess_node() (line 1039 of locale.module`

## Expected behavior

Before upgrade:
grafik

After upgrade:
grafik

## Additional information

- Drupal CMS version: 7.87
- Backdrop CMS version: 1.21.1
- MySQL version: 8.0.25
- PHP version: 7.4.27

indigoxela:
Fixes https://github.com/backdrop/backdrop-issues/issues/5498

It would have also been possible to just add the two missing colons in the placeholder arrays, but actually db_update seems the cleaner approach here.

### DIFF
--- a/core/modules/node/node.install
+++ b/core/modules/node/node.install
@@ -1085,8 +1085,15 @@ function node_update_1010() {
   $default_langcode = $site_config->get('language_default');
   $other_languages = db_query("SELECT COUNT(langcode) FROM {node} WHERE langcode NOT IN (:list)", array(':list' => array('und', $default_langcode)))->fetchField();
   if ($language_count == 1 && $other_languages == 0) {
-    db_query("UPDATE {node} SET langcode = 'und' WHERE langcode = :default", array('default' => $default_langcode));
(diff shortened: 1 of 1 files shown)


klonos (approved):
LGTM 👍🏼

laryn:
Thanks @indigoxela and @klonos! Merged into `1.x` and `1.21.x`.

PR #3958, opened 2022-02-11
Issue 5501: NULL layout paths (e.g., Default layout) cause problems in PHP 8.1

### LINKED ISSUE #5501: NULL layout paths (e.g., Default layout) cause problems in PHP 8.1
## Description of the bug

In PHP 8.1, if we pass a NULL path to various `path.inc` functions like `path_is_admin($path)` (and any function that ends up at `backdrop_match_path($path, $patterns)`, in PHP 8.1, any automated test that does this will fail with this error:

```
Exception Deprecated path.inc          315                                                                             
    preg_match(): Passing null to parameter #2 ($subject) of type string is deprecated
```

This showed up when working on PR https://github.com/backdrop/backdrop/pull/3955, and that contains a workaround. But there's a fundamental issue in core, which ultimately stems from the fact that for the default layout, `getPath()` will return NULL, rather than, say, an empty string.

## Steps To Reproduce

I've added a test in the PR for this issue that creates the problem. It contains the following:

```
    $layouts = layout_load_multiple_by_path('');
    $default_layout = $layouts['default'];
    $path = $default_layout->getPath(); // this will be NULL
    $this->assertTrue(is_null($path), 'Default layout path is NULL.');

    backdrop_static_reset('path_is_admin');
    $result = path_is_admin($path); // This should throw a deprecation warning in PHP 8.1
    $this->assertFalse($result, 'Default layout is not an admin path.');

    backdrop_match_path($path, "foo\nbar"); // This should throw a deprecation warning in PHP 8.1
```

## Actual behavior

The test will pass in PHP 5.3 and 7.4, but fails in PHP 8.1. 

## Expected behavior

Pass in all versions of PHP.

## Solution(s)

I could imagine two possible mitigations:

* Modify the default layout so that its path is an empty string, rather than NULL. (Not sure what other effects that might have.)
* Modify various functions in `path.inc` to gracefully handle NULL variables passed for the `$path` variable.

I solicit further input and thoughts on the best approach.

bugfolder:
Fixes https://github.com/backdrop/backdrop-issues/issues/5501. 

<strike>(Doesn't really fix, just demonstrates. Using "Fixes" to link it to the issue.)</strike> [Update: now fixes.]

### DIFF
--- a/core/modules/layout/tests/layout.test
+++ b/core/modules/layout/tests/layout.test
@@ -3435,7 +3435,7 @@ class LayoutPathTest extends BackdropWebTestCase {
   }
 
   /**
-   * Test that the correct renderer is used
+   * Test that null paths (in default layouts) don't cause problems.
    */
   function testLayoutPath() {
     $layouts = layout_load_multiple_by_path('');
@@ -3444,9 +3444,9 @@ class LayoutPathTest extends BackdropWebTestCase {
     $this->assertTrue(is_null($path), 'Default layout path is NULL.');
 
     backdrop_static_reset('path_is_admin');
-    $result = path_is_admin($path); // This should throw a deprecation warning in PHP 8.1
+    $result = path_is_admin($path);
     $this->assertFalse($result, 'Default layout is not an admin path.');
 
-    backdrop_match_path($path, "foo\nbar"); // This should throw a deprecation warning in PHP 8.1
+    backdrop_match_path($path, "foo\nbar");
   }
 }
--- a/core/modules/layout/tests/layout.test
+++ b/core/modules/layout/tests/layout.test
@@ -3435,18 +3435,20 @@ class LayoutPathTest extends BackdropWebTestCase {
   }
 
   /**
-   * Test that null paths (in default layouts) don't cause problems.
+   * Tests that null paths (in default layouts) don't cause problems.
    */
   public function testLayoutPath() {
     $layouts = layout_load_multiple_by_path('');
     $default_layout = $layouts['default'];
-    $path = $default_layout->getPath(); // this will be NULL
+    $path = $default_layout->getPath(); // This will be NULL.
     $this->assertNull($path, 'Default layout path is NULL.');
 
+    // Make sure path_is_admin can handle a NULL path.
     backdrop_static_reset('path_is_admin');
     $result = path_is_admin($path);
     $this->assertFalse($result, 'Default layout is not an admin path.');
 
+    // Make sure backdrop_match_path can handle a NULL path.
     backdrop_match_path($path, "foo\nbar");
   }
 }
--- a/core/modules/layout/tests/layout.test
+++ b/core/modules/layout/tests/layout.test
@@ -3422,8 +3422,8 @@ class LayoutPathTest extends BackdropWebTestCase {
   protected $profile = 'testing';
   protected $admin_user;
 
-  function setUp() {
-    parent::setUp(array('layout', 'layout_test'));
+  public function setUp() {
+    parent::setUp(array('layout_test'));
 
     // Create and login admin user.
     $this->admin_user = $this->backdropCreateUser(array(
@@ -3443,12 +3443,12 @@ class LayoutPathTest extends BackdropWebTestCase {
     $path = $default_layout->getPath(); // This will be NULL.
     $this->assertNull($path, 'Default layout path is NULL.');
 
-    // Make sure path_is_admin can handle a NULL path.
+    // Make sure path_is_admin() can handle a NULL path.
     backdrop_static_reset('path_is_admin');
     $result = path_is_admin($path);
     $this->assertFalse($result, 'Default layout is not an admin path.');
 
-    // Make sure backdrop_match_path can handle a NULL path.
+    // Make sure backdrop_match_path() can handle a NULL path.
     backdrop_match_path($path, "foo\nbar");
   }
 }
--- a/core/modules/layout/tests/layout.test
+++ b/core/modules/layout/tests/layout.test
@@ -3437,7 +3437,7 @@ class LayoutPathTest extends BackdropWebTestCase {
   /**
    * Test that null paths (in default layouts) don't cause problems.
    */
-  function testLayoutPath() {
+  public function testLayoutPath() {
     $layouts = layout_load_multiple_by_path('');
     $default_layout = $layouts['default'];
     $path = $default_layout->getPath(); // this will be NULL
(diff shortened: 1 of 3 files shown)


bugfolder:
Closing in favor of https://github.com/backdrop/backdrop/pull/4050.

bugfolder:
Reopening to respond to comments on https://github.com/backdrop/backdrop/pull/4050.

core/modules/layout/tests/layout.test
```diff
@@ -3414,3 +3414,39 @@ class LayoutFlexibleTemplateTest extends BackdropWebTestCase {
...
+    $this->backdropLogin($this->admin_user);
+  }
+
+  /**
+   * Test that null paths (in default layouts) don't cause problems.
+   */
+  public function testLayoutPath() {
+    $layouts = layout_load_multiple_by_path('');
+    $default_layout = $layouts['default'];
+    $path = $default_layout->getPath(); // this will be NULL
```
indigoxela: The comment should start uppercase and end with punctuation.

core/modules/layout/tests/layout.test
```diff
@@ -3414,3 +3414,39 @@ class LayoutFlexibleTemplateTest extends BackdropWebTestCase {
...
+    $this->assertTrue(is_null($path), 'Default layout path is NULL.');
+
+    backdrop_static_reset('path_is_admin');
+    $result = path_is_admin($path);
+    $this->assertFalse($result, 'Default layout is not an admin path.');
+
+    backdrop_match_path($path, "foo\nbar");
```
indigoxela, line 3452: What is this backdrop_match_path() supposed to do? It's not evaluated. Sure, deprecation warning, but it's not clear what the purpose is here.

core/modules/layout/tests/layout.test
```diff
@@ -3414,3 +3414,39 @@ class LayoutFlexibleTemplateTest extends BackdropWebTestCase {
...
 }
+
+/**
+ * Tests that default layout paths are handled properly.
+ */
+class LayoutPathTest extends BackdropWebTestCase {
+  protected $profile = 'testing';
+  protected $admin_user;
+
+  function setUp() {
```
indigoxela: Also here: `public function setUp()`...

core/modules/layout/tests/layout.test
```diff
@@ -3414,3 +3414,39 @@ class LayoutFlexibleTemplateTest extends BackdropWebTestCase {
...
+    parent::setUp(array('layout', 'layout_test'));
```
indigoxela: I don't think you have to enable the layout module explicitly, as it's a layout test, anyway. It should be sufficient to enable layout_test.

core/modules/layout/tests/layout.test
```diff
@@ -3414,3 +3414,39 @@ class LayoutFlexibleTemplateTest extends BackdropWebTestCase {
...
+  }
+
+  /**
+   * Test that null paths (in default layouts) don't cause problems.
+   */
+  public function testLayoutPath() {
+    $layouts = layout_load_multiple_by_path('');
+    $default_layout = $layouts['default'];
+    $path = $default_layout->getPath(); // this will be NULL
+    $this->assertTrue(is_null($path), 'Default layout path is NULL.');
```
argiepiano: Probably better to use assertNull? 

```suggestion
    $this->assertNull($path, 'Default layout path is NULL.');
```

core/modules/layout/tests/layout.tests.info
```diff
@@ -57,3 +57,9 @@ name = Layout Flexible Template Test
 description = Tests that flexible layout templates perform correctly.
 group = Layout
 file = layout.test
+
+[LayoutPathTest]
+name = Layout Paths Test
+description = Tests involving layout paths, like the Default Layout.
```
argiepiano: VERY minor: "Tests" is used as a verb in other tests above. 

```suggestion
description = Tests layout paths, like the Default Layout.
```

core/modules/layout/tests/layout.test
bugfolder: No other implementation of `setUp()` in that file is declared as a `public function`. Is this one different, or are they all wrong?

core/modules/layout/tests/layout.test
bugfolder: Most of the other `parent::setUp()` calls in this file include `'layout'`. Should they all be changed?

core/modules/layout/tests/layout.test
bugfolder: Will be fixed in next iteration, thanks!

core/modules/layout/tests/layout.test
bugfolder: Explanation added in next iteration.

core/modules/layout/tests/layout.test
indigoxela: Nice catch, taking a closer look - LayoutPathTest extends BackdropWebTestCase and there function setUp() is "protected".

Methods declared without any explicit visibility keyword are defined as public.
Most of the time in all the legacy code the visibility isn't explicitly declared. IMO this doesn't have to be addressed here. Only the newly added code should declare visibility.

core/modules/layout/tests/layout.test
indigoxela: Not in this issue. :wink:

core/modules/layout/tests/layout.test
bugfolder: OK, thx for explanation.

argiepiano: Wait a moment. setUp() overrides `BackdropWebTestCase::setUp()`, which is declared `protected function`. We can't use `public function setUp()` here - that's going to throw fatal errors

argiepiano: I'm wrong. Actually going from `protected` to `public` is legit

bugfolder: But that does trigger a thought: if the base class method was protected, should the derived class method be left protected, just for consistency? I.e., does it NEED to be public?

argiepiano: That's a good point. I don't think it needs to be public, but if we are going to change that, it should be done as a separate issue, addressing all tests that extend BackdropWebTestCase (given that many of the classes that extend BackdropWebTestCase have it as public)

indigoxela: So my suggestion is still to declare visibility for *this* setUp() and let further improvements of unrelated legacy code happen in a separate issue. Using "protected" seems more consistent, but "public" doesn't do any harm, either.

quicksketch:

PR #3959, opened 2022-02-12
Issue #5503: Add required indicator to vtabs and fieldsets

### LINKED ISSUE #5503: [UX] Vertical tabs and fieldsets that contain required fields should have a required indicator
Came across this UX flaw while working on #4302, but this UX improvement is broader than just that issue:

When editing a form that has required fields "hidden" in vertical tabs that are not selected by default, or fieldsets that are collapsed by default, there is no indication that there are in fact any required fields, so the user will most likely skip these tabs/fieldsets, and click the submit button, only to be faced with an error that a required field needs to be filled:

image

What I'm proposing is that we add the "required" indicator to such fieldsets/vtabs, so that:
- the user is likely to notice the requirement before they attempt to submit the form
- if they fail to notice the requirement, at least they get an indication of which tab may hold the required field that the form validation error is complaining about:

image

The fact that the validation does not automatically switch to the tab that contains the required field is a separate UX issue...

klonos:
Fixes: https://github.com/backdrop/backdrop-issues/issues/5503


core/includes/form.inc
```diff
@@ -3729,6 +3702,63 @@ function theme_details($variables) {
...
+  _form_set_class($element, array('form-wrapper'));
+
+  $output = '<fieldset' . backdrop_attributes($element['#attributes']) . '>';
+  if (!empty($element['#title'])) {
+    // Always wrap fieldset legends in a SPAN for CSS positioning.
+    $output .= '<legend><span class="fieldset-legend">' . $element['#title'];
+    // If the fieldset contains any required fields in it, add a "required"
+    // indicator in the legend.
+    $output .= _element_has_required_children($element) ? ' <abbr class="form-required" title="' . t('This fieldset contains required fields.') . '">*</abbr>' : '';
+    $output .= '</span></legend>';
```
klonos, line 3765: This bit here is the only actual change in the `theme_fieldset()` function - it was otherwise only moved to come after the `theme_details()` function.

core/misc/vertical-tabs.js
```diff
@@ -191,19 +190,28 @@ Backdrop.verticalTab.prototype = {
...
 Backdrop.theme.prototype.verticalTab = function (settings) {
+  // Separate the actual tab title text from any other elements that may have
+  // been added to the fieldset legend (such as "required" indicators).
+  var title_children = settings.title.children();
+  settings.title.children().remove();
+  var title_text = settings.title.text();
+
   var tab = {};
-  // Calculating height in em so CSS has a chance to update height
+  // Calculating height in em, so CSS has a chance to update the height.
```
klonos, line 210: Not sure how this comment here is relevant to what is happening in code 🤷🏼

yorkshire-pudding:
@klonos 
I've checked this out in tugboat.
Works well on desktop
Doesn't work on mobile (Chrome developer mode as proxy)

I wasn't able to test the standalone fieldset as not sure how to recreate what you had in your screenshot.

One think I did note is that both the blue text and the red asterisk fail contrast checks for accessibility.  Where the asterisk is larger and against white background (i.e. on forms themselves next to a field label) it passes.  I think, though, that this is a wider issue so don't think it should affect the implementation of this PR.

PR #3963, opened 2022-02-16
Issue #5512: Fix path in hook_views_api is not really optional

### LINKED ISSUE #5512: Fixed: Views hooks are not invoked without a defined `path` in `hook_views_api()`.
## Description of the bug

According to documentation of **hook_views_api**, the **path** parameter is **_optional_**, see https://docs.backdropcms.org/api/backdrop/core%21modules%21views%21views.api.php/function/hook_views_api/1

However, if it is not specified in the hook, it is not set at all and consequently the hook is not invoked as it should be.

## Steps To Reproduce

1. Define a **hook_views_data** as a function custom_module_views_data() in a module's custom_module.views.inc file.
2. Define a **hook_views_api** in a module's custom_module.module file, without giving a path:
```
function custom_module_views_api() {
   return array(
    'api' => 3,
  );
}
```
3. Check the return value of [views_module_include](https://docs.backdropcms.org/api/backdrop/core%21modules%21views%21views.module/function/views_module_include/1)('views', true);

## Actual behavior
The return value does not include a key for the hook_views_api defined in the custom module.
Subsequently, the function [views_include_handlers](https://docs.backdropcms.org/api/backdrop/core%21modules%21views%21views.module/function/views_include_handlers/1) called from [_views_fetch_data_build](https://docs.backdropcms.org/api/backdrop/core%21modules%21views%21includes%21cache.inc/function/_views_fetch_data_build/1)() does not load it and the hook is not invoked by [module_invoke_all](https://docs.backdropcms.org/api/backdrop/core%21includes%21module.inc/function/module_invoke_all/1)('[views_data](https://docs.backdropcms.org/api/backdrop/core%21modules%21views%21views.api.php/function/hook_views_data/1)').
The correctly defined hooks are not invoked by Views module.

## Expected behavior
The hooks defined in the custom module are invoked by Views module.
The return value of views_module_include('views', true) includes a key for the hook_views_api defined in the custom module.
This is the behavior in Drupal 7 and makes the path in hook_views_api optional, as documented (and used in Drupal contrib modules. Note: I just noticed this issue while trying to port a module)).

## Additional information
Backdrop version: 1.21.1 

This can be easily fixed by adding following (missing) code to function [views_get_module_apis](https://docs.backdropcms.org/api/backdrop/core%21modules%21views%21views.module/function/views_get_module_apis/1)() before or after "$api_info['module'] = $module;" (as in the original ctools_plugin_api_info function in Drupal 7)
```
        if (!isset($api_info['path'])) {
          $api_info['path'] = backdrop_get_path('module', $module);
        }
```

indigoxela:
Fixes https://github.com/backdrop/backdrop-issues/issues/5512

### DIFF
--- a/core/modules/views/views.module
+++ b/core/modules/views/views.module
@@ -966,11 +966,16 @@ function views_module_include($api, $reset = FALSE) {
   $included_modules = array();
   foreach ($api_cache as $module => $api_info) {
     if (isset($api_info['path'])) {
-      $path = './' . $api_info['path'] . '/' . $api_info['module'] . '.' . $api . '.inc';
-      if (is_file($path)) {
-        $included_modules[$module] = $api_info;
-        include_once($path);
-      }
+      // Custom location for MODULE.views.inc file.
+      $include_path = $api_info['path'];
+    }
+    else {
+      $include_path = backdrop_get_path('module', $module);
+    }
+    $path = './' . $include_path . '/' . $api_info['module'] . '.' . $api . '.inc';
+    if (is_file($path)) {
+      $included_modules[$module] = $api_info;
+      include_once($path);
     }
   }
   return $included_modules;


klonos (approved):
LGTM 👍🏼

quicksketch:

PR #3971, opened 2022-03-03
Issue 5528: don't renumber $roles numeric array keys

### LINKED ISSUE #5528: Fixed: Administrator Role form saves wrong value on sites with numeric role machine name
## Description of the bug

Upgrading a Drupal 7 site creates roles where the machine name is a number, like "3". This breaks the code that generates the Administrator Role form element in user_admin_roles() form. Specifically the call to array_unshift() renumbers numeric keys of the $roles array.  The select element generated is:
```
<select id="edit-user-admin-role" name="user_admin_role" class="form-select">
<option value="0">disabled</option>
<option value="1" selected="selected">Administrator</option>
</select>
```
The option value for the "Administrator" choice should be "3".

Note that this will also happen if you create a role with a numeric name, such as "7".

## Steps To Reproduce

To reproduce the behavior:
1. Upgrade a Drupal 7 site with an Administrator Role set
2. Visit Administration > Configuration > User accounts > Roles
3. Notice that the Administrator Role appears to be 'disabled'
4. Change the Administrator Role to your D7 admin role, and Save.

## Actual behavior

user_admin_role configuration setting is set to 1

## Expected behavior

user_admin_role configuration setting should be the array key for "Administrator" role, i.e., "3".

jromine:
array_unshift() will renumber numeric array keys, so use array plus (+)
instead which preserves keys.

Fixes https://github.com/backdrop/backdrop-issues/issues/5528

### DIFF
--- a/core/modules/user/user.admin.inc
+++ b/core/modules/user/user.admin.inc
@@ -527,7 +527,7 @@ function user_admin_roles($form, $form_state) {
   // Add the "disabled" option to the top of the roles list (makes sure that
   // this is the selected option if the role that was previously set as admin is
   // deleted).
-  array_unshift($roles, t('disabled'));
+  $roles = array(0 => t('disabled')) + $roles;
   $form['admin_role']['user_admin_role'] = array(
     '#type' => 'select',
     '#title' => t('The role that will be automatically assigned new permissions whenever a module is enabled:'),


core/modules/user/user.admin.inc
```diff
@@ -527,7 +527,7 @@ function user_admin_roles($form, $form_state) {
   // Add the "disabled" option to the top of the roles list (makes sure that
   // this is the selected option if the role that was previously set as admin is
   // deleted).
-  array_unshift($roles, t('disabled'));
+  $roles = array(0 => t('disabled')) + $roles;
```
docwilmot, line 530: why not `$roles = array('disabled' => t('disabled')) + $roles;`?

jromine: @docwilmot Backdrop seems to use 0 to mean user_admin_role setting is disabled. So perhaps open another issue to change that to a new value, since it would need to change in many places.

argiepiano: I agree with @jromine about opening this as a separate issue. Backdrop often does this (for example in filter.admin.inc):
`  elseif ($admin_role = config_get('system.core', 'user_admin_role')) {`
meaning that you need a 0 value to disable the admin role.

quicksketch:
Thanks @jromine! Merged into 1.x and 1.21.x.

PR #3972, opened 2022-03-04
Issue #5455: Add fallback to HTTP for Update Manager

### LINKED ISSUE #5455: Add fallback to HTTP for update manager
## Description of the need

We are seeing a new frustrated Forum post regarding this (https://forum.backdropcms.org/forum/not-possible-check-upgrades). This seems to come up regularly.

I experienced this issue and the problem was that Backdrop uses SSL to check for updates, and my local certificate used by stream_socket_client to make this socket connection with the server was expired. See https://forum.backdropcms.org/forum/issues-checking-updates-localhost

Other posters found similar solutions related to expired certificates. I think this is not uncommon.

## Proposed solution
Drupal 7, to this date, uses an HTTP connection to obtain the same module update information. Backdrop's defines the constant `UPDATE_DEFAULT_URL` as `https://updates.backdropcms.org/release-history`. Drupal 7 `http://updates.drupal.org/release-history`.

It seems like D8/9 does use SSL to fetch update data, but provides a `update_fetch_with_http_fallback` fallback option in settings.php. 

It'd be great to have this feature added to Backdrop.

This is the D8/9 doc explanation for this setting:

```php
/**
 * Fallback to HTTP for Update Manager and for fetching security advisories.
 *
 * If your site fails to connect to updates.drupal.org over HTTPS (either when
 * fetching data on available updates, or when fetching the feed of critical
 * security announcements), you may uncomment this setting and set it to TRUE to
 * allow an insecure fallback to HTTP. Note that doing so will open your site up
 * to a potential man-in-the-middle attack. You should instead attempt to
 * resolve the issues before enabling this option.
 */
```

klonos:
Fixes https://github.com/backdrop/backdrop-issues/issues/5455


core/modules/update/update.fetch.inc
```diff
@@ -332,6 +332,13 @@ function _update_build_fetch_url($project, $site_key = '') {
     $url .= '&list=';
     $url .= rawurlencode(implode(',', $list));
   }
+
+  // Fall back to the less secure HTTP version of the URL (if the respective
+  // setting in settings.php has been enabled).
+  if (settings_get('update_fetch_with_http_fallback', FALSE)) {
```
indigoxela: Hm... that's actually not a "fallback", but the insecure connection is *always* used if the setting is on. Confusing variable name and comment wording.

settings.php
```diff
@@ -58,6 +58,24 @@
...
+ * uncomment this setting (and make sure it is set to "TRUE"), to allow an
+ * insecure fallback to HTTP.
+ *
+ * SECURITY WARNING: Enabling this setting will open your site up to potential
+ * man-in-the-middle attacks! You should instead attempt to resolve the issues,
+ * and only use this option as a last resort.
+ *
+ * @see https://docs.backdropcms.org/documentation/system-requirements#php
+ * @see https://en.wikipedia.org/wiki/Man-in-the-middle_attack
+ * @see _update_build_fetch_url()
```
indigoxela, line 84: The last `@see` ... are people really supposed to search the code base for a helper function? :thinking: 

References to functions are great within the code base (for developers), but this is settings.php (for site builders).

core/modules/update/update.fetch.inc
klonos: I'm using the same wording and setting name as those used in D8/9. @indigoxela are you suggesting that the setting name would better be `'update_fetch_with_http'` instead, or something else? (I'd be fine with that if others agree)

settings.php
klonos: There are multiple other places within settings.php where we use `@see` to point to functions (although not helper functions indeed). Happy to remove this if others also find it problematic 👍🏼

core/modules/update/update.fetch.inc
klonos: Renamed to `'update_fetch_with_http'` and tweaked the comment accordingly.

klonos:
I've been getting update notification emails from this PR sandbox for some reason. Closing and re-opening the PR in order to refresh it.

PR #3974, opened 2022-03-06
Issue #5532: Fix inconsistencies in the Project Installer UI when it fails to fetch results

### LINKED ISSUE #5532: [UX] Fix inconsistencies in the Project Installer UI when it fails to fetch results.
Noticed these things while working on #5455.

image

Issues:
- The message should be an error instead of status.
- The server name (`Backdrop`) in the message could use some improvement, such as calling it `Backdrop CMS Projects` of something like that.
- The totals summary "Showing 0 to 0 of 0." makes no sense if there are no results, so it should be hidden.
- The "No results found." text appears to be custom - we should be using the `'empty'` variable in `theme_table()` instead.
- There were even some cases (not shown in the screenshot above) where the pager was shown, even without any results.

klonos:
Fixes https://github.com/backdrop/backdrop-issues/issues/5532

### DIFF
--- a/core/modules/installer/config/installer.settings.json
+++ b/core/modules/installer/config/installer.settings.json
@@ -2,7 +2,7 @@
     "_config_name": "installer.settings",
     "installer_server": {
         "url": "https://projects.backdropcms.org",
-        "name": "Backdrop"
+        "name": "Backdrop CMS Projects"
     },
     "core_update": true
 }
--- a/core/modules/installer/installer.browser.inc
+++ b/core/modules/installer/installer.browser.inc
@@ -252,8 +252,8 @@ function installer_browser_fetch_results($filters) {
     $results = backdrop_json_decode($response->data);
   }
   else {
-    backdrop_set_message(t("Encountered an error when trying to fetch results from @name. Error @code : @message",
-      array('@name' => $server['name'], '@code' => $response->code, '@message' => $response->error)));
+    backdrop_set_message(t('Encountered an error when trying to fetch projects from %installer_server. Error @code : @message',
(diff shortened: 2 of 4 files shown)


core/modules/installer/config/installer.settings.json
```diff
@@ -2,7 +2,7 @@
     "_config_name": "installer.settings",
     "installer_server": {
         "url": "https://projects.backdropcms.org",
-        "name": "Backdrop"
+        "name": "Backdrop CMS Projects"
```
indigoxela, line 5: Why that?

core/modules/installer/installer.browser.inc
```diff
@@ -252,8 +252,8 @@ function installer_browser_fetch_results($filters) {
     $results = backdrop_json_decode($response->data);
   }
   else {
-    backdrop_set_message(t("Encountered an error when trying to fetch results from @name. Error @code : @message",
-      array('@name' => $server['name'], '@code' => $response->code, '@message' => $response->error)));
+    backdrop_set_message(t('Encountered an error when trying to fetch projects from %installer_server. Error @code : @message',
+      array('%installer_server' => check_plain($server['name']), '@code' => $response->code, '@message' => $response->error)), 'error');
```
indigoxela, line 256: This is very hard to read, I'd suggest to display array items on their individual lines:

```
 backdrop_set_message(t('Encountered an error when trying to fetch projects from %installer_server. Error @code : @message', array(
   '%installer_server' => check_plain($server['name']),
   '@code' => $response->code,
   '@message' => $response->error,
 )), 'error');
```

PR #3977, opened 2022-03-08
Update views_plugin_style.inc

### LINKED ISSUE #5143: Fixed grammar error in Views row class description text
**Steps To Reproduce**

To reproduce the behavior:
1. Navigate to Structure -> Views (`admin/structure/views`)
2. Edit any view (click the "Configure" button)
3. Under the "Format" section, click "Settings"
4. In the pop-up that opens, scroll down to the "Row class" setting:
   image
5. The description text under Row class" says:
   > The class to provide on each row. You may use field tokens from as per the "Replacement patterns" used in "Rewrite the output of this field" for all fields.

   The "as per" is unexpected.

## Expected behavior

The description should probably change to something like:
> "The class to provide on each row. You may use field tokens from the "Replacement patterns" used in "Rewrite the output of this field" for all fields.

ayushi424:
Fixes: https://github.com/backdrop/backdrop-issues/issues/5143

(The line above creates an automatic link with the original issue)

Related Issue: Grammar error in Views display page style row class description text
Issue No:  #5143

Changes done:
* Optimized the file by removing unexpected "as per".

### DIFF
--- a/core/modules/views/plugins/views_plugin_style.inc
+++ b/core/modules/views/plugins/views_plugin_style.inc
@@ -256,7 +256,7 @@ class views_plugin_style extends views_plugin {
       );
 
       if ($this->uses_fields()) {
-        $form['row_class']['#description'] .= ' ' . t('You may use field tokens from as per the "Replacement patterns" used in "Rewrite the output of this field" for all fields.');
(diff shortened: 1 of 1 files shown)


ghost:
Thanks! Merged into 1.x and 1.21.x.

PR #3981, opened 2022-03-11
Issue #5525: Background processing support for LiteSpeed server.

### LINKED ISSUE #5525: Cached anonymous page and "background fetch" cause site rendering issues on LiteSpeed server
## Description of the bug

When using  "background fetch" (see admin/config/development/performance) of stale anonymous cached pages, Backdrop (or the LiteSpeed server?) seems to send headers and content twice. 

I've been able to reproduce in a LiteSpeed production server with different sites, ~~including a clean Backdrop install~~. Have NOT been able to reproduce on a local Apache one.
  
This issue causes duplication of content and a lot of problems when rendering the page (see attached image), among them also `Warning: session_set_save_handler(): Cannot change save handler when headers already sent in backdrop_session_initialize() (line 245 of /home/tenutowe/public_html/backdrop/core/includes/session.inc).`

## Steps To Reproduce
These are very specific steps. Use two browsers, or one browser window on "normal" mode and another on "incognito" . Again, this is an issue in LiteSpeed server as far as I know.

1. In "browser A", log in to your site, navigate to admin/config/development/performance and be sure "Cache pages for anonymous users" and "Use background fetch for cached pages" are on
2. Turn OFF "Compress cached pages." This will allow you to actually see the duplicated content (otherwise Backdrop sends the "doubled page" as a gzip, which is not displayed).
3. Make the expiration of cache pages to 1 minute (so that you don't have to wait too long to reproduce the error 😄 )
4. Save
5. Navigate to admin/config/development/logging and select "All messages" in "Error messages to display"

On a different browser (browser B):
1. Clear the memory cache of the browser.  This is done to assure the browser actually sends a request, rather than rendering the page from memory. This is done in different ways in different browsers. In Chrome, open the Developer Tools, and then Control Click the refresh button - select Empty Cache and Hard Reload
2. Navigate to the home page of your site
6. Click Log in
7. Log in as a user. DO NOT clear the Backdrop caches!
8. Navigate to Home
9. Clear the BROWSER memory cache one more time (see 1)
10. Sit patiently for one minute to wait for the cached anonymous pages to go stale (I put a timer!)
11. After one minute (or longer), click Log out

Upon logout you will be redirected to the Home page. This version of the Home page is retrieved from a stale entry in the cache_page table, which caches anonymous pages. You should see duplication of elements and the Warning I printed above.

You can verify that the response came from the Backdrop cache by opening the Developer Tools of the browser, going to "Network", selecting the document, and clicking "Header" to see the response headers received. They should indicate a cached page on the Response header `x-backdrop-cache: HIT`  (if you see MISS, you are not seeing a cached page).
 
## Actual behavior
See above

## Expected behavior
The page should load without any issues.

## Additional information
- Backdrop CMS version: 1.21.2
- Web server and its version: LiteSpeed V7.9 Cloudlinux 1.3
- Browser(s) and their versions: Chrome and Safari (same issue in both). 

Screen Shot 2022-03-02 at 8 13 04 AM

-----
ANOTHER SNAPSHOT, FROM A CLEAN INSTALL:

Screen Shot 2022-03-02 at 2 39 58 PM

quicksketch:
Fixes https://github.com/backdrop/backdrop-issues/issues/5525.

### DIFF
--- a/core/includes/bootstrap.inc
+++ b/core/includes/bootstrap.inc
@@ -3117,6 +3117,10 @@ function _backdrop_bootstrap_page_cache() {
       if (function_exists('fastcgi_finish_request')) {
         fastcgi_finish_request();
       }
+      // Flushing for LiteSpeed servers.
+      elseif (function_exists('litespeed_finish_request')) {
+        litespeed_finish_request();
+      }
       // Flushing for Apache mod_php installations.
       else {
         flush();
--- a/core/includes/common.inc
+++ b/core/includes/common.inc
@@ -3188,6 +3188,10 @@ function backdrop_page_footer() {
     if (function_exists('fastcgi_finish_request')) {
       fastcgi_finish_request();
     }
+    // Flushing for LiteSpeed servers.
+    elseif (function_exists('litespeed_finish_request')) {
+      litespeed_finish_request();
+    }
     // Flushing for Apache mod_php installations.
     else {
       flush();


quicksketch:
Merged into 1.x and 1.21.x. Thanks for reviewing @argiepiano and @yorkshire-pudding.

PR #3982, opened 2022-03-12
Issue #2062 - Change the label and description of maximum image size

### LINKED ISSUE #2062: [UX] Remove left-over help text on Maximum image size field
The description text for the Max image size field reads:

```
The maximum allowed image size expressed as WIDTHxHEIGHT (e.g. 640x480). Leave blank for no restriction. If a larger image is uploaded, it will be resized to reflect the given width and height. Resizing images on upload will cause the loss of EXIF data in the image.
```

We should remove the first part (`The maximum allowed image size expressed as WIDTHxHEIGHT (e.g. 640x480).`) because now that we have pretty integer fields that part of the explanation is nolonger necessary.

stpaultim:
Fixes: https://github.com/backdrop/backdrop-issues/issues/2062

### DIFF
--- a/core/modules/image/image.field.inc
+++ b/core/modules/image/image.field.inc
@@ -125,12 +125,12 @@ function image_field_instance_settings_form($field, $instance) {
   $min_dimensions[1] = empty($min_dimensions[1]) ? NULL : $min_dimensions[1];
   $form['min_dimensions'] = array(
     '#type' => 'item',
-    '#title' => t('Minimum image dimensions'),
+    '#title' => t('Minimum image dimensions (width x height)'),
     '#element_validate' => array('_image_field_resolution_validate'),
     '#weight' => 4.2,
     '#field_prefix' => '<div class="container-inline">',
     '#field_suffix' => '</div>',
-    '#description' => t('The minimum allowed image dimensions expressed as <code>WIDTH</code> x <code>HEIGHT</code>. Leave blank for no restriction. If a smaller image is uploaded, it will be rejected.'),
+    '#description' => t('Leave blank for no restriction. If a smaller image is uploaded, it will be rejected.'),
   );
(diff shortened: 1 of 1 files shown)


bugfolder:
Reopening this PR since there seems to be new interest in explicit "width × height".

core/modules/image/image.field.inc
```diff
@@ -94,12 +94,12 @@ function image_field_instance_settings_form($field, $instance) {
   $max_dimensions[1] = empty($max_dimensions[1]) ? NULL : $max_dimensions[1];
   $form['max_dimensions'] = array(
     '#type' => 'item',
-    '#title' => t('Maximum image dimensions'),
+    '#title' => t('Maximum image dimensions (width x height)'),
```
bugfolder, line 97:
```suggestion
    '#title' => t('Maximum image dimensions (width × height)'),
```
Suggest using Unicode "multiplication" rather than letter x.

core/modules/image/image.field.inc
```diff
@@ -125,12 +125,12 @@ function image_field_instance_settings_form($field, $instance) {
   $min_dimensions[1] = empty($min_dimensions[1]) ? NULL : $min_dimensions[1];
   $form['min_dimensions'] = array(
     '#type' => 'item',
-    '#title' => t('Minimum image dimensions'),
+    '#title' => t('Minimum image dimensions (width x height)'),
```
bugfolder, line 128:
```suggestion
    '#title' => t('Maximum image dimensions (width × height)'),
```
Suggest using Unicode "multiplication" rather than letter x.

jenlampton: This change recommends `Maximum` over `Minimum` (I think that was a typo)

bugfolder:
Closing in favor of https://github.com/backdrop/backdrop/pull/4523.

PR #3983, opened 2022-03-12
Issue #2062: Add placeholder text for image field width/height.

### LINKED ISSUE #2062: [UX] Remove left-over help text on Maximum image size field

ghost:
Fixes https://github.com/backdrop/backdrop-issues/issues/2062

### DIFF
--- a/core/modules/image/image.field.inc
+++ b/core/modules/image/image.field.inc
@@ -99,25 +99,29 @@ function image_field_instance_settings_form($field, $instance) {
     '#weight' => 4.1,
     '#field_prefix' => '<div class="container-inline">',
     '#field_suffix' => '</div>',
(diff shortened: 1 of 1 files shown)


quicksketch:
New version of this PR at https://github.com/backdrop/backdrop/pull/4523

PR #3988, opened 2022-03-13
Issue 5213: Hide the Check manually button if user doesn't have the r…

### LINKED ISSUE #5213: Dashboard: Either hide the "Check manually" button, or the entire "Available updates" block, if user doesn't have the required permission
1. Grant the "Access Dashboard" permission to the "Authenticated" role
2. Create a user; do not assign any role to them
3. Log into the site with the user you created in the previous step, and navigate to admin/dashboard ->
   - the "Available updates" block is there (I guess that's OK - should we not be showing this block at all if the user doesn't have the "View site reports" permission?)
   - clicking the "Check manually" button throws a 404 👍🏼 ...but in that case the button should not be there in the first place.

RakeshPotnuru:
…equired permission

Fixes https://github.com/backdrop/backdrop-issues/issues/5213

- User with Administration role
Screenshot 2022-03-13 212334

- User without Administration role
Screenshot 2022-03-13 212307

### DIFF
--- a/core/modules/update/update.theme.inc
+++ b/core/modules/update/update.theme.inc
@@ -42,8 +42,9 @@ function theme_update_last_check($variables) {
   $output = '<div class="update checked">';
 
   $output .= $last ? t('Last checked: @time ago', array('@time' => format_interval(REQUEST_TIME - $last))) : t('Last checked: never') . ' ';
-  $output .= l(t('Check manually'), 'admin/reports/updates/check', array('attributes' => array('class' => array('button', 'button-secondary')), 'query' => backdrop_get_destination()));
-
+  if (user_access('access site reports')) {
+    $output .= l(t('Check manually'), 'admin/reports/updates/check', array('attributes' => array('class' => array('button', 'button-secondary')), 'query' => backdrop_get_destination()));
+  }
   if (module_exists('installer') && user_access('access_site_reports')) {
     // Link to the updates installation page from the available updates report
     // (don't add this link to the "Available updates" Dashboard block).
--- a/core/modules/update/update.theme.inc
+++ b/core/modules/update/update.theme.inc
@@ -42,7 +42,7 @@ function theme_update_last_check($variables) {
   $output = '<div class="update checked">';
 
   $output .= $last ? t('Last checked: @time ago', array('@time' => format_interval(REQUEST_TIME - $last))) : t('Last checked: never') . ' ';
-  if (user_access('access site reports')) {
+  if (user_access('Administer site configuration')) {
     $output .= l(t('Check manually'), 'admin/reports/updates/check', array('attributes' => array('class' => array('button', 'button-secondary')), 'query' => backdrop_get_destination()));
   }
   if (module_exists('installer') && user_access('access_site_reports')) {
--- a/core/modules/update/update.theme.inc
+++ b/core/modules/update/update.theme.inc
@@ -42,7 +42,7 @@ function theme_update_last_check($variables) {
   $output = '<div class="update checked">';
 
   $output .= $last ? t('Last checked: @time ago', array('@time' => format_interval(REQUEST_TIME - $last))) : t('Last checked: never') . ' ';
-  if (user_access('Administer site configuration')) {
+  if (user_access('administer site configuration')) {
     $output .= l(t('Check manually'), 'admin/reports/updates/check', array('attributes' => array('class' => array('button', 'button-secondary')), 'query' => backdrop_get_destination()));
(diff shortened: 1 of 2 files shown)


core/modules/update/update.theme.inc
```diff
@@ -42,8 +42,9 @@ function theme_update_last_check($variables) {
   $output = '<div class="update checked">';
 
   $output .= $last ? t('Last checked: @time ago', array('@time' => format_interval(REQUEST_TIME - $last))) : t('Last checked: never') . ' ';
-  $output .= l(t('Check manually'), 'admin/reports/updates/check', array('attributes' => array('class' => array('button', 'button-secondary')), 'query' => backdrop_get_destination()));
-
+  if (user_access('access site reports')) {
```
stpaultim: It turns out that this is the wrong permission to check. We should be checking the permission for: `Administer site configuration`

That is the permission that controls if the user is able to manually check for updates.

RakeshPotnuru: Done

core/modules/update/update.theme.inc
```diff
@@ -42,8 +42,9 @@ function theme_update_last_check($variables) {
   $output = '<div class="update checked">';
 
   $output .= $last ? t('Last checked: @time ago', array('@time' => format_interval(REQUEST_TIME - $last))) : t('Last checked: never') . ' ';
-  $output .= l(t('Check manually'), 'admin/reports/updates/check', array('attributes' => array('class' => array('button', 'button-secondary')), 'query' => backdrop_get_destination()));
-
+  if (user_access('Administer site configuration')) {
```
stpaultim: Sorry, I gave you bad advice again. The permission must be in lower case. Simply change "Adminster" to "administer".

`if (user_access('administer site configuration')) {`

RakeshPotnuru: Can you just edit it? I will commit suggestion

stpaultim: This is quite strange. I didn't think I could, but I tried and I was in fact able to.

stpaultim:
@RakeshPotnuru - Yeah. Tests all passed!

stpaultim:
@RakeshPotnuru - I didn't notice this earlier, but we actually need to hide this entire block if the user does not have permission to view reports. We also need to keep the change you already made.

I believe that we do this by adding:

`if (user_access('access site reports')) {` 

At LINE 58 in core/modules/dashboard/includes/block.update.inc

I welcome other ideas.

RakeshPotnuru:
@stpaultim 
Is this it? After adding this I am still able to see **Available Updates**
```php
if (user_access('access site reports')) {
  $panel = array(
    '#theme' => 'dashboard_panel__update',
  );
}
```

stpaultim:
@RakeshPotnuru - Sorry, I should have been clear. What worked for me was to put the initial if statement at line 58 and close it on line 211. Basically, to put that entire function inside the if statement. 

Again - I am not sure if this is the best place to put the IF statement, but once we have a PR we can get feedback from others. It is always easier to get feedback if you have specific code changes for people to look at. 

For core development most Pull Requests are rewritten and improved several times, even those by our most experienced develoeprs.

Why don't you give my idea a try and we'll see if that works and if anyone has a better idea.

RakeshPotnuru:
@stpaultim 
function ends at line 207. So we can't close `if` block on line 211.

stpaultim:
@RakeshPotnuru - well, I might have added some extra lines in that were not necessary. Is this PR working for you?

RakeshPotnuru:
@stpaultim 
yes
image

stpaultim:
OK, I tested the code locally and it seems to work for me too. 

You only changed two lines of code in the core/modules/dashboard/includes/block.update.inc file, correct?
It just looks so complicated in the diff. I assume that is because of the shift in indents only.

I won't be able to look at this again until much later today. Maybe someone else will have time to look.

RakeshPotnuru:
Yes
On line 58 - `if (user_access('access site reports')) {`
Indented whole code block by two spaces
On line 207 - `}`

klonos (approved):
LGTM 👍🏼  ...thanks @RakeshPotnuru 🙏🏼

ghost (approved):
Looks great!

quicksketch:
Perfect! Thanks @RakeshPotnuru et al for fixing this bug! I've merged into 1.x and 1.21.x. I'm sorry for the delay that missed the OpenForce window. Really appreciate your work in joining us @RakeshPotnuru!

PR #3992, opened 2022-03-15
Issue #4645: Add non-US date formats to core.

### LINKED ISSUE #4645: Configure default date formats based on timezone/country.
**Description of the need**

As highlighted in https://github.com/backdrop/backdrop-issues/issues/989, we don't have any OOTB date formats that start with the day as is frequently used in European countries.

Available formats seem to be heavily orientated towards users in the USA. 

**Proposed solution**

Add at least one simple date format that starts follows day/month/year format to address this large European format. I recommend:

DD/MM/YYYY

Users can easier create their own date formats, which is useful because there are so many different options available. 

Given the almost infinite number of possible date formats, I understand that it might be difficult to settle on one date format that is common enough to warrant inclusion in core. But, I think that we should pick the most commonly used European format that starts with the date and include it. 

**Additional information**

"Official EU documents still tend to use DD.MM.YYYY but one document specifies the use of ISO 8601: "Dates should be formatted by the following format: YYYY-MM-DD."
https://en.wikipedia.org/wiki/Date_and_time_notation_in_Europe

I would also like a simple MM/DD/YYYY format as well, but agreeing to two new date formats at once is probably unlikely. 

Current options:

Date___Beetles

Would like the input of other European users on this regarding it's usefulness and/or which specific format would be most useful?

ghost:
Fixes https://github.com/backdrop/backdrop-issues/issues/4645

### DIFF
--- a/core/modules/system/config/system.date.json
+++ b/core/modules/system/config/system.date.json
@@ -8,17 +8,32 @@
     "user_empty_timezone_message": 0,
     "formats": {
         "long": {
-            "label": "Long date",
-            "pattern": "l, F j, Y - g:ia",
+            "label": "Long",
+            "pattern": "l, j F, Y - g:ia",
             "module": "system"
         },
         "medium": {
-            "label": "Medium date",
-            "pattern": "D, m/d/Y - g:ia",
+            "label": "Medium",
+            "pattern": "D, d/m/Y - g:ia",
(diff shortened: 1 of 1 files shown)


ghost:
Closing in favour of https://github.com/backdrop/backdrop/pull/4005

PR #3996, opened 2022-03-17
Issue 5003: Add a link to the BackdropCMS.org page on module list

### LINKED ISSUE #5003: Add a link to the BackdropCMS.org page on module list for each module
**Description of the need**
On the module list at `admin/modules` page would be nice to be able to see a link to the module's detail page on BackdropCMS.org and/or the repo on GH. 

**Proposed solution**
Add a link under the "more/less" toggle.

RakeshPotnuru:
Fixes https://github.com/backdrop/backdrop-issues/issues/5003

### DIFF
--- a/core/modules/system/system.theme.inc
+++ b/core/modules/system/system.theme.inc
@@ -178,6 +178,9 @@ function theme_system_modules_fieldset($variables) {
     if ($required_by) {
       $requirements .= '<div class="admin-requirements">' . t('Required by:') . ' ' . implode(', ', $module['#required_by']) . '</div>';
     }
+    if (!$module['#core']) {
+      $requirements .= '<div class="admin-requirements project-links">' . t('Links:') . ' ' . l(t("BackdropCMS.org"), 'https://backdropcms.org/project/' . $key) . ', ' . l(t("GitHub"), 'https://github.com/backdrop-contrib/' . $key) . '</div>';
+    }
     $requirements .= '<div class="admin-requirements table-filter-text-source">' . t('Tags:') . ' ' . backdrop_render($module['tags']) . '</div>';
     $requirements .= '</div>';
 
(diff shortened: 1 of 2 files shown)


core/modules/system/system.theme.inc
```diff
@@ -178,6 +178,9 @@ function theme_system_modules_fieldset($variables) {
     if ($required_by) {
       $requirements .= '<div class="admin-requirements">' . t('Required by:') . ' ' . implode(', ', $module['#required_by']) . '</div>';
     }
+    if (!$module['#project']) {
```
docwilmot: We want these links for contrib modules, and only contrib would have a `project` key, so this should be 
```
    if (isset($module['#project'])) {
```

RakeshPotnuru: Done!

core/modules/system/system.theme.inc
```diff
@@ -178,6 +178,9 @@ function theme_system_modules_fieldset($variables) {
     if ($required_by) {
       $requirements .= '<div class="admin-requirements">' . t('Required by:') . ' ' . implode(', ', $module['#required_by']) . '</div>';
     }
+    if (isset($module['#project'])) {
+      $requirements .= '<div class="admin-requirements project-links">' . t('Links:') . ' ' . l(t("BackdropCMS.org"), 'https://backdropcms.org/project/' . $key) . ', ' . l(t("GitHub"), 'https://github.com/backdrop-contrib/' . $key) . '</div>';
```
quicksketch, line 182: Because projects can contain multiple modules, we should use the `$module['#project']` variable here. This will result in multiple modules pointing at the same GitHub URL, but that is fine.

```suggestion
      $requirements .= '<div class="admin-requirements project-links">' . t('Links:') . ' ' . l(t("BackdropCMS.org"), 'https://backdropcms.org/project/' . $module['#project']) . ', ' . l(t("GitHub"), 'https://github.com/backdrop-contrib/' . $module['#project']) . '</div>';
```

quicksketch:
Replaced by https://github.com/backdrop/backdrop/pull/4063 (which used this as the base). Thanks @RakeshPotnuru!

PR #3998, opened 2022-03-18
Views cache key doesn't include enough uniqueness

### LINKED ISSUE #5551: Views cache causes validation errors
## Description of the bug

The References module can use a view to list valid users or nodes. Views caching can cause validation failures.

## Steps To Reproduce

To reproduce the behavior:

1. enable the references module
2. create a view that returns a list of some users including "user a" and "user b"
3. set the view to cache its results, time-based, 1hour
4. create a content type with a user reference field
5. configure the user reference field to use the user view as its allowed value list
6. create a node of the new content type with the user reference field value "user a"
7. edit the node, and attempt to change the user reference field value to "user b"
8. or create a new node of the new content type with the user reference field value "user b"

## Actual behavior

Validation fails, can not save the node.
image

## Expected behavior

Since the user is a valid user, validation should pass.

## Additional information

A workaround is to set the view cache to none.
Also, flushing the backdrop cache between steps 6 and 7 causes validation to succeed.
@jromine Investigated this issue and found a matching Drupal issue https://www.drupal.org/project/drupal/issues/1055616

This was originally reported in the References issue queue https://github.com/backdrop-contrib/references/issues/12

jromine:
Fixes https://github.com/backdrop/backdrop-issues/issues/5551

Applies patch from https://www.drupal.org/project/drupal/issues/1055616#comment-8610445

### DIFF
--- a/core/modules/views/plugins/views_plugin_cache.inc
+++ b/core/modules/views/plugins/views_plugin_cache.inc
@@ -284,59 +284,77 @@ class views_plugin_cache extends views_plugin {
     }
   }
 
-  function get_results_key() {
-    global $user;
-
+  public function get_results_key() {
     if (!isset($this->_results_key)) {
-
-      $build_info = $this->view->build_info;
-
-      $query_plugin = $this->view->display_handler->get_plugin('query');
-
-      foreach (array('query','count_query') as $index) {
-        // If the default query back-end is used generate SQL query strings from
-        // the query objects.
-        if ($build_info[$index] instanceof SelectQueryInterface) {
-          $query = clone $build_info[$index];
-          $query->preExecute();
-          $build_info[$index] = (string) $query;
-        }
-      }
-      $key_data = array(
-        'build_info' => $build_info,
-        'roles' => array_keys($user->roles),
-        'super-user' => $user->uid == 1, // special caching for super user.
-        'language' => $GLOBALS['language']->langcode,
-        'base_url' => $GLOBALS['base_url'],
-      );
+      $key_data = array();
       foreach (array('exposed_info', 'page', 'sort', 'order', 'items_per_page', 'offset') as $key) {
         if (isset($_GET[$key])) {
           $key_data[$key] = $_GET[$key];
         }
       }
 
-      $this->_results_key = $this->view->name . ':' . $this->display->id . ':results:' . md5(serialize($key_data));
+      $this->_results_key = $this->view->name . ':' . $this->display->id . ':results:' . $this->get_cache_key($key_data);
     }
 
     return $this->_results_key;
   }
 
-  function get_output_key() {
-    global $user;
+  public function get_output_key() {
     if (!isset($this->_output_key)) {
       $key_data = array(
         'result' => $this->view->result,
-        'roles' => array_keys($user->roles),
-        'super-user' => $user->uid == 1, // special caching for super user.
         'theme' => $GLOBALS['theme'],
-        'language' => $GLOBALS['language']->langcode,
-        'base_url' => $GLOBALS['base_url'],
       );
-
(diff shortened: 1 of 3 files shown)


core/modules/views/plugins/views_plugin_cache.inc
```diff
@@ -284,60 +284,78 @@ class views_plugin_cache extends views_plugin {
...
+   *
+   * @return string
+   */
+  public function get_cache_key($key_data = array()) {
+    global $user;
+
+    $key_data += array(
+      'roles' => array_keys($user->roles),
+      'super-user' => $user->uid == 1,
+    // special caching for super user.
```
klonos: This change seems to be a side-effect of the (HUGE) commit here: https://git.drupalcode.org/project/views/-/commit/84facc81fc2ff1e02be952758c07bb2736a0b987

There is nothing in neither Drupal nor Backdrop coding standards that suggests that comments should be wrapped this way, so we should revert this change. While at it, we should capitalize the first letter of the sentense:

```suggestion
      'super-user' => $user->uid == 1, // Special caching for super user.
```

core/modules/views/plugins/views_plugin_cache.inc
```diff
@@ -284,60 +284,78 @@ class views_plugin_cache extends views_plugin {
...
+      'language' => $GLOBALS['language']->langcode,
+      'language_content' => $GLOBALS['language_content']->langcode,
+      'base_url' => $GLOBALS['base_url'],
```
klonos, line 331: Thanks for including the changes from https://git.drupalcode.org/project/views/-/commit/aff92c08791e5bb01271e5e661375d04f46fcd8a in the line prior to this one @jromine 🙏🏼  ...can we please also include the line from https://git.drupalcode.org/project/views/-/commit/d5703d76b4ca079cd64ae1a64f7b62e779b50d69?:

```suggestion
      'base_url' => $GLOBALS['base_url'],
      'display_options' => $this->display->display_options,
```

core/modules/views/plugins/views_plugin_cache.inc
```diff
@@ -284,60 +284,78 @@ class views_plugin_cache extends views_plugin {
...
+        // the query objects.
+        if ($build_info[$index] instanceof SelectQueryInterface) {
+          $query = $build_info[$index];
+
+          // If the query was not yet prepared, work on a clone and run
+          // preExecute().
+          if (!$query->isPrepared()) {
+            $query = clone $build_info[$index];
+            $query->preExecute();
+          }
```
klonos, line 348: FTR, this is from https://git.drupalcode.org/project/views/-/commit/3f1efccdeed64ad339ea7e2e34968fa7ada5da7b

core/modules/views/plugins/views_plugin_query_default.inc
```diff
@@ -1369,6 +1369,16 @@ class views_plugin_query_default extends views_plugin_query {
     // Add all query substitutions as metadata.
     $query->addMetaData('views_substitutions', module_invoke_all('views_query_substitutions', $this));
 
+    if (!$get_count) {
+      if (!empty($this->limit) || !empty($this->offset)) {
+        // We can't have an offset without a limit, so provide a very large limit
+        // instead.
```
klonos: Comment exceeds the 80 character limit here:

```suggestion
        // We can't have an offset without a limit, so provide a very large
        // limit instead.
```

jromine:
@klonos thank you for the code review. I have accepted all your suggestions.

core/modules/views/plugins/views_plugin_cache.inc
jromine: Noted

klonos (approved):
LGTM 👍🏼

quicksketch:

PR #4000, opened 2022-03-18
Changed links to show underline on hover

tanishagupta0811:
Fixed issue [#4670](https://github.com/backdrop/backdrop-issues/issues/4670)


gitressa:
Thanks @tanisha-tech360, just a tip: The syntax to link an issue with a PR via the first comment is this:

`Fixes *ISSUE_URL*`

> **Development**
> bugfolder/backdrop

So if you update the syntax in this PR, your issue will probably also be linked, where it now says "No branches or pull requests".

PR #4009, opened 2022-03-23
5557 Adding link to full book outline

### LINKED ISSUE #5557: [UX] Book Outline - Make it easier to find and reorder book pages
## Description of the bug

I was just working on a book and wanted to find the place where I could drag and drop pages to change the order. This "Outline" tab really confused me for a long time. It seemed like I was in the right place, but this page just doen't do what I think it should do. 

image

In fact, I need to go here to find the drag and drop outline. /admin/content/book

image

## Expected behavior

At the very minimum, I think there should be a link on the `Outline` page that takes you to the place where you can actually edit the outline - drag and drop. 

Also, I think the "Outline" page/tab might need to be renamed. It seems that this is the page where you select the book and parent. But, you don't really have full control over where the page does in the Outline here. 

Also, why do we have a horizontal tab ad the bottom AND a tab at the top that do the same thing. Is there a reason for this, do we do this anywhere else?

image

PROPOSED FIRST STEPS:

1) Remove the "Outline" tab at the top of the page. It duplicates the vertical tab at the bottom of the page, and personally I think the tab at the bottom makes more sense. 

2) Add a link to the Outline tab that leads user to the Manage Books page. Something like this:

image

stpaultim:
Fixes: https://github.com/backdrop/backdrop-issues/issues/5557

### DIFF
--- a/core/modules/book/book.pages.inc
+++ b/core/modules/book/book.pages.inc
@@ -63,6 +63,14 @@ function book_outline_form($form, &$form_state, Node $node) {
   if (!isset($node->book['parent_depth_limit'])) {
     $node->book['parent_depth_limit'] = _book_parent_depth_limit($node->book);
   }
+
+  if ($node->book['bid'] != 0) {
+    $form['html'] = array(
+     '#type' => 'markup',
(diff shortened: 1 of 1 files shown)


stpaultim:
Closing this in favor of: https://github.com/backdrop/backdrop-issues/issues/5331

PR #4010, opened 2022-03-23
Issue #[5397]: ["Show as Expanded" help text needs revising]

### LINKED ISSUE #5397: Added: "Show as Expanded" help text needs revising.
## Description of the bug

The problem is that either this checkbox is not working, or the help text does a very poor job of explaining the purpose of this checkbox.

## Steps To Reproduce

1) Spin up a fresh site.
2) Create a test page with a menu link that is a child of "About"
3) Edit the menu link for "About" to "show as expanded" (see screenshot)

## Actual behavior

The About menu seems to act exactly the same whether or not this checkbox is checked. 

## Expected behavior

I could be wrong, but I expected that the child menu would be visible all the time - even without hovering. The help text says "If selected and this menu link has children, the menu will always appear expanded."

While this menu has a child, but it does not always appear expanded. In fact, it is only expanded when I hover over the link - as it was before I checked this box. 

I am open to the idea that I am completely misunderstanding what this checkbox should do, in which case I think we need to revise the help text, because I've spent 20 minutes trying to figure out what this checkbox does, to no avail.

## Additional information

image

AnjaliJha1221:
Fixes https://github.com/backdrop/backdrop-issues/issues/5397
The help text for this issue is revised from "If selected and this menu link has children, the menu will always appear expanded" to "When displayed in a 'hierarchial tree' menu style, children of expanded menu links will always be visible."

### DIFF
--- a/core/modules/menu/menu.admin.inc
+++ b/core/modules/menu/menu.admin.inc
@@ -403,7 +403,7 @@ function menu_edit_item($form, &$form_state, $type, $item, $menu) {
     '#type' => 'checkbox',
     '#title' => t('Show as expanded'),
     '#default_value' => $item['expanded'],
-    '#description' => t('When displayed in a 'hierarchial tree' menu style, children of expanded menu links will always be visible.'),
+    '#description' => t("When displayed in a 'hierarchial tree' menu style, children of expanded menu links will always be visible."),
   );
 
   // Generate a list of possible parents (not including this link or descendants).
--- a/core/modules/menu/menu.admin.inc
+++ b/core/modules/menu/menu.admin.inc
@@ -403,7 +403,7 @@ function menu_edit_item($form, &$form_state, $type, $item, $menu) {
     '#type' => 'checkbox',
     '#title' => t('Show as expanded'),
     '#default_value' => $item['expanded'],
-    '#description' => t('If selected and this menu link has children, the menu will always appear expanded.'),
+    '#description' => t('When displayed in a 'hierarchial tree' menu style, children of expanded menu links will always be visible.'),
   );
 
   // Generate a list of possible parents (not including this link or descendants).


core/modules/menu/menu.admin.inc
```diff
@@ -403,7 +403,7 @@ function menu_edit_item($form, &$form_state, $type, $item, $menu) {
     '#type' => 'checkbox',
     '#title' => t('Show as expanded'),
     '#default_value' => $item['expanded'],
-    '#description' => t('If selected and this menu link has children, the menu will always appear expanded.'),
+    '#description' => t('When displayed in a 'hierarchial tree' menu style, children of expanded menu links will always be visible.'),
```
stpaultim: The single quotes are causing a problem. I believe that you will need to use double quotes for "hierarchial tree" or for the entire string statement. Can't use the same quotes for both, because that causes confusion.

Current PR is generating this error:
image

stpaultim: @quicksketch or @BWPanda - I got this syntax error by simply trying to access 'admin/structure/menu'.

Does it indicate a lapse in test coverage that this page is not working and the tests did not catch it?

Maybe we should have a test to see if this page is accessible at it's normal path?

stpaultim: @quicksketch and @BWPanda - I take it back, I saw green and thought the tests had run. But, now I look closer and it appears that tests did not run at all. That is a different problem.

ghost: Tests probably didn't run because of the syntax error...

ghost:
```suggestion
    '#description' => t("When displayed in a 'hierarchial tree' menu style, children of expanded menu links will always be visible."),
```

ghost:
Ah, as this is a new contributor, I had to manually give access for the tests to run. Done now.

ghost:
Thanks, merged into 1.x and 1.21.x.

PR #4011, opened 2022-03-23
Issue #5561: Clean complete CSS class instead of subclasses

### LINKED ISSUE #5561: Fixed: `_views_tokenized_clean_css_identifier()` should clean the complete class.
## Description of the bug

The function [_views_tokenized_clean_css_identifier](https://docs.backdropcms.org/api/backdrop/core%21modules%21views%21views.module/function/_views_tokenized_clean_css_identifier/1)($class, $plugin, $row_index) splits classes into string and token parts, replaces the tokens with their values and then (wrongly) cleans the separated parts with [backdrop_clean_css_identifier](https://docs.backdropcms.org/api/backdrop/core%21includes%21common.inc/function/backdrop_clean_css_identifier/1). This can result in unexpected replacements, e.g. if the token [nid] is replaced with a node id (e.g. "123"), then this part would be "cleaned" to "_23". 
This breaks functionality of e.g. contrib module ip_geoloc which relies on a row class ".sync-id-[nid]". This contrib module is half-ported on https://github.com/backdrop-contrib/ip_geoloc. I'm trying to make it work, but this bug is preventing it.

## Steps To Reproduce

To reproduce the behavior:
1. Create a View from content nodes with Table display
2. Add the nid field
3. In Table Settings, add a row class ".sync-id-[nid]"
4. Check the HTML output of the table

## Actual behavior
The row classes of the table are not correct, e.g. for nid = 123 the row class is ".sync-id-_23"

## Expected behavior
The row classes of the table should be correct, e.g. for nid = 123 the row class should be ".sync-id-123"

## Additional information

I suggest to correct the function _views_tokenized_clean_css_identifier in views.module like this:

```
function _views_tokenized_clean_css_identifier($class, $plugin, $row_index) {
  if (!empty($class)) {
    // To support a class that is part hand-typed and part token, split out tokens.
    $subclasses = preg_split('/(\[[^\]]*])/i', $class, -1, PREG_SPLIT_NO_EMPTY | PREG_SPLIT_DELIM_CAPTURE);
    foreach ($subclasses as &$subclass) {
      $subclass = $plugin->tokenize_value($subclass, $row_index);
    }
    $class = implode('', $subclasses);
    return backdrop_clean_css_identifier($class);
  }
}
```

AlexHoebart-ICPDR:
Fixes https://github.com/backdrop/backdrop-issues/issues/5561

### DIFF
--- a/core/modules/views/views.module
+++ b/core/modules/views/views.module
@@ -2264,18 +2264,9 @@ function _views_tokenized_clean_css_identifier($class, $plugin, $row_index) {
     // To support a class that is part hand-typed and part token, split out tokens.
     $subclasses = preg_split('/(\[[^\]]*])/i', $class, -1, PREG_SPLIT_NO_EMPTY | PREG_SPLIT_DELIM_CAPTURE);
     foreach ($subclasses as &$subclass) {
-      $tokenized = $plugin->tokenize_value($subclass, $row_index);
-      if ($tokenized == $subclass) {
-        // No token replacement, all valid CSS characters allowed.
-        $subclass = backdrop_clean_css_identifier($tokenized, array());
-      }
-      else {
-        // Tokens that are replaced should be additionally sanitized, such
-        // as replacing underscores with hyphens.
-        $subclass = backdrop_clean_css_identifier($tokenized);
-      }
-      $class = implode('', $subclasses);
+      $subclass = $plugin->tokenize_value($subclass, $row_index);
     }
-    return $class;
+    $class = implode('', $subclasses);
+    return backdrop_clean_css_identifier($class);
   }
 }
--- a/core/modules/views/views.module
+++ b/core/modules/views/views.module
@@ -2262,11 +2262,23 @@ function views_autoload_info() {
 function _views_tokenized_clean_css_identifier($class, $plugin, $row_index) {
   if (!empty($class)) {
     // To support a class that is part hand-typed and part token, split out tokens.
-    $subclasses = preg_split('/(\[[^\]]*])/i', $class, -1, PREG_SPLIT_NO_EMPTY | PREG_SPLIT_DELIM_CAPTURE);
-    foreach ($subclasses as &$subclass) {
-      $subclass = $plugin->tokenize_value($subclass, $row_index);
+    $substrings = preg_split('/(\[[^\]]*])/i', $class, -1, PREG_SPLIT_NO_EMPTY | PREG_SPLIT_DELIM_CAPTURE);
+    foreach ($substrings as &$substring) {
+      $tokenized = $plugin->tokenize_value($substring, $row_index);
+      if ($tokenized == $substring) {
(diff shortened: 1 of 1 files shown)


core/modules/views/views.module
```diff
@@ -2262,20 +2262,23 @@ function views_autoload_info() {
...
+      $tokenized = $plugin->tokenize_value($substring, $row_index);
+      if ($tokenized == $substring) {
+        // No token replacement, keep entered string as it is.
+        $substring = $tokenized;
       }
       else {
         // Tokens that are replaced should be additionally sanitized, such
         // as replacing underscores with hyphens.
-        $subclass = backdrop_clean_css_identifier($tokenized);
+        // To avoid unexpected replacement at the start of this substring, we prefix two hyphens and remove them again.
```
indigoxela: Comments aren't supposed to be longer than 80 chars, so this comment needs a line break.

core/modules/views/views.module
```diff
@@ -2262,20 +2262,23 @@ function views_autoload_info() {
...
+        $substring = substr(backdrop_clean_css_identifier("--".$tokenized),2);
```
indigoxela: When concatenating string, the `.` char is supposed to be wrapped in single space.
```
'--' . $tokenized
```
As the placeholder contains no variables, it can get wrapped in `'` instead of `"`

core/modules/views/views.module
```diff
@@ -2262,20 +2262,23 @@ function views_autoload_info() {
...
       }
-      $class = implode('', $subclasses);
     }
-    return $class;
+    // Merge strings into class identifier
```
indigoxela: Comments are supposed to end with proper punctuation.
```
// Merge strings into class identifier.
```

core/modules/views/views.module
```diff
@@ -2262,20 +2262,23 @@ function views_autoload_info() {
...
+    $class = implode('', $substrings);
+    // Returned cleaned identifier, all valid CSS characters allowed
```
indigoxela: Also here. :wink:

core/modules/views/views.module
indigoxela: Oh, and another space is missing after the comma.

indigoxela:
Turn it off and on again.

AlexHoebart-ICPDR:
What is the problem with 8.1 and why is this request for fixing a bug resting in peace?

indigoxela:
> What is the problem with 8.1 and why is this request for fixing a bug resting in peace?

Nothings wrong with 8.1 - it seems to be a random failure.

The PR isn't "resting in peace" it's just pending review. :wink: Sometimes it needs a little patience, as we all do our Backdrop work in our spare time and we all have day jobs.

docwilmot:
Rebooting

quicksketch:
Merged into 1.x and 1.23.x. Thanks folks!

PR #4013, opened 2022-03-24
Issue #4364: Convert Drupal's error level integers to Backdrop's strings.

### LINKED ISSUE #4364: Fixed: Drupal 7 upgrade path must include setting for 'Error messages to display'.
In D7, the 'Error messages to display' setting (at `/admin/config/development/logging`) has values of 0, 1 & 2 (for 'None', 'Errors...' & 'All...' respectively). However in Backdrop these values have changed from integers to strings: 'hide', 'some' & 'all' respectively. That means that upgrading from D7 to Backdrop causes this setting to not be defined (at least in the UI):

errors_radios

We should add upgrade code that converts the integer values to the new string values.

ghost:
Fixes https://github.com/backdrop/backdrop-issues/issues/4364

### DIFF
--- a/core/modules/system/system.install
+++ b/core/modules/system/system.install
@@ -1946,8 +1946,10 @@ function system_update_1017() {
   switch ((int) update_variable_get('error_level', '2')) {
     case 0:
       $error_level = 'hide';
+      break;
     case 1:
       $error_level = 'some';
+      break;
     default:
       $error_level = 'all';
   }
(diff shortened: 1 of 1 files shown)


core/modules/system/system.install
```diff
@@ -1942,13 +1942,23 @@ function system_update_1016() {
  * Moves system logging settings from variables to config.
  */
 function system_update_1017() {
+  // Convert error levels.
+  switch ((int) update_variable_get('error_level', '2')) {
+    case 0:
+      $error_level = 'hide';
```
indigoxela, line 1948: Ahem... you forgot to add `break`, so this code always ends up with "all" :grin:

ghost: Oops, thanks!

quicksketch:

PR #4016, opened 2022-03-24
Issue #4482: Re-word 'Cache dates' field description.

### LINKED ISSUE #4482: 'Cache dates' setting description needs re-wording
Date fields have a checkbox setting called 'Cache dates' with the following description:

To me when reading this, I read 'as' as 'because', which means the sentance doesn't make sense. It took me a while to realise that 'as' means 'when', in which case it does make sense.

Having said that, the sentance is too long and a bit confusing in my opinion, so I believe we could re-word it to avoid future confusion.

Suggestions:

"Date objects can be created and cached when date fields are loaded rather than when they are displayed to improve performance."

"To improve performance, date objects can be created and cached when date fields are loaded (rather than when they are displayed)."

ghost:
Fixes https://github.com/backdrop/backdrop-issues/issues/4482

### DIFF
--- a/core/modules/date/date.admin.inc
+++ b/core/modules/date/date.admin.inc
@@ -557,7 +557,7 @@ function _date_field_settings_form($field, $instance, $has_data) {
   $form['cache_enabled'] = array(
     '#type' => 'checkbox',
     '#title' => t('Cache dates'),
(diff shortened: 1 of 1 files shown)


ghost:
Closing in favour of https://github.com/backdrop/backdrop/pull/4028

PR #4024, opened 2022-03-27
Issue #5567: Make admin bar's search component rely on menu component.

### LINKED ISSUE #5567: Fixed: Admin bar's 'search' component doesn't work if 'menu' component is disabled.
## Description of the bug

If you disable the 'Management menu' component of the Admin Bar, the 'Search bar' component no longer works. This was discovered in https://github.com/backdrop/backdrop-issues/issues/220

image

## Steps To Reproduce

To reproduce the behavior:
1. Go to `/admin/config/administration/admin-bar`
2. Make sure all components are enabled
3. Use the menu search to see that/how it works
4. Disable the Management menu component
5. Try the search again

## Actual behavior

Nothing happens - no drop-down list of results, and pressing 'Enter' doesn't show results either

## Expected behavior

Drop-down list of results show below search bar and/or pressing 'Enter' shows page of results.

## Proposed solution

Making the search work independently of the menu would (I imagine) be a massive undertaking, so as a simpler, more immediate solution to avoid this WTF moment, I suggest making the search checkbox only show when the menu checkbox is checked (using `#states`).

ghost:
Fixes https://github.com/backdrop/backdrop-issues/issues/5567

### DIFF
--- a/core/modules/admin_bar/admin_bar.inc
+++ b/core/modules/admin_bar/admin_bar.inc
@@ -752,11 +752,11 @@ function admin_bar_theme_settings() {
  * Submit handler for admin_bar_theme_settings().
  */
 function admin_bar_theme_settings_submit($form, &$form_state) {
-  $components = $form_state['values']['components'];
+  $components = &$form_state['values']['components'];
 
   // Disable the search bar if the management menu is also disabled.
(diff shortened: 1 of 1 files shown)


quicksketch:

PR #4028, opened 2022-03-30
Issue #4249: Date Field help text in UI

### LINKED ISSUE #4249: [UX] Fixed: Add better help text for Date fields.
It would certainly use some punctuation, and/or be split into two sentences or shortened:

...something like this perhaps:

> Create and cache date objects when date fields are being loaded, rather than when they are being displayed. This improves performance.

---
PR by @klonos: https://github.com/backdrop/backdrop/pull/3027

sudipto68:
Fixes: https://github.com/backdrop/backdrop-issues/issues/4249

### DIFF
--- a/core/modules/date/date.admin.inc
+++ b/core/modules/date/date.admin.inc
@@ -394,27 +394,25 @@ function _date_field_widget_settings_form($field, $instance) {
     '#fieldset' => 'date_format',
     '#weight' => 9,
   );
-  if (in_array($widget['type'], array('date_select'))) {
-    $options = array(
-      'above' => t('Above'),
-      'within' => t('Within'),
-      'none' => t('None'),
-    );
+  
+  $options = array(
+    'above' => t('Above'),
+    'none' => t('Hidden'),
+  ); 
+  $description = t('The location of date part labels, like "Year", "Month", or "Day" . "Above" displays the label as titles above each date part. "Hidden" doesn\'t visually label any of the date parts.');
+  if ($widget['type'] == 'date_select') {
+    $options['within'] = t('Within');
+    $options['none'] = t('None');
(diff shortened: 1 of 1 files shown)


ghost:
I think this whole section can be simplified:

```php
if (in_array($widget['type'], array('date_select'))) {
  $options = array(
    'above' => t('Above'),
    'within' => t('Within'),
    'none' => t('None'),
  );
  $description = t("The location of date part labels, like 'Year', 'Month', or 'Day'. 'Above' displays the label as titles above each date part. 'Within' inserts the label as the first option in the select list and in blank textfields. 'None' doesn't visually label any of the date parts.");
}
else {
  $options = array(
    'above' => t('Above'),
    'none' => t('Hidden'),
  );
  $description = t('The location of date part labels, like "Year", "Month", or "Day" . "Above" displays the label as titles above each date part. "Hidden" doesn\'t visually label any of the date parts.');
}
```

to something like this:
```php
$options = array(
  'above' => t('Above'),
  'within' => t('Within'),
  'none' => t('Hidden'),
);
if ($widget['type'] == 'date_select') {
  unset($options['within']);
}
```

stpaultim:
@BWPanda - @sudipto68 updated his PR based upon your feedback.

BWPanda:
@sudipto68 There's still too much code duplication in my opinion. Please see my suggestion above. I don't think we need to worry about changing the name of the `none` option, and you've still got the `$description` variable in there which isn't being used anymore.

ghost (changes requested):
@sudipto68 There's still too much code duplication in my opinion. Please see my suggestion above. I don't think we need to worry about changing the name of the `none` option, and you've still got the `$description` variable in there which isn't being used anymore.

stpaultim:
@BWPanda - Sorry, I should have noticed that the description field is no longer necessary. I missed that point. 

@sudipto68 @zubaer56 - 

1) Remove the old description fields and unecessary name change of the `none` field. Which means, that I think we can use the code provided by @BWPanda pretty much exactly as he gave it to us, despite our previous confusion on this point.

PR #4029, opened 2022-03-31
Issue #4284: Fix auto-upload with multiple file fields on the same form

### LINKED ISSUE #4284: With multiple image fields, only one thumbnail shows up per save
**Description of the bug**
If a content type has more than one image field. When a user goes to create a new node, only one of the images fields displays the thumbnail, per save. Each time I have the node, I can load one more image and see the thumbnail immediately, along with the thumbnails for all previously save images. 

NOTE: Even when thumbnails are not visible, they image is loaded upon save and the thumbnail will be visible upon editing the node. 

HOWEVER, on a field that allows more than one image per field, you can not add additional images until the thumbnail shows up. So if you have two images fields, one of which allows multiple images. You must load images to the multiple image field FIRST or you may not be able to enter all your images without saving. 

This issue based upon report in the forum: https://forum.backdropcms.org/forum/content-type-two-image-fields

**Steps To Reproduce**
To reproduce the behavior:
1. Create a new content type with three image fields. On any one of the images fields, allow unlimited images for that field. 
2. Create a new node and upload one image to each image field. Note that whichever field you use to upload your image first, will work as expected.  However, for the second and third fields, the thumbnail will not display. 

If you try to enter images to the multi-image image field AFTER adding an image to another field, the thumbnail will not display and you will only be able to one image.

3. Save your node. Notice, that all images are saved (even if there was no thumbnail). 
4. Edit node and notice that thumbnails are visible for all images. But, that if you try to add new images, the thumbnail will only show up in the first field you edit. 
5. Repeat. 

**Expected behavior**
I expect that I can upload all the images at once and that I will see all the image thumbnails before saving the node. 

**Additional information**
I tested this on a fresh version of Backdrop CMS 1.15.0 with no additional modules or customizations. 

Create_Test___Backdrop

Create_Test___Backdrop

indigoxela:
Fixes https://github.com/backdrop/backdrop-issues/issues/4284

It might be that the correct fix would be to remove that whole detach from fileUploadChange.

### DIFF
--- a/core/modules/file/js/file.js
+++ b/core/modules/file/js/file.js
@@ -19,13 +19,6 @@ Backdrop.behaviors.fileUploadChange = {
     $(context).find('input[data-file-auto-upload]').once('auto-upload').on('change', Backdrop.file.autoUpload).each(function() {
       $(this).closest('.form-item').find('.file-upload-button').hide();
     });
-  },
-  detach: function (context, settings, trigger) {
-    if (trigger === 'serialize') {
-      return;
-    }
-    $(context).find('input[data-file-extensions]').off('change', Backdrop.file.validateExtension);
-    $(context).find('input[data-file-auto-upload]').off('change', Backdrop.file.autoUpload);
   }
 };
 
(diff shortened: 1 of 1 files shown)


quicksketch:
Thanks @indigoxela! Merged into 1.x and 1.21.x.

PR #4042, opened 2022-04-08
Issue 5580. Correctly redirect field UI delete button

### LINKED ISSUE #5580: 'Page not found' when trying to delete field
## Description of the bug

If you're on the configuration page for a field that's attached to an entity other than a node (e.g. user account, taxonomy term, file entity), and you click the "Delete field" button at the bottom of the page, you'll get a "Page not found" error.

Doing the same thing for node fields works fine however, as does deleting any entity type's field via the drop-button on the Manage Fields page.

## Steps To Reproduce

To reproduce the behavior:
1. Add a field to a non-node entity (e.g. user account or taxonomy term)
2. On the Manage Fields page for that entity, expand the drop-button for the field and click "Delete"
3. Note that you're on the confirmation form 👍🏻 
4. Go back to the Manage Fields page and instead click "Configure" for the field
5. Scroll down and click the "Delete field" button

## Actual behavior

You get a "Page not found" error. 👎🏻 

## Expected behavior

You go to the confirmation form.

## Additional information

This looks to be a bug in the `field_ui_field_edit_form_delete_submit()` function, where the redirect to the confirmation form is hard-coded to use `admin/structure/types/manage/...` (i.e. the node type pages). I'm not sure how you'd go about specifying a different URL based on the entity type...

argiepiano:
Fixes https://github.com/backdrop/backdrop-issues/issues/5580

### DIFF
--- a/core/modules/field_ui/field_ui.admin.inc
+++ b/core/modules/field_ui/field_ui.admin.inc
@@ -2066,7 +2066,11 @@ function field_ui_field_edit_form_delete_submit($form, &$form_state) {
     unset($_GET['destination']);
   }
   $instance = $form['#instance'];
-  $form_state['redirect'] = array('admin/structure/types/manage/' . $instance['bundle'] . '/fields/' . $instance['field_name'] . '/delete', array('query' => $destination));
+  $entity_type = $instance['entity_type'];
+  $bundle = field_extract_bundle($entity_type, $instance['bundle']);
+  $admin_path = _field_ui_bundle_admin_path($entity_type, $bundle);
+  $admin_field_path = $admin_path . '/fields/' . $instance['field_name'] . '/delete';
+  $form_state['redirect'] = array($admin_field_path, array('query' => $destination));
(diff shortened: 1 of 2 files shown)


ghost:
@argiepiano Two things:
1. You still have the 2nd instance field code in the test. That's no longer necessary since adding the field to the user account acts as the 2nd instance. I'd suggest removing the 2nd field instance on the content type code.
2. I don't think we need a whole separate function to use/test the Delete Field button. Just add a test that clicks the button and confirms the form loads (i.e. we don't get the current 404), then proceed to delete the field as normal.

argiepiano:
Those have been addressed.

argiepiano:
Thank you, but I could have made those changes myself had you suggested them. Now, the question is, since you are committing changes, do we need someone else to review and test them?

argiepiano:
Also, now we have some failing tests after your changes...

quicksketch:

PR #4047, opened 2022-04-17
Issue 5589. Warn if permissions need rebuild

### LINKED ISSUE #5589: Backdrop fails to alert admin user to rebuild content access permissions
## Description of the bug

Backdrop fails to alert the admin users that content access permissions need to be rebuilt. This causes issues for modules that implement `hook_node_grants()` to control access. 

D7 includes a constant error message `The content access permissions need to be rebuilt.` on every page refresh after a new module that implement `hook_node_grants()` is enabled (like OG). This assures that the admin will (hopefully) at some point rebuild access permissions.

Backdrop does not show this warning; the admin in fact may never know that access permissions need to be rebuilt.

## Steps To Reproduce

Typically you would need to install a module that implements `hook_node_grants()` to see this problem (like Organic Groups). A hacky shortcut is the following:

(You can try this both in Backdrop and D7)

1. Install Devel, and use Execute PHP Code to run `node_access_needs_rebuild(TRUE)`. This will set the flag that a rebuilt is needed
2. Visit any admin page with the admin account. You can see a warning to rebuild in D7. No warning is shown in Backdrop. D7 shows the attached. 
3. You can manually check that the flag is still "on" by running `node_access_needs_rebuild()` with an empty parameter.
4. Rebuild permissions by visiting `admin/reports/status/rebuild`
5. Manually check that the flag has been turned off by running `node_access_needs_rebuild()` with an empty parameter.

## Actual behavior
Backdrop does not show any warning when the flag is "on"

## Expected behavior
Backdrop should warn the admin that a rebuilt is needed.

## Additional information

Note that neither cache clearing, nor cron running will ever rebuild the permissions. This HAS to be done manually both in B and D7. I believe this is so because this rebuilding can take a very long time if you have 1000s of nodes.

Possible solution:
1. Add the annoying nagging to Backdrop core so that admins know that permissions need to be rebuilt. This could be done through `hook_init()` to check 
2. Make rebuilding automatic on cache clearing (not ideal, as rebuilding can take a LONG time if you have 1000s of nodes)

Note: in D7 this nagging message was shown through `hook_help()`, which was removed in Backdrop. 

Screen Shot 2022-04-17 at 10 40 36 AM

argiepiano:
Fixes https://github.com/backdrop/backdrop-issues/issues/5589

### DIFF
--- a/core/modules/node/node.module
+++ b/core/modules/node/node.module
@@ -3514,15 +3514,16 @@ function node_requirements($phase) {
     $value .= ' (' . l(t('Rebuild permissions'), 'admin/reports/status/rebuild') . ')';
 
     if ($needs_rebuild) {
-      $description = t('Node access permissions need to be rebuilt. ');  
+      $description = t('Node access permissions need to be rebuilt.');  
       $severity = REQUIREMENT_ERROR;
-    } else {
-      $description = t('If the site is experiencing problems with permissions to content, you may have to rebuild the permissions cache. ');
+    } 
+    else {
+      $description = t('If the site is experiencing problems with permissions to content, you may have to rebuild the permissions cache.');
       $severity = REQUIREMENT_OK;
     }
 
     // Add further information.
-    $description .= t('Rebuilding will remove all privileges to content and replace them with permissions based on the current modules and settings. Rebuilding may take some time if there is a lot of content or complex permission settings. After rebuilding has completed, content will automatically use the new permissions.');
+    $description .= ' ' . t('Rebuilding will remove all privileges to content and replace them with permissions based on the current modules and settings. Rebuilding may take some time if there is a lot of content or complex permission settings. After rebuilding has completed, content will automatically use the new permissions.');
 
     $requirements['node_access'] = array(
       'title' => t('Node Access Permissions'),
--- a/core/modules/node/node.module
+++ b/core/modules/node/node.module
@@ -3495,23 +3495,40 @@ function node_action_info() {
 function node_requirements($phase) {
   $requirements = array();
   if ($phase === 'runtime') {
-    // Only show rebuild button if there are either 0, or 2 or more, rows
-    // in the {node_access} table, or if there are modules that
-    // implement hook_node_grants().
+    $needs_rebuild = node_access_needs_rebuild();
     $grant_count = db_query('SELECT COUNT(*) FROM {node_access}')->fetchField();
-    if ($grant_count != 1 || count(module_implements('node_grants')) > 0) {
+
+    // Show an error message if permissions need to be rebuilt. Otherwise
+    // show a permission count if there are either 0, or 2 or more, rows in the
+    // {node_access} table, or if there are modules that implement
+    // hook_node_grants().
+    if ($needs_rebuild) {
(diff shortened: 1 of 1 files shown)


core/modules/node/node.module
```diff
@@ -3495,23 +3495,40 @@ function node_action_info() {
...
       $value = format_plural($grant_count, 'One permission in use', '@count permissions in use', array('@count' => $grant_count));
     }
     else {
       $value = t('Disabled');
     }
     $value .= ' (' . l(t('Rebuild permissions'), 'admin/reports/status/rebuild') . ')';
-    $description = t('If the site is experiencing problems with permissions to content, you may have to rebuild the permissions cache. Rebuilding will remove all privileges to content and replace them with permissions based on the current modules and settings. Rebuilding may take some time if there is a lot of content or complex permission settings. After rebuilding has completed, content will automatically use the new permissions.');
+
+    if ($needs_rebuild) {
+      $description = t('Node access permissions need to be rebuilt. ');
```
indigoxela: Adding that white space (separator) as part of the translatable string is a bit fragile. Better add it explicitly when you concat the string below.

core/modules/node/node.module
```diff
@@ -3495,23 +3495,40 @@ function node_action_info() {
...
     else {
       $value = t('Disabled');
     }
     $value .= ' (' . l(t('Rebuild permissions'), 'admin/reports/status/rebuild') . ')';
-    $description = t('If the site is experiencing problems with permissions to content, you may have to rebuild the permissions cache. Rebuilding will remove all privileges to content and replace them with permissions based on the current modules and settings. Rebuilding may take some time if there is a lot of content or complex permission settings. After rebuilding has completed, content will automatically use the new permissions.');
+
+    if ($needs_rebuild) {
+      $description = t('Node access permissions need to be rebuilt. ');  
+      $severity = REQUIREMENT_ERROR;
+    } else {
```
indigoxela: That "else" should better go to a new line (see an example above).

core/modules/node/node.module
argiepiano: Thanks

core/modules/node/node.module
```diff
@@ -3495,23 +3495,41 @@ function node_action_info() {
...
-    if ($grant_count != 1 || count(module_implements('node_grants')) > 0) {
+
+    // Show an error message if permissions need to be rebuilt. Otherwise
+    // show a permission count if there are either 0, or 2 or more, rows in the
+    // {node_access} table, or if there are modules that implement
+    // hook_node_grants().
+    if ($needs_rebuild) {
+      $value = t('Access permissions need to be rebuilt.');
+    }
+    elseif ($grant_count != 1 || count(module_implements('node_grants')) > 0 || node_access_needs_rebuild()) {
```
docwilmot: `node_access_needs_rebuild()` is already set to a variable above.

argiepiano: In fact the last statement in the OR should be removed, as it's already checked in the if above. Thanks. I'll push a new commit

quicksketch:
Thanks folks! Merged into 1.x and 1.21.x.

PR #4053, opened 2022-04-26
Issue #5597 - changing default path pattern for cards

### LINKED ISSUE #5597: Remove the path pattern for Cards
## Description of the need

Cards were recently added to core, at least in part to provide a content type that will serve for hidden-path nodes (https://github.com/backdrop/backdrop-issues/issues/4903). Yet, by default all nodes get a default path assigned to them. It might be considered confusing to be assigning a special path for "path-less" nodes.

> Anyway, Backdrop comes with the Default URL alias pattern [node:content-type]/[node:title] which applies to all content types without specific patterns. So, to remove the pattern for cards, we had to remove the default pattern, and that's doesn't sound like a good idea to me.

We currently have this path.settings, despite the fact that Cards would get the path anyway - for the sake of consistency. 

```
  config('path.settings')
    ->set('node_page_pattern', '[node:title]')
    ->set('node_card_pattern', 'cards/[node:title]')
    ->save();
```

## Proposed solution

We could change this to 

```
  config('path.settings')
    ->set('node_page_pattern', '[node:title]')
    ->set('node_card_pattern', 'node/[node:nid]')
    ->save();
```

In essence, this should override the default pattern for all nodes. 

We could also decide to remove the default pattern for all new content types.

stpaultim:
Fixes: https://github.com/backdrop/backdrop-issues/issues/5597

### DIFF
--- a/core/profiles/standard/standard.install
+++ b/core/profiles/standard/standard.install
@@ -191,7 +191,7 @@ function standard_install() {
   config('path.settings')
     ->set('node_page_pattern', '[node:title]')
     ->set('node_post_pattern', 'posts/[node:title]')
-    ->set('node_card_pattern', 'cards/[node:title]')
+    ->set('node_card_pattern', 'node/[node:nid]')
     ->save();
 
   cache_clear_all();


core/profiles/standard/standard.install
```diff
@@ -191,7 +191,7 @@ function standard_install() {
   config('path.settings')
     ->set('node_page_pattern', '[node:title]')
     ->set('node_post_pattern', 'posts/[node:title]')
-    ->set('node_card_pattern', 'cards/[node:title]')
+    ->set('node_card_pattern', 'node/[node:nid]')
```
quicksketch, line 194: If we're going to remove the pattern, we should simply delete this line. Otherwise we'll just be making aliases to the original path, a confusing and unnecessary alias.

stpaultim: @quicksketch - removing that line does nothing. We have a default path pattern for all new content types. 

Line 193 does nothing, neither does line 194. @jenlampton suggested we remove line 194 to remove the pattern, but I left it in the PR because removing it does nothing except create an inconsistency in how we are treating posts and cards. 

To the best of my knowledge, to remove the default pattern for cards we either need to:

1) remove the default pattern for all content types OR
2) override as I did in this PR

stpaultim: Unless you have a better idea.

quicksketch: I think line 193 effectively changes the pattern from "post/[node:title]" to "posts/[node:title]" (pluralizing the path).

But you're right that because we have a default node pattern, all nodes will get a pattern even if we remove this line. I tested this, and sure enough, no new alias gets created (though I'm not sure how it works that way). It seems to work, so I retract my comment. I don't know of any other way to prevent the default from taking effect.

quicksketch:
Thanks @stpaultim! Merged into 1.x for 1.22.0.

PR #4056, opened 2022-04-29
Issue 5603. Fix views_handler_field_bulk_form

### LINKED ISSUE #5603: Bulk Operations views handler shows wrong actions when first adding a BO field
EDIT: In addition to the bug shown below, which I reported originally, please scroll down to see other bugs in this handler. I will provide a PR that tackles this and the other bugs.

## Description of the bug

When adding a Bulk Operation views field for an entity obtained through a "relationship" to the main one of the View, Bulk Operations shows the wrong actions. It shows the actions for the base entity, not for the related entity 

## Steps To Reproduce

1. Create a View with for a Content (Node) base entity
2. Add a relationship to `Content: Author`. Add a field from the user entity such as username
3. Add a Bulk Operation field for the User account entity

The actions shown in the "Configure field" modal for this Bulk Operations belong to the Node entity, not to the User one. 

## Actual behavior
The actions shown in the "Configure field" modal for this user Bulk Operations belong to Content (Node), not to User account. 

## Expected behavior

The actions for User account should be shown.

## Additional information
It seems like the method `views_handler_field_bulk_form::get_entity_type()` contemplated the ability to add a BO for fields available through relationships: 

```php
    // If the current field is under a relationship you can't be sure that the
    // base table of the view is the base table of the current field.
    // For example a field from a node author on a node view does have users as base table.
```

Obviously this is not working, as `$this->options['relationship']` is still set to `none` .
 

- Backdrop CMS version: 1.22.x-dev

argiepiano:
Fixes https://github.com/backdrop/backdrop-issues/issues/5603


core/modules/views/handlers/views_handler_field_bulk_form.inc
```diff
@@ -217,32 +222,4 @@ class views_handler_field_bulk_form extends views_handler_field {
 
     return $options;
   }
-
-  /**
-   * Determine the base table of the bulk operation field.
-   */
-  protected function get_entity_type() {
```
indigoxela, line 224: This removal is a low risk change, as probably nothing extends views_handler_field_bulk_form.

But for the sake of BC, can you restore it and let it just return `$this->entity_type;`?

core/modules/views/handlers/views_handler_field_bulk_form.inc
```diff
@@ -98,13 +98,19 @@ class views_handler_field_bulk_form extends views_handler_field {
     // Render checkboxes for all rows.
     $form[$this->options['id']]['#tree'] = TRUE;
     foreach ($this->view->result as $row_index => $row) {
+      // Store the row index to retrieve the entity from the results.
+      $this->view->row_index = $row_index;
```
indigoxela, line 102: Maybe a silly question, but why is this necessary?

argiepiano: This is needed because `views_handler_field_entity:get_value()` needs to know the row index in order to retrieve the value of the ID field for a particular entity. That row index points at the correct row in the results of the query. `get_value()` is invoked in `views_handler_bulk_form:views_form()` in order to build the checkboxes with the correct ID of the real field.

argiepiano: A bit more explanation - there is no assurance that the $values variable actually contains the value of $field when invoking `views_handler_field_entity:get_value()` - because the value of `real_field` may or may not have been included in the query. Therefore get_value needs to retrieve the actual entity (which was cache during  `views_handler_field_entity:pre_render()` in order to obtain the value of that field from the entity itself.

indigoxela: I see (sort of) :wink:

What baffled me was that inside a foreach the same property gets overridden again and again.

I'll probably have to locally stuff the code with debug()s - hopefully then I'll understand.

argiepiano: > What baffled me was that inside a foreach the same property gets overridden again and again.

Yup, that's how it's supposed to work. It's a way of passing the index of the result's row currently being processed. This is not something I made up - it's used this way by other handlers and style plugins, for example, look at `views_plugin_style_rss::render()`

core/modules/views/handlers/views_handler_field_bulk_form.inc
argiepiano: Thanks. Good suggestion. I've made that change.

indigoxela:
Ouch... this may be unrelated, but the NodeViewsRevisionRelationsTestCase fails. Actually in php 7.4 and 8.1, but it only shows up marked as "failed" in 7.4 (fraction 2/3).

argiepiano:
> Ouch... this may be unrelated, but the NodeViewsRevisionRelationsTestCase fails. Actually in all php versions, but it only shows up marked as "failed" in 7.4 (fraction 2/3).

Interesting. I'll be checking on that later today.

indigoxela:
> Interesting. I'll be checking on that later today.

All-clear - the failure is unrelated and gets already fixed in another PR. :+1:

argiepiano:
Curious: what was the issue and what's the PR that fixed it?

indigoxela:
> Curious: what was the issue and what's the PR that fixed it?

See here: https://backdrop.zulipchat.com/#narrow/stream/218635-Backdrop/topic/Hidden.20fatal.20error.20in.20test.20runs
It gets fixes "drive-by".

argiepiano:
I added a new test for the Bulk Form but that broke a couple of other tests. I will see what's going on.

core/modules/views/tests/handlers/views_handler_field_bulk_form.test
```diff
@@ -0,0 +1,247 @@
+<?php
+
+/**
+ * @file
+ * Definition of ViewsHandlerFieldBulkFormTest.
+ */
+
+require_once BACKDROP_ROOT . '/core/modules/views/tests/views_query.test';
```
indigoxela, line 8: As views_handler_field_bulk_form.test ships with the same module, another option for that require_once would be `__DIR__`.

core/modules/views/tests/views_test_entity/views_test_entity.entity.inc
```diff
@@ -0,0 +1,57 @@
+<?php
+/**
+ * @file
+ * Test entity classes.
+ */
+
+/**
+ * Test class for the test_entity entity type.
+ */
+class BulkTestEntity extends Entity {
```
indigoxela, line 10: `class BulkTestEntity extends Entity` ... strange, that's the same class as above. :thinking:

argiepiano: Oops! Yes, that's wrong. I forgot to remove the class from views_test.entity.inc 😓 

Thanks for catching!

argiepiano: I have removed the file views_test.entity.inc as it not needed. The class definition was moved to views_test_entity.entity.inc (and a new helper module)

core/modules/views/tests/handlers/views_handler_field_bulk_form.test
argiepiano: The issue is that `views_query.test` is not in the same directory as `views_handler_field_bulk_form.test`, but the one above it.  This exact same require_once line is used in many another existing views test, so I prefer not to mess up with it.

argiepiano:
Closing this PR in favor or an alternative one (#4066) that takes a lighter approach to fixing the original issue.

PR #4058, opened 2022-04-30
Issue #5607: Adding a new card image style and adding it to card view

### LINKED ISSUE #5607: Add a new cropped image style specifically for cards.
## Description of the need

With the addition of https://github.com/backdrop/backdrop-issues/issues/4903 into core, there has been some discussion about the need to add a new image style that we can use for our front page view of Cards. Various options have been discussed here:
https://github.com/backdrop/backdrop-issues/issues/5593

Of the options presented, this is the one that has the most traction so far. It would be nice to get this into 1.22.0, since the new cards content type will be so prominent and it would be really nice if the view layout (image alignment) doesn't "break" as people experiment with it. 

## Proposed solution

After discussing this with @quicksketch during our informal sprint today, we've come up with the following proposal to address the following goals.

1) Keep changes to a minimum
2) Future proof the changes as much as possible
3) Name the new image style in a way that helps people understand the difference between cropped and uncropped image styles
4) Provides a meaningful alternative or option to existing image styles

I'm about to submit a PR that does the following. Feedback welcome.

1) Creates a new cropped image style
2) Names it "Cards (600x400 cropped)
3) Machine name = "cards"

https://github.com/backdrop/backdrop-issues/issues/5593

## Additional information

We have deliberately added the word cropped into the label to help distinguish it from existing image styles. This one will behave differently, because it is cropped and the others are not, and we feel as if the label should somehow reflect that.

Sorted alphabetically, we would have a list of image styles that would appear as:

Cards (600x400 cropped)
Large (800x600)
Medium (300x250)
Thumbnail (100x100)

stpaultim:
Fixes: https://github.com/backdrop/backdrop-issues/issues/5607

### DIFF
--- a/core/profiles/standard/config/image.style.card.json
+++ b/core/profiles/standard/config/image.style.card.json
@@ -1,0 +1,19 @@
+{
+    "_config_name": "image.style.card",
+    "storage": 1,
+    "module": "image",
+    "label": "Card (800x533 cropped)",
+    "name": "card",
+    "effects": [
+        {
+            "name": "image_scale_and_crop",
+            "data": {
+                "width": "800",
+                "height": "533",
+                "anchor": "center-top",
+                "upscale": 1
+            },
+            "weight": 1
+        }
+    ]
+}
--- a/core/profiles/standard/config/views.view.promoted_cards.json
+++ b/core/profiles/standard/config/views.view.promoted_cards.json
@@ -195,7 +195,7 @@
                         "click_sort_column": "fid",
                         "type": "image",
                         "settings": {
-                            "image_style": "large",
+                            "image_style": "card",
                             "image_link": "",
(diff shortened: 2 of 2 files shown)


core/modules/image/config/image.style.card.json
```diff
@@ -0,0 +1,19 @@
+{
+    "_config_name": "image.style.card",
+    "storage": 4,
```
quicksketch: Can you change this to 1 (`IMAGE_STORAGE_NORMAL`), and move this entire file to within `profiles/standard/config`? If cards are only provided by the standard profile, this image style should only exist when using it.

quicksketch:
Merged into 1.x for 1.22.0. Thanks @stpaultim!

PR #4059, opened 2022-04-30
Chaning sort order by NID on cards view

### LINKED ISSUE #5608: Reverse NID search order for front page card view
## Description of the bug

We have two sort order rules for the front page view of cards. We word by both date and nid. We did this at least in part because sorting by date alone was not sufficient to get the right order in Tugboat sandboxes. 

Currently, if a user adds a new promoted card it shows up because of the date search, but if the date sort were to fail for some reason - a new card might not show up on the front page - because the NID sort is Ascending. @quicksketch suggested we reverse that search order and then also reverse the order in which the nodes are built. 

This will not effect the display on default sites, but it will make it less likely that a user adds a new promoted card in the future that does not show up on the front page. 

NOTE: If we do this, I am not sure we need the date sort at all. 

This PR was requested by @quicksketch

stpaultim:
Fixes: https://github.com/backdrop/backdrop-issues/issues/5608


quicksketch:
Thanks @stpaultim!

PR #4061, opened 2022-05-01
Issue #4569-1: Hide title on comments

### LINKED ISSUE #4569: Allow admin to hide the title on comments
**Description of the need**

The default situation in Backdrop CMS for comments is to display a title on all comments. If the user does not have the ability to manually create a custom title for the comment (this is an optional setting), then Backdrop CMS simply takes the first part of the first line and generates a title for the comment. 

This latter functionality doesn't seem very helpful to me. I believe that it should be possible to turn off the title on a comment field. 

**Proposed solutions**

We are currently discussing two different approaches towards resolving this issue and have two different PRs. 

4569-1  --  CSS Solution [edit: [PR 4061](https://github.com/backdrop/backdrop/pull/4061)]

This solution primarily relies on CSS and works pretty well with most core AND contrib themes (but not perfectly). 

- Update hook to replaces old comment_subject_field boolean setting with comment_title_options setting (3 choices - auto, custom, hidden)
- If "Hide Titles" is selected, a css class hide-comment-titles is added to the comment titles, it uses CSS to hide comment titles on any theme that is displaying comment titles and replaces them with a SVG "link" icon that acts as a permalink. 
- To accomplish css change in all contrib themes this PR adds a new css file at: core/modules/comment/css/comment_titles.css

4569-2  --  Template Solution [edit: [PR 4068](https://github.com/backdrop/backdrop/pull/4068)]

This solution updates the comment.tpl.php files for core, Bartik, and Basis and works with any contrib themes that inherit those templates/css. BUT, sites that use contrib templates that override comment.tpl.php files would not be able to hide comment titles unless their contrib theme is updated.

- Update hook to replaces old comment_subject_field boolean setting with comment_title_options setting (3 choices - auto, custom, hidden)
- If hidden is selected, it uses the comment.tpl.php file to hide the comment title in core, basis, and bartik.
- It does not have any effect on contrib themes using custom comment.tpl.php or comment.css files. Users will be unable to hide titles on those themes, unless the theme updates to support this feature.
- In BASIS, it uses CSS to add a permalink icon when the titles are hidden (see below).
- In core and Bartik, it keeps the same old permalink link that was there by default.

**Issues to Consider**

- By default comment titles also serve as a permalink. Simply hiding comment titles might leave some sites without any permalinks available for comments. Some core and contrib themes provide separate permalinks and this would not be a problem. 
- Because many contrib themes override the core comment.tpl.php and comment.css files, the first option above creates a NEW css file to make sure this solution works with as many contrib themes as possible.

**Additional information**

See [my comment](https://github.com/backdrop/backdrop-issues/issues/4413#issuecomment-678895796) in #4413 

> I am working on the website for Backdrop CMS Live event and I am running into this problem and it seems REALLY weird.
> 
> -1) "Allow comment title" is turned off (I see the point, that this might just be mislabeled).
> -2) My comments are all being auto populated by a title, which is really just the first line of text from the comment, formatted as a link which really does nothing but take me to the comment I'm already on.
> 
> This title and/or link looks really weird and doesn't serve any purpose that I can imagine. Also, I can't find anyway to hide this in the UI.
> 
> My screen shot comes from a default installation of Backdrop CMS with no changes to config, except enabling comments on this specific node and adding some comments.

Your_first_post____Comments

<hr />

To Do

- [ ] Decide if we should also add a new variable for permalink path. 

<hr />

**Next Steps**

The next step for this issue is to decide on the best approach (see options discussed above). 

Once we decide on an approach, we can work on the details of either PR and address issues with tests.

<hr />

- OLD PR by @VasasA https://github.com/backdrop/backdrop/pull/3280
- ADVOCATE: @stpaultim

stpaultim:
Fixes: https://github.com/backdrop/backdrop-issues/issues/4569


core/modules/comment/comment.theme.inc
```diff
@@ -109,6 +109,11 @@ function template_preprocess_comment(&$variables) {
   $variables['permalink'] = l(t('Permalink'), $uri['path'], $uri['options']);
   $variables['submitted'] = t('!datetime by !username', array('!datetime' => $variables['created'], '!username' => $variables['author']));
 
+  // Check to see if title should be hidden
+  if ($node_type->settings['comment_subject_hide']) {
+    $variables['title'] = '';
```
indigoxela: Note that this will still create an empty `<h3></h3>` tag in the output. Not sure if that matters.

stpaultim: Noted. Looking for ideas on how to avoid that.

core/modules/comment/comment.module
```diff
@@ -1165,6 +1160,20 @@ function comment_form_node_type_form_alter(&$form, $form_state) {
       '#default_value' => $node_type->settings['comment_close_days'],
       '#field_suffix' => t('days'),
     );
+    $form['comment']['titles'] = array(
+      '#markup' => t('<label>Comment titles</label>'),
+    );
+    $form['comment']['comment_subject_field'] = array(
```
docwilmot: I suggest switching the order of these two checkboxes, and hiding the `comment_subject_field` checkbox with states if the `comment_subject_hide` is checked, since it doesnt make much sense otherwise.

stpaultim: I thought about this, but didn't do it because comments still need titles for internal use. HOWEVER, now that I think about it, they are very unlikely to need custom titles for internal use. So this idea does make sense.

@argiepiano pointed out cases in which this could cause confusion if the box is checked and the other setting is hidden.

ghost: @docwilmot I disagree. There may be use-cases for wanting custom titles to be created, but not shown in comments themselves (perhaps they're used in admin views, etc.).

core/modules/comment/comment.module
```diff
@@ -1165,6 +1160,20 @@ function comment_form_node_type_form_alter(&$form, $form_state) {
...
+      '#type' => 'checkbox',
+      '#title' => t('Allow a custom title to be set for each comment'),
+      '#description' => t('When disabled, the comment title will be automatically generated from the comment text.'),
+    );
```
ghost: This is missing the default value: `$node_type->settings['comment_subject_field']`

core/modules/comment/comment.theme.inc
```diff
@@ -106,9 +106,16 @@ function template_preprocess_comment(&$variables) {
   $uri['options'] += array('attributes' => array('class' => array('permalink'), 'rel' => 'bookmark'));
 
   $variables['title'] = l($comment->subject, $uri['path'], $uri['options']);
+  // In case title is hidden, we can use 'created' as permalink
+  $variables['created_permalinked'] = l($variables['created'], $uri['path'], $uri['options']);
```
ghost: What about instead of adding a new parmalink version of the submitted date, we make `$uri['path']` a variable that themes can use to create permalinks of whatever they want? E.g.:

```php
$variables['permalink_path'] = $uri['path'];
```

core/modules/comment/comment.theme.inc
```diff
@@ -106,9 +106,16 @@ function template_preprocess_comment(&$variables) {
...
   $variables['permalink'] = l(t('Permalink'), $uri['path'], $uri['options']);
   $variables['submitted'] = t('!datetime by !username', array('!datetime' => $variables['created'], '!username' => $variables['author']));
 
+  // Check to see if title should be hidden
+  if ($node_type->settings['comment_subject_hide']) {
+    $variables['title'] = "";
+  }
```
ghost: I don't think we should do this. See issue discussion for reasoning.

ghost (dismissed):

core/modules/comment/comment.module
docwilmot: > There may be use-cases for wanting custom titles to be created, but not shown in comments themselves (perhaps they're used in admin views, etc.).

In that case we generate the title as we do when they dont type one. There is no point to a UI where you ask a user to type a tile (in the normal usage, ie a comment thread) and it doesnt show. If you have a fancy UI that requires to show a title somewhere, then dont hide them. 

I quite frankly dont see a reasonable use case.

stpaultim: @docwilmot - I see your point. Yes, if we could disable the custom titles completely when the "Hide titles" option is chosen, that might work. I'll see if I can figure out how to do that. I'm still a beginner at this stuff.

core/modules/system/css/system.css
```diff
@@ -370,3 +370,15 @@ section,
...
 }
+
+.hide_comment_title {
+  display: block;
+  text-indent: -99999px;
+  line-height: 0;
+}
+.hide_comment_title:after {
+  display: block;
+  content: "Permalink";
```
indigoxela: Caution, this won't be translatable.

stpaultim: @indigoxela - Thanks. I suspected already that this is the wrong way to do this. I'm looking for ideas on the right way. Otherwise, I will experiment more later. Just wanted to get the concept working first. Suggestions welcome.

core/modules/comment/comment.module
```diff
@@ -135,7 +151,7 @@ function comment_field_extra_fields() {
   $return = array();
 
   foreach (node_type_get_types() as $type) {
-    if ($type->settings['comment_subject_field']) {
+    if ($type->settings['comment_subject_options'] = 1) {
```
indigoxela: You assign the value `=` instead of comparison `==`?

core/modules/comment/css/comment_titles.css
```diff
@@ -0,0 +1,11 @@
...
+  display: block;
+  text-indent: -99999px;
+  line-height: 0;
+}
+.hide_comment_title:after {
+  display: block;
+  content: "Permalink";
+  text-indent: 0;
+  line-height: initial;
+}
```
indigoxela: Please add a newline after that last item.

core/modules/comment/comment.install
```diff
@@ -470,6 +470,26 @@ function comment_update_1007() {
...
+ * New setting for comment titles. Replacing comment_subject_field
+ * with comment_subject_options.
+ */
+function comment_update_1008() {
+  foreach (config_get_names_with_prefix('node.type.') as $config_name) {
+    $config = config($config_name);
+    $comment_subject = $config->get('settings.comment_subject_field');
+    if ($comment_subject == "0") {
+      // auto generated titles
+      $config->set('settings.comment_subject_options', 1);
```
docwilmot: Use the constant here instead of integer.

core/modules/comment/comment.install
```diff
@@ -470,6 +470,26 @@ function comment_update_1007() {
...
+    } else {
+      // custom titles
+      $config->set('settings.comment_subject_options', 0);
```
docwilmot: Use the constant here instead of integer.

core/modules/comment/comment.module
docwilmot: Use the constant here instead of integer.

core/modules/comment/comment.module
```diff
@@ -1864,7 +1887,7 @@ function comment_form($form, &$form_state, $comment) {
     '#title' => t('Title'),
     '#maxlength' => 64,
     '#default_value' => $comment->subject,
-    '#access' => $node_type->settings['comment_subject_field'],
+    '#access' => $node_type->settings['comment_subject_options'] == "0",
```
docwilmot: Use the constant here instead of integer.

core/modules/comment/comment.theme.inc
```diff
@@ -104,9 +104,14 @@ function template_preprocess_comment(&$variables) {
 
   $uri = $comment->uri();
   $uri['options'] += array('attributes' => array('class' => array('permalink'), 'rel' => 'bookmark'));
-
-  $variables['title'] = l($comment->subject, $uri['path'], $uri['options']);
+  
   $variables['permalink'] = l(t('Permalink'), $uri['path'], $uri['options']);
+  // Check to see if title should be hidden
+  if ($node_type->settings['comment_subject_options'] == 2) {
```
docwilmot: Use the constant here instead of integer.

docwilmot:
I like this new UI idea with the radios. Few areas where could use the new constants rather than writing out the integer value. 
Why is there a "link_icon.png" in the PR though?

core/modules/comment/comment.module
stpaultim: I belive that we've resolved this with @BWPanda idea

core/modules/comment/comment.theme.inc
stpaultim: Agreed

stpaultim:
The icon link is there, because I hoped to use that instead of the text "Permalink." Or at least, I'm considering it. But, I have yet to figure out how to insert the icon as a permalink using css.

core/modules/comment/comment.theme.inc
stpaultim: I added this to discuss as a todo in original issue.

core/modules/comment/comment.module
stpaultim: Done

core/modules/comment/comment.module
```diff
@@ -135,7 +151,7 @@ function comment_field_extra_fields() {
   $return = array();
 
   foreach (node_type_get_types() as $type) {
-    if ($type->settings['comment_subject_field']) {
+    if ($type->settings['comment_subject_options'] == COMMENT_TITLE_AUTO) {
```
indigoxela: Hm. To my understanding comment_subject_field would be same as COMMENT_TITLE_CUSTOM...

core/modules/comment/comment.theme.inc
```diff
@@ -104,9 +104,14 @@ function template_preprocess_comment(&$variables) {
 
   $uri = $comment->uri();
   $uri['options'] += array('attributes' => array('class' => array('permalink'), 'rel' => 'bookmark'));
-
-  $variables['title'] = l($comment->subject, $uri['path'], $uri['options']);
+  
   $variables['permalink'] = l(t('Permalink'), $uri['path'], $uri['options']);
+  // Check to see if title should be hidden
+  if ($node_type->settings['comment_subject_options'] == COMMENT_TITLE_HIDDEN) {
+    $uri['options']['attributes']['class'][] = 'hide_comment_title';
```
indigoxela: CSS classes usually should use hyphens, not underscores.

core/modules/comment/css/comment_titles.css
```diff
@@ -0,0 +1,22 @@
...
+  display: block;
+  /* Looking for a better idea here. */
+  text-indent: -99999px;
+  line-height: 0;
+}
+
+.hide_comment_title:after {
+  display: block;
+  /* Looking for a better/translatable idea here. */
+  content: "Permalink";
```
indigoxela: As the constant name AND help text currently are - there shouldn't be visible text:

COMMENT_TITLE_HIDDEN and t('Comment titles are hidden.') rather indicate, that there's NO visible text at all.

Another suggestion: as the constants are new, we could name them in a less ambiguous way:

COMMENT_SUBJECT_HIDDEN - it's the subject, not actually the title.

core/modules/comment/comment.module
stpaultim: @indigoxela I'm not sure if I understand your point, but maybe this will help explain what I did. That doesn't mean that what I did is correct or optimal.

comment_subject_field was a boolean setting with titles either 'auto generated' = false or 'custom" = true.

I replaced comment_subject_field with a config setting comment_subject_options that has three options.

1) COMMENT_TITLE_CUSTOM
2) COMMENT_TITLE_AUTO
3) COMMENT_TITLE_HIDDEN

Does this help? Am I missing something?

core/modules/comment/comment.theme.inc
stpaultim: I should know better. Fixed.

core/modules/comment/css/comment_titles.css
stpaultim: > As the constant name AND help text currently are - there shouldn't be visible text:

@indigoxela - Good points. The goal is to hide the title. BUT, I'm also trying to make sure we don't loose the permalink entirely.

My plan was to implement your idea that when the title is hidden, to replace it with a permalink icon. BUT, I don't know how to replace the title with an icon that is a link. So in it's current state, the "hidden" status actually replaces the title with the word "permalink." For me, this is an improvement, but not the end goal. Looking for better ideas or help replacing the title with an icon. 

> COMMENT_SUBJECT_HIDDEN - it's the subject, not actually the title.

I need guidance here, because I named the new setting "comment_subject_options" because it is more consistent with the old setting name. BUT, in the UI we don't use the word "subject" we use the term "title." 

1) I'm not sure if subject or title is more accurate and/or user friendly. I prefer title, in which case maybe I should drop "subject" completely OR the opposite. I can see how "subject" might be more accurate, especially when hidden, but I don't think it's as clear to end users.

stpaultim: For now, I'm using "Title" instead of subject - because that is what has been used in the UI for a long time prior to this PR. Happy to discuss alternatives, but for now at least it's consistent.

stpaultim:
Closing this PR, since we seem to be moving in the other direction.

stpaultim:
What this PR does:

1) Changes boolean setting "Comment_subject_field" to radio buttons "comment_title_options" with three choices. Previously the choices were default comment titles or custom comment titles. PR provides for three options "custom", "auto generated," or "hidden."

2) PR defines constants for each choice.

3) In comment.theme.inc we:
  - Create $variables['permalink_path']
  - Check to see if title should be hidden. If "yes" then add a class = 'hide-comment-title'
  - Create $variables['title_options']

4) Add a svg link icon

5) In the Bartik and Basis themes, it modifies the comment.tpl.php file and provides a class when the title is hidden AND does not render the title. In Basis it provides a new permalink icon, in Bartik it does not because there is already a permlink link.

PR #4066, opened 2022-05-06
Issue 5603. Fix views_handler_field_bulk_form

### LINKED ISSUE #5603: Bulk Operations views handler shows wrong actions when first adding a BO field

argiepiano:
Fixes https://github.com/backdrop/backdrop-issues/issues/5603

This is an alternative PR that takes a lighter approach. Instead of changing the class parent to `views_handler_field_entity` and making other modifications to adjust to this parent, it keeps the original parent class `views_handler_field` and modifies the method `get_entity_type` to provide the correct actions for the entity type to which this bulk form field belongs.


core/modules/views/tests/handlers/views_handler_field_bulk_form.test
```diff
@@ -0,0 +1,247 @@
...
+ */
+
+require_once BACKDROP_ROOT . '/core/modules/views/tests/views_query.test';
+
+class ViewsHandlerFieldBulkFormTest extends ViewsSqlTest {
+
+  public $view;
+
+  function setUp() {
+    parent::setUp();
```
indigoxela: This causes `PHP Fatal error:  Declaration of ViewsHandlerFieldBulkFormTest::setUp() must be compatible with ViewsSqlTest::setUp(array $modules = [])` (silent / hidden) problem.

Should probably rather be: `setUp($modules)`.

core/modules/views/handlers/views_handler_field_bulk_form.inc
```diff
@@ -232,6 +237,9 @@ class views_handler_field_bulk_form extends views_handler_field {
       $options = $relationships[$this->options['relationship']];
       $data = views_fetch_data($options['table']);
       $base_table = $data[$options['field']]['relationship']['base'];
+    } else {
```
indigoxela: This "else" should go to a new line ... (deja-vue? :stuck_out_tongue_winking_eye:)

core/modules/views/tests/views_test/views_test.module
```diff
@@ -35,6 +35,7 @@ function views_test_autoload_info() {
     'views_test_plugin_access_test_dynamic' => 'test_plugins/views_test_plugin_access_test_dynamic.inc',
     'views_test_plugin_access_test_static' => 'test_plugins/views_test_plugin_access_test_static.inc',
     'views_test_plugin_style_test_mapping' => 'test_plugins/views_test_plugin_style_test_mapping.inc',
+
   );
```
indigoxela: I guess, this removal of the empty line isn't intentional, but a leftover. If that gets reverted, the file won't show up here anymore.

argiepiano:
Just submitted a new PR with the couple of style fixes you pointed out. Thanks @indigoxela

core/modules/views/handlers/views_handler_field_bulk_form.inc
```diff
@@ -219,10 +219,15 @@ class views_handler_field_bulk_form extends views_handler_field {
...
 
   /**
-   * Determine the base table of the bulk operation field.
+   * Determine the entity type of the bulk operation field.
    */
   protected function get_entity_type() {
-    $base_table = $this->view->base_table;
+    // Bulk operations are tied to a specific entity type that defines them.
+    // If the entity type definition is available, return it.
+    $table_data = views_fetch_data($this->table);
```
indigoxela: It may be because I only tested with user and node, and because of my limited understanding of Views, but for me `$this->table` always returns the correct table for the chosen bulk handler.

You're absolutely right - `$this->view->base_table` is not correct here. But what about  `$this->table`? Wouldn't that alone be sufficient to get the right base_table?

argiepiano: Let me see if I can answer your question. 

This PR solves the problem of the Bulk Form field not presenting the correct actions when we are dealing with a relationship. So, assume that you create a view that shows Content.  Your views base table is `node`. Assume you create a relationship to `user`, for example by adding "Content: Author" in the Relationship setting. Then assume you add a "User account" Bulk Operations field to the view. The current bug is that the handler (before you save it) will show you the actions for the views' base table entity (node) instead of the actions for the related entity (user). 

So, the fix basically consists in retrieving the correct entity for the handler.  What these lines are doing: they use `views_fetch_data` to retrieve the views data for the `user` table (which is the handler's table). The "views data" is what you defined in `user.views.inc` under the function `user_views_data()`. If you look at that function, you'll see that `$data['table']['entity type']` is set to `user`. So, this is an efficient, and foolproof way of getting the correct entity type from the handler table name.

Once you have the correct entity type, then you can retrieve the correct actions.

I must add that in most cases the name of the handler table will be the same name as the entity it belongs to. But this may not be the case in some cases. Therefore it's important to really look at the basic definition of the views data for the given table, to retrieve the property `entity type`.

Let me know if that doesn't make sense...

indigoxela: @argiepiano I see, I have to be clearer in my questions. :wink:

I'm aware that the PR fixes the bug and I'm also aware why it fixes it. But my finding is that $this->table in views_handler_field_bulk_form::get_entity_type() *always* seems to return the correct db table (user, node, comment). So it might be that the whole part in this function that tries to fetch additional data explicitly, if there's a relationship could be redundant.

The base table retrieved via $this->table seems to be sufficient (according to my experiments). My suspicion is that the function code could get simplified. But I might be overlooking something.

indigoxela: Some more digging...

`$this->table` gets provided either by views_handler::init() or by views_handler_relationship::init(). That's why it's always the right table.

So the check could be: if `$this->table`, then the base table is `$this->table`, otherwise it's `$this->view->base_table`.

[This comment](https://github.com/backdrop/backdrop/blob/1.x/core/modules/views/includes/handlers.inc#L121) makes me think that `$this->table` might not always exist. So we should still fall back to $this->view->base_table.

But the fetching of additional data and also the extra check for options['relationship'] don't seem necessary. That would simplify the whole function.

argiepiano: @indigoxela, it IS possible for the table and the entity to have different names, both for core entities and for custom entities. For example, try the following command:
dpm(views_fetch_data('taxonomy_term_data'));

In the case of taxonomy terms, the bulk form handler table would be defined as `taxonomy_term_data`, while the entity type is `taxonomy_term`.

The redundancy is there to prevent those cases. Do you disagree?

indigoxela: > it IS possible for the table and the entity to have different names

Yes, of course. I'm not suggesting to only rely on the table name. That wouldn't even work for user (vs. users). :wink: 

My suggestion is to skip the extra fetch and fiddling with relationship data directly - that's unnecessary.

The part beginning with `foreach (entity_get_info() as $entity_type => $info) {` still *is* necessary, but the views_fetch_data() isn't and neither is `if (!empty($this->options['relationship']) ...`.

indigoxela: Maybe thing are easier to understand, if I provide my suggestion as code. Explaining things can be hard. :wink: 

```
  protected function get_entity_type() {
    $base_table = isset($this->table) ? $this->table : $this->view->base_table;
    foreach (entity_get_info() as $entity_type => $info) {
      if (isset($info['base table']) && $info['base table'] == $base_table) {
        return $entity_type;
      }
      elseif (isset($info['revision table']) && $info['revision table'] == $base_table) {
        $this->revision = TRUE;
        return $entity_type;
      }
    }
    return FALSE;
  }
```
There still might be edge cases, where this wouldn't work as expected. But it's simpler and much easier to read and understand.

argiepiano: @indigoxela the code above is make things clearer now.

OK, I'll make the change and then we can test. More later

argiepiano: @indigoxela I made that change. Can you please test? Thanks

core/modules/views/tests/handlers/views_handler_field_bulk_form.test
```diff
@@ -0,0 +1,247 @@
...
+
+/**
+ * @file
+ * Definition of ViewsHandlerFieldBulkFormTest.
+ */
+
+require_once BACKDROP_ROOT . '/core/modules/views/tests/views_query.test';
+
+class ViewsHandlerFieldBulkFormTest extends ViewsSqlTest {
+
```
indigoxela, line 11: Sorry that I didn't see that earlier. Newly added test classes are supposed to use the lightweight testing profile:

```
protected $profile = 'testing';
```
Not sure if your test needs further adaption (the testing profile is pretty empty, no out-of-the-box content types). But maybe you don't need any.

argiepiano: Thanks @indigoxela. Yeah, this creates an issue as it was relying on the standard installation nodes (although the tests seem to pass, but there are parts that are completely skipped since the view has no results). I wasn't aware of the 'testing' profile. How strict do you think adherence to that rule is? I guess I could manually create two nodes....

argiepiano: The solution was simple. I just created two nodes. I'll push those changes now. Tests are all passing.

quicksketch:
Merged into 1.x and 1.22.x. Great work @argiepiano and @indigoxela!

PR #4067, opened 2022-05-06
Issue 5615. Fix ViewsSqlTest::setUp regression

### LINKED ISSUE #5615: Regression in 1.22.0-preview for views-related tests
## Description of the bug

1.22.0-preview introduced a regression that breaks some tests for Views. All tests that override `ViewsSqlTest:setUp()` are currently failing. 

Version 1.22-0-preview added an argument to ViewsSqlTest::setUp. This creates a problem with all other classes that override ViewsSqlTest:setUp and have not been changed.

There are over 10 classes that have this problem, including `CommentViewsHandlerArgumentUserUidTest`, `UserViewsArgumentValidate`, `ViewsHandlerSortDateTest `, etc.

The tests fail in my local, and produce fatal errors in GHA (see for example [this](https://github.com/backdrop/backdrop/runs/6313809761?check_suite_focus=true)).

## Steps To Reproduce

Create an installation using 1.22.0-preview, enable `simpletest` and run, for example, the test `UserViewsArgumentValidate`.

## Actual behavior

Fatal errors

## Expected behavior

Tests should run normally.

PHP: 7.4

argiepiano:
Fixes https://github.com/backdrop/backdrop-issues/issues/5615

### DIFF
--- a/core/modules/views/tests/views_upgrade.test
+++ b/core/modules/views/tests/views_upgrade.test
@@ -13,10 +13,11 @@ require_once BACKDROP_ROOT . '/core/modules/views/tests/views_query.test';
  * You can find all conversions by searching for "moved to".
  */
 class ViewsUpgradeTestCase extends ViewsSqlTest {
-  protected function setUp() {
-//     // To import a view the user needs use PHP for settings rights, so enable php module.
-    parent::setUp();
+  protected function setUp(array $modules = array()) {
+    parent::setUp($modules);
 
+    // To import a view the user needs use PHP for settings rights, so enable php module.
+    // @TODO The php module does not exist in core in Backdrop!
     module_enable(array('php'));
     $this->resetAll();
   }
--- a/core/modules/comment/tests/comment_views_handler_argument_user_uid.test
+++ b/core/modules/comment/tests/comment_views_handler_argument_user_uid.test
@@ -33,8 +33,8 @@ class CommentViewsHandlerArgumentUserUidTest extends ViewsSqlTest {
     return comment_save($comment);
   }
 
-  function setUp() {
-    parent::setUp();
+  protected function setUp(array $modules = array()) {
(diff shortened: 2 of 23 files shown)


argiepiano:
I think this was a random test failure not related to the PR. Will close and open again to run tests

argiepiano:
Reopening to trigger

argiepiano:
Reopening to re-trigger the Tugboat instance

core/modules/views/tests/views_upgrade.test
```diff
@@ -13,10 +13,11 @@ require_once BACKDROP_ROOT . '/core/modules/views/tests/views_query.test';
...
 class ViewsUpgradeTestCase extends ViewsSqlTest {
-  protected function setUp() {
-//     // To import a view the user needs use PHP for settings rights, so enable php module.
-    parent::setUp();
+  protected function setUp(array $modules = array()) {
+    parent::setUp($modules);
 
+    // To import a view the user needs use PHP for settings rights, so enable php module.
+    // @TODO The php module does not exist in core in Backdrop!
     module_enable(array('php'));
```
indigoxela, line 21: Should we open a follow-up for that? It's unrelated here, but looks like an oversight.

quicksketch: Also there is no Views-specific importer any more, we just use the config-import to import views now, which does not use raw PHP code. Seems this line should be able to be removed? It appears this just fails silently when a module is not found.

indigoxela: New issue opened: https://github.com/backdrop/backdrop-issues/issues/5620

Someone with proper permissions can mark this comment as "resolved".

argiepiano: Thank you, @indigoxela  and @quicksketch.

PR #4068, opened 2022-05-09
Issue #4569-2: Hide title on comments

### LINKED ISSUE #4569: Allow admin to hide the title on comments

stpaultim:
Fixes: https://github.com/backdrop/backdrop-issues/issues/4569


core/themes/basis/templates/comment.tpl.php
```diff
@@ -67,13 +67,20 @@
   <div class="comment-text">
 
     <div class="comment-title">
+      <?php dpm($variables); ?>
```
indigoxela: That `dpm` doesn't belong here. :wink:

core/modules/comment/templates/comment.tpl.php
```diff
@@ -64,8 +64,10 @@
   <?php if ($new): ?>
     <mark class="new"><?php print $new; ?></mark>
   <?php endif; ?>
-  <h3><?php print $title; ?></h3>
-  <?php print render($title_suffix); ?>
+  <?php if ($title_options != COMMENT_TITLE_HIDDEN): ?>
```
indigoxela: Didn't you want to use an icon instead? This "if" will suppress the title completely if it's "hidden".

core/themes/basis/templates/comment.tpl.php
stpaultim: Where does it belong? ;-)

stpaultim: Removing...

core/modules/comment/templates/comment.tpl.php
stpaultim: No, it's a bit more complicated. The core comment.tpl.php file provides it's own permalink, which themes like Basis and several others overwrite. Automatically replacing the title with a link actually means that some themes end up with 2 permalinks (maybe better than none). 

PR #1 replaces title with icon with css across the board. This means that themes that have hard coded permalinks, will have a hard coded permalink AND an icon. Contrib themes can overwrite this behavior. 

PR #2 (this one) simply removes title and let's each theme handle the permalink on it's own. Both the core comment.tpl.php file as well as Bartik already have hard coded "permalink" links. So in this PR, I simply remove the title in the core and Bartik templates and leave the hard-coded permalink as it is. In Basis, which does not have a separate permalink, I add the icon. This PR (as it currently stands) will not have any effect on contrib themes that overwrite the comment.tpl.php and comment.css files. With this PR, users will not be able to hide comment titles in some themes, unless the theme is updated.

PR#1 just works in all themes (without the need for any changes to contrib themes), although not perfectly. 
PR#2 fixes the problem in core and core themes very nicely, but requires that contrib themes make adjustment.

There are still other possible combinations that we can discuss, my first goal was to get these two options working properly - and then see if we need even more options.

indigoxela: I'd like to suggest something: can you add (css) classes to that title? Alternative: add a css class to a parent div. That would simplify things for contrib themes.

But probably rather in a preprocessor (add a $title_attributes variable).

Background to this: this feature is based on constants, which will only exist in Backdrop starting with the next minor. If you add things in a preprocessor, themes don't have to also consider if the constants are defined.

Also: AFAIK constants have never been used in core like that (directly in a tpl.php), templates get variables via preprocessors. (TBH, that's how I always do it, I'm not 100% sure if core suggests that.)

stpaultim:
I'm removed constands from the tpl.php files based upon feedback from @indigoxela. It seems to me, that someone else asked me to put them there, but that might not be true. I can't find any specific reference to a discussion like that. 

I welcome feedback on my use of constants in this PR, because I'm not sure about what I'm doing.

core/modules/comment/tests/comment.test
```diff
@@ -190,13 +190,13 @@ class CommentHelperCase extends BackdropWebTestCase {
...
-   * Sets the value governing whether the subject field should be enabled.
+   * Sets the value governing the comment title field should be enabled.
    *
    * @param boolean $enabled
-   *   Boolean specifying whether the subject field should be enabled.
+   *   Boolean specifying whether the comment title field should be enabled.
    */
   function setCommentSubject($enabled) {
-    $this->setCommentSettings('comment_subject_field', ($enabled ? '1' : '0'), 'Comment subject ' . ($enabled ? 'enabled' : 'disabled') . '.');
+    $this->setCommentSettings('comment_title_options', ($enabled ? '1' : '0'), 'Comment subject ' . ($enabled ? 'enabled' : 'disabled') . '.');
```
stpaultim: This needs review. The test is passing, but I'm skeptical that I did what I need to do here. I don't quite understand it.

core/themes/bartik/templates/comment.tpl.php
```diff
@@ -18,19 +18,22 @@
...
  * - $permalink: Comment permalink.
+ * - $permalink_path: Available to build your own permalink.
  * - $submitted: Submission information created from $author and $created during
  *   template_preprocess_comment().
  * - $picture: Authors picture.
  * - $signature: Authors signature.
  * - $status: Comment status. Possible values are:
  *   comment-unpublished, comment-published or comment-preview.
  * - $title: Linked title.
+ * - $title_options: Auto-generated(0), custom(1) or hidden(2).
```
indigoxela: Already better, no more constants. But what I actually had in mind: provide something more "speaking"

"title_display":

- generated / custom = TRUE
- hidden = FALSE

core/themes/bartik/templates/comment.tpl.php
```diff
@@ -72,9 +75,11 @@
       <span class="new"><?php print $new; ?></span>
     <?php endif; ?>
 
-    <?php print render($title_prefix); ?>
-    <h3><?php print $title; ?></h3>
-    <?php print render($title_suffix); ?>
+    <?php if ($title_options != '2'): ?>
```
indigoxela: If it were "title_display", then this would become `if ($title_display)`.

Whether titles are generated or custom doesn't have any meaning here.

core/modules/comment/comment.theme.inc
```diff
@@ -145,5 +148,8 @@ function template_preprocess_comment(&$variables) {
       $variables['classes'][] = 'comment-by-viewer';
     }
   }
+  if ($variables['title_options'] == COMMENT_TITLE_HIDDEN) {
+    $variables['classes'][] = 'hide-comment-title';
```
indigoxela: What I'd find cool here for easier styling: be more specific with the css class:

- comment-title-generated
- comment-title-custom
- comment-title-hidden

I belief, that would provide maximum flexibility for themers. Even easier, if the `h3` tag had an extra css class like "comment-title" or "comment-subject".

core/themes/basis/css/component/comment.css
```diff
@@ -163,3 +163,16 @@
...
 }
+
+a.comment-permalink  {
+  width: .7em;
+  height: .7em;
+  padding: .1em;
+  display: inline-block;
+  background-color: currentColor;
+  -webkit-mask-image: url("/core/modules/comment/images/link-icon.svg");
+  mask-image: url("/core/modules/comment/images/link-icon.svg");
```
indigoxela: Unless I'm missing something, the png variant isn't used anymore and could get deleted from the images directory.

core/themes/bartik/templates/comment.tpl.php
stpaultim: @indigoxela - Thanks much for the comments. I'm learning new things and appreciate this feedback. 

In this case, I'm a bit confused. I don't know if you remember, but we started this issue (way, way back) with two boolean choices. 1) Auto generated titles vs custom 2) Hidden or not

There were some objections to this, because if we have hidden titles, we don't need auto or custom titles. At this point, we changed from two booleans to one radio button choice with three options - only one of which ever needs to be true at any given time. I've gotten very positive feedback on that change. 

It seems (and I may very well be wrong) that you are re-suggesting two boolean values. Now maybe your suggestion is compatible with the new Radio buttons, but if so, then I'm a bit confused about how to implement it. I might need a more specific example or some actual suggested code. 

This was one early idea, see discussion before and after this: https://github.com/backdrop/backdrop-issues/issues/4569#issuecomment-1114469244

core/themes/bartik/templates/comment.tpl.php
stpaultim: This comment is dependent upon us coming to an understanding about the previous comment, I think.

core/themes/basis/css/component/comment.css
stpaultim: @indigoxela you are 100% correct. Can't believe I didn't catch this. I think you pointed this out in the past. Will fix.

core/modules/comment/comment.theme.inc
stpaultim: @indigoxela - I'm not seeing the use case for a themer to care the difference between generated and custom titles. In my eyes, the only thing that matters is hidden or not. However, I have no objection to providing the additional classes.

I'll put them in here and see what folks think.

core/themes/bartik/templates/comment.tpl.php
indigoxela: I see, I need to be more clear about my suggestion. :wink:

It's about how it is used in the template (hence display). For that we only need some sort of on/off switch to either render the title (be it generated or custom), or not render the title (when it's hidden). So a single on/off thing seems sufficient for this particular usage in comment.tpl.php. The wrapper css class allows finer grained theming.

My suggestion is really only about the render switch in the preprocessor for the template, I'm fine with the three radios in the setting, which also control if the title form item is provided or not.

What may be a bit confusing is, that the radios (and constants) technically control two different things:

- The form item (textfield) in the comment form
- The title rendering when displaying comments

But I still think, that's a good idea - and that's what we settled on. In fact, form item and display rendering *are* related (as discussed). If we allow people to set a custom comment title (subject), we for sure don't want to hide it in display. If we generate it automatically, we might want to hide it to prevent that odd duplicate text.

core/modules/comment/comment.theme.inc
indigoxela: > I'm not seeing the use case for a themer to care the difference between generated and custom titles.

It doesn't hurt to provide css classes like that, simply for more flexibility for themers. If the css classes are there, they can decide how a generated or a human-provided title might look like without the need for an additional hook in their template.php (which might be a bit of a hurdle for front-enders with low or no experience with php).

core/modules/comment/tests/comment.test
indigoxela: This is correct for now, but currently only covers COMMENT_TITLE_AUTO vs. COMMENT_TITLE_CUSTOM.

When we're done with the overall implementation, the functional tests need some update, so all three constants are covered (not only on setCommentSettings, but in the whole test case). But that can be done by someone else - with more experience with functional tests.

core/themes/bartik/templates/comment.tpl.php
indigoxela: I just realized, I commented on the wrong tpl.php file. This should have gone to the main one in core/modules/comment/templates/comment.tpl.php. :wink:

core/themes/basis/css/component/comment.css
```diff
@@ -163,3 +163,16 @@
...
   float: left;
 }
+
+a.comment-permalink  {
+  width: .7em;
+  height: .7em;
+  padding: .1em;
+  display: inline-block;
+  background-color: currentColor;
+  -webkit-mask-image: url("/core/modules/comment/images/link-icon.svg");
```
indigoxela: ` url("/core/modules/comment/images/link-icon.svg");`

If you define the url like that, it won't work on installs in subdirectories.

core/modules/comment/comment.theme.inc
stpaultim: Understood. I did add three classes you suggested.

core/themes/bartik/templates/comment.tpl.php
stpaultim: @indigoxela - I THINK I get your point now. Thanks for your patience. 

If I understand you correctly, the change you are requesting does not functionally change the way the PR works but is more direct/understandable in the comment.tpl.php file. I think you are suggesting that instead of passing the original variable which handles more than just the title decision, I only pass need to pass a TRUE/FALSE varaible to the template regarding whether titles display or not. 

This will function the same, but will make more sense and be easier to understand for themers. 

I like this idea, which would not have occurred to me on my own. I've made that change. If I'm still missing any part of your suggestion, let me know. I also welcome comments on my implementation.

core/themes/basis/css/component/comment.css
stpaultim: I went ahead and changed this to a relative path. Let me know if that is the best option, keeping in mind that we may need to change this again once we decide on the best place for the icon. 

I know think that the icon should either be part of Basis or in core/misc for use by other themes or modules. Waiting for feedback on that.

core/modules/comment/comment.theme.inc
```diff
@@ -104,9 +104,16 @@ function template_preprocess_comment(&$variables) {
 
   $uri = $comment->uri();
   $uri['options'] += array('attributes' => array('class' => array('permalink'), 'rel' => 'bookmark'));
-
-  $variables['title'] = l($comment->subject, $uri['path'], $uri['options']);
+  
   $variables['permalink'] = l(t('Permalink'), $uri['path'], $uri['options']);
+  $permanlink_path = $uri['path'] . "#" . $uri['options']['fragment'];
+  $variables['permalink_path'] = $permanlink_path;
+  if ($node_type->settings['comment_title_options'] != COMMENT_TITLE_HIDDEN) {
```
docwilmot: This whole block could be shortened (doesnt have to be so, just saying) to 
`$variables['title_display'] = $node_type->settings['comment_title_options'] != COMMENT_TITLE_HIDDEN`

core/modules/comment/templates/comment.tpl.php
```diff
@@ -60,12 +63,14 @@
 ?>
 <article class="<?php print implode(' ', $classes); ?> clearfix"<?php print backdrop_attributes($attributes); ?>>
 
-  <?php print render($title_prefix); ?>
   <?php if ($new): ?>
     <mark class="new"><?php print $new; ?></mark>
   <?php endif; ?>
-  <h3><?php print $title; ?></h3>
-  <?php print render($title_suffix); ?>
+  <?php if ($title_display == TRUE): ?>
```
docwilmot: This could be shortened (doesnt have to be so, just saying) to 
`<?php if ($title_display): ?>`

core/themes/bartik/templates/comment.tpl.php
```diff
@@ -72,9 +75,11 @@
       <span class="new"><?php print $new; ?></span>
     <?php endif; ?>
 
-    <?php print render($title_prefix); ?>
-    <h3><?php print $title; ?></h3>
-    <?php print render($title_suffix); ?>
+    <?php if ($title_display == TRUE): ?>
```
docwilmot: This could be shortened (doesnt have to be so, just saying) to 
`<?php if ($title_display): ?>`

core/themes/basis/templates/comment.tpl.php
```diff
@@ -67,13 +70,18 @@
   <div class="comment-text">
 
     <div class="comment-title">
-      <?php print render($title_prefix); ?>
-      <h3><?php print $title; ?></h3>
-      <?php print render($title_suffix); ?>
+      <?php if ($title_display == TRUE): ?>
```
docwilmot: This could be shortened (doesnt have to be so, just saying) to 
`<?php if ($title_display): ?>`

core/themes/basis/templates/comment.tpl.php
```diff
@@ -67,13 +70,18 @@
...
+        <?php print render($title_prefix); ?>
+        <h3><?php print $title; ?></h3>
+        <?php print render($title_suffix); ?>
+      <?php endif; ?>
       <?php if ($new): ?>
         <span class="marker"><?php print $new; ?></span>
       <?php endif; ?>
       <span class="comment-time"><?php print $created; ?></span>
+      <?php if ($title_display == FALSE):  ?>
```
docwilmot: This could be shortened (doesnt have to be so, just saying) to 
`<?php if (!$title_display): ?>`

core/modules/comment/comment.install
```diff
@@ -375,7 +375,7 @@ function comment_update_1001() {
     update_variable_del('comment_default_per_page_' . $node_type->type);
     update_variable_del('comment_default_mode_' . $node_type->type);
     update_variable_del('comment_anonymous_' . $node_type->type);
-    update_variable_del('comment_subject_field_' . $node_type->type);
+    update_variable_del('comment_title_options_' . $node_type->type);
```
ghost: This doesn't make sense, as `comment_title_options_...` is a new setting that won't have existed in Drupal (so there's no variable to delete), while `comment_subject_field_...` does exist in Drupal (so people still need that variable deleted when upgrading).

core/modules/comment/comment.install
```diff
@@ -470,6 +470,26 @@ function comment_update_1007() {
...
+/**
+ * New setting for comment titles. Replacing comment_title_field
+ * with comment_subject_options.
+ */
+function comment_update_1008() {
+  foreach (config_get_names_with_prefix('node.type.') as $config_name) {
+    $config = config($config_name);
+    $comment_title = $config->get('settings.comment_subject_field');
+    if ($comment_title == COMMENT_TITLE_AUTO) {
+      // auto generated titles
```
ghost: Comments should form proper sentences (e.g. capital letter to start, full-stop to end): https://docs.backdropcms.org/doc-standards#general

```suggestion
      // Auto generated titles.
```

core/modules/comment/comment.install
```diff
@@ -470,6 +470,26 @@ function comment_update_1007() {
...
+      $config->set('settings.comment_title_options', COMMENT_TITLE_AUTO);
+    } else {
+      // custom titles
```
ghost: See above.

core/modules/comment/comment.install
```diff
@@ -470,6 +470,26 @@ function comment_update_1007() {
   }
 }
 
+/**
+ * New setting for comment titles. Replacing comment_title_field
+ * with comment_subject_options.
```
ghost: I think you have the setting names mixed up here...

core/modules/comment/comment.install
```diff
@@ -470,6 +470,26 @@ function comment_update_1007() {
...
+    $config = config($config_name);
+    $comment_title = $config->get('settings.comment_subject_field');
+    if ($comment_title == COMMENT_TITLE_AUTO) {
+      // auto generated titles
+      $config->set('settings.comment_title_options', COMMENT_TITLE_AUTO);
+    } else {
+      // custom titles
+      $config->set('settings.comment_title_options', COMMENT_TITLE_CUSTOM);
+    }
+    $config->clear('settings.comment_subject_field');
```
ghost, line 503: You're removing the old `comment_subject_field` config value here (which is fine), but since you also removed it from `hook_uninstall()`, then it needs to be removed from the node type settings here too.

core/modules/comment/comment.module
```diff
@@ -67,6 +67,22 @@ define('COMMENT_NODE_CLOSED', 1);
  */
 define('COMMENT_NODE_OPEN', 2);
 
+/**
+ * Auto generate comment titles.
```
ghost: I think best to try and have the descriptions of these 3 constants as similar as possible you it's easy to tell they're related:

```suggestion
 * Auto generated comment titles.
```

core/modules/comment/comment.module
```diff
@@ -67,6 +67,22 @@ define('COMMENT_NODE_CLOSED', 1);
...
+ */
+define('COMMENT_TITLE_AUTO', 0);
+
+/**
+ * Allow custom titles for comments.
```
ghost: As above:

```suggestion
 * Custom comment titles.
```

core/modules/comment/comment.module
```diff
@@ -67,6 +67,22 @@ define('COMMENT_NODE_CLOSED', 1);
...
+ */
+define('COMMENT_TITLE_CUSTOM', 1);
+
+/**
+ * Hide comment titles.
```
ghost: As above:

```suggestion
 * Hidden comment titles.
```

core/modules/comment/comment.module
```diff
@@ -135,7 +151,7 @@ function comment_field_extra_fields() {
   $return = array();
 
   foreach (node_type_get_types() as $type) {
-    if ($type->settings['comment_subject_field']) {
+    if ($type->settings['comment_title_options'] == COMMENT_TITLE_AUTO) {
```
ghost: This was previously checking if `comment_subject_field` was `TRUE` (i.e. `1`). But now it's checking if the new value equals `COMMENT_TITLE_AUTO` (which is `0`)... I don't think that's right.

core/modules/comment/comment.module
```diff
@@ -1165,6 +1175,19 @@ function comment_form_node_type_form_alter(&$form, $form_state) {
       '#default_value' => $node_type->settings['comment_close_days'],
       '#field_suffix' => t('days'),
     );
+    $form['comment']['comment_title_options'] = array(
+      '#type' => 'radios',
+      '#title' => t('Comment titles'),
+      '#default_value' => $node_type->settings['comment_title_options'],
+      '#options' => array(
+        COMMENT_TITLE_CUSTOM => t('Custom'),
+        COMMENT_TITLE_AUTO => t('Auto Generated'),
```
ghost: I don't think `Generated` should have a capital here...

```suggestion
        COMMENT_TITLE_AUTO => t('Auto generated'),
```

core/modules/comment/comment.module
```diff
@@ -1165,6 +1175,19 @@ function comment_form_node_type_form_alter(&$form, $form_state) {
...
+        COMMENT_TITLE_HIDDEN => t('Hidden'),
+      ),
+    );
+    $form['comment']['comment_title_options'][COMMENT_TITLE_CUSTOM]['#description'] = t('Comment titles are visible.A title field is added to comment forms to allow for custom titles.');
```
ghost: Need a space after the full-stop.

```suggestion
    $form['comment']['comment_title_options'][COMMENT_TITLE_CUSTOM]['#description'] = t('Comment titles are visible. A title field is added to comment forms to allow for custom titles.');
```

core/modules/comment/comment.module
```diff
@@ -1165,6 +1175,19 @@ function comment_form_node_type_form_alter(&$form, $form_state) {
...
+    $form['comment']['comment_title_options'][COMMENT_TITLE_AUTO]['#description'] = t('Comment titles are visible. Comment titles are automatically generated from the comment body.');
```
ghost: Not sure I like the repetition here...

```suggestion
    $form['comment']['comment_title_options'][COMMENT_TITLE_AUTO]['#description'] = t('Comment titles are visible and are automatically generated from the comment body.');
```

core/modules/comment/comment.module
```diff
@@ -1165,6 +1175,19 @@ function comment_form_node_type_form_alter(&$form, $form_state) {
...
+    $form['comment']['comment_title_options'][COMMENT_TITLE_HIDDEN]['#description'] = t('Comment titles are hidden (not supported by some contributed themes).');
```
ghost: As much as I'm against adding this line, if it has to be here I think this wording might be better...

```suggestion
    $form['comment']['comment_title_options'][COMMENT_TITLE_HIDDEN]['#description'] = t('Comment titles are hidden (this option might not be supported by some themes).');
```

core/modules/comment/comment.module
```diff
@@ -1864,7 +1887,7 @@ function comment_form($form, &$form_state, $comment) {
     '#title' => t('Title'),
     '#maxlength' => 64,
     '#default_value' => $comment->subject,
-    '#access' => $node_type->settings['comment_subject_field'],
+    '#access' => $node_type->settings['comment_title_options'] == COMMENT_TITLE_CUSTOM,
```
ghost, line 1890: I think this would appear when comment titles are hidden (i.e. set to `2`). May need to use `===` here instead...

```suggestion
    '#access' => $node_type->settings['comment_title_options'] === COMMENT_TITLE_CUSTOM,
```

core/modules/comment/comment.theme.inc
```diff
@@ -104,9 +104,16 @@ function template_preprocess_comment(&$variables) {
 
   $uri = $comment->uri();
   $uri['options'] += array('attributes' => array('class' => array('permalink'), 'rel' => 'bookmark'));
-
-  $variables['title'] = l($comment->subject, $uri['path'], $uri['options']);
+  
   $variables['permalink'] = l(t('Permalink'), $uri['path'], $uri['options']);
+  $permanlink_path = $uri['path'] . "#" . $uri['options']['fragment'];
+  $variables['permalink_path'] = $permanlink_path;
```
ghost: Simplification:

```suggestion
  $variables['permalink_path'] = $uri['path'] . "#" . $uri['options']['fragment'];
```

core/modules/comment/comment.theme.inc
```diff
@@ -104,9 +104,16 @@ function template_preprocess_comment(&$variables) {
...
+  if ($node_type->settings['comment_title_options'] != COMMENT_TITLE_HIDDEN) {
+    $variables['title_display'] = TRUE;
+  } else {
+    $variables['title_display'] = FALSE; 
+  }
```
ghost: I agree with @docwilmot:

```suggestion
  $variables['title_display'] = ($node_type->settings['comment_title_options'] != COMMENT_TITLE_HIDDEN);
```

core/modules/comment/comment.theme.inc
```diff
@@ -145,5 +152,11 @@ function template_preprocess_comment(&$variables) {
       $variables['classes'][] = 'comment-by-viewer';
     }
   }
+  if ($node_type->settings['comment_title_options'] == COMMENT_TITLE_HIDDEN) {
+    $variables['classes'][] = 'comment-title-hidden';
+  } elseif ($node_type->settings['comment_title_options'] == COMMENT_TITLE_AUTO) {
+    $variables['classes'][] = 'comment-title-auto';
+  } else {
```
ghost: Coding standards:

```suggestion
  }
  elseif ($node_type->settings['comment_title_options'] == COMMENT_TITLE_AUTO) {
    $variables['classes'][] = 'comment-title-auto';
  }
  else {
```

core/modules/comment/templates/comment.tpl.php
```diff
@@ -19,19 +19,22 @@
...
  *   comment-unpublished, comment-published or comment-preview.
  * - $title: Linked title.
+ * - $title_display: Should the title be displayed (TRUE or FALSE).
  * - $classes: Array of classes that can be used to style contextually through
  *   CSS. The default values can be one or more of the following:
  *   - comment: The current template type, i.e., "theme hook".
  *   - comment-by-anonymous: Comment by an unregistered user.
  *   - comment-by-node-author: Comment by the author of the parent node.
  *   - comment-preview: When previewing a new or edited comment.
+ *   - comment-title-hidden: Comment titles should be hidden.
```
ghost, line 38: What about the other classes?

core/modules/comment/templates/comment.tpl.php
ghost: Agreed:

```suggestion
  <?php if ($title_display): ?>
```

core/modules/comment/templates/comment.tpl.php
```diff
@@ -60,12 +63,14 @@
...
 
-  <?php print render($title_prefix); ?>
   <?php if ($new): ?>
     <mark class="new"><?php print $new; ?></mark>
   <?php endif; ?>
-  <h3><?php print $title; ?></h3>
-  <?php print render($title_suffix); ?>
+  <?php if ($title_display == TRUE): ?>
+    <?php print render($title_prefix); ?>
+    <h3 class="comment-title"><?php print $title; ?></h3>
```
ghost: I don't think we should be adding classes in here directly... Better to do it in the preprocess function so it can be overridden like all the other classes if need be.

core/modules/comment/tests/comment.test
```diff
@@ -190,13 +190,13 @@ class CommentHelperCase extends BackdropWebTestCase {
   }
 
   /**
-   * Sets the value governing whether the subject field should be enabled.
+   * Sets the value governing the comment title field should be enabled.
```
ghost:
```suggestion
   * Sets the value governing whether the comment title field should be enabled.
```

core/themes/bartik/templates/comment.tpl.php
```diff
@@ -18,19 +18,22 @@
...
  *   comment-unpublished, comment-published or comment-preview.
  * - $title: Linked title.
+ * - $title_display: Should the title be displayed (TRUE or FALSE).
  * - $classes: Array of classes that can be used to style contextually through
  *   CSS. The default values can be one or more of the following:
  *   - comment: The current template type, i.e., "theme hook".
  *   - comment-by-anonymous: Comment by an unregistered user.
  *   - comment-by-node-author: Comment by the author of the parent node.
  *   - comment-preview: When previewing a new or edited comment.
+ *   - comment-title-hidden: Comment titles should be hidden.
```
ghost, line 37: Other classes too (see above).

core/themes/bartik/templates/comment.tpl.php
```diff
@@ -72,9 +75,11 @@
...
     <?php endif; ?>
 
-    <?php print render($title_prefix); ?>
-    <h3><?php print $title; ?></h3>
-    <?php print render($title_suffix); ?>
+    <?php if ($title_display == TRUE): ?>
+      <?php print render($title_prefix); ?>
+      <h3 class="comment-title"><?php print $title; ?></h3>
+      <?php print render($title_suffix); ?>
+    <?php endif; ?>
```
ghost: Same changes needed here as core comment template file (above).

core/themes/basis/css/component/comment.css
```diff
@@ -163,3 +163,16 @@
 [dir="rtl"] .comment-footer {
   float: left;
 }
+
+a.comment-permalink  {
+  width: .7em;
+  height: .7em;
+  padding: .1em;
+  display: inline-block;
+  background-color: currentColor;
```
ghost, line 172: What is `currentColor`?

core/themes/basis/templates/comment.tpl.php
```diff
@@ -18,19 +18,22 @@
...
  *   comment-unpublished, comment-published or comment-preview.
  * - $title: Linked title.
+ * - $title_display: Should the title be displayed (TRUE or FALSE).
  * - $classes: Array of classes that can be used to style contextually through
  *   CSS. The default values can be one or more of the following:
  *   - comment: The current template type, i.e., "theme hook".
  *   - comment-by-anonymous: Comment by an unregistered user.
  *   - comment-by-node-author: Comment by the author of the parent node.
  *   - comment-preview: When previewing a new or edited comment.
+ *   - comment-title-hidden: Comment titles should be hidden.
```
ghost, line 37: As above.

core/themes/basis/templates/comment.tpl.php
```diff
@@ -67,13 +70,18 @@
...
+        <h3><?php print $title; ?></h3>
+        <?php print render($title_suffix); ?>
+      <?php endif; ?>
       <?php if ($new): ?>
         <span class="marker"><?php print $new; ?></span>
       <?php endif; ?>
       <span class="comment-time"><?php print $created; ?></span>
+      <?php if ($title_display == FALSE):  ?>
+        <a class="comment-permalink" href="/<?php print $permalink_path; ?>"></a>
+      <?php endif; ?>
```
ghost: As above (shortened as per @docwilmot's suggestion and remove hard-coded classes).

ghost (dismissed):
A few (:wink:) suggested changes...

core/modules/comment/templates/comment.tpl.php
indigoxela: Right, that also came to my mind.

The preprocess function could create an array of classes and here in the template they get flattened like:

`class="<?php print implode(' ', $title_classes); ?>"`

Another option would be to create a complete attributes array and apply them here with:
`<?php print backdrop_attributes($title_attributes); ?>`
But that may be a little too much, as we don't need anything besides css class here (I guess).

I'm aware that there are more hardcoded css classes or attributes in templates, but personally, I find such arrays more flexible.

core/modules/comment/comment.theme.inc
stpaultim: @docwilmot - I appreciate ideas for shortening code. I will look closer at this later and update PR.

core/modules/comment/comment.module
stpaultim: I appreciate that.

core/modules/comment/comment.theme.inc
stpaultim: I assumed there was a shorter way to write this. Does this assign the varaible a value for both TRUE and FALSE conditions? (I will test when I have time)

core/modules/comment/templates/comment.tpl.php
stpaultim: @BWPanda - Thanks. I added those after adding this comment. I will update them.

core/modules/comment/templates/comment.tpl.php
stpaultim: @BWPanda and @indigoxela - I considered that, but didn't know how. This helps.

core/modules/comment/comment.theme.inc
ghost: Yep

core/themes/basis/css/component/comment.css
stpaultim: @BWPanda - Truth is that I don't know, other than this - https://css-tricks.com/currentcolor/

I'm 99% sure that this was something @indigoxela recommended and part of why she suggested I use a .svg file instead of .png file. I'm quite sure I did not come up with this idea on my own. It might be worth checking to see if it's working as expected. @indigoxela can you confirm if this might have been your idea and if you think I've implemented it correctly?

stpaultim: @BWPanda - I found where @indigoxela recommended this: 
https://github.com/backdrop/backdrop-issues/issues/4569#issuecomment-1120381705

ghost: Ah, nice! Didn't know about that.

core/modules/comment/comment.install
stpaultim: @BWPanda - I am not sure what you are suggesting here - or is it possible I already implemented your suggestion?

core/modules/comment/comment.module
stpaultim: @BWPanda - I tried this option and it didn't work. The Title field did not display when it should.

stpaultim:
@BWPanda - Please, take a look at the update hook (function comment_update_1008()). Did I address your concerns or not?

Also, I tried chaning the "==" to "===" and it broke the comment form. https://github.com/backdrop/backdrop/pull/4068/files#r908003822

core/modules/comment/comment.install
ghost: It's a bit complicated... `hook_uninstall()` removes variables from `$node_type->settings`. But since the old `subject` variable was deleted from `hook_uninstall()`, it'll need to be removed from `$node_type->settings` somewhere else then. `comment_update_1008()` seems appropriate since this is where you're replacing the old `subject` variable with the new `title` one. So here you need to add similar code to what's in hook_uninstall()`, but just for the `subject` variable. Hope that makes sense.

stpaultim: I think it does. Processing now...

stpaultim: @BWPanda - I think got it. BUT, I'm stretching here and could easily be missing things. I would appreciate if you take another look.

core/modules/comment/comment.module
ghost: Ok, I was just looking at the code, didn't test. If it doesn't work feel free to revert then.

core/modules/comment/comment.install
```diff
@@ -470,6 +470,35 @@ function comment_update_1007() {
...
+      // Set for auto generated titles.
+      $config->set('settings.comment_title_options', COMMENT_TITLE_AUTO);
+    } 
+    else {
+      // Set for custom titles.
+      $config->set('settings.comment_title_options', COMMENT_TITLE_CUSTOM);
+    }
+    $config->clear('settings.comment_subject_field');
+    $config->save();
+
```
indigoxela: There's a syntax error, above "foreach" isn't closed, a `}` is missing here.

stpaultim: Dang this coding stuff can be difficult. Fixing now.

core/modules/comment/comment.install
ghost: Looks good to me!

core/modules/comment/comment.install
```diff
@@ -470,6 +470,36 @@ function comment_update_1007() {
...
+    }
+    $config->clear('settings.comment_subject_field');
+    $config->save();
+  }
+  
+  // Remove comment_subject_title variable now that it has been replaced.
+  $node_types = node_type_get_types();
+  foreach ($node_types as $type_name => $node_type) {
+    if (isset($node_type->settings['comment_subject_title'])) {
+      unset($node_type->settings['comment_subject_title']);
```
ghost: Oops, typo:

```suggestion
    if (isset($node_type->settings['comment_subject_field'])) {
      unset($node_type->settings['comment_subject_field']);
```

core/modules/comment/comment.install
```diff
@@ -470,6 +470,36 @@ function comment_update_1007() {
...
+      // Set for custom titles.
+      $config->set('settings.comment_title_options', COMMENT_TITLE_CUSTOM);
+    }
+    $config->clear('settings.comment_subject_field');
+    $config->save();
+  }
+  
+  // Remove comment_subject_title variable now that it has been replaced.
+  $node_types = node_type_get_types();
+  foreach ($node_types as $type_name => $node_type) {
```
ghost: This can be simplified slightly:

```suggestion
  foreach (node_type_get_types() as $type_name => $node_type) {
```

core/modules/comment/comment.install
```diff
@@ -470,6 +470,35 @@ function comment_update_1007() {
   }
 }
 
+/**
+ * New setting for comment titles. Replacing comment_subject_field
+ * with comment_title_options.
+ */
+function comment_update_1008() {
+  foreach (config_get_names_with_prefix('node.type.') as $config_name) {
```
jenlampton, line 477: As @stpaultim mentioned in the last comment on the issue, we'll need to add a check before this for the contrib module, and convert the current value from the contrib module (if Backdrop) or from the variables table (if D7), and then disable the module.

core/themes/basis/css/component/comment.css
```diff
@@ -163,3 +163,16 @@
...
+  -webkit-mask-image: url("../../images/link-icon.svg");
```
jenlampton, line 173: This needs a png fallback if we're going to use a SVG. Here's an example from messages where we also wanted to use an SVG. 
```
div.messages.info:before {
  background-image: url(../../../misc/message-32-info.png);
  background-color: #0074bd;
}
/* SVG overrides */
div.messages.info:before {
  background-image: linear-gradient(transparent, transparent), url(../../../misc/message-info.svg);
}
```

indigoxela: > This needs a png fallback.

I'm not sure. Why? Internet Explorer? :laughing: (RIP)

@jenlampton this is different. It's not a background image, but a mask image. We use a mask as that way the icon color will follow the link color. Keep in mind: Basis has color module support, so the link color isn't necessarily blue, but can be something completely different.

See also: https://developer.mozilla.org/en-US/docs/Web/CSS/mask-image
and https://caniuse.com/?search=mask-image

Seems like SVG as mask have better support than PNG.

core/modules/comment/comment.install
```diff
@@ -470,6 +470,35 @@ function comment_update_1007() {
...
+    $config->clear('settings.comment_subject_field');
+    $config->save();
+  }
+  
+  // Remove comment_subject_title variable now that it has been replaced.
+  foreach (node_type_get_types() as $type_name => $node_type) {
+    if (isset($node_type->settings['comment_subject_field'])) {
+      unset($node_type->settings['comment_subject_field']);
+    }
+    node_type_save($node_type);
```
quicksketch: We should only use `config` calls in update hooks, as using APIs like `node_type_save()` will trigger hooks and may call more unexpected code that does not work correctly in the context of update.php

docwilmot:
Was going to make an actual review comment but realized this was pre-existing. There is a lot of `$this->backdropLogout()` in comment tests. It's not necessary to log out before switching test users is it?

stpaultim:
Not ready for review yet.

jenlampton:
> There is a lot of $this->backdropLogout() in comment tests. It's not necessary to log out before switching test users is it?

I thought  it was? We copied a previous part of the test where a logout happened before a log-in. I haven't tried without logging out...

core/themes/basis/css/component/comment.css
jenlampton: Ah, I missed the mask image bit. If that's the case then maybe a fallback isn't necessary. ~What happens with non-webkit browsers?~ (nvm, I see we have both in here)

docwilmot:
I checked after I posted, seems `$this->backdropLogin()` calls `$this->backdropLogout()` first, so no, its not necessary to actually log out first, but there isnt any cost to doing so either.

core/themes/bartik/templates/comment.tpl.php
jenlampton: I'm not sure either variable is necessary. Can't we just check if the title itself is empty or not? If the title is set to hidden, then the `$title` is an empty string, isn't it? Why do we need another variable?

jenlampton:
> but there isnt any cost to doing so either.

well, a little clutter :)
Thanks! I just learned something :)

stpaultim:
@jenlampton - FYI - over the weekend, I was getting a bunch of test failures and I thought that some of them might have to do with the changes we made to the 4th comment test. So, I tried (in current PR) to leave that test the way was and add a 5th comment to accomplish our new tests. 

I then found some other variables that weren't up to date and might have been the real culprits. I may go back to what you and I did and see if I can get that to work. It's a little less code.

I've done a number of things that I'm unsure of and need time to review. If anyone wants to review the tests, feel free. But, keep in mind that I'm not sure I'm done with them yet. I expect to catch some silly mistakes on my own.

core/modules/comment/comment.install
stpaultim: DONE?

core/modules/comment/comment.module
```diff
@@ -67,6 +67,22 @@ define('COMMENT_NODE_CLOSED', 1);
...
+
+/**
+ * Custom comment titles.
+ */
+define('COMMENT_TITLE_CUSTOM', 1);
+
+/**
+ * Hidden comment titles.
+ */
+define('COMMENT_TITLE_HIDDEN', 2);
```
klonos: If we are fixed to using constants for these options, can we please not map them to "arbitrary" numbers? This causes things to be saved in config like `"comment_title_options": "2"`, which is very developer-centric and cryptic - I'd rather this was easy to read and self-explanatory instead, like `"comment_title_options": "hidden"` ...I think I would actually prefer it if it was something like `"comment_title_display": "hidden"`

```suggestion
/**
 * Auto generated comment titles.
 */
define('COMMENT_TITLE_AUTO', 'auto');

/**
 * Custom comment titles.
 */
define('COMMENT_TITLE_CUSTOM', 'custom');

/**
 * Hidden comment titles.
 */
define('COMMENT_TITLE_HIDDEN', 'hidden');
```

I would actually prefer it if we were even less vague where possible, having these options be `auto-generated` / `user-provided` / `hidden` (or something like that). Point being better DX, so that if someone inspects config diffs, they immediately know what is being changed to what, without having to know what 1,2,3 or "auto" are.

core/modules/comment/comment.module
```diff
@@ -1165,6 +1175,19 @@ function comment_form_node_type_form_alter(&$form, $form_state) {
...
+      '#default_value' => $node_type->settings['comment_title_options'],
+      '#options' => array(
+        COMMENT_TITLE_CUSTOM => t('Custom'),
+        COMMENT_TITLE_AUTO => t('Auto generated'),
+        COMMENT_TITLE_HIDDEN => t('Hidden'),
+      ),
+    );
+    $form['comment']['comment_title_options'][COMMENT_TITLE_CUSTOM]['#description'] = t('Comment titles are visible. A title field is added to comment forms to allow for custom titles.');
+    $form['comment']['comment_title_options'][COMMENT_TITLE_AUTO]['#description'] = t('Comment titles are visible and are automatically generated from the comment body.');
+    $form['comment']['comment_title_options'][COMMENT_TITLE_HIDDEN]['#description'] = t('Comment titles are hidden (this option might not be supported by some themes).');
```
klonos: I don't think that the "Comment titles are hidden" text provides any value/clarification for the "Hidden" option. I also don't think that repeating "Comment titles are visible." in the other two options is ideal - it could instead be incorporated in the radio option, which would also make the distinction be tween the 2 "visible" options and the "hidden" one. Something like this perhaps?:

```suggestion
    $form['comment']['comment_title_options'] = array(
      '#type' => 'radios',
      '#title' => t('Comment titles'),
      '#default_value' => $node_type->settings['comment_title_options'],
      '#options' => array(
        COMMENT_TITLE_CUSTOM => t('Visible, customizable'),
        COMMENT_TITLE_AUTO => t('Visible, auto-generated'),
        COMMENT_TITLE_HIDDEN => t('Hidden'),
      ),
    );
    $form['comment']['comment_title_options'][COMMENT_TITLE_CUSTOM]['#description'] = t('A title field is provided, to allow customizing comment titles.');
    $form['comment']['comment_title_options'][COMMENT_TITLE_AUTO]['#description'] = t('Comment titles are automatically generated from the comment body.');
    $form['comment']['comment_title_options'][COMMENT_TITLE_HIDDEN]['#description'] = t('Note: this option might not be supported by some themes.');
```

klonos: ...for visual comparison:

Before:
image

After:
image

stpaultim: IMHO - Good ideas. Added.

core/modules/comment/comment.module
stpaultim: I'd like another opinion on this as I was originally advised to do it the other way and am not qualified to say which is better.

docwilmot: It is not an unreasonable suggestion, as @klonos said this would make reading config files easier. But its not wrong the way it is either, as we have a mix of constant mappings in core it seems: sometimes we use integers (eg `define('LAYOUT_STORAGE_DEFAULT', 4)` and `define('NODE_PUBLISHED', 1)`), other times strings (eg `define('LAYOUT_TITLE_CUSTOM', 'custom')` and `define('NODE_ACCESS_ALLOW', 'allow')`). Again this could be another issue/initiative for switching to more readable config files but shouldnt block this.

klonos: @docwilmot agreed re the fact that we have either been adding or have inherited many such "arbitrary" values that make people scratch their heads when they see them. Also agreed that we could leave things as they are with the current PR in order to not block it, and then file a follow-up to clean things up 👍🏼 ...but since we have time till feature freeze, I say why add to the already big, messy pile of things we will need to fix in the future, when we have the opportunity to polish this PR now?

my 2c

stpaultim: OK, I implemented these changes. We'll see if anyone else has anything to add.

core/modules/comment/comment.module
```diff
@@ -67,6 +67,22 @@ define('COMMENT_NODE_CLOSED', 1);
...
+/**
+ * Custom comment titles.
+ */
+define('COMMENT_TITLE_CUSTOM', 'custom');
+
+/**
+ * Hidden comment titles.
+ */
+define('COMMENT_TITLE_HIDDEN', 'hidden');
+
```
indigoxela, line 84: Thinking this over... as we now switched to human readable values, there's actually no need for constants anymore.

Constants are usually only used if the value is something obscure, like 1, 4, 123, ...

core/modules/comment/comment.install
herbdool: @stpaultim it's missing a $config-save() but otherwise this hook looks good.

core/modules/comment/comment.module
herbdool: Not true. There are a number of places where constants store strings in core. It's useful for collecting them at the top of their file.

E.g.: `define('MAINTENANCE_MODE', 'install');`

herbdool: Not to mention that the values get referenced all over the place so using the constant is better practice rather than a string directly. It signals that the value is defined somewhere else and not just locally.

core/modules/comment/comment.install
stpaultim: @herbdool - Does this look right now?

herbdool: Good

core/modules/comment/comment.install
```diff
@@ -470,6 +470,45 @@ function comment_update_1007() {
...
 
+/**
+ * New setting for comment titles. Replacing comment_subject_field
+ * with comment_title_options.
+ */
+function comment_update_1008() {
+  foreach (config_get_names_with_prefix('node.type.') as $config_name) {
+    $config = config($config_name);
+    $comment_title = $config->get('settings.comment_subject_field');
+    if ($comment_title == COMMENT_TITLE_AUTO) {
```
quicksketch: This comparison does not seem correct. `comment_subject_field` has a value of either `TRUE` or `FALSE`, not a string value like `auto`. I think this is intended to do a check like this:

```suggestion
    $comment_title = $config->get('settings.comment_subject_field');
    if ($comment_title == FALSE) {
```

core/modules/comment/comment.install
```diff
@@ -470,6 +470,45 @@ function comment_update_1007() {
...
+      $config->set('settings.comment_title_options', COMMENT_TITLE_CUSTOM);
+    }
+    $config->save();
+    // Remove old comment_subject_title setting
+    foreach (node_type_get_types() as $type_name => $node_type) {
+      if (isset($node_type->settings['comment_subject_field'])) {
+        unset($node_type->settings['comment_subject_field']);
+      }
+    }
+    node_type_save($node_type);
```
quicksketch: This is unnecessarily looping through all node types within a loop that is already going through all node types. We're already working on the node type configuration above. Before calling `$config->save()`, the old value should be cleared with `$config->clear('settings.comment_subject_field')`.

quicksketch (commented):
There's a little work needed on the update hook yet.

core/modules/comment/comment.install
```diff
@@ -470,6 +470,48 @@ function comment_update_1007() {
...
+    $comment_title = $config->get('settings.comment_subject_field');
+
+    // If the old config value has already been removed, do nothing.
+    if (is_null($comment_title)) {
+      continue;
+    }
+
+    // If comment title was previously enabled, use the custom title option.
+    if ($comment_title) {
+      $config->set('settings.comment_title_options', COMMENT_TITLE_CUSTOM);
```
quicksketch:
```suggestion
      $config->set('settings.comment_title_options', 'custom');
```

core/modules/comment/comment.install
```diff
@@ -470,6 +470,48 @@ function comment_update_1007() {
...
+    }
+    else {
+      // Check to see if contrib module comment_hide_subject is installed and
+      // preserve the setting to hide comment titles entirely.
+      if (module_exists('comment_hide_subject')) {
+        $config->set('settings.comment_title_options', COMMENT_TITLE_HIDDEN);
```
quicksketch:
```suggestion
        $config->set('settings.comment_title_options', 'hidden');
```

core/modules/comment/comment.install
```diff
@@ -470,6 +470,48 @@ function comment_update_1007() {
...
+      }
+      // Otherwise, the previous behavior was to display an automatically
+      // generated title.
+      else {
+        $config->set('settings.comment_title_options', COMMENT_TITLE_AUTO);
```
quicksketch:
```suggestion
        $config->set('settings.comment_title_options', 'auto');
```

quicksketch (commented):
One more thing... update hooks can potentially run even if the comment module is disabled (but not uninstalled). In such a situation, the constants from the module won't be defined. So within the update hook, we should use the string values rather than the constants.

core/modules/comment/comment.module
quicksketch: I think using constants is a good practice, in theory that may reduce typo potential and gives us a place for documentation. However, I found in the .install file we can't (or shouldn't) use constants because the .module file might not always be loaded. So in the update hook we end up using the string values anyway.

PR #4069, opened 2022-05-09
Issue 5620. Clean up views_upgrade.test

### LINKED ISSUE #5620: ViewsUpgradeTestCase may be (partly) obsolete
## Description of the bug

This popped up in an unrelated test-only bug fix:

```
module_enable(array('php'));
```
But that makes no sense, as this module doesn't even exist in Backdrop.

Looking at the test, it seems completely obsolete - it deals with views upgrade (or import?) - not sure which one.

Let's verify that it's really obsolete and remove that test, if that's the case. Otherwise update the code so the check is useful.

## Additional information

- Code: https://github.com/backdrop/backdrop/blob/1.x/core/modules/views/tests/views_upgrade.test

argiepiano:
Fixes https://github.com/backdrop/backdrop-issues/issues/5620


core/modules/views/tests/views_upgrade.test
```diff
@@ -15,11 +15,6 @@ require_once BACKDROP_ROOT . '/core/modules/views/tests/views_query.test';
 class ViewsUpgradeTestCase extends ViewsSqlTest {
   protected function setUp(array $modules = array()) {
     parent::setUp($modules);
-
-    // To import a view the user needs use PHP for settings rights, so enable php module.
-    // @TODO The php module does not exist in core in Backdrop!
-    module_enable(array('php'));
-    $this->resetAll();
   }
```
ghost: My OOP isn't very strong so forgive me if I'm mistaken, but if we remove the custom bit then do we even need to keep this `setUp()` function if all it does is call the parent `setUp()` function? If we remove this function altogether, won't the parent function still be called? Or is that not the case...?

argiepiano: You are right. This override is not needed here. I'll remove it.

ghost:
Thanks. Merged into 1.x and 1.21.x.

PR #4073, opened 2022-05-12
Issue #5624: Updated Welcome Message

### LINKED ISSUE #5624: [UX] Updated welcome message to help folks find the Dashboard
## Help first time users find the dashboard quickly.

During a hack-a-thon a while back @jenlampton and I watched a first time contributor struggle to find basic information that is easily available on the dashboard, but it didn't seem to occur to this person that the dashboard might be a good place to get started. This user opened many admin menu items until we finally had to suggest looking at the dashboard.

In response, @jenlampton opened up [issue #4673](https://github.com/backdrop/backdrop-issues/issues/4673) to automatically redirect folks after installation to the dashboard. There were lots of opinions and little progress on that issue. One of the leading arguments (in my opinion) against that suggestion is the idea that immediately after installation a new site owner might rather see the front page of their website, rather than being taken to an administrative page. @klonos suggested a compromise of adding a one time message to the front page that encouraged new site owners to visit the dashboard. 

This seems like an easy change that can be improved upon at any point AND with the new content on the front page, I'm inclined to agree with this idea. 

I'm opening a new issue to focus on this specific solution in it's simplest form. If this idea is adopted, we might either close the other issue or continue to discuss a more complete solution for future releases. 

I suspect that this simple tip could go a long way towards making the first interaction with Backdrop CMS a more pleasant and productive experience for many people.

Here is a PR with my simple proposal.

BEFORE
image

AFTER:
image

stpaultim:
Fixes: https://github.com/backdrop/backdrop-issues/issues/5624

### DIFF
--- a/core/includes/install.core.inc
+++ b/core/includes/install.core.inc
@@ -1716,7 +1716,7 @@ function install_finished(&$install_state) {
 
   // Confirming that the Dashboard module is enabled. 
   if (module_exists('dashboard')) {
-    $message .= st('<br>If you are new to Backdrop, then the <a href="@dashboard">Dashboard</a> may be a good place to start.', array('@dashboard' => url('admin/dashboard')));
+    $message .= '<br>' . st('If you are new to Backdrop, then the <a href="@dashboard">Dashboard</a> may be a good place to start.', array('@dashboard' => url('admin/dashboard')));
   }
 
   if (!backdrop_is_cli()) {
--- a/core/includes/install.core.inc
+++ b/core/includes/install.core.inc
@@ -1712,7 +1712,8 @@ function install_finished(&$install_state) {
   // Will also trigger indexing of profile-supplied content or feeds.
   backdrop_cron_run();
 
-  $message = st('Thank you for installing @profile!', array('@profile' => backdrop_install_profile_distribution_name()));
+  $profile = backdrop_install_profile_distribution_name();
+  $message = st('Thank you for installing @profile!', array('@profile' => $profile));
 
   // Confirming that the Dashboard module is enabled. 
   if (module_exists('dashboard')) {
--- a/core/includes/install.core.inc
+++ b/core/includes/install.core.inc
@@ -1713,7 +1713,12 @@ function install_finished(&$install_state) {
   backdrop_cron_run();
 
   $message = st('Thank you for installing @profile!', array('@profile' => backdrop_install_profile_distribution_name()));
-  $message .= st(' If you are new to Backdrop, then <a href="@dashboard">the Dashboard</a> may be a good place to start.', array('@dashboard' => url('admin/dashboard')));
+
+  // Confirming that we are using standard profile with Dashboard. 
+  $profile = backdrop_get_profile();
+  if ($profile == 'standard') {
+    $message .= st(' If you are new to Backdrop, then <a href="@dashboard">the Dashboard</a> may be a good place to start.', array('@dashboard' => url('admin/dashboard')));
+  }
 
   if (!backdrop_is_cli()) {
     backdrop_set_message($message);
--- a/core/includes/install.core.inc
+++ b/core/includes/install.core.inc
@@ -1712,11 +1712,11 @@ function install_finished(&$install_state) {
   // Will also trigger indexing of profile-supplied content or feeds.
   backdrop_cron_run();
 
-  $message = st('<p>Thank you for installing @profile!</p>', array('@profile' => backdrop_install_profile_distribution_name()));
+  $message = st('Thank you for installing @profile!', array('@profile' => backdrop_install_profile_distribution_name()));
 
-  // Confirming that we are using standard profile with Dashboard. 
+  // Confirming that the Dashboard module is enabled. 
   if (module_exists('dashboard')) {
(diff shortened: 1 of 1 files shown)


core/includes/install.core.inc
```diff
@@ -1713,6 +1713,7 @@ function install_finished(&$install_state) {
   backdrop_cron_run();
 
   $message = st('Thank you for installing @profile!', array('@profile' => backdrop_install_profile_distribution_name()));
+  $message .= st(' If you are new to Backdrop, then <a href="@dashboard">the Dashboard</a> may be a good place to start.', array('@dashboard' => url('admin/dashboard')));
```
indigoxela: In theory the install profile may not be the "standard" one, and the Dashboard module may not be enabled - in that case the link would be 404.

stpaultim: I've added a profile check. Now trying to figure out how to test it.

core/includes/install.core.inc
```diff
@@ -1712,7 +1712,12 @@ function install_finished(&$install_state) {
   // Will also trigger indexing of profile-supplied content or feeds.
   backdrop_cron_run();
 
-  $message = st('Thank you for installing @profile!', array('@profile' => backdrop_install_profile_distribution_name()));
+  $message = st('<p>Thank you for installing @profile!</p>', array('@profile' => backdrop_install_profile_distribution_name()));
+
+  // Confirming that we are using standard profile with Dashboard.
```
ghost: Need to update this comment.

core/includes/install.core.inc
```diff
@@ -1712,7 +1712,12 @@ function install_finished(&$install_state) {
   // Will also trigger indexing of profile-supplied content or feeds.
   backdrop_cron_run();
 
-  $message = st('Thank you for installing @profile!', array('@profile' => backdrop_install_profile_distribution_name()));
+  $message = st('<p>Thank you for installing @profile!</p>', array('@profile' => backdrop_install_profile_distribution_name()));
+
+  // Confirming that we are using standard profile with Dashboard. 
+  if (module_exists('dashboard')) {
+    $message .= st('<p>If you are new to Backdrop, then <a href="@dashboard">the Dashboard</a> may be a good place to start.</p>', array('@dashboard' => url('admin/dashboard')));
```
ghost: Not sure I like the `<p>` tags, as they add a margin at the bottom whether it's needed or not... Can we do something else here instead? Maybe two `<br>`...?

ghost (changes requested):

stpaultim:
`$message .= st('<br>If you are new to Backdrop, then the <a href="@dashboard">Dashboard</a> may be a good place to start.', array('@dashboard' => url('admin/dashboard')));`

@BWPanda would this be better?

`$message .= st('<br>' . 'If you are new to Backdrop, then the <a href="@dashboard">Dashboard</a> may be a good place to start.', array('@dashboard' => url('admin/dashboard')));`

stpaultim:
Actually, I tried this and like this best.

`$message .= '<br>' . st('If you are new to Backdrop, then the <a href="@dashboard">Dashboard</a> may be a good place to start.', array('@dashboard' => url('admin/dashboard')));`

core/includes/install.core.inc
```diff
@@ -1714,6 +1714,11 @@ function install_finished(&$install_state) {
 
   $message = st('Thank you for installing @profile!', array('@profile' => backdrop_install_profile_distribution_name()));
```
quicksketch: Since this is part of the generic installation code, not part of a profile, we probably want to maintain the messaging that uses a customizable distribution name -- though Backdrop doesn't really support distributions in any official capacity. Perhaps that should be a separate issue. In the mean time, we should probably keep using a variable rather than saying "Backdrop" specifically.

```suggestion
  $profile = backdrop_install_profile_distribution_name();
  $message = st('Thank you for installing @profile!', array('@profile' => $profile));
```

core/includes/install.core.inc
```diff
@@ -1714,6 +1714,11 @@ function install_finished(&$install_state) {
...
 
+  // Confirming that the Dashboard module is enabled. 
+  if (module_exists('dashboard')) {
+    $message .= '<br>' . st('If you are new to Backdrop, then the <a href="@dashboard">Dashboard</a> may be a good place to start.', array('@dashboard' => url('admin/dashboard')));
```
quicksketch:
```suggestion
    $message .= '<br>' . st('If you are new to @profile, then the <a href="@dashboard">Dashboard</a> may be a good place to start.', array('@profile' => $profile, '@dashboard' => url('admin/dashboard')));
```

ghost (approved):

core/includes/install.core.inc
```diff
@@ -1712,7 +1712,13 @@ function install_finished(&$install_state) {
   // Will also trigger indexing of profile-supplied content or feeds.
   backdrop_cron_run();
 
-  $message = st('Thank you for installing @profile!', array('@profile' => backdrop_install_profile_distribution_name()));
+  $profile = backdrop_install_profile_distribution_name();
+  $message = st('Thank you for installing @profile!', array('@profile' => $profile));
+
+  // Confirming that the Dashboard module is enabled. 
+  if (module_exists('dashboard')) {
```
jenlampton, line 1719: Do we also need to check access to the dashboard? Or can we assume that the person who just istalled the site is user 1 and will have access to everything?

quicksketch:

PR #4075, opened 2022-05-15
Issue 5627. Use BlockBroken for unavailable fields

### LINKED ISSUE #5627: Deleting field creates warning and notice with Layouts that use that field in a block
## Description of the bug

This is very similar to #3263 - the difference is that the site doesn't crash in this case. Deleting a field that is displayed as a field block in a layout generates:

`Notice: Undefined index: node-field_test_field in Block->getBlockInfo() (line 198 of /Users/.../Sites/localhost/.../core/modules/layout/includes/block.class.inc). `

`Warning: array_merge(): Expected parameter 2 to be an array, null given in Block->getBlockInfo() (line 198 of /Users/.../Sites/localhost/.../core/modules/layout/includes/block.class.inc).
`

This was reported in a [forum post](https://forum.backdropcms.org/forum/error-after-deleting-node-field).

## Steps To Reproduce
1. Add a field to a content type
2. Create content
3. Create a layout and add the node you created above as Context
4. Place a field block for the field your created above anywhere in the layout
5. Visit the url of the layout to check that it displays correctly
6. Delete the field from the content type
7. Visit the same url above and see the warnings and notices  

## Actual behavior

This produces the notices above.

## Expected behavior

As suggested in #3263, this should not generate php warnings - perhaps it should display a broken block notice 

## Additional information

Backdrop 1.22.0-preview

argiepiano:
Fixes https://github.com/backdrop/backdrop-issues/issues/5627

This approach catches the issue early on, when deciding what class to use to create the Field Block. Additionally, this also catches a warning generated when creating the layout context for rendering.

### DIFF
--- a/core/modules/layout/layout.module
+++ b/core/modules/layout/layout.module
@@ -1728,7 +1728,8 @@ function layout_get_handler_name($plugin_type, $plugin_name) {
   // Block "plugins" for Layouts are wrappers around the generic block system.
   $broken_class = NULL;
   if ($plugin_type === 'block') {
-    list($module, $delta) = explode(':', $plugin_name);
+    $parts = explode(':', $plugin_name);
+    list($module, $delta) = $parts;
     $info = layout_get_block_info($module, $delta);
     if ($info) {
       if (isset($info['class'])) {
@@ -1737,8 +1738,8 @@ function layout_get_handler_name($plugin_type, $plugin_name) {
       else {
         $class_name = 'BlockLegacy';
       }
-      if ($delta == 'field_block') {
-        list( , , $field_name) = explode(':', $plugin_name);
+      if ($delta == 'field_block' && isset($parts[2])) {
+        $field_name = $parts[2];
         if (empty(field_get_block_list()[$field_name])) {
           $class_name = 'BlockBroken';
         }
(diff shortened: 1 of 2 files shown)


argiepiano:
Oops, many broken tests 👎🏽

core/modules/layout/layout.module
```diff
@@ -1737,6 +1738,12 @@ function layout_get_handler_name($plugin_type, $plugin_name) {
       else {
         $class_name = 'BlockLegacy';
       }
+      if ($delta == 'field_block' && isset($parts[2])) {
+        $field_name = $parts[2];
+        if (empty(field_get_block_list()[$field_name])) {
```
indigoxela: I know, we don't have to be compatible with php 5.3 anymore, but syntax-wise this looks odd (function inside function as array). An intermediate variable might look cleaner.

argiepiano: Thanks. I've added an intermediary variable.

argiepiano:
Closing and reopening to trigger Tugboat

core/modules/layout/layout.module
```diff
@@ -1737,6 +1738,13 @@ function layout_get_handler_name($plugin_type, $plugin_name) {
       else {
         $class_name = 'BlockLegacy';
       }
+      if ($delta == 'field_block' && isset($parts[2])) {
```
quicksketch, line 1743:
```suggestion
      // Field blocks use the field name as the third part of the plugin name.
      // If the field no longer exists, return a broken block.
      if ($delta == 'field_block' && isset($parts[2])) {
```

quicksketch:

PR #4080, opened 2022-05-27
Issue 5636. Fix layout relationship so that it works with the layout static cache

### LINKED ISSUE #5482: Fixed: Incorrect context data passed by `LayoutRelationshipAuthorFromNode`.

argiepiano:

IMPORTANT: this PR incorporates all the changes made in #4046. 

I suggest we close #4046. Alternatively @docwilmot  can modify #4046 to incorporate the couple of changes made here.

Explanation:
1. on Zulip @docwilmot had suggested inserting `backdrop_static_reset('getContexts');` in line 1378 or layout.module. This PR does that
2. In addition to that, it was necessary to change line 822 of `layout.class.inc` to remove the first part of the if statement, as this first part always returned true at this point of the code (see opening lines of method `Layout::getContexts()`), effectively canceling the second part of the if - `$include_types & LayoutContext::USAGE_TYPE_RELATIONSHIP`

(See also https://github.com/backdrop/backdrop-issues/issues/5636))


argiepiano:
Apologies for the long string of commits. That's an issue with my fork, but it does not affect the files changed in this PR.

argiepiano:
Closing the old approach in favor of the new one.

PR #4083, opened 2022-06-01
Issue #5482. Fix layout relationships to load correct context

### LINKED ISSUE #5482: Fixed: Incorrect context data passed by `LayoutRelationshipAuthorFromNode`.

argiepiano:

This is a different approach to solving the issue:
- Moved the code that sets the context data using the object stored in the router item.  This was moved from `layout_get_layout_by_path()` to a new method, `EntityLayoutContext::setDataFromRouterItem()`
- Added code to pass the $router_item array as a property of the Layout objects, so that it can be accessed by methods within the class
- Added code to set the data in `Layout::getContexts()` so that the object is fully loaded into the context BEFORE the call to `Layout::getContextsFromRelationships()`

This is a more organic way to approach this issue, rather than the previous attempt that selectively called getContexts() twice.


core/modules/layout/includes/layout.class.inc
```diff
@@ -815,7 +815,17 @@ class Layout {
...
     }
     elseif ($this->path) {
-      $this->contexts += layout_context_required_by_path($this->path);
+      $contexts_required_by_path = layout_context_required_by_path($this->path);
+
+      // Set the contexts' data from the router item if available. When this
+      // method is called from the admin form, there is no router item.
+      if (!empty($this->router_item)) {
+        foreach ($contexts_required_by_path as $context) {
+          $context->setDataFromRouterItem($this->router_item);
```
docwilmot: Context from paths may not all be `EntityLayoutContexts`. Path `node/%/%` would usually be a Node context and a string context.

argiepiano: Thanks, @docwilmot. I found the issue with the test and am working to solve it. I will push some changes probably this evening. I'm very close to making this work.

core/modules/layout/includes/layout.class.inc
```diff
@@ -810,16 +821,25 @@ class Layout {
...
+        $contexts = $this->menu_item->getContexts(); 
+      }
+      else {
+        $contexts = layout_context_required_by_path($this->path);
+      }
+      
+      // Set the contexts' data from the router item.
+      foreach ($contexts as $context) {
+        $context->setDataFromRouterItem($this->router_item);
+      }
```
quicksketch: Is it possible to have a `menu_item` or `path` but not have a `router_item`? For clarity, maybe we should wrap this in an `if` statement, since it doesn't look like `setDataFromRouterItem()` would handle being passed an empty value well.

```suggestion
      // Set the contexts' data from the router item.
      if ($this->router_item) {
        foreach ($contexts as $context) {
          $context->setDataFromRouterItem($this->router_item);
        }
      }
```

core/modules/layout/plugins/context/entity_layout_context_handler.inc
```diff
@@ -100,4 +100,42 @@ class EntityLayoutContext extends LayoutContext {
...
+        // new position, they need to have altered $context->position so that
+        // it points to the right place in the router item to get the data
+        // object.
+        $context_data = $router_item['map'][$this->position];
+        // If the router item contains the context object in the right
+        // position, then we can just set the data from that.
+        if (is_object($context_data)) {
+          $this->setData($context_data);
+        }
+        else { // TODO: this if/else may not be necessary
```
quicksketch: I think this should stay, even though I don't know of any core contexts that load this way. It can happen when a menu router has a path with a generic wildcard like `foo/%` rather than one that uses a specific loader like `foo/%node`. Without using the `%node` placeholder, the router system won't know to call `node_load()` and the value in the `$router_item['map']` would simply be the string from the URL, not the node object. The layout context `load callback` here would potentially take that string and load the full object.

core/modules/layout/plugins/context/entity_layout_context_handler.inc
```diff
@@ -100,4 +100,42 @@ class EntityLayoutContext extends LayoutContext {
...
     }
   }
+
+   /** 
+   * Set the context data from the router item array.
+   * 
+   * @param array $router_item
+   *   A router item array.
+   */
+  function setDataFromRouterItem($router_item) {
```
quicksketch: I don't think there's anything unique to Entities in this method. Router items can load arbitrary data, not only entities. For example, a menu item could be loaded in this way at `admin/structure/menu/item/%menu_link/edit` or a block at `admin/structure/block/manage/%`.

I think this would be better to move up to the generic `LayoutContext` class and omit any override here in `EntityLayoutContext`.

core/modules/layout/includes/layout.class.inc
```diff
@@ -119,6 +119,17 @@ class Layout {
    */
   var $menu_item = NULL;
 
+  /**
+   * A router item array.
+   * 
+   * This property holds a router item array like the one generated by
+   * menu_get_item(). This array is passed to its context objects
+   * to set the data when needed.
```
quicksketch: Slightly more verbose docblock:

```suggestion
   * A router item array as loaded by menu_get_item().
   * 
   * This value may be empty when configuring a Layout or showing an
   * administrative preview. It is only populated when loading the Layout
   * through layout_get_layout_by_path().
```

core/modules/layout/plugins/context/layout_context.inc
```diff
@@ -269,6 +269,14 @@ abstract class LayoutContext extends LayoutHandler {
...
    */
   function getAdminSummary() {}
+
+   /** 
+   * Set the context data from the router item array.
+   * 
+   * @param array $router_item
+   *   A router item array.
+   */
+  function setDataFromRouterItem($router_item) {}
```
quicksketch: Can you move this up under `setData()` so the two methods are next to each other?

quicksketch (commented):
The changes seem incremental and make sense to me. I'll post more feedback in the parent issue.

core/modules/layout/includes/layout.class.inc
argiepiano: Thanks @quicksketch. Rather than doing the check here, the new method `LayoutContext::setDataFromRouterMap`  (notice the change in name in the last push) does the empty check there and catches the empty value. I thought that would keep the code above cleaner, but I can change that if you feel strongly about this.

core/modules/layout/plugins/context/entity_layout_context_handler.inc
argiepiano: The new commit has changed this code. However, I did check what you pointed out, and noticed that the router_item will always contain the needed data :

1. In a custom module I created  the path `foo/%`  with `hook_menu()`. Notice: no autoloader. The callback manually loaded a node with the id provided by `%` and returned a renderable array
2. I overrode the path using Layout, with path `foo/%`. When you create a Layout with `%` you are required to provide a context for that `%`, which means that that particular context (whether it's a node, or a string passtrhough) WILL be included in the `$router_item['map']` array

I checked how this happens. The map is created early on in the page request process. This is the calll stack

- {main}
  - backdrop_bootstrap
    - _backdrop_bootstrap_full
      - menu_set_custom_theme
        - menu_get_custom_theme
          - menu_get_item
            - _menu_translate
              - _menu_load_object
                -  layout_path_load  (which loads all the objects and returns them to be part of the `map`)
  
the call to `layout_path_load` only happens when a Layout overrides a path (as in the `foo/%` override). This function takes care of loading the correct data and returning it to `menu_get_item()` which inserts it in the $router_item array.

So, from what I'm seeing here, your $router_item will contact exactly the objects (or string) you need as specified by the Layout context you chose for `%` in the configuration UI

core/modules/layout/plugins/context/entity_layout_context_handler.inc
argiepiano: I have remove the overridden method. `LayoutContext::setDataFromRouterMap()` now handles all the context data setting for both entity contexts and string passthrough.

core/modules/layout/includes/layout.class.inc
argiepiano: In the latest commit this property has changed name to $router_map (since it receives the  map sub-array). I've modified the doc block to closely match yours.

core/modules/layout/plugins/context/layout_context.inc
argiepiano: Done

PR #4086, opened 2022-06-06
Add help text for enabled components in admin bar theme settings

### LINKED ISSUE #5284: Rename Admin Bar components and add descriptions
**Description of the bug**

There is an Administration Bar menu setting that allows you to disable the "Home" menu in the admin toolbar. However, the option is labelled "Icon Menu." I believe that this is a holdover from Drupal 7, when this menu was the Druplicon. 

icon-menu

**Steps To Reproduce**

To reproduce the behavior:
1. Go to 'admin/config/administration/admin-bar'
2. Disable "Icon Menu" and save
3. Notice that the Home menu disappears

**Expected behavior**

I believe that the behavior is fine, but the label is incorrect and unhelpful. If I didn't remember Drupal 7, I would have no idea what the "Icon Menu" is.

Here is screenshot of Drupal 7 admin toolbar

D7-icon-menu

I believe that the only thing necessary here is to change the label to "Home Menu". I might also suggested removing this option completely, but I guess that would probably be a 2.0 issue, right?

RakeshPotnuru:
Fixes https://github.com/backdrop/backdrop-issues/issues/5284

image

### DIFF
--- a/core/modules/admin_bar/admin_bar.inc
+++ b/core/modules/admin_bar/admin_bar.inc
@@ -801,6 +801,13 @@ function admin_bar_theme_settings() {
     ),
   );
 
+  $form['components']['admin_bar.icon']['#description'] = t('A specialized menu that links to the home page and provides links to perform various other admin tasks (e.g.: clear caches, run cron, etc.).');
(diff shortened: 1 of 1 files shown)


quicksketch:
Merged into 1.x and 1.22.x. Thanks again @RakeshPotnuru!

PR #4094, opened 2022-06-16
Remove autofocus from node form title

### LINKED ISSUE #5386: Get rid of the distracting autofocus on node form titles
## Description of the bug

In Issue #2625 the autofocus on the node title field was introduced in node forms.
This led to a UX regression. Now every time someone opens a node form, the focus jumps to the title field.

That's (sort of) OK with new nodes, but when editing existing nodes, the title is one of the least likely candidate for changes - usually something in one of the other fields has to get updated (body or paragraphs...).

Additionally, the autofocus happens on the title field, regardless of its weight, which can cause irritating jumps, when the title field is further down the form.

There's already an issue for the weight (#3583), but I'd like to start a discussion on completely reverting this UX regression (IMO the autofocus on node titles is a regression).

Note that getting rid of it is only a matter of hook_form_alter() - which I use currently. But I'm pretty sure, I'm not the only one who thinks this autofocus is bad UX. So I start this as a question here.

olafgrabienski:
Fixes https://github.com/backdrop/backdrop-issues/issues/5386.

### DIFF
--- a/core/modules/node/node.module
+++ b/core/modules/node/node.module
@@ -3420,9 +3420,6 @@ function node_content_form(Node $node, $form_state) {
       '#default_value' => $node->title,
       '#maxlength' => 255,
       '#weight' => -5,
-      '#attributes' => array(
-        'autofocus' => TRUE,
-      ),
     );
   }
 


quicksketch:

PR #4100, opened 2022-06-22
Issue #5294: Hide Export links when Config module isn't installed.

### LINKED ISSUE #5294: [UX] Fixed: Export links are shown even when the Configuration Manager module is disabled.
**Description of the bug**

This is a minor issue, but from regular Backdrop user's perspective looks like a bug. When _Configuration Manager_ module is disabled, the Taxonomy page still shows the vocabulary export button pressing which gives no any warning or notice why user can not export.

**Steps To Reproduce**

To reproduce the behavior:
1. Go to /admin/structure/taxonomy > drop down _List Terms_ menu > export:

export

2. just to see:

export2

**Expected behavior**

Depending what the consensus on this will be either export link needs to be disabled (if _Configuration Manager_  is also disabled) or the next page needs to display an explanation to user what's going on and what s/he needs to do to be able to export taxonomy vocabularies.

ghost:
Fixes https://github.com/backdrop/backdrop-issues/issues/5294

### DIFF
--- a/core/modules/block/block.admin.inc
+++ b/core/modules/block/block.admin.inc
@@ -40,7 +40,7 @@ function block_admin_list() {
         'href' => 'admin/structure/block/manage/' . $delta . '/translation',
       );
     }
-    if (user_access('synchronize configuration')) {
+    if (module_exists('config') && user_access('synchronize configuration')) {
       $links['export'] = array(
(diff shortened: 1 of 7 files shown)


quicksketch:
Thanks! Merged into 1.x and 1.22.x.

PR #4103, opened 2022-06-24
Issue #3309: Remove tabs from user login pages.

### LINKED ISSUE #3309: [WP][UX] Remove "tabs" from log-in, register, and password reset pages.
A follow-up to https://github.com/backdrop/backdrop-issues/issues/3108

I'd really like to redesign the log-in page to remove the "tabs" entirely. The visual "tabs" are a strange way to navigate these three pages. Plus, they really give away that a site is Drupal/Backdrop.

I almost always remove these tabs in my custom themes, and replace them with a markup element in the form containing links to those pages instead. I'd love to see us do that by default for all Backdrop sites.

ghost:
Fixes https://github.com/backdrop/backdrop-issues/issues/3309

### DIFF
--- a/core/modules/user/user.module
+++ b/core/modules/user/user.module
@@ -1350,7 +1350,6 @@ function user_menu() {
     'page callback' => 'backdrop_get_form',
     'page arguments' => array('user_login'),
     'access callback' => 'user_is_anonymous',
-    'type' => MENU_DEFAULT_LOCAL_TASK,
   );
 
   $items['user/register'] = array(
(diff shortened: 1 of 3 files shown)


core/modules/user/user.module
```diff
@@ -1350,23 +1350,20 @@ function user_menu() {
     'page callback' => 'backdrop_get_form',
     'page arguments' => array('user_login'),
     'access callback' => 'user_is_anonymous',
-    'type' => MENU_DEFAULT_LOCAL_TASK,
```
indigoxela, line 1353: If you remove the type, the default is MENU_NORMAL_ITEM, which means that it will automatically show up.

Maybe MENU_CALLBACK, possibly MENU_SUGGESTED_ITEM is the better choice here.

PR #4104, opened 2022-06-24
Issue #3206: Add multisite check to telemetry.

### LINKED ISSUE #3206: Telemetry: Gather data to determine if multisites are popular or not
After commenting here https://github.com/backdrop/backdrop-issues/issues/3205, I'm wondering what people think of Backdrop's multisite feature...? Do you use it? If not, why not?

Apparently multisites aren't very popular (as indicated by @quicksketch's comment from https://github.com/backdrop/backdrop-issues/issues/2555#issuecomment-399227170 and my own experience with needing to ask for multisite support from modules like Drush).
But I think it's awesome! Allows me to have only one instance of Backdrop (for easier core updates), and I can put modules/themes/layouts that are common to all my sites in one place, rather than duplicating them.

Interested to hear what others think. Are there good reasons not to use multisites that I'm unaware of or ignoring? Or are people just not familiar with the concept? Feedback please!

ghost:
Fixes https://github.com/backdrop/backdrop-issues/issues/3206


quicksketch:
Merged into 1.x and 1.26.x. Thanks @BWPanda!

PR #4107, opened 2022-06-25
Issue #5185: Add core modules to telemetry.

### LINKED ISSUE #5185: Telemetry: Collect enabled core modules
This is part of the #285 meta.

ghost:
Fixes https://github.com/backdrop/backdrop-issues/issues/5185

### DIFF
--- a/core/modules/telemetry/telemetry.telemetry.inc
+++ b/core/modules/telemetry/telemetry.telemetry.inc
@@ -28,6 +28,11 @@ function telemetry_telemetry_info() {
     'description' => t('The web server that delivers HTTP pages to the web browser, such as Apache or nginx.'),
     'project' => 'backdrop',
   );
+  $info['modules_core'] = array(
+    'label' => t('Core modules'),
+    'description' => t('A list of core modules enabled on your site.'),
+    'project' => 'backdrop',
+  );
   return $info;
 }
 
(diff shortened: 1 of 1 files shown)


core/modules/telemetry/telemetry.telemetry.inc
```diff
@@ -55,5 +60,15 @@ function telemetry_telemetry_data($key) {
       else {
         return 'Other';
       }
+    case 'modules_core':
+      $modules = array();
+      $all_modules = system_rebuild_module_data();
+      foreach (module_list() as $module) {
+        if (strpos($all_modules[$module]->uri, 'core/') === 0) {
```
jenlampton, line 67: is this the best check we have to see if something is a core module? It seems as though someone could add a module in this location on their site and it would then report back as a core module. (I'm not sure if this is a legitimate concern)

klonos: I vaguely recall doing something smart for a PR for either https://github.com/backdrop/backdrop-issues/issues/498 or some similar/related issue, but I cannot find anything now 😞

klonos:
During this week's dev meeting, we've agreed that we should only collect metrics for disabled core modules. Also, we decided that an array of modules is not useful/ideal, as that would result in metrics that are combinations of modules, whereas we need individual metrics for each (specific, "of interest") module.

As per the above, I will go ahead and close this PR (which sends an array of all core modules).

PR #4124, opened 2022-07-03
Issue #5669: Properly indicate the "required" status

### LINKED ISSUE #5669: [A11Y] Properly indicate the "required" status of a form input
## Description of the bug

Currently Backdrop indicates required status of a form item:

- By an asterisk
- By a css class to style things
- By a title attribute on the asterisk

However, that's insufficient re web accessibility, when it comes to WCAG.

A `*` on its own means nothing, "title" attributes won't get read aloud by screen readers (at least in their default setting). EDIT wait, it's an `abbr` tag, so it would get read, but in an odd and distracting way.

What we need instead:

- Add a "required" html attribute to the input, select and textfield input, so assistive technologies can pick up that info
- For screen readers get rid of the "asterisk only" indicator display, provide actual text instead

## Additional information

- https://www.w3.org/WAI/GL/WCAG20/WD-WCAG20-TECHS/H90.html
- https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes/required
- https://stackoverflow.com/questions/45776312/how-do-screen-readers-read-abbr-tags

Affected functions: _form_set_class(), theme_form_required_marker().

See also: https://www.w3.org/WAI/WCAG21/quickref/?showtechniques=413#input-assistance

indigoxela:
Fixes https://github.com/backdrop/backdrop-issues/issues/5669

### DIFF
--- a/core/modules/simpletest/tests/form.test
+++ b/core/modules/simpletest/tests/form.test
@@ -66,7 +66,7 @@ class FormsTestCase extends BackdropWebTestCase {
     $elements['file']['empty_values'] = $empty_strings;
 
     // Regular expression to find the expected marker on required elements.
-    $required_marker_preg = '@<label.*<abbr class="form-required" title="This field is required\.">\*</abbr></label>@';
+    $required_marker_preg = '@<label.*<abbr class="form-required" title="This field is required\." aria-hidden="true">\*</abbr>.*</label>@';
 
     // Go through all the elements and all the empty values for them.
     foreach ($elements as $type => $data) {
--- a/core/includes/form.inc
+++ b/core/includes/form.inc
@@ -2822,6 +2822,9 @@ function form_process_select($element) {
 function theme_select($variables) {
   $element = $variables['element'];
   element_set_attributes($element, array('id', 'name', 'size'));
+  if ($element['#required']) {
+    $element['#attributes']['required'] = 'required';
+  }
   _form_set_class($element, array('form-select'));
 
   return '<select' . backdrop_attributes($element['#attributes']) . '>' . form_select_options($element) . '</select>';
@@ -3003,6 +3006,9 @@ function theme_radio($variables) {
 
   if (isset($element['#return_value']) && $element['#value'] !== FALSE && $element['#value'] == $element['#return_value']) {
     $element['#attributes']['checked'] = 'checked';
+  }
+  if ($element['#required']) {
+    $element['#attributes']['required'] = 'required';
   }
   _form_set_class($element, array('form-radio'));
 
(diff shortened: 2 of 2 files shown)


yorkshire-pudding:
@indigoxela - is it documented somewhere the correct/official form of the required HTML attribute?
The `required="required"` used appears [here](https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes/required)

However [this site](https://www.w3schools.com/tags/att_required.asp) and [this site](https://www.geeksforgeeks.org/html-input-required-attribute/) simply have `required`.

I think both work, but not sure if one version is more accepted than another.  This [stackoverflow discussion](https://stackoverflow.com/questions/18770369/how-to-set-html5-required-attribute-in-javascript) suggests that `required="true"` may be valid, which from a code readability perspective may be clearer.

indigoxela:
> @indigoxela - is it documented somewhere the correct/official form of the required HTML attribute? The `required="required"` used appears [here](https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes/required)


@yorkshire-pudding I'm relying on this info: https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes/required. I'm aware of discussions re simpler or more readable forms.

yorkshire-pudding:
I think this is functionally great.  I personally think `required="true"` is clearer for someone looking at the code and would prefer that, but I wouldn't let it stand in the way of an otherwise great addition to accessibility

indigoxela:
Hello random buddy "ddd_installer_test".

indigoxela:
And now hello to "Unable to find the wrapper "public"" :roll_eyes:

core/modules/simpletest/tests/form.test
```diff
@@ -66,7 +66,7 @@ class FormsTestCase extends BackdropWebTestCase {
     $elements['file']['empty_values'] = $empty_strings;
 
     // Regular expression to find the expected marker on required elements.
-    $required_marker_preg = '@<label.*<abbr class="form-required" title="This field is required\.">\*</abbr></label>@';
+    $required_marker_preg = '@<label.*<abbr class="form-required" title="This field is required\." aria-hidden="true">\*</abbr>.*</label>@';
```
klonos, line 69: Should we make this more explicit, by also adding the newly-introduced `<span>` to the regex?:

```suggestion
    $required_marker_preg = '@<label.*<abbr class="form-required" title="This field is required\." aria-hidden="true">\*</abbr><span class="element-invisible">.*</span></label>@';
```

indigoxela: Sure, let's try that.

yorkshire-pudding:
I've reviewed this again and it all looks good.  

The addition to regex in the test looks good.

indigoxela:
@yorkshire-pudding many thanks for reviving. It was time for a rebase, as the Tugboat sandbox has expired.

yorkshire-pudding:
I see phpcs strikes again.  As none of these apply to the changes, I don't think this should block.  Also the 5.6 test failure seems unrelated to the change.  I've tested this updated PR an all looks good.

indigoxela:
Closing  because of side effects of this approach.

PR #4126, opened 2022-07-04
Issue #5673: Allow Tugboat PR comments to be disabled. [skip tests]

### LINKED ISSUE #5673: Allow Tugboat PR comments to be disabled
When we make a change to our Tugboat config (https://github.com/backdrop/backdrop/blob/1.x/.tugboat), the base preview for the 1.x branch needs to be rebuilt so that the changes take effect. Doing this causes all PR sandboxes to expand in size because they're no longer linked to the base preview. This causes us to go over quota (ad means no more sandboxes can be created). Fixing this is simply a matter of rebuilding each PR sandbox (something that can be done automatically when rebuilding the base preview), however this triggers a new comment to be posted to each PR (with the new login details of the sandbox) which is very noisy (lots of people get a sudden influx of notifications/emails).

To avoid this, I want to try creating an environment variable in Tugboat that's set to TRUE by default. I'll then update the comment-posting code to exit if this variable is FALSE. That way, I hope to be able to better control when comments are posted or not.

For example, if a change is needed I can push the change to the repo, then update the env. variable in Tugboat to FALSE. I then then rebuild all previews which will import the change, prevent us going over quota, but not post updates to the PRs. I can then re-enable the env. variable.

I'll test all that in a PR here first...

ghost:
Attempt to fix https://github.com/backdrop/backdrop-issues/issues/5673 (purposefully not linking to that issue to avoid it being closed when this is merged as further changes might be needed)...

### DIFF
--- a/.tugboat/delete-comments.sh
+++ b/.tugboat/delete-comments.sh
@@ -4,6 +4,12 @@
  * Delete any previous comments in the PR about this preview.
  */
 
+// Don't run this if comment updating is disabled.
+// @see https://dashboard.tugboatqa.com/5f9cffcf6c879bf4534e573e/settings/
+if (!getenv('UPDATE_COMMENTS')) {
+  return;
+}
+
 $curl_header = array(
   'Authorization: token ' . getenv('BACKDROP_GITHUB_TOKEN'),
   'Content-Type: application/json',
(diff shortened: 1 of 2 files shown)


ghost:
Forgot the `[skip tests]` line in the PR title.

PR #4128, opened 2022-07-07
Issue #5677: Allows themes to add JS files to footer.

### LINKED ISSUE #5677: [DX] Allow theme JavaScript to specify header/footer scope.
## Description of the need

When adding a JS file to your theme via the `.info` file, it's automatically added to the header (i.e. between the `<head>` tags of the HTML). If you instead want to add it to the footer (i.e. just before the closing `<body>` tag in the HTML) you need to instead add the JS via `backdrop_add_js()` in something like `THEME_preprocess_page()`, which allows you to specify `'scope' => 'footer'`.

## Proposed solution

It'd be nice if we could instead do something like this in the `.info` file:

```
scripts[footer][] = js/my_footer.js
```

That'd be much simpler and, IMO, the more expected solution.

The current solution works well enough, it's just not as easy/obvious.

## Additional information

Here's a question in the forum of someone expecting this functionality but finding it didn't work: https://forum.backdropcms.org/forum/how-add-custom-js-template-head-or-footer

Backdrop now allows themes to add javascript files to the footer directly from the `.info` file.

ghost:
Fixes https://github.com/backdrop/backdrop-issues/issues/5677


core/includes/theme.inc
```diff
@@ -184,35 +179,53 @@ function _backdrop_theme_initialize($theme, $base_themes = array(), $registry_ca
...
+        }
       }
     }
   }
 
   // Add scripts used by this theme.
   if (!empty($theme->scripts)) {
-    foreach ($theme->scripts as $name => $script) {
-      $final_scripts[$name] = $script;
+    foreach ($theme->scripts as $key => $value) {
```
indigoxela, line 211: Unless I'm missing something, this allows any random string for $key, be it typo or misunderstanding... how about using switch instead?

Rough example:
```
case 'footer':
  // put it in footer
default:
  // put it in header, when empty, or header, or whatever string...
```
Or maybe catching things like that isn't what you want? It's just a suggestion. :wink: 

EDIT: I just realized that backdrop_get_js() catches that, anyway (page.tpl.php).

ghost: Also, the documentation for `backdrop_add_js()` says of the `scope` parameter:

> The location in which you want to place the script. Possible values are 'header' or 'footer'. **If your theme implements different regions, you can also use these.**

So this could be any string.

core/includes/theme.inc
```diff
@@ -133,22 +133,17 @@ function backdrop_theme_initialize() {
...
  * @param array $base_themes
- *    An optional array of objects that represent the 'base theme' if the
- *    theme is meant to be derivative of another theme. It requires
- *    the same information as the $theme object. It should be in
- *    'oldest first' order, meaning the top level of the chain will
- *    be first.
+ *   An optional array of objects that represent the 'base theme' if the theme
+ *   is meant to be derivative of another theme. It requires the same
+ *   information as the $theme object. It should be in 'oldest first' order,
+ *   meaning the top level of the chain will be first.
```
klonos, line 146: Nice 👍🏼

core/includes/theme.inc
```diff
@@ -184,35 +179,53 @@ function _backdrop_theme_initialize($theme, $base_themes = array(), $registry_ca
...
+      foreach ($base->scripts as $key => $value) {
+        if (is_string($value)) {
+          // Empty scope defaults to 'header'.
+          $final_scripts['header'][$key] = $value;
+        }
+        else {
+          foreach ($value as $name => $script) {
+            $final_scripts[$key][$name] = $script;
+          }
+        }
```
klonos, line 204: @BWPanda we seem to be duplicating this code for base themes and then also for the current theme. Couldn't we compile an `$all_themes` array, which is the `$base_themes` plus the `$theme`, and then do this iteration only once?

Same for the stylesheets further up in this function.

Feel free to ignore this if you think that that would be unnecessary refactoring.

klonos: ...actually this duplication of code happens again in the theme engine section later in the function. So perhaps this can be a follow up if you find it too much.

ghost: Hmm, good point. But I think a separate issue for refactoring this function might be better.

PR #4130, opened 2022-07-09
Issue #5678: Allow Link fields to restrict the URL type.

### LINKED ISSUE #5678: Allow Link fields to be restricted to only internal or external URLs
## Description of the need

If you want to add a Link field to a content type for storing URLs to external websites (e.g. like on the BackdropCMS.org Showcase nodes), users can still enter URLs for internal pages. Additionally, there's an autocomplete that displays a list of matching pages/content on your site (you may not want users seeing this list of content in this situation).

**Advocate:** @olafgrabienski

- Update 31 Aug 2023: see [#issuecomment-1701766330](https://github.com/backdrop/backdrop-issues/issues/5678#issuecomment-1701766330)

## Proposed solution

It'd be nice if there was an option for Link fields to restrict them to either external or internal URLs (or both, as is the situation currently). Then you wouldn't have the autocomplete showing when you only wanted users entering external URLs.

Drupal has the [Advanced Link](https://www.drupal.org/project/advanced_link) contrib module that achieves this, however it also does a bunch of other things too.

## Additional information

Here's a screencast showing the issue on Backdrop's Showcase node form:

Peek 2022-07-09 22-25

Backdrop now optionally allows Link fields to be restricted to just external or internal URLs.

ghost:
Fixes https://github.com/backdrop/backdrop-issues/issues/5678


core/modules/link/link.module
```diff
@@ -420,6 +439,20 @@ function _link_validate(&$item, $delta, $field, $entity, $instance, $langcode, &
         'error_element' => array('url' => FALSE, 'title' => TRUE),
       );
     }
+    // Validate the URL type.
+    if ($instance['settings']['type'] != LINK_BOTH) {
```
indigoxela, line 439: I guess, this should only happen, ` if ($instance['settings']['validate_url']) {...`, see the code block above.

core/modules/link/link.module
```diff
@@ -75,6 +77,18 @@ function link_field_instance_settings_form($field, $instance) {
     '#description' => t('If checked, the URL field is optional and submitting a title alone will be acceptable. If the URL is omitted, the title will be displayed as plain text.'),
   );
 
+  $form['type'] = array(
```
indigoxela, line 80: This form item should probably be (somehow) connected with `$form['validate_url']` above. Maybe use "#states"?

core/modules/link/link.module
```diff
@@ -335,6 +349,11 @@ function link_field_widget_form(&$form, &$form_state, $field, $instance, $langco
     '#type' => $instance['widget']['type'],
     '#default_value' => isset($items[$delta]) ? $items[$delta] : '',
   );
+  if ($instance['settings']['type'] != LINK_BOTH) {
+    $element['#description'] .= ' ' . t('Only %type URLs are allowed.', array(
```
indigoxela: Should this always be there - even if the site builder decided to not add any help text?

Anyway, it probably should only show up, if validation in general is turned on.

quicksketch: We should probably separate this out into two separate strings so it can be fully translated. Right now "internal" and "external" would be used regardless of translation. I would type out the entire string to provide context to translators:

```
$instance['settings']['type'] == LINK_INTERNAL ? t('Only internal paths are allowed.') : t('Only full URLs are allowed.);
```

Though I might lean towards not putting any automatic help text at all, like @indigoxela suggested.

core/modules/link/link.module
```diff
@@ -420,6 +439,20 @@ function _link_validate(&$item, $delta, $field, $entity, $instance, $langcode, &
         'error_element' => array('url' => FALSE, 'title' => TRUE),
       );
     }
+    // Validate the URL type.
+    if ($instance['settings']['type'] != LINK_BOTH) {
+      $url_type = link_validate_url(trim($item['url']));
+      if (($instance['settings']['type'] == LINK_INTERNAL && $url_type === LINK_EXTERNAL) || ($instance['settings']['type'] == LINK_EXTERNAL && in_array($url_type, array(LINK_INTERNAL, LINK_FRONT), TRUE))) {
+        $errors[$field['field_name']][$langcode][$delta][] = array(
+          'error' => 'link_required',
+          'message' => t('Only %type URLs are allowed in %field.', array(
```
quicksketch: This error message has the same problem as above where "internal" and "external" needs to be translatable. This would probably be best as two strings based on the link type.

core/modules/link/link.module
```diff
@@ -75,6 +77,18 @@ function link_field_instance_settings_form($field, $instance) {
     '#description' => t('If checked, the URL field is optional and submitting a title alone will be acceptable. If the URL is omitted, the title will be displayed as plain text.'),
   );
 
+  $form['type'] = array(
+    '#type' => 'radios',
+    '#title' => t('URL type'),
```
klonos:
```suggestion
    '#title' => t('Allowed URL types'),
```

core/modules/link/link.module
```diff
@@ -75,6 +77,18 @@ function link_field_instance_settings_form($field, $instance) {
...
+    '#default_value' => $instance['settings']['type'],
+    '#options' => array(
+      LINK_INTERNAL => t('Internal only'),
+      LINK_EXTERNAL => t('External only'),
+      LINK_BOTH => t('Both (internal and external)'),
+    ),
+    '#description' => t('The type of URLs this field will allow. Internal URLs point to pages on this website; an autocomplete displays matching pages to select from. External URLs point to other websites; no autocomplete is used.'),
+  );
```
klonos: The "The type of URLs this field will allow." bit does not provide any real value - we can move the word "allow"/"allowed" to be part of the radio set label, which will have the same effect.

Then can we please add individual, shorter descriptions to the various options, instead of a single lengthy one?:
```suggestion
  );
  $form['type'][LINK_INTERNAL]['#description'] = t('Internal URLs point to pages on this website; an autocomplete displays matching pages to select from.');
  $form['type'][LINK_EXTERNAL]['#description'] = t('External URLs point to other websites; no autocomplete is used.');
```

core/modules/link/link.module
ghost: Disagree, as per discussion in https://github.com/backdrop/backdrop-issues/issues/5678.

core/modules/link/link.module
ghost: As above.

core/modules/link/link.module
ghost: > Should this always be there - even if the site builder decided to not add any help text?

Not sure... Happy to hear what others have to say. Discuss in https://github.com/backdrop/backdrop-issues/issues/5678 ?

ghost: Ah, nevermind. Just saw @quicksketch's comment below. Will remove this line then.

klonos: [moving comment in the issue, for more visibility/feedback]

quicksketch:
Merged into 1.x for 1.26.0.

PR #4136, opened 2022-07-15
5421 improve cron page

### LINKED ISSUE #5413: Improve clarity of help information on the cron settings page
The current cron settings page at admin/config/system/cron is not as clear as it could be. Here's the current page:

before

Issues:

* It's not clear from this page that the "Cron configuration" is something that will be overridden by any external cron job.
* It's not clear that if an external cron job is set up, this won't be needed.
* Terminology: the user guide calls this "automated cron" (versus "native cron"); we should be consistent with what it's called.
* Terminology: the User Guide doesn't call itself a tutorial, and the linked page isn't a tutorial page, it's a documentation page in the User Guide.

### LINKED ISSUE #5421: [META][UX] Improve the admin/config/system/cron page
Before:

After:

Suggested improvements:
- [ ] Move the "Last run" indication to the top of the page, and make it be a status/warning/error, based on the same logic as the "Cron maintenance tasks" entry in `admin/reports/status` (see if we can reduce code and translatable string duplication along the way).
- [ ] #5413 | Improve clarity of help information on the cron settings page.
- [ ] Distinctly list the various methods to run cron (separate fieldsets, along with help text as needed).
- [ ] Convert the "Never" option in the "Run cron every" dropdown into a separate checkbox that enables/disables the feature - in other words, make it so that this bit in our [documentation page](https://backdropcms.org/cron) is not needed:
   > ## Disabling "automated cron"
   > For performance reasons, or if you want to ensure that cron can only ever run from an external trigger, it may be desirable to disable the automated cron system.
   > 
   > You can disable it by setting the "Run cron every" value to "Never" (e.g., at Administration > Configuration > System > Cron (admin/config/system/cron).
- [ ] Convert the hard-coded options in the "Run cron every" dropdown into a "natural language" set of settings (number field + dropdown for selecting minutes/hours/days/weeks)

bugfolder:
Fixes https://github.com/backdrop/backdrop-issues/issues/5421.
Fixes https://github.com/backdrop/backdrop-issues/issues/5413.


core/modules/system/system.admin.inc
```diff
@@ -1566,41 +1566,99 @@ function system_site_information_settings_submit($form, &$form_state) {
...
+  $form['last_run'] = array(
+    '#type' => 'help',
+    '#markup' => t('Last run: %cron-last ago.', array(
+      '%cron-last' => format_interval(REQUEST_TIME - state_get('cron_last')),)),
+  );
+
   $form['description'] = array(
     '#type' => 'help',
-    '#markup' => t('Cron takes care of running periodic tasks like checking for updates and indexing content for search.'),
+    '#markup' => t('Cron takes care of running periodic tasks like checking for updates and indexing content for search. The preferred way to run such tasks is by <a href="https://en.wikipedia.org/wiki/Cron">setting up native cron jobs</a>. If that is not an available option with your hosting provider, then the following methods are alternatives.'),
```
jenlampton: I don't think we should be linking to wikipedia. In fact, I don't think any of this new text adds value. "The following" also includes the link thats necessary for native cron jobs, so I worry that this might confuse people more than it helps.

core/modules/system/system.admin.inc
```diff
@@ -1566,41 +1566,99 @@ function system_site_information_settings_submit($form, &$form_state) {
 function system_cron_settings($form, &$form_state) {
   global $base_url;
   $form['#config'] = 'system.core';
+  $config = config('system.core');
+
+  $form['last_run'] = array(
+    '#type' => 'help',
```
jenlampton: This isn't help. Let's leave this type as `'#markup'`.

core/modules/system/system.admin.inc
```diff
@@ -1566,41 +1566,99 @@ function system_site_information_settings_submit($form, &$form_state) {
...
+    '#default_value' => $config->get('cron_interval_units'),
+    '#options' => array(
+      60 => t('minutes'),
+      3600 => t('hours'),
+      86400 => t('days'),
+      604800 => t('weeks'),
+    ),
+  );
+  $form['automatic']['interval']['cron_interval_post'] = array(
+    '#markup' => t('ago, run cron on the next page request.'),
```
jenlampton, line 1644: This doesn't translate cleanly. 
Let's move the "ago" to a `field_suffix` on the previous element. This can be more of a complete sentence: `t('run cron on the next page request.');`

core/modules/system/system.install
```diff
@@ -574,7 +574,10 @@ function system_requirements($phase) {
             break;
         }
         if ($change_notice) {
-          $description .= $t('Backdrop core provides a bundled @name module. A different copy of @name module is at %path. Remove this module from your installation to use the core @name module. For more information see the <a href="!url" target="_blank">@name change notice</a>.', array('@name' => $file->info['name'], '%path' => $path, '!url' => 'https://docs.backdropcms.org/node/' . $change_notice));
+          $description .= $t('Backdrop core provides a bundled @name module. A different copy of @name module is at %path. Remove this module from your installation to use the core @name module. For more information see the <a href="!url" target="_blank">@name change notice</a>.', array(
```
jenlampton, line 577: This change looks unrelated to this issue (and since it won't be QA'd as part of this issue, we shouldn't try to squeeze it in here)

core/modules/system/system.install
```diff
@@ -3343,6 +3346,78 @@ function system_update_1085() {
...
+  $config = config('system.core');
+  $threshold = $config->get('cron_safe_threshold');
+  // These are the former values of the cron_safe_threshold select options.
+  $enabled = !!$threshold;
+  switch($threshold) {
+    case 0:
+      $num = 0;
+      $units = 60;
+      break;
+
```
jenlampton, line 3363: We can probably remove some of these blank lines in the switch.

core/modules/comment/tests/comment.test
```diff
@@ -74,7 +74,7 @@ class CommentHelperCase extends BackdropWebTestCase {
 
     // Disable safety cron run that can affect the updating of comment "new"
     // status during shutdown functions when cron takes a long time.
-    config_get('system.core', 'cron_safe_threshold', 0);
+    config_set('system.core', 'auto_cron_enabled', 0);
```
jenlampton, line 77: I don't think we should be changing the variable name. This could cause a lot of problems if someone messes up a config deploy, and cron stops running on their site. I am going to work on an alternative PR with the same UI changes, but no changes to config. 

Also, why the change from get to set?  Is that a bug that should be fixed in it's own issue? 
edit: see https://github.com/backdrop/backdrop-issues/issues/6215

core/modules/system/system.admin.inc
bugfolder: OK, removed.

core/modules/system/system.admin.inc
bugfolder: Left as `#markup`.

core/modules/system/system.install
```diff
@@ -3343,6 +3346,78 @@ function system_update_1085() {
   }
 }
 
+/**
+ * Initialize the number and units of cron setting from the existing possible
+ * values.
+ */
+function system_update_1092() {
```
jenlampton, line 3353: no update hook needed if the original variable name stays.

bugfolder:
Closing in favor of https://github.com/backdrop/backdrop/pull/4512 (or its successors).

PR #4137, opened 2022-07-16
Issue #5684: Make alterable queries more robust against column ambiguity

### LINKED ISSUE #5684: Fixed: Make `path_autocomplete()` alterable queries more robust against column ambiguity.
path_autocomplete fails when acl, forum, and forums_access are installed. You get the following error on any field where path_autocomplete is called to provide suggestions. You receive the following log error:

```
PDOException: SQLSTATE[23000]: Integrity constraint violation: 1052 Column 'name' in where clause
is ambiguous:
SELECT DISTINCT taxonomy_term_data.`tid` AS `tid`, taxonomy_term_data.`vocabulary`
AS `vocabulary`, taxonomy_term_data.`name` AS `name` FROM {taxonomy_term_data} taxonomy_term_data
LEFT OUTER JOIN {forum_access} fa ON fa.tid = taxonomy_term_data.tid LEFT OUTER JOIN {acl} acl
ON acl.number = taxonomy_term_data.tid AND acl.module = 'forum_access' LEFT OUTER JOIN {acl_user} aclu
ON aclu.acl_id = acl.acl_id WHERE (`name` LIKE :db_condition_placeholder_0 ESCAPE '\\')
AND( (`fa`.`rid` IS NULL ) OR (`fa`.`grant_view` >= :db_condition_placeholder_1)
OR (`aclu`.`uid` IS NOT NULL ) )
ORDER BY tid DESC LIMIT 8 OFFSET 0; Array ( [:db_condition_placeholder_0] => %community%
[:db_condition_placeholder_1] => 1 ) in path_autocomplete()
(line 4226 of backdrop/core/modules/system/system.module).
```

The problem is that taxonomy_term_data and acl are joined in the query, they both have a column named 'name', and an unqualified 'name' is used in the where clause.

To reproduce the behavior:

1. Install and enable acl, forum, and forum_access on Backdrop.
2. Go to Configuration > URL Handling > URL Aliases > Add URL Alias.
3. Type a character in the "System existing path"  input and wait.
4. The autocomplete will trigger and you get an error.

I was able to fix this problem in Backdrop 1.22.0 by editing my core/modules/system/system.module and changing line 4222 from:

<code>    ->condition('name', '%' . db_like($string) . '%', 'LIKE')</code>
to:
<code>    ->condition('taxonomy_term_data.name', '%' . db_like($string) . '%', 'LIKE')</code>

After adding the table name, it prevents the ambiguous SQL 'name' column error and I get suggestions like you would expect for the path.

I know acl, forum, and forum_access appear to be pretty old and I plan long term to look at doing it the forum.backdropcms.org way. I like simple and using mostly core features when I can but I'm porting a Drupal 7 site and need to rely on the old forum module for now. Plus 'name' is a pretty generic column name so it would probably be good to qualify it in the where clause anyway. 

## Additional information

- Backdrop CMS version: 1.22.0
- ACL version: 1.x-1.0.0
- Forum version: 1.x-1.0.2
- Forum Access version: 1.x-1.4.2

indigoxela:
Fixes https://github.com/backdrop/backdrop-issues/issues/5684

### DIFF
--- a/core/modules/system/system.module
+++ b/core/modules/system/system.module
@@ -4146,12 +4146,12 @@ function path_autocomplete($string = '') {
 
   if ($string) {
     // Get node matches.
-    $results = db_select('node', 'n')
-      ->fields('n', array('nid', 'type', 'title', 'status'))
-      ->condition('n.title', '%' . db_like($string) . '%', 'LIKE')
-      ->condition('n.status', 1)
+    $results = db_select('node')
+      ->fields('node', array('nid', 'type', 'title', 'status'))
+      ->condition('node.title', '%' . db_like($string) . '%', 'LIKE')
+      ->condition('node.status', 1)
       ->addTag('node_access')
-      ->orderBy('n.nid', 'DESC')
+      ->orderBy('node.nid', 'DESC')
       ->range(0, $range)
       ->execute();
     foreach ($results as $result) {
@@ -4211,12 +4211,12 @@ function path_autocomplete($string = '') {
 
     // Get taxonomy term matches.
     if (module_exists('taxonomy') && $match_count < $range) {
-      $results = db_select('taxonomy_term_data', 'ttd')
-        ->fields('ttd', array('tid', 'vocabulary', 'name'))
-        ->condition('ttd.name', '%' . db_like($string) . '%', 'LIKE')
+      $results = db_select('taxonomy_term_data')
+        ->fields('taxonomy_term_data', array('tid', 'vocabulary', 'name'))
+        ->condition('taxonomy_term_data.name', '%' . db_like($string) . '%', 'LIKE')
         ->addTag('taxonomy_term_access')
         ->range(0, $range - $match_count)
-        ->orderBy('ttd.tid', 'DESC')
+        ->orderBy('taxonomy_term_data.tid', 'DESC')
         ->execute();
       foreach ($results as $result) {
         $path = backdrop_get_path_alias('taxonomy/term/' . $result->tid);
--- a/core/modules/system/system.module
+++ b/core/modules/system/system.module
@@ -4146,12 +4146,12 @@ function path_autocomplete($string = '') {
 
   if ($string) {
     // Get node matches.
-    $results = db_select('node')
-      ->fields('node', array('nid', 'type', 'title', 'status'))
-      ->condition('title', '%' . db_like($string) . '%', 'LIKE')
-      ->condition('status', 1)
+    $results = db_select('node', 'n')
+      ->fields('n', array('nid', 'type', 'title', 'status'))
+      ->condition('n.title', '%' . db_like($string) . '%', 'LIKE')
+      ->condition('n.status', 1)
       ->addTag('node_access')
-      ->orderBy('nid', 'DESC')
+      ->orderBy('n.nid', 'DESC')
       ->range(0, $range)
       ->execute();
     foreach ($results as $result) {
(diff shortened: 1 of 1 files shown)


klonos (approved):
LGTM 👍🏼

core/modules/system/system.module
```diff
@@ -4146,12 +4146,12 @@ function path_autocomplete($string = '') {
 
   if ($string) {
     // Get node matches.
-    $results = db_select('node')
-      ->fields('node', array('nid', 'type', 'title', 'status'))
-      ->condition('title', '%' . db_like($string) . '%', 'LIKE')
-      ->condition('status', 1)
+    $results = db_select('node', 'n')
```
argiepiano: Instead of creating a brand new alias for node, I propose using 'node', which was the default alias created before this change.

```suggestion
    $results = db_select('node')
```

core/modules/system/system.module
```diff
@@ -4146,12 +4146,12 @@ function path_autocomplete($string = '') {
...
+      ->fields('n', array('nid', 'type', 'title', 'status'))
```
argiepiano:
```suggestion
      ->fields('node', array('nid', 'type', 'title', 'status'))
```

core/modules/system/system.module
```diff
@@ -4146,12 +4146,12 @@ function path_autocomplete($string = '') {
...
+      ->condition('n.title', '%' . db_like($string) . '%', 'LIKE')
```
argiepiano:
```suggestion
      ->condition('node.title', '%' . db_like($string) . '%', 'LIKE')
```

core/modules/system/system.module
```diff
@@ -4146,12 +4146,12 @@ function path_autocomplete($string = '') {
...
+      ->condition('n.status', 1)
```
argiepiano:
```suggestion
      ->condition('node.status', 1)
```

core/modules/system/system.module
```diff
@@ -4146,12 +4146,12 @@ function path_autocomplete($string = '') {
...
       ->addTag('node_access')
-      ->orderBy('nid', 'DESC')
+      ->orderBy('n.nid', 'DESC')
```
argiepiano:
```suggestion
      ->orderBy('node.nid', 'DESC')
```

core/modules/system/system.module
```diff
@@ -4211,12 +4211,12 @@ function path_autocomplete($string = '') {
 
     // Get taxonomy term matches.
     if (module_exists('taxonomy') && $match_count < $range) {
-      $results = db_select('taxonomy_term_data')
-        ->fields('taxonomy_term_data', array('tid', 'vocabulary', 'name'))
-        ->condition('name', '%' . db_like($string) . '%', 'LIKE')
+      $results = db_select('taxonomy_term_data', 'ttd')
```
argiepiano: Instead of introducing a new alias, I suggest using the default provided by db_select
```suggestion
      $results = db_select('taxonomy_term_data')
```

core/modules/system/system.module
```diff
@@ -4211,12 +4211,12 @@ function path_autocomplete($string = '') {
...
+        ->fields('ttd', array('tid', 'vocabulary', 'name'))
```
argiepiano:
```suggestion
        ->fields('taxonomy_term_data', array('tid', 'vocabulary', 'name'))
```

core/modules/system/system.module
```diff
@@ -4211,12 +4211,12 @@ function path_autocomplete($string = '') {
...
+        ->condition('ttd.name', '%' . db_like($string) . '%', 'LIKE')
```
argiepiano:
```suggestion
        ->condition('taxonomy_term_data.name', '%' . db_like($string) . '%', 'LIKE')
```

core/modules/system/system.module
```diff
@@ -4211,12 +4211,12 @@ function path_autocomplete($string = '') {
...
         ->addTag('taxonomy_term_access')
         ->range(0, $range - $match_count)
-        ->orderBy('tid', 'DESC')
+        ->orderBy('ttd.tid', 'DESC')
```
argiepiano:
```suggestion
        ->orderBy('taxonomy_term_data.tid', 'DESC')
```

argiepiano (changes requested):

quicksketch:
Merged into 1.x and 1.22.x. Thanks @indigoxela!

PR #4138, opened 2022-07-19
Issue #5689: Fix type hint in doc block for function arg

### LINKED ISSUE #5689: [DX] Fixed: type hint in doc block for function arg().
## Description of the bug

Documentation task:

```
/**
 * Returns a component of the current Backdrop path.
 *
 * When viewing a page at the path "admin/structure/types", for example, arg(0)
 * returns "admin", arg(1) returns "structure", and arg(2) returns "types".
 *
 * Avoid use of this function where possible, as resulting code is hard to
 * read. In menu callback functions, attempt to use named arguments. See the
 * explanation in menu.inc for how to construct callbacks that take arguments.
 * When attempting to use this function to load an element from the current
 * path, e.g. loading the node on a node page, use menu_get_object() instead.
 *
 * @param $index
 *   The index of the component, where each component is separated by a '/'
 *   (forward-slash), and where the first component has an index of 0 (zero).
 * @param $path
 *   A path to break into components. Defaults to the path of the current page.
 *
 * @return string|NULL
 *   The component specified by $index, or NULL if the specified component was
 *   not found. If called without arguments, it returns an array containing all
 *   the components of the current path.
 */
function arg($index = NULL, $path = NULL) {
```

arg() may return an **array** when run without param.

```
 * @return string|NULL|array
```

indigoxela:
Fixes https://github.com/backdrop/backdrop-issues/issues/5689

### DIFF
--- a/core/includes/bootstrap.inc
+++ b/core/includes/bootstrap.inc
@@ -3788,7 +3788,7 @@ function request_path() {
  * @param $path
  *   A path to break into components. Defaults to the path of the current page.
  *
- * @return string|NULL
+ * @return string|NULL|array
  *   The component specified by $index, or NULL if the specified component was
  *   not found. If called without arguments, it returns an array containing all
  *   the components of the current path.


ghost:
FYI @indigoxela, for PRs like this that just changes comments/documentation and don't need tests, you can add `[skip tests]` in the PR title to skip them.

quicksketch:

PR #179, opened 2014-03-06
135-meta3-jen21

jenlampton:
Last PR for https://github.com/backdrop/backdrop-issues/issues/135

PR #4140, opened 2022-07-20
Security #179: Tighten image style derivative access checks.

quicksketch:
Fixes https://github.com/backdrop-ops/security/issues/179

### DIFF
--- a/core/modules/system/system.install
+++ b/core/modules/system/system.install
@@ -2477,6 +2477,7 @@ function system_update_1025() {
  */
 function system_update_1026() {
   $config = config('system.core');
+  $config->set('file_additional_public_schemes', update_variable_get('file_additional_public_schemes', array()));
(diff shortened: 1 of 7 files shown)


core/modules/system/system.install
```diff
@@ -3304,6 +3304,17 @@ function system_update_1082() {
   }
 }
 
+/**
+ * Set the default value for "file_additional_public_schemes" setting.
+ */
+function system_update_1083() {
+  $config = config('system.core');
+  if ($config->get('file_additional_public_schemes') == NULL) {
```
jenlampton, line 3313: Do we also need to do an `update_variable_get` (in the other system module conversion of system.core variables) so we don't loose this for sites upgrading from Drupal 7 tomorrow?

quicksketch: Good call. I'll add it to the earlier update that covers the other file system settings.

PR #4147, opened 2022-07-22
Issue 5696. Check installer.settings.json and fix if needed

### LINKED ISSUE #5696: Fixed potentially broken installer.settings.json configuration file.
## Description of the bug

This is a follow-up to #5690, where we discovered that under certain circumstances saving the update settings form can create a broken installer.settings.json file.

The immediate problem is fixed, but we need to take care of already broken files, as discovering what it needs to fix it isn't obvious.
We need an update hook.

argiepiano:
Fixes https://github.com/backdrop/backdrop-issues/issues/5696

### DIFF
--- a/core/modules/update/update.install
+++ b/core/modules/update/update.install
@@ -240,16 +240,27 @@ function update_update_1004() {
 }
 
 /**
- * Check installer.settings.json and fix it if broken.
+ * Check installer.settings.json and fix it or remove it if needed.
  */
 function update_update_1005() {
   $config = config('installer.settings');
-  if (!$config->isNew() && empty($config->get('installer_server'))) {
-    $config->set('installer_server', array(
-      'url' => 'https://projects.backdropcms.org',
-      'name' => 'Backdrop',
-    ));
-    $config->save();
+  if (!$config->isNew()) {
+    // Check to see if installer is installed.
+    // We must do this with a database query, as module_exists() doesn't tell us
+    // if the module is disabled or uninstalled.
+    $schema_version = db_query("SELECT schema_version FROM {system} WHERE name = 'installer' AND type = 'module'")->fetchObject();
+    if ($schema_version->schema_version < 0) {
+      // instaler is uninstalled. Remove the config file.
+      $config->delete();
+    } 
+    elseif (empty($config->get('installer_server'))) {
+      // Fix the config file.
+      $config->set('installer_server', array(
+        'url' => 'https://projects.backdropcms.org',
+        'name' => 'Backdrop',
+      ));
+      $config->save();
+    }
   }
 }
 
--- a/core/modules/update/update.install
+++ b/core/modules/update/update.install
@@ -250,9 +250,9 @@ function update_update_1005() {
     // if the module is disabled or uninstalled.
     $module_info = db_query("SELECT schema_version FROM {system} WHERE name = 'installer' AND type = 'module'")->fetchObject();
     if ($module_info->schema_version < 0) {
-      // instaler is uninstalled. Remove the config file.
+      // Installer is uninstalled. Remove the config file.
       $config->delete();
-    } 
+    }
     elseif (empty($config->get('installer_server'))) {
       // Fix the config file.
       $config->set('installer_server', array(
--- a/core/modules/update/update.install
+++ b/core/modules/update/update.install
@@ -240,9 +240,13 @@ function update_update_1004() {
 }
 
 /**
- * Check installer.settings.json and fix it or remove it if needed.
+ * Check for a faulty installer.settings.json and fix it or remove it if needed.
  */
 function update_update_1005() {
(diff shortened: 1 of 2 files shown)


core/modules/update/update.install
```diff
@@ -239,6 +239,31 @@ function update_update_1004() {
...
+ */
+function update_update_1005() {
+  $config = config('installer.settings');
+  if (!$config->isNew()) {
+    // Check to see if installer is installed.
+    // We must do this with a database query, as module_exists() doesn't tell us
+    // if the module is disabled or uninstalled.
+    $module_info = db_query("SELECT schema_version FROM {system} WHERE name = 'installer' AND type = 'module'")->fetchObject();
+    if ($module_info->schema_version < 0) {
+      // instaler is uninstalled. Remove the config file.
```
indigoxela: Oops, tiny typo:

instaler -> Installer :wink:

argiepiano: Fixed.

core/modules/update/update.install
```diff
@@ -239,6 +239,31 @@ function update_update_1004() {
   $config->save();
 }
 
+/**
+ * Check installer.settings.json and fix it or remove it if needed.
```
docwilmot: It took me a while reading the back issues to figure why we needed this function (about Installer, but in Update??), so I suspect others may benefit from a little bit more explanation here without needing to git blame in the future.

core/modules/update/update.install
```diff
@@ -239,6 +239,31 @@ function update_update_1004() {
...
+ */
+function update_update_1005() {
+  $config = config('installer.settings');
+  if (!$config->isNew()) {
```
docwilmot, line 254: This is TRUE if the settings file doesnt exist, right? Does this block need to run if the file doesnt exist?

core/modules/update/update.install
argiepiano: Thanks, @docwilmot. I had originally added this to `installer.intall`, but @indigoxela rightly pointed out that the faulty installer.settings.json is created even if `installer` is not enabled, meaning that the deletion of the faulty file must happen outside `installer.install` (otherwise, the faulty config file will not get fixed or removed if `installer` is uninstalled). 

The culprit for creating this faulty and unnecessary file is an ealier update hook inside `update.install`. Therefore we both thought that creating this update hook within `update.install` was the right way to process. 

I can add some explanations for this in the update hook. However I did not want to confuse or overwhelm the administrator running update with information in the doc block (which is shown when you run update.php). Do you suggest adding as inline comments within the function?

core/modules/update/update.install
argiepiano: The idea here is that the fixing or removal of the config file must happen only if the file actually exists, and if it's faulty (e.g. if it's there even when `installer` is disabled, or if the key `installer_server` doesn't exist in the file).

So, this `if` statement checks the config file. If the file is new, that means that the file doesn't exist. If this is the case, we don't need to do any fixing or removing. 

Side note: a different PR, already committed and released as part of 1.22.2, already fixed the code that created the faulty and unnecessary `installer.settings.json` file in the first place.

core/modules/update/update.install
docwilmot: I agree with everything you folks did actually, it just took me a while to get it, so wanted to ensure we help the next person to come across this function. And yes, I think some inline would be great.

core/modules/update/update.install
docwilmot: > If the file is new, that means that the file doesn't exist

That sounds so totally backwards! 😄 Sounds like we should have a `Config::file_exists()` or something. 

In either case, maybe a brief code comment for the rest of us please?

argiepiano: > That sounds so totally backwards!

Yup 😄 My wording didn't help. Rephrase: "if the config object returns TRUE for `isNew()` that means that it has not been saved, and the file doesn't exist." Makes more sense?  

I'll add better documentation later.

argiepiano: @docwilmot, I added more explanations. Please check to see if they make sense now.

core/modules/update/update.install
```diff
@@ -239,6 +239,35 @@ function update_update_1004() {
   $config->save();
 }
 
+/**
+ * Check installer.settings.json and fix it or remove it if needed.
+ */
+function update_update_1005() {
+  $config = config('installer.settings');
```
docwilmot, line 250: What I was actually suggesting earlier was a comment about why this update hook exists, to explain why Installer code is in Update. Something like "Saving the update settings form created an empty copy of installer settings in earlier Backdrop versions. This update hook fixes this." Sorry to nitpick and not explain the nitpick  properly.

argiepiano: OK, got it. Please check the latest version and let me know if this is what you had in mind.

quicksketch:

PR #4148, opened 2022-07-25
Issue #5700: Fix docblocks of hook_boot() and hook_init() [skip tests]

### LINKED ISSUE #5700: Fixed: Docblock should reference '$settings['page_cache_invoke_hooks']' (not '$settings['invoke_page_cache_hooks']').
## Description of the bug

The [documentation of `hook_boot()`](https://docs.backdropcms.org/api/backdrop/core%21modules%21system%21system.api.php/function/hook_boot/1) says:

> This hook by default is not called on pages served by the default page cache, but can be enabled through the `$settings['invoke_page_cache_hooks']` option in [settings.php](https://docs.backdropcms.org/api/backdrop/settings.php/1).

The correct name is `'page_cache_invoke_hooks'`, which is readily verified by doing a search on both names in Backdrop core. The only occurrence of `'invoke_page_cache_hooks'` is in the docblock of `hook_boot()`, also `hook_exit()`.

So the docblocks are the only thing that needs to be corrected. No change needed in functional code.

bugfolder:
Fixes https://github.com/backdrop/backdrop-issues/issues/5700.

Documentation-only change; no change to code.

### DIFF
--- a/core/modules/system/system.api.php
+++ b/core/modules/system/system.api.php
@@ -269,12 +269,13 @@ function hook_element_info_alter(&$type) {
  * the browser.
  *
  * This hook by default is not called on pages served by the default page cache,
- * but can be enabled through the $settings['invoke_page_cache_hooks'] option in
+ * but can be enabled through the $settings['page_cache_invoke_hook'] option in
  * settings.php.
  *
  * @param $destination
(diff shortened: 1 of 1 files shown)


quicksketch:
Merged into 1.x and 1.22.x. Thanks @bugfolder and @indigoxela!

PR #4149, opened 2022-07-25
Issue #5698: Ensure valid text format in file_admin view.

### LINKED ISSUE #5698: After D7 upgrade, no results in File Admin puts "Missing text format: filtered_html" in watchdog
## Description of the bug

After upgrading from D7, visiting the "Manage files" page at `admin/content/files` and entering a search term that gave no results puts a "Missing text format: filtered_html" error into watchdog.

Although recent D7 sites give the "Filtered HTML" text format the machine name `filtered_html`, older D7 sites used numeric machine names (for example, `1`).

## Steps To Reproduce

To reproduce the behavior:

1. Create a D7 website. Make sure it is old enough that its fallback text format (typically "Filtered HTML") has a numeric machine name (e.g., `1`), rather than `filtered_html` (if it has the latter, this issue won't arise). Or create a new text format to use as fallback format with a different machine name, and delete the `filtered_html` format.
2. Upgrade to Backdrop.
3. Visit `admin/content/files` and enter a search term that gives no results.
4. Visit watchdog at `admin/reports/dblog`. You will see the error.

## Additional information

Although recent D7 sites give the "Filtered HTML" text format the machine name `filtered_html`, older D7 sites used numeric machine names (for example, `1`). (Or site administrators might have created and designated a different fallback format.)

The View that provides this page (`views.view.file_admin.json`) is installed at upgrade time in `file_update_1000()`, and its "No results" value has the format set to `filtered_html`. If the Filtered HTML format had a machine name of `1`, though, it will result in this error.

A manual fix is easy enough to implement: edit the View and set the "No Results" format to one of the existing formats (one with proper sanitization, of course).

As a fix, the update should use `fiter_fallback_format()` to get the proper machine name and then use that.

bugfolder:
Fixes https://github.com/backdrop/backdrop-issues/issues/5698.


core/modules/file/file.install
```diff
@@ -1199,3 +1211,24 @@ function file_update_1008() {
...
+ * Ensures that the "no results" text in 'file_admin' view has a valid format.
+ */
+function file_update_1009() {
+  // Back in file_update_1000() we created this view containing 'filtered_html'
+  // as the "no results" text format, but if we were upgrading from a site that
+  // didn't contain that format, it results in an error. Here we ensure that the
+  // format for that text is a valid format and if it isn't, use the fallback
+  // format.
+  $view = views_get_view('file_admin');
+  if ($view) {
```
jenlampton: We need to check first if this view has been modified by site admins before we change it programatically. Can you pull its storage state first, wrap this in an "if default" check?  I believe there are some examples for other view updates (home page promoted view, for example) where you can see how this is done.

bugfolder: <strike>
Sorry, maybe I'm missing something. This update only changes the view if it has an invalid text format for the "No results" text (whether or not it has been modified by the site admin). So, IIUC, you're saying that if the View was modified by a site admin and has an invalid format, we should leave the invalid format alone? That we'd only fix the invalid format if the View hasn't been modified?

That seems like a less-desirable behavior. If the text format is invalid for any reason, surely using the fallback format is an improvement over the invalid format.

And it seems like a pretty common use case might be that the site admin has made a change elsewhere in the View without noticing that the "No results" format was invalid, and it would still be a good thing to correct the invalid format, even if they've modified the View.

Is there any use case where leaving the text format invalid is preferable to it being the fallback format?
</strike>

EDIT: Ah, I think I understand. They might not be using the `default` display.

bugfolder: Updated to check all displays of that view and fix if necessary.

jenlampton: You were right the first time. 

> So, IIUC, you're saying that if the View was modified by a site admin and has an invalid format, we should leave the invalid format alone? That we'd only fix the invalid format if the View hasn't been modified?

Yes, that's what I meant. This is the process we've used for all other config updates. If the [config thing] has been modified in any way from its default state, we should not update it in an update hook. We do this because it's possible that the [config thing] has been modified in some way that would make our change cause other problems, and since there are an infinite number of changes that we can't realistically test for, it's not worth worrying about the edgecases where the view has been modified. Instead, we catch the majority of cases where we know the change is safe to make (where the view is unchanged) and disregard any modified config.

bugfolder: This point was discussed in the issue (and I think is resolved), so I'm taking the liberty of marking the PR conversation resolved.

quicksketch:
Merged into 1.x and 1.26.x. Great work @bugfolder!

PR #4152, opened 2022-07-31
Issue #5701: Change message to "Your comment has been updated" on comment update

### LINKED ISSUE #5701: [UX] Show different messages for new comment posted vs. existing comment updated
## Description of the bug

We needed separate `created` and `updated` notifications for comments on a new Backdrop project. Unfortunately, the current state is that it always gives the "_Your comment has been posted._" message regardless if a comment was just created or updated. Nodes, on the other hand, have a separate message for update operation: "_Post Your first post! has been updated._". So I believe this needs to be brought into consistency.

## Additional information

This can be easily addressed by changing the following code in `comment.module`:

```
    if ($comment->status == COMMENT_NOT_PUBLISHED) {
      if (!user_access('administer comments')) {
        backdrop_set_message(t('Your comment has been queued for review by site administrators and will be published after approval.'));
      }
    }
    else {
      backdrop_set_message(t('Your comment has been posted.'));
    }
```

to 

```
    if ($comment->status == COMMENT_NOT_PUBLISHED) {
      if (!user_access('administer comments')) {
        backdrop_set_message(t('Your comment has been queued for review by site administrators and will be published after approval.'));
      }
    }
    elseif (isset($comment->new)) {
      backdrop_set_message(t('Your comment has been updated.'));
    }
    else {
      backdrop_set_message(t('Your comment has been posted.'));
    }
```

and I'll be filing a PR soon. However, what I find it strange is that the `$comment->new` is set and equals to FALSE only on updates whereas on new comment creation it's not there at all. So I am not sure if this is a separate bug, because semantically it makes more sense to use an if condition something like this:

```
if ($comment->new == TRUE) {
  print created message here
} 
else {
  print updated message here
}
```

Let me know if we should chase why the `$comment->new ` is not there for new comments or if the suggested PR is good enough.

alanmels:
Fixes https://github.com/backdrop/backdrop-issues/issues/5701


core/modules/comment/comment.module
```diff
@@ -2150,6 +2149,9 @@ function comment_form_submit($form, &$form_state) {
         backdrop_set_message(t('Your comment has been queued for review by site administrators and will be published after approval.'));
       }
     }
+    elseif (isset($comment->new)) {
```
argiepiano: Suggest using the method `isNew()` instead. The property `new` is used for a different purpose in comments. See my comment in the main issue.

```suggestion
    elseif (!$comment->isNew()) {
```

core/modules/comment/comment.module
```diff
@@ -2137,7 +2137,6 @@ function comment_form_submit($form, &$form_state) {
     if (user_is_anonymous()) {
       user_cookie_save(array_intersect_key($form_state['values'], array_flip(array('name', 'mail', 'homepage'))));
     }
-
     comment_save($comment);
```
argiepiano, line 2164: Since method isNew() after saving will return false, let's use a variable to store the return value here.

 ```suggestion
    $is_new = $comment->isNew();
    comment_save($comment);
```

File: core/modules/comment/comment.module [hunk 3]
```diff
@@ -2150,6 +2149,9 @@ function comment_form_submit($form, &$form_state) {
         backdrop_set_message(t('Your comment has been queued for review by site administrators and will be published after approval.'));
       }
     }
+    elseif (!$comment->isNew()) {
```
argiepiano: Then, here, check that variable.
```suggestion
    elseif (!$is_new) {
```

PR #4154, opened 2022-08-02
Fix #5704

### LINKED ISSUE #5704: [UX] Improve field block labels when copying the Layout title from a block.
## Description of the bug

When using Layouts to format node pages the Configure page title doesn't show meaningful options under "Block to copy" when "Copy from block" is selected.

## Steps To Reproduce

To reproduce the behavior:
1. (optional) Create a content type with multiple fields, or add various fields other than the default ones. This will make the issue more obvious.
2. Create a layout using Path: `node/%` and save.
3. Add two or more fields somewhere in the layout.
4. If the title combo block is available in the layout, remove it -> this action will the "Page title" element.
5. Click on Configure in the "Page title" element.
6. Select "Copy from block"
7. Click the "Block to copy" select that appears, to see the available options.

## Actual behavior

Multiple items are displayed as "Field block" with no way to know which field you are selecting.
image

## Expected behavior

Fields should be displayed with the field name. For example:  "Field block - field_1st_field", "Field bock - field_2nd_field"...

## Additional information

Backdrop version 1.22.2

kswan:
Fixes [#5704](https://github.com/backdrop/backdrop-issues/issues/5704)

### DIFF
--- a/core/modules/layout/layout.admin.inc
+++ b/core/modules/layout/layout.admin.inc
@@ -1468,6 +1468,7 @@ function layout_title_settings_form($form, &$form_state, Layout $layout) {
   $block_options = array();
   foreach ($layout->content as $uuid => $block) {
     $title = $all_blocks[$block->module][$block->delta]['info'];
+    if (isset($block->childDelta)) $title .= " - " . $block->childDelta;
     if (!empty($title) && $title !== 'Custom block' && !is_a($block, 'PageComponents')) {
       $block_options[$uuid] = $title;
     }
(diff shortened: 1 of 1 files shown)


core/modules/layout/layout.admin.inc
```diff
@@ -1468,6 +1468,7 @@ function layout_title_settings_form($form, &$form_state, Layout $layout) {
   $block_options = array();
   foreach ($layout->content as $uuid => $block) {
     $title = $all_blocks[$block->module][$block->delta]['info'];
+    if (isset($block->childDelta)) $title .= " - " . $block->childDelta;
```
indigoxela: Although this one-liner is correct, I'd suggest to use curly braces, newlines and proper indent. Just for better code readability.

PR #4155, opened 2022-08-03
 Issue #5704: Improve the list of block labels when copying Layout title from a block.

### LINKED ISSUE #5704: [UX] Improve field block labels when copying the Layout title from a block.

kswan:
Fixes [#5704](https://github.com/backdrop/backdrop-issues/issues/5704)

### DIFF
--- a/core/modules/layout/layout.admin.inc
+++ b/core/modules/layout/layout.admin.inc
@@ -1464,11 +1464,10 @@ function layout_title_settings_form($form, &$form_state, Layout $layout) {
     '#maxlength' => 255,
   );
 
-  $all_blocks = layout_get_block_info();
   $block_options = array();
   foreach ($layout->content as $uuid => $block) {
-    $title = $all_blocks[$block->module][$block->delta]['info'];
-    if (!empty($title) && $title !== 'Custom block' && !is_a($block, 'PageComponents')) {
+    $title = strip_tags($block->getAdminTitle());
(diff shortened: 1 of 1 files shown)


indigoxela:
@kswan mind to give the PR a more meaningful title? Preferably something like `Issue #5704: Better usability for layout page title configure form when using field values` (or improve as you think).

PR titles are base for the merge commit message - so ideally they describe what the PR does.

quicksketch:
Thanks @kswan! I merged into 1.x and 1.23.x. (I also updated the PR title and commit message, thanks @indigoxela!)

PR #4163, opened 2022-08-10
Issue #5554: Mark some Comment permissions as restricted

### LINKED ISSUE #5554: Mark some Comment permissions as restricted
In working on https://github.com/backdrop/backdrop-issues/issues/5536, I noticed that while permissions like `Administer content` are restricted (they have the security warning), equivalent Comment permissions aren't.

IMO, the following Comment permissions should be restricted:
- `Administer comments`
- `Administer comment settings` - allows approving and deleting comments
- `Skip comment approval` - allows bypassing approval process
- `Edit own comments` - allows getting approval for a comment, but then changing it to say something else afterwards (also bypassing approval)

Thoughts?

avpaderno:
Fixes https://github.com/backdrop/backdrop-issues/issues/5554


core/modules/comment/comment.module
```diff
@@ -415,9 +419,13 @@ function comment_permission() {
     ),
     'skip comment approval' => array(
       'title' => t('Skip comment approval'),
+      'restrict access' => TRUE,
+      'warning' => t('Users could post spam comments that are directly visibile to every user.),
```
argiepiano:
```suggestion
      'warning' => t('Users could post spam or offensive comments that become immediately visible to every user.),
```

core/modules/comment/comment.module
```diff
@@ -415,9 +419,13 @@ function comment_permission() {
...
     ),
     'edit own comments' => array(
       'title' => t('Edit own comments'),
+      'restrict access' => TRUE,
+      'warning' => t('Users could change their approved comments, without further approvals.'),
```
argiepiano:
```suggestion
      'warning' => t('Users could edit their approved comments, without further approvals.'),
```

core/modules/comment/comment.module
```diff
@@ -402,10 +402,14 @@ function comment_permission() {
...
       'title' => t('Administer comment settings'),
       'description' => t('Configure comment settings per content type.'),
+      'restrict access' => TRUE,
+      'warning' => t('Change the comment settings for each content type, including disabling comments.'),
     ),
     'administer comments' => array(
       'title' => t('Administer comments'),
       'description' => t('Edit, delete, unpublish, etc. comments at <em>/admin/content/comment</em>.'),
+      'restrict access' => TRUE,
+      'warning' => t('Allow a user to delete all the comments.'),
```
argiepiano:
```suggestion
      'warning' => t('Allow a user to delete all comments.'),
```

core/modules/comment/comment.module
```diff
@@ -402,10 +402,14 @@ function comment_permission() {
     'administer comment settings' => array(
       'title' => t('Administer comment settings'),
       'description' => t('Configure comment settings per content type.'),
+      'restrict access' => TRUE,
+      'warning' => t('Allows to change comment settings for each content type and disable comments.'),
```
klonos:
```suggestion
      'warning' => t('Allows changing comment settings for each content type, and disabling comments.'),
```

core/modules/comment/comment.module
```diff
@@ -402,10 +402,14 @@ function comment_permission() {
...
     ),
     'administer comments' => array(
       'title' => t('Administer comments'),
       'description' => t('Edit, delete, unpublish, etc. comments at <em>/admin/content/comment</em>.'),
+      'restrict access' => TRUE,
+      'warning' => t('Allows to delete all comments.'),
```
klonos:
```suggestion
      'warning' => t('Allows deleting/updating any comment, even those posted by other people.'),
```

core/modules/comment/comment.module
```diff
@@ -415,9 +419,13 @@ function comment_permission() {
     ),
     'skip comment approval' => array(
       'title' => t('Skip comment approval'),
+      'restrict access' => TRUE,
+      'warning' => t('Allows to post spam or offensive comments that become immediately visible to every visitor.'),
```
klonos:
```suggestion
      'warning' => t('Allows posting spam or offensive comments that become immediately visible to every visitor.'),
```

core/modules/comment/comment.module
```diff
@@ -415,9 +419,13 @@ function comment_permission() {
...
     ),
     'edit own comments' => array(
       'title' => t('Edit own comments'),
+      'restrict access' => TRUE,
+      'warning' => t('Allows to edit already approved comments, without the need of further approvals.'),
```
klonos:
```suggestion
      'warning' => t('Allows editing previously approved comments, bypassing the need for further approvals.'),
```

core/modules/comment/comment.module
jenlampton: This change looks great.

core/modules/comment/comment.module
jenlampton: Love this suggestion.

core/modules/comment/comment.module
```diff
@@ -415,9 +419,13 @@ function comment_permission() {
     ),
     'skip comment approval' => array(
       'title' => t('Skip comment approval'),
+      'restrict access' => TRUE,
```
jenlampton: Probably shouldn't restrict access for this one.

core/modules/comment/comment.module
jenlampton: This description isn't accurate. 

Nobody would ever intentionally choose to authorize "spam or offensive comments", so nobody would ever check this box if we described it this way. We do often choose to authorize trusted people to post comments without approval, however, so we should do a better job of describing  what this permission actually does, and not only state the risks.

How about something more descriptive like:
`Allows comments to be posted without approval, becoming immediately visible to everyone. Because spam and offensive comments are common, this permission should only be granted to those who are trusted, or on a site with additional spam protection measures in place."

core/modules/comment/comment.module
jenlampton: I think this description looks good.

core/modules/comment/comment.module
```diff
@@ -415,9 +419,13 @@ function comment_permission() {
...
+      'warning' => t('Allows to post spam or offensive comments that become immediately visible to every visitor.'),
     ),
     'edit own comments' => array(
       'title' => t('Edit own comments'),
+      'restrict access' => TRUE,
```
jenlampton: Probably no need to restrict access.

core/modules/comment/comment.module
avpaderno: I agree: The warning makes the permission sound like a permission to spam or post offensive comments, while the permission is just to post comments that don't need to be first approved. I would just use *Since spam and offensive comments are common, this permission should only be granted to those who are trusted, or on a site with additional spam protection measures in place.* or a shorter sentence. The warning is an addendum to the description, which should already say what the permission does.

quicksketch:
Merged into 1.x and 1.26.x. Thanks @kiamlaluno for sticking this one out!

PR #4164, opened 2022-08-12
Issue #5695: Prevent misleading watchdog when no join is needed

### LINKED ISSUE #5695: Missing Views handler: watchdog reports "Missing join" with relationship+contextual filter+taxonomy term
## Description of the bug

Creating a View with a Relationship to a taxonomy term and a Contextual filter on the term name results in a watchdog entry of type=`views`, message=`Missing join: taxonomy_term_data taxonomy_term_data` whenever the View is rendered with the contextual filter term.

## Steps To Reproduce

To reproduce the behavior:

1. Begin with a standard Backdrop installation (with Card, Page, and Post content types).
2. The `Tags` vocabulary already exists. Create two terms, `dog` and `cat`.
3. Create four Post nodes, two with `dog` tag, two with `cat` tag. (For example, "Retriever", "Spaniel", "Tabby", and "Siamese".
4. Create a new View of content of type Post, display teasers, with page path `animals`.
5. Add a relationship, "Content: Tags (field_tags)".
6. Add a contextual filter of type "Taxonomy term:name", using the Relationship "term from field_tags". Leave other settings at their default values.
7. Turn on "Auto preview" in the View. Initially, you will see all posts, the four you created plus "Your first post!".
8. Open `admin/reports/dblog` in a separate window. There's no notice yet.
9. To generate the notice, enter the contextual filter `dog` and click "Update Preview". Then refresh the watchdog page, where you will now see the notice.

Or, equivalently, save the View and then visit the page `animals/dog`. This, too, will put the notice into watchdog.

## Actual behavior

Every time you refresh the display with a contextual filter parameter, it puts another notice in watchdog. If you don't include the contextual filter parameter, there's no notice.

If you create a View with multiple relationships and corresponding contextual filters, there will be multiple watchdog notices with each page render, one for each relationship/filter combination.

Other than the watchdog notices, the View appears to function properly. In this example, using `dog` will show only the two dog posts; `cat` will show only the two cat posts; no parameter shows all posts.

A simple workaround is to just comment out the notice in function `views_get_table_join()` in file `core/modules/views/includes/utility.inc`:

```
  // DEBUG -- identify missing handlers
  watchdog('views', "Missing join: @table @base_table", array('@table' => $table, '@base_table' => $base_table));
```

But, presumably, it's there for a reason, and there's some deeper issue that needs to be addressed.

indigoxela:
Fixes https://github.com/backdrop/backdrop-issues/issues/5695

### DIFF
--- a/core/modules/views/includes/utility.inc
+++ b/core/modules/views/includes/utility.inc
@@ -123,8 +123,11 @@ function views_get_table_join($table, $base_table) {
     return $handler;
   }
 
-  // DEBUG -- identify missing handlers
-  watchdog('views', "Missing join: @table @base_table", array('@table' => $table, '@base_table' => $base_table));
+  // DEBUG: Identify missing handlers unless both tables are the same, because
+  // no join is needed in that case.
+  if ($table != $base_table) {
+    watchdog('views', "Missing join: @table @base_table", array('@table' => $table, '@base_table' => $base_table));
+  }
 }
 
 /**


quicksketch:

PR #4165, opened 2022-08-12
Issue #5686: Remove autocomplete from system path form input

### LINKED ISSUE #5686: Function path_admin_form should not have a path autocomplete for the source field
## Description of the bug

The form on "Add URL alias" (/admin/config/urls/path/add) has two fields, "Existing system path" and "URL alias".

The input for "Existing system path" has autocomplete functionality enabled, which is weird, as autocomplete will be an alias in most cases.

This leads to a "WTF" moment.

## Steps To Reproduce

1. Go to /admin/config/urls/path/add
2. Click into the first field and start typing and select something
3. Click into the next field and ... scratch your head

## Actual behavior

alias-an-alias

Sure, on submission the system path value gets submitted, so the alias ... for the (seemingly) alias will get created properly, but I find that very confusing.

## Expected behavior

Neither in D7 nor in D9 this field is path_autocomplete. And I think that makes much more sense. We need system paths here, but autocomplete *displays* aliases.

## Additional information

Here the autocomplete gets attached: https://github.com/backdrop/backdrop/blob/1.x/core/modules/path/path.admin.inc#L137

Besides the odd UX, this also leads to other problems, like #5684

indigoxela:
Fixes https://github.com/backdrop/backdrop-issues/issues/5686

### DIFF
--- a/core/modules/path/path.admin.inc
+++ b/core/modules/path/path.admin.inc
@@ -134,7 +134,6 @@ function path_admin_form($form, &$form_state, $path = array('source' => '', 'ali
     '#description' => t('Specify the existing path you wish to alias. For example: <code>node/28</code> or <code>taxonomy/term/1</code>.'),
     '#field_prefix' => $field_prefix,
     '#required' => TRUE,
-    '#autocomplete_path' => 'path-autocomplete',
   );
   $form['alias'] = array(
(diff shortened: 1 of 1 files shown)


quicksketch:

PR #4166, opened 2022-08-12
Issue #5720: Remove the "Ensure the current toolkit supports the effe…

### LINKED ISSUE #5720: Remove the "Ensure the current toolkit supports the effect." comment from image_effect_definitions()
[`image_effect_definitions()`](https://github.com/backdrop/backdrop/blob/1.x/core/modules/image/image.module#L1182) contains the following code.

```php
// Ensure the current toolkit supports the effect.
$effect['module'] = $module;
$effect['name'] = $name;
$effect['data'] = isset($effect['data']) ? $effect['data'] : array();
$effects[$name] = $effect;
```
The comment doesn't describe what the code is doing and it should be removed. 

Eventually, the comment should be *Make sure module, name, and data are set.* but that is probably not necessary, as that's clear from the code, except in the case we want to remember why those lines have been added.

avpaderno:
Fixes https://github.com/backdrop/backdrop-issues/issues/5720

### DIFF
--- a/core/modules/image/image.module
+++ b/core/modules/image/image.module
@@ -1179,7 +1179,6 @@ function image_effect_definitions() {
       include_once BACKDROP_ROOT . '/core/modules/image/image.effects.inc';
       foreach (module_implements('image_effect_info') as $module) {
         foreach (module_invoke($module, 'image_effect_info') as $name => $effect) {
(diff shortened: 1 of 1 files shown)


ghost:
Thanks! Merged into 1.x and 1.22.x.

PR #4170, opened 2022-08-15
Issue #5724: Change the word used for nodes

### LINKED ISSUE #5724: Node deletion message: "Deleted @count posts" even if it's not a Post
## Description of the bug

This is not a big deal, but I just noticed that when when one deletes one or more nodes, both the `backdrop_set_message()` and the watchdog entry say "Deleted &lt;count&gt; posts", which could be worrisome to someone who has Post, Page, and Card content types and deleted some of the latter two; the message suggests that it was the Post type that was deleted since the subtle distinction between "post" and "Post" might be lost on people.

This all takes place in `node_multiple_delete_confirm_submit()`, where we have access to the full list of nodes prior to deletion, so it would be straightforward to make the deletion message clearer. Perhaps something like:

"Deleted &lt;count&gt; Pages, Cards, and/or &lt;other type&gt;s", i.e., listing the total and what types were deleted, or

"Deleted &lt;count&gt; Pages, &lt;count&gt; Cards, and ...", i.e., listing how many of each type were deleted.

avpaderno:
Fixes https://github.com/backdrop/backdrop-issues/issues/5724

### DIFF
--- a/core/modules/node/node.admin.inc
+++ b/core/modules/node/node.admin.inc
@@ -199,7 +199,7 @@ function node_multiple_delete_confirm($form, &$form_state) {
     $form['node_list']['#items'][] = check_plain($title);
   }
 
-  $confirm_question = format_plural(count($rows), 'Are you sure you want to delete this item?', 'Are you sure you want to delete these items?');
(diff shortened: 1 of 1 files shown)


laryn:
Thanks  @kiamlaluno, @bugfolder, @olafgrabienski, @klonos and @stpaultim!

PR #4171, opened 2022-08-15
Issue #5726: Check what node_type_get_type() returns

### LINKED ISSUE #5726: node_type_delete() doesn't check the value returned from node_type_get_type()
## Description of the bug

[`node_type_delete()`](https://docs.backdropcms.org/api/backdrop/core%21modules%21node%21node.module/function/node_type_delete/1) doesn't check the value returned by [`node_type_get_type()`](https://docs.backdropcms.org/api/backdrop/core%21modules%21node%21node.module/function/node_type_get_type/1), which can either return an object with the required information, or `FALSE`. This also means that:

- `node_type_delete()` is passing `FALSE` to [`hook_node_type_delete()`](https://docs.backdropcms.org/api/backdrop/core!modules!node!node.api.php/function/hook_node_type_delete/1) which only expects a `stdClass` instance; this would either cause further errors, if those hook implementations don't check they are getting `FALSE` instead of the documented value they should get (in the worse case), or invoke the hook implementations which don't perform any action (in the best case)
- `node_type_delete()` tries to delete a configuration object that has been already deleted
- `node_type_delete()` calls [`field_attach_delete_bundle()`](https://docs.backdropcms.org/api/backdrop/core%21modules%21field%21field.attach.inc/function/field_attach_delete_bundle/1) for a content type that has been already deleted
- `node_type_delete()` calls [`node_type_cache_reset()`](https://docs.backdropcms.org/api/backdrop/core%21modules%21node%21node.module/function/node_type_cache_reset/1) which would again clear the cache and reset the *_node_types_build* static value
The function should first check what `node_type_get_type()` returns and return immediately when the returned value is `FALSE`.

avpaderno:
Fixes https://github.com/backdrop/backdrop-issues/issues/5726

### DIFF
--- a/core/modules/node/node.module
+++ b/core/modules/node/node.module
@@ -636,14 +636,15 @@ function node_field_extra_fields() {
  *   The machine-readable name of the node type to be deleted.
  */
 function node_type_delete($type) {
-  $info = node_type_get_type($type);
-  $config = config('node.type.' . $type);
-  $config->delete();
-  field_attach_delete_bundle('node', $type);
-  module_invoke_all('node_type_delete', $info);
+  if (($info = node_type_get_type($type)) !== FALSE) {
+    $config = config('node.type.' . $type);
+    $config->delete();
+    field_attach_delete_bundle('node', $type);
+    module_invoke_all('node_type_delete', $info);
 
-  // Clear the node type cache.
-  node_type_cache_reset();
+    // Clear the node type cache.
+    node_type_cache_reset();
+  }
 }
 
 /**
--- a/core/modules/node/node.module
+++ b/core/modules/node/node.module
@@ -636,7 +636,7 @@ function node_field_extra_fields() {
  *   The machine-readable name of the node type to be deleted.
  */
 function node_type_delete($type) {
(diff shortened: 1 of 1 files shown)


core/modules/node/node.module
```diff
@@ -636,14 +636,15 @@ function node_field_extra_fields() {
  *   The machine-readable name of the node type to be deleted.
  */
 function node_type_delete($type) {
-  $info = node_type_get_type($type);
-  $config = config('node.type.' . $type);
-  $config->delete();
-  field_attach_delete_bundle('node', $type);
-  module_invoke_all('node_type_delete', $info);
+  if (($info = node_type_get_type($type)) !== FALSE) {
```
argiepiano: Isn't it simpler to just do

```suggestion
  if (($info = node_type_get_type($type))) {
```

quicksketch: One more set of parenthesis can be removed:
```suggestion
  if ($info = node_type_get_type($type)) {
```

quicksketch:
Merged into 1.x and 1.23.x. Thanks @kiamlaluno!

PR #4172, opened 2022-08-15
Issue #5727: Passing null to parameter #1 of strlen() is deprecated

### LINKED ISSUE #5727: PHP 8.1: strlen(): Passing null to parameter #1 ($string) of type string is deprecated in backdrop_random_bytes()
## Description of the bug

[`backdrop_random_bytes()`](https://docs.backdropcms.org/api/backdrop/core!includes!bootstrap.inc/function/backdrop_random_bytes/1) uses the following code.

```php
    if (strlen($bytes) < $count) {
      // Initialize on the first call. The contents of $_SERVER includes a mix of
      // user-specific and system information that varies a little with each page.
      if (!isset($random_state)) {
        $random_state = print_r($_SERVER, TRUE);
        if (function_exists('getmypid')) {
          // Further initialize with the somewhat random PHP process ID.
          $random_state .= getmypid();
        }
        $bytes = '';
      }

      do {
        $random_state = hash('sha256', microtime() . mt_rand() . $random_state);
        $bytes .= hash('sha256', mt_rand() . $random_state, TRUE);
      } 
       while (strlen($bytes) < $count);
    }
```

It passes the static `$bytes` variable, which could still be uninitialized, to `strlen()`.

In PHP 8.1, passing `NULL` to `strlen()` throws an exception.

## Expected behavior
As done in one of the first function lines, the code should cast `$bytes` to `string` in the `strlen()` calls.

avpaderno:
Fixes https://github.com/backdrop/backdrop-issues/issues/5727

### DIFF
--- a/core/includes/bootstrap.inc
+++ b/core/includes/bootstrap.inc
@@ -2596,7 +2596,7 @@ function backdrop_random_bytes($count)  {
     // the microtime() - is prepended rather than appended. This is to avoid
     // directly leaking $random_state via the $output stream, which could
     // allow for trivial prediction of further "random" numbers.
-    if (strlen($bytes) < $count) {
+    if (strlen((string) $bytes) < $count) {
       // Initialize on the first call. The contents of $_SERVER includes a mix of
       // user-specific and system information that varies a little with each page.
       if (!isset($random_state)) {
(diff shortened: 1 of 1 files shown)


klonos (approved):
LGTM 👍🏼

laryn:
Thank you @kiamlaluno @indigoxela and @klonos -- merged into 1.x and 1.22.x!

PR #4173, opened 2022-08-15
Issue #5728: Handle unescaped less/greater than symbols in search indexing.

### LINKED ISSUE #5728: PHP notice when running search cron: Undefined
## Description of the bug

The `search_index()` function may throw a PHP notice if a `<` (less than) symbol is used in the body of indexed content. Search module tries to identify HTML tags such as `<h1>` to give them different rankings, but when searching for tags it can mistakenly identify an opening bracket that is not actually a tag, such as:

```
0 < 1
```

In this case, it tries to parse `< 1` as an opening HTML tag. Now granted the HTML markup should actually be escaped into `&lt;`, but this situation can and does occur in content and search module should not throw errors handling it.

## Steps To Reproduce

To reproduce the behavior:
1. Install search module
2. Create a node with the content `This is a test 0 < 1. With a <strong>bold tag</strong>.`
3. Run cron from the admin bar.
4. To repeat the error, invalidate the search index from the search settings page, then run cron again.

## Expected behavior

No PHP error should be shown after cron run.

image

## Additional information

quicksketch:
Fixes https://github.com/backdrop/backdrop-issues/issues/5728

### DIFF
--- a/core/modules/search/search.module
+++ b/core/modules/search/search.module
@@ -633,6 +633,10 @@ function search_index($sid, $module, $text) {
   $text = str_replace(array('<', '>'), array(' <', '> '), $text);
   $text = strip_tags($text, '<' . implode('><', array_keys($tags)) . '>');
 
+  // Remove any stray greater than or less than symbols that might not be
+  // properly HTML encoded.
+  $text = str_replace(array('< ', ' >'), ' ', $text);
+
   // Split HTML tags from plain text.
   $split = preg_split('/\s*<([^>]+?)>\s*/', $text, -1, PREG_SPLIT_DELIM_CAPTURE);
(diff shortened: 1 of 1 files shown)


core/modules/search/search.module
```diff
@@ -633,6 +633,10 @@ function search_index($sid, $module, $text) {
   $text = str_replace(array('<', '>'), array(' <', '> '), $text);
   $text = strip_tags($text, '<' . implode('><', array_keys($tags)) . '>');
 
+  // Remove any stray greater than or less than symbols that might not be
+  // properly HTML encoded.
+  $text = str_replace(array('< ', ' >'), ' ', $text);
+
```
quicksketch:
```suggestion
```

PR #4174, opened 2022-08-16
Issue #5728: Handle search terms filter empty values better.

### LINKED ISSUE #5729: Views: Search Filter: On Empty Input "Show None" option Shows All
## Description of the bug

When using a view to build a search page, the option for "On Empty Input" controls whether an empty dataset will be returned or if all results will be shown. This option seems to only work if the field is _also_ made required. This should not be a requirement, since landing on a search page URL directly should not immediately throw a validation error.

image

## Steps To Reproduce

To reproduce the behavior:
1. Enable search module.
2. Add a new view page of type content at any path, e.g. `/test-search`.
3. Add the filter for "Search: Search terms"
4. Change the "On empty input" option to "Show None". Make sure that the checkbox for "Required" is unchecked.
5. Save the view and visit the page.
6. With no search terms entered, no search results should be shown. Instead, all content is shown within the limit of the pager.

## Additional information

quicksketch:
Fixes https://github.com/backdrop/backdrop-issues/issues/5729

### DIFF
--- a/core/modules/search/views/views_handler_filter_search.inc
+++ b/core/modules/search/views/views_handler_filter_search.inc
@@ -13,6 +13,13 @@ class views_handler_filter_search extends views_handler_filter {
   var $always_multiple = TRUE;
 
   /**
+   * Set to always required, as the query for this filter should always run.
+   *
+   * Special handling of results is done with "On empty input" option.
+   */
+  var $always_required = TRUE;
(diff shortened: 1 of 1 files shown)


herbdool:
Replaced by https://github.com/backdrop/backdrop/pull/4175

PR #4175, opened 2022-08-16
Issue #5729: Properly follow Show All/Show None setting

### LINKED ISSUE #5729: Views: Search Filter: On Empty Input "Show None" option Shows All

indigoxela:
Fixes https://github.com/backdrop/backdrop-issues/issues/5729

### DIFF
--- a/core/modules/search/views/views_handler_filter_search.inc
+++ b/core/modules/search/views/views_handler_filter_search.inc
@@ -87,6 +87,18 @@ class views_handler_filter_search extends views_handler_filter {
     }
 
     $key = $this->options['expose']['identifier'];
+    // No search terms.
+    if (empty($form_state['values'][$key])) {
+      // The 'optional' property means: show all items if no input is set,
+      // 'required' means that no results should be shown if there's no input.
+      // @see views_handler_filter_search::operator_form()
+      if ($this->operator != 'optional' && empty($this->options['expose']['required'])) {
+        $this->view->built = TRUE;
+        $this->view->executed = TRUE;
+        $this->view->result = array();
+        return FALSE;
+      }
+    }
     if (!empty($form_state['values'][$key])) {
       $this->query_parse_search_expression($form_state['values'][$key]);
       if (count($this->search_query->words()) == 0) {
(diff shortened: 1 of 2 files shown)


core/modules/search/views/views_handler_filter_search.inc
```diff
@@ -87,6 +87,18 @@ class views_handler_filter_search extends views_handler_filter {
...
     $key = $this->options['expose']['identifier'];
+    // No search terms.
+    if (empty($form_state['values'][$key])) {
+      // The 'optional' property means: show all items if no input is set,
+      // 'required' means that no results should be shown if there's no input.
+      // @see views_handler_filter_search::operator_form()
+      if ($this->operator != 'optional' && empty($this->options['expose']['required'])) {
+        $this->view->built = TRUE;
+        $this->view->executed = TRUE;
+        $this->view->result = array();
```
quicksketch, line 98: Are setting these properties needed? What would happen if these were not set?

core/modules/search/views/views_handler_filter_search.inc
```diff
@@ -87,6 +87,18 @@ class views_handler_filter_search extends views_handler_filter {
...
+        return FALSE;
```
quicksketch: Maybe just return void, like the above return with no value.
```suggestion
        return;
```

core/modules/search/views/views_handler_filter_search.inc
indigoxela: Yes, they're needed.

Otherwise we end up with showing all, although "Show None" is checked with "On empty input".

indigoxela:
Random buddy BareUpgradePathTestCase strikes again. :stuck_out_tongue_winking_eye:

quicksketch:
> Random buddy BareUpgradePathTestCase strikes again. 😜

I updated this PR after https://github.com/backdrop/backdrop-issues/issues/7086 was fixed and the `BareUpgradePathTestCase` now seems to be fixed. :tada:

quicksketch:
Merged into 1.x and 1.33.x. Thanks @indigoxela!

PR #4181, opened 2022-08-26
Issue #5744: Datestamp default value2 not populated

### LINKED ISSUE #5744: Datestamp default values with value2 same as value - second value stays empty
## Description of the bug

While digging in the date module, I came across a bug. With a specific field + setup combination the default for value2 won't work.

## Steps To Reproduce

1. Attach a datestamp (Unix timestamp) field to a content type, collect end date
2. Set the field's "Default date" to some relative value
3. Set the field's "Default end date" to "_Same as Default date_"
4. Try to create a new node of that content type

## Actual behavior

The start default value is set, but the end date value is empty.

After digging a bit, it turned out that function [date_default_value_part](https://docs.backdropcms.org/api/backdrop/core%21modules%21date%21date.elements.inc/function/date_default_value_part/1) attempts to create a new BackdropDateTime from the timestamp, which is wrong for <s>two</s> one reason:

1. <s>That BackdropDateTime object has possibly already been created</s> nope, wrong
2. Creating a BackdropDateTime from a timestamp would require to prepend an `@`

The object exists, but has an error:

```
   'errors' => 
  array (
    'date' => 'The date "1661950450" does not match the expected format.',
  ),
```

empty-end-date

## Expected behavior

Both values should be set.

indigoxela:
Fixes https://github.com/backdrop/backdrop-issues/issues/5744

### DIFF
--- a/core/modules/date/date.elements.inc
+++ b/core/modules/date/date.elements.inc
@@ -905,8 +905,10 @@ function date_default_value_part($item, $field, $instance, $langcode, $part = 'v
       return NULL;
     }
     else {
+      // Consider timestamps.
+      $string = is_numeric($item[0]['value']) ? '@' . $item[0]['value'] : $item[0]['value'];
       // The date stored in 'value' has already been switched to the db timezone.
-      $date = new BackdropDateTime($item[0]['value'], $timezone_db, DATE_FORMAT_DATETIME);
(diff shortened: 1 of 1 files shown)


indigoxela:
My favorite random buddy: "Unable to find the wrapper "public""

quicksketch:
Merged into 1.x and 1.22.x. Thanks @indigoxela!

PR #4182, opened 2022-08-27
Issue 5731. Move page title to wrapper callback for user forms

### LINKED ISSUE #5731: Fixed: Forms `user_register_form` and `user_pass` hard-code the page title.
## Description of the bug

Forms `user_register_form` and `user_pass` hard-code the page title by using `backdrop_set_title()`. This messes up the "real" page title when these forms are used  as blocks in a Layout (with the contrib [Form Block](https://backdropcms.org/project/formblock)), or when adding the forms programmatically to a page in a custom module. 

(Clarification: the issue is a core problem, not one produced by Form Block).

## Steps To Reproduce

1. Install Form Block
2. Add the User Registration Form in a sidebar of the Default Layout 
3. Visit a node display (be sure you are logged out) and see the title of the node page being changed to "Create new account"
Screen Shot 2022-08-18 at 5 02 04 PM

## Actual behavior
The page title is force-changed by the block to "Create a new account"

## Expected behavior
The page title should still be the node's title.

## Additional information

The [line where the title is set](https://github.com/backdrop/backdrop/blob/8ce893d89106993c37ada5339355c9ead03bb8e8/core/modules/user/user.module#L3109) within the form function by using  `backdrop_set_title()` did not exist in Drupal 7.  This ws added 5 years ago as part of:

- https://github.com/backdrop/backdrop-issues/issues/260
and also as a reaction to:
- https://github.com/backdrop/backdrop-issues/issues/728
 
The advantage of using `backdrop_set_title()` in the forms is that it creates a more meaningful title for the user password and user register forms. Before `backdrop_set_title()` was added, Backdrop would produce the default page title of the parent item ("User account"),  since the paths for both `user/register` and `user/password`  are `MENU_LOCAL_TASK`.  

But the disadvantage of using `backdrop_set_title()` in the form is that it forces a specific page title onto other pages that contain the form as a block. This was [first reported in the forum](https://forum.backdropcms.org/forum/node-page-title-changes-create-new-account-when-user-reg-block-displayed). 

I really can't think of a way that preserves the nice form titles, and at the same time respects the page title where the form is present. 
 
- Backdrop CMS version: 1.22.2

argiepiano:
Fixes https://github.com/backdrop/backdrop-issues/issues/5731


nattywebdev:
Tested on Tugboat preview and works correctly (ie issue fixed). 
I'll test on my local dev shortly and report back.

nattywebdev:
I can confirm this fix also works on my local dev environment, correcting the original problem.
Many thanks to all involved and especially @argiepiano for a quick fix.

argiepiano:
Thanks. @nattywebdev . Can you please comment in the main issue rather than the pr? https://github.com/backdrop/backdrop-issues/issues/5731

Also if it works for you please add the appropriate WFM tag there. Thanks!

core/modules/user/user.module
```diff
@@ -3095,6 +3094,17 @@ function user_form_field_ui_field_edit_form_submit($form, &$form_state) {
   }
 }
 
+/** 
+ * Menu callback to create a new account. 
+ * 
+ * @see user_register_form()
+ */
```
docwilmot, line 3117: Could use an `@since 1.23 Replaces user_register_form() as menu callback` or something such. Same for other function.

argiepiano: This has been addressed in the latest push

argiepiano:
Test failure seems like an unrelated issue.

argiepiano (commented):
Update since label

argiepiano:
Reopening to trigger tests

quicksketch:

PR #4185, opened 2022-08-31
Issue #5752: Apply assertFieldChecked and assertNoFieldChecked to radios

### LINKED ISSUE #5752: [DX] `assertFieldChecked` and `assertNoFieldChecked` also apply to radios, but they only account for checkboxes
The `checked` property applies to both `type="checkbox"` and `type="radio"`, yet `BackdropWebTestCase::assertFieldChecked` and `BackdropWebTestCase::assertFieldChecked` only mention checkboxes.

The docblock description for these functions says:
```php
  /**
   * Asserts that a checkbox field in the current page is checked.
   *
```

And the default messages for these assertions are `Checkbox field @id is checked.` and `Checkbox field @id is not checked.`.

We should change that to account/mention both radios and checkboxes.

klonos:
Fixes https://github.com/backdrop/backdrop-issues/issues/5752

### DIFF
--- a/core/modules/simpletest/backdrop_web_test_case.php
+++ b/core/modules/simpletest/backdrop_web_test_case.php
@@ -3664,33 +3664,37 @@ class BackdropWebTestCase extends BackdropTestCase {
   /**
    * Asserts that a checkbox/radio field in the current page is checked.
    *
-   * @param $id
+   * @param string $id
    *   ID attribute of field to assert.
-   * @param $message
+   * @param string $message
    *   Message to display.
-   * @return
+   * @return boolean
    *   TRUE on pass, FALSE on fail.
    */
   protected function assertFieldChecked($id, $message = '', $group = 'Browser') {
     $elements = $this->xpath('//input[@id=:id]', array(':id' => $id));
     $type = $this->xpath('//input[@id=:id]/@type', array(':id' => $id));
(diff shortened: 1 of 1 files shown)


core/modules/simpletest/backdrop_web_test_case.php
```diff
@@ -3662,33 +3662,43 @@ protected function assertNoFieldById($id, $value = '', $message = '') {
...
+   *   The group this message belongs to.
+   *
+   * @return boolean
    *   TRUE on pass, FALSE on fail.
    */
-  protected function assertFieldChecked($id, $message = '') {
+  protected function assertFieldChecked($id, $message = '', $group = 'Browser') {
     $elements = $this->xpath('//input[@id=:id]', array(':id' => $id));
-    return $this->assertTrue(isset($elements[0]) && !empty($elements[0]['checked']), $message ? $message : t('Checkbox field @id is checked.', array('@id' => $id)), t('Browser'));
+    $message = $message ? $message : format_string('@type field @id is checked/selected.', array('@type' => $elements[0]['type'], '@id' => $id));
```
indigoxela, line 3679: This string assignment relies on array keys, that might eventually not be there.

```
'@type' => $elements[0]['type']
```
That can lead to "Trying to access array offset on value of type null" nagging, if "type" isn't set, or $elements is completely empty. Note that function xpath can return an empty array or even "false".

Probably an edge case, though.

PR #4189, opened 2022-09-03
Issue #5760: Show information about supported image types by GD

### LINKED ISSUE #5760: Status page: Show information about supported image types by GD
See: https://www.drupal.org/project/drupal/issues/3116611

We currently show this:
image

In Drupal, they currently have this:
image

And the suggestion for the new UI is this:
image

klonos:
Fixes https://github.com/backdrop/backdrop-issues/issues/5760

Alternative approach to https://github.com/backdrop/backdrop/pull/4188

### DIFF
--- a/core/modules/image/image.install
+++ b/core/modules/image/image.install
@@ -75,6 +75,7 @@ function image_field_schema($field) {
  */
 function image_requirements($phase) {
   $requirements = array();
+  $problems_detected = FALSE;
 
   if ($phase == 'runtime') {
     // Check for the PHP GD library.
@@ -141,7 +142,8 @@ function image_requirements($phase) {
         }
       }
 
-      if ($unsupported_formats) {
+      if (!empty($unsupported_formats)) {
+        $problems_detected = TRUE;
         $requirements['image_gd']['severity'] = REQUIREMENT_WARNING;
         $fix_info[] = t('If you want to add support for any unsupported image format, check the <a href="@url">PHP GD installation documentation</a>.', array('@url' => 'https://www.php.net/manual/image.installation.php'));
       }
@@ -169,12 +171,17 @@ function image_requirements($phase) {
         }
       }
       if (!empty($unsupported_effects)) {
+        $problems_detected = TRUE;
         $fix_info[] = t('The GD library was compiled without support for functions required by certain effects. It was probably compiled using the official GD libraries from http://www.libgd.org instead of the GD library bundled with PHP. If you want to add support for these effects, recompile PHP --with-gd using the bundled GD library. See <a href="http://www.php.net/manual/book.image.php">the PHP manual</a>.');
       }
 
-      $formats_and_effects = theme('item_list', array('items' => $descriptions, 'title' => t('Supported image formats and effects'), 'attributes' => array('class' => array('requirements'))));
+      $formats_and_effects = theme('item_list', array('items' => $descriptions, 'attributes' => array('class' => array('requirements'))));
       $fix_info = theme('item_list', array('items' => $fix_info, 'title' => t('Fixing issues')));
       $requirements['image_gd']['description'] = $formats_and_effects . $fix_info;
+
+      $support_status = $problems_detected ? t('Partial support of image formats and effects') : t('All image formats and effects supported');
+      $requirements['image_gd']['value'] .= ' - ' . $support_status;
+
     }
     else {
       $requirements['image_gd'] = array(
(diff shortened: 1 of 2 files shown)


core/modules/image/image.install
```diff
@@ -80,18 +80,101 @@ function image_requirements($phase) {
...
+            'data' => t('@effect effect supported.', array('@effect' => $effect)),
+            'class' => array('status-ok'),
+          );
+        }
+      }
+      if (!empty($unsupported_effects)) {
+        $fix_info[] = t('The GD library was compiled without support for functions required by certain effects. It was probably compiled using the official GD libraries from http://www.libgd.org instead of the GD library bundled with PHP. If you want to add support for these effects, recompile PHP --with-gd using the bundled GD library. See <a href="http://www.php.net/manual/book.image.php">the PHP manual</a>.');
       }
+
+      $formats_and_effects = theme('item_list', array('items' => $descriptions, 'title' => t('Supported image formats and effects'), 'attributes' => array('class' => array('requirements'))));
```
olafgrabienski: Under the heading "Supported image formats and effects" may be listed supported and unsupported things. I'd suggest to replace "Supported" with "Support of" in the heading: `Support of image formats and effects`

core/modules/system/css/system.admin.css
```diff
@@ -221,6 +221,43 @@ span.admin-missing {
...
+[dir="rtl"] .system-status-report ul.requirements li.status-error {
+  padding-right: 1em;
+}
+.system-status-report ul.requirements li.status-ok:before {
+  content: "✓";
+  color: green;
+  padding-right: 1ch; /* LTR */
+}
+.system-status-report ul.requirements li.status-error:before {
+  content: "✗";
```
indigoxela, line 252: Tiny accessibility nit-pick: contrast to the light grey background is only 3.73 (fails AA).

As far as I know, screen readers do read CSS content - not sure what they read for these characters, and if it's useful or distracting.

core/modules/image/image.install
klonos: Good point @olafgrabienski 👍🏼 ...not sure if either of these options are ideal. I have an idea to move this text from being the title of the list of image format/effects to being appended to the value of the requirement as a status (whether all or only some effects/formats are supported).

I'll reply to the issue, and add some screenshots as well.

core/modules/system/css/system.admin.css
klonos: Good point @indigoxela 👍🏼 ...working on it.

klonos: ...I've used https://webaim.org/resources/contrastchecker/ to check `#ad0000` and `#006100` against all possible table row colors in the Seven admin theme:
- `#ffffff` (odd table rows)
- `#f7f7f7` (even table rows)
- `#ffff99` (hovered-over rows)

klonos: ...I've also used ChromeVox on my Mac, and it read the items properly  as:
`List with 6 items - GIF image format supported - List item - JPEG image format supported - List item ...`

So I think we are good re screen readers 👍🏼 

PS: while using the screen reader, I spotted may other issues with the status report page and tables in general, but that's a larger issue that should be dealt with elsewhere - we are NOT making things worse with this PR here.

core/modules/system/css/system.admin.css
```diff
@@ -221,6 +221,43 @@ span.admin-missing {
...
+  list-style: none;
+  padding-left: 1em; /* LTR */
+  text-indent: -2ch;
+}
+[dir="rtl"] .system-status-report ul.requirements li.status-ok,
+[dir="rtl"] .system-status-report ul.requirements li.status-error {
+  padding-right: 1em;
+}
+.system-status-report ul.requirements li.status-ok:before {
+  content: "✓";
```
indigoxela, line 247: A quick question: How about adding an actual space? `content:  "✓ ";`

Wouldn't that reduce the necessary CSS code? Or is that a problem with RTL languages?

klonos: Sorry for taking that long to reply @indigoxela ...hmm 🤔 

I don't have a specific preference, but the CSS code reduction would basically only be the `padding-left`/`padding-right`. Right? ...is that worth it? 🤔 ...not sure about RTL - I'd have to test that and see. However, adding an actual space character in order to create CSS padding feels hackish to me, but searching google with "css content space instead of padding" didn't provide any specific articles saying whether this is bad practice or not. I'd say lets wait for more people with CSS/frontend experience to chime in. Happy to change once we have more feedback.

PR #4190, opened 2022-09-05
Issue #5761: Log completely unusable when an entry has corrupt serialized data

### LINKED ISSUE #5761: Log completely unusable when an entry has corrupt serialized data.
This is to crossport https://www.drupal.org/project/drupal/issues/2790857, which was recently committed to D7 core:

> If an entry in the dblog has corrupt serialized data, the dblog page (`/admin/reports/dblog`) will throw an exception and will not load.
> 
> If unserialize fails, then a boolean is sent to `t()` leading to a fatal exception.
> `$output = t($event->message, unserialize($event->variables));`

klonos:
Fixes https://github.com/backdrop/backdrop-issues/issues/5761

Crossport of this D7 commit: https://git.drupalcode.org/project/drupal/-/commit/30e76c6

### DIFF
--- a/core/modules/dblog/dblog.admin.inc
+++ b/core/modules/dblog/dblog.admin.inc
@@ -302,13 +302,19 @@ function dblog_format_message($event, $is_link = FALSE) {
   $output = '';
   // Check for required properties.
   if (isset($event->message) && isset($event->variables)) {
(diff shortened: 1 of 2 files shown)


quicksketch:
Thanks, merged into 1.x and 1.26.x!

PR #4193, opened 2022-09-05
Issue #5765: For accessibility, images that link should have the dest…

### LINKED ISSUE #5765: [A11Y] When an image links to content, the alt tag should be the title of the destination
## Description of the bug

When a screenreader is navigating a page and comes across an image that links somewhere, the screenreader needs to be able to explain where the link will take the navigator. 

This is commonly done by changing the `alt` tag on the image from a description of the image, to a description of the destination.

If the image is purely decorative, and is not necessary for navigation, another option would be to add a `tabindex=-1` on the link to skip it entirely for screenreaders. (If the post also has a text heading that links to the content, for example, it would be safe to skip the image for screenreaders)

Since the CMS cannot know when it is safe to skip the image (as in a teaser display mode) and when it is not safe to skip the image (as in a view, where the image/link is the only navigation) I think it would be safer to always add the alt text. This would ensure that all Backdrop sites have all the necessary navigation tools OOTB, and if people prefer to hide some elements to improve screen reader navigation, they can adjust their own sites further. 

I believe we should incorporate this change into [theme_image_formatter()](https://docs.backdropcms.org/api/backdrop/core%21modules%21image%21image.theme.inc/function/theme_image_formatter/1).

If we passed along an optional ALT tag as a 4th parameter, we could use that to override the image's specified ALT tag if provided.

For example, in [image_field_formatter_view()](https://docs.backdropcms.org/api/backdrop/core%21modules%21image%21image.field.inc/function/image_field_formatter_view/1)(
```php
    $element[$delta] = array(
      '#theme' => 'image_formatter',
      '#item' => $item,
      '#image_style' => $display['settings']['image_style'],
      '#path' => isset($uri) ? $uri : '',
      '#alt' => $entity->label(),
    );
```

and in [theme_image_formatter()](https://docs.backdropcms.org/api/backdrop/core%21modules%21image%21image.theme.inc/function/theme_image_formatter/1):
```php
function theme_image_formatter($variables) {
  $item = $variables['item'];

  // Do not output an empty 'title' attribute.
  if (isset($item['title']) && backdrop_strlen($item['title']) == 0) {
    unset($item['title']);
  }

  // Override the alt tag with destination information if the image is being used as a link. 
  if (isset($variables['path']['path']) && isset($variables['alt'])) {
    $item['alt'] = $variables['alt'];
  }

  if ($variables['image_style']) {
    $item['style_name'] = $variables['image_style'];
    $output = theme('image_style', $item);
  }
  else {
    $output = theme('image', $item);
  }

  // The link path and link options are both optional, but for the options to be
  // processed, the link path must at least be an empty string.
  if (isset($variables['path']['path'])) {
    $path = $variables['path']['path'];
    $options = isset($variables['path']['options']) ? $variables['path']['options'] : array();
    // When displaying an image inside a link, the html option must be TRUE.
    $options['html'] = TRUE;
    $output = l($output, $path, $options);
  }

  return $output;
}
```

Resources: 
* https://www.w3.org/WAI/tutorials/images/functional/

jenlampton:
Fixes https://github.com/backdrop/backdrop-issues/issues/5765

### DIFF
--- a/core/modules/image/image.theme.inc
+++ b/core/modules/image/image.theme.inc
@@ -386,6 +386,12 @@ function theme_image_formatter($variables) {
     unset($item['title']);
   }
 
+  // Override the alt tag with destination information when the image is linked. 
+  if (isset($variables['path']['path']) && isset($variables['alt'])) {
+    $item['alt'] = $variables['alt'];
+  }
+
+  // Use an image style if requested.
   if ($variables['image_style']) {
     $item['style_name'] = $variables['image_style'];
     $output = theme('image_style', $item);
--- a/core/modules/image/image.field.inc
+++ b/core/modules/image/image.field.inc
@@ -677,10 +677,12 @@ function image_field_formatter_settings_summary($field, $instance, $view_mode) {
  */
 function image_field_formatter_view($entity_type, $entity, $field, $instance, $langcode, $items, $display) {
   $element = array();
+  $alt = NULL;
 
   // Check if the formatter involves a link.
   if ($display['settings']['image_link'] == 'content') {
     $uri = entity_uri($entity_type, $entity);
+    $alt = $entity->label();
   }
   elseif ($display['settings']['image_link'] == 'file') {
     $link_file = TRUE;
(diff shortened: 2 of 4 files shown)


core/modules/image/image.theme.inc
```diff
@@ -386,6 +386,12 @@ function theme_image_formatter($variables) {
     unset($item['title']);
   }
 
+  // Override the alt tag with destination information when the image is linked.
```
klonos, line 389:
```suggestion
  // Override the alt attribute with destination information when the image is
  // linked.
```

PR #4197, opened 2022-09-08
Issue #5771: Add system requirements check for sufficient max_input_v…

### LINKED ISSUE #5771: [UX] Add Requirements check for max_input_vars to status report
Today @quicksketch mentioned that maybe we should split the status report update out of the PR for [[UX] set warning messages if max_input_vars is not sufficiently large](https://github.com/backdrop/backdrop-issues/issues/3824).

The status report message could stand on its own, and could get into the next minor version, and would be a benefit on its own.

We can keep working on the error messaging in [the other issue](https://github.com/backdrop/backdrop-issues/issues/3824) and improve as needed.

Advocate: @jenlampton

jenlampton:
Fixes https://github.com/backdrop/backdrop-issues/issues/5771

### DIFF
--- a/core/modules/system/system.install
+++ b/core/modules/system/system.install
@@ -113,7 +113,7 @@ function system_requirements($phase) {
   // Test PHP max_input_vars setting.
   $max_input_vars = ini_get('max_input_vars');
   $requirements['php_max_input_vars'] = array(
-    'title' => $t('PHP max_input_vars setting'),
+    'title' => $t('PHP <code>max_input_vars</code> setting'),
     'value' => $t('PHP limits the number of input variables in each form to @value items.', array('@value' => number_format($max_input_vars, 0, '.', ','))),
   );
   // 1000 is the default value and indicates potential issues.
--- a/core/modules/system/system.install
+++ b/core/modules/system/system.install
@@ -110,6 +110,19 @@ function system_requirements($phase) {
     $requirements['php_register_globals']['value'] = $t('Disabled');
   }
 
+  // Test PHP max_input_vars setting.
+  $max_input_vars = ini_get('max_input_vars');
+  $requirements['php_max_input_vars'] = array(
+    'title' => $t('PHP max_input_vars setting'),
+    'value' => $t('PHP limits the number of input variables in each form to @value items.', array('@value' => number_format($max_input_vars, 0, '.', ','))),
+  );
+  // 1000 is the default value and indicates potential issues.
+  // @see https://www.php.net/manual/en/info.configuration.php
+  if ($max_input_vars <= 1000) {
+    $requirements['php_max_input_vars']['description'] = $t('When set too low, the <code>max_input_vars</code> PHP setting can cause silent failures on large forms, such as the permissions page. A value of <code>10000</code> is recommended. <a href="@documentation" target="_blank">Read more about increasing the <code>max_input_vars</code> PHP setting</a>.', array('@documentation' => url('https://docs.backdropcms.org/max-input-vars')));
+    $requirements['php_max_input_vars']['severity'] = REQUIREMENT_WARNING;
+  }
+
   // Test for PHP extensions.
   $requirements['php_extensions'] = array(
     'title' => $t('PHP extensions'),
--- a/core/modules/system/system.install
+++ b/core/modules/system/system.install
@@ -119,8 +119,11 @@ function system_requirements($phase) {
   // 1000 is the default value and indicates potential issues.
   // @see https://www.php.net/manual/en/info.configuration.php
   if ($max_input_vars <= 1000) {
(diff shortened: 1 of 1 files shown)


core/modules/system/system.install
```diff
@@ -110,6 +110,19 @@ function system_requirements($phase) {
...
+  $max_input_vars = ini_get('max_input_vars');
+  $requirements['php_max_input_vars'] = array(
+    'title' => $t('PHP max_input_vars setting'),
+    'value' => $t('PHP limits the number of input variables in each form to @value items.', array('@value' => number_format($max_input_vars, 0, '.', ','))),
+  );
+  // 1000 is the default value and indicates potential issues.
+  // @see https://www.php.net/manual/en/info.configuration.php
+  if ($max_input_vars <= 1000) {
+    $requirements['php_max_input_vars']['description'] = $t('When set too low, the <code>max_input_vars</code> PHP setting can cause silent failures on large forms, such as the permissions page. A value of <code>10000</code> is recommended. <a href="@documentation" target="_blank">Read more about increasing the <code>max_input_vars</code> PHP setting</a>.', array('@documentation' => url('https://docs.backdropcms.org/max-input-vars')));
+    $requirements['php_max_input_vars']['severity'] = REQUIREMENT_WARNING;
```
quicksketch, line 123: Perhaps downgrade the notice so this does not throw a warning on installation and updates:

```suggestion
    $requirements['php_max_input_vars']['severity'] = $phase === 'runtime' ? REQUIREMENT_WARNING : REQUIREMENT_INFO;
```

Like @indigoxela said in the issue, that might be causing the failures. I think update.php puts in an interstitial warning page if you have active warnings during update.

core/modules/system/system.install
```diff
@@ -110,6 +110,19 @@ function system_requirements($phase) {
...
+  // Test PHP max_input_vars setting.
+  $max_input_vars = ini_get('max_input_vars');
+  $requirements['php_max_input_vars'] = array(
+    'title' => $t('PHP max_input_vars setting'),
+    'value' => $t('PHP limits the number of input variables in each form to @value items.', array('@value' => number_format($max_input_vars, 0, '.', ','))),
+  );
+  // 1000 is the default value and indicates potential issues.
+  // @see https://www.php.net/manual/en/info.configuration.php
+  if ($max_input_vars <= 1000) {
+    $requirements['php_max_input_vars']['description'] = $t('When set too low, the <code>max_input_vars</code> PHP setting can cause silent failures on large forms, such as the permissions page. A value of <code>10000</code> is recommended. <a href="@documentation" target="_blank">Read more about increasing the <code>max_input_vars</code> PHP setting</a>.', array('@documentation' => url('https://docs.backdropcms.org/max-input-vars')));
```
quicksketch: No need for `url()` when using an absolute URL for documentation.

```suggestion
    $requirements['php_max_input_vars']['description'] = $t('When set too low, the <code>max_input_vars</code> PHP setting can cause silent failures on large forms, such as the permissions page. A value of <code>10000</code> is recommended. <a href="@documentation" target="_blank">Read more about increasing the <code>max_input_vars</code> PHP setting</a>.', array('@documentation' => 'https://docs.backdropcms.org/max-input-vars'));
```

core/modules/system/system.install
jenlampton: > that might be causing the failures. 

But... there shouldn't be any active warnings if the server is set up correctly. Are our test servers not? Do we not have the ability to fix them?

> Perhaps downgrade the notice so this does not throw a warning on installation and updates:

This shouldn't be an "info" status if something is actually wrong -- people need to know about the problem in order to fix it. I don't think we should sacrifice the value of the warning just because we can't set up our test servers correctly?

jenlampton: Update: @quicksketch says that other updates also have a special status during install (and update) that are different from the actual status that appears on the status report. I'm going to rebase and first, and go from there :)

jenlampton: okay, I added 
```php
    if ($phase == 'install' || $phase == 'update') {
      $requirements['php_max_input_vars']['severity'] = REQUIREMENT_INFO;
    }
```

core/modules/system/system.install
```diff
@@ -110,6 +110,22 @@ function system_requirements($phase) {
     $requirements['php_register_globals']['value'] = $t('Disabled');
   }
 
+  // Test PHP max_input_vars setting.
+  $max_input_vars = ini_get('max_input_vars');
+  $requirements['php_max_input_vars'] = array(
+    'title' => $t('PHP max_input_vars setting'),
```
```suggestion
    'title' => $t('PHP <code>max_input_vars</code> setting'),
```

quicksketch:
Merged into 1.x and 1.23.x. Thanks @jenlampton and @klonos!

PR #4202, opened 2022-09-15
Issue #5780: Custom list types for nested items in theme_item_list()

### LINKED ISSUE #5780: [D8] Support custom list types in `theme_item_list()` for children/nested items
This is the respective issue for https://www.drupal.org/project/drupal/issues/1785310 (change record: [theme_item_list() supports custom list types for child elements](https://www.drupal.org/node/1816204)), which was actually part of our list in #166, but never actually got implemented/backported to Backdrop.

As it is currently in D7 and in Backdrop, nested list items within lists generated by `theme_item_list()` inherit their type (`<ul>`/`<ol>`) from the `type` variable specified in the parent items. This means that you cannot easily/properly create lists where the children items may need to be of a different type than the parent list.

So I was working on another issue, and I needed to generate this sort of a list:

```html
<ol>
  <li>parent item 1
    <ul>
      <li>child 1</li>
      <li>child 2</li>
    </ul>
  </li>
  <li>item 2</li>
  <li>item 3</li>
</ol>
```

Which I expected be rendered as follows:
```
1. parent item 1
   • child 1
   • child 2
3. item 2
4. item 3
```

So I have tried the following code (which works in D8/9):

```php
  $items = array();

  $items[] = array(
    'data' => 'parent item 1',
    'children' => array(
      array(
        'data' => 'child 1',
      ),
      array(
        'data' => 'child 2',
      ),
    ),
    'type' => 'ul',
  );
  $items[] = 'item 2';
  $items[] = 'Item 3';

  $return theme('item_list', array('items' => $items, 'type' => 'ol'));
```

Which rendered as follows (nested items ignored the `ul` type I've specified for them, and rendered as `<ol>` instead - inheriting that from their parent):
```
1. parent item 1
   1. child 1
   2. child 2
3. item 2
4. item 3
```

The workaround I resorted to was this:
```php
  $items = array();
  $children = array();

  $children[] = array(
    'data' => 'child 1',
  );
  $children[] = array(
    'data' => 'child 2',
  );

  $items[] = array(
    'data' => 'parent item 1' . theme('item_list', array('items' => $children, 'type' => 'ul')),
  );
  $items[] = 'item 2';
  $items[] = 'Item 3';

  $return theme('item_list', array('items' => $items, 'type' => 'ol'));
```

That works, but it feels hacky. We should support this the same way as in D9, both because it feels like the proper way to do it, but also because it would allow D9 snippets of code (either backports or things provided in how-to's and solution articles) to be easily converted to Backdrop.

klonos:
Fixes https://github.com/backdrop/backdrop-issues/issues/5780

Backport of https://www.drupal.org/project/drupal/issues/1785310


klonos:
Random failures in php 5.6. Restarting...

klonos:
Test gremlins: https://github.com/backdrop/backdrop-issues/issues/5721

core/includes/theme.inc
```diff
@@ -2391,10 +2391,13 @@ function theme_mark($variables) {
  *     - children: A list of nested child items to render that behave
  *       identically to 'items', but any non-numeric string keys are treated as
  *       HTML attributes for the child list that wraps 'children'.
+ *     - type: The type of list to return for the nested/children items (e.g.
+ *       "ul", "ol").
```
bugfolder: How about this:

`- type: the type of list to return for the children ('ul' or 'ol').`

Since you've already defined what "children" is in the preceding line, you can use that term. And since most people will single-quote 'ul' and 'ol' in usage, why not here? And 'ul' and 'ol' are the only choices, so no "e.g." needed.

Same comments would apply to the second (existing) usage of "type".

klonos:
```suggestion
 *     - type: The type of list to return for the children ('ul' or 'ol').
```

core/includes/theme.inc
```diff
@@ -2391,10 +2391,12 @@ function theme_mark($variables) {
  *     - children: A list of nested child items to render that behave
  *       identically to 'items', but any non-numeric string keys are treated as
  *       HTML attributes for the child list that wraps 'children'.
+ *     - type: The type of list to return for the children ('ul' or 'ol').
  *     Any other key/value pairs are used as HTML attributes for the list item
  *     in 'data'.
  *   - title: The title of the list.
- *   - type: The type of list to return (e.g. "ul", "ol").
+ *   - type: The type of list to return for the main/parent list (e.g. "ul",
+ *     "ol").
```
klonos:
```suggestion
 *   - type: The type of list to return for the parent list ('ul' or 'ol').
```

core/includes/theme.inc
```diff
@@ -2404,6 +2405,7 @@ function theme_mark($variables) {
  *
  * @since 1.21.0: added the "empty" variable.
  * @since 1.27.0: added the "wrapper_attributes" variable.
+ * @since ???: added support for a "type" variable in nested/children items.
```
klonos:
```suggestion
 * @since 1.28.3: added support for a "type" variable in nested/children items.
```

core/themes/seven/css/style.css
```diff
@@ -22,7 +22,11 @@
 [dir="rtl"] .item-list ul {
   margin: 0.25em 1.5em 0.25em 0;
 }
-.item-list ul li,
+.item-list ol>li {
+  list-style-type: decimal;
+  list-style-image: none;
+}
+.item-list ul>li,
```
quicksketch, line 29: I think we could avoid adding new more-specific selectors here by eliminating the unnecessary selector in `seven.base.css` instead. I filed a PR against this one at https://github.com/klonos/backdrop/pull/13

PR #4207, opened 2022-09-16
Issue #5706: Update all documentation URLs in core

### LINKED ISSUE #5706: Update all documentation URLs in core
We have a lot of links to Backdrop CMS documentation throughout core. The documentation site once lived at https://api.backdropcms.org. We've since moved it to https://docs.backdropcms.org. 

We should do a find/replace for api.backdropcms.org/docs.backdropcms.org in all files in core, so all links can point to the correct destination, without needing to leverage a redirect. 

Example page: Configuration manager (admin/config/development/configuration)

Example text: Read more about [managing configuration](https://api.backdropcms.org/documentation/working-with-configuration).

CrisT195:
Fixes https://github.com/backdrop/backdrop-issues/issues/5706

### DIFF
--- a/core/profiles/standard/standard.install
+++ b/core/profiles/standard/standard.install
@@ -114,7 +114,7 @@ function standard_install() {
 
   // Create default pre-defined node types. For a complete list of available
   // node type attributes please refer to the node type API documentation at:
-  // https://api.backdropcms.org/api/backdrop/core!modules!node!node.module/function/node_type_save
+  // https://docs.backdropcms.org/api/backdrop/core!modules!node!node.module/function/node_type_save
   $types = array(
     array(
       'type' => 'page',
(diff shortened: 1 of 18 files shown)


core/profiles/standard/standard.install
```diff
@@ -248,7 +248,7 @@ function standard_install() {
 
   // Create an image field named "Image", enabled for the 'post' content type.
   // Many of the following values will be defaulted, they're included here as an illustrative examples.
-  // See https://api.backdropcms.org/api/function/field_create_field/1
+  // See https://docs.backdropcms.org/api/backdrop/core%21modules%21field%21field.crud.inc/function/field_create_field/1
```
jenlampton, line 251: This looks funny to me, but the link is correct. I think we have a problem with the API module that's generating this link. I have opened https://github.com/backdrop-contrib/api/issues/7 to investigate further, but that should not block this issue from proceeding.

quicksketch:
Merged into 1.x and 1.23.x. Thanks @CrisT195!

PR #4212, opened 2022-09-24
Issue 5777. Batch-process file type classification

### LINKED ISSUE #5777: File Access Checks Fail if no Type is defined (throw a warning or watchdog)?
## Description of the bug

When upgrading from Drupal 7 or early versions of Backdrop (before File Entities were introduced), if using the "Download Link" formatter, the link is never displayed unless the user has `bypass file access`. This seems to be because `file_access()` with an `$op = 'download'` fails when files have no `type` attribute specified (which is done on cron jobs).

Note this only affects the "Download link" formatter. If the formatter is "Generic file" or "URL to file", both show and the file can be downloaded.

## Steps To Reproduce

- Upgrade from Drupal 7 with a simple site that includes a file field.
- After upgrade, set the file formatter to "Download link"
- Now log out and view as anonymous. The download link is hidden.
- Grant the "Audio: Download any files", "Document: Download any files", "Image: Download any files", and "Video: Download any files" permissions. The download link is still hidden.

This appears to be caused by upgraded files not yet having a "type" specified. The type gets specified over a series of cron jobs, but the situation of a site upgrade with thousands of files, that could take several days for the type data to fully get populated.

## Additional information

- Backdrop CMS version: 1.22.3
- Browser(s) and their versions: All

argiepiano:
Fixes https://github.com/backdrop/backdrop-issues/issues/5777

### DIFF
--- a/core/modules/file/file.admin.inc
+++ b/core/modules/file/file.admin.inc
@@ -677,3 +677,41 @@ function _file_sort_weight_label($a, $b) {
     return $a_weight < $b_weight ? -1 : 1;
   }
 }
+
+/**
+ * Confirm form for file type classification process.
+ *
+ * @see file_type_classify_confirm_submit()
+ * @see file_menu()
+ *
+ * @ingroup forms
+ */
+function file_type_classify_confirm($form, $form_state) {
+  return confirm_form(array(), t('Are you sure you want to classify the file types?'),
+  'admin/reports/status', t('This action will classify all files that are missing a type, and may be a lengthy process. This action cannot be undone.'), t('Classify file types'), t('Cancel'));
+}
+
+/**
+ * Submit handler for file type classification confirmation form.
+ * 
+ * @see file_type_classify_confirm()
+ */
+function file_type_classify_confirm_submit($form, &$form_state) {
+  $result = db_query('SELECT fid FROM {file_managed} WHERE `type` = :db_condition_placeholder', array(':db_condition_placeholder' => 'undefined'));
+  $fids = $result->fetchCol();
+  if (!empty($fids)) {
+    $batch = array(
+      'title' => t('Classifying files'),
+      'operations' => array(
+        array('file_type_classify_batch', array($fids)),
+      ),
+      'finished' => 'file_type_classify_finished',
+    );
+    batch_set($batch);
+  }
+  else {
+    file_needs_type_classification(FALSE);
+  }
+
+  $form_state['redirect'] = 'admin/reports/status';
+}
--- a/core/modules/file/file.admin.inc
+++ b/core/modules/file/file.admin.inc
@@ -687,8 +687,17 @@ function _file_sort_weight_label($a, $b) {
  * @ingroup forms
  */
 function file_type_classify_confirm($form, $form_state) {
-  return confirm_form(array(), t('Are you sure you want to classify the file types?'),
-  'admin/reports/status', t('This action will classify all files that are missing a type, and may be a lengthy process. This action cannot be undone.'), t('Classify file types'), t('Cancel'));
+  $form = confirm_form(
+    array(),
+    t('Are you sure you want to classify the file types?'),
+    'admin/reports/status',
+    t('This action will classify all files that are missing a type, and may be a lengthy process.'),
+    t('Classify file types'),
+    t('Cancel')
+  );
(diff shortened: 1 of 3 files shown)


core/modules/file/file.module
```diff
@@ -429,6 +429,15 @@ function file_menu() {
     'file' => 'file.pages.inc',
   );
 
+  $items['admin/reports/status/classify'] = array(
```
quicksketch: The closest example I can think of to this kind of batch update is the UTF8-MB4 database update, which lives at `admin/config/development/utf8mb4-upgrade`, basically in the menu where it would be if it were a normal menu item.

Since this is file-type related, I'd probably put it at:
```suggestion
  $items['admin/structure/file-types/classify'] = array(
```

core/modules/file/file.module
```diff
@@ -429,6 +429,15 @@ function file_menu() {
...
+    'title' => 'Classify file types',
+    'page callback' => 'backdrop_get_form',
+    'page arguments' => array('file_type_classify_confirm'),
+    'access arguments' => array('access administration pages'),
+    'type' => MENU_CALLBACK,
```
quicksketch: Also from the UTF8-MB4 update:

```suggestion
    'type' => MENU_VISIBLE_IN_BREADCRUMB,
```

This keeps the breadcrumb working unlike `MENU_CALLBACK` but does not show up in menus.

core/modules/file/file.module
```diff
@@ -2613,16 +2622,6 @@ function _file_sort_array_by_weight(&$a) {
...
 
-/*
- * Implements hook_cron_queue_info().
- */
-function file_cron_queue_info() {
-  $queues['file_type_determine'] = array(
-    'worker callback' => 'file_type_determine',
-  );
-  return $queues;
-}
```
quicksketch, line 2624: I think removing this queue is a good move, the batch processing page is a lot more accessible. But we might want to have another update hook to remove this queue if it exists? Otherwise either an error might occur trying to process the queue on cron jobs, or maybe the queue will just never be emptied?

core/modules/file/file.admin.inc
```diff
@@ -677,3 +677,41 @@ function _file_sort_weight_label($a, $b) {
...
+ * Confirm form for file type classification process.
+ *
+ * @see file_type_classify_confirm_submit()
+ * @see file_menu()
+ *
+ * @ingroup forms
+ */
+function file_type_classify_confirm($form, $form_state) {
+  return confirm_form(array(), t('Are you sure you want to classify the file types?'),
+  'admin/reports/status', t('This action will classify all files that are missing a type, and may be a lengthy process. This action cannot be undone.'), t('Classify file types'), t('Cancel'));
```
quicksketch: This confirm form is unnecessarily scary:
image

I think we can remove "This action cannot be undone", since it's non-destructive. And we can change the submit button to remove the `button-danger` class, so it shows as a primary (blue) button.

core/modules/file/file.module
argiepiano: Done

core/modules/file/file.module
argiepiano: Done

core/modules/file/file.module
argiepiano: Added a new update hook implementation to remove the queue.

core/modules/file/file.admin.inc
argiepiano: I had to manually change the css type of the button, as `confirm_form` uses `button-danger` by default.

core/modules/file/file.install
```diff
@@ -1199,3 +1201,11 @@ function file_update_1008() {
   }
   db_create_table('cache_entity_file', $table);
 }
+
+/** 
+ * Remove the file_type_determine queue.
+ */
+function file_update_1009() {
+  $queue = BackdropQueue::get('file_type_determine');
```
quicksketch, line 1209: If there are still items in the queue, then the batch needs to be run to finish it:

```suggestion
  $queue = BackdropQueue::get('file_type_determine');
  if ($queue->numberOfItems() > 0) {
    file_needs_type_classification(TRUE);
  }
```

argiepiano: Of course. Added!

quicksketch:

PR #4213, opened 2022-09-26
Issue 4182. Add manual rebuild of search indexes

### LINKED ISSUE #4182: Change the "Invalidate search index" process to a batched "Rebuild search index"
I like the neatness of the first screenshot that @jenlampton proposed on #1321 (comment). 

screen shot 2015-11-05 at 3 08 11 pm

Placing the same warning message twice on the search settings form:

Screen Shot 2019-10-31 at 10 13 54 PM

and then again on the confirm page:

 screen shot 2015-11-05 at 3 08 46 pm

looks really too excessive. 

I do understand the initial concern for some people who could press the button without realizing the consequences. However, in fact the consequences are not so scary, because (1) no data is lost anyway and then (2) those who pressed the button will see the warning on the confirm page anyway.

Making the button outstanding with the scary text in red color was sufficient enough, so if possible please remove the description text.

argiepiano:
Fixes https://github.com/backdrop/backdrop-issues/issues/4182


core/modules/search/search.admin.inc
```diff
@@ -5,26 +5,60 @@
...
  */
 function search_reindex_confirm_submit(&$form, &$form_state) {
   if ($form['confirm']) {
-    search_reindex();
-    backdrop_set_message(t('The search index will be invalidated.'));
+    $config = config('search.settings');
+    $limit_per_pass = $config->get('search_cron_limit');  
+
+    // Wipe out all indexes.
+    module_invoke_all('search_reset');
```
quicksketch: This line accomplishes the same thing and it might be a good idea to re-use the existing API rather than duplicating calls to module_invoke_all():

```suggestion
    // Mark all content as needing to be re-indexed.
    search_reindex();
```

core/modules/search/search.module
```diff
@@ -1422,3 +1422,61 @@ function search_config_info() {
...
+    $config = config('search.settings');
+    // Collect some stats
+    $remaining = 0;
+    $total = 0;
+    foreach ($config->get('search_active_modules') as $module) {
+      if ($status = module_invoke($module, 'search_status')) {
+        $remaining += $status['remaining'];
+        $total += $status['total'];
+      }
+    }
```
quicksketch, line 1496: This same bit of code is in `search_admin_settings()` and something similar in `search_reindex_confirm_submit()` above. Could we put this into a new API function? Maybe named `search_get_status($module = NULL)` with a parameter for which module to check, or NULL for all modules like this.

core/modules/search/search.admin.inc
argiepiano: Done.

core/modules/search/search.module
argiepiano: Done. Added the function `search_get_stats()`  at the end of search.module

argiepiano:
Re opening to run tests again

core/modules/search/tests/search.test
```diff
@@ -1384,13 +1384,30 @@ class SearchConfigSettingsForm extends BackdropWebTestCase {
...
-    // Test the Invalidate search index button.
-    $this->backdropPost('admin/config/search/settings', array(), t('Invalidate search index'));
-    $this->assertText(t('Are you sure you want to invalidate the search index?'));
-    $this->backdropPost('admin/config/search/settings/reindex', array(), t('Invalidate search index'));
-    $this->assertText(t('The search index will be invalidated'));
+    // Reset indexes and verify the count of items left to index is correct.
+    search_reindex();
+
+    // Create a new node to test index rebuilding.
+    $body_text = 'Glutten-free.';
```
quicksketch:
```suggestion
    $body_text = 'Gluten-free.';
```

:joy:

argiepiano: Fixed!

quicksketch:
Merged into 1.x for 1.24.0. Thanks @argiepiano!!

PR #4214, opened 2022-09-26
Issue #5800: Enable static caching on system.date config file.

### LINKED ISSUE #5800: Enable Static Caching on system.date.json
## Description of the bug

While performance testing date output in #5788, I found that the "system.date.json" file is parsed and read from disk _twice_ every time a date field is displayed. On a page with 1500 dates, this means reading `system.date.json` from disk 3000 times.

## Steps To Reproduce

To reproduce the behavior:
1. Add a date field to any content type.
2. Create several sample pieces of content.
3. Make a view that lists the date fields.
5. Either using a debugger or with a debug statement like `dpm()` in `ConfigFileStorage::read()` (line 1305 of config.inc):
```
if ($name === 'system.date') {
  dpm('Read system.date.json');
}
```
6. Check how many times `system.date.json` is read.

## Actual behavior

File is read _from disk_ twice per date field.

## Expected behavior

File should only be read once per page load.

## Additional information

quicksketch:
Fixes https://github.com/backdrop/backdrop-issues/issues/5800


core/modules/system/system.install
```diff
@@ -3343,6 +3343,22 @@ function system_update_1085() {
...
+ */
+function system_update_1086() {
+  $config = config('system.date');
+  $data = $config->get();
+  // Put the _config_static option at the top of the file.
+  unset($data['_config_name']);
+  $data = array_merge(array(
+    '_config_name' => 'system.date',
+    '_config_static' => TRUE,
+  ), $data);
```
klonos, line 3513: Wouldn't `array_splice()` be a cleaner way to do this here, instead of unsetting the first item and adding it back? Something like this perhaps? (untested):

```suggestion
  // Insert the "_config_static" option at the top of the file, right after
  // "_config_name".
  $static_cache = array('_config_static' => TRUE);
  array_splice($data, 1, 0, $static_cache);
```

klonos: ...hmm, `array_splice()` won't play that nice with associative arrays 🤔 ...see: https://stackoverflow.com/questions/1783089/array-splice-for-associative-arrays

So you'd need to do something like this instead (using `array_slice()` instead of `array_splice()`):

```suggestion
  // Insert the "_config_static" option at the top of the file, right after
  // "_config_name".
  $new_data = array_slice($data, 0, 1, TRUE) + array('_config_static' => TRUE) + array_slice($data, 1, NULL, TRUE);
```

...which seems equally messy/confusing.

quicksketch: Yeah, assigning items to the middle of an array is often tricky. I think leaving this as-is might be the easiest solution.

quicksketch:
This issue has been marked RTBC and I merged this into 1.x and 1.26.x.

PR #4215, opened 2022-09-27
Issue #5803: Only disable mod_php if it is on

### LINKED ISSUE #5803: GHA - our change does not work anymore after another change in the underlying system
## Description of the bug

Sort of a follow-up to #5790...

Beginning with 2022-09-18 we experienced a problem with the php module for 5.6 being turned on - which broke our Apache setup in the php 5.6 runner.

NOW there's: `ERROR: Module php5.6 does not exist!`, which means they turned it off again.

Need to figure out a more robust way - probably check first, if the module in on.

Another option is to just drop that step again - but then it might fail again, if the module is on (for whatever reason). Annoying.

indigoxela:
Fixes https://github.com/backdrop/backdrop-issues/issues/5803

### DIFF
--- a/.github/workflows/functional-tests.yml
+++ b/.github/workflows/functional-tests.yml
@@ -49,13 +49,9 @@ jobs:
           coverage: none
           tools: none
 
-      - name: Switch mpm if necessary
+      - name: Disable mod_php if necessary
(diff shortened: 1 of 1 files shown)


quicksketch:
Thank you! Merged into 1.x and 1.23.x.

PR #4216, opened 2022-09-27
Issue 5805. DatabaseCondition::__toString() should return string

### LINKED ISSUE #5805: Method DatabaseCondition::__toString() must return a string value
This is the corresponding fix for a [Drupal 7 issue](https://www.drupal.org/node/3004215). Currently `DatabaseCondition::__toString()` returns NULL.

Official PHP documentation for __toString() magic method:

__toString() method must return a string, as otherwise a fatal E_RECOVERABLE_ERROR level error is emitted.

argiepiano:
Fixes https://github.com/backdrop/backdrop-issues/issues/5805


quicksketch:

PR #4219, opened 2022-09-28
Issue #5839: Switch to testing profile in comment views tests

### LINKED ISSUE #5839: Use the testing profile for `CommentViewsHandlerArgumentUserUidTest`.
## Description of the task

Spin off from #5799, which gets fixed in a more generic way.

However, we still want to switch to the testing profile, which is the preferred one for Simpletest.

indigoxela:
Fixes https://github.com/backdrop/backdrop-issues/issues/5839

(changed issue)

### DIFF
--- a/core/modules/comment/tests/comment_views_handler_argument_user_uid.test
+++ b/core/modules/comment/tests/comment_views_handler_argument_user_uid.test
@@ -53,9 +53,8 @@ class CommentViewsHandlerArgumentUserUidTest extends ViewsSqlTest {
       'skip comment approval',
     ));
 
-    // Add two users, create a node with the user1 as author and another node
-    // with user2 as author.
-    // For the second node add a comment from user1.
+    // Add two users, create a node with user1 as author and another node with
+    // user2 as author. For the second node, add a comment from user1.
     $this->account = $this->backdropCreateUser();
     $this->account2 = $this->backdropCreateUser();
     $this->backdropLogin($this->account);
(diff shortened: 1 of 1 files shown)


klonos (approved):
LGTM 👍🏼

core/modules/comment/tests/comment_views_handler_argument_user_uid.test
```diff
@@ -33,15 +34,36 @@ class CommentViewsHandlerArgumentUserUidTest extends ViewsSqlTest {
...
+      'is_new' => TRUE,
+    ));
+    user_role_grant_permissions(BACKDROP_AUTHENTICATED_ROLE, array(
+      'access content',
+      'access comments',
+      'post comments',
+      'skip comment approval',
+    ));
+
+    // Add two users, create a node with the user1 as author and another node
```
avpaderno: It should be *with user1 as author*, instead of *with the user1 as author*, in the same way it is *with user2 as author*.

core/modules/comment/tests/comment_views_handler_argument_user_uid.test
```diff
@@ -33,15 +34,36 @@ class CommentViewsHandlerArgumentUserUidTest extends ViewsSqlTest {
...
+    // with user2 as author.
     // For the second node add a comment from user1.
```
klonos: @indigoxela if you agree, please merge this and let's wrap it up.

```suggestion
    // Add two users. Create a node with user1 as author and another node with
    // user2 as author. For the second node, add a comment from user1.
```

quicksketch:

PR #4221, opened 2022-10-02
Issue 5815. Fix SelectQuery::__clone()

### LINKED ISSUE #5815: Fix SelectQuery::_clone() (D7 issue: "Complex cloned query dependent on __toString() call")
This is the corresponding issue to D7's https://www.drupal.org/node/3000191

The clone operation for `SelectQuery` objects is done in incomplete fashion. The fix needs to make sure that subqueries stored in the `tables` property are also cloned properly, and also that the class parent's `__clone()` method is invoked.

argiepiano:
Fixes https://github.com/backdrop/backdrop-issues/issues/5815

### DIFF
--- a/core/includes/database/select.inc
+++ b/core/includes/database/select.inc
@@ -1829,6 +1829,8 @@ class SelectQuery extends Query implements SelectQueryInterface {
   }
 
   public function __clone() {
+    parent::__clone();
+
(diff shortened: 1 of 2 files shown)


quicksketch:
Merged into 1.x and 1.23.x. Thanks @argiepiano for porting this!

PR #4223, opened 2022-10-02
Issue 5816. Fix DatabaseConnection_mysql::createDatabase

### LINKED ISSUE #5816: Creation of database during installation fails
## Description of the bug

When you install Backdrop without previously creating a database, Backdrop tries creating it for you in `DatabaseConnection_mysql::createDatabase()`. A bug in this method results in a fatal error:

`Error: Call to undefined method DatabaseConnection_mysql::exec() in DatabaseConnection_mysql->createDatabase() (line 217 of /Users/XXX/Sites/localhost/panettone/core/includes/database/mysql/database.inc).`

## Steps To Reproduce
1. Do a clean install WITHOUT manually creating a database previously 
2. Enter the name of the (new) database, database username and database password in the second step 
3. See the error appear

## Actual behavior
Fatal error when trying to create the database.

## Expected behavior
Backdrop should create the database.

## Additional information
Backdrop 1.23.0
PHP  7.4.21
MySQL, MariaDB, or equivalent version 5.7.34

This was first reported in the [Forum](https://forum.backdropcms.org/forum/call-undefined-method-databaseconnection)

The `DatabaseConnection_mysql::createDatabase()` method contains calls to the method `exec()` which doesn't exist. I believe this should be `$this->pdo-exec()`. Changing this still produces a non-fatal Backdrop error message  about a test table not having been created, but the database is actually created just fine, and the installation process can be continued by submitting the form again.

argiepiano:
Fixes https://github.com/backdrop/backdrop-issues/issues/5816

### DIFF
--- a/core/includes/database/mysql/database.inc
+++ b/core/includes/database/mysql/database.inc
@@ -214,8 +214,8 @@ class DatabaseConnection_mysql extends DatabaseConnection {
   public function createDatabase($database) {
     try {
       // Create the database and set it as active.
-      $this->exec("CREATE DATABASE IF NOT EXISTS $database");
-      $this->exec("USE $database");
+      $this->pdo->exec("CREATE DATABASE IF NOT EXISTS $database");
+      $this->pdo->exec("USE $database");
     }
     catch (\Exception $e) {
       throw new DatabaseNotFoundException($e->getMessage());


klonos (approved):
LGTM 👍🏼

quicksketch:

PR #4224, opened 2022-10-03
Issue 5817. Re-run connect after creating new database

### LINKED ISSUE #5817: Creation of table `backdrop_install_test` fails upon installation without a pre-created database
## Description of the bug

This bug only appears after a fatal error reported in #5816 is fixed.

When installing Backdrop without having manually created a database prior to installation, the process first throws a `backdrop_set_message()` error. This happens despite the fact that the username and password ARE correct.

```
Failed to CREATE a test table on your database server with the command CREATE TABLE {backdrop_install_test} (id int NULL). The server reports the following message: SQLSTATE[3D000]: Invalid catalog name: 1046 No database selected.
```
Screen Shot 2022-10-02 at 4 56 01 PM

The installation proceeds normally if you submit the Database configuration form one more time.

## Steps To Reproduce
1. Patch Backdrop using #5816 
2. Install Backdrop on a completely clean install (do not edit settings.php) without manually creating a database before the installation
3. On the Database configuration screen enter the name of the NEW database, databse username and password
4. See error appear

## Actual behavior
Backdrop shows a soft error. Database is created normally but the test table is not.
 
## Expected behavior
Error should not happen. The test table should be created normally.

## Additional information

Backdrop 1.23.0
PHP 7.4.21
MySQL, MariaDB, or equivalent version 5.7.34

argiepiano:
Fixes https://github.com/backdrop/backdrop-issues/issues/5817


quicksketch:

PR #4225, opened 2022-10-04
Issue #5809: Make sure entity is loaded in checkAccess.

### LINKED ISSUE #5809: Fixed: Layouts and Taxonomy Term view context conflict.
## Description of the bug

When using the core-provided (but disabled by default) Taxonomy Term view, conflicts can arise with Layouts.

## Steps To Reproduce

To reproduce the behavior:
1. Enable the "Taxonomy Term" view
2. Create a tag on a Post
3. Create a custom Layout with the `taxonomy/term/%` context and provide a visibility condition of `Taxonomy vocabulary = Tags`
4. Visit the term page for the tag you created in step 2.

## Actual behavior

> Error
> Call to a member function bundle() on string 

## Expected behavior

The new layout should be used without a problem when viewing terms from the chosen vocabulary.

## Additional information

In `core/modules/layout/plugins/access/entity_bundle_layout_access.inc`, modifying `checkAccess()` as follows is a quick-and-dirty fix. It may not be the best long-term fix (hopefully discussion here can help identify the best long-term fix).

Change:

```
    if (!$this->settings['bundles']) {
      return FALSE;
    };
```

To:
```
    if (!$this->settings['bundles']) {
      return FALSE;
    }
    else if (empty($entity->bundle)) {
      return TRUE;
    }
```

laryn:
Fixes https://github.com/backdrop/backdrop-issues/issues/5809

### DIFF
--- a/core/modules/layout/plugins/access/entity_bundle_layout_access.inc
+++ b/core/modules/layout/plugins/access/entity_bundle_layout_access.inc
@@ -70,10 +70,15 @@ class EntityBundleLayoutAccess extends LayoutAccess {
   function checkAccess() {
     if (!$this->settings['bundles']) {
       return FALSE;
-    };
+    }
 
     /* @var Entity $entity */
     $entity = $this->contexts[$this->entity_type]->data;
+    // Make sure we're testing an entity here. Views could override system
+    // paths and cause difficulties. See issue #5809.
+    if (!is_object($entity)) {
+      $entity = entity_load($this->entity_type, $entity);
+    }
     $bundle = $entity->bundle();
     if (!in_array($bundle, $this->settings['bundles'])) {
       return FALSE;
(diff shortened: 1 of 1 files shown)


core/modules/layout/plugins/access/entity_bundle_layout_access.inc
```diff
@@ -70,10 +70,15 @@ class EntityBundleLayoutAccess extends LayoutAccess {
   function checkAccess() {
     if (!$this->settings['bundles']) {
       return FALSE;
-    };
+    }
 
     /* @var Entity $entity */
     $entity = $this->contexts[$this->entity_type]->data;
+    // Make sure we're testing an entity here. Views could override system
+    // paths and cause difficulties. See issue #5809.
```
klonos: When viewing this via GitHub diff online, the hash-tagged issue number is linkified - not the same when someone might be viewing the inline comments with a text editor. May I suggest that we actually use the link to the issue here?:

```suggestion
    // paths and cause difficulties.
    // See https://github.com/backdrop/backdrop-issues/issues/5809
```

quicksketch:
Thanks @laryn! Merged into 1.x and 1.23.x.

I made a typo in the commit message and re-pushed as 0fd1554ef.

PR #4226, opened 2022-10-05
Issue #5818: Preserve utf8mb4 value when upgrading from Drupal.

### LINKED ISSUE #5818: Upgrade from Drupal 7 should respect UTF8-MB4 variable
## Description of the bug

Backdrop includes a built-in migration tool to convert old databases to use UTF8-MB4 (to support emojis and Asian character sets). This tool is based on the Drupal 7 project: https://www.drupal.org/project/utf8mb4_convert

That project when run on a D7 site, sets a variable named `drupal_all_databases_are_utf8mb4`. We should respect that variable on upgrades from Drupal 7 and set our matching `state` variable to match: `database_utf8mb4_active`.

## Steps To Reproduce

To reproduce the behavior:
1. Install a Drupal 7 site with UTF8-MB4 disabled by default.
2. Run the drush command to convert tables to UTF8-MB4: https://www.drupal.org/project/utf8mb4_convert

OR

1. Install a Drupal 7 site with UTF8-MB4 installed by default.
2. No need to run the drush command.
3. Point Backdrop at the D7 site.
4. Run core/update.php

## Actual behavior

5. The Backdrop site after upgrading will show a notice that the UTF8-MB4 upgrade needs to be performed, even if all tables are already converted.

## Expected behavior

5. Backdrop should not throw a status report warning that the UTF8-MB4 upgrade needs to be run.

## Additional information

quicksketch:
Fixes https://github.com/backdrop/backdrop-issues/issues/5818

### DIFF
--- a/core/modules/system/system.install
+++ b/core/modules/system/system.install
@@ -3344,6 +3344,23 @@ function system_update_1085() {
 }
 
 /**
+ * Preserve 4-byte UTF-8 database update value if set.
+ */
+function system_update_1086() {
+  $connection = Database::getConnection();
+  $utf8mb4_active = $connection->utf8mb4IsActive();
+  $utf8mb4_supported = $connection->utf8mb4IsSupported();
+  // This variable is set by Drupal 7 on new installations that support UTF8-MB4
+  // and when the update is performed by the "utf8mb4_convert" project.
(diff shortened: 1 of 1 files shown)


quicksketch:
Merged into 1.x and 1.23.x.

PR #4227, opened 2022-10-06
Issue #5791: Provide a list of enabled modules to help with debugging

### LINKED ISSUE #5791: Provide an easy way to get ready-to copy/paste debugging information (core/php/db/web server versions, enabled themes/modules etc.)
## Description of the need

We frequently ask folks who are having difficulties with Backdrop to share a list of module they are using. I was asked for this today. 

In fact, this is not an easy thing to do and really it should be. Yes, it's easy to see all enabled modules on the admin/modules/list page. But, it's not easy to share that list in a forum post or Github issue. Not nearly as easy as it should be. 

I would like a text only or bulleted list of my enabled modules that I can easily export or share in a forum post. I would assume this is pretty easy to provide.. 

I looked for an existing feature request but could not find one. I was a bit surprised. Maybe it already exists and I haven't found it. Or maybe there is an easy way to do this already that I'm not aware of.

## Proposed solution

I'm going to suggest a report under the admin/reports menu that provides a list of all modules that can be filtered in any of the following ways:

1) All modules
2) All enabled modules
3) All enabled contrib modules
4) All enabled core modules

I think this should be available in the CORE UI because often the folks who need to share this information with someone do NOT have command line access to drush or Bee and may not be able to add a contributed module themselves.

For Drupal 7 there is a Drush command for this, I'm not sure if it works on Backdrop. But, even if it does, many of the users that need this feature are not going have access to a Drush or Bee based solution. I think this should be possible in the UI. 

It looks like this Drupal 7 module might do the trick, but I have not tested it. 
https://www.drupal.org/project/enabled_modules

Based upon discussion in this issue, I've released a recipe that generates a view for this:
https://backdropcms.org/project/enabled_modules_report_recipe

## Additional information

Related idea - list of "unused" modules - https://github.com/backdrop/backdrop-issues/issues/5020
Contrib proof of concept: https://github.com/backdrop-contrib/unused

Backdrop CMS now includes a copy and paste debug report that can be used when reporting errors or seeking support in forums or chat. The new debug report includes server info, database version, jQuery info, a list of enabled modules and themes, and more. We expect that this report will make it much easier for site owners to get answers to technical questions, by making it easier for them to share the specifics of their site installation.

indigoxela:
Fixes https://github.com/backdrop/backdrop-issues/issues/5791

### DIFF
--- a/core/modules/system/system.install
+++ b/core/modules/system/system.install
@@ -3344,6 +3344,260 @@ function system_update_1085() {
 }
 
 /**
+ * Activate the views admin listing of enabled modules.
+ */
+function system_update_1086() {
+  $view_config = array(
+    'name' => 'list_modules',
+    'description' => 'Administrative listing of enabled modules.',
+    'module' => 'system',
+    'storage' => 4,
+    'tag' => 'default',
+    'disabled' => FALSE,
+    'base_table' => 'system',
+    'human_name' => 'List of modules',
+    'core' => '1.24.0',
+    'display' => array(
+      'default' => array(
+        'display_title' => 'Master',
+        'display_plugin' => 'default',
+        'display_options' => array(
+          'query' => array(
+            'type' => 'views_query',
+            'options' => array(),
+          ),
+          'access' => array(
+            'type' => 'perm',
+            'perm' => 'administer modules',
+          ),
+          'cache' => array(
+            'type' => 'none',
+          ),
+          'exposed_form' => array(
+            'type' => 'basic',
+          ),
+          'pager' => array(
+            'type' => 'none',
+            'options' => array(
+              'items_per_page' => '',
+            ),
+          ),
+          'style_plugin' => 'list',
+          'row_plugin' => 'fields',
+          'fields' => array(
+            'name' => array(
+              'id' => 'name',
+              'table' => 'system',
+              'field' => 'name',
+              'relationship' => 'none',
+              'group_type' => 'group',
+              'ui_name' => '',
+              'label' => '',
+              'exclude' => 0,
+              'alter' => array(
+                'alter_text' => 0,
+                'text' => '',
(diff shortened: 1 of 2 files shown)


core/modules/system/system.install
```diff
@@ -3343,6 +3343,260 @@ function system_update_1085() {
   }
 }
 
+/**
+ * Activate the views admin listing of enabled modules.
+ */
+function system_update_1086() {
+  $view_config = array(
```
argiepiano, line 3350: Rather than duplicating all of this, I suggest you use `  config_install_default_config('system', 'views.view.list_modules');`

This function will essentially do the same thing you do here (check of the view exists, and if so, avoid installing it)

indigoxela: There's a function for that!! :tada: 

Hm, nobody seems to have used that in install files so far - is that because it's not safe to use it in install, or because all the others also weren't aware of that function? :thinking:

docwilmot: https://github.com/backdrop/backdrop-issues/issues/352 maybe?

stpaultim:
I tried to generate a fresh sandbox to review this issue, but I believe that a merge conflict is blocking progress. @indigoxela do you have time to reroll this patch?

I'd like to take a look at it.

argiepiano:
I'm not sure this was intended to be merged. There is additional discussion in the issue after this PR was submitted. Also https://github.com/backdrop/backdrop-issues/issues/5820 is related - that PR provides Views fields that can expand what this PR was trying to do.

indigoxela:
I don't think, this issue has in any way consensus. It's probably not worth to put further effort in it, as it doesn't provide what has been requested in the meantime.

indigoxela:
Closing in favor of ... any other approach. :smirk_cat:

PR #4228, opened 2022-10-06
Issue #2054: Preserve image aspect ratio in editor dialog

### LINKED ISSUE #2054: [UX] Add an option to "lock" aspect ratio when resizing images in the editor dialog
When resizing images in CKEditor, 99% of the time the user will want the aspect ratio to remain the same. We should make an option to do this (and enable it by default) so that people have a better experience resizing images in Backdrop.

Note that resizing within the editor iframe already maintains aspect ratio, this is just about the aspect ratio within the image _dialog_.

indigoxela:
Fixes https://github.com/backdrop/backdrop-issues/issues/2054


core/modules/filter/js/filter.js
```diff
@@ -333,6 +329,87 @@ Backdrop.behaviors.editorImageLibrary = {
...
+
+    // Lock image aspect ratio.
+    var naturalDimensions = {
+      width: null,
+      height: null
+    };
+    // When editing a previously added image or upload a new one.
+    var existingFile = $('.filter-format-editor-image-form .form-managed-file a').attr('href');
+    if (typeof existingFile !== 'undefined') {
+      let img = new Image();
```
argiepiano, line 309: ~~This seems a bit hacky, but I can't think of another way, other than perhaps attaching a listener to the actual img element in the document... but because this img is added by ckeditor, it may be difficult to do.~~

Disregard this comment. I misunderstood the goal of this.

core/modules/filter/js/filter.js
```diff
@@ -333,6 +329,87 @@ Backdrop.behaviors.editorImageLibrary = {
...
+      img.onload = function() {
+        naturalDimensions.width = this.width;
+        naturalDimensions.height = this.height;
+        if (!$('.filter-format-editor-image-form [name="attributes[width]').val().length) {
```
argiepiano: Why is length needed here? Wouldn't checking for val() be enough?

core/modules/filter/js/filter.js
```diff
@@ -333,6 +329,87 @@ Backdrop.behaviors.editorImageLibrary = {
...
+      img.onerror = function() {
+        naturalDimensions.width = null;
+        naturalDimensions.height = null;
+      }
+      img.src = existingFile;
+    }
+    else if ($('.filter-format-editor-image-form [name="attributes[width]').length) {
+      // After an image has been removed via button, and the managed form item
+      // reloads, reset width and height.
+      if ($('.filter-format-editor-image-form [name="attributes[width]').val().length) {
```
argiepiano: There seems to be some unnecessary logic here. The parameter of the if statement here can be moved above to the else if, and remove this if.

And once again, is length needed?

core/modules/filter/js/filter.js
```diff
@@ -333,6 +329,87 @@ Backdrop.behaviors.editorImageLibrary = {
...
+      }
+      img.src = this.value;
+    });
+
+  },
+  /**
+   * Helper function to empty the width and height form items.
+   */
+  imageDimensionsEmpty: function() {
+    $('.filter-format-editor-image-form [name="attributes[width]').val('');
```
argiepiano: There seems to be a " and ] missing in the jquery selector after [width]

core/modules/filter/js/filter.js
```diff
@@ -333,6 +329,87 @@ Backdrop.behaviors.editorImageLibrary = {
...
+    $('.filter-format-editor-image-form [name="attributes[height]').val('');
```
argiepiano: There seems to be a " and ] missing in the jquery selector after [height]

core/modules/filter/js/filter.js
```diff
@@ -333,6 +329,87 @@ Backdrop.behaviors.editorImageLibrary = {
...
+  },
+  /**
+   * Helper funtion to set width and height values.
+   */
+  imageDimensionsSet: function(values) {
+    $('.filter-format-editor-image-form [name="attributes[width]').val(values.width);
```
argiepiano: There seems to be a " and ] missing in the jquery selector after [width]

core/modules/filter/js/filter.js
```diff
@@ -333,6 +329,87 @@ Backdrop.behaviors.editorImageLibrary = {
...
+    $('.filter-format-editor-image-form [name="attributes[height]').val(values.height);
```
argiepiano: There seems to be a " and ] missing in the jquery selector after [height]

core/modules/filter/js/filter.js
```diff
@@ -333,6 +329,87 @@ Backdrop.behaviors.editorImageLibrary = {
...
+  /**
+   * Remove previous event listeners, add new ones with current dimensions.
+   *
+   * Keep width and height input values in sync based on the natural image
+   * dimensions.
+   */
+  syncAspectRatio: function(naturalDimensions) {
+    $('.filter-format-editor-image-form [name="attributes[width]"]').off('change').on('change', function() {
+      var newHeight = Math.round(this.value / naturalDimensions.width * naturalDimensions.height);
+      $('.filter-format-editor-image-form [name="attributes[height]').val(newHeight);
```
argiepiano: missing " and ]?

core/modules/filter/js/filter.js
```diff
@@ -333,6 +329,87 @@ Backdrop.behaviors.editorImageLibrary = {
...
+    });
+    $('.filter-format-editor-image-form [name="attributes[height]"]').off('change').on('change', function() {
+      var newWidth = Math.round(this.value / naturalDimensions.height * naturalDimensions.width);
+      $('.filter-format-editor-image-form [name="attributes[width]').val(newWidth);
```
argiepiano: missing " and ]?

core/modules/filter/js/filter.js
```diff
@@ -333,6 +329,87 @@ Backdrop.behaviors.editorImageLibrary = {
...
+      // reloads, reset width and height.
+      if ($('.filter-format-editor-image-form [name="attributes[width]').val().length) {
+        naturalDimensions.width = null;
+        naturalDimensions.height = null;
+        Backdrop.behaviors.editorImageLibrary.imageDimensionsEmpty();
+      }
+    }
+
+    // Selecting an image from library updates width and height values.
+    $('.filter-format-editor-image-form [name="attributes[src]"]').once().on('change', function() {
```
argiepiano: Oh, one more thing. I don't have a ton of experience with `once()`, but I believe it's better to provide an id in `once()`, otherwise the default id `once` will be used. Not sure about this, but it may create problems if there are more than one `once()` using the default id in the document. This id can be anything (it's not a DOM element id, but rather an internal id to keep track of listeners that have been already attached). 

All uses of `once` in Backdrop provide an id

core/modules/filter/js/filter.js
indigoxela: An extra check that this has a length.

core/modules/filter/js/filter.js
indigoxela: Right... :+1:

core/modules/filter/js/filter.js
indigoxela: Good idea. :+1:

core/modules/filter/js/filter.js
indigoxela: Too much copy/paste :wink:
Funny that it still worked.

indigoxela:
There's both, lack of interest and difficulties to find a good compromise, so I'm moving this PR out of the way. :wink:

PR #4229, opened 2022-10-06
Issue 5820. Provide views field handler for module info settings

### LINKED ISSUE #5820: Provide views field handler to access project properties from the `info` column in the `system` table
## Description of the need

Backdrop currently provides Views definitions to display module and theme information stored in the `system` database. Currently, it's possible to display theme/module `filename`, `name`, `schema_version`, `status` and `type`. 

However, there is no way to display any of the information stored in the `info` column, which includes version, dependencies, backdrop version, tags, etc.

This is a request to add such a handler. 

## Proposed solution

Write a special field  handler to display those values. Since the `info` column is a serialized array, there needs to be a special option definition and option form within the handler to select the specific setting to be displayed. 

None.

Backdrop now includes a Views field handler to display setting information stored in the module and theme info files.

argiepiano:
Fixes https://github.com/backdrop/backdrop-issues/issues/5820


core/modules/system/views/system.views.inc
```diff
@@ -120,6 +120,17 @@ function system_views_data() {
     ),
   );
 
+  // - info
+  $data['system']['system_info'] = array(
+    'title' => t('Module/Theme/Theme engine settings'),
+    'help' => t("Selected settings from the info file."),
```
indigoxela: Nothing substantial, but single quotation marks are sufficient here.

core/modules/system/views/views_handler_field_system_info.inc
```diff
@@ -0,0 +1,105 @@
...
+    $form['label']['#states'] = array(
+      'visible' => array(
+        ':input[name="options[custom_label]"]' => array('value' => 'custom')
+      ),
+    );
+    $form['label']['#title'] = t('Custom label');
+
+    $form['info_setting'] = array(
+      '#type' => 'select',
+      '#title' => 'Field',
```
indigoxela: This form item title should be translatable (`t()`).

core/modules/system/views/system.views.inc
```diff
@@ -120,6 +120,17 @@ function system_views_data() {
     ),
   );
 
+  // - info
+  $data['system']['system_info'] = array(
```
klonos: Why not just `'info'`?

```suggestion
  $data['system']['info'] = array(
```

core/modules/system/views/views_handler_field_system_info.inc
```diff
@@ -0,0 +1,105 @@
+<?php
+/**
+ * @file
+ * Definition of views_handler_field_system_info.
+ */
+
+/**
+ * A handler to display module and theme settings stored in the info column.
```
klonos: 1. What we are retrieving/displaying here is not settings - it is information/details/properties.
2. In Backdrop-world, we also have a 3rd type of project: layout templates - not just modules and themes.
3. The words "info column" can be perceived as vague/ambiguous if the word "database" is not mentioned as well. Since we mention `the info column of the system table.` in the next line of this docblock, we can skip details here. How about this?:

```suggestion
 * A handler to display module, theme, and layout template information stored in
 * the database.
```

klonos: ...alternatively this (also adheres to the standard of the first description of functions/classes being a short, single line):

```suggestion
 * A handler to display project information stored in the database.
```

core/modules/system/views/views_handler_field_system_info.inc
```diff
@@ -0,0 +1,105 @@
...
+ *
+ * Provides access to several settings stored as a serialized array in the info
+ * column of the system table.
```
klonos:
```suggestion
 * Provides access to several information stored as a serialized array in the
 * info column of the system table.
```

core/modules/system/views/views_handler_field_system_info.inc
```diff
@@ -0,0 +1,105 @@
...
+
+  /**
+   * {@inheritdoc}
+   */
+  function options_form(&$form, &$form_state) {
+    parent::options_form($form, $form_state);
+
+    $form['custom_label']['#type'] = 'radios';
+    $form['custom_label']['#options'] = array(
+      'default' => t('Use setting label'),
```
klonos: How about something like this? (this is not a setting - it is information)

```suggestion
      'default' => t('Same as in the .info file'),
```

core/modules/system/views/views_handler_field_system_info.inc
```diff
@@ -0,0 +1,105 @@
...
+      case 'tags':
+      case 'dependencies':
+        return implode(', ', $info_array[$info_setting]);
+      default:
+        return $info_array[$info_setting];
+    } 
+  }
+
+  /**
+   * * Get this field's label based on the selected setting.
```
klonos:
```suggestion
   * Get this field's label based on the selected information.
```

core/modules/system/views/views_handler_field_system_info.inc
```diff
@@ -0,0 +1,105 @@
...
+      return $label;
+    }
+    elseif ($this->options['custom_label'] === 'custom' && isset($this->options['label'])) {
+      return $this->options['label'];
+    }
+    return '';
+  }
+
+  /**
+   * Helper function to return the option array.
```
klonos:
```suggestion
   * Helper function to return the options array.
```

core/modules/system/views/views_handler_field_system_info.inc
```diff
@@ -0,0 +1,105 @@
...
+    return array(
+      'name' => t('Name'),
+      'description' => t('Description'),
+      'package' => t('Package'),
+      'configure' => t('Configuration path'),
+      'tags' => t('Tags'),
+      'version' => t('Version'),
+      'backdrop' => t('Backdrop version'),
+      'dependencies' => t('Dependencies'),
+    );
```
klonos: Initially I thought of suggesting that we should use `ucfirst()` here, but then realized that by using specific strings in `t()`, we gain the following:
- these can be translatable
- we can customize the string that comes from the .info file (like in the case of `'configure' => t('Configuration path'),`)
...so good call here overall👍🏼

Having said that, can we please account for any other information that may be added in the .info files in the future? (by issues like https://github.com/backdrop-ops/backdropcms.org/issues/487 and/or https://github.com/backdrop/backdrop-issues/issues/5367). So the array with these default core-provided properties is perfectly fine, but can we account for other properties by converting their names to labels in this array via `ucfirst()`?

core/modules/system/views/views_handler_field_system_info.inc
klonos: ...or:
```suggestion
   * Get this field's label based on the selected property.
```

core/modules/system/views/views_handler_field_system_info.inc
klonos: ...or:
```suggestion
 * Provides access to several properties stored as a serialized array in the
 * info column of the system table.
```

core/modules/system/views/views_handler_field_system_info.inc
klonos: ...or:
```suggestion
 * A handler to display project properties stored in the database.
```

core/modules/system/views/system.views.inc
```diff
@@ -120,6 +120,17 @@ function system_views_data() {
...
+    'title' => t('Module/Theme/Theme engine settings'),
```
klonos:
```suggestion
    'title' => t('Project/Theme engine properties'),
```

core/modules/system/views/system.views.inc
klonos:
```suggestion
    'help' => t('Selected properties from the project .info file.'),
```

core/modules/system/views/views_handler_field_system_info.inc
klonos: Also, can we add `required` to the list of properties:
```suggestion
    return array(
      'name' => t('Name'),
      'description' => t('Description'),
      'package' => t('Package'),
      'configure' => t('Configuration path'),
      'tags' => t('Tags'),
      'version' => t('Version'),
      'backdrop' => t('Backdrop version'),
      'dependencies' => t('Dependencies'),
      'required' => t('Required'),
    );
```

core/modules/system/views/system.views.inc
argiepiano: Fixed

core/modules/system/views/views_handler_field_system_info.inc
argiepiano: Fixed

core/modules/system/views/system.views.inc
argiepiano: Changed

core/modules/system/views/views_handler_field_system_info.inc
argiepiano: I've changed the documentation a bit following your suggestions. Layout templates are not stored in the `system` table, and therefore are beyond the scope of this PR.

core/modules/system/views/views_handler_field_system_info.inc
argiepiano: Changed the default with a clearer wording, though not exactly what you suggested.

core/modules/system/views/views_handler_field_system_info.inc
argiepiano: Fixed

core/modules/system/views/views_handler_field_system_info.inc
argiepiano: >   So the array with these default core-provided properties is perfectly fine, but can we account for other properties by converting their names to labels in this array via ucfirst()?

With the way these properties are declared, we can't. There is no central repository or definition of all possible info properties. You would have to scan all the info files for themes and modules to find out all the possible properties, as some properties are not required (e.g. tags. Some modules info files have tags, others don't). 

Instead, I've added an "Other" input, that allows you to type the name of any additional property not included in the array. This way you can possibly type new properties. And if necessary, this handler can be modified in the future to expand this array.

I've added several other improvements for flat array properties like tags and dependencies. You can now specify the separator, or order/unordered list. Also dependencies give you the option of outputting machine names or human labels.
 
But, since some properties can be multilevel arrays (like `spreadsheets`) I've accounted for that by outputting a `print_r` output. It doesn't look pretty, but it's better than just showing `Array` in the output. Things get just too complicated with multilevel arrays, or with arrays that use keys.

core/modules/system/views/system.views.inc
argiepiano: Since other system handlers use `Module/Theme/Theme engine` in their names, I think we should either modify all of them, or none. I've chosen to keep the current name. To change the other ones we would need to open another issue.

core/modules/system/views/system.views.inc
```diff
@@ -120,6 +120,16 @@ function system_views_data() {
     ),
   );
 
+  // - info
+  $data['system']['info'] = array(
+    'title' => t('Module/Theme/Theme engine properties'),
```
jenlampton: I agree that this should be called "Project properties". People are not able to use alternate theme engines in backdrop, so including that label here could be confusing for people. I don't think the omission of some project types is a problem, and would prefer this be consistent with other uses of "Project".

argiepiano: @jenlampton I can see that this change would make things clearer. Please notice that there are a bunch of pre-existing views data in `system.views.inc` that already used "Module/Theme/Theme engine..." wording. I can make the change you request here, but do you also want me to change the rest of the data there to match?

core/modules/system/views/system.views.inc
```diff
@@ -19,15 +19,15 @@ function system_views_data() {
...
     'help' => t('Modules/Themes/Theme engines in your codebase.'),
   );
 
   // fields
   // - filename
   $data['system']['filename'] = array(
-    'title' => t('Module/Theme/Theme engine filename'),
-    'help' => t('The path of the primary file for this item, relative to the Backdrop root; e.g. modules/node/node.module.'),
+    'title' => t('Project filename'),
+    'help' => t('The path of the primary file for this module/theme/theme engine, relative to the Backdrop root; e.g. modules/node/node.module.'),
```
quicksketch, line 30: Perfect! Much easier to read and find.

quicksketch:
Merged into 1.x for 1.24.0. Thanks @argiepiano!

PR #4232, opened 2022-10-14
Issue #5832: Prevent passing null to preg_replace()

### LINKED ISSUE #5832: Deprecated function: preg_replace(): Passing null to parameter #3
## Description of the bug

Function system_modules() assumes that all modules that are a dependency of another module have a version set, which is not always the case.
https://github.com/backdrop/backdrop/blob/1.x/core/modules/system/system.admin.inc#L630

This leads to php deprecation notices on php 8+

## Steps To Reproduce

1. Use php 8.1
1. Enable error display
1. Install a module without version info
1. Install another module, which depends on the module without version set
1. Go to /admin/modules/list

## Actual behavior

Deprecated function: preg_replace(): Passing null to parameter #3 ($subject) of type array|string is deprecated in system_modules() (630 ...core/modules/system/system.admin.inc).

## Expected behavior

No deprecation warning.

## Additional information

- Backdrop CMS version: 1.23

indigoxela:
Fixes https://github.com/backdrop/backdrop-issues/issues/5832

### DIFF
--- a/core/modules/system/system.admin.inc
+++ b/core/modules/system/system.admin.inc
@@ -627,8 +627,12 @@ function system_modules($form, $form_state = array()) {
         $requires_name = $files[$requires]->info['name'];
 
         // Disable this module if it is incompatible with the dependency's version.
-        $version_string = preg_replace('/^' . preg_quote(BACKDROP_CORE_COMPATIBILITY, '/') . '-/', '', $files[$requires]->info['version']);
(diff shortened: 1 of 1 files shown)


klonos (approved):
LGTM 👍🏼

quicksketch:
Thanks! Merged into 1.x and 1.23.x.

PR #4235, opened 2022-10-17
Issue #5837: [DX] Allow disabling individual options in select elements

### LINKED ISSUE #5837: [DX] Form API: Allow individual options in select elements to be set to disabled
This was inspired by the discussion in https://backdrop.zulipchat.com/#narrow/stream/218635-Backdrop/topic/.24from.20and.20disable.20a.20select.20list.20item

Consider this form element:
```php
  $form['my_checkboxes'] = array(
    '#type' => 'checkboxes',
    '#title' => t('my Checkboxes'),
    '#options' => array(
      1 => 'one',
      2 => 'two',
      3 => 'three',
      4 => 'four',
    ),
    '#description' => t('This is the description for the entire set of checkboxes'),
    '#default_value' => array(1, 3),
  );
```

One can set descriptions for each individual checkbox, and also disable only one/some of these checkboxes:

```php
  $form['my_checkboxes'][1]['#description'] = t('This is the description for the 1st checkbox');
  $form['my_checkboxes'][3]['#description'] = t('This is the description for the 3rd checkbox');
  $form['my_checkboxes'][2]['#disabled'] = TRUE;
```
The above would result in this:
image

---

Although individual descriptions do not make sense for select elements, there are perfectly valid cases where you need to:
- make sure that one/some of the options are disabled/locked (user cannot select them, but they are still shown)
- make sure that one/some of the options are selected + at the same time disabled/locked (so that you can force one or more options that must be selected in addition to those that the user can select).

However, if you consider this select element:
```php
  $form['my_select'] = array(
    '#type' => 'select',
    '#title' => t('my Select'),
    '#options' => array(
      1 => 'one',
      2 => 'two',
      3 => 'three',
      4 => 'four',
    ),
    '#default_value' => array(1, 3),
  );
```
...or this multi-select element:
```php
  $form['my_select'] = array(
    '#type' => 'select',
    '#multiple' => TRUE,
    '#title' => t('my multi-Select'),
    '#options' => array(
      'group1' => array(
        1 => 'one',
        2 => 'two',
      ),
      'group2' => array(
        3 => 'three',
        4 => 'four',
      ),
    ),
    '#default_value' => array(1, 3),
  );
```

Then this has absolutely no effect:
```php
$form['my_select'][2]['#disabled'] = TRUE;
```

klonos:
Fixes https://github.com/backdrop/backdrop-issues/issues/5837

### DIFF
--- a/core/includes/form.inc
+++ b/core/includes/form.inc
@@ -2836,7 +2836,7 @@ function theme_select($variables) {
  *
  * @param array $element
  *   An associative array containing the following key-value pairs:
- *   - #multiple: Optional Boolean indicating if the user may select more than
+ *   - #multiple: (optional) Boolean indicating if the user may select more than
  *     one item.
  *   - #options: An associative array of options to render as HTML. Each array
  *     value can be a string, an array, or an object with an 'option' property:
@@ -2854,7 +2854,7 @@ function theme_select($variables) {
  *       object with an 'option' property, the key is ignored, the contents of
  *       the option property are interpreted as $element['#options'], and the
  *       resulting HTML is added to the output.
- *   - #value: Optional integer, string, or array representing which option(s)
+ *   - #value: (optional) Integer, string, or array representing which option(s)
  *     to pre-select when the list is first displayed. The integer or string
  *     must match the key of an option in the '#options' list. If '#multiple' is
  *     TRUE, this can be an array of integers or strings.
@@ -2876,6 +2876,7 @@ function form_select_options($element, $choices = NULL) {
   $value_is_array = $value_valid && is_array($element['#value']);
   $options = '';
   foreach ($choices as $key => $choice) {
+    $disabled = '';
     if (is_array($choice)) {
       $options .= '<optgroup label="' . check_plain($key) . '">';
       $options .= form_select_options($element, $choice);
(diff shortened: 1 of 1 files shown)


core/includes/form.inc
```diff
@@ -2876,6 +2876,7 @@ function form_select_options($element, $choices = NULL) {
   $value_is_array = $value_valid && is_array($element['#value']);
   $options = '';
   foreach ($choices as $key => $choice) {
+    $disabled = '';
```
argiepiano, line 2879: Since $disabled is only used within the scope of the `else{}` below, I suggest moving the initialization to within the `else{}`

core/includes/form.inc
```diff
@@ -2892,7 +2893,12 @@ function form_select_options($element, $choices = NULL) {
       else {
         $selected = '';
       }
-      $options .= '<option value="' . check_plain($key) . '"' . $selected . '>' . check_plain($choice) . '</option>';
+      if (isset($element[$key])) {
+        if (isset($element[$key]['#disabled']) && $element[$key]['#disabled'] == TRUE) {
```
argiepiano, line 2897: These two nested `if` statements can be combined and simplified into:

```php
      if (!empty($element[$key]['#disabled'])) {
        $disabled = ' disabled';
      }
```

In PHP, `!empty(TRUE)` always returns true, and doing `!empty` for an array element that is not set or doesn't exist always returns false.  

From the [PHP manual](https://www.php.net/manual/en/function.empty.php):
<blockquote>
A variable is considered empty if it does not exist or if its value equals false. empty() does not generate a warning if the variable does not exist.
</blockquote>

PR #4236, opened 2022-10-19
2709: Replaces Home with Back to Site in the admin bar while on admin…

### LINKED ISSUE #2709: [UX] Admin bar: Consider replacing the home button with a "Back to site" button.

This is basically a request to have https://www.drupal.org/project/escape_admin in core instead of (or in addition to) the home button. From that project's description:

>This is a Drupal 7 backport of the functionality added to the revamped toolbar: [~#787896: Add a link so that administrators can return to their most recently visited non-admin page~](https://www.drupal.org/node/787896) which adds a 'Back to site' button when the user has clicked on an administration URL from a non-administration page. The 'Back to site' link will persist until the user returns to a non-administration page.

### Steps to reproduce

The UX- with the current situation is that we are assuming that the user wants to return to the home page after dealing with any admin/backend task, whereas the most likely thing is that they want to return back to the page where they started from.

For example: as a novice admin, go to any page of your site -> realize that you want comments enabled -> edit the page -> realize that comments are not enabled for the page content type -> go to Structure -> Content types -> Page -> make the changes -> Save. Now, if you click the Home button, you are taken to the home page, while you most likely needed to navigate back to the page you started from.

### Actual behavior

The "Home" button takes the user to their home page. This is fine, but they already have a "home" trail in the admin theme that they can use if they want to go to the home page.

### Expected behavior

A much better UX would be to have it so that when the user navigates to any admin page from the frontend, the path of the page they started from is stored in their browsing session. This path persists as they navigate through the admin interface, until they hit the "Back to site" button. What happens then is that they are taken to the page where they started from. As the user navigates through the frontend, their session is updated to store the last non-admin page they visit before switching to the admin UI.

If we think that changing the current behavior of the "Home" button would be disrupting the way current users are used for things to work, then we can either make this an option in `admin/config/administration/admin-bar` or add a separate "Back to site" button.

---

PR by @alexfinnarn : ~https://github.com/backdrop/backdrop/pull/2040~
PR by @keiserjb: https://github.com/backdrop/backdrop/pull/4236

keiserjb:
… screens.

Fixes [[UX] Admin bar: Consider replacing the home button with a "Back to site" button.](https://github.com/backdrop/backdrop-issues/issues/2709)


core/modules/admin_bar/js/admin_bar.js
```diff
@@ -500,6 +500,46 @@ Backdrop.adminBar.behaviors.search = function (context, settings, $adminBar) {
...
+    // Saves the last non-administrative page in the browser to be able to link back
+    // to it when browsing administrative pages. If there is a destination parameter
+    // there is not need to save the current path because the page is loaded within
+    // an existing "workflow".
+    if (!settings.admin_bar.current_path_is_admin && !/destination=/.test(window.location.search)) {
+      sessionStorage.setItem('escapeAdminPath', settings.admin_bar.current_path);
+    }
+
+
+
```
klonos: Don't need more than a single empty line.

core/modules/admin_bar/tests/admin_bar.test
```diff
@@ -490,4 +490,60 @@ class AdminBarCustomizedTestCase extends AdminBarWebTestCase {
...
+   * @broken
+   * @todo Figure out how to test for HTML in the admin bar. The debug output for
+   *   the failed tests doesn't seem to give me that information.
+   */
+  /*
+  function testEscapeAdmin() {
+
+    // Create a sample page.
+    $settings = array(
+      'title'     => 'Foo',
```
klonos:
```suggestion
      'title' => 'Foo',
```

core/modules/admin_bar/tests/admin_bar.test
```diff
@@ -490,4 +490,60 @@ class AdminBarCustomizedTestCase extends AdminBarWebTestCase {
...
+
+    // Set theme to something other than Stark.
+    config_set('system.core', 'theme_default', 'basis');
+    config_set('system.core', 'admin_theme', 'seven');
+
+    // Login user and go to created page.
+    $permissions = $this->basePermissions + array(
+        'access content overview',
+        'view the administration theme'
+      );
```
klonos: Unnecessary indentation here:

```suggestion
      'access content overview',
      'view the administration theme'
    );
```

core/modules/admin_bar/tests/admin_bar.test
```diff
@@ -490,4 +490,60 @@ class AdminBarCustomizedTestCase extends AdminBarWebTestCase {
...
     ))->fetchField();
   }
+  /**
+   * Test the "Back To Site" link functionality.
+   *
+   * @broken
+   * @todo Figure out how to test for HTML in the admin bar. The debug output for
+   *   the failed tests doesn't seem to give me that information.
+   */
+  /*
```
klonos: Remove this.

core/modules/admin_bar/tests/admin_bar.test
```diff
@@ -490,4 +490,60 @@ class AdminBarCustomizedTestCase extends AdminBarWebTestCase {
...
+
+    // Go to the front page and then back to the admin.
+    $front = config_get('system.core', 'site_frontpage');
+    $this->backdropGet($front);
+    $this->backdropGet('admin/content');
+
+    // See that the Home link is still there.
+    $this->assertRaw('<li class="admin-bar-icon expandable"><a href="/" class="">');
+  }
+  */
```
klonos: Remove this too.

core/modules/admin_bar/tests/admin_bar.test
argiepiano: @klonos, the whole test needs to be removed (not just the commenting line). This test always fails because the replacement of the icon with "Back to site" is done by Javascript on the client side. So, at the moment, it's impossible to test this functionality (with the tests we use in Backdrop). 

I think @keiserjb is aware of this.

core/modules/admin_bar/tests/admin_bar.test
argiepiano: All the comments and suggestions for the test are moot, as the whole test needs to be removed.

core/modules/admin_bar/js/admin_bar.js
```diff
@@ -500,6 +500,46 @@ Backdrop.adminBar.behaviors.search = function (context, settings, $adminBar) {
...
+    // We only want to change the first anchor tag in the admin bar icon sub-menu.
+    var $toolbarEscape = $('.admin-bar-icon a').first();
+
+    // If the current page is admin, then switch the path.
+    if ($toolbarEscape.length
+      && settings.admin_bar.current_path_is_admin
+      //&& settings.admin_bar.frontpage !== escapeAdminPath
```
argiepiano: This line, which you commented out, is needed. The goal of this line is to avoid replacing "Home" with "Back to site" when your original page was the home page of your site. The "Home" text and link should remain there (not be replaced with "Back to site", since it will take you to the home page anyway).

core/modules/admin_bar/admin_bar.module
```diff
@@ -244,6 +244,15 @@ function admin_bar_preprocess_page(&$variables) {
     $settings['position_fixed'] = $setting;
   }
 
+  // Send whether this is an admin page to JS for back to site functionality.
+  $settings['current_path_is_admin'] = FALSE;
+  $settings['current_path'] = current_path();
```
argiepiano: This line here causes the "Back to site" link to always point at the **path** of the original page, INSTEAD of the **alias** .

It would be desirable if this functionality used the alias rather than the path. For example if your original path was `example.com/my-page` you want the link to point at that, rather than at, say, `example.com/node/15`.

I will suggest code to do this later today.

argiepiano: This is the suggested code (replace lines 249 - 252). This will make the link point at the alias when available (rather than the real path).

```php
  $current_path = current_path();
  if (path_is_admin($current_path)) {
    $settings['current_path_is_admin'] = TRUE;
  }
  $alias = backdrop_get_path_alias($current_path, $language->langcode);
  $settings['current_path'] = $alias;
```

core/modules/admin_bar/js/admin_bar.js
keiserjb: Yes, it was like that originally and I took it out. It can go back.

core/modules/admin_bar/js/admin_bar.js
keiserjb: Lines removed.

core/modules/admin_bar/tests/admin_bar.test
keiserjb: Test removed.  It apparently did not work.

core/modules/admin_bar/admin_bar.module
keiserjb: Thanks for the suggestion.  I had noticed that but did not know how to fix.

core/modules/admin_bar/css/admin_bar.css
```diff
@@ -83,6 +83,16 @@
   padding-right: 28px;
 }
 
+/* Back to Site arrow. */
+
+#admin-bar .dropdown li.admin-bar-icon > a.escape {
+  background-image: url("../icons/chevron-disc-left.png");
```
olafgrabienski: The new icons (added below) don't look good on high resolution displays. To make them look better, the images should be 32 x 32px instead of 16px.

keiserjb: I've also wondered if it is better to place these in the images folder with everything else rather than a separate icons folder.  The icons came from the [Drupal module](https://www.drupal.org/project/escape_admin).

olafgrabienski: > if it is better to place these in the images folder with everything else rather than a separate icons folder.

Agreed! (The other images, or most of them, are also icons.)

core/modules/admin_bar/admin_bar.inc
```diff
@@ -779,11 +779,25 @@ function admin_bar_theme_settings() {
...
+    '#title' => t('Add Back to site link'),
+    '#default_value' => $config->get('back_to_site_link'),
+    '#description' => t('Changes the "Home" to a "Back to site" link that links to the last non-admin page you visited.'),
+    '#states' => array(
+          'visible' => array(
+            ':input[name="components[admin_bar.icon]"]' => array('checked' =>
+                                                                TRUE),
+          ),
+        ),
+  );*/
```
argiepiano: Please remove this and all other commented code.

core/modules/admin_bar/admin_bar.inc
```diff
@@ -802,6 +823,8 @@ function admin_bar_theme_settings() {
   );
 
   $form['components']['admin_bar.icon']['#description'] = t('A specialized menu that links to the home page and provides links to perform various other admin tasks (e.g.: clear caches, run cron, etc.).');
+  $form['components']['admin_bar.back_to_site_link']['#description'] = t
+  ('Changes the "Home" to a "Back to site" link that links to the last non-admin page you visited.');
```
argiepiano: I'm not sure if this is the best wording. Thoughts?  ("link that links" doesn't work well). And I realize I probably suggested this 😆

core/modules/admin_bar/admin_bar.info
```diff
@@ -9,3 +9,8 @@ version = BACKDROP_VERSION
 backdrop = 1.x
 
 configure = admin/config/administration/admin-bar
+
+; Added by Backdrop CMS packaging script on 2022-09-15
+project = backdrop
+version = 1.23.0
+timestamp = 1663295416
```
argiepiano: Remove this and all other packaging information.

core/modules/admin_bar/admin_bar.module
```diff
@@ -244,6 +244,27 @@ function admin_bar_preprocess_page(&$variables) {
...
+  if (path_is_admin($settings['current_path'])) {
+    $settings['current_path_is_admin'] = TRUE;
+  }
+  /*$settings['back_to_site_link'] = config_get('admin_bar.settings',
+   'back_to_site_link');*/
+
+  //$settings['back_to_site_link'] = config_get('admin_bar.settings',
+  // 'components . [admin_bar.back_to_site_link]');
```
argiepiano: Remove commented code.

core/modules/admin_bar/admin_bar.module
argiepiano: @keiserjb, my original comment (which you resolved) still applies here. The Back to site link is not using the alias. You had fixed this, but not have regressed to the original problematic code.

core/modules/admin_bar/js/admin_bar.js
```diff
@@ -500,6 +500,54 @@ Backdrop.adminBar.behaviors.search = function (context, settings, $adminBar) {
...
+   *
+   * Back to site link points to the last non-administrative page the user visited
+   * within the same browser tab.
+   */
+  Backdrop.adminBar.behaviors.escapeAdmin = function (context, settings) {
+
+    // Don't do anything if variable isn't set.
+    /*if (settings.admin_bar.back_to_site_link && settings.admin_bar.back_to_site_link === 0) {
+      return;
+    }*/
```
argiepiano: Remove

core/modules/admin_bar/js/admin_bar.js
```diff
@@ -500,6 +500,54 @@ Backdrop.adminBar.behaviors.search = function (context, settings, $adminBar) {
...
+
+    if (!settings.admin_bar.back_to_site_link) {
+      return;
+    }
+
+    /*if (settings.admin_bar.back_to_site_link = FALSE) {
+      return;
+    }*/
```
argiepiano: Remove

core/modules/admin_bar/tests/admin_bar.tests.info
```diff
@@ -20,4 +20,8 @@ file = admin_bar.test
 name = Customized links
 description = Tests customized menu links.
 group = Administration bar
-file = admin_bar.test
\ No newline at end of file
+file = admin_bar.test
+; Added by Backdrop CMS packaging script on 2022-09-15
+project = backdrop
+version = 1.23.0
+timestamp = 1663295416
```
argiepiano: Remove

core/modules/admin_bar/admin_bar.module
```diff
@@ -244,6 +244,27 @@ function admin_bar_preprocess_page(&$variables) {
...
+
+  $components = config_get('admin_bar.settings', 'components');
```
argiepiano: I suggest replacing with:

```php
$components = $config->get('components');
```

(watch your spaces between closing parentheses in the following line).

core/modules/admin_bar/js/admin_bar.js
argiepiano: This commented line has returned? This causes the Home icon menu to be replaced even when the link points to the Home page. I don't know if this is intentional or not. I think it can go either way. If you don't like that, then remove the commented code.

core/modules/admin_bar/admin_bar.inc
keiserjb: Changes the "Home" link to a "Back to site" link. It returns to the last non-admin page visited.

keiserjb:
I think I got things back to the way they were.  I pushed from the wrong local site the other day.

core/modules/admin_bar/js/admin_bar.js
```diff
@@ -500,6 +500,46 @@ Backdrop.adminBar.behaviors.search = function (context, settings, $adminBar) {
...
 
+  /**
+   * Replaces the "Home" link with "Back to site" link.
+   *
+   * Back to site link points to the last non-administrative page the user visited
+   * within the same browser tab.
+   */
+  Backdrop.adminBar.behaviors.escapeAdmin = function (context, settings) {
+
+
```
argiepiano: Remove these two empty lines

core/modules/admin_bar/js/admin_bar.js
```diff
@@ -500,6 +500,46 @@ Backdrop.adminBar.behaviors.search = function (context, settings, $adminBar) {
...
+    if (!settings.admin_bar.back_to_site_link) {
+      return;
+    }
+
+        // Grab the stored path of the last non-admin page.
```
argiepiano: Comment indentation should coincide with the next line.

core/modules/admin_bar/js/admin_bar.js
```diff
@@ -500,6 +500,46 @@ Backdrop.adminBar.behaviors.search = function (context, settings, $adminBar) {
...
+    var escapeAdminPath = sessionStorage.getItem('escapeAdminPath');
+
+    // Saves the last non-administrative page in the browser to be able to link back
+    // to it when browsing administrative pages. If there is a destination parameter
+    // there is not need to save the current path because the page is loaded within
+    // an existing "workflow".
+    if (!settings.admin_bar.current_path_is_admin && !/destination=/.test(window.location.search)) {
+      sessionStorage.setItem('escapeAdminPath', settings.admin_bar.current_path);
+    }
+
```
argiepiano: Too many empty lines.

core/modules/admin_bar/js/admin_bar.js
```diff
@@ -500,6 +500,46 @@ Backdrop.adminBar.behaviors.search = function (context, settings, $adminBar) {
...
+
+
+    // We only want to change the first anchor tag in the admin bar icon sub-menu.
+    var $toolbarEscape = $('.admin-bar-icon a').first();
+
+    // If the current page is admin, then switch the path.
+    if ($toolbarEscape.length
+      && settings.admin_bar.current_path_is_admin
+      //&& settings.admin_bar.frontpage !== escapeAdminPath
```
argiepiano: So, if you prefer to always show the "Back to site" link (even when we are going back home), please remove this commented code.

core/modules/admin_bar/admin_bar.inc
```diff
@@ -826,6 +836,10 @@ function admin_bar_theme_settings() {
 function admin_bar_theme_settings_submit($form, &$form_state) {
   $components = &$form_state['values']['components'];
 
+  // Disable back to site if icon bar is also disabled.
+  if (!$components['admin_bar.icon']) {
+    $components['admin_bar.back_to_site_link'] = FALSE;
```
argiepiano, line 840: So, here's an idea about the behavior of this checkbox. Typically, when a checkbox depends on another and we hide it, we don't want to change the value of the hidden checkbox. So when/if the user changes their mind, they can re-enable to top checkbox and see that the lower one has the same value as before. So, my suggestion is to leave this value alone as is, and modify the preprocess function below to check for both, the `admin_bar.back_to_site_link` component AND the `admin_bar.icon` one. See my next  comment below.

If this sounds good to you, then these 4 lines can be removed.

core/modules/admin_bar/admin_bar.module
```diff
@@ -244,6 +244,22 @@ function admin_bar_preprocess_page(&$variables) {
...
+  $settings['current_path_is_admin'] = FALSE;
+  $current_path = current_path();
+  if (path_is_admin($current_path)) {
+    $settings['current_path_is_admin'] = TRUE;
+  }
+  $alias = backdrop_get_path_alias($current_path, $language->langcode);
+  $settings['current_path'] = $alias;
+
+  $components = $config->get('components');
+  if (in_array('admin_bar.back_to_site_link', $components) ) {
```
argiepiano: See my comment above in `admin_bar.inc`. If we decide to go the route of what I explained there, then this line should be:

```php
if (in_array('admin_bar.icon', $components) && in_array('admin_bar.back_to_site_link', $components)) {
```

Also, check the unnecessary space between parentheses at the end.

core/modules/admin_bar/admin_bar.module
```diff
@@ -244,6 +244,22 @@ function admin_bar_preprocess_page(&$variables) {
...
+    $settings['back_to_site_link'] = TRUE;
+  }
+
+  $settings['frontpage'] = config_get('system.core', 'site_frontpage');
+
```
argiepiano: Remove empty line.

core/modules/admin_bar/admin_bar.inc
keiserjb: I copied this from the lines underneath it which already has the same behavior.   The search bar is disabled when the management menu is deselected and setting saved.

argiepiano: You are right. Then let's just keep it as is.

core/modules/admin_bar/admin_bar.module
```diff
@@ -199,8 +200,7 @@ function admin_bar_preprocess_page(&$variables) {
   $args = explode('/', $_GET['q']);
   if ($args[0] == 'admin' && !empty($args[1])) {
     $settings['activeTrail'] = url($args[0] . '/' . $args[1]);
-  }
-  elseif (backdrop_is_front_page()) {
+  } elseif (backdrop_is_front_page()) {
```
argiepiano: elseif should start in a new line. Please check the [coding standards for Backdrop CMS](https://docs.backdropcms.org/php-standards#control-structures).

core/modules/admin_bar/js/admin_bar.js
```diff
@@ -1,547 +1,543 @@
...
+                            $adminBar.css({ visibility: '' });
+                        }, 0);
+                    }
+                });
+            }
+            // If the menu is in the output already, this means there is a new version.
+            else if (!$adminBar.hasClass('admin-bar-processed')) {
+                // Apply our behaviors.
+                Backdrop.adminBar.attachBehaviors(context, settings, $adminBar);
+            }
```
argiepiano: Another coding standard problem: tabs should be only 2 spaces (you are using 4).

keiserjb: My vscode has a mind of its own when it comes to this.

core/modules/admin_bar/css/admin_bar.css
```diff
@@ -83,6 +83,16 @@
   padding-right: 28px;
 }
 
+/* Back to Site arrow. */
+
+#admin-bar .dropdown li.admin-bar-icon > a.escape {
+  background-image: url("../images/chevron-disc-left.png");
```
argiepiano: Now that the file names have changed, you need to change the file names here as well

core/modules/admin_bar/css/admin_bar.css
```diff
@@ -83,6 +83,16 @@
...
+}
+[dir="rtl"] #admin-bar .dropdown li.admin-bar-icon > a.escape {
+  background-image: url("../images/chevron-disc-right.png");
```
argiepiano: File name needs to change too.

core/modules/admin_bar/tests/admin_bar.tests.info
```diff
@@ -20,4 +20,8 @@ file = admin_bar.test
 name = Customized links
 description = Tests customized menu links.
 group = Administration bar
-file = admin_bar.test
\ No newline at end of file
+file = admin_bar.test
+; Added by Backdrop CMS packaging script on 2022-09-15
```
argiepiano: Remove packaging information

core/modules/admin_bar/js/admin_bar.js
```diff
@@ -1,507 +1,634 @@
...
-      position_fixed: false,
-      destination: '',
-      basePath: settings.basePath,
-      hash: 0,
-      replacements: {}
-    }, settings.admin_bar || {});
-    // Check whether administration bar should be suppressed.
-    if (settings.admin_bar.suppress) {
+(function ($) {
+  Backdrop.adminBar = Backdrop.adminBar || {};
```
argiepiano: There are still issues with indentation in this js file. Notice that the diff view is showing you are making changes to ALL of the file. Can you restore the original admin_bar.js file and only change the lines that need changing? Right now it looks like you've made change to the whole file.

Backdrop js files most often start with:

```js
(function($) {
```

The lines that come after that are all flushed left (no indentation). This is done to increase readability, since the first line is just a way to pass the jQuery object as variable `$` to the function. The indentation  should only start with each individual `Backdrop.behaviors.adminBar = {` line and similar. 

Can you please restore the correct indentation of the file as it was? Thanks!

argiepiano (commented):
@keiserjb, this is another round of comments. Looking good, but there are still a couple of needed changes (now that the file names have changed, you need to change the names in the css file too). 

Also, many issues remain with the js file. It may be easiest to copy-drop the original js file and only change the lines that need change. No reflowing. See my comment about indentation. Thanks

core/modules/admin_bar/css/admin_bar.css
keiserjb: Now I make stupid mistakes.

core/modules/admin_bar/js/admin_bar.js
keiserjb: I'm honestly not doing that intentionally.  My editors are reformatting it, even when I try to fix it.

core/modules/admin_bar/admin_bar.module
```diff
@@ -1,4 +1,5 @@
 <?php
+
```
argiepiano: No line between <?php and the first line of the docblock below.

core/modules/admin_bar/tests/admin_bar.tests.info
argiepiano: Packaging info still present in this PR... Please remove.

core/modules/admin_bar/admin_bar.module
```diff
@@ -244,6 +245,21 @@ function admin_bar_preprocess_page(&$variables) {
...
+  }
+  $alias = backdrop_get_path_alias($current_path, $language->langcode);
+  $settings['current_path'] = $alias;
+
+  $components = $config->get('components');
+  if (in_array('admin_bar.back_to_site_link', $components)) {
+    $settings['back_to_site_link'] = TRUE;
+  }
+
+  $settings['frontpage'] = config_get('system.core', 'site_frontpage');
```
argiepiano: What is the purpose of this? `frontpage` is not used anywhere...

keiserjb: I think this was for when back to site didn’t replace home when going back home but I’ll make sure before I remove it.

keiserjb: Yes this was not necessary.

core/modules/admin_bar/admin_bar.inc
```diff
@@ -784,6 +784,7 @@ function admin_bar_theme_settings() {
     '#title' => t('Enabled components'),
     '#options' => array(
       'admin_bar.icon' => t('Icon menu'),
+      'admin_bar.back_to_site_link' => t('Add Back to site link'),
```
jenlampton: All of the other items under "Enabled components" list only the name of the item, so we can remove "Add" here, it should simply be 'Back to site link'.

core/modules/admin_bar/admin_bar.inc
```diff
@@ -802,6 +810,8 @@ function admin_bar_theme_settings() {
   );
 
   $form['components']['admin_bar.icon']['#description'] = t('A specialized menu that links to the home page and provides links to perform various other admin tasks (e.g.: clear caches, run cron, etc.).');
+  $form['components']['admin_bar.back_to_site_link']['#description'] = t
+  ('Changes the "Home" link to a "Back to site" link. It returns to the last non-admin page visited.');
```
jenlampton: Alternative text:
`Replaces the "Home" link with a "Back to site" link that returns to the last non-admin page visited.`

jenlampton:
I found two minor changes for the admin text (noted above) the first (Remove "Add" in the list of "Enabled components") is necessary, the second I think is an improvement, but could use more eye's on. 

@keiserjb for some reason github would not allow me to create a PR against your PR, but it did allow me to add a commit to your branch. I don't usually like doing that... but I got stuck today for some reason.

core/modules/admin_bar/admin_bar.inc
keiserjb: Yes, this is better.

keiserjb: It may have been that at one point.

quicksketch:
Merged into 1.x for 1.24.0. Great work @keiserjb (and @alexfinnarn who made the original PR)! Full list of thanks in https://github.com/backdrop/backdrop-issues/issues/2709#issuecomment-1335940356

PR #4237, opened 2022-10-20
Issue #5813: Prefer native random_bytes(), prefer /dev/urandom as fallback

### LINKED ISSUE #5813: [SR] Switch to using `random_bytes()` where available (built into php7)
This issue is a spin-off of a discussion that started in #5812, and a task to crossport https://www.drupal.org/project/drupal/issues/2550519 that's planned for D7.

> ## Problem/Motivation
> Drupal prefers `openssl_random_pseudo_bytes()` if available in `Crypt::randomBytes()` in 8.0.x or `drupal_random_bytes()` in 7.x and 6.x.
> 
> PHP used the wrong method in the openssl library now fixed in 5.6.12, 5.5.28, 5.4.44 see: https://bugs.php.net/bug.php?id=70014 but it is **NOT** classified as a security hole so backports are to older (ie distro) versions are less likely. Just because PHP didn't declare this a security hole doesn't mean it is not. But also read on for other scenarios where this class creates a security hole.
> 
> The returned pseudo-random bytes were NOT necessarily cryptographically secure.
> 
> ## Proposed resolution
> Add [random_compat](https://github.com/paragonie/random_compat) v2.0.2 to core + use it for Crypt::randomBytes()
> Adding v2 should be ok here because in drupal 7.x this includes a fallback.
> (so not a concern like in [#2763787: Upgrade random_compat to latest version](https://www.drupal.org/project/drupal/issues/2763787) which reverted the upgrade to v2.0.2 in drupal 8.2.x and 8.3.x)

indigoxela:
Fixes https://github.com/backdrop/backdrop-issues/issues/5813

And (unrelated) coding standard fixes in that function.

### DIFF
--- a/core/includes/bootstrap.inc
+++ b/core/includes/bootstrap.inc
@@ -2559,7 +2559,7 @@ function backdrop_base64_encode($string) {
  * function because it can return a long string of bytes (compared to < 4 bytes
  * normally from mt_rand()), and uses the best available pseudo-random source.
  *
- * @param $count
+ * @param int $count
  *   The number of characters (bytes) to return in the string.
  *
  * @return string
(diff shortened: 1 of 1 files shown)


klonos (approved):
LGTM 👍🏼

quicksketch:
Merged into 1.x and 1.23.x. Thanks @indigoxela!

PR #4239, opened 2022-10-23
Issue 5841. Add absolute position to ui-dialog in style.css

### LINKED ISSUE #5841: Fixed: Token browser disappears below viewport when resized.
## Steps To Reproduce
1. Go to the configuration page of any content type (or any page that has a token browser link, such as `admin/config/urls/path/patterns`)
2. Click on "Browse available tokens"
3. Try dragging and dropping the dialog that pops up -> that works fine 👍🏼 
4. Try resizing the dialog -> it disappears! 👎🏼 

## Actual behavior
The token browser dialog disappears (is moved outside viewport actually*)

## Expected behavior
The token browser dialog stays in place when/after resizing it.

## Additional information
If you have a big monitor and it's set to very high resolution, you will see that the dialog actually gets moved to the bottom of the page. If it is not visible, then simply scroll to the bottom of the page, and you can find and drag it back within viewport ...but once you try to resize it, this whole disappearing/moving of the dialog keeps happening.

argiepiano:
Fixes https://github.com/backdrop/backdrop-issues/issues/5841

### DIFF
--- a/core/themes/seven/css/style.css
+++ b/core/themes/seven/css/style.css
@@ -1782,6 +1782,7 @@ div.add-or-remove-shortcuts {
   background: #e9e8e8;
 }
 .ui-dialog {
+  position: absolute;
   border: 1px solid #d0d0d0;
   -webkit-border-radius: 4px;
   -moz-border-radius: 4px;


quicksketch:

PR #4240, opened 2022-10-23
Issue 5842. Fix misspelling of dialog's resizable

### LINKED ISSUE #5842: Fixed: misspelling of dialog's "resizable" JS setting option in `system_token_output()`.
## Description of the bug

The jQuery UI dialog plugin contains a setting option called `resizable` [documented here](https://api.jqueryui.com/dialog/#option-resizable). That option is misspelled in the settings done for the token tree plugin [here](https://github.com/backdrop/backdrop/blob/08d182106e82521a39e6fe11f1ed49a61d77fa8f/core/modules/system/system.pages.inc#L27) in file `system.pages.inc`.

This misspelling has basically no effect; it's ignored. The default for resizable is true anyways.

## Steps To Reproduce

There is nothing to reproduce, as the misspelling doesn't affect the functioning of the dialog ("resizeable" is ignored, and the default `true` takes effect for `resizable`).

argiepiano:
Fixes https://github.com/backdrop/backdrop-issues/issues/5842

### DIFF
--- a/core/modules/system/system.pages.inc
+++ b/core/modules/system/system.pages.inc
@@ -24,7 +24,7 @@ function system_token_output() {
       'dialogClass' => 'token-dialog',
       'modal' => FALSE,
       'draggable' => TRUE,
-      'resizeable' => TRUE,
+      'resizable' => TRUE,
       'autoResize' => FALSE,
       'width' => '600',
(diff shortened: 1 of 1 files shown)


quicksketch:

PR #4242, opened 2022-10-27
Issue #5843: Linked files lose permanent status after edit

### LINKED ISSUE #5843: Fixed: Files lose permanent status when in edited in CKEditor link dialog.
## Description of the bug

With file uploads for a text format enabled, CKEditor can be used to upload files through the link dialog. Such files lose their permanent status, when the correspondent link is edited using the CKEditor link dialog.

## Steps To Reproduce

To reproduce the behavior:
1. Enable file uploads on `admin/config/content/formats/filtered_html`.
2. Add a Test post, insert a file link in the Body field, using the link dialog in CKEditor and its tab "Upload a file and link to it". Save the link dialog.
3. Check the Source code, it should be something like `<a data-file-id="4" href="/files/inline-files/dummy.pdf">`.
4. Save the node.
5. Check `admin/content/files`. The Status of the file should be `Permanent`, Use count should be `1`.
6. Edit the test post. Edit the link using the link dialog. Save the dialog. (It doesn't matter if you actually change something.)
7. Check the Source code: the `data-file-id` from step 3 is missing.
8. Save the node.
9. Check `admin/content/files`. Status of the file is `Temporary`, Use count is none.

## Actual behavior

On link edting, the `data-file-id` is removed.

## Expected behavior

The `data-file-id` should not be removed.

## Additional information

- Tested in the Tugboat driven demo site.
- Not yet tested, but I suspect data loss, as temporary files get removed at a certain point.

indigoxela:
Fixes https://github.com/backdrop/backdrop-issues/issues/5843

### DIFF
--- a/core/modules/filter/filter.pages.inc
+++ b/core/modules/filter/filter.pages.inc
@@ -656,9 +656,16 @@ function _filter_url_has_protocol($url) {
 function _filter_get_file_id($href) {
   $result = 0;
 
+  // Make sure this is a file.
+  global $base_path;
+  $file_path = $base_path . config_get('system.core', 'file_public_path');
+  if (strpos($href, $file_path) !== 0) {
+    return $result;
+  }
+
   // Extract file name and directory from $href.
   $pos = strrpos($href, "/");
-  $filename = substr($href,$pos + 1);
+  $filename = substr($href, $pos + 1);
 
   // Now find this file name in uri field to get fid.
   if (!empty($filename)) {
--- a/core/modules/filter/filter.pages.inc
+++ b/core/modules/filter/filter.pages.inc
@@ -659,6 +659,8 @@ function _filter_get_file_id($href) {
   // Make sure this is a file, either in public or in private path.
   global $base_path;
   $public_path = config_get('system.core', 'file_public_path');
+  // Private files have "system/files/" in their URL.
+  // @see BackdropPrivateStreamWrapper::getExternalUrl()
   $pattern = '#^' . $base_path . '(\?q=)?(' . $public_path . '|system/files/' . ')#';
   if (preg_match($pattern, $href) === FALSE) {
     return $result;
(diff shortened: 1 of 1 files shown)


core/modules/filter/filter.pages.inc
```diff
@@ -656,14 +656,16 @@ function _filter_url_has_protocol($url) {
 function _filter_get_file_id($href) {
   $result = 0;
 
-  // Check URI for valid file scheme.
-  if (!file_valid_uri($href)){
+  // Make sure this is a file.
+  global $base_path;
+  $file_path = $base_path . config_get('system.core', 'file_public_path');
```
klonos: Why are we assuming that the public files scheme will be the one used for files in all use cases? ...file uploads in CKEditor can be configured to use the private files directory:

image

core/modules/filter/filter.pages.inc
```diff
@@ -656,14 +656,16 @@ function _filter_url_has_protocol($url) {
...
+  if (strpos($href, $file_path) !== 0) {
     return $result;
   }
 
   // Extract file name and directory from $href.
   $pos = strrpos($href, "/");
-  $filename = substr($href,$pos + 1);
+  $filename = substr($href, $pos + 1);
```
klonos, line 671: Good catch 👍🏼

core/modules/filter/filter.pages.inc
indigoxela: Oops, you're right. We have to consider that.

core/modules/filter/filter.pages.inc
```diff
@@ -656,20 +656,23 @@ function _filter_url_has_protocol($url) {
 function _filter_get_file_id($href) {
   $result = 0;
 
-  // Check URI for valid file scheme.
-  if (!file_valid_uri($href)){
+  // Make sure this is a file, either in public or in private path.
+  global $base_path;
+  $public_path = config_get('system.core', 'file_public_path');
+  $pattern = '#^' . $base_path . '(\?q=)?(' . $public_path . '|system/files/' . ')#';
```
klonos, line 664:
```suggestion
  // Private files have "system/files/" in their URL.
  // @see BackdropPrivateStreamWrapper::getExternalUrl()
  $pattern = '#^' . $base_path . '(\?q=)?(' . $public_path . '|system/files/' . ')#';
```

quicksketch:

PR #4247, opened 2022-10-31
Issue 5835. Another js approach.

### LINKED ISSUE #5835: Fixed: Existing Image library popup does not show all images.
## Description of the bug

When there are amount of images in the existing image library and I try to open the image library popup. It shows only part of the row of the images. But when I scroll down the pop-up resizes and shows all the images and the entire image library popup modal.

## Steps To Reproduce

To reproduce the behavior:
1. Add content "Post". 
2. Create “title” and then “Select existing file” for an image field.
3. Image library only shows part of one row.
4. When scrolling down or top it shows the entire image library Pop-up.

The existing Image library should have a number of images to reproduce this.

## Actual behavior
This is how it looks when the pop-up opens first.
edit1

## Expected behavior

The existing image library should show the full window
edit2

## Additional information

- Web server and its version: Nginx/1.18.0
- Operating System and its version: Ubuntu 20.04 lts
- Browser(s) and their versions: Google Chrome V106.0

argiepiano:
Fixes https://github.com/backdrop/backdrop-issues/issues/5835

This is a modified approach as posted by @indigoxela's #4243

### DIFF
--- a/core/misc/dialog.js
+++ b/core/misc/dialog.js
@@ -75,7 +75,35 @@ Backdrop.dialog = function (element, options) {
         })
         .dialog('widget').css('position', 'fixed');
       Backdrop.optimizedResize.add(resetPosition, 'dialogResize.' + dialogId);
-      resetPosition();
+
+      // Because of a bug in Chromium, we must wait until the ajax-loaded
+      // jquery.ui.dialog.css is applied to the content of the modal through the
+      // class .ui-dialog-content. 
+
+      // First, get the computed style for the content element. Note: this
+      // variable changes automatically when the element's style is changed. 
+      var computedStyle = getComputedStyle($element[0]);
+
+      // Get the current value for the overflow property, which must by 'auto'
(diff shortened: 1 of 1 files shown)


core/misc/dialog.js
```diff
@@ -75,7 +75,35 @@ Backdrop.dialog = function (element, options) {
         })
         .dialog('widget').css('position', 'fixed');
       Backdrop.optimizedResize.add(resetPosition, 'dialogResize.' + dialogId);
-      resetPosition();
+
+      // Because of a bug in Chromium, we must wait until the ajax-loaded
+      // jquery.ui.dialog.css is applied to the content of the modal through the
+      // class .ui-dialog-content.
```
indigoxela: Can we reduce this comment a bit?

I'm not actually sure, if I'd call that a bug BTW. It's just "behavior".

argiepiano: I've trimmed the documentation.

PR #4248, opened 2022-11-01
issue 5847. Fix how date field HTML5 gets form values

### LINKED ISSUE #5847: Fixed: HTML5 date field widget produces a fatal error on a Paragraph.
## Description of the bug

When a date field is attached to a Paragraph and the field uses the new HTML5 date widget, saving the node throws a fatal error `EntityStorageException: SQLSTATE[21S01]: Insert value list does not match column list: 1136 Column count doesn't match value count at row 1 of NodeStorageController->save() (line 510 of XXX/modules/node/node.entity.inc)`.

## Steps To Reproduce

1. Create a Paragraph type and attach a date field that uses the HTML5 widget
2. Add a paragraph reference field to a node type (unlimited cardinality) that references the new Paragraph type
3. Add a node of that note type, and enter a value for the date field
4. Save the node and see the fatal error

## Actual behavior

Fatal error.

## Expected behavior

The node should be saved with all its values, including the date field.

## Additional information

This was first reported in the [Paragraph issue queue](https://github.com/backdrop-contrib/paragraphs/issues/148).

The issue happens because of how `date_html5_element_validate()` handles the submitted form values. This validation function takes care of transforming the HTML5 date widget submitted values into a structure that can be saved to the date field table. But the problem is that this function assumes the values for the field are stored in `$form_state['values'][FIELD_NAME][LANGUAGE_NONE]`.

Unfortunately, this is not the case when the date is attached to another entity referenced by a field in the node (like the Paragraph Item entity referenced by a paragraph reference field). I expect this will also fail with other types of referenced entities that attach their forms to node forms, such as Field Collection entities.

The solution is to avoid hard-coding the location of the values. PR coming.

argiepiano:
Fixes https://github.com/backdrop/backdrop-issues/issues/5847

### DIFF
--- a/core/modules/date/date.elements.inc
+++ b/core/modules/date/date.elements.inc
@@ -1837,9 +1837,10 @@ function date_hidden_element($element) {
 function date_html5_element_validate(array $element, array &$form_state) {
   $field_name = $element['#field_name'];
 
-  if (!empty($form_state['values'][$field_name][LANGUAGE_NONE])) {
+  if (!empty(backdrop_array_get_nested_value($form_state['values'], $element['#parents']))) {
     $values = array();
-    $items = $form_state['values'][$field_name][LANGUAGE_NONE];
+    $trimmed_parents = array_slice($element['#parents'], 0, -1);
+    $items = backdrop_array_get_nested_value($form_state['values'], $trimmed_parents);
     $field_type = $element['#field_type'];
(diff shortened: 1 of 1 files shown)


robertgarrigos:
I can confirm that this pf fixes the problem. Thanks!

quicksketch:

PR #4254, opened 2022-11-12
Issue #5858: System metrics in status page

### LINKED ISSUE #5858: Provide a "System metrics" entry in the status page, to show disk space, RAM usage, and database size
This feature request was inspired by [this discussion in Zulip](https://backdrop.zulipchat.com/#narrow/stream/218635-Backdrop/topic/Minimal.20Backdrop) around "Minimal Backdrop", which touched on topics around what would be the very minimum system specs where Backdrop core would run, and how well it would perform.

image

Something like this might be helpful to allow determining requirements when:
- moving a locally-built site to an actual hosting provider
- moving a live site to a different provider
- upgrading the hosting plan of an existing site (when certain specs seem to have been reaching capacity/limits)

klonos:
Fixes https://github.com/backdrop/backdrop-issues/issues/5858

### DIFF
--- a/core/modules/system/system.install
+++ b/core/modules/system/system.install
@@ -66,9 +66,11 @@ function system_requirements($phase) {
     $directory_sizes = array();
     $total_dir_size = 0;
     foreach ($directories as $directory => $path) {
-      $dir_size = _system_get_dir_size($path);
-      $directory_sizes[$directory] = $t('@dir: @size', array('@dir' => $directory, '@size' => format_size($dir_size)));
-      $total_dir_size += $dir_size;
+      if (!empty($path)) {
+        $dir_size = _system_get_dir_size($path);
+        $directory_sizes[$directory] = $t('@dir: @size', array('@dir' => $directory, '@size' => format_size($dir_size)));
+        $total_dir_size += $dir_size;
+      }
     }
     $dir_size_breakdown = theme('item_list', array('items' => $directory_sizes));
 
--- a/core/modules/system/system.install
+++ b/core/modules/system/system.install
@@ -14,7 +14,111 @@ function system_requirements($phase) {
   // Ensure translations don't break during installation.
   $t = get_t();
 
-  // Report Backdrop version
+  // System metrics.
+  if ($phase == 'runtime') {
+    $metrics = array();
+
+    $free_output = shell_exec('free');
+    $free_output = (string) trim($free_output);
+    $free_lines = explode("\n", $free_output);
+
+    $result = array();
+    foreach ($free_lines as $line => $output) {
+      // Convert to lowercase.
+      $output = strtolower($output);
+      // Remove multiple spaces.
+      $output = preg_replace('/[\s]+/mu', ' ', $output);
+      // Convert any special characters to underscore.
+      $output = preg_replace('/[^a-zA-Z0-9_\s]/', '_', $output);
+
+      $result[$line] = explode(" ", $output);
+    }
+
+    // Grab the 1st line of the output (which is the heading of the various
+    // columns), then remove that from the rest of the output.
+    $header = $result[0];
+    unset($result[0]);
+    // Grab each RAM metric (per column).
+    $mem_metrics = array();
+    foreach ($result as $result_line => $data) {
+      // The first column contains the type of memory ("mem" or "swap").
+      $prefix = $data[0];
+      array_splice($data, 0, 1);
+      foreach ($data as $column => $metric) {
+        $mem_metrics[$prefix . $header[$column]] = $metric;
+      }
+    }
+
+    $mem_usage = round(($mem_metrics['mem_total'] - $mem_metrics['mem_available']) / $mem_metrics['mem_total'] * 100);
+
+    // Disk metrics. Should work the same across operating systems.
+    $disk_total = disk_total_space('.');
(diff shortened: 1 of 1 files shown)


core/modules/system/system.install
```diff
@@ -23,8 +127,8 @@ function system_requirements($phase) {
       'weight' => -10,
     );
 
-    // Display the currently active installation profile, if the site
-    // is not running the default installation profile.
+    // Display the currently active installation profile, if the site is not
```
bugfolder:
```suggestion
    // Display the currently active installation profile if the site is not
```

No comma if the clause is dependent (which is the case here).

core/modules/system/system.install
```diff
@@ -14,7 +14,129 @@ function system_requirements($phase) {
   // Ensure translations don't break during installation.
   $t = get_t();
 
-  // Report Backdrop version
+  // System metrics.
+  if ($phase == 'runtime') {
+    $metrics = array();
+
+    $free_output = shell_exec('free');
```
indigoxela, line 21: Caution! On shared hosting `shell_exec` might be turned off for security reasons. In that case the install would break.

core/modules/system/system.install
```diff
@@ -14,7 +14,129 @@ function system_requirements($phase) {
...
+    // CPU load - not implemented on Windows platforms.
+    if (stristr(PHP_OS, 'win') === FALSE) {
+      $cpu_load = sys_getloadavg();
+      // This is not a percentage. For details, see:
+      // - https://stackoverflow.com/questions/9719155/understading-sys-getloadavg
+      // - https://www.php.net/manual/function.sys-getloadavg.php
+      $cpu_metric = $t('CPU load (over 1, 5 and 15 minutes): @load', array('@load' => implode(' / ', $cpu_load)));
+    }
+    else {
+      $cpu_metric = $t('CPU load cannot be estimated on Windows platforms');
```
indigoxela, line 102: Backdrop hosting on Windows? Really?

core/modules/system/system.install
```diff
@@ -14,7 +14,129 @@ function system_requirements($phase) {
...
+      if (!empty($path)) {
+        $dir_size = _system_get_dir_size($path);
+        $directory_sizes[$directory] = $t('@dir: @size', array('@dir' => $directory, '@size' => format_size($dir_size)));
+        $total_dir_size += $dir_size;
+      }
+    }
+    $dir_size_breakdown = theme('item_list', array('items' => $directory_sizes));
+
+    // Database metrics.
+    $rows = db_query('SHOW TABLE STATUS')->fetchAll();
```
indigoxela, line 78: There might be a smarter way to do this.

core/modules/system/system.install
```diff
@@ -664,6 +786,23 @@ function system_requirements($phase) {
...
+ * Gets the size of a specified directory.
+ *
+ * @param string $directory
+ *
+ * @return integer
+ */
+function _system_get_dir_size($directory) {
+  $size = 0;
+
+  foreach (new RecursiveIteratorIterator(new RecursiveDirectoryIterator($directory)) as $file) {
```
indigoxela, line 799: This means, on *every* call of the status page, we do a full iteration, which can be an expensive task with many files in public/private.

core/modules/system/system.install
klonos: Apparently some people do that 🤷🏼 ...Amazon offers Windows servers on the cloud: https://aws.amazon.com/windows

...and our telemetry shows that there are few sites running on Windows (although I'm assuming that this is most likely on local, before pushing changes to a linux seever)
image

Anyway, this was some "defensive coding", to account the display of this when developing on Windows local, running WAMPs etc.

core/modules/system/system.install
klonos: ...yup, should add some defensive coding there, since `free` is not available on Windows either.

core/modules/system/system.install
klonos: This was based on some quick research on the internet. Happy to be provided with other ways/ideas.

core/modules/system/system.install
klonos: Yeah. ...wonder if caching the results makes sense in this scenario 🤔 ...on the other hand, how often would an admin visit that page? It's not like they'd sit there refreshing the page all day, right?

indigoxela: My concern is that the status page is already sort of "expensive" - many checks are done in one request. Now think of a site with thousands of files... it might even be that the status page runs into a timeout (or PHP out of resources, worst case).

core/modules/system/system.install
indigoxela: > ...and our telemetry shows that there are few sites running on Windows...

One percent. :smirk_cat: And, yes, I agree that these are very likely local dev instances.

core/modules/system/system.install
indigoxela: A quick experiment:

```
global $databases;
$db_name = $databases['default']['default']['database'];

$size = db_query('SELECT ROUND(SUM(data_length + index_length) / 1024 / 1024, 2) as size_mb FROM information_schema.tables WHERE table_schema = :dbname GROUP BY table_schema', array(':dbname' => $db_name))->fetchField();

debug($size);
```

PR #4256, opened 2022-11-19
Issue #5863: Consolidate PHP-related entries in status page

### LINKED ISSUE #5863: [UX] Consolidate PHP-related information in the status report page, and add some basic PHP memory usage metrics
This is a spin off of #5858, where we considered adding some system metrics to the status page, to help site admins/owners get a better understanding of the minimum-recommended and minimum-possible specs required for a server that would run their site (when they are looking to move their locally-developed site to a live server, or when switching hosting providers).

In #5858 we decided that showing system RAM usage may not be relevant, but showing some basic PHP memory usage would be helpful. While working on the code, I've noticed some code repetition and certain things that could be consolidated. For example, things like the PHP extensions, memory limit, and register globals are not frequently used past the initial site installation, so I thought that they could be placed under the "more" toggle in the main PHP entry (where we show the PHP version). That, plus some other minor tweaks.

Also, I wanted to split this from the rest of the things we considered (such as db size and files directory size), in order to not hold back one thing while discussing/bikeshedding the others.

klonos:
Fixes https://github.com/backdrop/backdrop-issues/issues/5863

### DIFF
--- a/core/modules/system/system.install
+++ b/core/modules/system/system.install
@@ -50,72 +50,120 @@ function system_requirements($phase) {
     'severity' => REQUIREMENT_INFO,
   );
 
-  // Test PHP version and show link to phpinfo() if it's available
-  $phpversion = phpversion();
-  if (function_exists('phpinfo')) {
-    if ($phase == 'runtime') {
-      $php_value = $t('Version: !version (<a href="!infourl">PHP information</a>)', array('!version' => $phpversion, '!infourl' => url('admin/reports/status/php')));
-    }
-    else {
-      $php_value = $t('Version: !version', array('!version' => $phpversion));
-    }
-    $requirements['php'] = array(
-      'title' => $t('PHP'),
-      'value' =>  $php_value,
-      'severity' => REQUIREMENT_INFO,
-    );
-  }
-  else {
-    $requirements['php'] = array(
-      'title' => $t('PHP'),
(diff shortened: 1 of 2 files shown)


core/modules/system/system.install
```diff
@@ -50,72 +50,120 @@ function system_requirements($phase) {
...
+      $requirements['php']['value'] .= ' (' . l(t('PHP information'), 'admin/reports/status/php') . ')';
+    }
+  }
+  else {
+    $requirements['php']['description'] = $t('The <code>phpinfo()</code> function has been disabled for security reasons. To see your server\'s <code>phpinfo()</code> information, change your PHP settings. For more information, see the <a href="@phpinfo">Enabling and disabling phpinfo()</a> handbook page.', array('@phpinfo' => 'https://backdropcms.org/user-guide/enabling-and-disabling-phpinfo'));
+  }
+
+  if ($phase === 'runtime') {
+    $php_mem_usage_info = $t('PHP memory usage: @usage | @peak (peak) | @true (true)', array(
+      '@usage' => format_size(memory_get_usage()),
```
indigoxela, line 96: @klonos do you happen to know what memory_get_usage() measures?

I suppose it's the memory usage of the current process. Wouldn't that be the identical number every time? The amount of memory the status page needed?

PR #4258, opened 2022-11-26
Issue #5835: Resize dialogs after jQuery UI CSS has been loaded.

### LINKED ISSUE #5835: Fixed: Existing Image library popup does not show all images.

quicksketch:
Fixes https://github.com/backdrop/backdrop-issues/issues/5835

Replaces https://github.com/backdrop/backdrop/pull/4247

### DIFF
--- a/core/misc/dialog.js
+++ b/core/misc/dialog.js
@@ -75,7 +75,35 @@ Backdrop.dialog = function (element, options) {
         })
         .dialog('widget').css('position', 'fixed');
       Backdrop.optimizedResize.add(resetPosition, 'dialogResize.' + dialogId);
-      resetPosition();
+
+      // Because of a bug in Chromium, we must wait until the ajax-loaded
+      // jquery.ui.dialog.css is applied to the content of the modal through the
+      // class .ui-dialog-content. 
+
+      // First, get the computed style for the content element. Note: this
+      // variable changes automatically when the element's style is changed. 
+      var computedStyle = getComputedStyle($element[0]);
+
+      // Get the current value for the overflow property, which must by 'auto'
+      // after jquery.ui.dialog.css has been applied. 
+      var cssOverflowProperty = computedStyle.getPropertyValue('overflow');
+      if (cssOverflowProperty != 'auto') {
+        setTimeout(
+          function waitForCss() {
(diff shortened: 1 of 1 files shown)


core/misc/dialog.js
```diff
@@ -75,7 +75,20 @@ Backdrop.dialog = function (element, options) {
         })
         .dialog('widget').css('position', 'fixed');
       Backdrop.optimizedResize.add(resetPosition, 'dialogResize.' + dialogId);
-      resetPosition();
+
+      // Because of Chromium's behavior, we must wait until the ajax-loaded
+      // jquery.ui.dialog.css is applied.
+      var waitForCss = function() {
+        var cssOverflowProperty = getComputedStyle($element[0]).getPropertyValue('overflow');
```
argiepiano: In my original PR I had this variable assignment outside the closure. Scope-wise, your approach may be better, as it keeps the variable within the scope of the closure.  I approached this differently because one of the features of `getComputedStyle()` is that it creates a live object "which updates automatically when the element's styles are changed" according to the [docs](https://developer.mozilla.org/en-US/docs/Web/API/Window/getComputedStyle). So, this way, we avoid making a new assignment every time the function is invoked. 

Both approaches work, and I guess it's just a matter of preference.

quicksketch: Good idea. I didn't know that about `getComputedStyle()`. I moved it to a separate variable again. :+1:

quicksketch:

PR #4265, opened 2022-12-02
Issue #5872: Fix backdrop_is_cli for PHP8.1

### LINKED ISSUE #5872: PHP 8.1 and backdrop_is_cli()
## Description of the bug

When getting `bee` ready for PHP 8.1, we set `$_SERVER['SERVER_SOFTWARE'] = ''` rather than `$_SERVER['SERVER_SOFTWARE'] = NULL` as several functions in `core/includes/bootstrap.inc` use `preg_match()` which under PHP 8.1 will not accept `NULL` as a string.

We get deprecation notices:
`preg_match(): Passing null to parameter #2 ($subject) of type string is deprecated` - see https://github.com/backdrop-contrib/bee/issues/222

Hence why we changed to `$_SERVER['SERVER_SOFTWARE'] = ''`

However, `backdrop_is_cli()` does not evaluate `''` to being `!isset()` and this (tracked down by @indigoxela ) then causes `bee` to freeze when the command would generate a prompt for the user to confirm an action.

See:
- https://github.com/backdrop-contrib/bee/issues/244
- https://github.com/backdrop-contrib/bee/issues/234

If the function could be changed to:
```php
function backdrop_is_cli() {
  return (empty($_SERVER['SERVER_SOFTWARE']) && (php_sapi_name() == 'cli' || (is_numeric($_SERVER['argc']) && $_SERVER['argc'] > 0)));
}
```
Then this would work whether the value was `NULL` or `''` 

---
PR added - https://github.com/backdrop/backdrop/pull/4265

yorkshire-pudding:
Fixes [https://github.com/backdrop/backdrop-issues/issues/5872](https://github.com/backdrop/backdrop-issues/issues/5872)

### DIFF
--- a/core/includes/bootstrap.inc
+++ b/core/includes/bootstrap.inc
@@ -4215,7 +4215,7 @@ function backdrop_static_reset($name = NULL) {
  * Detects whether the current script is running in a command-line environment.
  */
 function backdrop_is_cli() {
-  return (!isset($_SERVER['SERVER_SOFTWARE']) && (php_sapi_name() == 'cli' || (is_numeric($_SERVER['argc']) && $_SERVER['argc'] > 0)));
+  return (empty($_SERVER['SERVER_SOFTWARE']) && (php_sapi_name() == 'cli' || (is_numeric($_SERVER['argc']) && $_SERVER['argc'] > 0)));
 }
 
 /**


quicksketch:
Merged into 1.x and 1.23.x. Thanks @yorkshire-pudding!

PR #4271, opened 2022-12-09
Issue #5748: Add a description option to user roles.

### LINKED ISSUE #5748: Add descriptions for user roles.
This is the same as https://www.drupal.org/project/drupal/issues/256287:

> It would be useful to be able to document on the site why a certain role exists.

The feature is currently provided by the contrib module https://github.com/backdrop-contrib/role_help (that module also allows for a separate help text field that is shown to the user, but I don't think that this should be provided by core).

---

Advocate: @quicksketch

---

Minor release update:
- [x] Add a conclusion suitable for blog post + release notes
- [x] Add Documentation label if the feature needs to be documented (or needs docs to be updated)
- [x] N/A Add Needs change record label if the the issue contains an API change.
- [x] Add Needs translation label if the the issue contains string translations.

---

String translations:
* `Summary of this role's purpose, shown on the role listing page and when assigning this role on user account forms.`
* `Backdrop core now provides most of the functionality of Role Help module. See the <a href="!url">change record for more information</a>.`

---

Backdrop now includes the ability to add descriptions for user roles.  This will help administrators understand the differences between roles and assign them appropriately.

quicksketch:
Fixes https://github.com/backdrop/backdrop-issues/issues/5748

### DIFF
--- a/core/modules/user/config/user.role.authenticated.json
+++ b/core/modules/user/config/user.role.authenticated.json
@@ -3,7 +3,7 @@
     "name": "authenticated",
     "label": "Authenticated",
     "admin_description": "Special role assigned to all accounts. Permissions assigned to this role are given to all other roles except the anonymous role.",
-    "description": "Automatically assigned role to all accounts.",
+    "description": "Automatically assigned to all accounts.",
     "weight": "1",
     "permissions": []
 }
--- a/core/modules/user/config/user.role.authenticated.json
+++ b/core/modules/user/config/user.role.authenticated.json
@@ -2,8 +2,8 @@
     "_config_name": "user.role.authenticated",
     "name": "authenticated",
     "label": "Authenticated",
-    "admin_description": "Special role assigned to all users who have signed into the site. Permissions assigned to this role are given to all other roles except the anonymous role.",
-    "description": "",
+    "admin_description": "Special role assigned to all accounts. Permissions assigned to this role are given to all other roles except the anonymous role.",
+    "description": "Automatically assigned role to all accounts.",
     "weight": "1",
     "permissions": []
 }
--- a/core/modules/user/config/user.role.authenticated.json
+++ b/core/modules/user/config/user.role.authenticated.json
@@ -2,6 +2,7 @@
     "_config_name": "user.role.authenticated",
     "name": "authenticated",
     "label": "Authenticated",
+    "description": "Special role assigned to all users who have signed into the site. Permissions assigned to this role are given to all other roles except the anonymous role.",
     "weight": "1",
     "permissions": []
 }
--- a/core/modules/user/config/user.role.authenticated.json
+++ b/core/modules/user/config/user.role.authenticated.json
@@ -2,7 +2,8 @@
     "_config_name": "user.role.authenticated",
     "name": "authenticated",
     "label": "Authenticated",
-    "description": "Special role assigned to all users who have signed into the site. Permissions assigned to this role are given to all other roles except the anonymous role.",
+    "admin_description": "Special role assigned to all users who have signed into the site. Permissions assigned to this role are given to all other roles except the anonymous role.",
+    "description": "",
     "weight": "1",
     "permissions": []
 }
--- a/core/modules/user/tests/user.test
+++ b/core/modules/user/tests/user.test
@@ -2181,7 +2181,7 @@ class UserRoleAdminTestCase extends BackdropWebTestCase {
 
   function setUp() {
     parent::setUp();
-    $this->admin_user = $this->backdropCreateUser(array('administer permissions', 'administer users'));
+    $this->admin_user = $this->backdropCreateUser(array('administer permissions', 'administer users', 'assign roles'));
   }
 
   /**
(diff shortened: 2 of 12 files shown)


core/modules/user/user.module
```diff
@@ -2503,7 +2505,7 @@ function user_role_load_by_name($role_name) {
  *
  * @param $role
  *   A role object to modify or add.
- * @return
+ * @return int
```
klonos, line 2519:
```suggestion
 * @return int | bool
```

core/modules/user/user.theme.inc
```diff
@@ -107,7 +107,12 @@ function theme_user_admin_permissions($variables) {
     $row = array();
     // Module name
     if (is_numeric($key)) {
-      $row[] = array('data' => backdrop_render($form['permission'][$key]), 'class' => array('module'), 'id' => 'module-' . $form['permission'][$key]['#id'], 'colspan' => count($form['roles']['#value']) + 1);
+      $row[] = array(
+        'data' => backdrop_render($form['permission'][$key]),
+        'class' => array('module'),
+        'id' => 'module-' . $form['permission'][$key]['#id'],
+        'colspan' => count($form['roles']['#value']) + 1,
+      );
```
klonos, line 115: nice 👍🏼

core/modules/user/user.theme.inc
```diff
@@ -125,7 +130,11 @@ function theme_user_admin_permissions($variables) {
   }
   $header[] = (t('Permission'));
   foreach (element_children($form['role_names']) as $role_name) {
-    $header[] = array('data' => backdrop_render($form['role_names'][$role_name]), 'class' => array('checkbox'));
+    $header[] = array(
+      'data' => backdrop_render($form['role_names'][$role_name]),
+      'title' => check_plain(strip_tags($form['role_names'][$role_name]['#description'])),
+      'class' => array('checkbox')
```
klonos:
```suggestion
      'class' => array('checkbox'),
```

core/modules/user/user.theme.inc
```diff
@@ -179,20 +188,37 @@ function theme_user_permission_description($variables) {
 function theme_user_admin_roles($variables) {
   $form = $variables['form'];
 
-  $header = array(t('Name'), t('Weight'), t('Operations'));
+  $header = array(
+    array('data' => t('Name'), 'class' => array('role-name')),
+    array('data' => t('Description'),'class' => array('role-description', 'priority-low')),
```
klonos:
```suggestion
    array('data' => t('Description'), 'class' => array('role-description', 'priority-low')),
```

core/modules/user/user.theme.inc
```diff
@@ -179,20 +188,37 @@ function theme_user_permission_description($variables) {
...
-    $row[] = backdrop_render($form['roles'][$role_name]['operations']);
+    $row[] = array(
+      'data' => theme('label_machine_name__role', array(
+        'label' => $label,
+        'machine_name' => $form['roles'][$role_name]['#role']->name,
+      )),
+      'class' => array('role-name'),
+    );
+    $row[] = array(
+      'data' => filter_xss_admin(t($form['roles'][$role_name]['#role']->description)),
```
klonos: We don't seem to be using `filter_xss_admin()`/`filter_xss()` for the label and machine name here. Should we? ...or should we be handling all this on form save instead? The only way for an attack vector here seems if someone manually edited the values of `user.role.*.json` config files.

core/modules/user/user.theme.inc
```diff
@@ -125,7 +130,11 @@ function theme_user_admin_permissions($variables) {
   }
   $header[] = (t('Permission'));
   foreach (element_children($form['role_names']) as $role_name) {
-    $header[] = array('data' => backdrop_render($form['role_names'][$role_name]), 'class' => array('checkbox'));
+    $header[] = array(
+      'data' => backdrop_render($form['role_names'][$role_name]),
+      'title' => check_plain(strip_tags($form['role_names'][$role_name]['#description'])),
```
klonos: While testing on the PR sandbox, this line here throws this:

```
Notice: Undefined index: #description in theme_user_admin_permissions()
(line 135 of /var/lib/tugboat/core/modules/user/user.theme.inc).
```

Got that when tried to edit a content type.

quicksketch: Thanks, fixed in the new commits!

core/modules/user/user.theme.inc
quicksketch: Looks like theme_label_machine_name() is what runs the check_plain() in this case, so we don't need to escape it here.

core/modules/user/user.module
quicksketch: Turns out this only ever returns an integer. If it fails, a fatal error is thrown.

core/modules/user/tests/user.test
```diff
@@ -2204,9 +2204,17 @@ class UserRoleAdminTestCase extends BackdropWebTestCase {
 
     // Test adding a role using the 'Save role' button.
     $role1_name = '123';
-    $edit = array('name' => $role1_name, 'label' => $role1_name);
+    $role1_admin_description = 'Admin description for 123 role.';
+    $role1_description = 'Help text for 456 role.';
```
klonos:
```suggestion
    $role1_description = 'Help text for 123 role.';
```

core/modules/user/config/user.role.authenticated.json
```diff
@@ -2,6 +2,8 @@
     "_config_name": "user.role.authenticated",
     "name": "authenticated",
     "label": "Authenticated",
+    "admin_description": "Special role assigned to all accounts. Permissions assigned to this role are given to all other roles except the anonymous role.",
+    "description": "Automatically assigned role to all accounts.",
```
klonos: I would either lose the word "role" altogether (it is mentioned in the field label when assigning roles, so it's redundant), or move it to the beginning of the sentence.

quicksketch:
Merged into 1.x for 1.24.0. Thanks @klonos for the reviews!

core/profiles/standard/standard.install
```diff
@@ -559,7 +561,9 @@ function standard_install() {
   // permissions assigned.
   $admin_role = new stdClass();
   $admin_role->name = 'administrator';
-  $admin_role->label = st('Administrator');
+  $admin_role->label = 'Administrator';
+  $admin_role->admin_description = 'This role receives all permissions on the site.';
```
jenlampton, line 565: We need to remove this. I'll file a follow-up.

PR #4273, opened 2022-12-09
Issue #5885: Auto-assign permissions for default editor role

### LINKED ISSUE #5885: [UX] When a new content type is created, automatically assign permissions to the Editor role.
## Description of the need

This is a sibling issue to #5350, and this is a related issue against Drupal 7 core: [Dynamic permissions cannot be automatically assigned to or removed from roles.](https://www.drupal.org/project/drupal/issues/787152)

The 3 content types that come OOTB with a default Backdrop installation all have these permissions already granted to the Editor role, but that is not the case when creating new content types:
image

When creating a new content type, these permissions need to be manually adjusted each time. With #900, we have improved the UX in Backdrop, since you don't need to navigate away from the content type creation/edit form in order to assign the required permissions to the various roles, nor do you have to search through a HUGE list of permissions to find the correct ones 👍🏼 ...this issue here is to improve the UX a big further, by allowing some sensible defaults (that way, site builders will have to scroll/click less when configuring content types).

## Proposed solution

- Automatically grant these permissions to the Editor role.
- Perhaps add a setting for this in `admin/config/people/roles`, similar to the "The role that will be automatically assigned new permissions whenever a module is enabled:" setting for admin roles.
- Also throw a notice/info message in the node creation form, letting the user know that automatic permissions assignment has happened, and that they might want to review/tweak them as needed.

---
Advocate for this feature: @klonos 
---

Related issues
* https://github.com/backdrop/backdrop-issues/issues/5031 
* https://github.com/backdrop/backdrop-issues/issues/1285
* https://github.com/backdrop/backdrop-issues/issues/6078

klonos:
Fixes https://github.com/backdrop/backdrop-issues/issues/5885


core/modules/user/user.admin.inc
```diff
@@ -550,29 +550,37 @@ function user_admin_roles($form, $form_state) {
...
+  // Do not allow the anonymous or authenticated roles to be set as the default
+  // administrator or content editor roles.
   unset($roles[BACKDROP_ANONYMOUS_ROLE]);
   unset($roles[BACKDROP_AUTHENTICATED_ROLE]);
   // Add the "disabled" option to the top of the roles list (makes sure that
-  // this is the selected option if the role that was previously set as admin is
-  // deleted).
+  // this is the selected option if the role that was previously set as admin or
+  // content editor is deleted).
   $roles = array(0 => t('disabled')) + $roles;
```
quicksketch: I realize this is an existing string, but I think we should change this to either `- None-` to match other disabled options, or if we want to keep this word, change to `- Disabled -`.

```suggestion
  $roles = array(0 => t('- None -')) + $roles;
```

core/modules/node/node.types.inc
```diff
@@ -130,6 +130,92 @@ function node_type_form($form, &$form_state, $type = NULL) {
...
+    // page, which would make it unnecessarily prominent and potentially
+    // distracting. We are placing it right above the vertical tabs section
+    // instead (closer to where permissions can be configured). Intentionally
+    // not placing it inside the "Permissions" tab either, since that is not the
+    // default tab, so it wouldn't be visible, and we need to show some warnings
+    // to people even if they don't have access to the "Permissions" tab.
+    $form['permissions_message_container'] = array(
+      '#type' => 'container',
+      '#attributes' => array('class' => array('messages', $permissions_message_type)),
+    );
```
quicksketch, line 212: It would be a nice follow-up to establish a pattern for this, or possibly a new element type similar to `#type => help` (maybe `#type => messages`)? I think we only do this in one or two other places in all of core.

core/modules/node/node.types.inc
```diff
@@ -130,6 +130,92 @@ function node_type_form($form, &$form_state, $type = NULL) {
...
+    $permissions_message_type = 'info';
+  }
+
+  if (isset($permissions_message)) {
+    if (user_access('administer permissions')) {
+      $permissions_message .= ' ' . t('You can make any necessary adjustments in the <em>Permissions</em> tab below.');
+    }
+    else {
+      $user_module_permissions = module_invoke('user', 'permission');
+      $permissions_message .= ' ' . t('Only roles with the %permission permission can make any necessary adjustments.', array('%permission' => $user_module_permissions['administer permissions']['title']));
```
quicksketch, line 198: Also a follow-up issue, there should be a way to get a permission label.

core/modules/node/node.types.inc
klonos: We already have https://github.com/backdrop/backdrop-issues/issues/3329 @quicksketch, which is on my radar 👀

core/modules/node/node.types.inc
klonos: Done: https://github.com/backdrop/backdrop-issues/issues/6097

core/modules/user/tests/user.test
```diff
@@ -1400,11 +1405,80 @@ class UserPermissionsTestCase extends BackdropWebTestCase {
...
+
+    // Now check permissions were correctly assigned.
+    $user_accounts = array(
+      $this->admin_user,
+      $this->editor_user,
+    );
+    foreach ($content_type_permissions as $type_permission) {
+      $permission_name = backdrop_strtolower(str_replace('content', "$content_type_name content", $type_permission));
+      foreach ($user_accounts as $user_account) {
+        user_access($permission_name, $user_account);
```
klonos: This should be an assertion @quicksketch. Right? ...something like this I believe:
```suggestion
        $this->assertTrue(user_access($permission_name, $user_account));
```

quicksketch: Ah, yes, you're right!

quicksketch:
Merged into 1.x for 1.25.0. Yay! Last PR before feature freeze.

core/modules/user/tests/user.test
```diff
@@ -1400,11 +1405,82 @@ class UserPermissionsTestCase extends BackdropWebTestCase {
...
+    // Opposite of the above assertNoFieldByXPath(), confirm that there are at
+    // least some permission checkboxes checked.
+    $this->assertFieldByXPath("//table[@id='permissions']//input[@checked]");
+
+    $content_type_permissions = array(
+      'create content',
+      'edit own content',
+      'edit any content',
+      'delete own content',
+      'delete any content',
```
jenlampton, line 1442: looks like we forgot to remove this.

PR #4274, opened 2022-12-10
Issue 5886. Avoid adding the base path twice to ckeditor links

### LINKED ISSUE #5886: CKEditor Link Continuously Adds Base URL
After editing a link using the CKEditor link icon, the Base URL string is added (/smbmc) - each time when it is saved.  After the second save this produces a broken link.

To reproduce the behavior:

1. Edit a Page with a working link using CKEditor - Full HTML
2. Click on the link to open the dialog box but make no changes
3. Click on "Save" in the dialog box
4. Click on "Save" in CKEditor
5. Clicking on the link on the page now gives an error

## Actual behavior

Link Dialog after 1 save:

working_link

Link Dialog after 2nd Save (no editing):

broken_link

## Expected behavior

Either 1) BASE URL is never added

or

2) no BASE URL added if it's already present

## Additional information

- Backdrop CMS version: 1.23.1 
- Web server/Operating: Apache/2.4.6 (CentOS) OpenSSL/1.0.2k-fips PHP/7.4.27 
- Browser(s) and their versions: Firefox 102.4.0esr (64-bit)

argiepiano:
Fixes https://github.com/backdrop/backdrop-issues/issues/5886

### DIFF
--- a/core/modules/filter/filter.pages.inc
+++ b/core/modules/filter/filter.pages.inc
@@ -461,9 +461,7 @@ function filter_format_editor_link_form_submit($form, &$form_state) {
     }
   }
   else {
-    // Modify references to other URIs if not an external link.
-    $url = (substr($uri, 0, 4) == 'http') ? $uri : substr(base_path(), 0, -1) . $uri;
-    $form_state['values']['attributes']['href'] = $url;
+    $form_state['values']['attributes']['href'] = $uri;
     // Remove any reference to a data file.
     unset($form_state['values']['fid']);
     $form_state['values']['attributes']['data-file-id'] = NULL;


core/modules/filter/filter.pages.inc
```diff
@@ -461,9 +461,7 @@ function filter_format_editor_link_form_submit($form, &$form_state) {
     }
   }
   else {
-    // Modify references to other URIs if not an external link.
-    $url = (substr($uri, 0, 4) == 'http') ? $uri : substr(base_path(), 0, -1) . $uri;
-    $form_state['values']['attributes']['href'] = $url;
+    $form_state['values']['attributes']['href'] = $uri;
```
jenlampton, line 464: Rather than never adding the base path, we should check to see if the person (or autocpmplete) providing the link included it, or not, and add it only if it is not already present.

argiepiano: I'm not sure I understand. The base path is always added. I took this out because this adds it a second time.   

Please read the [comment (and subsequent)](https://github.com/backdrop/backdrop-issues/issues/5886#issuecomment-1345147356) that explain this in the main conversation.

laryn:
Thanks all! Merged into `1.x` and `1.26.x`.

PR #4275, opened 2022-12-12
Issue #4097: Introduce a new user_permission_get_info() function

### LINKED ISSUE #4097: [DX] Introduce a new `user_permission_get_info()` function
Modules can specify permissions via [`hook_permission()`](https://api.backdropcms.org/api/backdrop/1/search/hook_permission):

```php
function my_awesome_module_permission() {
  return array(
    'permission x' => array(
      'title' => t('Permission to do x'),
      'description' => t('Allows people to do x.'),
      'restrict access' => TRUE,
      'warning' => t('Allows people to do ...'),
    ),
    'permission y' => array(
      'title' => t('Permission to do y'),
      'description' => t('Allows people to do y.'),
    ),
  );
}
```

Now, we have [`user_access()`](https://api.backdropcms.org/api/backdrop/1/search/user_access), which allows us to determine if a user has a specific permission:

```php
$user_can_do_x => user_access('permission x');
$user_can_do_y => user_access('permission y');
```

...and we also have [`user_permission_get_modules()`](https://api.backdropcms.org/api/backdrop/1/search/user_permission_get_modules), which returns all permissions, along with which module implements each permission.

...but we don't have any function that returns the properties of a permission. So each time we need to reference a permission say in some help text, we do this:

```php
$form['whatever'] = array(
  ...
  '#description' => t('This option depends on the "Permission to do y" permission.',
  ...
);
```

This is OK, but if at some point we rename the permission (change it's `'title'`), then we need to do a search and replace (at best), in order to find all instances of the literal permission title in the codebase, and update it to the new name/label. Ideally though, we should be able to do something like this:

```php
$permission_info = user_permission_get_info('permission x');
$form['whatever'] = array(
  ...
  '#description' => t('This option depends on the @permission permission.', array('@permission' => $permission_info['title'],
  ...
);
```

That way, we would only need to update any permission info in a single place (in its `hook_permission()` function), which also makes translators' life easier.

As a developer, at minimum I would like this new function to return:

- the title of the permission
- the description of the permission

...additional nice-to-have's:

- an admin path, which can be used to provide a link to `/admin/config/people/permissions#permission x` (scrolls the permissions listing page to the specific permission)
- the module that specifies the hook of the permission (alternatively, have `user_permission_get_modules()` accept an optional `$permission` parameter, so to be able to do `user_permission_get_modules('permission x')`). This could also be used as a search/filter query when #980 gets implemented 😉 
- a list of all user roles this permission has been assigned to (we have [`user_roles()`](https://api.backdropcms.org/api/backdrop/1/search/user_roles) for that - still nice to have)

klonos:
Fixes https://github.com/backdrop/backdrop-issues/issues/4097

### DIFF
--- a/core/modules/user/user.module
+++ b/core/modules/user/user.module
@@ -2598,18 +2598,132 @@ function user_role_delete_access($role) {
 /**
  * Determine the modules that permissions belong to.
  *
- * @return
+ * @return array
  *   An associative array in the format $permission => $module.
  */
 function user_permission_get_modules() {
-  $permissions = array();
+  return user_permission_get_info(NULL, 'module');
+}
+
+/**
+ * Determine the module providing a specific permission.
+ *
+ * @param string $permission_name
+ *   The machine name of the target permission.
+ *
+ * @return string
+ *   The machine name of the module the permission belongs to.
+ *
+ * @since 1.24.0
+ */
+function user_permission_get_module($permission_name) {
+  return user_permission_get_info($permission_name, 'module');
+}
+
+/**
+ * Get the titles (human-readable names) of all permissions.
+ *
+ * @return array
+ *   An associative array in the format $permission => $title.
+ *
+ * @since 1.24.0
+ */
+function user_permission_get_titles() {
+  return user_permission_get_info(NULL, 'title');
+}
+
+/**
+ * Get the title (human-readable name) of a specific permission.
+ *
+ * @param string $permission_name
+ *   The machine name of the target permission.
+ *
+ * @return string
+ *   The human-readable title of the permission.
+ *
+ * @since 1.24.0
+ */
+function user_permission_get_title($permission_name) {
+  return user_permission_get_info($permission_name, 'title');
+}
+
+/**
+ * Retrieve information about a given permission or all permissions.
+ *
+ * @param string $permission_name
+ *   (optional) The name of a speciff permission for which to return
+ *   information. If ommitted, an array of all permissions found will be
+ *   returned.
+ * @param string $property
+ *   (optional) The name of a specific property to be retrieved. Can be:
+ *   - 'title' (string)
+ *   - 'description' (string)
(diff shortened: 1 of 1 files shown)


core/modules/user/user.module
```diff
@@ -2598,18 +2598,134 @@ function user_role_delete_access($role) {
...
+      if ($property == NULL) {
+        $all_permissions[$permission] = $permission_info;
+      }
+      // Specific permission property requested.
+      elseif (array_key_exists($property, $permission_info)) {
+        $all_permissions[$permission] = $permission_info[$property];
+      }
+
+      // Target permission found - no need to keep looking.
+      if ($permission_name != NULL && $permission == $permission_name) {
```
docwilmot, line 2698: This could be just `if ($permission == $permission_name)` since `$permission` will never be NULL.

core/modules/user/user.module
```diff
@@ -2598,18 +2598,134 @@ function user_role_delete_access($role) {
...
+  }
+  else {
+    // Requested permission not found.
+    if (!array_key_exists($permission_name, $all_permissions)) {
+      return FALSE;
+    }
+
+    // Specific property of a permission requested.
+    if ($property != NULL) {
+      if (array_key_exists($property, $permission_info)) {
```
docwilmot, line 2717: `$permission_info` isn't defined outside the `foreach()` loop above. Should this be `array_key_exists($property, $all_permissions[$permission_name]`? And the return below `$all_permissions[$permission_name][$property]`?

core/modules/user/user.module
```diff
@@ -2598,18 +2598,134 @@ function user_role_delete_access($role) {
...
+        return $permission_info[$property];
+      }
+      // Requested property not found.
+      return FALSE;
+    }
+    // No specific property requested - return all permission properties as an
+    // array.
+    else {
+      return $permission_info;
```
docwilmot, line 2726: See above.

PR #4282, opened 2022-12-17
Issue #5722: Update the links pointing to php.net

### LINKED ISSUE #5722: Update the links pointing to php.net
The links pointing to php.net used in comments and in the user interface should be changed to:

### Links without sub-domain (http://php.net)
- Use https://
- Use www as sub-domain
- Don't have any language selector

### Links with sub-domain (for example, http://bugs.php.net)
- Use https://

Removing the language selector in the link allows users to see the page in the language they selected. Providing the language selector isn't necessary as, for example, http://www.php.net/manual/en/function.rawurlencode.php#86506 and http://www.php.net/manual/fr/function.rawurlencode.php#86506 will both show the same comment.

avpaderno:
Fixes https://github.com/backdrop/backdrop-issues/issues/5722


core/modules/system/system.tar.inc
```diff
@@ -36,14 +36,14 @@
...
  */
 
 /**
  * Backdrop-specific notes on this port from the original source.
  *
  * This file origin is Tar.php, release 1.4.9 (stable) with some code
- * from PEAR.php, release 1.10.10 (stable) both at http://pear.php.net.
+ * from PEAR.php, release 1.10.10 (stable) both at http://pear.php.net/.
  * To simplify future porting from pear of this file, you should not
  * do cosmetic or other non significant changes to this file.
```
indigoxela: I think, this file shouldn't get changed at all - the two lines explain it: "you should not do consmetic or other...".

core/includes/unicode.inc
```diff
@@ -110,15 +110,15 @@ function _unicode_check() {
 
   // Check for mbstring extension
   if (!function_exists('mb_strlen')) {
-    return array(UNICODE_SINGLEBYTE, $t('Operations on Unicode strings are emulated on a best-effort basis. Install the <a href="@url">PHP mbstring extension</a> for improved Unicode support.', array('@url' => 'http://www.php.net/mbstring')));
+    return array(UNICODE_SINGLEBYTE, $t('Operations on Unicode strings are emulated on a best-effort basis. Install the <a href="@url">PHP mbstring extension</a> for improved Unicode support.', array('@url' => 'https://www.php.net/mbstring')));
```
indigoxela: @kiamlaluno if you run phpcs locally with B coding standards - isn't this line and the two following changes violations? The line is longer than 80 chars and in this case you should use line breaks in the array.

core/modules/date/date.admin.inc
```diff
@@ -207,7 +207,7 @@ function _date_field_instance_settings_form($field, $instance) {
     '#fieldset' => 'default_values',
   );
 
-  $description = t("Describe a time by reference to the current day, like '+90 days' (90 days from the day the field is created) or '+1 Saturday' (the next Saturday). See !strtotime for more details.", array('!strtotime' => l(t('strtotime'), 'http://www.php.net/manual/en/function.strtotime.php')));
+  $description = t("Describe a time by reference to the current day, like '+90 days' (90 days from the day the field is created) or '+1 Saturday' (the next Saturday). See !strtotime for more details.", array('!strtotime' => l(t('strtotime'), 'https://www.php.net/manual/function.strtotime.php')));
```
indigoxela: Also try line breaks for the array here.

core/modules/date/views/date_views_argument_handler_simple.inc
```diff
@@ -175,7 +175,7 @@ class date_views_argument_handler_simple extends views_handler_argument_date {
       '#type' => 'textfield',
       '#title' => t('Custom summary date format'),
       '#default_value' => $this->options['title_format_custom'],
-      '#description' => t("A custom format for the title and summary date format. Define a php date format string like 'm-d-Y H:i' (see <a href=\"@link\">http://php.net/date</a> for more details).", array('@link' => 'http://php.net/date')),
+      '#description' => t("A custom format for the title and summary date format. Define a php date format string like 'm-d-Y H:i' (see <a href=\"@link\">https://www.php.net/date</a> for more details).", array('@link' => 'https://www.php.net/date')),
```
indigoxela: Another item with long line and array...

core/modules/dblog/dblog.module
```diff
@@ -168,7 +168,7 @@ function dblog_form_system_logging_settings_alter(&$form, $form_state) {
     '#type' => 'textfield',
     '#title' => t('Date format for log messages'),
     '#maxlength' => 100,
-    '#description' => t('A date format using PHP date and time codes. See the <a href="@url">PHP manual</a> for available options.', array('@url' => 'http://php.net/manual/function.date.php')),
+    '#description' => t('A date format using PHP date and time codes. See the <a href="@url">PHP manual</a> for available options.', array('@url' => 'https://www.php.net/manual/function.date.php')),
```
indigoxela: Another long line with array...

core/modules/system/system.admin.inc
```diff
@@ -1900,7 +1900,7 @@ function system_image_toolkit_settings() {
   if (count($toolkits_available) == 0) {
     config_set('system.core', 'image_toolkit', '');
     $form['image_toolkit_help'] = array(
-      '#markup' => t("No image toolkits were detected. Backdrop includes support for <a href='!gd-link'>PHP's built-in image processing functions</a> but they were not detected on this system. You should enable them, or try using a third party toolkit.", array('gd-link' => url('http://php.net/gd'))),
+      '#markup' => t("No image toolkits were detected. Backdrop includes support for <a href='!gd-link'>PHP's built-in image processing functions</a> but they were not detected on this system. You should enable them, or try using a third party toolkit.", array('gd-link' => url('https://www.php.net/gd'))),
```
indigoxela: Another long line + array...

core/includes/unicode.inc
avpaderno: I didn't touch issues reported from `phpcs` because there are many warnings and errors in the 34 changed files. There are more than 150 warnings and errors for a single file.

I think it's probably better to first understand why actions throw a so generic error that doesn't even report what exactly is the issue, before starting working on all the issues reported by `phpcs`.

core/modules/date/date.admin.inc
avpaderno: I changed the code as suggested.

core/modules/dblog/dblog.module
avpaderno: I changed the code as suggested.

core/modules/system/system.admin.inc
avpaderno: I changed the code as suggested.

core/includes/unicode.inc
```diff
@@ -110,15 +110,18 @@ function _unicode_check() {
 
   // Check for mbstring extension
   if (!function_exists('mb_strlen')) {
-    return array(UNICODE_SINGLEBYTE, $t('Operations on Unicode strings are emulated on a best-effort basis. Install the <a href="@url">PHP mbstring extension</a> for improved Unicode support.', array('@url' => 'http://www.php.net/mbstring')));
+    return array(UNICODE_SINGLEBYTE, $t(
+      'Operations on Unicode strings are emulated on a best-effort basis. Install the <a href="@url">PHP mbstring extension</a> for improved Unicode support.',
+      array('@url' => 'https://www.php.net/mbstring')),
+    );
```
indigoxela: I'd suggest different line breaks. An example...
```php
return array(UNICODE_SINGLEBYTE, $t('Operations.... and so on ...', array(
  '@url' => 'http://www.php.net/mbstring',
)));
```
So "array(" still on the same line, the values - one per line - indented, ending with a comma per line. No newline *inside the function*, but only around array items. Does that help?

avpaderno: I changed the indentation too.

core/modules/system/system.admin.inc
indigoxela: No line breaks, though.

core/modules/simpletest/simpletest.install
```diff
@@ -27,7 +27,7 @@ function simpletest_requirements($phase) {
   );
   if (!$has_curl) {
     $requirements['curl']['severity'] = REQUIREMENT_ERROR;
-    $requirements['curl']['description'] = $t('The testing framework could not be installed because the PHP <a href="@curl_url">cURL</a> library is not available.', array('@curl_url' => 'http://php.net/manual/en/curl.setup.php'));
+    $requirements['curl']['description'] = $t('The testing framework could not be installed because the PHP <a href="@curl_url">cURL</a> library is not available.', array('@curl_url' => 'https://www.php.net/manual/curl.setup.php'));
```
indigoxela: Still no line breaks in here and below.

core/modules/image/image.install
```diff
@@ -90,14 +90,14 @@ function image_requirements($phase) {
...
+        $requirements['image_gd']['description'] = t('The GD Library for PHP is enabled, but was compiled without support for functions used by the rotate and desaturate effects. It was probably compiled using the official GD libraries from http://www.libgd.org instead of the GD library bundled with PHP. You should recompile PHP --with-gd using the bundled GD library. See <a href="https://www.php.net/manual/book.image.php">the PHP manual</a>.');
       }
     }
     else {
       $requirements['image_gd'] = array(
         'value' => t('Not installed'),
         'severity' => REQUIREMENT_ERROR,
-        'description' => t('The GD library for PHP is missing or outdated. Check the <a href="@url">PHP image documentation</a> for information on how to correct this.', array('@url' => 'http://www.php.net/manual/book.image.php')),
+        'description' => t('The GD library for PHP is missing or outdated. Check the <a href="@url">PHP image documentation</a> for information on how to correct this.', array('@url' => 'https://www.php.net/manual/book.image.php')),
```
indigoxela: Line breaks... :smile_cat: I know, this might get frustrating, so thanks again for your effort and patience.

avpaderno: Essentially, you are suggesting to change where an array definition starts in every line where I changed a php.net url, aren't you?

indigoxela: Yes, essentially that. :+1: 

The automated action fails to annotate, but I'd like you to fix all coding standard issues in your PR - in code you touched, so we can see, if that makes the check pass.

core/modules/date/views/date_views_argument_handler_simple.inc
```diff
@@ -175,7 +175,10 @@ class date_views_argument_handler_simple extends views_handler_argument_date {
       '#type' => 'textfield',
       '#title' => t('Custom summary date format'),
       '#default_value' => $this->options['title_format_custom'],
-      '#description' => t("A custom format for the title and summary date format. Define a php date format string like 'm-d-Y H:i' (see <a href=\"@link\">http://php.net/date</a> for more details).", array('@link' => 'http://php.net/date')),
+      '#description' => t(
+        "A custom format for the title and summary date format. Define a php date format string like 'm-d-Y H:i' (see <a href=\"@link\">https://www.php.net/date</a> for more details).",
+        array('@link' => 'https://www.php.net/date')
```
indigoxela: The line break right after t( isn't actually correct.

Example:
```
'#description' => t("A custom format for the...", array(
  '@link' => 'https://www.php.net/date',
)),
```

core/modules/system/system.tokens.inc
```diff
@@ -189,14 +189,16 @@ function system_token_info() {
   foreach ($date_format_types as $machine_name => $date_format_type_info) {
     if (!isset($date_format_type_info['hidden']) || $date_format_type_info['hidden'] == FALSE) {
       $date[$machine_name] = array(
-         'name' => check_plain($date_format_type_info['label']),
-         'description' => t("A date in '@type' format. (%date)", array('@type' => $machine_name, '%date' => format_date(REQUEST_TIME, $machine_name))),
-       );
+        'name' => check_plain($date_format_type_info['label']),
+        'description' => t("A date in '@type' format. (%date)", array('@type' => $machine_name, '%date' => format_date(REQUEST_TIME, $machine_name))),
```
indigoxela: This change is actually unrelated. Either fix it completely, or leave it untouched.

core/includes/unicode.inc
indigoxela: Two more thingies, and we should be able to see, if phpcs / the action passes and the problem is just with annotations (so, the action), or something underneath is broken (phpcs / the rule(s)). That's important to figure out, where to report / fix.

indigoxela:
So, now we have a totally different problem: `✗ phpcs 403: rate limit exceeded` - we didn't get phpcs...

That tool gets pulled via shivammathur/setup-php - another action we're using. :shrug:

avpaderno:
I guess that *rate limit exceeded* means the action has been used too often or too much.

core/modules/date/date.admin.inc
```diff
@@ -207,7 +207,10 @@ function _date_field_instance_settings_form($field, $instance) {
     '#fieldset' => 'default_values',
   );
 
-  $description = t("Describe a time by reference to the current day, like '+90 days' (90 days from the day the field is created) or '+1 Saturday' (the next Saturday). See !strtotime for more details.", array('!strtotime' => l(t('strtotime'), 'http://www.php.net/manual/en/function.strtotime.php')));
+  $description = t(
+    "Describe a time by reference to the current day, like '+90 days' (90 days from the day the field is created) or '+1 Saturday' (the next Saturday). See !strtotime for more details.",
```
indigoxela: Still wrong usage of line breaks here - please use them around array items, not inside the function.
```
  $description = t( "Describe a time by...", array(
```
No line break right after `t(`, but after `array(`.

avpaderno: I fixed how the function calls changed by previous commits are formatted.

PR #4283, opened 2022-12-17
Document that node_load_multiple(FALSE) returns all the nodes

### LINKED ISSUE #5889: Document that node_load_multiple(FALSE) returns all the nodes
As reported in #5437, calling [`node_load_multiple()`](https://docs.backdropcms.org/api/backdrop/core%21modules%21node%21node.module/function/node_load_multiple/1) with no arguments returns an empty array, while [`entity_load_multiple('node')`](https://docs.backdropcms.org/api/backdrop/core%21modules%21entity%21entity.module/function/entity_load_multiple/1), the function called by `node_load_multiple()`) returns all nodes as expected.

This happens because the default value for `entity_load_multiple()`'s `$ids` is `FALSE`, while the default value for `node_load_multiple()`'s `$nids` is `array()`.

`$nids` should be documented to accept `FALSE` as value, which is the value used to return all nodes.

avpaderno:
Fixes https://github.com/backdrop/backdrop-issues/issues/5889


core/modules/node/node.module
```diff
@@ -856,15 +856,15 @@ function node_invoke($node, $hook, $a2 = NULL, $a3 = NULL, $a4 = NULL) {
  * from the database. Nodes are loaded into memory and will not require database
  * access if loaded again during the same page request.
  *
- * @param $nids
- *   (optional) An array of node IDs.
- * @param $conditions
+ * @param array|false $nids
```
argiepiano, line 859: Core seems to use `array|FALSE` instead of `array|false`. I couldn't fine any instance of the second.

avpaderno: [Indicating data types in documentation](https://docs.backdropcms.org/doc-standards#types) in the coding standards says it is `false` not `FALSE`. I think that is the same type-hinting used in PHP documentation.

argiepiano: Got it.

avpaderno: I take that new code should follow what reported in the coding standards. Is it preferable to keep the code consistent even if it does not follow the coding standards? (I am asking just to understand what to do in future PRs.)

argiepiano: I think it's better to follow the coding standards, and perhaps open a separate issue to correct the rest of the `FALSE` instances in core separately.

ghost: @kiamlaluno Yep, what @argiepiano said. Always follow the coding standards for new/updated code. Then there'll be less issues to fix later.

core/modules/node/node.module
```diff
@@ -856,15 +856,15 @@ function node_invoke($node, $hook, $a2 = NULL, $a3 = NULL, $a4 = NULL) {
...
+ *   (optional) An array of node IDs or FALSE to load all the nodes.
```
ghost: It might also be worth noting what an empty array does... I believe @argiepiano's PR addressed this.

avpaderno: When an empty array is passed as first argument to [`node_load_multiple()`](https://docs.backdropcms.org/api/backdrop/core%21modules%21node%21node.module/function/node_load_multiple/1), the function returns an empty array, not `NULL`. At least, that is what I understand from [`DefaultEntityController::load()`](https://docs.backdropcms.org/api/backdrop/core%21modules%21entity%21entity.controller.inc/function/DefaultEntityController%3A%3Aload/1), where the returned value (`$entities`) is initialized to an empty array.  
If that is the case, it should not need to be documented, since it is the expected behavior.  
Is that correct?

avpaderno: The return value for `node_load_multiple()` is type-hinted with `Node[]`. Does that include an empty array as I am thinking it does?

argiepiano: Yes, the return value is an array and, technically, doesn't need to be documented (it's covered by `Node[]`). Still, I think it would be helpful to mention it. I've been "bitten" by this behavior: the default value for the 2nd parameter in `entity_load_multiple('node')` returns all nodes, but the default value for `node_load_multiple()` returns an empty array. It's a bit inconsistent.

But if you feel strongly about not mentioning it, that's fine.

avpaderno: @argiepiano I am not contrary to mention that. I was trying to understand what the exact `node_load_multiple()` behavior is (in the case I misunderstood it) and what exactly should be documented (which is what your last comment made me understand).

quicksketch:

PR #4285, opened 2022-12-18
Passing null to parameter of type string is deprecated

### LINKED ISSUE #5890: Deprecated function: mb_strlen(): Passing null to parameter #1 ($string) of type string is deprecated in mb_strlen() (line 479 of ../unicode.inc)
## Description of the bug

In certain conditions the deprecated function error in subject is triggered. 

## Steps To Reproduce

Not clear how to reproduce this - I'll update if it comes up again - but the last time I caught this was a combination of a custom module being ported and Devel module. In particular, the error was caused by a call from `/var/www/html/docroot/modules/devel/devel.module line 1232`  which is the` krumo($object);` and the whole function is:

```
function krumo_ob($object) {
  ob_start();
  krumo($object);
  $output = ob_get_contents();
  ob_end_clean();
  return $output;
}
```

The [line](https://github.com/backdrop/backdrop/blob/c27cd94ff25f633d5a9903abf0085f6f1c0724f6/core/includes/unicode.inc#L479) in the subject is `return mb_strlen($text);` and the whole function is:

```
function backdrop_strlen($text) {
  global $multibyte;
  if ($multibyte == UNICODE_MULTIBYTE) {
    return mb_strlen($text);
  }
  else {
    // Do not count UTF-8 continuation bytes.
    return strlen(preg_replace("/[\x80-\xBF]/", '', $text));
  }
}
```

## Additional information

This kind of errors started to flow in with projects switching to PHP 8.1. I believe the same error was reported on https://www.drupal.org/project/feeds/issues/3266703 with the patch offered on https://www.drupal.org/files/issues/2022-02-26/core_unicode_deprecated.patch, but the issue itself being marked as duplicate of https://www.drupal.org/project/feeds/issues/3266704

Wanted to offer a PR using the null coalescing operator as suggested on https://stackoverflow.com/questions/71707325/migration-to-php-8-1-how-to-fix-deprecated-passing-null-to-parameter-error-r, but then because of potential comparability issues decided to hear from the community on what would be the best way of addressing this issue.

avpaderno:
Fixes https://github.com/backdrop/backdrop-issues/issues/5890

### DIFF
--- a/core/includes/unicode.inc
+++ b/core/includes/unicode.inc
@@ -474,8 +474,9 @@ function decode_entities($text) {
  * @ingroup php_wrappers
  */
 function backdrop_strlen($text) {
-  // Bail early if no string provided (if $text is NULL for example).
-  if (!is_string($text)) {
+  // Bail early if $text is NULL.
+  // @todo Remove this in 2.x.
+  if (is_null($text)) {
     return 0;
   }
 
@@ -492,7 +493,7 @@ function backdrop_strlen($text) {
 /**
  * Uppercase a UTF-8 string.
  *
- * @param $text
+ * @param string $text
  *   The string to run the operation on.
  *
  * @return string
@@ -501,6 +502,12 @@ function backdrop_strlen($text) {
  * @ingroup php_wrappers
  */
 function backdrop_strtoupper($text) {
+  // Bail early if $text is NULL.
+  // @todo Remove this in 2.x.
+  if (is_null($text) {
+    return "";
+  }
+
   global $multibyte;
   if ($multibyte == UNICODE_MULTIBYTE) {
     return mb_strtoupper($text);
@@ -517,7 +524,7 @@ function backdrop_strtoupper($text) {
 /**
  * Lowercase a UTF-8 string.
  *
- * @param $text
+ * @param string $text
  *   The string to run the operation on.
  *
  * @return string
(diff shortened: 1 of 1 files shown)


klonos (approved):
LGTM 👍🏼

core/includes/unicode.inc
```diff
@@ -475,6 +475,9 @@ function decode_entities($text) {
  */
 function backdrop_strlen($text) {
   global $multibyte;
+  if (!isset($text)) {
```
klonos, line 478: @kiamlaluno given the discussion in the issue, I am wondering if we'd be better off using `is_string()` here (instead of type hinting):

```suggestion
  if (!is_string($text)) {
```

klonos: ...and since we are bailing early, we might wanna move this conditional at the beginning of the function, before `global $multibyte;`.

core/includes/unicode.inc
```diff
@@ -475,6 +475,9 @@ function decode_entities($text) {
...
+    return 0;
+  }
```
klonos: @kiamlaluno given the discussion in the issue, I am wondering if we'd be better off using `is_string()` here (instead of type hinting):

...and since we are bailing early, we might wanna move this conditional at the beginning of the function, before `global $multibyte;`.

```suggestion
  // Bail early if no string provided (if $text is NULL for example).
  if (!is_string($text)) {
    return 0;
  }

  global $multibyte;
```

quicksketch:

PR #4287, opened 2022-12-25
Issue #1583: Site default RSS settings do not work for teaser

### LINKED ISSUE #1583: Views RSS "Use site default RSS setting" does not work.
I changed my site default RSS setting to "Titles plus teaser" but the whole node was rendered in my view.

Then, I changed my side default RSS setting to "Titles only" but the whole node was still rendered in my view.

This option does not work.  Let's remove it?

kbahey:
Fixes https://github.com/backdrop/backdrop-issues/issues/1583

### DIFF
--- a/core/modules/node/node.module
+++ b/core/modules/node/node.module
@@ -2189,7 +2189,11 @@ function node_feed($nids = FALSE, $channel = array()) {
       ->fetchCol();
   }
 
-  $item_length = $rss_config->get('rss_viewmode');
+  $display_mode = $rss_config->get('rss_viewmode');
+  if ($display_mode == 'fulltext') {
+    $display_mode = 'full';
+  }
+  
   $namespaces = array('xmlns:dc' => 'http://purl.org/dc/elements/1.1/');
 
   // Load all nodes to be rendered.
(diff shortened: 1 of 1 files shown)


core/modules/node/node.module
```diff
@@ -2189,7 +2189,11 @@ function node_feed($nids = FALSE, $channel = array()) {
       ->fetchCol();
   }
 
-  $item_length = $rss_config->get('rss_viewmode');
+  $display_mode = $rss_config->get('rss_viewmode');
+  if ($display_mode == 'fulltext') {
```
jenlampton, line 2193: Why do we need to change from `fulltext` to `full`, was this one of the setting values in Drupal 7?  I'll go check...

jenlampton: Ah, I see. It was the default value for `feed_item_length`... 
```
$item_length = variable_get('feed_item_length', 'fulltext');
```
And in the update from Drupal 7 we switched that over to `rss_viewmode`
```
$config->set('rss_viewmode', update_variable_get('feed_item_length', 'fulltext'));
```
huh.

kbahey: @jenlampton 

I thought about changing the setting from "fulltext" to "full", but that will break things for people upgrading from previous Backdrop versions or migrating from D7. 

That can be avoided by having an update hook to change it, but at this point, it is more work for something that may eventually be all replaced with views anyway (see discussion in #1583 for more)

Another point: the automated tests keep failing, and I have no idea why ...

PR #4290, opened 2022-12-31
Issue #5899: Better help text for time zone selection in the account edit form

### LINKED ISSUE #5899: Wrong help text in the timezone selector in user edit form
image

The text reads:
> Select the desired local time and time zone. Dates and times throughout this site will be displayed using this time zone.

The user is selecting one thing: timezone - not local time. Also, if you view the second sentence from an admin point of view (while editing another person's user account), it sounds as if this setting will affect everything on the site, whereas this will only affect the user for the account being edited.

klonos:
Fixes https://github.com/backdrop/backdrop-issues/issues/5899


quicksketch:
Thanks @klonos! Merged into 1.x for 1.24.0.

PR #4292, opened 2022-12-31
Issue #980: Add a quick filter to permissions page.

### LINKED ISSUE #980: [UX] Add an instant search filter to the permissions page.
Part of #503.

[Module Filter](https://www.drupal.org/project/module_filter) module for D7 does this:

module_filter-permissions_page-search_working_on_module_sections

...although unfortunately the search is limited to the module sections/headers and not to each of the individual permissions (that's what we should implement):

module_filter-permissions_page-search_not_working_on_individual_permissions

---

Minor release update:
- [x] Add a conclusion suitable for blog post + release notes
- [x] N/AAdd Documentation label if the feature needs to be documented (or needs docs to be updated)
- [x] N/A Add Needs change record label if the the issue contains an API change.
- [x] Add Needs translation label if the the issue contains string translations.

You can now quickly search permissions by name, module name, or description.

quicksketch:
Fixes https://github.com/backdrop/backdrop-issues/issues/980


core/modules/user/user.admin.inc
```diff
@@ -421,6 +421,36 @@ function user_admin_permissions($form, $form_state, $role_name = NULL) {
...
+  );
+  $form['filter']['search'] = array(
+    '#type' => 'textfield',
+    '#title' => t('Filter'),
+    '#size' => 30,
+    '#placeholder' => t('Search...'),
+    '#attributes' => array(
+      'class' => array('table-filter-text'),
+      'autocomplete' => 'off',
+      'title' => t('Enter a part of the module name or description to filter.'),
```
klonos:
```suggestion
      'title' => t('Enter a part of the permission name, module name, or description to filter.'),
```

core/modules/user/user.theme.inc
```diff
@@ -104,21 +104,42 @@ function theme_user_admin_permissions($variables) {
...
+      );
     }
+    // Permission row.
     else {
-      // Permission row.
-      $row[] = array(
-        'data' => backdrop_render($form['permission'][$key]),
-        'class' => array('permission'),
+      $permission_text = '<div class="permission-name table-filter-text-source">' . $form['permission'][$key]['#markup'] . '</div>';
+      $permission_text .= '<div class="permission-description">' . $form['permission'][$key]['#description'] . '</div>';
```
klonos: This results in unnecessary nested divs, and removes the previous `'description'` class from the text (which gives it the smaller font size and gray color). Can we instead do something like this here (avoid adding any extra divs around the description):

```suggestion
      $permission_text .= $form['permission'][$key]['#description'];
```

...and then add the `'table-filter-text-source'` class either directly in `theme_user_permission_description()`? ...or try passing it as parameter in the `$variables` via `user_admin_permissions()`?

The above assumes that the `'description'` class is added automatically in the resulting div - if not, then that will also need to be added along with `'table-filter-text-source'`.

In general, this is what we had before:
```html
<td class="permission">
  <div id="edit-upload-editor-images" class="form-item form-type-item">
    PERMISSION NAME
    <div class="description">PERMISSION DESCRIPTION</div>
  </div>
</td>
```
This is what we get with this PR:

```html
<td class="permission table-filter-text" id="module-edit-upload-editor-images">
  <div class="permission-name table-filter-text-source">PERMISSION NAME</div>
  <div class="permission-description">
    <div class="table-filter-text-source">PERMISSION DESCRIPTION</div>
  </div>
</td>
```

...and this is what we want ideally:
```html
<td class="permission table-filter-text" id="module-edit-upload-editor-images">
  <div class="permission-name table-filter-text-source">PERMISSION NAME</div>
  <div class="permission-description description table-filter-text-source">PERMISSION DESCRIPTION</div>
</td>
```

klonos: ...not sure if we need to retain the `form-item` and `form-type-item` classes in the permission name div like before.

And another diff to consider: the CSS ID was previously being added to the div with the permission name, and it was the name of the permission. Now, it is being added in the `<td>` that contains the divs with the permission name and description, and it is prefixed with `module-`. Not sure what these changes may break.

quicksketch: Thanks @klonos, great catches. I updated the PR:

* Description markup updated as you suggest, with the applied styling now staying the same.
* Removed `module-` prefix from permission rows. Now the IDs are the same as before but they are on the rows instead of on a div.

klonos:
Hey @quicksketch our little "Start and end date on same day" friend is back 😅 ...can you please trigger the tests again?

quicksketch:
Merged into 1.x for 1.24.0. Thanks for the reviewing @klonos!

PR #4293, opened 2023-01-01
Issue #5903: In Seven theme, respect the position settings on admin pages

### LINKED ISSUE #5903: [DX] Remove the `position` key from `hook_menu()`.
## Description of the bug

In [the documentation of `hook_menu()`](https://docs.backdropcms.org/api/backdrop/core%21modules%21system%21system.api.php/function/hook_menu/1), the `'position'` entry is described thus:

> **"position"**: Position of the block ('left' or 'right') on the system administration page for this item.

However, the Seven theme, via `seven_admin_page()`, ignores the position values and simply puts the first half of the blocks on the left and the second half on the right. This can result in a badly unbalanced admin page if it is a mix of long and short blocks, and of course, if the module developer has tried to balance their admin page by judicious selection of left and right position, it will be ignored.

## Expected behavior

If all of the blocks have been positioned as `left` or `right`, I would expect those positions to be honored when creating the admin page.

Of course if no blocks have been positioned, it's reasonable to use a simple heuristic like the current behavior to divide them up between left and right; but if they've already been positioned, those positions should be respected.

## Update

Through the discussion below the community collectively decided that it was strange to have anything related to page display stored in the menu system, so the old Drupal `position` key has been removed. 

The admin pages are now using CSS to divide items into multiple columns, which should work better in most scenarios.

bugfolder:
Fixes https://github.com/backdrop/backdrop-issues/issues/5903.

### DIFF
--- a/core/themes/seven/template.php
+++ b/core/themes/seven/template.php
@@ -96,12 +96,9 @@ function seven_admin_page($variables) {
   );
   foreach ($blocks as $key => $block) {
     if ($block_output = theme('admin_block', array('block' => $block))) {
-      if (($key+1) < count($blocks)/2) {
-        $container['left'] .= $block_output;
-      }
-      else {
-        $container['right'] .= $block_output;
-      }
+      $default_position = (($key + 1) < count($blocks) / 2) ? 'left' : 'right';
+      $position = !empty($block['position']) ? $block['position'] : $default_position;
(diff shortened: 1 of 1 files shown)


bugfolder:
Closing in favor of https://github.com/backdrop/backdrop/pull/4294.

PR #4297, opened 2023-01-03
Issue #5909: Prevent incorrect redirect when resetting permissions filter

### LINKED ISSUE #5909: Instant search filter for the permissions page doesn't reset on role-specific pages
## Description of the bug

https://github.com/backdrop/backdrop-issues/issues/980 introduced an instant filter for the permissions page (/admin/config/people/permissions). This instant filter also appears on role-specific permissions pages, e.g., (/admin/config/people/permissions/administrator). However, on these pages, clicking the "Reset" button takes you back to the all-roles permissions page, rather than leaving you on the role-specific page you were on.

## Steps To Reproduce

To reproduce the behavior:

1. Install a Backdrop version that includes the change, e.g., 1.24.0-preview.
2. Go to /admin/config/people/permissions/administrator.
3. Click the "Reset" button on the filter.

## Actual behavior

You'll be taken to /admin/config/people/permissions.

## Expected behavior

You should remain on the page you were on, /admin/config/people/permissions/administrator.

## Additional information

For people using the role-specific permissions pages to avoid the long load times of the full permissions matrix, this is problematic. The Role Listings module makes it easy to get to the role-specific pages (by providing tabs for them all), but the pages are accessible even without Role Listings module installed.

bugfolder:
Fixes https://github.com/backdrop/backdrop-issues/issues/5909.

### DIFF
--- a/core/modules/user/user.admin.inc
+++ b/core/modules/user/user.admin.inc
@@ -443,7 +443,7 @@ function user_admin_permissions($form, $form_state, $role_name = NULL) {
   $form['filter']['reset'] = array(
     '#theme' => 'link',
     '#text' => t('Reset'),
-    '#path' => 'admin/config/people/permissions',
+    '#path' => $_GET['q'],
     '#options' => array(
       'attributes' => array(
         'class' => array('button', 'button-secondary', 'search-reset'),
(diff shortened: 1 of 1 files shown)


klonos (approved):
LGTM 👍🏼

PR #4299, opened 2023-01-04
Issue #5902: Document the global $conf variable.

### LINKED ISSUE #5902: Add documentation for the global `$conf`.
The global variable `$conf` (which typically holds a copy of the (deprecated) `variables` table) is undocumented (no description) in the [Globals variable listing](https://docs.backdropcms.org/api/backdrop/globals).

It should have a docblock added in core/authorize.php, where it is introduced, and then the API documentation would pick it up. It would also be nice to have some explanation of when and where it could still be used (if at all?). I see there are several usages in core, presumably all related to backward compatibility with use of `variable_*` functions.

Would it be fair to say that if one is not using the backward-compatibility layer for `variable_*` functions, one should not be using the global `$conf` variable either?

bugfolder:
Fixes https://github.com/backdrop/backdrop-issues/issues/5902.

### DIFF
--- a/core/authorize.php
+++ b/core/authorize.php
@@ -76,7 +76,21 @@ require_once BACKDROP_ROOT . '/core/includes/ajax.inc';
 // variables, however, so we have access to the class autoloader registry.
 backdrop_bootstrap(BACKDROP_BOOTSTRAP_SESSION);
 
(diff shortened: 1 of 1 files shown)


quicksketch:

PR #4300, opened 2023-01-08
Issue #5273: Fix unreachable nested contextual links triggers.

### LINKED ISSUE #5273: Nested contextual links triggers sometimes unreachable

bugfolder:

Rebase of https://github.com/backdrop/backdrop/pull/3765.

### DIFF
--- a/core/modules/contextual/js/contextual.js
+++ b/core/modules/contextual/js/contextual.js
@@ -37,8 +37,26 @@ Backdrop.behaviors.contextualLinks = {
       // Prepend the trigger.
       $wrapper.prepend($trigger);
 
-      // Prevent child contextual links overlapping their parent ones.
-      $wrapper.siblings().find('.contextual-links-wrapper').eq(0).css('margin-top', '20px');
+      // Prevent child contextual link triggers from overlapping their parent
+      // ones by shifting their position.
+      var $child = $wrapper.siblings().find('.contextual-links-wrapper').eq(0);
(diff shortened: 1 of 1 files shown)


jenlampton:
Code looks great.

core/modules/contextual/js/contextual.js
```diff
@@ -37,8 +37,46 @@ Backdrop.behaviors.contextualLinks = {
       // Prepend the trigger.
       $wrapper.prepend($trigger);
 
-      // Prevent child contextual links overlapping their parent ones.
-      $wrapper.siblings().find('.contextual-links-wrapper').eq(0).css('margin-top', '20px');
+      // Prevent child contextual link triggers from overlapping their parent
+      // ones by shifting their position.
+      var $child = $wrapper.siblings().find('.contextual-links-wrapper').eq(0);
```
quicksketch, line 42: Is `.siblings()` necessary here? I could see this causing potentially a lot of processing if you had a view that listed a lot of nodes using a teaser display mode. You'd end up looping through all contextual links `(number of nodes) ^ 2` and repeatedly calculating the offsets.

bugfolder:
Closing in favor of https://github.com/backdrop/backdrop/pull/4518.

PR #4302, opened 2023-01-10
Issue #5918: Add role and permissions links when assigning roles

### LINKED ISSUE #5918: [UX] Improve the default user management view and account edit form with links to manage permissions/roles
## Use case

(This makes more sense on complex/bigger sites with many user accounts and multiple roles, but still applies to simpler sites)

A person that has an account on the site contacts you (being the site admin or the developer that is looking after their site), and requests that they get access to perform task x. What usually happens is that the site admin will:

1. go to the the user account management view at `/admin/people/list`
2. filter the list by the person's name/email etc.
3. examine the role(s) the person has, then determine if:
   - the person needs an existing role assigned to them in addition to the ones they already have (a role that already has the permission to do task x)
   - or if a new role needs to be created (because the existing roles that have that permission are overly permissive, with more permissions than what the person should have)
   - or an existing role needs to be edited to add the permission that was requested
4. depending on the case form the previous step, the site admin will need to do any of the following:
   - a. additional role assignment: click the existing "edit" button in the user account screen, add/remove existing roles as required -> done 👍🏼 
   - b. new role needs to be created: need to head to the roles management page at `/admin/config/people/roles` (see #3814, ??? and other issues relevant to improving that workflow)
   - c. permissions of an existing role need to be updated: need to head to `/admin/config/people/permissions` to edit the permissions of the existing user role(s) assigned to the user

## Proposed improvement

Since the other workflows in point 4 above are either being handled OK already, or we have existing issues for them, I would like us streamline the last workflow a bit (case 4c), so that there's less clicking around, less page loads, and less back and forth between admin pages. I propose that we consider implementing the following:
1. Make the role names in the "roles" column of the view in `/admin/people/list` links that take you directly to the permissions management form, filtered by the respective role column only. Alternatively, we could add a single "manage permissions" link to each user account, which filter the permissions management page only by the role(s) already assigned to the user.
2. Either also make the roles in the account edit form links (not a big fan of this, since these are labels of checkboxes), or add a set of two edit links right after the checkboxes: a "manage permissions" and an "edit roles".

klonos:
Fixes https://github.com/backdrop/backdrop-issues/issues/5918

### DIFF
--- a/core/modules/user/user.module
+++ b/core/modules/user/user.module
@@ -946,8 +946,23 @@ function user_account_form(&$form, &$form_state) {
   // Add each role as a checkbox option with a description.
   foreach ($roles as $role_name => $role) {
     $form['account_settings']['roles']['#options'][$role_name] = $role->label;
+    $description = '';
     if (strlen($role->description)) {
-      $form['account_settings']['roles'][$role_name]['#description'] = filter_xss_admin($role->description);
+      $description .= filter_xss_admin($role->description);
+    }
+    // Add quick links to manage roles and they permissions.
+    if (user_access('administer permissions')) {
+      // Make sure that links open in new tabs, to prevent any current edits
+      // from being lost.
+      $open_in_tab = array('attributes' => array('target' => '_blank'));
+      $configure_role_link = l('Configure role', 'admin/config/people/roles/configure/' . $role_name, $open_in_tab);
+      $manage_permissions_link = l('Manage permissions', 'admin/config/people/permissions/' . $role_name, $open_in_tab);
+      // Add the links in a new line if there is a text description.
+      $description .= strlen($description) ? '<br>' : '';
+      $description .= $configure_role_link . ' | ' . $manage_permissions_link;
+    }
+    if (strlen($description)) {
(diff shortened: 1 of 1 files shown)


core/modules/user/user.module
```diff
@@ -946,8 +946,23 @@ function user_account_form(&$form, &$form_state) {
...
+    $description = '';
     if (strlen($role->description)) {
-      $form['account_settings']['roles'][$role_name]['#description'] = filter_xss_admin($role->description);
+      $description .= filter_xss_admin($role->description);
+    }
+    // Add quick links to manage roles and they permissions.
+    if (user_access('administer permissions')) {
+      // Make sure that links open in new tabs, to prevent any current edits
+      // from being lost.
+      $open_in_tab = array('attributes' => array('target' => '_blank'));
```
docwilmot, line 957: Maybe open in dialog might be slicker than new tab/window?

klonos: Thanks for reviewing @docwilmot 🙏🏼 

I've added your comment in the issue, where others can chime in.

PR #4305, opened 2023-01-10
Issue #1360: Add link to layouts page from custom blocks list

### LINKED ISSUE #1360: [UX] Link to the Layouts admin page from the Custom blocks admin page.
@stephit, https://github.com/backdrop/backdrop-issues/issues/1359#issue-118069106

> Put a link in the Blocks menu (or at least on the Blocks page) that directs people to Layouts where they can add blocks. This may be redundant but for the average user used to Drupal it is very confusing at first.

@quicksketch, https://github.com/backdrop/backdrop-issues/issues/1359#issuecomment-158604336

> Great idea. Just a note at the top of the page would be adequate. Like: `Custom blocks may be reused throughout your site on different pages. To place a custom block, configure a layout from the Layouts configuration page.` (with a link on the last 3 words). Let's make a separate issue.

Here it is :wink:

klonos:
Fixes https://github.com/backdrop/backdrop-issues/issues/1360

### DIFF
--- a/core/modules/block/block.admin.inc
+++ b/core/modules/block/block.admin.inc
@@ -12,7 +12,7 @@ function block_admin_list() {
 
   $help_text = array(
     '#theme' => 'help',
-    '#markup' => t('Custom blocks allow placing arbitrary text in layout regions. Blocks created here become available to be placed in any layout in the <a href="@link">Layouts</a> page.', array('@link' => 'admin/structure/layouts')),
+    '#markup' => t('These blocks are available to be placed in any layout in the <a href="@link">Layouts</a> page.', array('@link' => url('admin/structure/layouts'))),
   );
 
   $header = array(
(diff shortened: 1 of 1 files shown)


core/modules/block/block.admin.inc
```diff
@@ -59,12 +64,14 @@ function block_admin_list() {
...
-  return array(
+  $block_admin_list = array(
     '#theme' => 'table__block_admin_list',
-    '#rows' => $rows,
     '#header' => $header,
+    '#rows' => $rows,
     '#empty' => t('No custom blocks have been created yet.'),
   );
+
+  return render($help_text) . render($block_admin_list);
```
argiepiano: Since you are creating two render arrays that contain the key `#theme`, it's more efficient to return the arrays and let the Theme layer take care of rendering them, rather than invoke `render` here. It also allows contrib modules to modify the output before it's rendered. 

I suggest changing this to:

```php
  return array('help_text' => $help_text, 'block_admin_list' => $block_admin_list);
```

klonos: Of course 👍🏼

quicksketch:

PR #4306, opened 2023-01-12
Issue 5925. Avoid passing null as parameter 3 to preg_replace

### LINKED ISSUE #5925: PHP 8.1: Deprecated function: `preg_replace()`: Passing null to parameter #3 in `system_requirements()`
## Description of the bug
When a custom module depends on another custom module that doesn't have a version number, PHP 8.1 throws

`Deprecated function: preg_replace(): Passing null to parameter #3 ($subject) of type array|string is deprecated in system_requirements() (line 472 of...`

This is only encountered when updating the site. 

## Steps To Reproduce

To reproduce you need two custom modules without versions, where one depends on another. Both should be enabled. Then visit `core/update.php` and see the notice.

## Actual behavior

Deprecation notice is shown.

## Expected behavior

No deprecation notice.

## Additional information

This is similar to https://github.com/backdrop/backdrop/pull/4232 (fixed), but is happening in a different place.
Backdrop 1.23.1
PHP 8.1

argiepiano:
Fixes https://github.com/backdrop/backdrop-issues/issues/5925

This is a simple fix. The result is that $version remains an empty string when there is no version, which is exactly what happens with PHP 7.4, when `preg_replace` accepted an empty 3rd argument.

### DIFF
--- a/core/modules/system/system.install
+++ b/core/modules/system/system.install
@@ -485,7 +485,10 @@ function system_requirements($phase) {
         // Check for an incompatible version.
         $required_file = $files[$required_module];
         $required_name = $required_file->info['name'];
-        $version = preg_replace('/^' . preg_quote(BACKDROP_CORE_COMPATIBILITY, '/') . '-/', '', $required_file->info['version']);
+        $version = '';
+        if (!empty($required_file->info['version'])) {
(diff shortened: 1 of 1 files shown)


klonos (approved):
LGTM 👍🏼

quicksketch:
Merged into 1.x and 1.23.x. Thanks @argiepiano and @klonos!

PR #4311, opened 2023-01-13
Issue #5932: Remove confusing secondary role description.

### LINKED ISSUE #5932: [UX] Remove "help text" from new User Role Description
## Description of the bug

I really dislike having two different descriptions for one thing (the recent role help text addition, for example) because it gets very confusing for both users and developers to determine when to use which one (and it's often handled incorrectly). 

Even when in the issue where we added the new feature to add role text, core developers got confused several times as to which description was which, and where each should be shown. This problem is worse for contrib developers deciding which to display on their own interfaces, and worse still for people trying to use the software.

The best solution is often the simplest: Have only one description so there is no confusion where to use which one, and everyone always uses it correctly.

I'd like to propose that we remove the "help" text, and leave only the "description" for the user roles. If an additional description is necessary for other purposes, that seems like a good candidate for contrib. 

## Additional information

- Web server and its version:

## Related

This is a follow-up to https://github.com/backdrop/backdrop-issues/issues/5748

jenlampton:
Fixes https://github.com/backdrop/backdrop-issues/issues/5932

### DIFF
--- a/core/modules/contact/contact.admin.inc
+++ b/core/modules/contact/contact.admin.inc
@@ -159,7 +159,7 @@ function contact_form_permissions() {
     );
     $form['role_names'][$role_name] = array(
       '#markup' => check_plain($role->label),
-      '#description' => filter_xss_admin($role->admin_description),
+      '#description' => filter_xss_admin($role->description),
       '#tree' => TRUE,
     );
   }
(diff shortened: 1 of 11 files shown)


quicksketch:
Merged into 1.x for 1.24.0.

PR #4318, opened 2023-01-26
Issue #5942: Remove "database" from the hook_user_role_*() and hook_node_type_*() documentation

### LINKED ISSUE #5942: [DX] Many of the functions that save to config still save they save in the database.
This is a function docblock/documentation task. Not sure how we can get all of these instances, but I'll start by making a list of them as I come across them:

- `hook_user_role_presave()`
  > Modules implementing this hook can act on the user role object before it has been saved to the database.
- `user_role_save()`
  > Save a user role to the database.
- `hook_node_type_insert()`
  > This hook is invoked from `node_type_save()` after the node type is added to the database.
- ...

avpaderno:
Fixes https://github.com/backdrop/backdrop-issues/issues/5942


quicksketch:
Merged into 1.x and 1.33.x. Thank you @avpaderno and @herbdool!

PR #4320, opened 2023-01-30
Issue #5915: Use the correct layouts on node revision pages.

### LINKED ISSUE #5915: Node layout not being used when viewing a node revision
## Description of the need

When a node display is configured with a layout, the layout is not used when viewing inactive revisions of the node. Instead of the node layout controlling the display, it falls back to the content type display mode. This is counterintuitive to the user.

## Proposed solution

When a node path is entered in a layout as "node/%", that layout should also be applied to the path "node/%/revisions/%/view".

A workaround/alternative is to clone the layout and change the path of the cloned layout. This means any layout edits need to be applied to both layouts.

Node layouts are now applied when an inactive revision is displayed.

bugfolder:
Fixes https://github.com/backdrop/backdrop-issues/issues/5915.

Also added a test for this behavior.


core/modules/node/node.module
```diff
@@ -1665,32 +1665,56 @@ function node_layout_context_info() {
  * tempstore ID of the node begin previewed. But we want to display it using a
```
kswan: Not part of this PR, but I noticed a typo.
begin -> being

bugfolder: Thanks, fixed.

core/modules/node/tests/node.test
```diff
@@ -887,6 +890,70 @@ class NodeLayoutPreviewTestCase extends BackdropWebTestCase {
...
+  /**
+   * Tests that the correct layout is used for node revisions.
+   */
+  public function testNodeLayoutRevisions() {
+    $nodes = $this->nodes;
+
+    // Confirm that our layout shows up on both the original node and its
+    // revision.
+    foreach ($nodes as $node) {
+      $node = $nodes[0];
```
klonos: This line should be removed, as we are already iterating through the 2 node revisions:
```suggestion
```

klonos (approved):
LGTM now 👍🏼

core/modules/node/tests/node.test
```diff
@@ -884,9 +911,76 @@ class NodeLayoutPreviewTestCase extends BackdropWebTestCase {
...
     // Look for the body text, which appears only in that field if the context
     // was set properly.
-    $this->assertText($body, "\"{$body}\" found.");
+    $this->assertText($body, "Body text \"{$body}\" found.");
+
+    // Now enable the 'custom_node_preview' layout; give it path
+    // node/preview/{$this->content_type->type}/%, and run the same tests.
+
+    $layouts = layout_load_all();
+    $preview_layout = &$layouts['custom_node_preview'];
```
argiepiano: The `&` is not really needed, as the value stored at `$layouts['custom_node_preview']` is a PHP object.

core/modules/node/tests/node_layout_preview_revision_test/node_layout_preview_revision_test.module
```diff
@@ -0,0 +1,24 @@
+<?php
+/**
+ * @file
+ * Node layout preview/revision test module.
+ */
+
+/**
+ * Implements hook_config_info().
+ */
+function node_layout_preview_revision_test_config_info() {
```
argiepiano, line 10: I'm wondering if this hook implementation is really needed in the test?

bugfolder: It's useful when you are manually testing out the actions with the test module, because the layouts get removed when the module is uninstalled. (By "manually testing," I mean "doing all the steps in the test manually", not running the tests from the UI.)

core/modules/node/tests/node.test
bugfolder: Agreed.

core/modules/node/tests/node.test
```diff
@@ -950,6 +1046,31 @@ class NodeLayoutRevisionTestCase extends BackdropWebTestCase {
       // was set properly.
       $this->assertText($body, "\"{$body}\" found.");
     }
+
+    // Enable the node revision override layout 'custom_node_revision' with
+    // path node/%/revisions/%/view.
+
+    $layouts = layout_load_all();
+    $preview_layout = &$layouts['custom_node_revision'];
```
argiepiano: This `&` is not needed...

bugfolder: Agreed. (Also renamed the variable: should be `$revision_layout`.)

quicksketch:
Merged into 1.x and 1.24.x. Thanks so much!

PR #4323, opened 2023-02-01
Issue #5955: Clear url query when reset on modules and permissions.

### LINKED ISSUE #5955: Filter Reset for Modules and Permissions doesn't clear url search query
## Description of the bug

This relates to the new functionality on the Permissions page (through #980) and the existing filter on the Modules page.  In #5933, the instant reset was improved to clear the input without reloading the page.  This works, however, if the search filter was added through a URL query (modules can pass url search to the pages or url query is added by pressing enter on field which some users may do), then this is not cleared when Reset is pressed.

## Steps To Reproduce

To reproduce the behavior:
1. Go to `admin/config/people/permissions?search=dashboard`
2. Click on 'Reset'
3. Reload the page

## Actual behavior
Page reloads with the search filter reapplied

## Expected behavior
Page reloads with the search filter remaining cleared.

## Additional information

---

[PR](https://github.com/backdrop/backdrop/pull/4323)

yorkshire-pudding:
Fixes https://github.com/backdrop/backdrop-issues/issues/5955


core/modules/system/js/modules.js
```diff
@@ -90,10 +90,19 @@ Backdrop.behaviors.moduleFilter = {
       }
     }
 
-    // Clear out the input field when clicking the reset button.
+    // Clear out the input field and search query when clicking the reset
+    // button.
     function resetModuleList(e) {
+      // Clear input field.
```
klonos:
```suggestion
      // Clear the input field.
```

core/modules/system/js/modules.js
```diff
@@ -90,10 +90,19 @@ Backdrop.behaviors.moduleFilter = {
...
       $input.val('').triggerHandler('keyup');
       e.preventDefault();
+
+      // Clear search query.
```
klonos:
```suggestion
      // Clear the search query.
```

core/modules/system/js/modules.js
```diff
@@ -90,10 +90,19 @@ Backdrop.behaviors.moduleFilter = {
...
+      var uri = window.location.toString();
+      if (uri.indexOf("?") > 0) {
+          var clean_uri = uri.substring(0, uri.indexOf("?"));
+          window.history.replaceState({}, document.title, clean_uri);
```
klonos:
```suggestion
        var clean_uri = uri.substring(0, uri.indexOf("?"));
        window.history.replaceState({}, document.title, clean_uri);
```

klonos: ...I believe that we can reduce the entire code here simply by chaining `.split('?')[0]` to the `uri` variable, which will work even when there is no query string in the URL, and which won't require any conditional check:

`var uri = window.location.toString().split('?')[0];`

klonos: ...actually, turns out that we don't need to do any variable declaration and processing at all - we can use native functions and simply do this:

`window.location.href = window.location.origin + window.location.pathname;`

- Courtesy of https://stackoverflow.com/a/48542058
- [`location.origin`](https://developer.mozilla.org/en-US/docs/Web/API/Location/origin): ...a string containing the Unicode serialization of the origin of the represented URL. ... That is: ... for URL using the `http` or `https`, the scheme followed by '://', followed by the domain, followed by `':'`, followed by the port (the default port, `80` and `443` respectively, if explicitly specified);
- [`location.pathname`](https://developer.mozilla.org/en-US/docs/Web/API/Location/pathname): ...a string containing the path of the URL for the location. If there is no path, `pathname` will be empty: otherwise, `pathname` contains an initial '/' followed by the path of the URL, not including the query string or fragment.

I've tested this on my local and it works fine 😉

core/modules/system/js/modules.js
yorkshire-pudding: Done

core/modules/system/js/modules.js
yorkshire-pudding: Done

core/modules/system/js/modules.js
yorkshire-pudding: Great find of the much simpler function @klonos - Done

klonos (approved):
LGTM 👍🏼

core/modules/system/js/modules.js
```diff
@@ -90,10 +90,15 @@ Backdrop.behaviors.moduleFilter = {
...
-    // Clear out the input field when clicking the reset button.
+    // Clear out the input field and search query when clicking the reset
+    // button.
     function resetModuleList(e) {
+      // Clear the input field.
       $input.val('').triggerHandler('keyup');
       e.preventDefault();
+
+      // Clear the search query.
+      window.location.href = window.location.origin + window.location.pathname;
```
quicksketch: I _think_ this will cause the page to reload? Setting `window.location` directly will only not reload the page if modifying the anchor. Modifying the query string will cause a reload.

We can use [History.pushState()](https://developer.mozilla.org/en-US/docs/Web/API/History/pushState#change_a_query_parameter) to remove the query string instead. A bit more verbose but I think it does the job without reloading:

```
var currentUrl = new URL(window.location);
currentUrl.searchParams.delete('search');
window.history.replaceState({}, '', currentUrl);
```

yorkshire-pudding: This is similar to my original approach but looks simpler and cleaner. Done

PR #4324, opened 2023-02-04
Issue 5958. Provide method to get the pager's element ID

### LINKED ISSUE #5958: Provide a method to get the pager's element ID.
## Description of the need

This is a cross-port from https://www.drupal.org/node/702778, and part of the task list #5770. The D7 issue added a method, `getElement()`, to `PagerDefault` that returns the element ID of the pager. Having this method allows developers an easier way to have more than one pager in the same page without conflicts. 

Before this new method, the only way to handle more than one pager in the same page was to hard-code the element ID through `PagerDefault::element()`. This doesn't make much sense and can lead to unexpected conflicts if two different modules need to add a pager to the same page. `PagerDefault` already has a way to avoid those conflicts by keeping track of IDs in a static variable, but until this feature was added in D7, it was impossible to get the internally assigned element ID (which is a protected property). 

PR coming.

argiepiano:
Fixes https://github.com/backdrop/backdrop-issues/issues/5958

### DIFF
--- a/core/includes/pager.class.inc
+++ b/core/includes/pager.class.inc
@@ -178,6 +178,8 @@ class PagerDefault extends SelectQueryExtender {
    * @return int
    *   Element ID that is used to differentiate between different pager
    *   queries.
+   *
+   * @since 1.24.1 Method added.
    */
   public function getElement() {
     $this->ensureElement();
--- a/core/includes/pager.class.inc
+++ b/core/includes/pager.class.inc
@@ -168,4 +168,19 @@ class PagerDefault extends SelectQueryExtender {
     }
     return $this;
   }
+
+  /**
+   * Gets the element ID for this pager query.
+   *
+   * The element is used to differentiate different pager queries on the same
+   * page so that they may be operated independently.
+   *
+   * @return
+   *   Element ID that is used to differentiate between different pager
+   *   queries.
+   */
+  public function getElement() {
+    $this->ensureElement();
+    return $this->element;
+  }
 }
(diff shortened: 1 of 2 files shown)


core/includes/pager.class.inc
```diff
@@ -168,4 +168,19 @@ class PagerDefault extends SelectQueryExtender {
...
     return $this;
   }
+
+  /**
+   * Gets the element ID for this pager query.
+   *
+   * The element is used to differentiate different pager queries on the same
+   * page so that they may be operated independently.
+   *
+   * @return
```
klonos: This should stop PHPCS from complaining:
```suggestion
   * @return int
```

klonos (approved):
LGTM now 👍🏼

core/includes/pager.class.inc
```diff
@@ -168,4 +168,19 @@ class PagerDefault extends SelectQueryExtender {
...
+  /**
+   * Gets the element ID for this pager query.
+   *
+   * The element is used to differentiate different pager queries on the same
+   * page so that they may be operated independently.
+   *
+   * @return int
+   *   Element ID that is used to differentiate between different pager
+   *   queries.
+   */
```
quicksketch:
```suggestion
   *   queries.
   *
   * @since 1.24.1 Method added.
   */
```

quicksketch:
Looks good to me! I added a `@since` docblock line to indicate this will be in 1.24.1.

quicksketch:

PR #4328, opened 2023-02-05
Issue 5960. Fix handling of files that exceed php upload limit

### LINKED ISSUE #5960: Incorrect handling of file upload PHP limit exceeded
This is a cross-port of https://www.drupal.org/project/drupal/issues/2749245, which is part of the #5770 task list.

This issue was partly fixed in Backdrop:
- Unlike D7 the widget does not disappear upon receiving the error `An unrecoverable error occurred. The uploaded file likely exceeded the maximum file size`
- Like D7, you can't save the node after this error has occurred. 

To reproduce, follow the steps in the D7 issue. Copying and adapting:

1. Turn off `display_errors` in php.ini.
2. Make sure the post_max_size is set to equal or lower than the `upload_max_filesize` in php.ini. The bug only manifests itself when the `post_max_size` is exceeded.
3. Create a file upload field that is not limited in file size.
4. Upload any file larger than the `post_max_size`

In Backdrop, the result: the error is shown correctly, and the widget does not disappear, but (like D7) when you click Save, the form is reloaded and nothing is saved.

PR coming.

argiepiano:
Fixes https://github.com/backdrop/backdrop-issues/issues/5960

### DIFF
--- a/core/modules/file/file.module
+++ b/core/modules/file/file.module
@@ -3343,13 +3343,13 @@ function file_view_file($file, $display = 'full', $langcode = NULL) {
  * That form will trigger the type classification of managed files that do not
  * currently have a type. This is often the case after upgrading from Drupal 7.
  *
- * @param $classify
+ * @param bool|NULL $rebuild
  *   (optional) The boolean value to be written.
  *
  * @return bool|NULL
  *   The current value of the flag if no value was provided for $rebuild.
  *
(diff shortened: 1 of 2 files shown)


quicksketch:
Looks great! Merged into 1.x and 1.26.x. Thanks @argiepiano!

PR #4332, opened 2023-02-09
Issue 5969. Remove dangerous protocols from $path in `l()`

### LINKED ISSUE #5969: [SR] Security improvement for l() function
This is a cross-port for the D7 issue https://www.drupal.org/project/drupal/issues/2990723. Tracked by the task list #5770.

From the D7 issue:

> The issue was reported first on security.drupal.org where we agreed to fix it in the public queue. Members who suggested it: greggles, larowlan and mlhess. 
> 
> <h3 id="summary-problem-motivation">Problem/Motivation</h3>
> l() function does not strip dangerous protocols when external parameter is set to TRUE.
> 
> When it returns the output it only calls check_plain for sanitize the path that is given from url() - because comment also indicates it's a plain text URL, it should be sanitized before, but currently it's not.
> 
> For reproduction steps, you can try one of the following:
> 
> Scenario No 1:
> 1. Create a new test module in drupal.
> 2. Implement a hook_menu that will call a custom function.
> 3. In the page callback return the output of l() like this:
> <code>l('link', "javascript:alert('XSS')", array('external' => TRUE))</code>

> Scenario No 2:
> 1. Turn on PHP module.
> 2. Create a new article content and use the php filter format and insert to body the following:
> ```php
> print l('link', "javascript:alert('XSS')", array('external' => TRUE));
> ```
> 3. Save the content, visit its page and click on the 'link' link, then the alert will be executed.
> 
> Both of the above scenarios need more permissions on the website that could lead more serious issues, that's also the reason why we turned to public hardening.

> ## Proposed resolution
> As l() function is widely used in drupal core, also in contribs, changing the way how it prints the output could lead to breaking functionalities, the proposal is to sanitize the path before it's given to sanitization functions that do the other jobs. See currently attached patch where the drupal_strip_dangerous_protocols() function is directly called on $path.

argiepiano:
Fixes https://github.com/backdrop/backdrop-issues/issues/5969


core/includes/common.inc
```diff
@@ -2913,7 +2913,16 @@ function l($text, $path, array $options = array()) {
   $attributes = backdrop_attributes($options['attributes']);
   unset($options['attributes']);
 
-  $url = check_plain(url($path, $options));
+  $url = url($path, $options);
+  $skip_js_paths = array(
+    'javascript:void()',
+    'javascript:void();',
+    'javascript:void(0)',
+    'javascript:void(0);'
```
bugfolder:
```suggestion
    'javascript:void(0);',
```

core/modules/simpletest/tests/common.test
```diff
@@ -76,6 +76,26 @@ class CommonURLWebTestCase extends BackdropWebTestCase {
...
+    $link = l($text, $path, array('external' => TRUE));
+    $this->assertTrue(strpos($link, 'javascript:') === FALSE, 'Dangerous protocol javascript: was sanitized.');
+
+    // Verify that these harmless javascript paths are left intact for BC.
+    $special_case_js_paths = array(
+      'javascript:void()',
+      'javascript:void();',
+      'javascript:void(0)',
+      'javascript:void(0);',
+      'JavaScript:Void(0)'
```
bugfolder:
```suggestion
      'JavaScript:Void(0)',
```

core/includes/common.inc
argiepiano: Change pushed.

core/modules/simpletest/tests/common.test
argiepiano: Change pushed.

quicksketch:
Thanks @argiepiano! Merged into 1.x for 1.26.1.

PR #4333, opened 2023-02-11
#5971: Avoid username enumeration via password reset URLs.

### LINKED ISSUE #5971: Username enumeration via password reset URL
I found [this issue](https://www.drupal.org/project/drupal/issues/2828724) in Drupal's issue queue and discovered that Backdrop has the same problem. Since it's already public on D.org, and since [username enumeration isn't considered a security issue in Backdrop](https://github.com/backdrop/backdrop-issues/issues/4696), I'm posting this here instead of the security queue.

The issue is that if you access a password reset URL for any user ID, when logged in as another user, you'll see the username matching the user ID. For example:

image

More background about this particular problem can be found in the Drupal issue, and a discussion about username enumeration security/privacy concerns can be found in #4696.

The thing some people took issue with about the #4696 suggestion is that the inconvenience of the solution apparently outweighs the benefit. What makes *this* issue different, however, is that I don't see any downside to fixing this. There's no real need/benefit of mentioning the username of the user you're trying to reset the password for. We could simply change this message to something like:

> You cannot use a password reset link while logged into the site. Please logout and try using the link again.

ghost:
Fixes https://github.com/backdrop/backdrop-issues/issues/5971

### DIFF
--- a/core/modules/user/user.pages.inc
+++ b/core/modules/user/user.pages.inc
@@ -59,7 +59,7 @@ function user_pass() {
     $form['name']['#value'] = $user->mail;
     $form['mail'] = array(
       '#prefix' => '<p>',
-      '#markup' =>  t('Password reset instructions will be mailed to %email. You must log out to use the password reset link in the email.', array('%email' => $user->mail)),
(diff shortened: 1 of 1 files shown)


klonos (approved):
LGTM 👍🏼

hosef:
Thanks! Merged into 1.x and 1.24.x.

PR #4340, opened 2023-02-14
Issue #1830: Remove mention of hook_views_default_views(). [skip tests]

### LINKED ISSUE #1830: Remove mention of `hook_views_default_views()` from docs
In the docblock for `hook_views_api` we mention `hook_views_default_views()` but that function does not exist in backdrop. I recommend we change 

```
This is required for your module to have its include files loaded; for example, when implementing hook_views_default_views().
```

to 

```
This is required for your module to have its include files loaded.
```

Also, we should add a change record to help new Backdrop developers transition from using `hook_views_default_views` to config.

Releveant page on api site: https://api.backdropcms.org/api/backdrop/core!modules!views!views.api.php/function/hook_views_api/1

ghost:
Fixes https://github.com/backdrop/backdrop-issues/issues/1830

### DIFF
--- a/core/modules/views/views.api.php
+++ b/core/modules/views/views.api.php
@@ -632,8 +632,7 @@ function hook_views_plugins_alter(&$plugins) {
 /**
  * Register View API information.
  *
- * This is required for your module to have its include files loaded; for
- * example, when implementing hook_views_default_views().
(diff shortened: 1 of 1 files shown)


ghost:
Merged into 1.x.

PR #4341, opened 2023-02-14
Issue #1662: Reword config backup instructions during update.

### LINKED ISSUE #1662: Upgrade instructions - folder(s) to back-up
I noticed a difference in the core upgrade instructions.

On https://backdropcms.org/upgrade, step #2 is
**Make a backup of the "config_*" directory**

During the actual upgrade process, instead, the procedure suggests to back-up only the active subdrir:
**Back up the entire './files/config_220bfb6f0fc5e3b34319afbba8b63ffb/active**

Would the second option be enough, in case of trouble?
Conflicting info could be avoided, anyway.

ghost:
Fixes https://github.com/backdrop/backdrop-issues/issues/1662

### DIFF
--- a/core/update.php
+++ b/core/update.php
@@ -305,7 +305,7 @@ function update_info_page() {
   $db_name = $databases['default']['default']['database'];
 
   // Get the config path
-  $config_dir = config_get_config_directory('active');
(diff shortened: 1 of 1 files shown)


quicksketch:
Closed in favor of keeping in-core documentation as-is and updating the online documentation to match.

PR #4346, opened 2023-02-16
Issue #5447: Report db type and version (2 digits)

### LINKED ISSUE #5447: Fix mysql/mariadb version for Telemetry and the Status Report

jenlampton:
Fixes https://github.com/backdrop/backdrop-issues/issues/5447

Replaces https://github.com/backdrop/backdrop/pull/3904

### DIFF
--- a/core/modules/system/system.module
+++ b/core/modules/system/system.module
@@ -1122,6 +1122,46 @@ function _system_batch_theme() {
 }
 
 /**
+ * Helper function to deterine the MySQL or MariaDB versionl
+ *
+ * @see system_requirements().
+ * @see telemetry_telemetry_data().
+ */
+function _system_get_database_version() {
+  $version = Database::getConnection()->databaseType() === 'mysql' ? Database::getConnection()->version() : NULL;
+  // Backdrop core only supports MySQL and MariaDB. If there is no "maria"
+  // in the version string, assume it is MySQL.
+  $type = strpos(strtolower($version), 'maria') !== FALSE ? 'MariaDB' : 'MySQL';
+  // MariaDB version strings tend to have 2 versions, one for the equivalent
+  // MySQL version + another for the actual MariaDB version. For example, in
+  // 5.5.5-10.5.11-MariaDB:
+  // - 5.5.5 would denote the MySQL equivalency
+  // - 10.5.11 would be the actual MariaDB version.
+  // Grab all strings within the version that are a sequence of numbers and
+  // dots.
+  preg_match_all('/[\d\.]+/', $version, $versions);
+  if (count($versions[0]) == 1) {
+    $version = $versions[0][0];
+  }
+  else {
+    // Some variations of the MariaDB version string contain the version
+    // number more than once. For example:
+    // 5.5.5-10.5.11-MariaDB-1:10.5.11+maria~focal-log.
+    $versions = array_unique($versions[0]);
+    // Sort versions in reverse (descending) order.
+    rsort($versions, SORT_NATURAL);
+    // Get the greatest version (or the only version).
+    $version = $versions[0];
+  }
+  // Trim the version number to only major + minor, otherwise we'd flood the
+  // table with dozens of bug fix versions.
+  $version = explode(".", $version);
+  $version = array_slice($version, 0, 2);
+  $version = implode('.', $version);
+  return $type . ' v' . $version;
+}
+
+/**
  * Implements hook_page_delivery_callback_alter().
  *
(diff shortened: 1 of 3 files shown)


core/modules/system/system.module
```diff
@@ -1121,6 +1121,46 @@ function _system_batch_theme() {
...
+    $versions = array_unique($versions[0]);
+    // Sort versions in reverse (descending) order.
+    rsort($versions, SORT_NATURAL);
+    // Get the greatest version (or the only version).
+    $version = $versions[0];
+  }
+  // Trim the version number to only major + minor, otherwise we'd flood the
+  // table with dozens of bug fix versions.
+  $version = explode(".", $version);
+  $version = array_slice($version, 0, 2);
```
klonos: Since this version number is going to be used in multiple places, where in some of them the bug version may be important, we shouldn't be doing this here. Perhaps this function should be returning an array of separate type and version number, and then we can be doing the concatenation and trimming of the less significant version numbers only where appropriate. For example, use only 2 digits in telemetry, but the entire version number in the site status report.

core/modules/system/system.module
```diff
@@ -1121,6 +1121,46 @@ function _system_batch_theme() {
   }
 }
 
+/**
+ * Helper function to deterine the MySQL or MariaDB versionl
```
klonos: Surprised how the missing dot at the end didn't trigger PHPCS 🤔 

```suggestion
 * Helper function to determine the MySQL or MariaDB version.
```

core/modules/system/system.module
klonos: ...here's a PR against this: https://github.com/jenlampton/backdrop/pull/17

core/modules/telemetry/telemetry.telemetry.inc
```diff
@@ -44,7 +44,13 @@ function telemetry_telemetry_data($key) {
     case 'php_version':
       return PHP_MAJOR_VERSION . '.' . PHP_MINOR_VERSION . '.' . PHP_RELEASE_VERSION;
     case 'mysql_version':
-      return Database::getConnection()->databaseType() === 'mysql' ? Database::getConnection()->version() : NULL;
+      // Trim the version number to only major + minor, otherwise we'd flood the
+      // table with dozens of bug fix versions.
```
klonos, line 48: @jenlampton I'm not sure if we should be doing this. If we ever need to troubleshoot issues that happen in specific bug fix versions (like in https://github.com/backdrop/backdrop-issues/issues/5496 for instance), then we won't have the data. The borg side of our telemetry implementation already allows us to choose to aggregate on either 0/1/2 decimals for the versions:
image

Also, if you take a look at the `case 'php_version':` above, you'll see that we aren't trimming the release (bug fix) version there either.

Point being: lets collect more accurate data in case we ever need it - the granularity on display on the metrics table can easily be limited to only major/minor.

quicksketch: I agree, I don't think this trimming is necessary. It's trimmed on the display side like @klonos says above.

quicksketch:
Replaced by https://github.com/backdrop/backdrop/pull/4853.

PR #4348, opened 2023-02-20
Issue #5986: Allow token browser collapse/expand strings to be translated

### LINKED ISSUE #5986: Token browser: Explicitly set the `stringCollapse` and `stringExpand` settings and pass them through `t()` to allow for them to be translated
The jQuery treetable plugin allows for `stringCollapse` and `stringExpand` as settings(*), which can be customized. The default strings for them are "Collapse" and "Expand" respectively, and these are fine - they cannot be translated though, so we should be setting them explicitly in `core/modules/system/js/token.js` and pass them through `Backdrop.t()` to allow for that.

*See: http://ludo.cubicphuse.nl/jquery-treetable for details.

klonos:
Fixes https://github.com/backdrop/backdrop-issues/issues/5986


quicksketch:
Thanks @klonos and @bugfolder! I merged this into 1.x and 1.26.x.

PR #4349, opened 2023-02-20
Issue #5987: The great Backdrop spell check.

### LINKED ISSUE #5987: The great Backdrop spell check!
Backdrop has quite a few spelling mistakes/typos. Let's fix them!

---

When I saw a typo in core while working on another issue, I wondered how long it'd been there. Git Blame pointed to Drupal, where I noticed it'd since been fixed as part of a larger issue to fix lots of spelling mistakes. I then wondered how many Backdrop had...

After doing some research, I came across the [`cspell`](https://cspell.org/) command-line tool which looked promising. So I setup a local Lando env. where I installed cspell and configured it to run against all of core. After *much* configuring and tweaking, I finally got the following:

- A PR of fixed typos in core (see below)
- A cspell config file that:
  - Uses Drupal's own cspell dictionary
  - Uses a bunch of other dictionaries
  - Uses a custom dictionary for Backdrop
  - Ignores certain files/folders (e.g. SVG files, core's included transliteration directory, etc.)
  - Ignores specific words in specific files (e.g. `dont` is considered a typo in most of Backdrop, except `core/modules/simpletest/tests/session*` files where it's used in a variable name and hence can't have an apostrophe, so it's legitimate there)
- A custom Backdrop dictionary that includes ~550 words included in Backdrop core that aren't typos but aren't recognised by any of the other dictionaries

Only later did I find [this issue](https://github.com/backdrop/backdrop-issues/issues/5811) to add cspell as a GH automation for PRs (that's what led me to Drupal's dictionary). So I'll likely go and post my cspell files there and try to get that moving next, but for *this* issue let's just fix all the typos in core in bulk before adding an automation.

The below PR mostly fixes documentation typos (comments, docblocks, etc.), but there are also some typos in code (`*.api.php` files, strings shown to the user, and even variable/function/class names). For the latter, I checked each instance and determined that it didn't really constitute an API change to fix (most were only used in tests anyway), so I went ahead with including them here. But happy for others to review and provide feedback in that regard.

ghost:
Fixes https://github.com/backdrop/backdrop-issues/issues/5987


core/includes/bootstrap.classes.inc
```diff
@@ -14,7 +14,7 @@ function backdrop_class_list() {
     'BatchQueue' => $path . 'batch.queue.inc',
     'BatchMemoryQueue' => $path . 'batch.queue.inc',
     'BackdropCacheArray' => $path . 'bootstrap.inc',
-    'SchemaCache' => $path . 'boostrap.inc',
+    'SchemaCache' => $path . 'bootstrap.inc',
```
avpaderno, line 17: Good catch! This surely is causing issues. 👍🏻

core/includes/color.inc
```diff
@@ -47,7 +47,7 @@ class Color {
     // Ignore '#' prefixes.
     $hex = ltrim($hex, '#');
 
-    // Convert shorhands like '#abc' to '#aabbcc'.
```
avpaderno, line 50: I think the plural is more correct, in this case. I would rather put a comma after *shorthands* and use *for example* instead of *like*.

core/includes/theme.inc
```diff
@@ -591,7 +591,7 @@ function _theme_process_registry(&$cache, $name, $type, $theme, $path) {
 
       // Allow variable preprocessors for all theme hooks, whether the hook is
       // implemented as a template or as a function.
-      // Check for existing variable preprocessors. Ensure arrayness.
+      // Check for existing variable preprocessors. Ensure arrays.
```
avpaderno: Both sentences do not seem right. I would rather remove that, as *Ensure arrays* just describes something already clear from the code. A variable is changed to an array in other cases, but no comment says that.

core/modules/entityreference/tests/entityreference.handlers.test
```diff
@@ -480,12 +480,12 @@ class EntityReferenceHandlersTestCase extends BackdropWebTestCase {
         ),
       ),
     );
-    $this->assertReferencable($field, $referencable_tests, 'Comment handler (comment admin)');
+    $this->assertReferencable($field, $referenceable_tests, 'Comment handler (comment admin)');
```
avpaderno: The method name should also be changed.

core/modules/field/tests/field.test
```diff
@@ -1409,7 +1409,7 @@ class FieldInfoTestCase extends FieldTestCase {
     );
     field_create_instance($instance_definition);
 
-    // Disable coment module. This clears field_info cache.
+    // Disable comment module. This clears field_info cache.
```
avpaderno: This should probably be *Disable the Comment module.*

core/includes/bootstrap.classes.inc
ghost: I know, right!

core/includes/color.inc
ghost: I'm not sure I like the word 'shorthands' in this context - doesn't seem right to me. But I see your point, so I've made it 'Convert shorthand colors' instead. I agree with your other suggestions.

core/includes/theme.inc
ghost: Agreed.

core/modules/entityreference/tests/entityreference.handlers.test
ghost: `assertReferencable()` is only used in this file, so I've gone ahead and updated it. However the reason I didn't do so initially was because `Referencable` is used in 3 other function names which are used across 7 different files. So I think that, unfortunately, needs to stay.

core/includes/color.inc
```diff
@@ -47,7 +47,7 @@ class Color {
...
+    // Convert shorthand colors, for example '#abc', to '#aabbcc'.
```
klonos: @BWPanda I believe that @kiamlaluno meant this:
```suggestion
    // Convert shorthand colors. For example, '#abc' to '#aabbcc'.
```

klonos: ...or something this?:
```suggestion
    // Convert shorthand colors (for example '#abc' to '#aabbcc').
```

core/includes/bootstrap.inc
```diff
@@ -726,7 +726,7 @@ function backdrop_environment_initialize() {
 
   // When clean URLs are enabled, emulate ?q=foo/bar using REQUEST_URI. It is
   // not possible to append the query string using mod_rewrite without the B
-  // flag (this was added in Apache 2.2.8), because mod_rewrite unescapes the
+  // flag (this was added in Apache 2.2.8), because mod_rewrite un-escapes the
```
klonos: @BWPanda this seems to be legit: https://en.wiktionary.org/wiki/unescape

So perhaps add it to the Backdrop-specific dictionary and allow it?

core/includes/common.inc
```diff
@@ -5829,7 +5829,7 @@ function backdrop_json_decode_unicode($json) {
     }
     return $match[1] . $string;
   };
-  // To prevent unintentionally unescaping the string "\u", be sure that the
+  // To prevent unintentionally un-escaping the string "\u", be sure that the
```
klonos: As per previous comment, this word seems to be legit without the dash: https://en.wiktionary.org/wiki/unescape

core/includes/diff.inc
```diff
@@ -144,16 +144,16 @@ class DiffOpChange extends DiffOp {
...
- * Some ideas are (and a bit of code) are from from analyze.c, from GNU
- * diffutils-2.7, which can be found at:
- *     ftp://gnudist.gnu.org/pub/gnu/diffutils/diffutils-2.7.tar.gz
+ * Some ideas (and a bit of code) are from analyze.c, from GNU diffutils-2.7,
+ * which can be found at:
+ * ftp://gnudist.gnu.org/pub/gnu/diffutils/diffutils-2.7.tar.gz
  *
- * closingly, some ideas (subdivision by NCHUNKS > 2, and some optimizations)
+ * Finally, some ideas (subdivision by NCHUNKS > 2, and some optimizations)
  * are my own.
```
klonos, line 157: Since we are fixing this, this should either indicate the who the person is and give due credits, or simply say "are original" or something like that.

klonos: ...never mind. There's a list of authors below, which should be enough.

core/includes/diff.inc
```diff
@@ -774,9 +774,8 @@ class MappedDiff extends Diff {
    *
    * Computes diff between sequences of strings.
    *
-   * This can be used to compute things like
-   * case-insensitve diffs, or diffs which ignore
-   * changes in white-space.
+   * This can be used to compute things like case-insensitive diffs, or diffs
+   * which ignore changes in white-space.
```
klonos, line 778: Is the dash in `white-space` really necessary here?

core/includes/install.inc
```diff
@@ -114,7 +114,7 @@ function backdrop_get_schema_versions($module) {
     // Narrow this down to functions ending with an integer, since all
     // hook_update_N() functions end this way, and there are other
     // possible functions which match '_update_'. We use preg_grep() here
-    // instead of foreaching through all defined functions, since the loop
+    // instead of foreach-ing through all defined functions, since the loop
```
klonos: I like word inventing as much as the next person 😅  ...but can't we just use "iterating through" here?

core/includes/color.inc
avpaderno: I was thinking of *Convert shorthand colors, for example '#abc' to '#aabbcc'.* but that is not a typo issue, but a "writing style" issue. Actually, there is nothing wrong in using *like*; to me it seems too *colloquial*, but it is not an error, when it is followed by a pronoun or a noun. (I always think to *Spies like us*, a movie title.)

core/includes/theme.inc
```diff
@@ -591,7 +591,7 @@ function _theme_process_registry(&$cache, $name, $type, $theme, $path) {
 
       // Allow variable preprocessors for all theme hooks, whether the hook is
       // implemented as a template or as a function.
-      // Check for existing variable preprocessors. Ensure arrayness.
+      // Check for existing variable preprocessors.
```
klonos: We could say "and valid" here, since that's what "ensuring arrayness" tried to say I believe.

core/misc/autocomplete.js
```diff
@@ -145,7 +145,7 @@ Backdrop.jsAC.prototype.selectUp = function () {
  * Highlights a suggestion.
  */
 Backdrop.jsAC.prototype.highlight = function (node) {
-  // Unhighlights a suggestion for "keyup" and "keydown" events.
+  // Un-highlights a suggestion for "keyup" and "keydown" events.
```
klonos: Legit: https://en.wiktionary.org/wiki/unhighlight

core/misc/autocomplete.js
```diff
@@ -155,7 +155,7 @@ Backdrop.jsAC.prototype.highlight = function (node) {
 };
 
 /**
- * Unhighlights a suggestion.
+ * Un-highlights a suggestion.
```
klonos: Legit: https://en.wiktionary.org/wiki/unhighlight

core/includes/bootstrap.inc
avpaderno: I would trust more the Merriam-Webster dictionary, which say [*unescape* does not exist](https://www.merriam-webster.com/dictionary/unescape). 😅

core/modules/date/date.elements.inc
```diff
@@ -2160,7 +2161,7 @@ function _date_html5_get_date_limits(array $limit_settings) {
 /**
  * Helper function to construct form item attributes.
  *
- * Takes care of nested (html_datetime) and unnested (html_date) form items.
+ * Takes care of nested (html_datetime) and un-nested (html_date) form items.
```
klonos: Legit: https://en.wiktionary.org/wiki/unnested

core/modules/date/date.module
```diff
@@ -545,7 +545,7 @@ function date_format_type_format($format_type, $langcode = NULL) {
   if (!isset($static[$langcode][$format_type])) {
     $date_format = system_date_format_load($format_type);
 
-    // Fallback to (undeletable) medium format if the requested one does not
+    // Fallback to (un-deletable) medium format if the requested one does not
```
klonos: Legit: https://en.wiktionary.org/wiki/undeletable

core/includes/diff.inc
avpaderno: Maybe this file should not be changed, since it comes from an external repository. (The comment in the code says it originally came from `https://git.wikimedia.org/raw/mediawiki%2fcore.git/HEAD/includes%2fdiff%2fDairikiDiff.php`.)

core/modules/entityreference/plugins/behavior/EntityReferenceBehaviorHandler.inc
```diff
@@ -59,7 +59,7 @@ interface EntityReferenceBehaviorHandlerInterface {
   public function validate($entity_type, $entity, $field, $instance, $langcode, $items, &$errors);
 
   /**
-   * Act on presaving an entity reference field.
+   * Act on pre-saving an entity reference field.
```
klonos, line 62: 👍🏼

core/modules/entityreference/tests/entityreference.handlers.test
```diff
@@ -21,7 +21,7 @@ class EntityReferenceHandlersTestCase extends BackdropWebTestCase {
     }
   }
 
-  protected function assertReferencable($field, $tests, $handler_name) {
+  protected function assertReferenceable($field, $tests, $handler_name) {
```
klonos, line 24: If we wanted to be really safe about this (contrib/custom modules might be using it), we should add a wrapper function and a `@todo` item to remove in 2.0

klonos: ...oh, this is a protected function though. So perhaps not.

core/modules/field/field.api.php
```diff
@@ -298,7 +298,7 @@ function hook_field_schema($field) {
  *   FIELD_LOAD_REVISION to load the version indicated by each entity.
  */
 function hook_field_load($entity_type, $entities, $field, $instances, $langcode, &$items, $age) {
-  // Sample code from text.module: precompute sanitized strings so they are
+  // Sample code from text.module: pre-compute sanitized strings so they are
```
klonos: Legit: https://en.wiktionary.org/wiki/precompute

core/modules/field/field.attach.inc
```diff
@@ -869,7 +869,7 @@ function field_attach_submit($entity_type, $entity, $form, &$form_state, $option
 function field_attach_presave($entity_type, $entity) {
   _field_invoke('presave', $entity_type, $entity);
 
-  // Let other modules act on presaving the entity.
+  // Let other modules act on pre-saving the entity.
```
klonos, line 872: 👍🏼

core/misc/autocomplete.js
avpaderno: There is [*unhighlighted*](https://www.merriam-webster.com/dictionary/unhighlight), but not *unhighlight*.

core/modules/field/modules/text/text.module
```diff
@@ -316,7 +316,7 @@ function text_field_formatter_view($entity_type, $entity, $field, $instance, $la
  *  The sanitized string.
  */
 function _text_sanitize($instance, $langcode, $item, $column) {
-  // If the value uses a cacheable text format, text_field_load() precomputes
+  // If the value uses a cacheable text format, text_field_load() pre-computes
```
klonos: Legit: https://en.wiktionary.org/wiki/precompute

core/modules/field/tests/field.test
```diff
@@ -1409,7 +1409,7 @@ class FieldInfoTestCase extends FieldTestCase {
     );
     field_create_instance($instance_definition);
 
-    // Disable coment module. This clears field_info cache.
+    // Disable the comment module. This clears field_info cache.
```
klonos: Module names should be properly capitalized:
```suggestion
    // Disable the Comment module. This clears field_info cache.
```

core/modules/field/tests/field.test
```diff
@@ -1587,7 +1587,7 @@ class FieldFormTestCase extends FieldTestCase {
     $this->assertFieldByName("{$this->field_name}[$langcode][0][value]", '', 'Widget 1 is displayed');
     $this->assertFieldByName("{$this->field_name}[$langcode][1][value]", '', 'New widget is displayed');
     $this->assertNoField("{$this->field_name}[$langcode][2][value]", 'No extraneous widget is displayed');
-    // TODO : check that non-field inpurs are preserved ('title')...
+    // TODO : check that non-field inputs are preserved ('title')...
```
klonos:
```suggestion
    // @todo: Check that non-field inputs are preserved ('title').
```

core/modules/date/date.elements.inc
avpaderno: There is [*unnest*](https://www.merriam-webster.com/dictionary/unnest) but it has a different meaning. 😉

core/modules/field_ui/field_ui.admin.inc
```diff
@@ -1499,7 +1499,7 @@ function field_ui_field_type_options() {
     $field_type_options = array();
     foreach ($field_types as $name => $field_type) {
       // Skip field types which have no widget types, or should not be add via
-      // uesr interface.
+      // user interface.
```
klonos: Grammar: "added" (in the line above)

I would also expect a "the" before "user interface".

core/modules/file/file.field.inc
```diff
@@ -127,7 +127,7 @@ function file_field_instance_settings_form($field, $instance) {
 }
 
 /**
- * Render API callback: Validates the maximum uplodad size field.
+ * Render API callback: Validates the maximum upload size field.
```
klonos: This doesn't make sense. Perhaps "for the field"?

core/modules/file/js/file.js
```diff
@@ -129,7 +129,7 @@ Backdrop.file = Backdrop.file || {
     // do not get enabled when we re-enable these fields at the end of behavior
     // processing. Re-enable in a setTimeout set to a relatively short amount
     // of time (1 second). All the other mousedown handlers (like Backdrop's
-    // Ajax behaviors) are excuted before any timeout functions are called, so
+    // Ajax behaviors) are executed before any timeout functions are called, so
```
klonos:
```suggestion
    // AJAX behaviors) are executed before any timeout functions are called, so
```

core/modules/field_ui/field_ui.admin.inc
avpaderno: *via user interface* is acceptable, in the same way *stay up to date with the latest headlines via email* is.

core/modules/layout/tests/layout.test
```diff
@@ -1981,6 +1981,8 @@ class LayoutBlockTest extends BackdropWebTestCase {
     $hero_title_1 = 'Wonderwoman';
     $hero_content_1 = 'Youth is your greatest weapon, your greatest tool.';
     $hero_title_2 = 'Black Widow';
+    // "isnt" is purposefully missing an apostrophe as it was causing tests to
+    // fail.
```
klonos, line 1985: Then why not change this to "This is not gonna end well"?

core/modules/menu/tests/menu_language.test
```diff
@@ -89,7 +89,7 @@ class MenuLanguageTestCase extends BackdropWebTestCase {
     // First remove all existing menu links.
     db_truncate('menu_links')->execute();
 
-    // Add some (unnested) items to main menu.
+    // Add some (un-nested) items to main menu.
```
klonos: Legit: https://en.wiktionary.org/wiki/unnested

core/modules/node/tests/node.test
```diff
@@ -2032,7 +2032,7 @@ class NodeTypePersistenceTestCase extends BackdropWebTestCase {
     $this->assertFalse($config->isNew(), 'Test node type found in config.');
     $this->assertEqual($config->get('disabled'), 0, 'Test node type is not disabled');
 
-    // Check that test node type (uncustomized) shows up.
+    // Check that test node type (un-customized) shows up.
```
klonos: Legit: https://en.wiktionary.org/wiki/uncustomized

core/modules/node/views/views_handler_filter_history_user_timestamp.inc
```diff
@@ -55,9 +55,8 @@ class views_handler_filter_history_user_timestamp extends views_handler_filter {
       return;
     }
 
-    // Hey, Backdrop kills old history, so nodes that haven't been updated
-    // since NODE_NEW_LIMIT are bzzzzzzzt outta here!
```
klonos, line 59: 🤣 ...we should change the language in all comments to this!

core/modules/simpletest/tests/common.test
```diff
@@ -1500,7 +1500,7 @@ class CommonJavaScriptTestCase extends BackdropWebTestCase {
       '<script src="' . file_create_url('core/misc/ajax.js') . '?' . $default_query_string . '"></script>',
       '<script src="' . file_create_url('core/misc/autocomplete.js') . '?' . $default_query_string . '"></script>',
     ));
-    $this->assertTrue(strpos($javascript, $expected) > 0, t('Unaggregated JavaScript is added in the expected group order.'));
+    $this->assertTrue(strpos($javascript, $expected) > 0, t('Un-aggregated JavaScript is added in the expected group order.'));
```
klonos: Legit: https://en.wiktionary.org/wiki/unaggregated

core/modules/simpletest/tests/menu.test
```diff
@@ -772,7 +772,7 @@ class MenuLinksUnitTestCase extends BackdropWebTestCase {
     $this->assertMenuLinkParents($links, $expected_hierarchy);
 
     // Now delete 'child-2' directly from the database, simulating a database
-    // crash. 'child-1-2' will get reparented under 'child-1' based on its
+    // crash. 'child-1-2' will get re-parented under 'child-1' based on its
```
klonos: Legit: https://en.wiktionary.org/wiki/reparent

core/modules/simpletest/tests/simpletest.tests.info
```diff
@@ -822,7 +822,7 @@ file = file.test
 
 [FileNameMungingTest]
 name = File naming
-description = Test filename munging and unmunging.
+description = Test filename munging and un-munging.
```
klonos: Legit: https://en.wiktionary.org/wiki/unmunged

core/modules/system/system.tar.inc
```diff
@@ -185,7 +185,7 @@ class Archive_Tar
                     if ($data == "\37\213") {
                         $this->_compress = true;
                         $this->_compress_type = 'gz';
-                        // No sure it's enought for a magic code ....
+                        // No sure it's enough for a magic code ....
```
klonos:
```suggestion
                        // Not sure it's enough for a magic code.
```

core/modules/views/tests/views_groupby.test
```diff
@@ -288,7 +288,7 @@ class ViewsQueryGroupByTest extends ViewsSqlTest {
     $handler->display->display_options['fields']['type']['hide_empty'] = 0;
     $handler->display->display_options['fields']['type']['empty_zero'] = 0;
     $handler->display->display_options['fields']['type']['link_to_node'] = 0;
-    /* Filtrar: Nodo: Nid */
+    /* Filter: Node: Nid */
```
klonos:
```suggestion
    // Filter: Node: NID.
```

All comments above should also be updated to start with `//` and end with a dot.

core/modules/views_ui/css/views_ui.admin.theme.css
```diff
@@ -776,11 +776,11 @@ td.group-title {
 
 /* @group Attachment bucket overridden
  *
- * Applies a overriden(italics) font style to overridden buckets.
+ * Applies a overridden (italics) font style to overridden buckets.
```
klonos:
```suggestion
 * Applies an overridden font style (italics) to overridden buckets.
```

core/modules/views_ui/js/views-admin.js
```diff
@@ -367,7 +367,7 @@ $.extend(Backdrop.viewsUi.OptionsSearch.prototype, {
     words = search.split(' ');
     wordsLength = words.length;
 
-    // Start the counter for restriping rows.
+    // Start the counter for re-striping rows.
```
klonos: Legit: https://en.wiktionary.org/wiki/restriping

core/modules/views_ui/views_ui.admin.inc
```diff
@@ -2503,7 +2503,7 @@ function views_ui_standard_form_buttons(&$form, &$form_state, $form_id, $name =
   // Finally, we never want these cached -- our object cache does that for us.
   $form['#no_cache'] = TRUE;
 
-  // If this isn't an ajaxy form, then we want to set the title.
+  // If this isn't an ajax form, then we want to set the title.
```
klonos:
```suggestion
  // If this isn't an AJAX form, then we want to set the title.
```

core/modules/views_ui/wizards/views_ui_base_views_wizard.php
```diff
@@ -763,7 +763,7 @@ protected function default_display_sorts_user($form, $form_state) {
     // Don't add a sort if there is no form value or the user selected none as sort.
     if (!empty($form_state['values']['show']['sort']) && $form_state['values']['show']['sort'] != 'none') {
       list($column, $sort) = explode(':', $form_state['values']['show']['sort']);
-      // Column either be a column-name or the table-columnn-ame.
+      // Column either be a column-name or the table-column-name.
```
klonos: Isn't there a "should" or "must" or something missing from this sentence?

core/modules/file/file.field.inc
avpaderno: Either that, or *the field's maximum upload size*.

core/includes/color.inc
ghost: I decided to use the word 'expand' instead of convert, because then it still makes sense if you remove the example.

core/includes/bootstrap.inc
ghost: Since this is a programming term, and since at least one other place on the internet uses it and considers it a word, let's use it and I'll add it to Backdrop's dictionary.

core/includes/diff.inc
ghost: @kiamlaluno I have been tossing back and forth with this file. Whether to include it or not... But this is written towards the top of the file, so I think that means it's been changed enough that we should be responsible for spelling mistakes in it:

> Backdrop's implementation is directly adopted from https://git.wikimedia.org/raw/mediawiki%2fcore.git/HEAD/includes%2fdiff%2fDairikiDiff.php
> Code has been reformatted to match Backdrop standards, and MediaWiki-specific profiling code has been removed.

core/includes/diff.inc
ghost: Maybe not, but considering it's used a few times throughout core, I'd suggest that might belong to a different issue (it's not a spelling mistake).

core/includes/install.inc
ghost: Good point.

core/includes/theme.inc
ghost: @klonos Are you suggesting to change it to the following?

> Check for existing and valid variable preprocessors.

core/modules/file/file.field.inc
ghost: It does, since this function is the validator for the 'Maximum upload field':

image

I'll add quotes to make it more obvious, but I think the wording is correct.

core/modules/layout/tests/layout.test
ghost: I feel like this might be a quote from the aforementioned character, so didn't want to go changing the actual words...

core/modules/node/views/views_handler_filter_history_user_timestamp.inc
ghost: :stuck_out_tongue_closed_eyes:

core/modules/file/file.field.inc
avpaderno: In that case, it should be *Render API callback: Validates the "Maximum upload size" field.* It is not a typo, so it is not important for this issue.

core/includes/theme.inc
klonos: @BWPanda yes 👍🏼 ...either that, or something this:
> Check for existing variable preprocessors, and ensure they are in the expected format.

core/includes/diff.inc
bugfolder: Grammatically, one uses a hyphen when the two words are used as an adjective to modify another, but no hyphen if the adjective is modifying a noun. Thus:

* Ignore changes in white space.
* That was a white-space change.

core/includes/theme.inc
```diff
@@ -591,7 +591,8 @@ function _theme_process_registry(&$cache, $name, $type, $theme, $path) {
 
       // Allow variable preprocessors for all theme hooks, whether the hook is
       // implemented as a template or as a function.
-      // Check for existing variable preprocessors. Ensure arrayness.
+      // Check for existing variable preprocessors, and ensure they're in the
```
bugfolder:
```suggestion
      // Check for existing variable preprocessors and ensure they're in the
```

For short independent clauses, you don't need a comma between them.

core/misc/backdrop.js
```diff
@@ -760,7 +760,7 @@ Backdrop.featureDetect.flexbox = function() {
  * Example use:
  * @code
  *   Backdrop.optimizedResize.add(function() {
- *     console.log('Smooth AND resource effecient!');
+ *     console.log('Smooth AND resource efficient!');
```
bugfolder:
```suggestion
 *     console.log('Smooth AND resource-efficient!');
```

Here the noun is used as a modifier, so a hyphen would be appropriate.

core/modules/simpletest/backdrop_web_test_case_cache.php
```diff
@@ -159,7 +159,9 @@ protected function tearDown() {
   /**
    * Alter tables to MyISAM engine to speed up tests.
    *
-   * MyISAM is faster to delete and copy tables. It gives small adventage when /var/lib/mysql on SHM (memory) device, but much bigger when tests run on regular device.
+   * MyISAM is faster to delete and copy tables. It gives small advantage when
+   * /var/lib/mysql on SHM (memory) device, but much bigger when tests run on
+   * regular device.
```
bugfolder:
```suggestion
   * MyISAM is faster to delete and copy tables. Its advantage is small when 
   * using /var/lib/mysql on an SHM (memory) device but is much bigger when
   * tests are run on a regular device.
```

core/modules/simpletest/tests/common.test
```diff
@@ -599,7 +599,7 @@ class CommonSizeUnitTestCase extends BackdropUnitTestCase {
       $size = 79691776,
       $string . ' == ' . $parsed_size . ' bytes'
     );
-    $string = '76.24 Giggabyte'; // Misspeld text -> 76.24 GB
+    $string = '76.24 Giggabyte'; // Misspelt text -> 76.24 GB
```
bugfolder:
```suggestion
    $string = '76.24 Giggabyte'; // Misspelled text -> 76.24 GB
```

core/modules/update/tests/update.test
```diff
@@ -281,7 +281,7 @@ class UpdateCoreTestCase extends UpdateTestHelper {
    */
   function testGetUpdaterFromDirectory() {
     module_load_include('inc', 'installer', 'installer.manager');
-    // Extract our stub sample backdop zip archive to temporary.
+    // Extract our stub sample backdrop zip archive to temporary.
```
bugfolder:
```suggestion
    // Extract our stub sample Backdrop zip archive to temporary.
```

I think the usage here is "Backdrop" as the proper noun (even though the file name is lower-case).

core/modules/views/handlers/views_handler_field.inc
```diff
@@ -1232,7 +1232,7 @@ If you would like to have the characters \'[\' and \']\' please use the html ent
         $more_link_path = $this->options['alter']['more_link_path'];
         $more_link_path = strip_tags(decode_entities(strtr($more_link_path, $tokens)));
 
-        // Take sure that paths which was runned through url() does work as well.
+        // Make sure that the path which was run through url() works as well.
```
bugfolder:
```suggestion
        // Make sure that the path that was run through url() works as well.
```

This is a defining clause, so use "that." ("Which" clauses are always optional.)

core/modules/views/handlers/views_handler_field_entity.inc
```diff
@@ -23,12 +23,12 @@ class views_handler_field_entity extends views_handler_field {
   public $entity_type;
 
   /**
-   * Stores all entites which are in the result.
+   * Stores all entities which are in the result.
```
bugfolder:
```suggestion
   * Stores all entities that are in the result.
```

core/modules/views/handlers/views_handler_field_machine_name.inc
```diff
@@ -5,7 +5,7 @@
  */
 
 /**
- * Field handler whichs allows to show machine name content as human name.
+ * Field handler which allows showing machine name content as human names.
```
bugfolder:
```suggestion
 * Field handler that allows showing machine name content as human names.
```

core/modules/views/includes/view.inc
```diff
@@ -144,7 +144,7 @@ class view {
   // To avoid recursion in views embedded into areas.
   var $parent_views = array();
 
-  // Is the current stored view runned as an attachment to another view.
+  // Is the current stored view run as an attachment to another view.
```
bugfolder:
```suggestion
  // Is the current stored view run as an attachment to another view?
```

When phrased as a question (like this), it should end in a question mark.

core/modules/views/plugins/views_plugin_query_default.inc
```diff
@@ -170,7 +159,7 @@ class views_plugin_query_default extends views_plugin_query {
    * based on $pure_distinct.
    *
    * @param bool $value
-   *   Should the view by distincted.
+   *   Should the view by distinct.
```
bugfolder:
```suggestion
   *   Should the view be distinct?
```

core/modules/views/plugins/views_plugin_query_default.inc
```diff
@@ -170,7 +159,7 @@ class views_plugin_query_default extends views_plugin_query {
...
    * @param bool $pure_distinct
    *   Should only the sql keyword be added.
```
bugfolder:
```suggestion
   *   Should only the SQL keyword be added?
```

core/modules/views/tests/views_groupby.test
```diff
@@ -235,8 +235,8 @@ class ViewsQueryGroupByTest extends ViewsSqlTest {
   }
 
   public function testGroupByCountOnlyFilters() {
-    // Check if GROUP BY and HAVING are included when a view
-    // Doesn't display SUM, COUNT, MAX... functions in SELECT statment
+    // Check if GROUP BY and HAVING are included when a view.
+    // Doesn't display SUM, COUNT, MAX... functions in SELECT statement.
```
bugfolder:
```suggestion
    // Check if GROUP BY and HAVING are included when a view doesn't display
    // SUM, COUNT, MAX... functions in SELECT statement.
```

core/modules/simpletest/backdrop_web_test_case_cache.php
ghost: @bugfolder I think the point of this issue is to remove typos, not add more... :wink: 

> Its *adventage* is...

bugfolder: Doi!

core/modules/views/includes/view.inc
klonos: @bugfolder @BWPanda should we use "Whether ..." instead of asking a question here?

Also, isn't the past tense of "run" "ran"? ..although I'd prefer to use "added" or "loaded" instead, since views don't "run" - the queries that are generated by Views are executed. Right?

core/modules/views/plugins/views_plugin_query_default.inc
klonos: Same as my previous comment, should we use "Whether" here instead of asking a question?

It would be nice if we were specific/explicit about the values as well. Something like this perhaps?:

`   *   Whether the view is distinct (TRUE) or not (FALSE)`

or this?:

`   *   TRUE if the view should be distinct; FALSE if not.`

core/modules/views/plugins/views_plugin_query_default.inc
klonos: Same as previous comments re using "whether" instead of asking a question.

core/modules/views/includes/view.inc
bugfolder: In fact, "run" is unneeded:

```
// Is the current stored view an attachment to another view?
```

But yes, "Whether" is generally better than a question, so:

```
  // Whether the current stored view is an attachment to another view.
```

core/modules/views/plugins/views_plugin_query_default.inc
bugfolder: Since it's declared as a bool, saying "TRUE or FALSE" is a bit unnecessary. It could be made clearer if we changed the variable name. Your point about "Whether" is good. I suggest:

```
   /**
   * Set the view to be distinct.
   *
   * There are either distinct per base field or distinct in the pure sql way,
   * based on $pure_distinct.
   *
   * @param bool $is_distinct
   *   Whether the view is distinct.
   * @param bool $pure_distinct
   *   Whether only the SQL keyword be added.
   */
  function set_distinct($is_distinct = TRUE, $pure_distinct = FALSE) {
    if (!(isset($this->no_distinct) && $is_distinct)) {
      $this->distinct = $is_distinct;
      $this->pure_distinct = $pure_distinct;
    }
  }
```

core/modules/views/plugins/views_plugin_query_default.inc
bugfolder: Yes, suggested rewrite of the function above includes this.

core/modules/views/includes/view.inc
avpaderno: > Also, isn't the past tense of "run" "ran"?

It is the Past Participle that is used as adjective, in this case _run_.

*Whether the current stored view is an attachment to another view.* makes sense for me, together *Whether the current stored view is used as attachment to another view.* (Maybe it is *as attachment for another view*?)

core/modules/views/plugins/views_plugin_query_default.inc
ghost: @bugfolder I don't agree with changing the variable name here. For a function called `set_distinct()`, I think the variable name `$value` makes sense (because it's the value you're setting).

bugfolder: OK. The rename better sense to me, but you did the heavy lifting on this one (really heavy!), so I can live with it.

quicksketch:
Merged into 1.x! And it cherry-picked perfectly back into 1.24.x as well. Nice! Thanks @BWPanda, @klonos, @kiamlaluno, and @bugfolder!

PR #4351, opened 2023-02-22
Issue #3911: Display reminder to uninstall disabled modules.

### LINKED ISSUE #3911: [UX] Display a reminder to uninstall modules after disabling them
After a module is disabled, we should redirect people to the uninstall page if that module ~~has an uninstall hook~~ needs to be uninstalled. If it does not need to be uninstalled, we can reload the modules list (as we do now).

In order to do this, we would need to test for the ability to uninstall specific modules, and set a `#redirect` on the form conditionally. 

I don't think this will be too hard, so adding the good first issue tag. If it turns out to be difficult we can remove it.

ghost:
Fixes https://github.com/backdrop/backdrop-issues/issues/3911


stpaultim:
Closing in favor of newer PR.

PR #4355, opened 2023-02-22
Issue #5893: Fix Coding Standards test not handling large number of files. [skip tests]

### LINKED ISSUE #5893: GHA: action thenabeel/action-phpcs seems unmaintained
## Description of the bug

Not actually sure, what's going on, but [one PR](https://github.com/backdrop/backdrop/pull/4282) just doesn't pass the phpcs check - for no obvious reason. And the phpcs output is odd. No annotations, but still a failure.

All the information we get is:

```
Error: Error: Command failed: phpcs --report=json -q --encoding=undefined --standard=../phpcs/Backdrop ...
```
That reads like: "something went wrong, lol". Not really helpful.

What's different in this PR compared to the other (passing) ones: A lot more files changed (35). But that doesn't explain anything.

Now the problem: the action thenabeel/action-phpcs, which we use to only run checks on actually changed or added code seems unmaintained (by @thenabeel). It doesn't even have an issue queue enabled. And it's a fork of an also unmaintained action.

The action still runs based on [nodejs 12](https://github.com/thenabeel/action-phpcs/blob/master/action.yml#L25), which is deprecated in GHA as of September this year.

https://github.blog/changelog/2022-09-22-github-actions-all-actions-will-begin-running-on-node16-instead-of-node12/
Not sure, if this has impact here.

The action provides *very little* options to debug, so it's tricky to figure out what fails, anyway.
What are our options:

- Ignore failures like that, which means, we're back to manual review.
- Drop the whole phpcs check - same problem as above: exhausting manual review.
- Search for an alternative action (there seems to be none, that's maintained).
- Write and maintain the whole check ourselves (not that we really have resources for that)
- ... did I miss anything?

ghost:
Fixes https://github.com/backdrop/backdrop-issues/issues/5893

### DIFF
--- a/.github/workflows/phpcs-lint.yml
+++ b/.github/workflows/phpcs-lint.yml
@@ -31,8 +31,9 @@ jobs:
         with:
           fetch-depth: 0
 
-      - uses: thenabeel/action-phpcs@v8
-        with:
-          files: "**.php,**.inc,**.module,**.install,**.profile,**.test"
-          standard: ../phpcs/Backdrop
-          fail_on_warnings: false
+      - name: Get changed files
+        id: changed-files
+        uses: tj-actions/changed-files@v35
+
+      - name: Run CodeSniffer
+        run: phpcs -n --colors --standard=../phpcs/Backdrop --basepath=. --extensions=php,inc,module,install,profile,test ${{ steps.changed-files.outputs.all_changed_files }}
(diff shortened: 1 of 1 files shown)


ghost:
Hmm, looks like we hit a limit...

image

ghost:
Ah, that's better :slightly_smiling_face:

PR #4358, opened 2023-02-24
Issue #3292: Indent taxonomy forms for checkboxes and radios.

### LINKED ISSUE #3292: [UX] Add proper indentation to taxonomy checkboxes/radios to denote child terms.
[RJL edit: advocate is @bugfolder.]

This is how a form widget looks like for a vocabulary with nested terms, when configured to be rendered as a checkbox list:

screen shot 2018-09-18 at 5 52 51 am

For D7, you can use the [Taxonomy Term Reference Tree Widget](https://www.drupal.org/project/term_reference_tree) module to improve how this widget works, but I believe that all the bell and whistles it offers would be too much for core. We should just get rid of the dashes from the checkbox labels and add indentation to the checkboxes. It would be so much better and more readable like so:

screen shot 2018-09-18 at 5 56 44 am

Related:
https://drupal.stackexchange.com/questions/85011/how-to-get-rid-of-the-sign-before-sub-taxonomy-term
https://drupal.stackexchange.com/questions/121053/add-class-to-sub-taxonomy-terms
https://www.drupal.org/project/drupal/issues/648264 (Core taxonomy.module issue, pending since 2009 - although for dropdowns)

bugfolder:
Fixes https://github.com/backdrop/backdrop-issues/issues/3292

### DIFF
--- a/core/modules/taxonomy/taxonomy.module
+++ b/core/modules/taxonomy/taxonomy.module
@@ -1751,6 +1751,9 @@ function taxonomy_field_formatter_settings_summary($field, $instance, $view_mode
  *   The field definition.
  * @param bool $indent
  *   Whether to add indentation dots to the terms based on their depth.
+ *
+ * @since 1.x.x Added $indent parameter.
+ *
  * @return array
  *   The array of valid terms for this field, keyed by term id.
  */
--- a/core/modules/taxonomy/taxonomy.module
+++ b/core/modules/taxonomy/taxonomy.module
@@ -1862,20 +1862,18 @@ function taxonomy_field_widget_form(&$form, &$form_state, $field, $instance, $la
  */
 function taxonomy_field_widget_options_buttons_form_alter(&$element, &$form_state, $context) {
   if ($context['field']['type'] == 'taxonomy_term_reference') {
-    $option_extras = array();
+    $options = taxonomy_allowed_values($context['field'], FALSE);
     foreach ($context['field']['settings']['allowed_values'] as $tree) {
       if ($vocabulary = taxonomy_vocabulary_load($tree['vocabulary'])) {
         if ($terms = taxonomy_get_tree($vocabulary->machine_name, $tree['parent'])) {
           foreach ($terms as $term) {
-            $option_extras[$term->tid] = array(
-              '#field_prefix' => theme('indentation', array('size' => $term->depth)),
-            );
+            if (isset($options[$term->tid])) {
+              $element['#options'][$term->tid] =  theme('indentation', array('size' => $term->depth)) . check_plain($options[$term->tid]);
+            }
           }
         }
       }
     }
-    $element += $option_extras;
-    $element['#options'] = taxonomy_allowed_values($context['field'], FALSE);
   }
 }
 
--- a/core/modules/taxonomy/taxonomy.module
+++ b/core/modules/taxonomy/taxonomy.module
@@ -1867,7 +1867,9 @@ function taxonomy_field_widget_options_buttons_form_alter(&$element, &$form_stat
       if ($vocabulary = taxonomy_vocabulary_load($tree['vocabulary'])) {
         if ($terms = taxonomy_get_tree($vocabulary->machine_name, $tree['parent'])) {
           foreach ($terms as $term) {
-            $option_extras[$term->tid] = array('#attributes' => array('class' => array('taxonomy-term-depth-' . $term->depth)));
+            $option_extras[$term->tid] = array(
+              '#field_prefix' => theme('indentation', array('size' => $term->depth)),
+            );
           }
         }
       }
--- a/core/modules/taxonomy/taxonomy.module
+++ b/core/modules/taxonomy/taxonomy.module
@@ -1868,7 +1868,7 @@ function taxonomy_field_widget_options_buttons_form_alter(&$element, &$form_stat
         if ($terms = taxonomy_get_tree($vocabulary->machine_name, $tree['parent'])) {
           foreach ($terms as $term) {
             if (isset($options[$term->tid])) {
-              $element['#options'][$term->tid] =  theme('indentation', array('size' => $term->depth)) . check_plain($options[$term->tid]);
+              $element['#options'][$term->tid] = theme('indentation', array('size' => $term->depth)) . check_plain($options[$term->tid]);
             }
           }
         }
--- a/core/modules/taxonomy/taxonomy.module
+++ b/core/modules/taxonomy/taxonomy.module
@@ -1858,10 +1858,15 @@ function taxonomy_field_widget_form(&$form, &$form_state, $field, $instance, $la
  */
 function taxonomy_field_widget_options_buttons_form_alter(&$element, &$form_state, $context) {
   if ($context['field']['type'] == 'taxonomy_term_reference') {
-    if (!empty($context['field']['settings']['allowed_values'])) {
+    if (empty($context['field']['settings']['option_list_callback'] || $context['field']['settings']['option_list_callback'] == 'taxonomy_allowed_values')) {
+      // With the default callback, we can rely on 'allowed_values' to provide
+      // the vocabularies needed to compute depth.
       $trees = $context['field']['settings']['allowed_values'];
     }
     else {
+      // If a module has overridden 'option_list_callback', we can't rely on
+      // 'allowed_values' to provide the vocabularies, so we'll have to fetch
+      // them.
       $tids = array_keys($element['#options']);
       $vocabularies = db_query('
         SELECT DISTINCT vocabulary
--- a/core/modules/taxonomy/taxonomy.module
+++ b/core/modules/taxonomy/taxonomy.module
@@ -1757,7 +1757,7 @@ function taxonomy_field_formatter_settings_summary($field, $instance, $view_mode
  * @return array
  *   The array of valid terms for this field, keyed by term id.
  */
-function taxonomy_allowed_values($field, $indent = TRUE) {
+function taxonomy_allowed_values($field) {
   $options = array();
   foreach ($field['settings']['allowed_values'] as $tree) {
     if ($vocabulary = taxonomy_vocabulary_load($tree['vocabulary'])) {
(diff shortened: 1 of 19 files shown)


core/modules/taxonomy/taxonomy.module
```diff
@@ -1852,6 +1854,32 @@ function taxonomy_field_widget_form(&$form, &$form_state, $field, $instance, $la
...
+ * Implements hook_field_widget_form_alter().
+ */
+function taxonomy_field_widget_form_alter(&$element, &$form_state, $context) {
+  if ($context['field']['type'] == 'taxonomy_term_reference' && in_array($element['#type'], array(
+    'checkboxes',
+    'radios',
+  ))) {
+    $option_extras = array();
+    foreach ($context['field']['settings']['allowed_values'] as $tree) {
+      if ($vocabulary = taxonomy_vocabulary_load($tree['vocabulary'])) {
```
klonos, line 1886: I haven't checked everything in depth, but calling `taxonomy_vocabulary_load()` for each option seems a bit excessive. Can't we load the vocabulary once outside the `foreach` loop? Is there any situation where the options loaded are coming from a different vocabulary?

core/modules/taxonomy/taxonomy.module
```diff
@@ -1747,18 +1747,20 @@ function taxonomy_field_formatter_settings_summary($field, $instance, $view_mode
...
+ * @param array $field
  *   The field definition.
- * @return
+ * @param bool $indent
+ *   Whether to add indentation dots to the terms based on their depth.
+ * @return array
  *   The array of valid terms for this field, keyed by term id.
  */
-function taxonomy_allowed_values($field) {
+function taxonomy_allowed_values($field, $indent = TRUE) {
```
klonos: We should add an `@since` for this parameter in the docblock of the function. We can leave it as a placeholder if we don't know the exact version when this'll get in, and the core committer that merges is can adjust the version number.

core/modules/taxonomy/taxonomy.module
bugfolder: `taxonomy_vocabulary_load()` is also called in `taxonomy_allowed_values()` (which is where I cribbed this bit of code from) when the original list of options is built. `taxonomy_vocabulary_load()` calls `taxonomy_vocabulary_load_multiple()`, which statically caches its result, so I think this call is cheap here.

core/modules/taxonomy/taxonomy.module
bugfolder: Added.

core/modules/taxonomy/taxonomy.module
```diff
@@ -1852,6 +1857,32 @@ function taxonomy_field_widget_form(&$form, &$form_state, $field, $instance, $la
   return $element;
 }
 
+/**
+ * Implements hook_field_widget_form_alter().
+ */
+function taxonomy_field_widget_form_alter(&$element, &$form_state, $context) {
```
argiepiano: Instead of using the generic alter hook, I suggest you use `taxonomy_field_widget_options_buttons_form_alter()` and remove the check for `radios` and `checkboxes` in the if below. This hook is an implementation of `hook_field_widget_WIDGET_TYPE_form_alter()`. The widget `options_buttons` is what Backdrop uses for both radios and checkboxes.

core/modules/taxonomy/taxonomy.module
```diff
@@ -1852,6 +1857,32 @@ function taxonomy_field_widget_form(&$form, &$form_state, $field, $instance, $la
...
+  if ($context['field']['type'] == 'taxonomy_term_reference' && in_array($element['#type'], array(
```
argiepiano: This if can be simplified by removing the check for element type (see my comment above).

core/modules/taxonomy/taxonomy.module
```diff
@@ -1852,6 +1857,32 @@ function taxonomy_field_widget_form(&$form, &$form_state, $field, $instance, $la
...
+      if ($vocabulary = taxonomy_vocabulary_load($tree['vocabulary'])) {
+        if ($terms = taxonomy_get_tree($vocabulary->machine_name, $tree['parent'])) {
+          foreach ($terms as $term) {
+            $option_extras[$term->tid] = array('#attributes' => array('class' => array('taxonomy-term-depth-' . $term->depth)));
+          }
+        }
+      }
+    }
+    $element += $option_extras;
+    $element['#attached'] = array(
```
argiepiano: I suggest you don't attach css this way. This will "wipe out" all other css attachments done by other widget alter hooks.

Instead I suggest you use:

`$element['#attached']['css'][] = backdrop_get_path('module', 'taxonomy') . '/css/taxonomy.css';`

core/modules/taxonomy/taxonomy.module
argiepiano: Additionally, by doing this alter hook this way, you avoid the potential issue of messing up a custom taxonomy widget that happens to use element types `checkboxes` or `radios`. You really want to act when the site is using the widget provided by core for taxonomy selection.

bugfolder: That's pretty slick, thanks. Agreed that the narrower the hook's scope is, the better.

core/modules/taxonomy/taxonomy.module
bugfolder: Yes, thanks!

core/modules/taxonomy/taxonomy.module
bugfolder: Agreed.

core/modules/taxonomy/taxonomy.module
```diff
@@ -1852,6 +1857,26 @@ function taxonomy_field_widget_form(&$form, &$form_state, $field, $instance, $la
   return $element;
 }
 
+/**
+ * Implements hook_field_widget_options_buttons_form_alter().
+ */
+function taxonomy_field_widget_options_buttons_form_alter(&$element, &$form_state, $context) {
+  if ($context['field']['type'] == 'taxonomy_term_reference') {
+    $options = taxonomy_allowed_values($context['field'], FALSE);
+    foreach ($context['field']['settings']['allowed_values'] as $tree) {
```
argiepiano: This line assumes that the field info will always have `'allowed_values'` set. This is not always the case. It's possible to use `$field['settings']['options_list_callback']` instead, to specify a callback that's different from the default `taxonomy_allowed_values` to obtain the options. In those cases, `$field['settings']['allowed_values']` will be empty, or will be ignored.

bugfolder: I've added a check for the existence of `$context['field']['settings']['allowed_values']`, and if it's not present, added code to rebuild the essential bits of it. It seems a bit heavy (fuly loading all the taxonomy terms to get their vocabularies), and I'm thinking it might be doable with a single direct db query; so looking into that.

core/modules/taxonomy/taxonomy.module
```diff
@@ -1852,6 +1853,44 @@ function taxonomy_field_widget_form(&$form, &$form_state, $field, $instance, $la
...
 }
 
+/**
+ * Implements hook_field_widget_options_buttons_form_alter().
+ */
+function taxonomy_field_widget_options_buttons_form_alter(&$element, &$form_state, $context) {
+  if ($context['field']['type'] == 'taxonomy_term_reference') {
+    if (!empty($context['field']['settings']['allowed_values'])) {
+      $trees = $context['field']['settings']['allowed_values'];
+    }
```
argiepiano, line 1865: I don't understand why you are loading the trees from `['settings']['allowed_values']`. Is there any scenario when the options won't be available in `$element['#options']`? 

In any case, I think loading directly from `['settings']['allowed_values']` is not a good idea either, because a contrib module like `taxonomy_access` or `i18n_taxonomy` may have defined a callback within `['settings']['options_list_callback']` that is supposed to override the default way of getting options, or do some additional processing.  I believe `['settings']['allowed_values']` will still not be empty, even when there is a callback in `options_list_callback`. Which means that the code above will be ignoring that special callback and basically loading all the options in the "default" way provided by this module.

bugfolder: > Is there any scenario when the options won't be available in `$element['#options']`?

Not that I know of, but we need more than just the tids (which are in #options); we need the vocabularies themselves (which are in #allowed_values) in order to compute the term depth. (I originally took that code from `taxonomy_allowed_values()`, which was this:

```
function taxonomy_allowed_values($field) {
  $options = array();
  foreach ($field['settings']['allowed_values'] as $tree) {
    if ($vocabulary = taxonomy_vocabulary_load($tree['vocabulary'])) {
        foreach ($terms as $term) {
          $options[$term->tid] = str_repeat('· ', $term->depth) . $term->name;
        }
      }
    }
  }
  return $options;
}
```

I'd assumed (incorrectly, it appears) that if that callback was being overidden, that the `empty()` check above would detect whether I can rely on `allowed_values`, and save the extra db call(s) to get the term vocabularies.

At this point, though, it might be simpler to just use the current `else` clause for all cases to get the `$trees`. It's only one extra db call. On the other hand, for the vast majority of use cases, people won't be overriding the callback, so it would be save to use `allowed_values` to get the $trees.

What do you think?

bugfolder: Meanwhile, I've updated the check for the override to check for the callback, not `allowed_values`.

core/themes/seven/css/style.css
```diff
@@ -889,8 +889,15 @@ div.teaser-checkbox .form-item,
 .form-item input.form-radio,
 .form-item input.form-checkbox {
   vertical-align: middle;
+}
+[dir="ltr"] .form-item input.form-radio,
+[dir="ltr"] .form-item input.form-checkbox {
   margin-right: 5px;
```
klonos: Our coding standards require that we add `/* LTR */` for things that have a respective RTL:
```suggestion
  margin-right: 5px; /* LTR */
```

bugfolder:
Closing in favor of rebase onto 1.x with indentation support.

PR #4362, opened 2023-02-25
Issue #5997: Document the backdrop_install_schema('system') call

### LINKED ISSUE #5997: Document why the backdrop_install_schema('system') in system_install() is necessary
[`system_install()`](https://docs.backdropcms.org/api/backdrop/core%21modules%21system%21system.install/function/system_install/1) calls `backdrop_install_schema('system')`, which is usually done from Backdrop core before installing a module.

Since this is an exception, a comment in the code should remember us that.

avpaderno:
Fixes https://github.com/backdrop/backdrop-issues/issues/5997


quicksketch:
Merged into 1.x and 1.32.x. Thank you @avpaderno!

PR #4364, opened 2023-02-26
Issue 5999. Add indentation to checkboxes, radios and select FAPI ele…

### LINKED ISSUE #5999: [DX] Add `#indentation` property to FAPI elements.
## Description of the need

This request arose from #3292. The solution proposed there by @bugfolder alters to the core radio/checkbox widget form used for displaying the taxonomy term widget.  But there was discussion about the possibility of adding a `#indentation` property to make this feature available to all checkbox and radio elements beyond the taxonomy term widget.

## Proposed solution

Adding a `#indentation` property that would further theme the element using `theme_indentation()`. This function adds `<div>` elements with css class `indentation` to create indentation. 

The new `#indentation` property would take an integer >= 0. 

The alternative is to manually add indentations to the `#options` values before outputting the form element. This is the current solution used in #3292. 

Alternatively, it may be possible to add different css classes to each element depending on their depths. This was an earlier approach in #3292, but it was cumbersome, as it required the creation of several css classes, one for each level.

I will try to provide a PR later today.

argiepiano:
Fixes https://github.com/backdrop/backdrop-issues/issues/5999


core/modules/system/css/system.css
```diff
@@ -138,7 +138,6 @@ a.tabledrag-handle-hover .handle {
 }
 div.indentation {
   float: left; /* LTR */
-  height: 1.7em;
```
klonos: Nice 👍🏼 ...I also wondered why we are using this when I saw it in the output of menu/taxonomy reorder-able tables.

core/modules/system/css/system.css
```diff
@@ -148,6 +148,14 @@ div.indentation {
...
   padding: 1em 0.6em 1em 0;
 }
+div.form-item-indentation {
+  display: inline-block;
+  width: 1em;
+}
+[dir="rtl"] div.form-item-indentation {
+  display: inline-block;
+  width: 1em;
+}
```
quicksketch: This can be removed because it's identical to the LTR version. LTR/RTL versions are only needed when left/right padding/margin/border/etc are affected. A 1em inline-block space will work the same regardless of text direction.

```suggestion
```

quicksketch:
Merged into 1.x for 1.25.0. Great work @argiepiano!

PR #4367, opened 2023-02-26
Issue 6000. Document use of FALSE in node_load_multiple.

### LINKED ISSUE #6000: Update docblock for `node_load_multiple()` to explain use of `FALSE`as parameter 1.
This is a follow-up #5437 (which won't be fixed). 

Currently, the doc block for `node_load_multiple()` doesn't explain that using FALSE as the first parameter will retrieve all nodes in the site. The default value for parameter 1 (an empty array) will actually return NULL, which seems counter-intuitive, but apparently can't be fixed in version 1. This isn't explained either.

argiepiano:
Fixes https://github.com/backdrop/backdrop-issues/issues/6000.

### DIFF
--- a/core/modules/node/node.module
+++ b/core/modules/node/node.module
@@ -856,15 +856,16 @@ function node_invoke($node, $hook, $a2 = NULL, $a3 = NULL, $a4 = NULL) {
  * from the database. Nodes are loaded into memory and will not require database
  * access if loaded again during the same page request.
  *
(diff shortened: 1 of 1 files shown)


ghost:
Closing in favour of https://github.com/backdrop/backdrop/pull/4283 for issue https://github.com/backdrop/backdrop-issues/issues/5889

PR #4368, opened 2023-02-27
Document $require_settings

### LINKED ISSUE #6002: Document `$require_settings` for `find_conf_path()`.
[`find_conf_path()`](https://docs.backdropcms.org/api/backdrop/core%21includes%21bootstrap.inc/function/find_conf_path/1) accept `$require_settings` as last parameter, but that parameter is not documented.
From [`conf_path()`](https://docs.backdropcms.org/api/backdrop/core%21includes%21bootstrap.inc/function/conf_path/1), the function that calls `find_conf_path()`, that description could be  the following one.

> bool $require_settings: Only configuration directories with an existing [settings.php](https://docs.backdropcms.org/api/backdrop/settings.php/1) file will be recognized. Defaults to TRUE. During initial installation, this is set to FALSE so that Backdrop can detect a matching directory, then create a new [settings.php](https://docs.backdropcms.org/api/backdrop/settings.php/1) file in it.

Alternatively, it can be slightly simplified to:

> bool $require_settings: Only configuration directories with an existing [settings.php](https://docs.backdropcms.org/api/backdrop/settings.php/1) file will be recognized. Defaults to TRUE.

It should not be necessary to repeat that, on initial installation, the function is called with `$require_settings` set to `FALSE`.

avpaderno:
Fixes https://github.com/backdrop/backdrop-issues/issues/6002


klonos (approved):
LGTM 👍🏼

quicksketch:
Merged into 1.x and 1.24.x. Thanks @kiamlaluno and @klonos!

PR #4369, opened 2023-02-28
Issue #6004: Fix coding standards.

ghost:


ghost:
The test failure is a bug: https://github.com/backdrop-ops/phpcs/issues/4

yorkshire-pudding (approved):
I have reviewed all the changes, noted the phpcs error and the related bug that incorrectly causes that error.  All changes look good. I note all functional tests have passed.

herbdool:
In core we are capitalizing `NULL` so it should be (for example): `@param array|NULL`.

herbdool:
Maybe that also means the Coding Standards needs a fix? Not sure if that's possible. But it should not accept lowercase `null`.

herbdool:
Looks like it's incorrect here?: https://github.com/backdrop-ops/phpcs/blob/a588dc22a117593b85cc52eec7d472f4698e2a54/Backdrop/Sniffs/Commenting/FunctionCommentSniff.php#L34. If I understand correctly, it's saying it should be `null` not `NULL` and `false` / `true` not `FALSE` / `TRUE`. But I don't think that is correct.

yorkshire-pudding:
@herbdool - I may be wrong, but this bit of docs - https://docs.backdropcms.org/doc-standards#types - suggests that this is indeed how **types** should be written.
I think TRUE, FALSE and NULL relate to values or constants - https://docs.backdropcms.org/php-standards#naming-conventions

ghost:
@yorkshire-pudding Correct. I've explained in @herbdool's phpcs issue.

herbdool:
Interesting, though the more recent code has been using all-caps, though maybe not consistently. I'll move my response to the issue. https://github.com/backdrop-ops/phpcs/issues/5

herbdool:
Toggling to trigger tests again.

herbdool:
phpcs has the error: Unknown type hint "FileTransfer" found for $filetransfer

PR #4370, opened 2023-02-28
Issue #6004: Fix coding standards.

ghost:

### DIFF
--- a/core/includes/bootstrap.inc
+++ b/core/includes/bootstrap.inc
@@ -288,17 +288,18 @@ if (version_compare(PHP_VERSION, BACKDROP_MINIMUM_PHP) < 0) {
  * static caches of that same data.
  *
  * Note that array_* functions do not work with ArrayAccess. Systems using
- * BackdropCacheArray should use this only internally. If providing API functions
- * that return the full array, this can be cached separately or returned
- * directly. However since BackdropCacheArray holds partial content by design, it
- * should be a normal PHP array or otherwise contain the full structure.
+ * BackdropCacheArray should use this only internally. If providing API
+ * functions that return the full array, this can be cached separately or
+ * returned directly. However since BackdropCacheArray holds partial content by
+ * design, it should be a normal PHP array or otherwise contain the full
+ * structure.
  *
  * Note also that due to limitations in PHP prior to 5.3.4, it is impossible to
  * write directly to the contents of nested arrays contained in this object.
  * Only writes to the top-level array elements are possible. So if you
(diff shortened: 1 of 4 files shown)


core/includes/bootstrap.inc
```diff
@@ -3307,15 +3333,15 @@ function _backdrop_bootstrap_sanitize_request() {
 /**
  * Sanitizes unsafe keys from request input.
  *
- * @param mixed $input
+ * @param array $input
  *   The input data array from a request to be sanitized.
- * @param string[] $allowlist
+ * @param array $allowlist
```
herbdool, line 3338: I believe `string[]` is fine here and elsewhere. `array` would be less precise. From what I understand `string[]` means an array of strings.

core/includes/bootstrap.inc
```diff
@@ -3545,17 +3571,17 @@ function language_multilingual() {
...
+ * @param bool $option_list
  *   (optional) Whether to return a list of options instead of objects. This
  *   list is suitable for use in a select list #options array. This list is
  *   ordered by weight and then language name.
  * @param bool $native_names
  *   (optional) Whether the option_list labels should use the native name
  *   instead of the English language name. Only applies if $option_list is TRUE.
  *
- * @return (stdClass|string)[]
+ * @return array
```
herbdool, line 3584: I think this can be instead: `(object|string)[]`.

I found this reference: https://docs.phpdoc.org/3.0/guide/references/phpdoc/types.html

core/includes/bootstrap.inc
```diff
@@ -1521,13 +1528,13 @@ function backdrop_load($type, $name) {
  * Note: When sending a Content-Type header, always include a 'charset' type,
  * too. This is necessary to avoid security bugs (e.g. UTF-7 XSS).
  *
- * @param $name
+ * @param string $name
  *   The HTTP header name, or the special 'Status' header name.
- * @param $value
+ * @param string|false $value
```
herbdool, line 1533: I'm not sure about having `false` or `true` in `@param`. It might be allowed by our standards but I think I read somewhere how it can be set for the return value: `@return false` but not for `@param`. But now I can't find it or maybe it was just my impression. 

Maybe https://docs.phpdoc.org/3.0/guide/guides/types.html#supported-types?

herbdool: Maybe it's okay. I still haven't ready anything either way so I assume it might be ok.

https://github.com/php-fig/fig-standards/blob/master/proposed/phpdoc.md#53-tags

herbdool (changes requested):

PR #4373, opened 2023-03-03
Issue #5995: Adds context support to views blocks associated with layouts

### LINKED ISSUE #5995: Add context handling to Views blocks
Views blocks currently don't allow context to be passed in from a Layout. The only way to accomplish this now is to extract context from the URL based on position. This is less than ideal because it forces the block to be locked to a specific path structure that is separate from the Layout configuration.

Drupal 7 had the option to choose a data source "From context" in a View block. This code exists in Backdrop but has been commented out with a TODO. Perhaps this should be revisited to add context handling for Views blocks.

Update (April 3, 2023): The latest pull request for this issue adds context support to views blocks associated with layouts. There has been many revisions made to this pull request to address concerns from this wonderful Backdrop community. Some of these revisions have addressed improvements, efficiency changes, updated coding standards, etc. As of today, this PR has passed all the validation checks except one that seems to be some kind of external issue with that particular check.

Update (April 11, 2023): The latest pull request for this issue includes automated testing and is awaiting review. All validation tests have successfully passed.

---
Advocate for this feature: @rbargerhuff

rbargerhuff:
Fixes https://github.com/backdrop/backdrop-issues/issues/5995


core/modules/layout/layout.module
```diff
@@ -1543,13 +1542,22 @@ function layout_get_layout_info($layout_name = NULL) {
...
  * Load all layout context information from modules.
  *
+ * @param string $plugin_name
+ *   The name of the plugin. If no plugin is specified, all context info will
+ *   be returned.
+ *
  * @return array
  */
-function layout_get_context_info($plugin_name) {
+function layout_get_context_info($plugin_name = NULL) {
```
docwilmot: This change isn't really needed though, see below.

core/modules/layout/layout.module
```diff
@@ -1543,13 +1542,22 @@ function layout_get_layout_info($layout_name = NULL) {
...
   $info = _layout_get_all_info('layout_context');
-  return isset($info[$plugin_name]) ? $info[$plugin_name] : FALSE;
-}
 
+  if (!$plugin_name) {
+    return $info;
+  }
+  else {
+    return isset($info[$plugin_name]) ? $info[$plugin_name] : FALSE;
```
docwilmot: You could achieve the same result by calling `_layout_get_all_info('layout_context')`.

core/modules/views/plugins/views_plugin_display_block.inc
```diff
@@ -266,6 +277,18 @@ class views_plugin_display_block extends views_plugin_display {
         $form['#title'] .= t('Choose the data sources for contextual filters');
         $argument_input = $this->get_argument_input();
 
+        $layout_context_info = layout_get_context_info();
+        $layout_context_options = array();
+
+        foreach ($layout_context_info as $key => $value) {
```
docwilmot, line 285: I think it may be best, instead of only including these two context types, we just excluded `LayoutOverridesPathContext`, and allow all others. That way things like dashboard context, and contrib contexts, can be used by Views blocks as well.

core/modules/layout/layout.module
rbargerhuff: We understand that this function exists, `_layout_get_all_info('layout_context')`, however this function's name indicates that the function is private. We did not feel it was a good idea to have the views module call a private function buried in the layout module.

We chose to modify the function `layout_get_context_info($plugin_name = NULL) ` because we felt it was more in-line with what the function is doing, in this case retrieving context info. 

We also concluded that because the function `layout_get_layout_info($layout_name = NULL)` has $layout_name optional, it would be okay for `layout_get_context_info($plugin_name)` to have its argument optional as well.

No existing implementation of the function would break.

All that being said, if calling the private function aligns with Backdrop CMS practices, we'll happily make the adjustment. :)

core/modules/views/plugins/views_plugin_display_block.inc
rbargerhuff: This is a great suggestion and we will try to implement it soon.

core/modules/layout/layout.module
docwilmot: Layout is a required module so I think realistically "private" functions wouldn't really mean much. And yes, private functions do seem to get called from other files so IMO I think this is OK. Would welcome other opinions though.

rbargerhuff: Thank you. This has been reverted back.

core/modules/layout/plugins/context/layout_context.inc
```diff
@@ -257,6 +259,9 @@ abstract class LayoutContext extends LayoutHandler {
           $this->setData($context_data);
         }
       }
+
+      // Set the raw argument in the context.
+      $this->argument = $router_original_map[$this->position];
```
docwilmot: Its extremely unlikely, but contrib could be calling `setDataFromRouterMap()` without this new parameter, so we should wrap this in conditions. 
```
if (isset($router_original_map)) {
      $this->argument = $router_original_map[$this->position];
}
```

Edit: if contrib is calling with one argument, it would also cause a fatal error, so would need `$router_original_map` to be set as optional (`$router_original_map = NULL`).

core/modules/views/plugins/views_plugin_display_block.inc
rbargerhuff: This is now implemented using a different method that should both be more efficient and respectful to the existing Context foundation that has been established. This new method will show all contexts defined in `hook_layout_context_info` where the hidden attribute is not set to `true`. This new method will cause the specific `LayoutOverridesPathContext` above to be excluded.

core/modules/layout/plugins/context/layout_context.inc
rbargerhuff: This has been addressed in the latest update to this PR.

PR #4374, opened 2023-03-03
Issue #6016: Add block/layout visibility condition based on HTTP response code

### LINKED ISSUE #6016: [D10] Layouts/blocks: Add visibility condition for 200/403/404 response codes
This is the respective issue as https://www.drupal.org/project/drupal/issues/2245767, and would be a good addition to the features of our layouts system.

> ## Problem/Motivation
> Some blocks, like user login, should be shown on 403/404 response pages.
> Others, like the help block, should not.
> 
> ## Proposed resolution
> Add a visibility setting to blocks that restricts visibility to any combination of 200/403/404 responses. Default this to show the block for all responses.

Perhaps in our case it also makes sense to add that visibility conditions to the layouts (besides blocks), so that if people really have the need for it, they can configure a different layout to be used based on response.

klonos:
Fixes https://github.com/backdrop/backdrop-issues/issues/6016

### DIFF
--- a/core/modules/layout/includes/layout.layout.inc
+++ b/core/modules/layout/includes/layout.layout.inc
@@ -168,6 +168,11 @@ function layout_layout_access_info() {
     'description' => t('Control access by the current path.'),
     'class' => 'PathLayoutAccess',
   );
+  $info['http_response_code'] = array(
+    'title' => t('Server response'),
+    'description' => t('Control access by the HTTP response status code being returned by the server.'),
+    'class' => 'HttpResponseCodeLayoutAccess',
+  );
   return $info;
 }
 
--- a/core/modules/layout/plugins/access/http_response_code_layout_access.inc
+++ b/core/modules/layout/plugins/access/http_response_code_layout_access.inc
@@ -3,11 +3,12 @@
  * @file
  * Class for loading, modifying, and saving a layout access rule (condition).
  */
+
 class HttpResponseCodeLayoutAccess extends LayoutAccessNegatable {
   /**
    * {@inheritdoc}
    */
-  function form(&$form, &$form_state) {
+  private function form(&$form, &$form_state) {
     parent::form($form, $form_state);
     $response_codes = isset($this->settings['http_response_codes']) ? $this->settings['http_response_codes'] : array();
 
@@ -47,7 +48,7 @@ class HttpResponseCodeLayoutAccess extends LayoutAccessNegatable {
   /**
    * {@inheritdoc}
    */
-  function checkAccess() {
+  private function checkAccess() {
     $allowed_status_codes = array_filter($this->settings['http_response_codes']);
 
     $router = menu_get_item();
@@ -74,7 +75,7 @@ class HttpResponseCodeLayoutAccess extends LayoutAccessNegatable {
   /**
    * {@inheritdoc}
    */
-  function summary() {
+  private function summary() {
     $allowed_status_codes = array_filter($this->settings['http_response_codes']);
 
     if (count($allowed_status_codes)) {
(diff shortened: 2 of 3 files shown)


core/modules/layout/includes/layout.layout.inc
```diff
@@ -168,6 +168,11 @@ function layout_layout_access_info() {
     'description' => t('Control access by the current path.'),
     'class' => 'PathLayoutAccess',
   );
+  $info['http_response_code'] = array(
+    'title' => t('Server response'),
+    'description' => t('Control access by the HTTP response status code being returned by the server.'),
```
jenlampton, line 173: we could shorten this up a little bit by removing `being`: `Control access by the HTTP response status code returned by the server`

core/modules/layout/plugins/access/http_response_code_layout_access.inc
```diff
@@ -0,0 +1,98 @@
+<?php
+/**
+ * @file
+ * Class for loading, modifying, and saving a layout access rule (condition).
```
jenlampton, line 4: Most of our other class files don't have both a file docblock and a class docblock. Should be consistent with the rest of core, or make the code pass code sniffer?

core/modules/layout/plugins/access/http_response_code_layout_access.inc
```diff
@@ -0,0 +1,98 @@
...
+      $checkboxes_description = t('Shows the block on pages with any matching response code. If nothing is selected, the block will be shown on all pages where this layout applies.');
+      $negate_description = t('Hides the block on pages with any matching response code selected from the options above.');
+    }
+    elseif (isset($form_state['layout'])) {
+      $checkboxes_description = t('This layout will apply on pages with any matching response code. If nothing is selected, the layout will apply on all pages (subject to other visibility conditions).');
+      $negate_description = t('This layout will apply on all pages, except those with any matching response code selected from the options above (subject to other visibility conditions).');
+    }
+    $form['http_response_codes'] = array(
+      '#type' => 'checkboxes',
+      '#title' => t('HTTP response status is'),
```
jenlampton, line 28: How about `HTTP response status code` so we don't end with "is"?

core/modules/layout/plugins/access/http_response_code_layout_access.inc
```diff
@@ -0,0 +1,98 @@
...
+    // "Access denied" and "Page not found" get "system/403" and "system/404" as
+    // their path respectively. See backdrop_deliver_html_page().
+    switch ($router_path) {
+      case 'system/403':
+        $current_status_code = 403;
+        break;
+      case 'system/404':
+        $current_status_code = 404;
+        break;
+      default:
```
jenlampton, line 68: What about the other response codes? maybe this should be a case sate also and not `default`. I also wonder if maybe we shouldn't be actually be checking the header, and not the system path determined by the header. Either that, or we remove 200 as one of the options.

core/modules/layout/plugins/access/http_response_code_layout_access.inc
```diff
@@ -0,0 +1,98 @@
...
+    if (count($allowed_status_codes)) {
+      if ($this->settings['negate']) {
+        $summary = format_plural(count($allowed_status_codes), 'Server response code is not @response', 'Server response code is not one of: @response', array('@response' => implode(', ', $allowed_status_codes)));
+      }
+      else {
+        $summary = format_plural(count($allowed_status_codes), 'Server response code is @response', 'Server response code is one of: @response', array('@response' => implode(', ', $allowed_status_codes)));
+      }
+    }
+    else {
+      $summary = t('Shown on all pages where this layout applies.');
```
jenlampton, line 93: This doesn't make sense for a layout. "Layout is shown where layout is shown". 

Instead, I think we should only add the visibility rule if it's configured to actually do something. Having a rule that does nothing doesn't really make sense. 

Let's add a `formValidate()` that checks to see that one of the options was checked and not save until it is.

quicksketch:
Replaced by https://github.com/backdrop/backdrop/pull/4855 which addresses all of @jenlampton's feedback.

PR #4376, opened 2023-03-05
Remove "return TRUE;"

### LINKED ISSUE #6019: Do not return TRUE from backdrop_session_regenerate()
[`backdrop_session_regenerate()`](https://docs.backdropcms.org/api/backdrop/core%21includes%21session.inc/function/backdrop_session_regenerate/1) returns `TRUE`, but none of the functions calling it uses the returned value, which is not even documented.
Even [`drupal_session_regenerate()`](https://api.drupal.org/api/drupal/includes%21session.inc/function/drupal_session_regenerate/7.x), the function that in Backdrop is just a wrapper for `backdrop_session_regenerate()` returns `TRUE` or `FALSE`.

The first `return TRUE;` in `backdrop_session_regenerate()` should be changed to `return;`, while the other ones should be removed.

avpaderno:
Fixes https://github.com/backdrop/backdrop-issues/issues/6019


quicksketch:
Thanks @kiamlaluno! I've merged into 1.x and 1.25.x.

PR #4379, opened 2023-03-12
Issue #6024: handle integer 0 key in #options for radios form element

### LINKED ISSUE #6024: #default_value may or may not work in #options for radios depending on PHP version, option order, and/or key of 0
## Description of the bug

In checkboxes and radios, if the `#options` property contains a key of `0` and the `#default_value` is set to a different option, whether that default value is applied depends on (a) the order of the options, (b) the PHP version.

To test this, put this code into a form and render it:

```
  $types = array(
    'select' => t('Select'),
    'radios' => t('Radios'),
    'checkboxes' => t('Checkboxes'),
  );
  $default_values = array(
    'select' => 'foo',
    'radios' => 'foo',
    'checkboxes' => array('foo'),
  );
  foreach ($types as $type => $name) {
    $form[$type] = array(
      '#type' => 'fieldset',
      '#title' => $name,
    );
    $form[$type]["{$type}_1"] = array(
      '#type' => $type,
      '#title' => t('Group 1'),
      '#options' => array(
        'foo' => t('Foo option'),
        0 => t('Zero option'),
      ),
      '#default_value' => $default_values[$type],
    );
    $form[$type]["{$type}_2"] = array(
      '#type' => $type,
      '#title' => t('Group 2'),
      '#options' => array(
        0 => t('Zero option'),
        'foo' => t('Foo option'),
      ),
      '#default_value' => $default_values[$type],
    );
  }
```

Here's the result under PHP 7.4.33. Note that in every case, `'foo'` is the `#default_value`.

options

If the `0` option is first, `#default_value` doesn't work in radios. If the `0` option is second, it works properly in radios. Selects and checkboxes all work.

That was in PHP 7.4. It works properly in PHP 8.1. Here's a summary of PHP versions I tested:

* 5.6.40: radios fail
* 7.4.33: radios fail
* 8.0.24: all work
* 8.1.13: all work

So the problem can be avoided by either (a) not putting your 0 key first, or (b) using an empty string, rather than 0, as the key. But it would be nice if the problem wasn't there, as it's $#@!! puzzling when it crops up (e.g., https://github.com/backdrop-contrib/ubercart/issues/436#issuecomment-1464941747).

## Additional information

- PHP version: various (see above)

bugfolder:
Fixes https://github.com/backdrop/backdrop-issues/issues/6024.


core/includes/form.inc
```diff
@@ -3001,7 +3001,7 @@ function theme_radio($variables) {
   $element['#attributes']['type'] = 'radio';
   element_set_attributes($element, array('id', 'name', '#return_value' => 'value'));
 
-  if (isset($element['#return_value']) && $element['#value'] !== FALSE && $element['#value'] == $element['#return_value']) {
+  if (isset($element['#return_value']) && $element['#value'] !== FALSE && !strcmp($element['#value'], $element['#return_value'])) {
```
klonos, line 3032: Can we please add an inline comment here, explaining the issue and why exactly we are using `strcmp` instead of a simple comparison operator?

bugfolder: Comment added.

TBH, [the documentation page for `#options` property](https://docs.backdropcms.org/form_api#options_property) should warn people off about mixing integer and string keys, because even apart from this one, someone, somewhere is going to try to compare two key values for some reason and encounter this peculiarity.

bugfolder: Warning has been added to the documentation page, so marking this as resolved.

quicksketch:
Merged into 1.x and 1.32.x. Thank you @bugfolder!

PR #4380, opened 2023-03-13
Issue #6026: Cleanup diff.inc Doxygen documentation. [skip tests]

### LINKED ISSUE #6026: Cleanup `diff.inc` Doxygen documentation
The [`diff.inc` file](https://github.com/backdrop/backdrop/blob/1.x/core/includes/diff.inc) has a few issues with its Doxygen documentation we could fix:

- `@file` tag isn't at the top of the file docblock
- `@ingroup` in file docblock doesn't make sense (as the group hasn't been defined yet)
- `@defgroup` group description is the same as the group's name (`DifferenceEngine`, so superfluous)
- `@private` tags aren't used anywhere else in core (remove)
- Adding `@ingroup` to each class in the file is pointless - we can just wrap the whole file in a proper `@defgroup` declaration instead:
  ```php
  /**
   * @defgroup DifferenceEngine
   * @{
   */

  // File contents...

  /**
   * @} End of "defgroup DifferenceEngine".
   */
  ```

ghost:
Fixes https://github.com/backdrop/backdrop-issues/issues/6026

### DIFF
--- a/core/includes/diff.inc
+++ b/core/includes/diff.inc
@@ -1,5 +1,6 @@
 <?php
 /**
+ * @file
  * A PHP diff engine for phpwiki. (Taken from phpwiki-1.3.3)
  *
  * Copyright © 2000, 2001 Geoffrey T. Dairiki <dairiki@dairiki.org>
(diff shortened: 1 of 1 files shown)


klonos (approved):
LGTM 👍🏼

quicksketch:

PR #4381, opened 2023-03-13
Issue #6028: FAPI: Follow-up to the addition of #indentation

### LINKED ISSUE #6028: [DX] FAPI improvements follow-up to the addition of the `#indentation` property
This is a follow-up + cleanup issue for #5999, which also slightly improves/optimizes things in certain aspects. The changes I am proposing are not all related directly to the change that went in with that issue, rather than cleaning up things I came across while fiddling in my local for a possible PR for that issue. I just wouldn't want these to be lost.

Suggested changes:
- Add attributes like `selected` and `value` to select elements properly (via `backdrop_attributes()`)
- Add an `indent` class, to allow targeting all indented elements en masse.
- Add a `data-indentation-depth` attribute to the same form elements that received the `form-item-indentation-*` class (this is the follow-up bit actually relevant to #5999). Since the current solution was based on CSS classes, it is limited to only handling indentation depths from 0 to 5, which is not as robust. I'm hoping that the addition of this new `data-*` attribute, which holds the indentation depth as its value, can be used by contrib and custom code to provide a better solution for indentation, either based on custom JS, or in core in the future (when there's better browser support for CSS calculations that include `data-*` attributes).

This issue is hard to describe - better provide a PR for it so that people can see it in code...

klonos:
Fixes https://github.com/backdrop/backdrop-issues/issues/6028

Follow-up to https://github.com/backdrop/backdrop-issues/issues/5999


core/includes/form.inc
```diff
@@ -2886,17 +2886,19 @@ function form_select_options($element, $choices = NULL) {
...
-      }
-      else {
-        $selected = '';
+        $option_attributes['selected'] = 'selected';
       }
 
       if (!empty($element[$key]['#indentation'])) {
         $choice = str_repeat('· ', $element[$key]['#indentation']) . $choice;
+        $option_attributes['data-indentation-depth'] = $element[$key]['#indentation'];
+        $option_attributes['class'] = 'indent';
```
quicksketch: Maybe be consistent with indent/indentation here:
```suggestion
        $option_attributes['class'] = 'indentation';
```

klonos: Fixed all 3 instances of this class with [`c275dc6` (#4381)](https://github.com/backdrop/backdrop/pull/4381/commits/c275dc628a1083ee94102d6ba56a7f80b3ce2502)

quicksketch:
I merged this into 1.x for 1.25.0. Thanks @klonos!

PR #4382, opened 2023-03-14
Issue #3292: Indent taxonomy forms for checkboxes and radios

### LINKED ISSUE #3292: [UX] Add proper indentation to taxonomy checkboxes/radios to denote child terms.

bugfolder:
Fixes https://github.com/backdrop/backdrop-issues/issues/3292.

### DIFF
--- a/core/modules/taxonomy/taxonomy.module
+++ b/core/modules/taxonomy/taxonomy.module
@@ -1751,6 +1751,9 @@ function taxonomy_field_formatter_settings_summary($field, $instance, $view_mode
  *   The field definition.
  * @param bool $indent
  *   Whether to add indentation dots to the terms based on their depth.
+ *
+ * @since 1.x.x Added $indent parameter.
+ *
  * @return array
  *   The array of valid terms for this field, keyed by term id.
  */
--- a/core/modules/taxonomy/taxonomy.module
+++ b/core/modules/taxonomy/taxonomy.module
@@ -1854,41 +1854,63 @@ function taxonomy_field_widget_form(&$form, &$form_state, $field, $instance, $la
 }
 
 /**
- * Implements hook_field_widget_options_buttons_form_alter().
+ * Implements hook_field_widget_WIDGET_TYPE_form_alter().
  */
 function taxonomy_field_widget_options_buttons_form_alter(&$element, &$form_state, $context) {
-  if ($context['field']['type'] == 'taxonomy_term_reference') {
-    if (empty($context['field']['settings']['option_list_callback']) || $context['field']['settings']['option_list_callback'] == 'taxonomy_allowed_values') {
-      // With the default callback, we can rely on 'allowed_values' to provide
-      // the vocabularies needed to compute depth.
-      $trees = $context['field']['settings']['allowed_values'];
+  $field = $context['field'];
+  if ($field['type'] == 'taxonomy_term_reference') {
+    _taxonomy_field_widget_options_process($field, $element);
+  }
+}
+
+/**
+ * Implements hook_field_widget_WIDGET_TYPE_form_alter().
+ */
+function taxonomy_field_widget_options_select_form_alter(&$element, &$form_state, $context) {
+  $field = $context['field'];
+  if ($field['type'] == 'taxonomy_term_reference') {
+    _taxonomy_field_widget_options_process($field, $element);
+  }
+}
+
+/**
+ * Adds indentation to the options items for taxonomy radios/checkboxes/selects.
+ *
+ * @see taxonomy_field_widget_options_buttons_form_alter()
+ * @see taxonomy_field_widget_options_select_form_alter()
+ */
+function _taxonomy_field_widget_options_process($field, &$element) {
+  if (empty($field['settings']['option_list_callback']) || $field['settings']['option_list_callback'] == 'taxonomy_allowed_values') {
+    // With the default callback, we can rely on 'allowed_values' to provide
+    // the vocabularies needed to compute depth.
+    $trees = $field['settings']['allowed_values'];
+  }
+  else {
+    // If a module has overridden 'option_list_callback', we can't rely on
+    // 'allowed_values' to provide the vocabularies, so we'll have to fetch
+    // them.
+    $tids = array_keys($element['#options']);
+    $vocabularies = db_query('
+      SELECT DISTINCT vocabulary
+      FROM {taxonomy_term_data}
+      WHERE tid IN ( :tids )
+      ', array(':tids' => $tids))
+      ->fetchCol();
+    $trees = array();
+    foreach ($vocabularies as $vocabulary) {
+      $trees[] = array(
+        'vocabulary' => $vocabulary,
+        'parent' => 0,
+      );
     }
-    else {
-      // If a module has overridden 'option_list_callback', we can't rely on
-      // 'allowed_values' to provide the vocabularies, so we'll have to fetch
-      // them.
-      $tids = array_keys($element['#options']);
-      $vocabularies = db_query('
-        SELECT DISTINCT vocabulary
-        FROM {taxonomy_term_data}
-        WHERE tid IN ( :tids )
-        ', array(':tids' => $tids))
-        ->fetchCol();
-      $trees = array();
-      foreach ($vocabularies as $vocabulary) {
-        $trees[] = array(
-          'vocabulary' => $vocabulary,
-          'parent' => 0,
-        );
-      }
-    }
-    foreach ($trees as $tree) {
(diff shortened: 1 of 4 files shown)


core/modules/taxonomy/taxonomy.module
```diff
@@ -1852,6 +1853,49 @@ function taxonomy_field_widget_form(&$form, &$form_state, $field, $instance, $la
...
+          foreach ($terms as $term) {
+            if (isset($element['#options'][$term->tid])) {
+              $element[$term->tid] = array('#indentation' => $term->depth);
+            }
+          }
+        }
+      }
+    }
+  }
+}
```
klonos, line 1897:
```suggestion
/**
 * Implements hook_field_widget_form_alter().
 */
function taxonomy_field_widget_form_alter(&$element, &$form_state, $context) {
  $field = $context['field'];
  $widget_type = $context['instance']['widget']['type'];

  if ($field['type'] == 'taxonomy_term_reference' && in_array($widget_type, array('options_buttons', 'options_select'))) {
    if (empty($field['settings']['option_list_callback']) || $field['settings']['option_list_callback'] == 'taxonomy_allowed_values') {
      // With the default callback, we can rely on 'allowed_values' to provide
      // the vocabularies needed to compute depth.
      $trees = $field['settings']['allowed_values'];
    }
    else {
      // If a module has overridden 'option_list_callback', we can't rely on
      // 'allowed_values' to provide the vocabularies, so we'll have to fetch
      // them.
      $tids = array_keys($element['#options']);
      $vocabularies = db_query('
        SELECT DISTINCT vocabulary
        FROM {taxonomy_term_data}
        WHERE tid IN ( :tids )
        ', array(':tids' => $tids))
        ->fetchCol();
      $trees = array();
      foreach ($vocabularies as $vocabulary) {
        $trees[] = array(
          'vocabulary' => $vocabulary,
          'parent' => 0,
        );
      }
    }

    foreach ($trees as $tree) {
      if ($vocabulary = taxonomy_vocabulary_load($tree['vocabulary'])) {
        if ($terms = taxonomy_get_tree($vocabulary->machine_name, $tree['parent'])) {
          foreach ($terms as $term) {
            if (isset($element['#options'][$term->tid])) {
              $element[$term->tid] = array('#indentation' => $term->depth);
            }
          }
        }
      }
    }
  }
}
```

core/modules/taxonomy/taxonomy.module
```diff
@@ -1852,6 +1853,49 @@ function taxonomy_field_widget_form(&$form, &$form_state, $field, $instance, $la
   return $element;
 }
 
+/**
+ * Implements hook_field_widget_options_buttons_form_alter().
+ */
+function taxonomy_field_widget_options_buttons_form_alter(&$element, &$form_state, $context) {
```
klonos: We should be using a `hook_field_widget_form_alter()` instead . Please see https://github.com/bugfolder/backdrop/pull/7/files.

core/modules/taxonomy/taxonomy.module
```diff
@@ -1852,6 +1853,53 @@ function taxonomy_field_widget_form(&$form, &$form_state, $field, $instance, $la
...
 
+/**
+ * Implements hook_field_widget_form_alter().
+ */
+function taxonomy_field_widget_form_alter(&$element, &$form_state, $context) {
+  $field = $context['field'];
+  $widget_type = $context['instance']['widget']['type'];
+
+  if ($field['type'] == 'taxonomy_term_reference' && in_array($widget_type, array('options_buttons', 'options_select'))) {
+    if (empty($field['settings']['option_list_callback']) || $field['settings']['option_list_callback'] == 'taxonomy_allowed_values') {
```
argiepiano: The correct key is `'options_list_callback'`, not `'option_list_callback'`.

HOWEVER, I'm thinking that all this `if / else` business is not needed. In the end, what you are doing here is simply adding the indentations to existing options stored   in `$element['#options']`. This means that we are not really worried about how those options are built (all access checks and translations happened earlier, when that options array was built). 

So, I think all is needed here is to simply do `$trees = $field['settings']['allowed_values'];` and remove all the `if/else` stuff below. I haven't tested this yet, but let me know if there is a reason why you need to do all this.

bugfolder: The reason for the if/else was to save the extra db call to find the vocabularies for terms. If `allowed_values` is set and `taxonomy_allowed_values` is the callback, then we can get the vocabularies from `allowed_values`. But if there's a different callback, we can't count on anything in `allowed_values` (at least, that's what I understood you to say in some earlier comments), and so we have to get the vocabularies for all of the terms in order to get the term depths.

If we don't care about that extra db call, we can drop the if/else and just always get the vocabularies via the db call.

bugfolder: Or conversely, if we can count on `allowed_values` even when there's a different callback, we can use it always. (That's what I was doing in an earlier version of this PR.)

argiepiano: I'm so sorry @bugfolder. I think what I said earlier doesn't quite make sense. `allowed_values` will always contain the selected vocabularies, since that's determined in the field settings form. I got confused earlier, thinking that `allowed_values` actually contained the options - thus my confusion.

 > if we can count on allowed_values even when there's a different callback, we can use it always

Yes. No need to do a db call.

🙏🏽

core/modules/taxonomy/taxonomy.module
```diff
@@ -1852,6 +1853,53 @@ function taxonomy_field_widget_form(&$form, &$form_state, $field, $instance, $la
   return $element;
 }
 
+/**
+ * Implements hook_field_widget_form_alter().
+ */
+function taxonomy_field_widget_form_alter(&$element, &$form_state, $context) {
```
argiepiano, line 1859: @bugfolder, this is what I think the function should look like (which is, I think, what you had earlier before I confused things with my erroneous comments):

```php
/**
 * Implements hook_field_widget_form_alter().
 */
function taxonomy_field_widget_form_alter(&$element, &$form_state, $context) {
  $field = $context['field'];
  $widget_type = $context['instance']['widget']['type'];


  if ($field['type'] == 'taxonomy_term_reference' && in_array($widget_type, array('options_buttons', 'options_select'))) {
    $trees = $field['settings']['allowed_values'];
    
    foreach ($trees as $tree) {
      if ($vocabulary = taxonomy_vocabulary_load($tree['vocabulary'])) {
          foreach ($terms as $term) {
            if (isset($element['#options'][$term->tid])) {
              $element[$term->tid] = array('#indentation' => $term->depth);
            }
          }
        }
      }
    }
  }
}
```

core/modules/taxonomy/taxonomy.module
bugfolder: Got it.

core/modules/taxonomy/taxonomy.module
bugfolder: PR is updated.

core/modules/taxonomy/taxonomy.module
bugfolder: Implemented.

core/themes/seven/css/style.css
```diff
@@ -890,7 +890,14 @@ div.teaser-checkbox .form-item,
 .form-item input.form-radio,
 .form-item input.form-checkbox {
   vertical-align: middle;
-  margin-right: 5px;
+}
+[dir="ltr"] .form-item input.form-radio,
+[dir="ltr"] .form-item input.form-checkbox {
```
quicksketch: Although the result is the same, we don't usually specify `[dir="ltr"]` we should remove this prefix for consistency.

```suggestion
.form-item input.form-radio,
.form-item input.form-checkbox {
```

quicksketch:
Merged into 1.x for 1.25.0. Thanks @bugfolder, @argiepiano, and @klonos!

PR #4385, opened 2023-03-17
Issue #5903: Style admin page columns with CSS.

### LINKED ISSUE #5903: [DX] Remove the `position` key from `hook_menu()`.

ghost:
Fixes https://github.com/backdrop/backdrop-issues/issues/5903

### DIFF
--- a/core/modules/system/system.api.php
+++ b/core/modules/system/system.api.php
@@ -772,8 +772,6 @@ function hook_menu_get_item_alter(&$router_item, $path, $original_map) {
  *     the default parent for 'admin/people/create' is 'admin/people').
  *   - tab_root: (optional) For local task menu items, the path of the closest
  *     non-tab item; same default as "tab_parent".
- *   - position: (optional) Position of the block ('left' or 'right') on the
- *     system administration page for this item.
  *   - type: (optional) A bitmask of flags describing properties of the menu
  *     item. Many shortcut bitmasks are provided as constants in menu.inc:
  *     - "MENU_NORMAL_ITEM": (default) Normal menu items show up in the menu
--- a/core/modules/system/system.api.php
+++ b/core/modules/system/system.api.php
@@ -795,6 +795,8 @@ function hook_menu_get_item_alter(&$router_item, $path, $original_map) {
  * For a detailed usage example, see page_example.module.
  * For comprehensive documentation on the menu system, see
  * http://drupal.org/node/102338.
+ *
+ * @since 1.24.2 Support for the "position" key removed.
  */
 function hook_menu() {
   $items['example'] = array(
--- a/core/modules/system/system.theme.inc
+++ b/core/modules/system/system.theme.inc
@@ -519,11 +519,7 @@ function theme_admin_block_content($variables) {
  */
 function theme_system_admin_index($variables) {
   $menu_items = $variables['menu_items'];
-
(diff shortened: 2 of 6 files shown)


core/modules/system/system.api.php
```diff
@@ -772,8 +772,6 @@ function hook_menu_get_item_alter(&$router_item, $path, $original_map) {
  *     the default parent for 'admin/people/create' is 'admin/people').
  *   - tab_root: (optional) For local task menu items, the path of the closest
  *     non-tab item; same default as "tab_parent".
- *   - position: (optional) Position of the block ('left' or 'right') on the
- *     system administration page for this item.
```
klonos, line 776: Can we please add an `@since 1.24.2 Support for the "position" key removed.`

ghost: Done.

core/modules/system/system.theme.inc
```diff
@@ -393,37 +393,24 @@ function theme_status_report($variables) {
...
-      $container[$block['position']] .= $block_output;
+      $container[] = $block_output;
     }
   }
 
   if (!empty($container)) {
     $output = '<div class="admin clearfix">';
-      foreach ($container as $id => $data) {
-        $output .= '<div class="' . $id . ' clearfix">';
+      foreach ($container as $data) {
```
klonos: @BWPanda this coding standards violation needs to be fixed.

klonos: ...never mind. I was looking at the commit instead of the entire PR. Please ignore, and sorry 😬

klonos (approved):
LGTM 👍🏼

core/modules/system/system.api.php
```diff
@@ -797,6 +795,8 @@ function hook_menu_get_item_alter(&$router_item, $path, $original_map) {
  * For a detailed usage example, see page_example.module.
  * For comprehensive documentation on the menu system, see
  * http://drupal.org/node/102338.
+ *
+ * @since 1.24.2 Support for the "position" key removed.
```
klonos, line 799: @BWPanda and @backdrop/core-committers in general, this line should be changed to reflect whichever version this ends up being merged into (1.24.2 or the next minor 1.25.0). Can be done with a commit before merging 😉

PR #4388, opened 2023-03-20
Issue 6036. Remove use of D7 function from Entity Reference

### LINKED ISSUE #6036: Fixed: EntityReference invokes deprecated Drupal function instead of Backdrop alternative.
## Description of the bug

Method `EntityReferenceSelectionHandlerGeneric_user::entityFieldQueryAlter`, which is part of core's Entity Reference module, makes a call to the deprecated Drupal 7 function `format_username()`. This is usually not a problem, unless you have turned off `backdrop_drupal_compatibility` in `settings.php`. Then you get a fatal error, `Call to undefined function format_username()`

## Steps To Reproduce
1. Enable Entity Reference
2. Turn off Drupal compatibility by editing, in`settings.php`:

`$settings['backdrop_drupal_compatibility'] = FALSE;`

3. Add an entity reference field to the `page` content type
4. In "Target type" select `User account`
5. Save the field

## Actual behavior

Fatal error: `Call to undefined function format_username() in EntityReferenceSelectionHandlerGeneric_user->entityFieldQueryAlter() (line 432 of /Users/XXX/Sites/localhost/bdtesting/core/modules/entityreference/plugins/selection/EntityReferenceSelectionHandlerGeneric.inc).`

## Expected behavior

Saving the field should not throw the error.

## Additional information

Obviously a workaround is to turn Drupal compatibility back on.

But a better solution is to avoid using `format_username()`. The Backdrop equivalent is `user_format_name()`

- Backdrop CMS version: 1.24.1
- PHP version: 7.4

====
PR https://github.com/backdrop/backdrop/pull/4388

argiepiano:
Fixes https://github.com/backdrop/backdrop-issues/issues/6036

### DIFF
--- a/core/modules/entityreference/plugins/selection/EntityReferenceSelectionHandlerGeneric.inc
+++ b/core/modules/entityreference/plugins/selection/EntityReferenceSelectionHandlerGeneric.inc
@@ -429,7 +429,7 @@ class EntityReferenceSelectionHandlerGeneric_user extends EntityReferenceSelecti
           $value_part->condition('anonymous_name', $condition['value'], $condition['operator']);
           $value_part->compile(Database::getConnection(), $query);
           $or->condition(db_and()
-            ->where(str_replace('`anonymous_name`', ':anonymous_name', (string) $value_part), $value_part->arguments() + array(':anonymous_name' => format_username(user_load(0))))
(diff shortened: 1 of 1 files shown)


klonos (approved):
LGTM 👍🏼

quicksketch:

PR #4397, opened 2023-04-09
Issue #6040: Use path aliases if available in admin menu items

### LINKED ISSUE #6040: Admin menu doesn't use URL aliases, but other menus do
## Description of the bug

If you create a Page node with a URL alias, then add it to the Administration menu, the destination link uses the normal path, i.e., `http://mysite/node/NNN`, rather than the URL alias. The normal path shows if you hover over the menu item, and if you click on the link, the normal path shows in the browser bar.

This is different from other menus. For example, if you add the same Page node to the Primary navigation menu, the alias path shows up when you hover and when you follow the link, the browser shows the alias path as well.

The admin bar menu should behave the same as other menus; it should use the alias.

## Steps To Reproduce

To reproduce the behavior:

1. Create a page node, say, "User accounts page", with some content.
2. Give it a custom URL alias, say, `admin/people/user-accounts-page`, and save.
3. Add a link to this page to the Administration menu, say, at Administration > User accounts > User accounts page.
4. Also add a link to this page to the Primary navigation menu, say, at [menu root] > User accounts page.

Hover over both links in the menu and observe the URL, and/or visit the page via either link.

## Actual behavior

Hovering over the admin menu link shows this:

image

Hovering over the primary navigation link shows this:

image

## Expected behavior

Both links should show the URL alias for the link.

bugfolder:
Fixes https://github.com/backdrop/backdrop-issues/issues/6040.


quicksketch:
Merged into 1.x and 1.26.x. Thanks @bugfolder!

PR #4399, opened 2023-04-10
Issue 6051. Remove faulty initialization of locale_translation_plurals

### LINKED ISSUE #6051: Plural formulas are wiped out upon D7 upgrade
## Description of the bug

When you upgrade a multi-language site that contains plural formulas, those formulas are not transferred to Backdrop.

## Steps To Reproduce
1. Do a vanilla install of D7
2. Install module `l10n_pconfig`. This module provides defaults for plural formulas, and also provides a UI to edit those
3. Install a predefined "standard" language, for example Spanish. The module `l10n_pconfig` will automatically define a plural formula. You can check this by inspecting the table `languages`. You'll see values in columns `plurals` (e.g. 2), and `formula` (e.g. `($n!=1)`)
4. Proceed to upgrade your site to Backdrop
NOTE: In Backdrop, plural formulas are stored in  the `state` table, under name `locale_translation_plurals`, keyed by langcode. 
5. Inspect the `state` table after the upgrade. You will see the name `locale_translation_plurals`, but its value is blank.

## Actual behavior
The formulas are not transferred.

## Expected behavior
The formulas should be transferred. The `state` table should contain a value for `locale_translation_plurals`

## Additional information

The issue is `locale_update_1004()`. This is what's happening during the upgrade:

1. `update_prepare_language()` is called. The `state` table stores `locale_translation_plurals` with the correct values from the D7 `languages` table 👍🏽 
2. At some point `locale_update_1004()` is run. This hook contains [a line](https://github.com/backdrop/backdrop/blob/1.x/core/modules/locale/locale.install#L346) that wipes out the value previously stored :-1:
`  state_set('locale_translation_plurals', update_variable_get('locale_translation_plurals', array()));`

The issue is that the variable  `locale_translation_plurals` **does not exist in D7**. It never existed, in any version of D7. The plural formulas have **always** been stored in the `languages` table in D7. Thus, since the variable does not exist, `state_set()` simply wipes out the previously stored value.

This [line](https://github.com/backdrop/backdrop/blob/1.x/core/modules/locale/locale.install#L346) should be removed.

BTW, the origin of `locale_translation_plurals` as a state variable is murky. I believe it was present in the first alpha versions of Drupal 8, and was [later removed](https://www.drupal.org/files/1751348-locale-settings-49.patch) in favor of a config setting. Apparently this never made it into Backdrop.

------
PR: https://github.com/backdrop/backdrop/pull/4399

argiepiano:
Fixes https://github.com/backdrop/backdrop-issues/issues/6051

### DIFF
--- a/core/modules/locale/locale.install
+++ b/core/modules/locale/locale.install
@@ -343,8 +343,6 @@ function locale_update_1004() {
 
   $config->save();
 
-  state_set('locale_translation_plurals', update_variable_get('locale_translation_plurals', array()));
-
   // Moved to config:
   update_variable_del('locale_language_negotiation_url_type');
   update_variable_del('locale_language_providers_weight_language');
@@ -358,7 +356,6 @@ function locale_update_1004() {
   update_variable_del('locale_cache_strings');
 
   // Moved to state:
-  update_variable_del('locale_translation_plurals');
   update_variable_del('locale_translation_javascript');
 }
 


quicksketch:
Looks great. Thanks! Merged into 1.x and 1.25.x.

PR #4400, opened 2023-04-13
Issue #6059: Streamline aggregated CSS/JS file RewriteRules in .htaccess

### LINKED ISSUE #6059: [PS] .htaccess rewrite rules for gzipped CSS and JS aggregates cause lots of lstats for files that will never exist
This is a request to backport this change that recently went into D9.5.x/10.0.x/10.1.x: https://www.drupal.org/project/drupal/issues/1040534 

> ...gzipped versions of the files are served by default when available, but there are a couple of issues:
> 
> 1. The regexp matches `.css.gz`, so apache checks if `css_blah.gz.gz` exists.
> 2. The regexp matches any `.css` file anywhere in the filesystem, so unaggregated files (like modernizr) are checked too - i.e. `modernizr.min.js.gz` which is never going to exist.
>
> We can tweak the regular expression to only look for files that are very likely to be css or js aggregates.

[Change record](https://www.drupal.org/node/3308335):

> ## Description: 
> The htaccess rewrite rule that enables static gzipped JavaScript and CSS aggregates to be served to browsers that accept gzip encoding has been modified so that it performs less checks for file existence. This should not cause any functional change except for the performance improvement, however sites should ensure that the change is applied:
> 
> ```diff
> -    RewriteRule ^(.*)\.css $1\.css\.gz [QSA]
> +    RewriteRule ^(.*css_[a-zA-Z0-9-_]+)\.css$ $1\.css\.gz [QSA]
>
> -    RewriteRule ^(.*)\.js $1\.js\.gz [QSA]
> +    RewriteRule ^(.*js_[a-zA-Z0-9-_]+)\.js$ $1\.js\.gz [QSA]
> ```

klonos:
Fixes https://github.com/backdrop/backdrop-issues/issues/6059

### DIFF
--- a/.htaccess
+++ b/.htaccess
@@ -144,12 +144,12 @@ DirectoryIndex index.php index.html index.htm
     # Serve gzip compressed CSS files if they exist and the client accepts gzip.
     RewriteCond %{HTTP:Accept-encoding} gzip
     RewriteCond %{REQUEST_FILENAME}\.gz -s
-    RewriteRule ^(.*)\.css $1\.css\.gz [QSA]
+    RewriteRule ^(.*css_[a-zA-Z0-9-_]+)\.css$ $1\.css\.gz [QSA]
 
     # Serve gzip compressed JS files if they exist and the client accepts gzip.
     RewriteCond %{HTTP:Accept-encoding} gzip
     RewriteCond %{REQUEST_FILENAME}\.gz -s
-    RewriteRule ^(.*)\.js $1\.js\.gz [QSA]
+    RewriteRule ^(.*js_[a-zA-Z0-9-_]+)\.js$ $1\.js\.gz [QSA]
 
     # Serve correct content types, and prevent double compression.
     RewriteRule \.css\.gz$ - [T=text/css,E=no-gzip:1,E=no-brotli:1]


.htaccess
```diff
@@ -144,12 +144,12 @@ DirectoryIndex index.php index.html index.htm
     # Serve gzip compressed CSS files if they exist and the client accepts gzip.
     RewriteCond %{HTTP:Accept-encoding} gzip
     RewriteCond %{REQUEST_FILENAME}\.gz -s
-    RewriteRule ^(.*)\.css $1\.css\.gz [QSA]
+    RewriteRule ^(.*css_[a-zA-Z0-9-_]+)\.css$ $1\.css\.gz [QSA]
```
avpaderno, line 147: `Rewrite` would handle a `randomcharacterscss_aZ09.css` filename too. If that is expected to handle the path too, the regular expression needs to be changed.

.htaccess
```diff
@@ -144,12 +144,12 @@ DirectoryIndex index.php index.html index.htm
...
 
     # Serve gzip compressed JS files if they exist and the client accepts gzip.
     RewriteCond %{HTTP:Accept-encoding} gzip
     RewriteCond %{REQUEST_FILENAME}\.gz -s
-    RewriteRule ^(.*)\.js $1\.js\.gz [QSA]
+    RewriteRule ^(.*js_[a-zA-Z0-9-_]+)\.js$ $1\.js\.gz [QSA]
```
avpaderno, line 152: The same is true here.

.htaccess
klonos: @kiamlaluno are you suggesting something like this then?:

```suggestion
    RewriteRule ^(.*)/(css_[a-zA-Z0-9-_]+)\.css$ $1/$2\.css\.gz [QSA]
```

avpaderno: Since that would handle values like `/files/css/css_xxGQ2.css`, that should work. I think those files will never be accessed with a URL like `http://example.com/css_file_to_use.css`.

PR #4404, opened 2023-04-13
Issue #6050: Avoid static positioning on autocomplete for mobile Safari support.

### LINKED ISSUE #6050: Fixed: Adjust autocomplete dropdown positioning on iPad and iPhone.
## Description of the bug

On iPads and iPhones, the dropdown menu for an autocomplete path doesn't always appear under the field, and if you scroll the image, the dropdown menu moves twice as fast as the scrolling, so that even if it started out appearing in the right place, it moves to the wrong place.

This behavior happens on iPad and iPhone in Safari and Chrome, but it does not manifest on desktop, even when using Safari's Developer tools to emulate an iPad or iPhone screen size.

## Steps To Reproduce

To reproduce the behavior:

Go to any Backdrop site that has an autocomplete field. You need to be able to access the site with an iPad or iPhone. In the example below, I'm visiting b.org.

Find a field that has an autocomplete path. A core example is to edit a menu link, at admin/structure/menu/item/NNN/edit. The Path field is an autocomplete. Type in a character, watch the menu appear, and then try scrolling while the menu is visible.

## Actual behavior

Here's an example (editing a menu item) on iPhone. Note that the autocomplete menu completely covers up the field that it's matching to.

IMG_7437

Here's an example on iPad:

IMG_0049

And here's a video of scrolling behavior on iPad:

https://user-images.githubusercontent.com/29214084/230798376-182fe93e-ec4e-4f0b-a3b7-d05398635dc5.mov

Not shown here, but if the autocomplete field is lower on the page so that when you start typing the keyboard pops up and force-scrolls the window to keep the textfield visible, the autocomplete menu will be pushed so high that it's not even visible upon first appearance.

## Expected behavior

I'd expect the autocomplete menu to appear in the right place under all circumstances, and to work when one is scrolling the image.

## Additional information

Screenshots above are taken from b.org. The site where this first popped up used a Bootstrap Lite-based theme, but I've verified that the behavior happens across themes (e.g., also on Seven).

quicksketch:
Fixes https://github.com/backdrop/backdrop-issues/issues/6050

### DIFF
--- a/core/misc/autocomplete.js
+++ b/core/misc/autocomplete.js
@@ -217,13 +217,13 @@ Backdrop.jsAC.prototype.populatePopup = function () {
     var paddingRight = parseInt($input.css('padding-right').replace('px', ''), 10);
     var paddingWidth = paddingLeft + paddingRight;
 
-    // Because we use "fixed" position, the final location is the offset from
-    // the document and the height of the element, minus scroll bar position.
+    // Position "absolute" instead of "fixed" for mobile Safari compatibility.
+    // See https://github.com/backdrop/backdrop-issues/issues/6050
     $(autocompleteInstance.popup).css({
-      top: ($input.outerHeight() + offset.top - $(document).scrollTop()) + 'px',
-      left: (offset.left - $(document).scrollLeft()) + 'px',
+      top: ($input.outerHeight() + offset.top) + 'px',
+      left: offset.left + 'px',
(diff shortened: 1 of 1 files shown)


klonos (approved):
LGTM 👍🏼

quicksketch:
Merged into 1.x and 1.24.x. Thanks for reviewing @klonos!

PR #4415, opened 2023-04-29
Update install.sh to fix deprecation notices

### LINKED ISSUE #6087: Fix install.sh with PHP 8.1 causes deprecation notices
## Description of the bug

When installing a new Backdrop CMS site using PHP 8.1, the install.sh generates Deprecated Notices:

```
PHP Deprecated:  urldecode(): Passing null to parameter #1 ($string) of type string is deprecated in /var/www/docroot/core/scripts/install.sh on line 134
PHP Deprecated:  urldecode(): Passing null to parameter #1 ($string) of type string is deprecated in /var/www/docroot/core/scripts/install.sh on line 134
PHP Deprecated:  urldecode(): Passing null to parameter #1 ($string) of type string is deprecated in /var/www/docroot/core/scripts/install.sh on line 134
```

## Steps To Reproduce

To reproduce the behavior:
1. Install a new instance of a Backdrop CMS site.  I used the Docksal Boilerplate template that runs this command:
```
bee.php si \--db-name=default --db-user=user \
	  --db-host=db --db-pass=user --url="${VIRTUAL_HOST}" \
		--site-name='My Backdrop Site' \
		--username="admin" --password="admin" --auto
```
3. The script /core/scripts/install.sh calls `array_map` which passes empty values to the `urldecode` function.   

## Actual behavior

Deprecation notice is generated, but the script continues executing, and finishes the site install.

## Expected behavior

No notices should be generated.  

## Additional information

On line 125 the empty value that trigger the error are added to the array.  Moving these lines below will rectify this error, and only add null values.

Wylbur:
Fixes https://github.com/backdrop/backdrop-issues/issues/6087

### DIFF
--- a/core/scripts/install.sh
+++ b/core/scripts/install.sh
@@ -126,13 +126,27 @@ $url = parse_url($options['db-url']);
 $url = (object)array_map('urldecode', $url);
 
 // Assign null values to all required fields that do not already exist
-$url->driver = $url->driver ?? NULL ;
-$url->user = $url->user ?? NULL ;
-$url->pass = $url->pass ?? NULL ;
-$url->host = $url->host ?? NULL ;
-$url->port = $url->port ?? NULL ;
-$url->path = $url->path ?? NULL ;
-$url->database = $url->database ?? NULL ;
+if(!isset($url->driver) && empty($url->driver)) {
+	$url->driver = NULL ;
+}
+if(!isset($url->user) && empty($url->user)) {
+	$url->user = NULL ;
+}
+if(!isset($url->pass) && empty($url->pass)) {
+	$url->pass = NULL ;
+}
+if(!isset($url->host) && empty($url->host)) {
+	$url->host = NULL ;
+}
+if(!isset($url->port) && empty($url->port)) {
+	$url->port = NULL ;
+}
+if(!isset($url->path) && empty($url->path)) {
+	$url->path = NULL ;
+}
+if(!isset($url->database) && empty($url->database)) {
+	$url->database = NULL ;
+}
 
 $db_spec = array(
   'driver' => 'mysql',
--- a/core/scripts/install.sh
+++ b/core/scripts/install.sh
@@ -123,30 +123,22 @@ if (empty($options['db-url'])) {
 }
 $url = parse_url($options['db-url']);
 
-$url = (object)array_map('urldecode', $url);
+$url = array_map('urldecode', $url);
 
-// Assign null values to all required fields that do not already exist
-if(!isset($url->driver) && empty($url->driver)) {
-	$url->driver = NULL ;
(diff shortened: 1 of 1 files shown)


core/scripts/install.sh
```diff
@@ -122,16 +122,18 @@ if (empty($options['db-url'])) {
...
+
+// Assign null values to all required fields that do not already exist
+$url->driver = $url->driver ?? NULL ;
+$url->user = $url->user ?? NULL ;
+$url->pass = $url->pass ?? NULL ;
+$url->host = $url->host ?? NULL ;
+$url->port = $url->port ?? NULL ;
+$url->path = $url->path ?? NULL ;
+$url->database = $url->database ?? NULL ;
+
```
kiamlaluno, line 150: The null coalescing operator is supported started from PHP 7. Backdrop still requires PHP 5.6.

core/scripts/install.sh
```diff
@@ -122,16 +122,32 @@ if (empty($options['db-url'])) {
...
+if(!isset($url->port) && empty($url->port)) {
+	$url->port = NULL ;
+}
+if(!isset($url->path) && empty($url->path)) {
+	$url->path = NULL ;
+}
+if(!isset($url->database) && empty($url->database)) {
+	$url->database = NULL ;
+}
+
```
argiepiano, line 133: Instead of all the above, I suggest a more compact version (starting after `$url = parse_url($options['db-url']);`)

```php
$url = array_map('urldecode', $url);
$url += array(
  'driver' => NULL,
  'user' => NULL,
  'pass' => NULL,
  'host' => NULL,
  'port' => NULL,
  'path' => NULL,
  'database' => NULL,
);

$url = (object)$url;
```

This avoids all the `if` checks, and avoids sending NULLs to `urldecode`

avpaderno: The next lines do not use `$url->database`. Does it make sense to set it to `NULL`?
The same is true for `$url->driver`.
(The array returned by `parse_url()` will not contain any `'driver'` nor `'database'` indexes.)

core/scripts/install.sh
```diff
@@ -122,16 +122,24 @@ if (empty($options['db-url'])) {
   exit;
 }
 $url = parse_url($options['db-url']);
+
+$url = array_map('urldecode', $url);
+
+// Check if the driver is set to mysql and report error if it is not.
+if ( $url['scheme'] != 'mysql' ) {
```
klonos:
```suggestion
if ($url['scheme'] != 'mysql') {
```

klonos (commented):
LGTM 👍🏼

quicksketch:
Merged into 1.x and 1.25.x for 1.25.1. Thanks @Wylbur!

PR #4417, opened 2023-04-30
Issue #6083: Search module: Add mini pager option

### LINKED ISSUE #6083: Add a mini pager option to Search module results.
## Description of the need

All other pagers on my site (mostly generated using Views) have the "mini" pager. It's frustrating to have this one page use only the "full" style pager, and it generates a LOT of buttons which looks pretty silly.

## Proposed solution

Ideally, the search module config would have an option to do a "mini" pager like Views module does.

I have used Views with exposed filters for search in the past (like many years ago before Drupal's search was polished) but I would prefer to use the search module  because it works a lot faster.

## Additional information

Desired look: 
image

Current look:
image

Backdrop now includes the option to use mini pagers on search results pages for a sleek appearance.

---

TRANSLATION UPDATE

Changed strings:
* old: `The maximum number of items indexed in each pass of a <a href="@cron">cron maintenance task</a>. If necessary, reduce this number to prevent timeouts and memory errors while indexing.`
* new: `The maximum number of items indexed in each pass of a <a href="@cron">cron maintenance task</a>. If necessary, reduce this number to prevent timeouts and memory errors while indexing.`
* old2: `Whether to apply a simple Chinese/Japanese/Korean tokenizer based on overlapping sequences. Turn this off if you want to use an external preprocessor for this instead. Does not affect other languages. <strong>Changing this setting will invalidate the search index.</strong>`
* new2: `Whether to apply a simple Chinese/Japanese/Korean tokenizer based on overlapping sequences. Turn this off if you want to use an external preprocessor for this instead. Does not affect other languages. <strong>Changing this setting will invalidate the search index.</strong>'`

New strings:
* `Pager type`
* `Full pager`
* `Mini pager`
* `Includes "Previous"/"Next" and "First"/"Last" links, as well as numbered links for each page of the search results.`
* `Shows only "Previous"/"Next" links, and indications of the current and total number of search results pages.`

klonos:
Fixes https://github.com/backdrop/backdrop-issues/issues/6083


core/modules/search/search.admin.inc
```diff
@@ -137,9 +137,30 @@ function search_admin_settings($form, &$form_state) {
...
+  $form['pager']['search_pager_type'] = array(
+    '#type' => 'radios',
+    '#title' => t('Pager type'),
+    '#title_display' => 'invisible',
+    // The options keys here should be kept the same as the name of theme
+    // function to be called for the pager. See how $variables['pager'] is being
+    // set in template_preprocess_search_results().
+    '#options' => array(
+      'pager' => t('Full pager'),
+      'views_mini_pager' => t('Mini pager'),
```
quicksketch, line 154: Can we change these keys to "full" and "mini"? Ideally we'd rename `theme_views_mini_pager()` to `theme_mini_pager()` and remove it from being views-specific. That can be a follow-up.

quicksketch:
Replaced by https://github.com/backdrop/backdrop/pull/4424.

core/modules/search/search.admin.inc
klonos: Follow-up here: https://github.com/backdrop/backdrop-issues/issues/6096

PR #4423, opened 2023-05-01
Issue 2615

### LINKED ISSUE #2615: [UX] Add the option to limit text formats for each field instance.

quicksketch:
Fixes https://github.com/backdrop/backdrop-issues/issues/2615

Replaces https://github.com/backdrop/backdrop/pull/2894

This builds on @BWPanda's PR. Tests to be added.


core/modules/field/modules/text/text.module
```diff
@@ -83,13 +103,31 @@ function text_field_instance_settings_form($field, $instance) {
...
+  $form['allowed_formats'][filter_fallback_format()]['#value'] = TRUE;
+  $form['allowed_formats'][filter_fallback_format()]['#disabled'] = TRUE;
+
   if ($field['type'] == 'text_with_summary') {
     $form['display_summary'] = array(
       '#type' => 'checkbox',
       '#title' => t('Summary input'),
       '#default_value' => $settings['display_summary'],
       '#description' => t('This allows authors to input an explicit summary, to be displayed instead of the automatically trimmed text when using the "Summary or trimmed" display type.'),
+      '#weight' => 0,
```
klonos, line 137: This places this checkbox between the "Text processing" radios and the newly-introduced "Allowed text formats" checkboxes (which is no good because we are using `#states` and the latter is dependant on a choice of the former). Can we try either giving it a weight of `1`, or giving the other two a weight of `-1`?

quicksketch: Thanks, good suggestion. I wasn't paying very close attention to the form element ordering. Fixed in latest commit.

core/modules/filter/filter.module
```diff
@@ -1249,10 +1252,13 @@ function check_markup($text, $format_id = NULL, $langcode = '', $cache = FALSE)
  *   The form element to process. Properties used:
  *   - #base_type: The form element #type to use for the 'value' element.
  *     'textarea' by default.
- *   - #format: (optional) The text format ID to preselect. If NULL or not set,
- *     the default format for the current user will be used.
+ *   - #format: (optional) The text format name to preselect. If NULL or not
+ *     set, the default format for the current user will be used.
+ *   - #allowed_formats: (optional) An array of format IDs that should be
```
quicksketch:
```suggestion
 *   - #allowed_formats: (optional) An array of format names that should be
```

quicksketch:
Merged into 1.x for 1.25.0. Thanks @BWPanda for making the original PR and @klonos for reviewing here!

PR #2869, opened 2019-09-12
Issue #4029: Add a compare user layout visibility condition.

jenlampton:
Fixes https://github.com/backdrop/backdrop-issues/issues/4029

PR #4426, opened 2023-05-02
4029 layout user compare

### LINKED ISSUE #4029: [UX][D7] Add a 'User account: compare' visibility rule
**Description of the need**
In Drupal 7 / Panels you could choose to show or hide a content pane based on whether a `User context` was equal to the `logged in User context`. This panels access plugin allowed you to create a 'my own dashboard' separately from `the global dashboard`, or to show people their `own` blocks, without showing that block to anybody else. 

Now that we have a dashboard in core, I suspect more and more people are going to notie that this feature is missing from Backdrop.

**Proposed solution**
We should copy the User: compare visibility rule and turn it into a Visibility condition for the Backdrop block/layout system.

**Steps To Test**

To reproduce the behavior:
1. Apply the patch from https://github.com/backdrop/backdrop-issues/issues/4029 (or test in the sandbox)
2. Create a layout at the path `user/%` -- confirm `User` context is automatically added.
3. Scenario 1: Place a block I can only see on my own page
  a) Place a custom block (text: `my own page`)
  b) Add a visibility condition `User: Compare`
  c) Select **First User** Logged in user account
  d) Select **Second user** User account from path (position 2)
  e) Select **Grant access if the two are** The same user account
  f) Save
4. Scenario 2: Place a block I can only see on someone else's page
  a) Place a custom block (text: `someone else's page`)
  b) Add a visibility condition `User: Compare`
  c) Select **First User** Logged in user account
  d) Select **Second user** User account from path (position 2)
  e) Select **Grant access if the two are** Different user accounts
  f) Save
5. Save the layout.
6. Visit your own user profile -- see your own block
7. Visit someone else's user profile -- see the other block

---

**Please do not merge this until https://github.com/backdrop/backdrop-issues/issues/4594 has been merged**

---
PR by @jenlampton: https://github.com/backdrop/backdrop/pull/2869

quicksketch:
Fixes https://github.com/backdrop/backdrop-issues/issues/4029

Replaces/alternative to https://github.com/backdrop/backdrop/pull/2869

This PR builds on @jenlampton's approach with an alternative UI. Fundamentally it's the same. Differences:

- Both user context fields are shown with labels First User Account and Second User Account
- But duplicate options are removed from each field, so in the most common use case, there is only one option in each field.
- The wording on the equality check is changed.


stpaultim:
@quicksketch Can you get a new sandbox working for this PR?

quicksketch:
Let's forego this approach in favor of https://github.com/backdrop/backdrop/pull/2869.

PR #4431, opened 2023-05-05
Update color.module

### LINKED ISSUE #6057: Color Module cannot save colors if uppercase hex codes used in theme's color.inc
Creating a separate issue for this comment raised here: https://github.com/backdrop/backdrop-issues/issues/1905

## Description of the bug

Color Module is intolerant of upper case hex values.

I used uppercase hex values in my css file, only to learn (after three hours of head-scratching and hair-pulling) that the Color Module is intolerant of upper case letters. It reads #45C690 as #000000, and proceeds from there. So you basically get kind of a dark mode thing going for all your color configurations. Once I replaced my biggie letters with smallies, everything worked great.

The issue shows up when you include upper case hex colors in your css file, and the Backdrop Color Module tries to process it (doing color substitutions). It croaks on upper case text. 

This issue also discussed here: https://forum.backdropcms.org/forum/color-module-cant-handle-uppercase-hex

ericfoy:
To repair the failure of Test case 3.

Fixes [#6057](https://github.com/backdrop/backdrop-issues/issues/6057)

See the comments, particularly the last two as of today.

There is a remaining test case, for which a subsequent PR might be raised...

Update: It appears that further patches are likely unnecessary...


core/modules/color/color.module
```diff
@@ -411,7 +411,9 @@ function color_get_palette($theme, $default = FALSE) {
   // Fetch and expand default palette.
   $info = color_get_info($theme);
   $palette = $info['schemes']['default']['colors'];
-
+  foreach ($palette as $k => $v) {
+  $palette[$k] = backdrop_strtolower($v);
```
klonos: Can we please give some meaningful names to these variables? Something like this perhaps?:
```suggestion
  foreach ($palette as $item => $color) {
    $palette[$item] = backdrop_strtolower($color);
```

Also, according to our coding standards, the lines should be indented by 2 spaces. See https://docs.backdropcms.org/php-standards#indenting

klonos: @ericfoy another thought here would be to use `array_map()` instead of the `foreach`. Something like this (untested though):
`$palette = array_map('strtolower', $info['schemes']['default']['colors']);`

avpaderno: The `array_map()` callback can also be `'backdrop_strtolower'`.

ericfoy: 1. Meaningful names: Yes, I agree, and normally I would have done exactly that. In this case, however, I had simply lifted a snippet from line 664 in this same file. This explains my use of the cryptic $k and $v... I actually was following an establish convention ;-) But I'm in favor of the proposed change.
2. The indentation: My bad. Sorry about that. I'll be more careful in this regard in the future.
3. array_map vs. foreach: Personally, I like foreach here because (to me) it is a clearly procedural-looking, straight-forward manipulation of the data at hand, whereas (probably due to my PHP inexperience) the array_map is less obvious to me. This might be simply because this kind of construct (passing a function name as a parameter), though obviously amazingly powerful, is somewhat new to me. That being said, I looked into the speed comparisons of the two functions, and there appears to be no real advantage either way. I would therefore argue in favor of foreach, also just for the sake of symmetry with the other sections of code in this file.

So... how does this work? Should I re-edit, and resubmit? Or do you guys just take it from here? Please advise.

P.S: I like this array_map thing, and I can already tell you: I'm gonna use that sucker, and I already know several places where...

ericfoy: Also...
1. Should I commit @klonos suggestion? 
2. Should I click Resolve below?

I'm still learning the workflow, so I don't know what is appropriate, and what might be an egregious breach of collectively understood protocol...;-)

ericfoy:
Just wondering why "All checks have failed," and what can be done about it... I'm clueless at this point.

avpaderno:
The fork is 22 commits behind the forked repository. It needs to be synchronized.

ericfoy:
Okay.
I'll "start over" (should be super easy) and re-submit.

quicksketch:
Thanks @ericfoy, I've merged this into 1.x and 1.25.x!

ericfoy:
Thanks, everyone! This is a big thrill for me.

PR #4436, opened 2023-05-15
Issue #6095: Views renders empty previous/next links as nbsp.

### LINKED ISSUE #6095: Mini pager: do not render the "previous" link on the first page of results, nor the "next" link in the last page
This is a follow-up to #6083...

On the first page of results the empty "previous" button needs to not be rendered:
image
...same with the "next" button on the last page of results:
image

PS: need to check if this is happening with the mini pager in Views, or if it was only an issue when the mini pager is being used in the search results.

quicksketch:
Fixes https://github.com/backdrop/backdrop-issues/issues/6095

### DIFF
--- a/core/modules/views/templates/views.theme.inc
+++ b/core/modules/views/templates/views.theme.inc
@@ -1157,9 +1157,6 @@ function theme_views_mini_pager($variables) {
         'parameters' => $parameters,
       )
     );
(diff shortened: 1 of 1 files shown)


quicksketch:
Reviewed by @leeksoup and @laryn. I'm merging this into 1.x for 1.25.0.

PR #4421, opened 2023-05-01
Issue 6093: Improve the parent item behavior UI for menu blocks

klonos:
Fixes https://github.com/backdrop/backdrop-issues/issues/6093

Adding `'#indentation'` is currently blocked on https://github.com/backdrop/backdrop-issues/issues/6090

PR #4437, opened 2023-05-15
Issue #4421: Reword menu behavior descriptions.

### LINKED ISSUE #6093: [UX] Menu block settings: Improve the UI for the parent link behavior
This is a follow-up for the feature that got into v1.25.0 with #2370.

This is a side-by-side of the current UI vs. the what I propose we change:
image

List of changes (numbered, so that I can get feedback for each one of the from others):
1. "Merge" the `Parent item's behavior in collapsible (mobile) view.` description into the `Collapsible behavior` label of the radios. Something like `Parent item behavior in collapsible/mobile mode`.
2. Tweak the labels and descriptions of the various options so that they make sense to site builders. As they are currently, the labels are single words that are mirroring Smartmenus settings, which will only make sense to people that are familiar with the "internals" of the library.
3. Add the newly-introduced `#indentation` property to the radio options and the "Accordion-style" checkbox, so that they are nested under the "Display menu toggle button on small screens" checkbox, on which they depend (via `#states` that is). ~~This is currently blocked on #6090~~.

quicksketch:
Fixes https://github.com/backdrop/backdrop-issues/issues/6093

### DIFF
--- a/core/modules/system/system.menu.inc
+++ b/core/modules/system/system.menu.inc
@@ -163,7 +163,7 @@ function system_menu_block_form($config) {
     ),
   );
   $form['collapse']['default']['#description'] = t('Parent expands on first tap, acts as a link on second tap.');
-  $form['collapse']['toggle']['#description'] = t(' Parent expands or collapses on each tap and does not act a link.');
+  $form['collapse']['toggle']['#description'] = t('Parent expands or collapses on each tap and does not act a link.');
   $form['collapse']['link']['#description'] = t('Parent text is a link, +/- element expands or collapses.');
 
   $form['accordion'] = array('#type' => 'checkbox',
--- a/core/modules/system/system.menu.inc
+++ b/core/modules/system/system.menu.inc
@@ -155,17 +155,16 @@ function system_menu_block_form($config) {
       'toggle' => t('Toggle'),
       'link' => t('Link'),
     ),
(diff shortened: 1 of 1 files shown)


core/modules/system/system.menu.inc
```diff
@@ -155,25 +155,23 @@ function system_menu_block_form($config) {
...
       'visible' => array(
-        ':input[name="block_settings[toggle]"]' => array('checked' => TRUE),
         'select[name="block_settings[style]"]' => array('value' => 'dropdown'),
       ),
     ),
   );
   $form['collapse']['default']['#description'] = t('Parent expands on first tap, acts as a link on second tap.');
-  $form['collapse']['toggle']['#description'] = t('Parent functions as toggle without a link.');
-  $form['collapse']['link']['#description'] = t('Parent text is a link, +/- acts as a toggle.');
+  $form['collapse']['toggle']['#description'] = t(' Parent expands or collapses on each tap and does not act a link.');
```
laryn: Extra space snuck in here it looks like.

laryn (commented):
Looks great to me. One minor whitespace issue, and some unrelated code warnings (do we need to fix those in this PR?)

core/modules/system/system.menu.inc
quicksketch:
```suggestion
  $form['collapse']['toggle']['#description'] = t('Parent expands or collapses on each tap and does not act a link.');
```

PR #4442, opened 2023-05-25
Issue #6126: Remove double-encoding of option values.

### LINKED ISSUE #6126: Fix regression: double encoding of option values
## Description of the bug

I noticed this on a views filter for a date field, which has a `>=` in the option value. Improvements that came in with [this commit](https://github.com/backdrop/backdrop/commit/5d202e38f2eca1714055505a4675a61f726eb4c6) accidentally introduced a double-encoding of the option values. In this case, that means `>=` becomes `&gt;=` and no longer validates.

Within `form_select_options` the key value is run through `check_plain` and then used in `backdrop_attributes`, which runs the value through `check_plain` again. 

PR coming.

laryn:
Fixes https://github.com/backdrop/backdrop-issues/issues/6126

### DIFF
--- a/core/includes/form.inc
+++ b/core/includes/form.inc
@@ -2887,7 +2887,7 @@ function form_select_options($element, $choices = NULL) {
     else {
       $key = (string) $key;
       $option_attributes = array();
-      $option_attributes['value'] = check_plain($key);
+      $option_attributes['value'] = $key;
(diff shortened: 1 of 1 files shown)


quicksketch:
I merged this into 1.x and 1.25.x for 1.25.1. Thanks @laryn!

PR #4446, opened 2023-05-28
Issue #5791: Provide a way to quickly grab debug info

### LINKED ISSUE #5791: Provide an easy way to get ready-to copy/paste debugging information (core/php/db/web server versions, enabled themes/modules etc.)

klonos:
Fixes https://github.com/backdrop/backdrop-issues/issues/5791


core/modules/ckeditor/ckeditor.module
```diff
@@ -31,6 +31,7 @@ function ckeditor_editor_info() {
   $editors['ckeditor'] = array(
     'label' => t('CKEditor 4'),
     'library' => array('ckeditor', 'backdrop.ckeditor'),
+    'library_version' => CKEDITOR_VERSION,
```
quicksketch, line 34: We should be able to extract this information from the `library` value, which already provides the CKEDITOR_VERSION in `ckeditor_library_info()`.

core/modules/ckeditor5/ckeditor5.module
```diff
@@ -39,6 +39,7 @@ function ckeditor5_editor_info() {
   $editors['ckeditor5'] = array(
     'label' => t('CKEditor 5'),
     'library' => array('ckeditor5', 'backdrop.ckeditor5'),
+    'library_version' => CKEDITOR5_VERSION,
```
quicksketch, line 42:
```suggestion
```

core/modules/filter/filter.api.php
```diff
@@ -142,6 +142,8 @@ function hook_filter_info_alter(&$info) {
  *     the editor, to be applied when the editor has not been configured yet.
  *   - file: The name of a file containing the editor settings callback.
  *   - library: An associative array containing an optional library.
+ *   - library_version: An optional string containing the current version of the
+ *     editor library.
```
quicksketch, line 146:
```suggestion
```

core/modules/system/system.admin.inc
```diff
@@ -2328,6 +2328,244 @@ function system_status($check = FALSE) {
...
+  $modules = system_rebuild_module_data();
+
+  foreach ($text_formats as $format => $format_info) {
+    if (!empty($format_info->editor)) {
+      $editor = $text_editors[$format_info->editor];
+      $editor_label = $editor['label'];
+      // Add a suffix with the editor version or the editor library version
+      // to the editor label.
+      if (!empty($editor['library_version'])) {
+        $editor_label .= ', ' . $editor['library_version'];
```
quicksketch: I don't think this is exactly right, but something like this would save the need to specify library_version in two places:

```suggestion
      if (!empty($editor['library'])) {
        $library = backdrop_get_library($editor['library']);
        $editor_label .= ', ' . $library['version'];
```

core/modules/system/system.admin.inc
```diff
@@ -2328,6 +2328,244 @@ function system_status($check = FALSE) {
...
+ *   - min_spacing: An integer indicating the minimum spaces added between the
+ *     first column of the output and the second column.
+ *   - rtl: A boolean indicating whether the output should be prepared for RTL
+ *     languages.
+ *
+ * @see system_debug_info()
+ *
+ * @since 1.28.0 function added.
+ */
+function _system_format_array_to_plain_text_columns($lines, $format_settings) {
```
quicksketch: Code hints to make phpcs happy:

```suggestion
function _system_format_array_to_plain_text_columns(array $lines, array $format_settings) {
```

core/modules/filter/filter.api.php
quicksketch: Sorry, I see why this is needed now.

core/modules/ckeditor/ckeditor.module
quicksketch: Sorry, I see why this is needed now. This can't be pulled from `hook_library_info()` because the library specified is the _Backdrop Library for CKEditor_, not the CKEditor 5 library itself (which is a dependency of the Backdrop Library). I can't think of a better way to get this value, so this might be fine after all.

core/modules/system/system.admin.inc
quicksketch: This doesn't work, as I noted in previous comments, so resolving this.

core/modules/system/tests/system.test
```diff
@@ -2512,6 +2512,57 @@ class SystemAdminTestCase extends BackdropWebTestCase {
...
+ * Tests administrative overview pages.
+ */
+class DebugReportTestCase extends BackdropWebTestCase {
+  protected $profile = 'testing';
+
+  /**
+   * @var User
+   */
+  protected $admin_user;
+  function setUp() {
```
docwilmot: Need newline here?

quicksketch:
Merged into 1.x for 1.28.0. Thanks @klonos!

PR #4447, opened 2023-05-30
Issue #6130: Optimize settings.php discovery

### LINKED ISSUE #6130: [PS] Optimize settings.php discovery
I just bumped into this old issue again: https://www.drupal.org/project/drupal/issues/1055856

It was initially filed against D7, and there was a patch for it, with some benchmark results as well:
> ...running benchmarks several times I see the patched version consistently 5-10% faster except for a couple of outliers. xhprof suggests about the same saving (some from file_exists(), some from the surrounding logic).

The issue eventually got moved to the 8.x-dev queue and lost traction since.

Might be worth porting the D7 patch and doing some benchmarks to see if there's actually any performance gain 😉

klonos:
Fixes https://github.com/backdrop/backdrop-issues/issues/6130


core/includes/bootstrap.inc
```diff
@@ -623,9 +623,10 @@ function find_conf_path($http_host, $script_name, $require_settings = TRUE) {
     for ($i = count($uri) - 1; $i > 0; $i--) {
       for ($j = count($server); $j > 0; $j--) {
         $dir = implode('.', array_slice($server, -$j)) . implode('.', array_slice($uri, 0, $i));
-        if (isset($sites[$dir]) && file_exists(BACKDROP_ROOT . '/sites/' . $sites[$dir])) {
-          $dir = $sites[$dir];
-        }
+        if (isset($sites[$dir])) {
+          $conf = './sites/' . $sites[$dir];
+          return $conf;
+       }
```
avpaderno: There is a space less in the last line.

avpaderno (approved):

core/includes/bootstrap.inc
```diff
@@ -623,9 +623,10 @@ function find_conf_path($http_host, $script_name, $require_settings = TRUE) {
...
         if (file_exists(BACKDROP_ROOT . '/sites/' . $dir . '/settings.php') || (!$require_settings && file_exists(BACKDROP_ROOT . '/sites/' . $dir))) {
           $conf = './sites/' . $dir;
           return $conf;
```
avpaderno: Why is not the `file_exists()` call removed from these lines too?

klonos: Not sure. This PR is a direct backport of the latest patch available in https://www.drupal.org/project/drupal/issues/1055856, and this the only bit that is changed there as well.

I am not claiming to understand everything around this - just want to check if there is in fact a performance improvement as suggested over in that issue.

core/includes/bootstrap.inc
klonos: Good catch 👍🏼 ...not sure why PHPCS didn't catch this 🤔 

Fixed with [`267dea2` (#4447)](https://github.com/backdrop/backdrop/pull/4447/commits/267dea273abb460c0705a828c728beaaf1cd3e42)

quicksketch:
Closed as won't fix per https://github.com/backdrop/backdrop-issues/issues/6130#issuecomment-3757881655

PR #4449, opened 2023-06-02
Issue #6132: Allow same vars for tableselect as for table

### LINKED ISSUE #6132: When using `theme_tableselect()`, the `footer`, `caption`, `colgroups`, and `sticky` parameters cannot be set for the table
There is this issue against D7 core, but it is only about not being able to set `sticky`: https://www.drupal.org/node/1788278.

We should fix things so that every possible options allowed in `theme_table()` can be set via `theme_tableselect()`.

klonos:
Fixes https://github.com/backdrop/backdrop-issues/issues/6132


klonos:
Failure seems random. Restarting tests...

klonos:
Random failure with php 8.2 (?)...

klonos:
Different random in php 8.2 (?)...

PR #4450, opened 2023-06-03
Issue #6131: Tableselect: Merge existing child items instead of ignoring them

### LINKED ISSUE #6131: form_process_tableselect() should merge existing child items, not ignore them
I got bitten by this while working on #5085. I was trying to do a couple of things:
- disable one checkbox for a single row in the tableselect (core updates when the core self-update option is disabled)
- make sure that core updates are at the top of the list, followed by any security updates, followed by any unsupported version updates, then lastly listing all regular updates alphabetically.

I was able to individually set a single row to be disabled (checkbox not rendered for it), via `'#disabled'` 👍🏼  ...but the ordering did not work, despite having set `#weight` for the tableselect rows properly. I started doing some research, and came across this series of comments in the [Drupal 7 API page for `theme_tableselect()`](https://api.drupal.org/api/drupal/includes%21form.inc/function/theme_tableselect/7.x), which seemed very similar to the issue I was facing:
> ### How do I disable individual rows in tableselect or set defaults?
> ...
> ### This worked great, but the sort is weird
> ...
> Adding an explicit weight solves the sort issue for me. So in the example above that would be:
> ...
> ```php
> //disable checkboxes
> foreach( $optout as $key => $value ) {
>   $form['table'][$key]['#disabled'] = true;
>   $form['table'][$key]['#weight'] = $rows[$key]['#weight'];
> }
> ```

The first line in that loop was exactly what I was doing in my code, so I followed the advice to set an explicit `#weight`, which fixed the order of the rows in the tableselect 👍🏼 ...but it also removed all the checkboxes from the first column of the table 👎🏼 😞 

So I then started looking at the code of our `form_process_tableselect()` and `theme_tableselect()` functions, and I noticed that there was this odd `// Do not overwrite manually created children.` inline comment in the foreach loop that was setting the `#weight` of the various tableselect rows 🤔 ...so I then tried moving the logic that is adding the `#weight` out of that whole conditional, and everything started working as expected 🎉 

Further search revealed that this issue was already reported by our Drupal brethren here: https://www.drupal.org/project/drupal/issues/2101357 (currently against Drupal core 9.5.x, but also tagged as "Needs backport to D7"):

> In form_process_table() and form_process_tableselect():
> ```php
>     foreach (element_children($element) as $key) {
>       // Do not overwrite manually created children.
>       if (!isset($element[$key]['select'])) {
> ```
> This is the opposite of the behaviour of checboxes and radios elements.

> With those, it's established and desired behaviour that the #process callback for the element merges in any existing child elements. This allows you to do things like add description to single radio buttons or checkboxes, or mark them disabled:
> ```php
>   $form['checkboxes'] = array(
>     '#type' => 'checkboxes',
>     '#options' => $options,
>   );
>   // Make the foobar option disabled for some reason.
>   $form['checkboxes']['foobar']['#disabled'] = TRUE;
>   // The bizbax checkbox needs further description.
>   $form['checkboxes']['foobar']['#description'] = t('More details here');
> ```
> This is not possible with tableselect / table, because they specifically skip existing child elements.

klonos:
Fixes https://github.com/backdrop/backdrop-issues/issues/6131

### DIFF
--- a/core/modules/simpletest/tests/form_test.module
+++ b/core/modules/simpletest/tests/form_test.module
@@ -772,6 +772,13 @@ function _form_test_tableselect_js_select_form($form, $form_state, $action) {
 }
 
 /**
+* AJAX callback that returns the form element.
+*/
+function _form_test_tableselect_ajax_callback($form, &$form_state) {
+  return $form['tableselect'];
+}
+
+/**
  * Tests functionality of vertical tabs.
  */
 function _form_test_vertical_tabs_form($form, &$form_state) {
@@ -799,13 +806,6 @@ function _form_test_vertical_tabs_form($form, &$form_state) {
     '#type' => 'textfield',
   );
   return $form;
-}
-
-/**
-* Ajax callback that returns the form element.
-*/
-function _form_test_tableselect_ajax_callback($form, &$form_state) {
-  return $form['tableselect'];
 }
 
 /**
--- a/core/includes/form.inc
+++ b/core/includes/form.inc
@@ -3838,7 +3838,6 @@ function theme_tableselect($variables) {
  *   The processed element.
  */
 function form_process_tableselect($element) {
-
   if ($element['#multiple']) {
     $value = is_array($element['#value']) ? $element['#value'] : array();
   }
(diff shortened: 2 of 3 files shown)


core/modules/simpletest/tests/form_test.module
```diff
@@ -771,6 +771,13 @@ function _form_test_tableselect_js_select_form($form, $form_state, $action) {
   return _form_test_tableselect_form_builder($form, $form_state, $options);
 }
 
+/**
+* AJAX callback that returns the form element.
+*/
+function _form_test_tableselect_ajax_callback($form, &$form_state) {
+  return $form['tableselect'];
+}
+
```
klonos, line 780: Simply moving this slightly up, to be with the rest of the `_form_test_tableselect_*()` functions.

PR #4458, opened 2023-06-15
Issue #6148: Add wrapper_attributes variable to theme_item_list()

### LINKED ISSUE #6148: [DX] Provide an easy way to specify attributes for the wrapper div in the output of `theme_item_list()`
Consider this code:
```php
  // Prepare the various items for the list.
  $things_to_create_a_list_from = array(
    ...
  );
  $list_items = array();
  foreach ($things_to_create_a_list_from as $thing => $thing_info) {
    $item_class = do something with $thing_info['abc'];
    $list_items[] = array(
      'data' => do something with $thing_info['xyz'],
      // Add CSS classes to each individual <li> item.
      'class' => $item_class,
    );
  }
  $my_list_of_items = theme('item_list', array(
    'items' => $list_items,
    // Add CSS id and classes to the <ul> element.
    'attributes' => array('id' => 'ul-css-id', 'class' => array('some-css-class', 'another-css-class')),
  ));
```

The output of the above implementation of `theme('item_list', ...)` looks like this:
```html
  <div class="item-list">
    <ul id="ul-css-id" class="some-css-class another-css-class">
      <li class="class-for-item-1 odd first">...</li>
      <li class="class-for-item-2 even">...</li>
      <li class="class-for-item-3 odd">...</li>
      <li class="class-for-item-3 even last">...</li>
    </ul>
  </div>
```

I was going through the comments in https://api.drupal.org/api/drupal/includes%21theme.inc/function/theme_item_list/7.x and there are many comments to ask how to add a class, id, or other property to the wrapper div, because CSS doesn't have a way to target parents - see: https://stackoverflow.com/questions/1014861/is-there-a-css-parent-selector

> As of October 2023, there is no way to select the parent of an element in CSS in a way that works across all major browsers.
> 
> The Selectors Level 4 Working Draft specifies [a :has() pseudo-class](https://www.w3.org/TR/selectors-4/#relational) that provides this capability, among others:
> 
> ```css
> li:has(> a.active) { /* styles to apply to the li tag */ }
> ```
> According to https://caniuse.com/css-has, most major browsers added support for this pseudo-class in 2022; however, as of October 2023, it is not yet enabled by default in Firefox.
> 
> So if you want a solution that supports Firefox, minor browsers, and/or old browser versions, then for now, you'll have to resort to JavaScript.

So, each time that someone asks how to add an attribute to the wrapper `<div>` of the list, the answer is that you have to write a custom `MYTHEME_item_list()` function. DX 👎🏼 

Instead of having the output of that wrapper `<div>` be hard-coded, I would like to propose that we introduce a new 'wrapper_attributes' property in `theme_item_list()`, to allow for this to happen:
```php
  $my_list_of_items = theme('item_list', array(
    'items' => $list_items,
    // Add CSS id and classes to the <ul> element.
    'attributes' => array('id' => 'ul-css-id', 'class' => array('some-css-class', 'another-css-class')),
    // Add CSS id and classes to the wrapper <div> element.
    'wrapper_attributes' => array('id' =>'div-css-id', 'class' => array('some-css-classes', 'for-the-div')),
  ));
```

...so that people can achieve the following output without having to write custom theme functions:
```html
  <div id="div-css-id" class="some-css-classes for-the-div item-list">
    <ul id="ul-css-id" class="some-css-class another-css-class">
      <li class="class-for-item-1 odd first">...</li>
      <li class="class-for-item-2 even">...</li>
      <li class="class-for-item-3 odd">...</li>
      <li class="class-for-item-3 even last">...</li>
    </ul>
  </div>
```

klonos:
Fixes https://github.com/backdrop/backdrop-issues/issues/6148


core/includes/theme.inc
```diff
@@ -2260,8 +2260,8 @@ function theme_table(array $variables) {
     $sections['tfoot'] = $variables['footer'];
   }
 
-  // tbody and tfoot have the same structure, and are built using the same
-  // procedure.
+  // The TBODY and TFOOT HTML tags have the same structure, and are built
```
bugfolder:
```suggestion
  // The tbody and tfoot HTML tags have the same structure, and are built
```

core/includes/theme.inc
```diff
@@ -2395,16 +2395,21 @@ function theme_mark($variables) {
...
  *   - type: The type of list to return (e.g. "ul", "ol").
  *   - attributes: The attributes applied to the list element.
+ *   - wrapper_attributes: Any additional attributes that should be applied to
+ *       the wrapper DIV around the list element (the "item-list" CSS class is
+ *       added by default).
  *   - empty: A message to display when there are no items. Allowed value is a
  *     string or render array.
  *
  * @since 1.21.0: added the "empty" variable.
+ * @since 1.26.2: added the "wrapper_attributes" variable.
```
```suggestion
 * @since 1.27.0: added the "wrapper_attributes" variable.
```

quicksketch:
Merged into 1.x for 1.27.0. Thanks @klonos and @bugfolder!

PR #4463, opened 2023-06-16
Issue #6149: Replace deprecated jQuery methods used in core

### LINKED ISSUE #6149: Follow-up: Replace deprecated jQuery methods used in core scripts
## Description of the Task

This is a follow-up to #3106, which added jQuery 3.7.0 to core (with a fallback to 1.12.4).

Methods that have been removed from jQuery already got handled there, but many deprecated functions are still in use in core.
And this issue exists to address that.

Examples for such deprecated functions:

- bind() instead of on()
- click() instead of on('click'...
- $.isArray() instead of Array.isArray()
- $.isFunction instead of typeof
- $.trim() instead of the native string .trim()

These deprecated functions don't cause any problems currently, but getting rid of them is necessary to be ready for jQuery updates, especially the ones that will finally remove those methods.

Luckily, there's a helper tool to find such usages: https://github.com/jquery/jquery-migrate/

None of the changes will (and should) have any impact on functionality, nor should it change anything for contrib.

indigoxela:
Fixes https://github.com/backdrop/backdrop-issues/issues/6149


core/misc/collapse.js
```diff
@@ -69,8 +69,8 @@ Backdrop.behaviors.collapse = {
 
       var summary = $('<span class="summary"></span>');
       $fieldset.
-        bind('summaryUpdated', function () {
-          var text = $.trim($fieldset.backdropGetSummary());
+        on('summaryUpdated', function () {
+          var text = $fieldset.backdropGetSummary().trim();
```
argiepiano: This is probably beyond the scope of this PR, but the custom jQueyr method `backdropGetSummary` defined in `form.js` already trims the return.

core/modules/date/js/date-range.js
```diff
@@ -140,10 +140,10 @@ Backdrop.date.EndDateHandler.prototype.startClickHandler = function () {
 Backdrop.date.EndDateHandler.prototype.endFocusHandler = function (event) {
   for (var id in this.selects) {
     if (this.selects.hasOwnProperty(id)) {
-      this.selects[id].start.unbind('click.endDateHandler');
+      this.selects[id].start.off('click.endDateHandler');
     }
   }
-  $(event.target).unbind('focus', this.endFocusHandler);
+  $(event.target).off('focus');
```
argiepiano: I noticed here that you removed the handler name when you converted this to `off`. I believe the handler name is needed here. That way you unbind only that event handler from the event. As it is right now, this will remove all listeners to the `focus` event.

core/misc/collapse.js
indigoxela: Yep, I know. I didn't dare to skip the second run, but you give me more confidence that it won't cause too much discussion. :relaxed:

core/modules/date/js/date-range.js
indigoxela: Good reminder: this is something where I want to take a second and closer look. :+1:

core/misc/collapse.js
indigoxela: I removed the redundant trim().

core/modules/date/js/date-range.js
indigoxela: Added it back :+1:

core/misc/form.js
```diff
@@ -5,7 +5,7 @@
  */
 $.fn.backdropGetSummary = function () {
   var callback = this.data('summaryCallback');
-  return (this[0] && callback) ? $.trim(callback(this[0])) : '';
+  return (this[0] && callback) ? callback(this[0]).trim() : '';
```
quicksketch: I could see certain situations where this may throw an error. If a summary callback incorrectly returns `false` or does not return anything at all (`null`), then `(false).trim()` or `(null).trim()` will throw a console error. `$.trim(null)` returns an empty string.

There's probably a shorter/cleaner way to write this, but this would be more robust against incorrect return values:
```suggestion
  var returnValue = (this[0] && callback) ? callback(this[0]) : '';
  return typeof returnValue === 'string' ? returnValue.trim() : '';
```

core/modules/admin_bar/js/admin_bar.admin.js
```diff
@@ -18,14 +18,14 @@ Backdrop.behaviors.adminBarPermissionsSetupHelp = {
       var $menu = $(this).find('input[name$="[access administration bar]"]');
 
       // Retrieve the permission label - without description.
-      var adminPermission = $.trim($admin.eq(0).parents('td').prev().children().get(0).firstChild.textContent);
-      var menuPermission = $.trim($menu.eq(0).parents('td').prev().children().get(0).firstChild.textContent);
+      var adminPermission = $admin.eq(0).parents('td').prev().children().get(0).firstChild.textContent.trim();
+      var menuPermission = $menu.eq(0).parents('td').prev().children().get(0).firstChild.textContent.trim();
```
quicksketch, line 22: All other `.trim()` calls (including these) look fine since the variable should always be a string.

quicksketch (commented):
I did a full read-through and everything looks great! I found this one possible problem.

quicksketch:
Merged into 1.x for 1.26.0. Great work @indigoxela and thanks so much!

PR #4465, opened 2023-06-19
Issue #6153 - attributes to username

### LINKED ISSUE #6153: User login form should not use spellcheck or autocorrect
## Description of the bug

Two use cases:
1. Device, particularly mobile autocorrects the username - see https://backdrop.zulipchat.com/#narrow/stream/218635-Backdrop/topic/Mysterious.20login.20failures

2. User has enhanced spell checking enabled and sensitive details are sent for spell checking:
This was raised with security team in December but confirmed (by @jenlampton) that could be in public queue; I forgot about it until today.
   >If site has a password box with a 'show password' toggle (which Backdrop does):
   > 

   > AND if user has enabled Enhanced Spell Check in Chrome, another browser or some spell checking extensions:
   > image
   >
   > Then passwords may be sent in plain text to Google/Microsoft or other services.
   > This could also apply to name, email address, or any fields where personally identifiable information is input (i.e. login, registration, > account creation, webforms, ubercart).
   > 
   > It can be easily prevented by sites adding in the spellcheck=false attribute to sensitive input fields.

https://www.otto-js.com/news/article/spell-jacking-enhanced-spellcheck-features-send-pii-even-passwords

## Actual behavior
On some devices and with some usernames they are autocorrected
With enhanced spellcheck enabled usernames and passwords sent to public service.

## Expected behavior
Usernames and passwords are neither checked for spelling nor autocorrected

## Additional information

See https://www.drupal.org/project/drupal/issues/2537190

---
PR added - https://github.com/backdrop/backdrop/pull/4465

yorkshire-pudding:
Fixes https://github.com/backdrop/backdrop-issues/issues/6153

### DIFF
--- a/core/includes/form.inc
+++ b/core/includes/form.inc
@@ -4755,10 +4755,8 @@ function theme_textarea($variables) {
  */
 function theme_password($variables) {
   $element = $variables['element'];
-  $element['#attributes'] += array(
-    'type' => 'password',
-    'spellcheck' => 'false',
-  );
+  $element['#attributes']['type'] = 'password';
+  $element['#attributes']['spellcheck'] = 'false';
   element_set_attributes($element, array('id', 'name', 'size', 'maxlength', 'placeholder'));
   _form_set_class($element, array('form-text'));
 
--- a/core/includes/form.inc
+++ b/core/includes/form.inc
@@ -3023,22 +3023,26 @@ function form_process_password_confirm($element) {
     '#title' => t('New password'),
     '#value' => empty($element['#value']) ? NULL : $element['#value']['pass1'],
     '#required' => $element['#required'],
-    '#attributes' => array('class' => array('password-field')),
+    '#attributes' => array(
+      'class' => array('password-confirm'),
+      'spellcheck' => 'false',
+    ),
     '#password_strength' => TRUE,
     '#password_shown' => FALSE,
     '#password_toggle' => FALSE,
-    '#spellcheck' => 'false',
   );
   $element['pass2'] =  array(
     '#type' => 'password',
     '#title' => t('Confirm new password'),
     '#value' => empty($element['#value']) ? NULL : $element['#value']['pass2'],
     '#required' => $element['#required'],
-    '#attributes' => array('class' => array('password-confirm')),
+    '#attributes' => array(
+      'class' => array('password-confirm'),
+      'spellcheck' => 'false',
+    ),
     '#password_strength' => FALSE,
     '#password_shown' => FALSE,
     '#password_toggle' => FALSE,
-    '#spellcheck' => 'false',
   );
   $element['#element_validate'] = array('password_confirm_validate');
   $element['#tree'] = TRUE;
@@ -4751,7 +4755,7 @@ function theme_textarea($variables) {
  */
 function theme_password($variables) {
   $element = $variables['element'];
-  $element['#attributes'] = array(
+  $element['#attributes'] += array(
     'type' => 'password',
     'spellcheck' => 'false',
   );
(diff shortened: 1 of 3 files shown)


core/includes/form.inc
```diff
@@ -4749,7 +4751,10 @@ function theme_textarea($variables) {
  */
 function theme_password($variables) {
   $element = $variables['element'];
-  $element['#attributes']['type'] = 'password';
+  $element['#attributes'] = array(
+    'type' => 'password',
+    'spellcheck' => 'false',
+  );
```
argiepiano: The problem with this change is that you are "wiping out" all the other key/values of the `$element['#attributes]` array  (including "disabled"). This is why the test is failing.

Instead do this:
```php
$element['#attributes']['type'] = 'password'l
$element['#attributes']['spellcheck'] = 'false';
```

or

```php
  $element['#attributes'] += array(
    'type' => 'password',
    'spellcheck' => 'false',
  );
```

The same thing applies to all other changes in this PR.

argiepiano: (Notice the `+=`)

core/includes/form.inc
```diff
@@ -3037,6 +3038,7 @@ function form_process_password_confirm($element) {
     '#password_strength' => FALSE,
     '#password_shown' => FALSE,
     '#password_toggle' => FALSE,
+    '#spellcheck' => 'false',
```
avpaderno: Is this line required? `theme_password()` does not use `$element['#spellcheck']`; it sets `$element['#attributes']['spellcheck']` to `'false'`.

core/includes/form.inc
```diff
@@ -3027,6 +3027,7 @@ function form_process_password_confirm($element) {
     '#password_strength' => TRUE,
     '#password_shown' => FALSE,
     '#password_toggle' => FALSE,
+    '#spellcheck' => 'false',
```
avpaderno: Is this line required?

core/includes/form.inc
yorkshire-pudding: Thanks. Done.

yorkshire-pudding:
"Tests" passing, but not quite working on all fields so taking another look.

core/includes/form.inc
```diff
@@ -4750,6 +4756,7 @@ function theme_textarea($variables) {
 function theme_password($variables) {
   $element = $variables['element'];
   $element['#attributes']['type'] = 'password';
+  $element['#attributes']['spellcheck'] = 'false';
```
argiepiano: @yorkshire-pudding I really prefer the second way I suggested, with `=+`. The problem of doing it this way is that you will force false on spellcheck, regardless of what the form definition has. I realize this is really an edge case, but the way element definition work is that the preprocess and theming functions always respect the original definition of the element in the form function. Right now this code doesn't.

For example, if I do (I know this is a crazy edge case):

```php
function testing_stuff_test_form($form, &$form_state) {
  $form['password'] = array(
    '#type' => 'password',
    '#title' => 'Password',
    '#maxlength' => 64,
    '#size' => 15,
    '#disabled' => TRUE,
    '#spellcheck' => 'true',
  );

  return $form;
}
```

I expressly set `#spellcheck` to 'true'. Yet the theming function ignores this and forces `false` on me. 

Again, I know this is an edge case, but for the sake of consistency, we should code theming and preprocess functions to respect the original definition of the element.

argiepiano: @yorkshire-pudding forget it - I'm wrong! spellcheck can't be configured in the form definition. Resolving this comment now...

argiepiano: Actually only you can resolve it - can't do that on my end.

argiepiano: Not sure why this was posted twice. The above is incorrect.

quicksketch:
Merged into 1.x and 1.25.x. Great work here @yorkshire-pudding!

PR #4468, opened 2023-06-30
Change the default value for flood_uid_only

### LINKED ISSUE #6162: On /admin/config/people/login, the default value for the "Identify users attempting to log in, using" field is not shown
On /admin/config/people/login, the *Identify users attempting to log in, using* is rendered as in the following screenshot.

> screenshot

This happens because the value for *flood_uid_only* in [user.flood.json](https://github.com/backdrop/backdrop/blob/1.x/core/modules/user/config/user.flood.json) is `"false"`, not `'0'` nor `'1'` [`user_login_settings()`](https://docs.backdropcms.org/api/backdrop/core%21modules%21user%21user.admin.inc/function/user_login_settings/1) uses for the *flood_uid_only* form element.

```php
  $form['user_limit_settings']['flood_uid_only'] = array(
    '#type' => 'radios',
    '#title' => t('Identify users attempting to log in, using'),
    '#options' => array(
      '0' => t('User ID and IP address combination'),
      '1' => t('User ID only'),
    ),
    '#default_value' => $flood_config->get('flood_uid_only'),
    FALSE => array(
      '#description' => ' Less secure, less likely to lock out users.',
    ),
    TRUE => array(
      '#description' => 'More secure, more likely to lock out users.',
    ),
  );
```

I verified this in a preview site created for a different issue.

I also checked what happens on my site.
With PHP 5.6 the default value for that form element is shown.

> screenshot

If I change the PHP version to PHP 8.0 when the form is shown, and then refresh the page from the browser, the form shows this error on the top.

> Invalid option false in Identify users attempting to log in, using element.

> screenshot

If I change PHP version to PHP 8.1 before going to /admin/config/people/login, when I go to /admin/config/people/login, I do not get the error message, but the form element does not show the default value, like with the preview site.

> 
screenshot

avpaderno:
Changed the default value for flood_uid_only

Fixes https://github.com/backdrop/backdrop-issues/issues/6162

### DIFF
--- a/core/modules/user/user.install
+++ b/core/modules/user/user.install
@@ -2358,13 +2358,13 @@ function user_update_1025() {
 }
 
 /**
- * Correct the value for flood_uid_only.
+ * Correct the default config value for "flood_uid_only" setting.
  */
 function user_update_1026() {
   $config = config('user.flood');
 
-  // Only update if the value is "false".
-  if ($config->get('flood_uid_only') == 'false') {
+  // Only update if the value is exactly the string "false".
+  if ($config->get('flood_uid_only') === 'false') {
     $config->set('flood_uid_only', '0');
     $config->save();
   }
--- a/core/modules/user/user.install
+++ b/core/modules/user/user.install
@@ -2358,5 +2358,18 @@ function user_update_1025() {
 }
 
 /**
+ * Correct the value for flood_uid_only.
+ */
+function user_update_1026() {
+  $config = config('user.flood');
+
+  // Only update if the value is "false".
+  if ($config->get('flood_uid_only') == 'false') {
+    $config->set('flood_uid_only', '0');
+    $config->save();
+  }
+}
+
+/**
  * @} End of "addtogroup updates-7.x-to-1.x"
  */
--- a/core/modules/user/user.install
+++ b/core/modules/user/user.install
@@ -2358,7 +2358,7 @@ function user_update_1025() {
 }
 
 /**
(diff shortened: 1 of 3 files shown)


core/modules/user/user.install
```diff
@@ -2357,6 +2357,19 @@ function user_update_1025() {
   $config->save();
 }
 
+/**
+ * Correct the value for flood_uid_only.
```
quicksketch:
```suggestion
 * Correct the default config value for "flood_uid_only" setting.
```

core/modules/user/user.install
```diff
@@ -2357,6 +2357,19 @@ function user_update_1025() {
...
+ */
+function user_update_1026() {
+  $config = config('user.flood');
+
+  // Only update if the value is "false".
+  if ($config->get('flood_uid_only') == 'false') {
```
quicksketch:
```suggestion
  // Only update if the value is exactly the string "false".
  if ($config->get('flood_uid_only') === 'false') {
```

core/modules/user/user.install
```diff
@@ -2357,6 +2357,19 @@ function user_update_1025() {
   $config->save();
 }
 
+/**
+ * Correct the default config value for "flood_uid_only" setting.
```
quicksketch:
```suggestion
 * Correct the default config value for the "flood_uid_only" setting.
```

quicksketch:
I made some minor tweaks for clarity and merged into 1.x and 1.26.x. Thanks @kiamlaluno!

PR #4473, opened 2023-07-14
Issue 6170. Fix user name filter description

### LINKED ISSUE #6170: User name views autocomplete filter always shows the same description
## Description of the bug

Even if the description for the user name exposed filter is left blank, the following is shown:

Screen Shot 2023-07-14 at 7 30 06 AM

This makes it impossible to customize the description for this filter, plus it causes quite a bit of alignment problems in Seven if the other filters don't contain descriptions.

There is a [D7 issue](https://www.drupal.org/project/views/issues/1700194) for this.

## Steps To Reproduce
1. Create a user View
2. Add a user name exposed filter, leave the description blank
3. Add another exposed filter without a description
4. Visit the view 

## Actual behavior

The description for the user name is shown. 

## Expected behavior

If left blank no description should be shown. And if the description is customized, that customization should be shown verbatim. 

## Additional information

- Backdrop CMS version: 1.25.0

----
PR https://github.com/backdrop/backdrop/pull/4473

argiepiano:
Fixes https://github.com/backdrop/backdrop-issues/issues/6170

### DIFF
--- a/core/modules/user/views/views_handler_filter_user_name.inc
+++ b/core/modules/user/views/views_handler_filter_user_name.inc
@@ -31,13 +31,15 @@ class views_handler_filter_user_name extends views_handler_filter_in_operator {
     $form['value'] = array(
       '#type' => 'textfield',
       '#title' => t('Usernames'),
-      '#description' => t('Enter a comma separated list of user names.'),
       '#default_value' => $default_value,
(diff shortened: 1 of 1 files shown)


quicksketch:
Merged into 1.x and 1.26.x. Thanks!

PR #4476, opened 2023-07-17
Issue #6121: Simpletest cURL status warning should be silent on success.

### LINKED ISSUE #6121: cURL line in status report (from Simpletest module) needs more (or less?) info
## Description of the bug

The Simpletest module places a line in the Status Report like this if cURL is installed:

image

This is not terribly informative. If cURL isn't installed, then there is an error message that explains the need, 

`'The testing framework could not be installed because the PHP <a href="@hash_url">hash</a> extension is disabled.'`

but when it is installed, there is no indication of who needs or cares about it.

(This came up in [this Ubercart issue](https://github.com/backdrop-contrib/ubercart/issues/450), which used the same idiom and ended up with up to four identical (and similarly opaque) status lines.)

We could make this line a little bit more useful by adding a `'description'` field that says something like "cURL is required by the Simpletest testing framework," so that people will know where that line of the report is coming from.

Although, in the interest of decluttering the status report, I'd raise the question of whether we even need to list cURL in the status report if it is properly installed. It makes sense to list things like PHP version, MySQL version, etc., but for a simple "either it's there or it's not", maybe "silence on success" would be a good philosophy.

bugfolder:
Fixes https://github.com/backdrop/backdrop-issues/issues/6121.


.cspell/cspell.json
```diff
@@ -76,6 +76,10 @@
       "filename": "/core/modules/simpletest/tests/file.test",
       "ignoreWords": ["NONEXISTINGFILENAME", "ufail"]
     },
+    {
+      "filename": "/core/modules/simpletest/simpletest.install",
+      "ignoreWords": ["libssh"]
+    },
```
klonos: I would be fine with adding this to our custom dictionary, as we may use that word elsewhere in the future.

core/modules/system/system.install
```diff
@@ -203,12 +203,40 @@ function system_requirements($phase) {
     $tasks = new $class();
     $dbinfo = $t('!system version !version', array('!system' => $tasks->name(), '!version' => Database::getConnection()->version()));
     $requirements['database_system'] = array(
-      'title' => $t('MySQL Database'),
+      'title' => $t('MySQL database'),
```
klonos, line 206: 👍🏼

core/modules/system/system.install
```diff
@@ -203,12 +203,40 @@ function system_requirements($phase) {
...
+      if (!empty($curl_version['ssl_version'])) {
+        $version_info[] = $t('SSL version:') . ' ' . $curl_version['ssl_version'];
+      }
+      if (!empty($curl_version['libssh_version'])) {
+        $version_info[] = $t('libssh version:') . ' ' . $curl_version['libssh_version'];
+      }
+      $curl_status = implode(', ', $version_info);
+    }
+    else {
+      $curl_status = $t('cURL version not available');
```
klonos: This makes it sound as if this is a problem. I would simply leave that as `$t('Enabled')` or `$t('Available')`. The version is a nice to have information - not necessary in any way, and not a problem if we cannot determine it.

core/modules/system/system.install
```diff
@@ -676,7 +704,7 @@ function system_requirements($phase) {
       }
       $description .= $t('You can change this by editing the <code>trusted_host_patterns</code> setting in <code>settings.php</code>.') . ' ' . $more_info;
       $requirements['trusted_host_patterns'] = array(
-        'title' => $t('Trusted Host Settings'),
+        'title' => $t('Trusted host settings'),
```
klonos, line 703: 👍🏼

core/modules/system/system.install
```diff
@@ -707,7 +735,7 @@ function _system_check_db_utf8mb4_requirements($phase) {
 
   $connection = Database::getConnection();
   $t = get_t();
-  $requirements['title'] = $t('MySQL Database 4-byte UTF-8 support');
+  $requirements['title'] = $t('MySQL database 4-byte UTF-8 support');
```
klonos, line 734: 👍🏼

.cspell/cspell.json
klonos: ...now you're getting CSpell failures again because the path where these words are used changed to `/core/modules/system/system.install`. So yeah, better to have the word added in `.cspell/backdrop.dic` instead.

core/modules/system/system.install
```diff
@@ -203,12 +203,40 @@ function system_requirements($phase) {
...
     $dbinfo = $t('!system version !version', array('!system' => $tasks->name(), '!version' => Database::getConnection()->version()));
     $requirements['database_system'] = array(
-      'title' => $t('MySQL Database'),
+      'title' => $t('MySQL database'),
       'severity' => REQUIREMENT_INFO,
       'value' => $dbinfo,
     );
   }
 
+  // Report cURL version or absence.
```
klonos: Perhaps compliment this inline comment by saying that we are intentionally not adding any severity on absence of cURL, since core doesn't require this to function.

core/modules/system/system.install
bugfolder: I was unable to find in any documentation the circumstances under which PHP's `curl_version()` would exist, but would return `FALSE` instead of an array. If it returns FALSE, can we be assured that cURL exists?

.cspell/cspell.json
bugfolder: Gotcha. Done.

core/modules/system/system.install
bugfolder: Good idea. Done.

core/modules/system/system.install
```diff
@@ -203,12 +203,43 @@ function system_requirements($phase) {
...
+      if (!empty($curl_version['version'])) {
+        $version_info[] = $t('Version:') . ' ' . $curl_version['version'];
+      }
+      if (!empty($curl_version['ssl_version'])) {
+        $version_info[] = $t('SSL version:') . ' ' . $curl_version['ssl_version'];
+      }
+      if (!empty($curl_version['libssh_version'])) {
+        $version_info[] = $t('libssh version:') . ' ' . $curl_version['libssh_version'];
+      }
+      $curl_status = implode(', ', $version_info);
```
klonos: Can we please add the ssl and libssh versions in `$requirements['curl']['description]` instead, so that they are shown if someone clicks the "more" link? They would also be easier to read as a bulleted list (`theme('item_list', ...)`).

bugfolder: Sounds good, will update shortly (airport/airplane WiFi permitting...)

core/modules/system/system.install
bugfolder: It looks like it returns FALSE if curl is not installed, so that's a better check for its existence. Reordered the logic to reflect that.

core/modules/system/system.install
bugfolder: Done.

core/modules/simpletest/simpletest.install
```diff
@@ -13,33 +13,32 @@ define('SIMPLETEST_MINIMUM_PHP_MEMORY_LIMIT', '64M');
...
-  $open_basedir = ini_get('open_basedir');
-
-  $requirements['curl'] = array(
-    'title' => $t('cURL'),
-    'value' => $has_curl ? $t('Enabled') : $t('Not found'),
-  );
   if (!$has_curl) {
-    $requirements['curl']['severity'] = REQUIREMENT_ERROR;
-    $requirements['curl']['description'] = $t('The testing framework could not be installed because the PHP <a href="@curl_url">cURL</a> library is not available.', array('@curl_url' => 'http://php.net/manual/en/curl.setup.php'));
+    $requirements['simpletest_curl'] = array(
```
klonos, line 26: Can we please add a `@todo` here to cycle back to this and merge it with the `$requirements['curl']` array provided by the System module, or hook-alter this once we have https://github.com/backdrop/backdrop-issues/issues/6241?

bugfolder: Added.

PR #4477, opened 2023-07-19
Issue #5511: Layout deletion confirmation form: add more explanation …

### LINKED ISSUE #5511: Add more explanation and options to layout deletion confirmation form
This is a follow-up to https://github.com/backdrop/backdrop-issues/issues/5077, collecting some suggestions made in that issue about the confirmation page for the "Delete Layout" action (function `layout_delete_form()` and its processing). Suggestions from that issue:

* If the layout creates a page AND it's the last layout with that path, on the confirm form, add "Warning: the page at path mypage will no longer exist" or such. (H/T @docwilmot.) 
* Provide another button in the confirmation form, that disables the layout instead of deleting it. (H/T @klonos.) 
* Add more explanation to the warning message, something like this: `Deleting this layout will permanently delete the page at /%path. This action cannot be undone. If you are unsure, then consider disabling the layout instead of deleting it.` (H/T @klonos.) 
* Consider making the "destructive" bit of the text a warning message. (H/T @klonos.) 
* Check whether the page being deleted was selected on `admin/config/system/site-information` as the home page or either error page and either warn user or prevent its deletion (which?).

---

I've updated the PR and its behavior, which is now:

- Added a disable option to the "Delete layout?" confirmation page.
- If deleting (or disabling) a layout would remove a page that is used on the site info page, a warning is issued on the "Delete layout?" confirmation page, and after confirming deletion or disabling, a warning message is put up directing the user to the site info page to make the appropriate change.
- If the user disables a layout directly (using the dropbutton link, so there's no confirmation page), if that action removes a page that is used on the site info page, a warning message is put up directing the user to the site info page to make the appropriate change.

bugfolder:
Fixes https://github.com/backdrop/backdrop-issues/issues/5511.

### DIFF
--- a/core/modules/layout/layout.admin.inc
+++ b/core/modules/layout/layout.admin.inc
@@ -2558,7 +2558,6 @@ function layout_clone_form_submit($form, &$form_state) {
 function layout_delete_form($form, &$form_state, $layout) {
   $form_state['layout'] = $layout;
 
-  $form['warn_page_removal'] = FALSE;
   $delete_layout = ($layout->storage === LAYOUT_STORAGE_NORMAL);
   if ($delete_layout) {
     $question = t('Delete layout @title?', array('@title' => $layout->title));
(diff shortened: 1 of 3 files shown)


core/modules/layout/layout.admin.inc
```diff
@@ -2564,27 +2594,140 @@ function layout_delete_form($form, &$form_state, $layout) {
...
+*
+* @return Boolean.
+*   TRUE if the layout creates a page that is the last enabled layout with the
+*   same path. FALSE if not.
+*/
+function layout_removal_removes_page($layout) {
+  $path = $layout->getPath();
+  $form['layout_path'] = $path;
+  $router_item = menu_get_item($path);
+  if ($router_item != FALSE) {
```
argiepiano, line 2660: Minor suggestion, to make this more compact:
```php
if ($router_item != FALSE && $router_item['page_callback'] == 'layout_page_callback') {}
```

And remove the subsequent lines and unneeded variable $is_layout_page.

And a question: is there any instance when the menu_router's page_callback entry for a layout is not `layout_page_callback`?

bugfolder: Any time a layout overrides an existing path (like `node/%`), the `page_callback` is that of the existing path (in the case of `node/%`, the `page_callback` will be `'node_page_view'`). So this check determines whether the page was created by the layout or some other module.

If the page was created by the layout, then deleting the layout entirely removes the page. But if was created by some other module (like `node/%`), then deleting the layout won't remove the page.

core/modules/layout/layout.admin.inc
```diff
@@ -2564,27 +2594,140 @@ function layout_delete_form($form, &$form_state, $layout) {
...
+ * Submit handler for layout_delete_form(). Delete, revert, or disable a layout.
  */
 function layout_delete_form_submit(&$form, &$form_state) {
   /* @var Layout $layout */
   $layout = $form_state['layout'];
+  $title = $layout->title;
+  $path = $layout->getPath();
+  $removal_removes_page = layout_removal_removes_page($layout);
+
+  switch($form_state['clicked_button']['#value']) {
```
```suggestion
  switch ($form_state['clicked_button']['#value']) {
```

quicksketch:
Thanks @bugfolder! Lots of nice tidying in this PR. I merged this into 1.x and 1.26.x. Thanks @argiepiano for the final review!

PR #4479, opened 2023-07-22
Issue #6176: Fix duplicate validation error for user account pictures

### LINKED ISSUE #6176: User pictures: Uploading an unsupported image format throws irrelevant error (about non-image extensions)
This is a follow-up to #4937 and #5917.

## Steps To Reproduce
1. Navigate to `/admin/config/people/settings`, and make sure that the "Enable user pictures" setting is enabled (enable it and save the settings if not).
2. Edit a user account, and try to upload an .svg image in the "Personalization" vertical tab.

## Actual behavior

A validation error is thrown 👍🏼 ...however only the first item in the error is relevant:

image

Here's the same error in textual format:

> The specified file android.svg could not be uploaded.
> - Only images with the following extensions are allowed: png, gif, jpg, jpeg, webp.
> - Only files with the following extensions are allowed: png gif jpg jpeg webp txt doc xls pdf ppt pps odt ods odp docx docm pptx pptm xlsx xlsm.

## Expected behavior
The second bullet point in the error should not be thrown, as the field only allows for images - not any type of supported document/file types.

klonos:
Fixes https://github.com/backdrop/backdrop-issues/issues/6176

### DIFF
--- a/core/includes/file.inc
+++ b/core/includes/file.inc
@@ -1577,26 +1577,33 @@ function file_save_upload($form_field_name, $validators = array(), $destination 
     }
   }
   else {
-    // No validator was provided, so add one using the default list.
-    // Build a default non-munged safe list for file_munge_filename().
-    $extensions = implode(' ', image_get_supported_extensions()) . ' txt doc xls pdf ppt pps odt ods odp docx docm pptx pptm xlsx xlsm';
-
-    $validators['file_validate_extensions'] = array();
-    $validators['file_validate_extensions'][0] = $extensions;
+    // Build a default non-munged safe extension list for file_munge_filename(),
+    // and conditionally add a validator, depending on whether a check for
+    // 'file_validate_is_image' has been added or not.
+    if (isset($validators['file_validate_is_image'])) {
+      $extensions = implode(' ', image_get_supported_extensions());
+      // No need to add a validator for extensions, as that happens for images
+      // in file_validate_is_image().
+    }
+    else {
+      $extensions = implode(' ', file_get_default_extensions());
+      // No validator was provided, so add one.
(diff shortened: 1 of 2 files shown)


core/includes/file.inc
```diff
@@ -1577,26 +1577,33 @@ function file_save_upload($form_field_name, $validators = array(), $destination
...
-    // Build a default non-munged safe list for file_munge_filename().
-    $extensions = implode(' ', image_get_supported_extensions()) . ' txt doc xls pdf ppt pps odt ods odp docx docm pptx pptm xlsx xlsm';
-
-    $validators['file_validate_extensions'] = array();
-    $validators['file_validate_extensions'][0] = $extensions;
+    // Build a default non-munged safe extension list for file_munge_filename(),
+    // and conditionally add a validator, depending on whether a check for
+    // 'file_validate_is_image' has been added or not.
+    if (isset($validators['file_validate_is_image'])) {
+      $extensions = implode(' ', image_get_supported_extensions());
```
indigoxela, line 1584: Why are you setting this variable, when it's actually not used?
Edit: oh, wait, it's used further down the code...

Didn't check, but is $validators even set and an array, if there's no validator set, yet?
Edit: actually it's always supposed to be an array (function param).

core/includes/file.inc
```diff
@@ -1705,6 +1712,13 @@ function file_save_upload($form_field_name, $validators = array(), $destination
   }
 }
 
+function file_get_default_extensions() {
```
indigoxela, line 1715: At first sight I liked the idea of having a function for that. (With doc block, though.)

However, isn't this info already available via config: `config_get('system.core', 'file_default_allowed_extensions');`?

PR #4483, opened 2023-07-27
Issue #6137: Uses database defaults for number field min and max values.

### LINKED ISSUE #6137: number.module is enforcing a maximum value even if one is not specified
## Description of the bug

This issue/bug concerns the `number.module` within Backdrop Core. When defining a` Number (integer)` field, if states that if a maximum value is not specified for the field, then there will not be a maximum value set for the field. Below is the help text for the `Maximum `setting associated with the `Number `field.

"The maximum value that should be allowed in this field. Leave blank for no maximum."

Here is the code that is enforcing the maximum when one is not required:

```
function number_field_widget_form(&$form, &$form_state, $field, $instance, $langcode, $items, $delta, $element) {
  $value = isset($items[$delta]['value']) ? $items[$delta]['value'] : '';

  $element += array(
    '#type' => 'number',
    '#default_value' => $value,
    '#max' => 1000000,
  );

  ....

  // Set minimum and maximum.
  if (is_numeric($instance['settings']['min'])) {
    $element['#min'] = $instance['settings']['min'];
  }
  if (is_numeric($instance['settings']['max'])) {
    $element['#max'] = $instance['settings']['max'];
  }
```

## Expected Behavior

Honor not having a maximum value defined on a `Number (integer)` field.

## Additional information

- Backdrop CMS version: 1.25

smaiorana:
Fixes [#6137](https://github.com/backdrop/backdrop-issues/issues/6137)

### DIFF
--- a/core/modules/field/modules/number/number.module
+++ b/core/modules/field/modules/number/number.module
@@ -76,14 +76,14 @@ function number_field_instance_settings_form($field, $instance) {
     '#type' => 'textfield',
     '#title' => t('Minimum'),
     '#default_value' => $settings['min'],
-    '#description' => t('The minimum value that should be allowed in this field. Leave blank for no minimum.'),
+    '#description' => t('The minimum value that should be allowed in this field. Leave blank for the lowest possible value that can be saved to the database.'),
     '#element_validate' => array('form_validate_number'),
   );
   $form['max'] = array(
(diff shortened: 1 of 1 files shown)


quicksketch:
Thanks @smaiorana! Merged into 1.x and 1.25.x.

PR #4484, opened 2023-07-29
Issue #6180: Set dropbutton widths in Backdrop core JS.

### LINKED ISSUE #6180: Views dropbuttons have bad widths and text alignment in Basis theme
## Description of the bug

Views dropbuttons in the Basis theme have bad widths and text alignment. The width of the dropbutton changes when clicked on; it can extend outside of its enclosing table cell; and the text alignment of the individual entries is right-justified (should be left).

## Steps To Reproduce

To reproduce the behavior:

Install the attached View, which creates two otherwise identical pages with paths `dropbutton-test` and `admin/dropbutton-test`. The first will render with Basis theme, the second with Seven theme (this one renders correctly).

## Actual behavior

Path `admin/dropbutton-test`, in Seven: this renders correctly.

seven

Path `dropbutton-test`, in Basis. You can see all of the problems.

basis

## Expected behavior

Basis should render the dropbuttons as nicely as Seven.

## Additional information

A difference between Seven and Basis is that Seven contains a JS behavior, `Backdrop.behaviors.sevenDropButtonWidths`, in file seven/js/script.js, that sets the min-width of the dropbutton and its parent element, which addresses the changing width and extension outside of the table column. There is also additional CSS in Seven that takes care of the text alignment. So the apparent solution would be to copy the JS and CSS (suitably modified) from Seven into Basis.

The existing issue https://github.com/backdrop/backdrop-issues/issues/6010 mentions several D* issues that apply to dropbuttons, and one of them, https://www.drupal.org/project/drupal/issues/3168326, shows similar width-changing behavior.

This is a View that demos the two pages. Drop the `.txt` extension, add it to staging config and synchronize.
[views.view.dropbutton_test.json.txt](https://github.com/backdrop/backdrop-issues/files/12139239/views.view.dropbutton_test.json.txt)

bugfolder:
Fixes https://github.com/backdrop/backdrop-issues/issues/6180.

### DIFF
--- a/core/misc/dropbutton.js
+++ b/core/misc/dropbutton.js
@@ -23,6 +23,46 @@ Backdrop.behaviors.dropButton = {
 };
 
 /**
+ * Set the min-width of dropbuttons (and their parent) to handle all of the
+ * link widths.
+ */
+Backdrop.behaviors.dropButtonWidths = {
+  attach: function(context, settings) {
+    function adjustDropButtonWidths() {
+      var $dropbutton = $(this);
+
+      // Get widest item width.
+      var widestItem = 0, $item;
+      $dropbutton.find('li:hidden').each(function() {
+        // Use a clone element to avoid altering element CSS properties.
+        $item = $(this).clone().insertAfter($(this)).show().css('visibility','hidden');
+        widestItem = Math.max($item.outerWidth(), widestItem);
+        $item.remove();
+      });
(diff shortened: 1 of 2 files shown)


laryn:
Thank you @bugfolder and @kiamlaluno -- merged into `1.x` and `1.25.x`!

PR #4487, opened 2023-08-04
Issue #1821902: Fixed false positives in recursive token check in _token_build_tree()

### LINKED ISSUE #5014: Commits from token 7.x-1.7
- [ ] https://github.com/backdrop/backdrop/pull/4487 | [d04c7b8](https://git.drupalcode.org/project/token/-/commit/d04c7b83e330f65aa5f6f91858d72f8098651c8a) | [Issue #1821902 by Antonin Brunon, emattias: Fixed palse positives in recursive token check in _token_build_tree() for tokens that had a name of '0' (e.g. entity_token module's delta '0' field tokens).](https://drupal.org/node/1821902)
- [ ] https://github.com/backdrop/backdrop/pull/4488 | [2187bd6](https://git.drupalcode.org/project/token/-/commit/2187bd6a59fbd9378671028ec57024789a13bbe9) | Fixed tokens without a description value were treated as wrong.
- [ ] https://github.com/backdrop/backdrop/pull/4489 | [b189c11](https://git.drupalcode.org/project/token/-/commit/b189c11160e8ef4a517f3e749ac8ebb791c29a99) | Soften token status report messages, split into individual items with different severity values.
- [ ] https://github.com/backdrop/backdrop/pull/4490 | [e6c2702](https://git.drupalcode.org/project/token/-/commit/e6c2702a63877a3eea6b14c8b8d73e087f2b8833) | [Issue #2299039 by Elijah Lynn: Remove unused $sanitize variable from field_tokens().](https://drupal.org/node/2299039)
- [ ] https://github.com/backdrop/backdrop/pull/4491 | [4c39c41](https://git.drupalcode.org/project/token/-/commit/4c39c41c481956b61433c63130c34eae69aaa20b) | [Issue #2485221 by Berdir: Fixed token_pre_render_field_token() parameter should not be passed by reference.](https://drupal.org/node/2485221)
- [ ] N/A??? - see [comment below](https://github.com/backdrop/backdrop-issues/issues/5014#issuecomment-1665075799) | [0a620c8](https://git.drupalcode.org/project/token/-/commit/0a620c8df77dd60f0a54b7c6d02b43107db342a4) | [Issue #2474403 by kfritsche, mariamania, Zekvyrin, B-Prod: Fixed token replacement of field descriptions should not replace i18n_field.module's translations.](https://drupal.org/node/2474403)
- [ ] https://github.com/backdrop/backdrop/pull/4493 | [d9f4e54](https://git.drupalcode.org/project/token/-/commit/d9f4e54356922db0b3c0a746095ce1c05ef26b78) | [Issue #2630378 by hussainweb: Token "cannot contain fewer than one token" error misleading](https://drupal.org/node/2630378)
- [ ] https://github.com/backdrop/backdrop/pull/4493 | [cf53b40](https://git.drupalcode.org/project/token/-/commit/cf53b4089dd7d3821aeb93f237eb7c8d656b878b) | [Issue #2702379 by Dave Reid: Review @todos in token_element_validate()](https://drupal.org/node/2702379)
- [x] N/A - already addressed in a different way in #2697 | [6fa49e9](https://git.drupalcode.org/project/token/-/commit/6fa49e98a450f6df34f9e1ab9abea1721b105085) | [Issue #2825841 by hbemtec: Fixed PHP 7.1 compatibility error: Cannot use string offset as an array in _token_token_tree_format_row().](https://drupal.org/node/2825841)

This list generated with (and then d.org issue links manually added):
`git log 7.x-1.6..7.x-1.7 --reverse --no-merges --pretty=format:"- [ ] #xx | [%h](https://git.drupalcode.org/project/token/-/commit/%H) | [%s](https://drupal.org/node/)"`

klonos:
Fixes [5014](https://github.com/backdrop/backdrop-issues/issues/5014)

Issue #1821902: Fixed false positives in recursive token check in `_token_build_tree()` for tokens that had a name of `'0'`.

See:
- https://www.drupal.org/i/1821902
- https://git.drupalcode.org/project/token/-/commit/d04c7b83e330f65aa5f6f91858d72f8098651c8a

### DIFF
--- a/core/includes/token.inc
+++ b/core/includes/token.inc
@@ -609,7 +609,7 @@ function _token_build_tree($token_type, array $options, $parent_restricted = FAL
       // parent.
       $token_parents[] = $token_type;
     }
-    elseif (in_array($token, array_slice($token_parents, 1))) {
+    elseif (in_array($token, array_slice($token_parents, 1), TRUE)) {
       // Prevent duplicate recursive tokens. For example, this will prevent
       // the tree from generating the following tokens or deeper:
       // [comment:parent:parent]


argiepiano:
Hello @klonos. Thanks for tackling these changes. I'm commenting here rather than in the META issue, since that issue refers to several children issues.

The fix here is harmless and simple, but...

I would caution about transferring changes from Token into Backdrop core without first being able to reproduce the D7 problem in the Backdrop environment. Token was not transferred verbatim into core. 

In this case, the D7 PR fixed a problem that doesn't exist in Backdrop, because of the way tokens for fields with unlimited cardinality work in Backdrop. In Backdrop, those types of fields **do not get listed with the deltas at all** in the token tree. So, the problem that this PR fixes in D7, doesn't exist in Backdrop.

In fact, in Backdrop it's impossible to do a token replacement using a delta for a multiple cardinality field. You can see this by:

1. Be sure you start from a clean install of Backdrop (do not install Entity Tokens)
2. Add a text field to Page with multiple cardinality, for example, a Short text field. Call it Multiple text (`field_multiple_text`)
3. Create a node and enter 2 or 3 values in that field. Take note of the nid of the node (if you hover over "Edit" you'll see the nid).
4. Run the following on Devel (assuming the node nid is 7)

```php
$node = node_load(7);
dpm(token_replace('[node:field_multiple_text:0]', array('node' => $node)));
```

Result: the token is not replaced! If instead you run

```php
$node = node_load(7);
dpm(token_replace('[node:field_multiple_text]', array('node' => $node)));
```

Then you'll get an output, based on the view mode`Tokens`. 

So, you see, in Backdrop, we can't use deltas for multiple cardinality fields. So, the problem in the D7 queue doesn't really apply to any useful case in Backdrop.

In Backdrop, if you want to access deltas, then you have to install Entity Tokens. AND, entity tokens displays the help for  delta replacement in a different way:

Screen Shot 2023-08-04 at 8 22 03 AM


All of this to say: let's please reproduce the original problem from D7 before providing a PR that doesn't fix a Backdrop problem. But then, perhaps you DO have a way to reproduce that problem?

klonos:
@argiepiano you are right, I was working under the assumption that the issue exists in Backdrop (given the similarity in the relevant code bits), same as it did in 7.x Token. Thanks for testing this, and you are right in that we should be first checking to see if the issue applies to us first before "blindly" crossporting commits 👍🏼 ...so fair enough.

I was merely going for "dumb" legwork (in an effort to save time for others), in the hope that it would help us get things done sooner. I was hoping that others would come and do the testing/checking, and someone did 😅 ...thank you 🙏🏼 😄 

So, do you think that we should close this PR and mark the commit as N/A? My concern (not only for this PR - in general) is that there might be a remote chance that this might affect contrib that was built around the 7.x Token and then ported over to Backdrop. So I think that if we have as much code parity as possible, then we might minimize the chances for that to happen. Open to go either way.

argiepiano:
Hi @klonos. Thanks for the reply and for working on these cross-ports! 

In this case, the code only affects the building of the token tree (which is only used for display - not for computing tokens). So, basically, this was done to avoid a "false positive" when doing something like:

```php 
$token_parents = array('site', 'name');
in_array(0, array_slice($token_parents, 1)); // Results in TRUE, but it should actually be FALSE
```

This was a problem when a site had both Token and Entity Tokens enabled (BTW, there was a long feud between maintainers). The problem in D7 was that Entity Tokens created "delta access" for field tokens, while Token didn't. But Token was the module in charge of creating the token display tree! So, the lack of "0" delta in the display tree happened only when both modules were enabled - but this only affected the tree, not the functionality. In Backdrop, Entity Tokens takes a different approach to display the "delta" of multiple value fields, so this fix is unnecessary, but harmless.

In short, the fix is harmless, but I agree about keeping parity with Token, so go for it.

PR #4488, opened 2023-08-04
Fixed tokens without a description value were treated as wrong.

klonos:
Fixes https://github.com/backdrop/backdrop-issues/issues/5014

See: https://git.drupalcode.org/project/token/-/commit/2187bd6a59fbd9378671028ec57024789a13bbe9

PR #4489, opened 2023-08-04
Issue #5014: Soften token status report messages, split into individual items with different severity values.

### LINKED ISSUE #5014: Commits from token 7.x-1.7

klonos:
Fixes https://github.com/backdrop/backdrop-issues/issues/5014

See: https://git.drupalcode.org/project/token/-/commit/b189c11160e8ef4a517f3e749ac8ebb791c29a99

NOTE!!!: This includes some of the changes in https://github.com/backdrop/backdrop/pull/4488 (which is an earlier commit than this one here) - pay attention to not revert changes, depending on which PR gets merged first.

Example of a status warning before:
image

After this change it becomes an "info" message:
image

### DIFF
--- a/core/includes/token.inc
+++ b/core/includes/token.inc
@@ -405,19 +405,24 @@ function token_get_token_problems() {
   $token_info = token_info();
   $token_problems = array(
     'not-array' => array(
-      'label' => t('The following tokens or token types are not defined as arrays:'),
+      'label' => t('Tokens or token types not defined as arrays'),
+      'severity' => REQUIREMENT_ERROR,
     ),
     'missing-info' => array(
-      'label' => t('The following tokens or token types are missing required name and/or description information:'),
+      'label' => t('Tokens or token types missing name property'),
+      'severity' => REQUIREMENT_WARNING,
     ),
     'type-no-tokens' => array(
-      'label' => t('The following token types do not have any tokens defined:'),
+      'label' => t('Token types do not have any tokens defined'),
+      'severity' => REQUIREMENT_INFO,
     ),
     'tokens-no-type' => array(
-      'label' => t('The following token types are not defined but have tokens:'),
+      'label' => t('Token types are not defined but have tokens'),
+      'severity' => REQUIREMENT_INFO,
     ),
     'duplicate' => array(
-      'label' => t('The following token or token types are defined by multiple modules:')
+      'label' => t('Token or token types are defined by multiple modules'),
+      'severity' => REQUIREMENT_ERROR,
     ),
   );
 
@@ -428,8 +433,11 @@ function token_get_token_problems() {
       $token_problems['not-array']['problems'][] = "\$info['types']['$type']";
       continue;
     }
-    elseif (!isset($type_info['name']) || !isset($type_info['description'])) {
+    elseif (!isset($type_info['name'])) {
       $token_problems['missing-info']['problems'][] = "\$info['types']['$type']";
+    }
+    elseif (is_array($type_info['name'])) {
+      $token_problems['duplicate']['problems'][] = "\$info['types']['$type']";
     }
     elseif (empty($token_info['tokens'][$real_type])) {
       $token_problems['type-no-tokens']['problems'][] = "\$info['tokens']['$real_type']";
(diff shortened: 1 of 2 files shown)


avpaderno:
I comment here because the related issue is for multiple PRs. I wonder if it makes sense to show, on /admin/reports/status, code errors as errors.  

For an error caused by a missing dependency, I could install the module or the PHP extension that is required; in the case the error is about a PHP extension I could not install (for example because my hosting provider does not include it between the available extensions), I can at least uninstall the module showing the error on the reports page.  
If the error is about the `$settings['trusted_host_patterns']` value that is not defined, I can open the *settings.php* file and change the lines that define `$settings['trusted_host_patterns']`.

For a *Tokens or token types not defined as arrays* error, which does not identify the module causing it, I would not know what to do. I should disable all the modules that define tokens, one by one, and check all the times if the error vanished, which is not that easy if the module that causes the error is required by another module (and I am not allowed to disable it before disabling all the dependent modules).

klonos:
> I wonder if it makes sense to show, on /admin/reports/status, code errors as errors.

Sorry @kiamlaluno, can you please clarify? Are you saying that all these problems should be flagged as errors? Before this change, everything was flagged as a warning. With this change, some will be shown as errors, some as warnings, and some as info. Is that what you are suggesting to be changed?

Since the respective 7.x commit was not linked to any issue in d.org, I couldn't find any discussion around this change 🤷🏼 ...do others think that we should omit it?

avpaderno:
I meant that I would not show, on the reports, errors that say that some hook did not return the right value, especially when the error does not make clear which module caused it. If showing those errors on the reports page is mandatory, I would rather show them as warnings.

avpaderno:
The issue that introduced that code in the Token module is [Improve checking for token definition problems](https://www.drupal.org/project/token/issues/1061714), where the last comment is the following one.

> Committed #᠎﻿5 to Git. Still needs improved reporting on 'which' modules are the violators.
http://drupalcode.org/project/token.git/commit/6cfd59b

That issue's status is still _Need work_.

Drupal 8, although it has tokens in core, does not contain code similar to that one.

PR #4492, opened 2023-08-04
Issue #2630378: Token "cannot contain fewer than one token" error misleading

klonos:
Replaced by https://github.com/backdrop/backdrop/pull/4493


klonos:
Replaced by https://github.com/backdrop/backdrop/pull/4493

PR #4495, opened 2023-08-08
Issue #6196: Fix token replacement for help text in regular field types

### LINKED ISSUE #6196: Token in field descriptions not working unless in fields that handle multiple values
## Description of the bug

Tokens in field descriptions (on node form) aren't replaced in several field types.

## Steps To Reproduce

1. Add fields of types that do not handle multiple values on their own (like text) to content type
2. Add a description (help text) to that field via field UI (on /admin/structure/types/manage/TYPE/fields/field_FIELD)
3. Note that every field type has that token browser available
4. Use tokens from the token browser on that admin form (date or whatever)
5. Save the field instance
6. Now create a new node of that type (/node/add/TYPE)

## Actual behavior

The tokens are there plain text, no replacement, unless it's a field widget that handles multiple values, like the file or image field.

widget-token

## Expected behavior

As the token browser is available, I'd also expect the tokens to get replaced.

## Additional information

- PHP version: 8.1

The token handling came into core with https://github.com/backdrop/backdrop-issues/issues/480, but the only place function token_replace() is used is [here](https://github.com/backdrop/backdrop/blob/1.x/core/modules/field/field.form.inc#L106), which means, only when some conditions are met. NOT in field_multiple_value_form(), which runs on most field types.

indigoxela:
Fixes https://github.com/backdrop/backdrop-issues/issues/6196

### DIFF
--- a/core/modules/field/field.form.inc
+++ b/core/modules/field/field.form.inc
@@ -193,7 +193,7 @@ function field_multiple_value_form($field, $instance, $langcode, $items, &$form,
   }
 
   $title = check_plain($instance['label']);
-  $description = field_filter_xss($instance['description']);
+  $description = field_filter_xss(token_replace($instance['description']));
 
   $id_prefix = implode('-', array_merge($parents, array($field_name)));
   $wrapper_id = backdrop_html_id($id_prefix . '-add-more-wrapper');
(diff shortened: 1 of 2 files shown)


indigoxela:
This time it's UpdateScriptFunctionalTest :shrug:

laryn:
Thank you @indigoxela and @argiepiano -- merged into `1.x` and `1.25.x`.

PR #4499, opened 2023-08-17
Issue 6198. Provide value array structure when date is empty.

### LINKED ISSUE #6198: Hidden date field saves null values when editing an existing node, producing notices
## Description of the bug

Editing an existing node with a date field that's hidden via `'#access' => FALSE` saves a null value to the database. This, in turn, produces notices when viewing the node: `Warning: Trying to access array offset on value of type null in theme_date_display_combination() (line 219 of /Users/XXX/Sites/localhost/kmta/core/modules/date/date.theme.inc).` This happens regardless of the widget [EDIT: the HTML5 widget handles this correctly].

## Steps To Reproduce

In order to reproduce the problem, you have to alter the node_form to hide the date field.

1. Add a field called `field_test_date` to the Page content type. Use any widget [EDIT: OTHER THAN HTML5]
2. Set the Default Valuie in the field UI to "No default value"
3. In a custom module, implement `hook_form_node_form_alter(&$form, &$form_state);`
4. In that hook, set the access to false, as in `$form['field_test_date'][LANGUAGE_NONE][0]['#access'] = FALSE;`
5. Create a page node. Check the database table `field_data_field_test_date`. Nothing is saved for the new node. 👍🏽 
6. Edit that same page node (no need to change anything). Click `Save`

You'll get the notices above. If you inspect the table for the field you'll see a NULL value being stored. 👎🏽 

## Actual behavior

Null value gets saved for the date. Notices are thrown when viewing the node.

## Expected behavior

Nothing should get saved, since the field is hidden and its default value is none. 

## Additional information

- PHP 8.1
- Backdrop 1.25.1

I've tested the same in Drupal 7. As expected, D7 does not save a null value for the date, and therefore the notices are not produced

---------

PR https://github.com/backdrop/backdrop/pull/4499

argiepiano:
Fixes https://github.com/backdrop/backdrop-issues/issues/6198

Related to https://github.com/backdrop/backdrop-issues/issues/4060

### DIFF
--- a/core/modules/date/date.install
+++ b/core/modules/date/date.install
@@ -113,6 +113,9 @@ function date_update_1002() {
       db_delete('field_data_' . $field_name)
         ->isNull($field_name . '_value')
         ->execute();
+      db_delete('field_revision_' . $field_name)
+        ->isNull($field_name . '_value')
+        ->execute();
     }
   }
 }
--- a/core/modules/date/date.install
+++ b/core/modules/date/date.install
@@ -101,3 +101,18 @@ function date_update_1001() {
   // Remove old system table entries for merged in modules.
   db_query("DELETE FROM {system} WHERE name IN ('date_api', 'date_popup', 'date_views') AND type = 'module'");
 }
+
+/**
+ * Remove entries with null values from date field tables.
+ */
+function date_update_1002() {
+  // Get all date fields.
+  $fields = field_info_fields();
+  foreach ($fields as $field_name => $field) {
+    if (in_array($field['type'], array('date', 'datetime', 'datestamp'))) {
+      db_delete('field_data_' . $field_name)
+        ->isNull($field_name . '_value')
+        ->execute();
+    }
+  }
+}
--- a/core/modules/date/date.elements.inc
+++ b/core/modules/date/date.elements.inc
@@ -2234,9 +2234,17 @@ function _date_default_field_widget_form($field, $instance, $langcode, $items, $
   // @see http://drupal.org/node/1478848.
 
   $is_default = FALSE;
-  if (!empty($instance['widget']['is_new']) && !isset($items[$delta]['value'])) {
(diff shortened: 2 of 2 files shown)


argiepiano:
Reopening to check the random test failure.

core/modules/date/date.install
```diff
@@ -101,3 +101,18 @@ function date_update_1001() {
...
+
+/**
+ * Remove entries with null values from date field tables.
+ */
+function date_update_1002() {
+  // Get all date fields.
+  $fields = field_info_fields();
+  foreach ($fields as $field_name => $field) {
+    if (in_array($field['type'], array('date', 'datetime', 'datestamp'))) {
+      db_delete('field_data_' . $field_name)
```
indigoxela, line 113: We also need to consider `field_revision...`, otherwise we get inconsistencies between the two tables.

core/modules/date/date.install
```diff
@@ -101,3 +101,18 @@ function date_update_1001() {
...
+        ->isNull($field_name . '_value')
+        ->execute();
+    }
+  }
```
indigoxela, line 120: We're almost there. One last thing to take care of, is the persistent entity cache (cache_entity_node);

The "NULL" values are still cached. Loading a node will still contain the faulty values, and the nagging's still there.

core/modules/date/date.install
argiepiano: Good catch! Will do.

core/modules/date/date.install
argiepiano: When updates run, the entity caches are cleared. No need to do that again in the update hook.

core/modules/date/date.install
argiepiano: Done

quicksketch (approved):
Looks good to me!

PR #4501, opened 2023-08-19
Issue 6076. Check also for hidden overflow to trigger resetPosition()

### LINKED ISSUE #6076: [UX] Regression: Views dialogs are rendered partially off-screen and submit buttons are unreachable
## Description of the bug

I think a recent update may have caused this, but views modals seem to be taking up more of my browser window than the whole "admin area" of the site, which seems a little weird for a modal (which is usually smaller) - but most importantly, 
**I can't use the modal because it's now pushed below the part of the screen I can see, so I can't reach the submit buttons on it to submit the modal form.**

Screenshot 2023-04-21 at 9 50 35 PM

## Steps To Reproduce

To reproduce the behavior:
1. Update to 1.24.1 (or newer)
2. Attempt to add a filter to the user/account admin listing

## Actual behavior

Unable to "reach" submit buttons.

## Expected behavior

It should be possible to use the modal. 

## Additional information

- Backdrop CMS version: 1.24.1 (or newer)

## Related issues

* https://github.com/backdrop/backdrop-issues/issues/6077

------

PR https://github.com/backdrop/backdrop/pull/4501

argiepiano:
Fixes https://github.com/backdrop/backdrop-issues/issues/6076

### DIFF
--- a/core/misc/dialog.js
+++ b/core/misc/dialog.js
@@ -79,17 +79,20 @@ Backdrop.dialog = function (element, options) {
       // Because of Chromium's behavior, we must wait until the ajax-loaded
       // jquery.ui.dialog.css is applied.
       var computedStyle = getComputedStyle($element[0]);
-      var waitForCss = function() {
+      var waitForCss = function(counter) {
         var cssOverflowProperty = computedStyle.getPropertyValue('overflow');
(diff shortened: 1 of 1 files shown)


quicksketch:
Merged into 1.x and 1.25.x. Great work @argiepiano!

PR #4503, opened 2023-08-26
Issue #6203: Update format settings if editor changed

### LINKED ISSUE #6203: Filter module: Reset text format editor settings if editor is changed
## Description of the bug

Currently we have CKEditor 4, which is possibly the only editor with settings via the filter module admin UI (could be wrong), which also uses a setting named "toolbar".

If switching the editor, for instance on /admin/config/content/formats/filtered_html, the callback fires to load the new editor's settings. However, the `$format` variable of the form (filter_admin_format_form) isn't updated, which leads to all sorts of errors because of the wrong data structure.

## Steps To Reproduce

Yeah, that's the tricky part. Install the new CKEditor5 module from contrib, and try to get along (patches) with all its quirks. Just to figure out, that the filter module sticks with outdated and incompatible settings in the editor settings form.

_Edit:_ take the mini-module "dummyeditor" from below comment for local testing.

The PHP version might also play a role, didn't test below PHP 8.1.

## Actual behavior

All sorts of PHP errors and warnings, depending in which direction one tries to switch (4 to 5, or 5 to 4).

## Expected behavior

Smooth switching. :wink: 

## Additional information

This is a blocker for progress on the new [CKEditor5](https://github.com/backdrop-contrib/ckeditor5) module, which is a candidate to get merged to core with our 1.27 release next January.

indigoxela:
Fixes https://github.com/backdrop/backdrop-issues/issues/6203

### DIFF
--- a/core/modules/filter/filter.admin.inc
+++ b/core/modules/filter/filter.admin.inc
@@ -209,7 +209,7 @@ function filter_admin_format_form($form, &$form_state, $format) {
     $current_editor = $form_state['values']['editor'];
     if ($format->editor != $current_editor && isset($editors[$current_editor])) {
       $format->editor = $current_editor;
-      $format->editor_settings = $editors[$current_editor]['default settings'];
+      $format->editor_settings = array();
     }
   }
   else {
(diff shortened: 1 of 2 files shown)


quicksketch:
Merged into 1.x and 1.26.x. Thanks @indigoxela!

PR #4512, opened 2023-09-05
Issue #5421: Improve the cron page.

### LINKED ISSUE #5421: [META][UX] Improve the admin/config/system/cron page

jenlampton:
Fixes https://github.com/backdrop/backdrop-issues/issues/5421


core/modules/system/system.admin.inc
```diff
@@ -1566,41 +1566,174 @@ function system_site_information_settings_submit($form, &$form_state) {
...
+  $sentence = 'If the last cron run was more than !number !unit ago, run cron on the next page request.';
+  $form['automatic']['interval']['sentence'] = array(
+    '#type' => 'container',
+    '#theme' => 'cron_interval_sentence',
+    '#children' => array(
+      'number' => $number_element,
+      'unit' => $unit_element,
+      'sentence' => $sentence,
+    ),
+  );
```
quicksketch, line 1696: Setting `#children` manually, especially with a non-element like `$sentence` looks likely to cause unexpected problems. Maybe just set the nested elements normally and set `#sentence` as it's own property.

```suggestion
  $form['automatic']['interval']['sentence'] = array(
    '#type' => 'container',
    '#theme' => 'cron_interval_sentence',
    '#sentence' => $sentence,
  );
  $form['automatic']['interval']['sentence']['number'] = $number_element;
  $form['automatic']['interval']['sentence']['unit'] = $unit_element;
```

PR #4514, opened 2023-09-10
Issue #1275: Add documentation to theme_username().

### LINKED ISSUE #1275: [DX] Document Node passed into theme_username as 'account'.
Use case: I'm trying to override the `theme_username` function to display a user's name as "First Last" instead of as "Username". There are fields for first and last name on the user object, so as long as the user object is passed into `theme_username` I should be golden.

Unfortunately, on a node, instead of passing the user object into `theme_username`, a node is passed in as the 'account'. 

``` php
$variables['name'] = theme('username', array(
    'account' => $node,
    'link_attributes' => array('rel' => 'author'),
  ));
```

(See https://api.backdropcms.org/api/backdrop/core!modules!node!node.theme.inc/function/template_preprocess_node/1)

I found the same thing happening in Drupal 7:

``` php
$variables ['name'] = theme('username', array('account' => $node));
```

(see https://api.drupal.org/api/drupal/modules!node!node.module/function/template_preprocess_node/7) 

This was obviously a deliberate choice, but I wonder why. What does `theme_username` need to know about the node here? Wouldn't it be more valuable to pass in the user account as 'account'? 

Is this worth changing?

bugfolder:
Fixes https://github.com/backdrop/backdrop-issues/issues/1275.

### DIFF
--- a/core/modules/user/user.theme.inc
+++ b/core/modules/user/user.theme.inc
@@ -66,7 +66,11 @@ function theme_user_signature($variables) {
  *   - attributes: An array of attributes to pass to the
  *     backdrop_attributes() function if not linking to the user's page.
  *
+ * $account is typically a User object, but it can be any object that provides
+ * the uid and name properties, for example, a Node.
+ *
  * @see template_preprocess_username()
+ * @see template_preprocess_node()
  */
 function theme_username($variables) {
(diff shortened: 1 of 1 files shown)


quicksketch:
Thanks @bugfolder! Merged into 1.x and 1.25.x.

PR #4517, opened 2023-09-15
Issue #6221: Restore the bottom margin on filter feildsets, and bring…

### LINKED ISSUE #6221: [UX] Bartik Fieldset bottom margin accidentally overwritten
## Description of the bug

If you are using the node form in the Bartik theme, and have a field that comes after the body, the label for that field gets hidden by the text format wrapper (see screenshots below).

This happens because the CSS from filter module adds a `margin: 0;` with the selector `.filter-wrapper`. This selector is more specific than the one the bartik module uses `fieldset` to add the bottom margin that's needed here: 	`margin-bottom: 32px;`

**Before the fix:**

Screenshot 2023-09-14 at 4 36 56 PM

**After the fix:**

Screenshot 2023-09-14 at 5 26 11 PM

## Steps To Reproduce

To reproduce the behavior:

1. Install backdrop
2. Switch admin theme to bartik
3. visit node/add/post
4. check positioning of the label for image field (should be below body field)

## Additional information

- Backdrop CMS version:
- Web server and its version:
- PHP version: 

jenlampton:
Fixes https://github.com/backdrop/backdrop-issues/issues/6221

### DIFF
--- a/core/themes/bartik/css/style.css
+++ b/core/themes/bartik/css/style.css
@@ -1340,6 +1340,7 @@ fieldset {
 }
 .filter-wrapper {
   padding: .5em 0;
+  margin: 0 0 32px 0;
 }
 .filter-wrapper legend,
 .filter-wrapper.collapsed legend {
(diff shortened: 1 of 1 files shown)


quicksketch:
Merged into 1.x and 1.26.x. Thanks @jenlampton!

PR #4518, opened 2023-09-15
Issue #5273: Adjust contextual link triggers to prevent overlaps.

### LINKED ISSUE #5273: Nested contextual links triggers sometimes unreachable

bugfolder:

### DIFF
--- a/core/modules/contextual/js/contextual.js
+++ b/core/modules/contextual/js/contextual.js
@@ -32,7 +32,8 @@ Backdrop.behaviors.contextualLinks = {
         $region.removeClass('contextual-links-region-active');
       });
 
-      // Hide the contextual links when user clicks a link or rolls out of the .contextual-links-region.
+      // Hide the contextual links when user clicks a link or rolls out of the
+      // .contextual-links-region.
       $region.on('mouseleave click', Backdrop.contextualLinks.mouseleave);
       $region.on('mouseenter', function() {
         $trigger.addClass('contextual-links-trigger-active');
@@ -53,7 +54,7 @@ Backdrop.behaviors.contextualLinks = {
       var allWrappers = [];
       $('.contextual-links-wrapper', context).each(function() {
         allWrappers.push({
-          'wrapper': $(this),
+          '$wrapper': $(this),
           'regionOffsetBottom': $(this).parent().offset().top + $(this).parent().height(),
           'hShift': 0,
           'vShift': 0
@@ -62,7 +63,7 @@ Backdrop.behaviors.contextualLinks = {
 
       // Reset margins on all wrappers.
       allWrappers.forEach(function(info) {
-        info.wrapper.css('margin', '0');
+        info.$wrapper.css('margin', '0');
       });
 
       // Recalculate margins to avoid collisions.
@@ -78,8 +79,8 @@ Backdrop.behaviors.contextualLinks = {
           var leader = allWrappers[j];
           // Adjust the position of follower if necessary to avoid collision
           // with leader.
-          var leaderOffset = leader.wrapper.offset();
-          var followerOffset = follower.wrapper.offset();
+          var leaderOffset = leader.$wrapper.offset();
+          var followerOffset = follower.$wrapper.offset();
           // Check vertical overlap.
           if (!(followerOffset.top >= leaderOffset.top && followerOffset.top < leaderOffset.top + vSize)) {
             continue;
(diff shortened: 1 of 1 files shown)


bugfolder:
I added the test text to node/1 in the Tugboat preview.

core/modules/contextual/js/contextual.js
```diff
@@ -16,30 +16,107 @@ Backdrop.behaviors.contextualLinks = {
...
+      var $trigger = $('<a class="contextual-links-trigger" href="#" />').text(Backdrop.t('Configure')).click(
         function () {
           $links.stop(true, true).slideToggle(100);
           $wrapper.toggleClass('contextual-links-active');
           return false;
         }
       );
       // Attach hover behavior to trigger and ul.contextual-links.
-      $trigger.add($links).on('hover',
+      $trigger.add($links).hover(
```
quicksketch: The hover problem is being addressed separately in https://github.com/backdrop/backdrop-issues/issues/6223. We'll need to revert these after that PR is merged.

core/modules/contextual/js/contextual.js
```diff
@@ -16,30 +16,107 @@ Backdrop.behaviors.contextualLinks = {
...
+     */
+    function adjustContextualLinks() {
+      // Get all wrappers anywhere on the page and some info about each.
+      var allWrappers = [];
+      $('.contextual-links-wrapper', context).each(function() {
+        allWrappers.push({
+          'wrapper': $(this),
+          'regionOffsetBottom': $(this).parent().offset().top + $(this).parent().height(),
+          'hshift': 0,
+          'vshift': 0
```
quicksketch: To avoid these spelling notices, maybe change these variables to `xShift` and `yShift` or `hShift` and `vShift`

core/modules/contextual/js/contextual.js
```diff
@@ -16,30 +16,107 @@ Backdrop.behaviors.contextualLinks = {
...
+                follower.wrapper.css('margin-top', follower.vshift);
+                follower.wrapper.css('margin-left', follower.hshift);
+              }
+            }
+          }
+        }
+      }
+    }
+    $(document).ready(adjustContextualLinks);
+    $(window).on('resize', adjustContextualLinks);
```
quicksketch: We have a method for preventing the resize handler from firing repeatably on resize:

```suggestion
    Backdrop.optimizedResize.add(adjustContextualLinks, 'adjustContextualLinks');
```

core/modules/contextual/js/contextual.js
bugfolder: Will do.

core/modules/contextual/js/contextual.js
bugfolder: Yes, I saw the changes and will do the rebase against 1.x after it's merged.

bugfolder: New code incorporated.

core/modules/contextual/js/contextual.js
```diff
@@ -23,23 +23,110 @@ Backdrop.behaviors.contextualLinks = {
...
+
+    /**
+     * Adjusts trigger positions in contextual links to avoid overlaps.
+     */
+    function adjustContextualLinks() {
+      // Get all wrappers anywhere on the page and some info about each.
+      var allWrappers = [];
+      $('.contextual-links-wrapper', context).each(function() {
+        allWrappers.push({
+          'wrapper': $(this),
```
quicksketch: It might be good for clarity to prefix this property with a `$`, as is convention for variables that represent jQuery objects.

```suggestion
          '$wrapper': $(this),
```

core/modules/contextual/js/contextual.js
```diff
@@ -23,23 +23,110 @@ Backdrop.behaviors.contextualLinks = {
...
+          'regionOffsetBottom': $(this).parent().offset().top + $(this).parent().height(),
+          'hShift': 0,
+          'vShift': 0
+        });
+      });
+
+      // Reset margins on all wrappers.
+      allWrappers.forEach(function(info) {
+        info.wrapper.css('margin', '0');
```
quicksketch: Then use this everywhere below where `.wrapper` was referenced:

```suggestion
        info.$wrapper.css('margin', '0');
```

core/modules/contextual/js/contextual.js
```diff
@@ -23,23 +23,110 @@ Backdrop.behaviors.contextualLinks = {
...
+                follower.wrapper.css('margin-top', follower.vShift);
+                follower.wrapper.css('margin-left', follower.hShift);
+              }
+            }
+          }
+        }
+      }
+    }
+    $(document).ready(adjustContextualLinks);
+    Backdrop.optimizedResize.add(adjustContextualLinks, 'adjustContextualLinks');
```
quicksketch: In testing, I found that optimizedResize() fires a lot more often than I would have thought. It also fires on scroll, which we don't want for something like this. So instead a debounce with some documentation might be the best course:
 
```suggestion

    // Usually Backdrop.optimizedResize() would be used for a window resize
    // event, but this potentially expensive operation should be limited to
    // firing infrequently, so Backdrop.debounce() is used here instead.
    $(window).on('resize', Backdrop.debounce(adjustContextualLinks, 500));
```

core/modules/contextual/js/contextual.js
bugfolder: Done.

core/modules/contextual/js/contextual.js
bugfolder: Done.

quicksketch:
Merged into 1.x and 1.26.x. Thanks @bugfolder!

PR #4519, opened 2023-09-15
Issue #6223: Fix jQuery .hover() replacement calls.

### LINKED ISSUE #6223: Contextual Links not appearing on hover
## Description of the bug

A regression was introduced in https://github.com/backdrop/backdrop-issues/issues/6149, where contextual links no longer show up on hover. This appears to be because the replacement for the jQuery `.hover()` is not simply `.on('hover')` but combining `.on('mouseenter')` and `.on('mouseleave')`. See the deprecation notice at https://api.jquery.com/hover/

## Steps To Reproduce

To reproduce the behavior:
1. Log in as an admin user that can use contextual links.
2. Hover over the default homepage blocks and views, you should get contextual menus on hover.
3. No menus are shown.

## Additional information

- Backdrop CMS version: 1.26.0-preview
- Browser(s) and their versions: Firefox, Chromium. Latest versions

quicksketch:
Fixes https://github.com/backdrop/backdrop-issues/issues/6223

### DIFF
--- a/core/misc/tabledrag.js
+++ b/core/misc/tabledrag.js
@@ -302,9 +302,10 @@ Backdrop.tableDrag.prototype.makeDraggable = function (item) {
   }
 
   // Add hover action for the handle.
-  handle.on('hover', function () {
+  handle.on('mouseenter', function () {
     self.dragObject == null ? $(this).addClass('tabledrag-handle-hover') : null;
-  }, function () {
+  });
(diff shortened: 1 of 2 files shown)


laryn:
Thanks @quicksketch and @bugfolder -- merged into `1.x` in time for `1.26.0` 🎉

PR #4520, opened 2023-09-16
Issue #206: Introduce backdrop_include_file().

### LINKED ISSUE #206: [DX][BC] Replace module_load_include() with new, more generic function

bugfolder:
Fixes https://github.com/backdrop/backdrop-issues/issues/206.

Based on https://github.com/backdrop/backdrop/pull/3512, but with further changes to try to get tests to pass.

### DIFF
--- a/core/includes/bootstrap.inc
+++ b/core/includes/bootstrap.inc
@@ -4088,7 +4088,7 @@ function backdrop_get_complete_schema($rebuild = FALSE) {
       // complete list of modules.
       $modules = module_list(TRUE);
       foreach ($modules as $module) {
-        backdrop_include_file('module', $module, $module . '.install');
+        backdrop_include_file($module, $module . '.install');
       }
 
       require_once BACKDROP_ROOT . '/core/includes/common.inc';
--- a/core/includes/bootstrap.inc
+++ b/core/includes/bootstrap.inc
@@ -4081,15 +4081,14 @@ function backdrop_get_complete_schema($rebuild = FALSE) {
     else {
       $schema = array();
       // Load the .install files to get hook_schema.
-      // On some databases this function may be called before bootstrap has
-      // been completed, so we force the functions we need to load just in case.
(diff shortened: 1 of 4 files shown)


core/includes/bootstrap.inc
```diff
@@ -3975,15 +3975,14 @@ function backdrop_get_complete_schema($rebuild = FALSE) {
...
-        module_list(TRUE);
-        module_load_all_includes('install');
+      // On some databases this function may be called before bootstrap has been
+      // completed, so we force the functions we need to load just in case. This
+      // function can be called very early in the bootstrap process, so we force
+      // the module_list() cache to be refreshed to ensure that it contains the
+      // complete list of modules.
+      $modules = module_list(TRUE);
+      foreach ($modules as $module) {
+        backdrop_include_file('module', $module, $module . '.install');
```
avpaderno: The documentation comment for `backdrop_include_file()` says *Do not use this function to include an install file.* but that is what this code does.

bugfolder: Thanks for pointing this out; new PR expands the comment to clarify where and when one can/not use the function to include an install file.

PR #4524, opened 2023-09-28
Issue #6233: Add documentation to hook_update_N().

### LINKED ISSUE #6233: [DX] Explain why you can't count on module functions in update hooks
I just learned the reason why one can't rely on module functions in update hooks. This seems like an easy error for someone to make since doing so would work normally most of the time (updating enabled modules), but could crash the update under particular circumstances (updating a disabled module), and crashing updates are particularly hard to debug and particularly anxiety-producing for those that they happen to.

To help forestall this issue, I suggest adding to the documentation of `hook_update_N()` to explain why using module functions is problematic. PR to follow.

bugfolder:
Fixes https://github.com/backdrop/backdrop-issues/issues/6233.

### DIFF
--- a/core/modules/system/system.api.php
+++ b/core/modules/system/system.api.php
@@ -2878,9 +2878,16 @@ function hook_install() {
  * name, you should never renumber update functions. It may result in updates
  * being either skipped or run twice.
  *
- * Not all module functions are available from within a hook_update_N() function.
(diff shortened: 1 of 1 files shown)


laryn:
Thanks @bugfolder and @argiepiano -- merged!

PR #4525, opened 2023-09-29
Issue #6234: Set aria-current on active links.

### LINKED ISSUE #6234: [A11Y] aria-current on active menu item
## Description of the bug

An accessibility review on a site in development called out the following:

> No aria dictating the "active" element in the navigation for a non-visual user using a screen reader

The solution?

> `aria-current="page"` on the active `<a>`

 - https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Attributes/aria-current

laryn:
Fixes https://github.com/backdrop/backdrop-issues/issues/6234


quicksketch:
Thanks @laryn! Merged into 1.x and 1.26.x.

PR #4532, opened 2023-10-03
Issue #4108: [UX] Improve the UI around reusable blocks

### LINKED ISSUE #4108: [UX] The "Make this block reusable" checkbox has no value once the block is made reusable.
I have started working on the feedback from the code review in #4071, and this UX WTF came up (again) 😜😓 ... #2733 still has merit for all the other things mentioned there, but I am splitting this UX problem here, to it own issue:

> ...you also have that "Make this block reusable" checkbox there that converts the block to something else. When that action is done, that checkbox gets disabled, and the action cannot be reverted. So, what's the point of having that thing be a checkbox in the first place?

I think that:

- Once a block has been made reusable, this checkbox should be hidden/removed.
- The "This option may not be unchecked once enabled." bit of help text is currently only shown when the block has been made reusable. It should be shown *before* it is made reusable; possibly also as a warning, as this action is not reversible once done.
- The "You may configure or delete this reusable block from the [Custom blocks]() page." bit of the help text could become an info message at the top of the block.

Bonus (while at it):

- The "**Translation notice**: The settings in this form are for the source translation only." bit of text should also be a proper message; sounds like a warning. Even in code, we have this comment: `Notify the user that they may be about to edit something they weren't expecting.`.

---
PR by @klonos: ...

klonos:
Fixes https://github.com/backdrop/backdrop-issues/issues/4108


klonos:
Seemingly random failures around `testValidateFormStorageOnCachedPage()`. Kicking it again...

izmeez:
Very nice.

izmeez:
@klonos Thank you this is working. A few points:
1. Looking at the report page (/admin/reports/blocks) is a bit distracting because it seems to default to reverse alphabetical and even on returning to the page it seems to revert to that sort order. I would have expected it to be normal alphabetical sort on first viewing and saved sort order if changed. Nonetheless that is clearly a separate issue.
2. Creating a custom block under structure (/admin/structure/block/add) works smoothly. Only question was should the description be non caps? But, leave caps in warnings with blue background. Here the field for Admin label has the red asterix to indicate it is required.
3. If a custom layout is created through a page layout with the add block button in a region and then selecting "Custom block": the more help text looks good, and if Make reusable is cheked, the extra fields are shown. Howver, the field "Admin label does not have the asterix to indicate it is required even though the description text says it is. This field does not appear to be required if the block is not reusable but is required when reusable, so the field doesn't have an asterix and would have to show one if the Make reusable box is checked. Can this be done with ajax? Even if a custom block is not reusable, shouldn't it still have an admin label and optional description to show in the admin page of the layout anyway?

klonos:
@izmeez sorry for the late reply here 😞 ...to address your points:
1. Yes, separate issue. I actually came across that sorting problem myself while working on something entirely different last week (https://github.com/backdrop/backdrop-issues/issues/6513 in case anyone is interested) - it is an oversight from back when that report page was created or re-styled.
2. I actually think that the part of the message that is bolded ("Updating its content in one place will affect all places where this block is used!") should not be shown at all in `admin/structure/block/add` since it only makes sense when updating a block - it should only be shown when editing existing custom blocks. I'll adjust accordingly.
3. You are right. You might wanna have a look at the following existing issues 😉:
   - https://github.com/backdrop/backdrop-issues/issues/2731
   - https://github.com/backdrop/backdrop-issues/issues/2732

PR #4538, opened 2023-10-08
Issue #6257: Handle case of no visible dropbutton elements.

### LINKED ISSUE #6257: Regression: JS libraries crash if dropbutton library is loaded when there are no dropbuttons on the page
## Description of the bug

This bug happens in combination with #6256. As reported there, the dropbutton JS library is loaded even when the Manage taxonomies Dashboard block is not actually rendered. 

When the dropbutton library is loaded, and when there are no dropbuttons on the page, there is a JS failure in the newly introduced method `Backdrop.behaviors.dropButtonWidths` - see #6180.

This was first reported in [Zulip](https://backdrop.zulipchat.com/#narrow/stream/218635-Backdrop/topic/Dashboard.20.2F.20Admin.20bar.20issue.20after.20update.20to.201.2E26.2E1)

## Steps To Reproduce

To reproduce this, we'll force an error that happens because of the combination of this bug, and another reported in https://github.com/backdrop/backdrop-issues/issues/6256. 

To reproduce the behavior:

1. Be sure you have a vanilla, clean install of Backdrop 1.26.1
2. Create an Editor user
3. Remove all taxonomy privileges from the Editor role
4. Log in as Editor, head to the Dashboard, and inspect the browser dev console
5. You should see `jQuery.Deferred exception: Cannot read properties of undefined (reading 'split') TypeError: Cannot read properties of undefined (reading 'split')`

The admin bar stops working, since JS crashed

## Actual behavior
JS crashes, admin bar stops working.

## Expected behavior
Things should still work even if there are no dropbuttons in the page.

## Additional information

- Backdrop CMS version: 1.26.1
- PHP version: 8.1
- Browser(s) and their versions: Chrome latest

I believe the solution is simple: after doing this in that method
```javascript
    var link = $('.dropbutton li a').first();
```

We should check to see if link actually has content. 

By the way, following an unwritten Backdrop convention, the variable should be `$link`, not `link`, since this holds a jQuery object.

bugfolder:
Fixes https://github.com/backdrop/backdrop-issues/issues/6257.

Also took the opportunity to adhere more closely to coding standards.

### DIFF
--- a/core/misc/dropbutton.js
+++ b/core/misc/dropbutton.js
@@ -29,16 +29,6 @@ Backdrop.behaviors.dropButton = {
 Backdrop.behaviors.dropButtonWidths = {
   attach: function(context, settings) {
     function adjustDropButtonWidths() {
-      // Get the font size and family for an element in a dropbutton if one
-      // exists. If none exist, then we don't need to do any processing.
-      var $link = $('.dropbutton li a').first();
-      if ($link.length == 0) {
-        return;
-      }
-      var fontSize = $link.css('font-size');
-      var fontList = $link.css('font-family').split(', ');
-      var actualFont = fontList.find(font => document.fonts.check(fontSize + ` ${font}`));
-
       var $dropbutton = $(this);
 
       // Get widest item width.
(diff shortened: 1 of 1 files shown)


quicksketch:
Thanks @bugfolder! I've merged this into 1.x and 1.26.x.

PR #4539, opened 2023-10-09
Issue #5541: Support SVG

### LINKED ISSUE #5541: Add support for SVG images.
## Description of the need

At present, if you have uploaded SVG files to use on site, these show as though the file can't be found:
image

As a content author
I want to be able to preview SVG images when I go to add an image to a content item through the "Select from Image Library dialog"
So that I can check I'm inserting the right image

As a content author
I want to see light coloured SVG images against a dark background and vice versa
So that I can easily see what the SVG will look like

## Proposed solution

When viewing the Select from Image Library dialog, SVG files are rendered against a light or dark background depending on the dominant colour, or user could have an option to switch background colour.

If you go to manage files, you can click on view and see them displayed as an entity.

Not aware of any.
The [SVG Image Support](https://github.com/backdrop-contrib/svg_image) module adds support for fields but doesn't have any impact here.

## Additional information

Backdrop now has improved support for SVG files. SVG files (Structured Vector Graphics) have the advantage of being able to scale to any size without distortion or needing a larger file size.  SVGs have been added to the list of default allowed extensions for new sites and you can preview your SVG both in the field where it is added and the Image Library where you can select previously uploaded images.

docwilmot:
Fixes https://github.com/backdrop/backdrop-issues/issues/5541


docwilmot:
Try again TubGoat

docwilmot:
Trying for a Tugboat again!

core/modules/image/image.field.inc
```diff
@@ -617,22 +624,29 @@ function _image_field_widget_alt_validate($element, &$form_state) {
...
+      'image_svg_settings' => array(),
+    ),
+    'description' => t('Format the file as an image. The image can be displayed using an image style and can optionally be linked to the image file itself or its parent content.'),
+    'file formatter' => array(
+      'mime types' => array('image/*'),
     ),
   );
+  $formatters['image'] = $image_formatter;
+
+  $image_formatter['label'] = t('Image (SVG Support)');
```
klonos: I understand what's going on here, however the choice of the variable name here made me pause for a minute and go re-read through the code. Can we please start with a new `$svg_formatter` variable instead of adding keys to the previously defined `$image_formatter` (and adjust the surrounding code accordingly)?

Also, the word `Support` in the label should not be capitalized.

core/modules/image/image.field.inc
```diff
@@ -695,6 +709,28 @@ function image_field_formatter_settings_form($field, $instance, $view_mode, $for
   $element['image_load']['auto']['#description'] = t('Each browser will determine whether or not to lazily load the image.');
   $element['image_load']['eager']['#description'] = t('Load this image immediately, regardless of where it is located on the page.');
 
+  $element['image_svg_settings'] = array(
+    '#type' => 'fieldset',
+    '#title' => t('SVG Images dimensions (attributes)'),
```
klonos:
```suggestion
    '#title' => t('SVG images dimensions (attributes)'),
```

core/modules/image/image.field.inc
```diff
@@ -695,6 +709,28 @@ function image_field_formatter_settings_form($field, $instance, $view_mode, $for
...
+    '#tree' => TRUE,
+  );
+
+  $element['image_svg_settings']['width'] = array(
+    '#type' => 'textfield',
```
klonos: Can we please use a number field here instead?

core/modules/image/image.field.inc
```diff
@@ -695,6 +709,28 @@ function image_field_formatter_settings_form($field, $instance, $view_mode, $for
...
+    '#title' => t('Width'),
+    '#size' => 10,
+    '#field_suffix' => 'px',
+    '#default_value' => $settings['image_svg_settings']['width'],
+  );
+
+  $element['image_svg_settings']['height'] = array(
+    '#type' => 'textfield',
```
klonos: Same as previous comment. We should be using a number field.

core/modules/image/image.field.inc
```diff
@@ -737,6 +773,19 @@ function image_field_formatter_settings_summary($field, $instance, $view_mode) {
     $summary[] = t('Loading: @speed', array('@speed' => $settings['image_load']));
   }
 
+  $formatter_settings = $instance['display'][$view_mode]['settings'];
+
+  if (!empty($formatter_settings['image_svg_settings']['width'])
+    || !empty($formatter_settings['image_svg_settings']['width'])) {
+    $summary[] = t(
+      'SVG images attributes: <br /> Width: @widthpx,<br /> Height: @heightpx',
```
klonos: Lets try this to see if cspell is happy:

```suggestion
      'SVG images attributes: <br /> Width: @width_px,<br /> Height: @height_px',
```

...also, these names do not match the array keys below, so the string replacement won't happen.

core/modules/image/image.field.inc
```diff
@@ -737,6 +773,19 @@ function image_field_formatter_settings_summary($field, $instance, $view_mode) {
...
+      array(
+        '@width' => (int) $formatter_settings['image_svg_settings']['width'],
+        '@height' => (int) $formatter_settings['image_svg_settings']['height'],
```
klonos: ...as per my previous comment, these do not match the placeholders used in `t()`.

Also, if we use number fields instead of textfields, we might not need to be explicitly type casting here.

core/modules/image/image.field.inc
```diff
@@ -755,26 +804,42 @@ function image_field_formatter_view($entity_type, $entity, $field, $instance, $l
...
-      );
+      $is_svg = image_is_svg($item['uri']);
+
+      // Add custom attributes only for SVG images.
+      if ($is_svg) {
+        $image_svg_settings = $display['settings']['image_svg_settings'];
+
+        foreach ($image_svg_settings as $attribute => $attribute_value) {
+          if (!empty($attribute)) {
+            $item[$attribute] = (int) $attribute_value;
```
klonos: I haven't actually tested this, but from reading the code it seems that this would add the attributes with only numeric values. Shouldn't we be adding `px` suffix here?

core/modules/image/image.module
```diff
@@ -1349,3 +1359,173 @@ function image_filter_keyword($value, $current_pixels, $new_pixels) {
...
+      $file->metadata['height'] = $svg_dimensions['height'];
+    }
+  }
+}
+
+
+/**
+ * Implements hook_field_attach_presave().
+ *
+ * When saving a entity, this hook is called to make any adjustments to the
```
klonos, line 1408:
```suggestion
 * When saving an entity, this hook is called to make any adjustments to the
```

core/modules/image/image.module
```diff
@@ -1349,3 +1359,173 @@ function image_filter_keyword($value, $current_pixels, $new_pixels) {
...
+ * entire entity's fields before attempting to write to the database. When
+ * Image module encounters an SVG image, it sets the height and width values
```
klonos:
```suggestion
 * When saving an entity, this hook is called to make any adjustments to the
 * entire entity's fields before attempting to write to the database. When the
 * Image module encounters an SVG image, it sets the height and width values
```

core/modules/image/image.module
```diff
@@ -1349,3 +1359,173 @@ function image_filter_keyword($value, $current_pixels, $new_pixels) {
...
+ * to be empty strings, which would cause a database constraint error when
+ * a string is inserted into an integer column.
+ *
+ * This implementation loops through all fields attached to an entity, finds
+ * those that are an image field, then sets the width and height values within
+ * that field. These changes are made by reference, so they are in the $entity
+ * object when Field module saves to the database.
```
klonos:
```suggestion
 * object when the Field module saves to the database.
```

core/modules/image/image.module
```diff
@@ -1349,3 +1359,173 @@ function image_filter_keyword($value, $current_pixels, $new_pixels) {
...
+        $variables['height'] = SVG_IMAGE_DEFAULT_HEIGHT;
+      }
+    }
+  }
+}
+
+/**
+ * Implements hook_file_presave().
+ *
+ * Set the dimensions in the "metadata" property used by File Entity module.
```
klonos:
```suggestion
 * Set the dimensions in the "metadata" property used by the File Entity module.
```

core/modules/image/image.module
```diff
@@ -1349,3 +1359,173 @@ function image_filter_keyword($value, $current_pixels, $new_pixels) {
...
+ *   NULL values. If the image is not an SVG, FALSE will be return.
+ */
+function image_get_svg_dimensions($uri) {
+  // Safety check.
+  if (!image_is_svg($uri)) {
+    return FALSE;
+  }
+
+  // Return cached dimensions if already retrieved once this request.
+  $cached_images = &drupal_static(__FUNCTION__, array());
```
klonos:
```suggestion
  $cached_images = &backdrop_static(__FUNCTION__, array());
```

In https://github.com/backdrop/backdrop-issues/issues/6262#issuecomment-1760503696 I proposed adding `drupal` as a forbidden word in cspell, to catch these as "spelling errors" in our tests 😉

core/modules/image/image.field.inc
```diff
@@ -695,6 +709,28 @@ function image_field_formatter_settings_form($field, $instance, $view_mode, $for
...
+    '#type' => 'fieldset',
+    '#title' => t('SVG Images dimensions (attributes)'),
+    '#tree' => TRUE,
+  );
+
+  $element['image_svg_settings']['width'] = array(
+    '#type' => 'textfield',
+    '#title' => t('Width'),
+    '#size' => 10,
+    '#field_suffix' => 'px',
```
klonos: In the image field settings form, we are using the full word "pixels" and using a "compound field" for both width/height:
image

It seems like a more streamlined UI, so perhaps we should use it here as well for consistency.

core/modules/image/image.module
```diff
@@ -1349,3 +1359,173 @@ function image_filter_keyword($value, $current_pixels, $new_pixels) {
...
+ * an SVG image, attempt to set those dimensions. This helps in particular when
+ * uploading an image for the first time in a content form.
+ */
+function image_preprocess_image_style(&$variables) {
+  $uri = $variables['path'];
+  // Only try to get file information if $uri is not empty.
+  if (!empty($uri)) {
+    $is_svg = image_is_svg($uri);
+
+    if ($is_svg && empty($variables['width']) && empty($variables['height'])) {
```
klonos: Is it possible to have an svg with only one of the width/height dimension defined? If so, then we should be checking if either is empty - not both. Then we should be setting any missing dimension to its default value.

core/modules/image/image.field.inc
docwilmot: I'm removing the `Image (SVG Support)` formatter completely. This was added by the SVG Image module, but since we're merging into core, we just need one image formatter.

core/modules/image/image.field.inc
docwilmot: DOne

core/modules/image/image.field.inc
docwilmot: Good idea, done.

core/modules/image/image.field.inc
docwilmot: Fixed

core/modules/image/image.field.inc
docwilmot: Fixed

core/modules/image/image.field.inc
docwilmot: No, `px` not needed. This results in (for example) `height='500'` attribute which defaults to `500px`.

core/modules/image/image.module
docwilmot: This is actually checking if you're calling `theme_image_style()` with neither width nor height, in which case we use defaults. If either one is available, then the image would be scaled to that dimension.

docwilmot:
Preview not found!

indigoxela:
I do have some questions re the current implementation:

Currently the library location is 'core/modules/image/libraries', and hook_autoload_info() is also in the image module.
That means, that the functionality to sanitize SVG is only available, if the image (field) module is enabled. That's for sure usually the case, but the module's not required. Should we consider that?

Besides that I wonder, if hook_autoload_info is actually necessary in this case. And I also wonder, why there's an abstraction layer for the sanitizer class. (Is someone supposed to come up with an alternative sanitizer class?) And then, still all (sub-)classes of this sanitizer would get loaded on every page call (because of hook_autoload_info). Even if it wouldn't ever be needed, for instance if no field allows svg's.

core/modules/image/image.module
```diff
@@ -1349,3 +1369,242 @@ function image_filter_keyword($value, $current_pixels, $new_pixels) {
...
+ * Implements hook_file_url_alter().
+ *
+ * Prevent image_style_url() from using a different URL for SVG images. Because
+ * image styles cannot operate on SVGs, set the URL back to the original image.
+ * This affects theme_image_style(), which normally would point to a new image.
+ */
+function image_file_url_alter(&$url) {
+  if (!empty($url)) {
+    // If this is an SVG image in the styles directory...
+    if (strpos($url, '/styles/') && strpos($url, '.svg')) {
```
indigoxela: strpos() is one of those functions, that require strict comparison to avoid "misunderstandings", example: 'strpos($url, '/styles/') !== FALSE'

core/includes/image.inc
```diff
@@ -526,6 +540,62 @@ function image_hex2rgba(&$hex) {
...
+    'width' => IMAGE_SVG_DEFAULT_WIDTH,
+    'height' => IMAGE_SVG_DEFAULT_HEIGHT,
+  );
+
+  $file_contents = file_get_contents($uri);
+  if ($file_contents) {
+    $svg = simplexml_load_string($file_contents);
+    foreach ($svg->attributes() as $attribute => $value) {
+      if ($attribute === 'width' || $attribute === 'height') {
+        $svg_dimensions[$attribute] = (integer) $value;
```
indigoxela: This long form for type casting is quite uncommon in B. Until now I only saw `(int)`. Maybe be consistent with that?

core/modules/system/system.module
```diff
@@ -3916,6 +3916,16 @@ function system_image_toolkits() {
...
 }
 
+/**
+ * Get core SVG support.
+ */
+function system_get_svg_support() {
+  // @todo: static? Check how often this would be called.
+  // @todo: Maybe this should be in image module.
+  $image_svg_support = config_get('system.core', 'image_svg_support');
+  return $image_svg_support;
```
indigoxela: Currently this can be true or null (and whatever people set the config value to), maybe make it true boolean (true/false)?

core/modules/image/config/image.settings.json
```diff
@@ -0,0 +1,4 @@
+{
+    "_config_name": "image.settings",
+    "svg_sanitizer_class": "enshrined\\svgSanitize\\Sanitizer"
```
indigoxela: What was the intention behind that extra config file? Is it supposed to contain other items?
To me it seems a bit overdone to create an extra json file for one config option.

Oh, and don't forget the update hook for existing sites, as otherwise there won't be any svg sanitizing. And admins wouldn't even know.

core/modules/file/config/file.settings.json
```diff
@@ -3,7 +3,7 @@
     "_module": "file",
     "max_filesize": "",
     "default_file_directory": "",
-    "default_allowed_extensions": "jpg jpeg gif png txt doc docx xls xlsx pdf ppt pptx pps ppsx odt ods odp mp3 mov mp4 m4a m4v mpeg avi ogg oga ogv weba webp webm",
+    "default_allowed_extensions": "jpg jpeg gif png txt doc docx xls xlsx pdf ppt pptx pps ppsx odt ods odp mp3 mov mp4 m4a m4v mpeg avi ogg oga ogv weba webp webm svg",
```
indigoxela, line 6: Shouldn't that svg have been added to "file_default_allowed_extensions" in system.core.json instead? This might be the wrong spot, if it's about extensions for /file/add.

core/includes/image.inc
docwilmot: To be corrected next PR, thanks.

core/modules/file/config/file.settings.json
docwilmot: Interesting question, but I cannot find where in core `file_default_allowed_extensions` is actually used. Could you point me to it?

indigoxela: Hm... I think, you're right. It *is* default_allowed_extensions from file.settings.json. I always struggle to figure out, which one strikes.
Consider this as resolved (I have no permission to set this here):

docwilmot: Created https://github.com/backdrop/backdrop-issues/issues/6298 to explore what thats all about.

core/modules/image/config/image.settings.json
docwilmot: > What was the intention behind that extra config file? Is it supposed to contain other items?

I feel it would be helpful to expose some way to either change or extend the sanitizer. This is the easiest way, I suspect. Alternatives considered:
- just this: a config file entry in Image module
- a config entry in another module (File, System, etc). But the sanitizer is just used by Image fields so if Image is disabled, its not needed. Still could go in System.
- just force this sanitizer, dont allow extending or changing, so no config file needed
- create a new hook `hook_svg_sanitizer()`
- add a configuration form to select available sanitizers provided by contrib or custom (would still save to config)

For something that is unlikely to be frequently changed, I thought just the config entry should be good enough. Any other ideas welcome.

core/modules/image/image.module
docwilmot: Agreed, will change.

core/modules/system/system.module
docwilmot: Agreed, will change.

core/modules/file/config/file.settings.json
indigoxela: Good idea, this is really confusing. :+1:

core/modules/image/config/image.settings.json
indigoxela: Right, no image module - no sanitizer... :thinking:

I don't think, we need an admin UI for sanitizer class selection. I'm not even sure, if we need the option to replace the class, anyway. But if so, and given it's a setting hardly ever changed, ... a hook instead of config? That seems worth considering.

If someone wants to _replace_ the class, there's some coding required, anyway.

core/modules/image/image.module
docwilmot: Actually, this is OK as is since neither string can ever be at the start of the string.

core/modules/image/config/image.settings.json
docwilmot: Thanks @indigoxela I'm leaving this open for other opinions for now.

core/includes/file.inc
```diff
@@ -1881,6 +1881,9 @@ function file_validate_size(File $file, $file_limit = 0, $user_limit = 0) {
  */
 function file_validate_is_image(File $file) {
   $errors = array();
+  if ('image/svg+xml' === mime_content_type($file->uri)) {
```
herbdool: Instead of this, have you considered, adding the svg support to `image_get_info` so that svg support is more integrated? A good reason to make this more integrated is that the toolkits might also support various svg commands. I think ImageMagick might.

core/modules/image/config/image.settings.json
herbdool: I agree about simplifying the sanitizer. No need for a UI or config. And perhaps don't even need the class. We might only need to pass it through something like: `$svg_contents = filter_xss($svg_contents, array('svg', 'use', 'title', 'desc', 'defs', 'linearGradient', 'stop', 'rect', 'circle', 'path'));`, similar to how it's done in this PR https://github.com/backdrop/backdrop/blob/df47d525efd5f074b4a73161e20c80612cd83909/core/includes/icon.inc. Maybe there could be a hook so that a contrib module could override that.

core/modules/image/image.field.inc
```diff
@@ -458,6 +462,10 @@ function image_field_widget_form(&$form, &$form_state, $field, $instance, $langc
     backdrop_alter('image_supported_extensions', $supported_extensions);
     $extensions = isset($elements[$delta]['#upload_validators']['file_validate_extensions'][0]) ? $elements[$delta]['#upload_validators']['file_validate_extensions'][0] : implode(' ', $supported_extensions);
     $extensions = array_intersect(explode(' ', $extensions), $supported_extensions);
+    // Todo: Decide if we're keeping this permission.
+    if (!user_access('upload svg images')) {
```
herbdool: If we're defaulting to stripping out all xss then I'd say this isn't necessary.

core/modules/image/image.module
```diff
@@ -1349,3 +1359,186 @@ function image_filter_keyword($value, $current_pixels, $new_pixels) {
...
+    if ($svg_dimensions) {
+      $file->metadata['width'] = $svg_dimensions['width'];
+      $file->metadata['height'] = $svg_dimensions['height'];
+    }
+  }
+}
+
+
+/**
+ * Implements hook_field_attach_presave().
```
herbdool: This feels intensive - like it was a contrib module that couldn't change core in a better way. Isn't there a way to set the width and height in the image or file element earlier? Instead of having to find image fields this way.

herbdool (changes requested):

core/modules/image/image.module
docwilmot: In progress, but removed that `hook_field_attach_presave()` implementation, now that we get image dimensions from `image_get_info()`

core/modules/image/image.field.inc
docwilmot: Removed sanitizer completely.

core/includes/file.inc
docwilmot: Made sense, done.

core/modules/image/config/image.settings.json
docwilmot: I think the weight of opinion is firmly against this sanitizer, so, its gone! 😄

core/modules/image/image.field.inc
```diff
@@ -759,32 +824,74 @@ function image_field_formatter_view($entity_type, $entity, $field, $instance, $l
...
-    $element[$delta] = array(
-      '#theme' => 'image_formatter',
-      '#item' => $item,
-      '#image_style' => $display['settings']['image_style'],
-      '#path' => isset($uri) ? $uri : '',
-    );
+    else {
+      $svg_path = file_create_url($item['uri']);
+      $svg = file_get_contents($svg_path);
+      $svg_clean = filter_xss($svg, array('svg', 'use', 'title', 'desc', 'defs', 'linearGradient', 'stop', 'rect', 'circle', 'path'));
```
indigoxela: Oops, this does a little too much. It breaks perfectly harmless and valid svg's. :stuck_out_tongue_winking_eye: 

Maybe after some discussion we'll end up to always just display SVG in the image :point_left: formatter as img tag, and leave plain SVG as markup to contrib... :shrug:

herbdool: Oh, it's only sanitizing on inline display. If that's the case, only having an inline img tag option is ok by me.

indigoxela: The latest commit seems to have fixed the sanitizer - it no longer breaks a valid SVG. :+1: 

Whether we allow direct markup display in core, or leave it to contrib, is something still to determine, I guess.

core/modules/image/image.field.inc
```diff
@@ -759,33 +822,153 @@ function image_field_formatter_view($entity_type, $entity, $field, $instance, $l
...
+      }
+    }
   }
 
   return $element;
 }
+
+/**
+ * Provides a list of allowed SVG tags.
+ */
```
indigoxela: Could we add some documentation link to the comment - where this tag list info's from and could get verified? In case, such a link is available, of course.

core/modules/image/image.field.inc
```diff
@@ -647,6 +653,10 @@ function image_field_formatter_info() {
 function image_field_formatter_settings_form($field, $instance, $view_mode, $form, &$form_state) {
   $display = $instance['display'][$view_mode];
   $settings = $display['settings'];
+  $svg_allowed = FALSE;
+  if (isset($instance['settings']['file_extensions'])) {
+    $svg_allowed = strpos($instance['settings']['file_extensions'], 'svg') !== FALSE;
+  }
```
quicksketch: Let's be more precise here to match exactly `svg` and not other extensions like `foosvg` or `svg.xml`.

core/modules/image/image.field.inc
```diff
@@ -759,33 +822,153 @@ function image_field_formatter_view($entity_type, $entity, $field, $instance, $l
...
+    'feOffset',
+    'fePointLight',
+    'feSpecularLighting',
+    'feSpotLight',
+    'feTile',
+    'feTurbulence',
+    'filter',
+    'foreignObject',
+    'g',
+    'iframe',
```
quicksketch: Seems like maybe iframe shouldn't be allowed in SVGs? `video`, `audio`, and `style` also give me pause.

indigoxela: `foreignObject` is another candidate for removal here. The [default example on developer.mozilla.org](https://developer.mozilla.org/en-US/docs/Web/SVG/Element/foreignObject) looks harmless, but there's more things one can do with it.

core/modules/image/image.field.inc
docwilmot: https://www.w3.org/TR/SVG/eltindex.html

core/modules/image/image.field.inc
docwilmot: The list is from here: https://www.w3.org/TR/SVG/eltindex.html

I dont know how we can reach consensus, but if you all dont mind I'm going to leave this list as is until there is agreement as to which should be in here and which should not, rather than pecking them out one by one (and probably putting them back in again). It'll only be a simple delete commit and push, so no rush.

core/modules/system/config/system.core.json
```diff
@@ -25,7 +25,7 @@
     "field_purge_batch_size": 200,
     "file_not_normalized_schemes": [],
     "file_additional_public_schemes": [],
-    "file_default_allowed_extensions": "jpg jpeg gif png txt doc docx xls xlsx pdf ppt pptx pps ppsx odt ods odp mp3 mov mp4 m4a m4v mpeg avi ogg oga ogv weba webp webm",
+    "file_default_allowed_extensions": "jpg jpeg gif png txt doc docx xls xlsx pdf ppt pptx pps ppsx odt ods odp mp3 mov mp4 m4a m4v mpeg avi ogg oga ogv svg weba webp webm",
```
docwilmot, line 28: Just a note again (its been mentioned elsewhere I'm sure) that this is not used anywhere in Backdrop); should we delete it?

core/modules/image/image.field.inc
indigoxela: Ah, many thanks...
I actually meant to add it to the doc block with an `@see` :grinning:

core/modules/image/image.field.inc
indigoxela: > I'm going to leave this list as is until there is agreement as to which should be in here and which should not...

Agreed. :+1: Let's discuss this properly.

core/modules/system/config/system.core.json
indigoxela: This should probably get its own issue for proper discussion. Yes, this config item caused me confusion more than once. :wink:

core/includes/image.inc
```diff
@@ -526,6 +548,62 @@ function image_hex2rgba(&$hex) {
...
+  }
+
+  $svg_dimensions = array(
+    'width' => IMAGE_SVG_DEFAULT_WIDTH,
+    'height' => IMAGE_SVG_DEFAULT_HEIGHT,
+  );
+
+  $file_contents = file_get_contents($uri);
+  if ($file_contents) {
+    $svg = simplexml_load_string($file_contents);
```
indigoxela: We have to catch here, that simplexml_load_string() returns FALSE on invalid (unparseable) input.
Otherwise the page breaks with `Error: Call to a member function attributes() on bool in image_get_svg_dimensions()`

herbdool: Looks like this is still to do.

core/modules/system/config/system.core.json
docwilmot: Issue exists, closing.

core/includes/image.inc
docwilmot: Done

core/includes/image.inc
```diff
@@ -162,6 +174,18 @@ function image_get_info($filepath, $toolkit = FALSE) {
     if (isset($details) && is_array($details)) {
       $details['file_size'] = filesize($filepath);
     }
+    elseif (image_is_svg($filepath)) {
+      // Dimensions could be FALSE if the SVG is invalid, although we should
+      // have caught that on upload.
+      $dimensions = image_get_svg_dimensions($filepath);
+      $details = array(
+        'width'     => isset($dimensions['width']) ? $dimensions['width'] : IMAGE_SVG_DEFAULT_WIDTH,
```
indigoxela, line 182: I guess, this extra check (isset) is unnecessary now that a FALSE returned by simplexml_load_string() is already caught in function image_get_svg_dimensions(), which could already make sure, valid dimensions are returned.

core/includes/image.inc
```diff
@@ -526,6 +550,65 @@ function image_hex2rgba(&$hex) {
...
+  $file_contents = file_get_contents($uri);
+  if ($file_contents && $svg = simplexml_load_string($file_contents)) {
+    foreach ($svg->attributes() as $attribute => $value) {
+      if ($attribute === 'width' || $attribute === 'height') {
+        $svg_dimensions[$attribute] = (int) $value;
+      }
+    }
+  }
+  else {
+    return FALSE;
```
indigoxela, line 647: The return of FALSE doesn't provide value here, I belief.
We could simply stick with defaults in this case. Or am I misunderstanding something?

And also: this happens before the static cache is populated... :thinking:

core/includes/file.inc
```diff
@@ -2018,6 +2018,64 @@ function file_validate_image_orientation(File $file, $orientation = 0) {
...
+
+    $errors = libxml_get_errors();
+
+    if (!$xml || !empty($errors)) {
+      return array(t('Invalid SVG file.'));
+    }
+    else {
+      $namespaces = $xml->getNamespaces();
+      if (!in_array('http://www.w3.org/2000/svg', $namespaces)) {
+        $errors[] = t('Invalid SVG namespace.');
```
indigoxela, line 2052: Hm... it's actually not invalid, but missing. That's not per se forbidden, as far as I read in the docs, but extremely uncommon. Needs more recherche. :thinking:

core/includes/image.inc
```diff
@@ -112,7 +122,9 @@ function image_get_supported_extensions() {
   $toolkit = image_get_toolkit();
   $function = 'image_' . $toolkit . '_supported_extensions';
   if (function_exists($function)) {
-    return $function();
+    $extensions = $function();
+    $extensions[] = 'svg';
```
indigoxela, line 126: I'm not sure yet, if always adding svg is a good idea... But then - we don't have a condition to apply here. Or do we?

core/includes/file.inc
```diff
@@ -2018,6 +2018,64 @@ function file_validate_image_orientation(File $file, $orientation = 0) {
...
+ *   dangerous content, it will contain an error message.
+ *
+ * @see hook_file_validate()
+ */
+function file_validate_svg(File $file) {
+  $errors = array();
+
+  if (image_is_svg($file->uri)) {
+    $file_contents = file_get_contents($file->uri);
+    $xml = simplexml_load_string($file_contents);
```
indigoxela, line 2042: OK, we have to wrap this in libxml_use_internal_errors() to capture errors (as in my example).

core/includes/file.inc
```diff
@@ -2018,6 +2018,64 @@ function file_validate_image_orientation(File $file, $orientation = 0) {
...
+ * Validate uploaded SVG files.
+ *
+ * @param File $file
+ *   A file entity.
+ *
+ * @return
+ *   An array. If the file is an SVG and is either not valid or contains
+ *   dangerous content, it will contain an error message.
+ *
+ * @see hook_file_validate()
```
indigoxela, line 2034: The function name below doesn't seem like actually having anything to do with hook_file_validate, though.

Is this supposed to implement the hook?

docwilmot: All the other `file_validate_*()` functions in `file.inc` reference ` @see hook_file_validate()`, so I think this one should follow same pattern.

core/includes/file.inc
docwilmot: > we have to wrap this in libxml_use_internal_errors() to capture errors 

It wasntnecessary to do so on my local machine. The reason the upload validation wasnt working was actually the `if (image_is_svg($file->uri))` check was failing. Fixed it, but no sandbox yet to test (need rebase again!).

core/includes/image.inc
docwilmot: `$extensions` array holds all image types that image fields support. The others depend on toolkit support of course. I cant think of any condition to add here either. Either add svg or don't.

core/includes/file.inc
docwilmot: Yeah, I had to think about this. It seems the namespace declaration is recommended, but not required? Maybe should remove this.

core/includes/image.inc
docwilmot: My thinking is that without explicitly returning FALSE here, for an invalid or empty file, the function would return the default dimensions, which wouldnt make sense really.

core/includes/image.inc
docwilmot: Actually if `$dimensions` is FALSE then means its an invalid file, so the whole function should just return FALSE, come to think of it.

core/includes/file.inc
indigoxela: Removing seems like an option, yes. Will test possible side effects.

core/includes/image.inc
indigoxela: Makes sense...
This utility function could get called on arbitrary URIs or actually strings.

core/includes/file.inc
```diff
@@ -2018,6 +2018,66 @@ function file_validate_image_orientation(File $file, $orientation = 0) {
...
+ */
+function file_validate_svg(File $file) {
+  $errors = array();
+
+  if ($file->filemime == 'image/svg+xml') {
+    $file_contents = file_get_contents($file->uri);
+    $xml = @simplexml_load_string($file_contents);
```
indigoxela: Suppressing errors with `@` here contradicts the way we try to evaluate errors in the next condition.

I'd still strongly recommend to wrap our libxml checks inside libxml_use_internal_errors() to actually *capture* them.
I can provide a sample file, where this makes a real difference.

core/includes/file.inc
indigoxela: We could drop that check and make our xpath queries a little more greedy to still catch naughty content.
However, I can't get Firefox nor Chrome to display such svg *as image*. Both seem to be like "svg namespace or bust". :wink: Locally, in other software, those crippled svg's still render fine as image, but not in browsers. And browser rendering is what people need in a web CMS.

So I'd say let's stick with the namespace check - which also validates that it's svg and not some arbitrary xml.

docwilmot: Yep I found the same, lots of errors without that namespace, so keeping, agreed.

core/includes/file.inc
docwilmot: The `@` just suppresses *displaying* the errors, just to hide all PHP error messages from displaying (if the file is invalid, PHP throws a *whole bunch* of messages about xml elements. we dont need those here, theyre in the log) . All we need for the upload validator is the simple message. Using `libxml_use_internal_errors()` doesnt help with this AFAICT.

indigoxela: > we dont need those here, theyre in the log...

TBH, I wouldn't want a crappy file to clutter the log, when uploading.

> Using libxml_use_internal_errors() doesnt help with this AFAICT.

It does in my local testing. We could even process the nagging to display a more verbose message, to tell people why this file is crap.
This error handling might eventually be different in different PHP versions. Will do further testing later.

indigoxela: A lot more testing... When using libxml_use_internal_errors() (true/false) wrapped around, *nothing* shows up in the logs.

Tested with PHP 5.6.40, 7.4.33 and 8.2.18 in combination with libxml version 2.9.14.
Can we please use libxml_use_internal_errors? It's more useful here than just the `@` operator.

docwilmot: Youre right of course. Maybe didnt flush cache or something before. Will PR later. 

P.s. I can only work on this for brief moments before and after my day job (I'm off now for the next 8-9 hours). If you want to take over this one it would be great, I am also working on other PRs too and would like to see this get in by feature freeze.

indigoxela: > If you want to take over this one it would be great, I am also working on other PRs too...

I belief, this PR is in a good shape and already a valid candidate for 1.28. Remember: it doesn't have to be fully done by feature freeze. Just ready enough to look promising - and I think it is. :wink: 

Yes, I'm aware that you're involved in other 1.28 candidates. That's a bit unfortunate, and I hope, Backdrop doesn't eat too much of your sparse spare time. :pleading_face:

I don't think, it's necessary that I take over. I'll just try to be more patient. :innocent: Didn't mean to stress you out.

docwilmot: Just to be clear, I was genuinely hoping you could take this over, because I am concerned I might not have enough time and you're well familiar with it already. I wasnt at all complaining that you were stressing me out. 😄

indigoxela: :smiley: phew!
Will think it over.

herbdool: If there are just small changes here and there I could edit @docwilmot's branch.

core/includes/file.inc
herbdool: I'm going to tentatively close this. I assume it's because `file_validate()` calls all these other validation functions and allow the hook to change the results.

core/includes/file.inc
herbdool: I wrapped `libxml_use_internal_errors()` and it's capturing invalid svg and returning "Invalid SVG file." There's nothing in the logs so I removed that. We can have a TODO to output the svg errors to the screen in a friendly way.

It would be good to have a test with an invalid svg and a dirty svg.

I noticed that this function isn't called if creating a file from `admin/content/files`.

core/modules/image/image.field.inc
```diff
@@ -452,6 +454,7 @@ function image_field_widget_form(&$form, &$form_state, $field, $instance, $langc
     if ($settings['max_dimensions'] || $settings['min_dimensions']) {
       $elements[$delta]['#upload_validators']['file_validate_image_resolution'] = array($settings['max_dimensions'], $settings['min_dimensions']);
     }
+    $elements[$delta]['#upload_validators']['file_validate_svg'] = array();
```
herbdool: This doesn't seem necessary now that it's called from `file_save_upload()`.

core/includes/file.inc
herbdool: I fixed it so it gets called for file uploads too.

core/includes/image.inc
herbdool: I think this is resolved.

core/includes/image.inc
herbdool: I think this is okay as-is. If no toolkit needed then it's always supported.

core/includes/file.inc
```diff
@@ -2018,6 +2021,67 @@ function file_validate_image_orientation(File $file, $orientation = 0) {
...
+ */
+function file_validate_svg(File $file) {
+  $errors = array();
+
+  if ($file->filemime == 'image/svg+xml') {
+    $file_contents = file_get_contents($file->uri);
+    libxml_use_internal_errors(TRUE);
+    $xml = simplexml_load_string($file_contents);
+    $errors = libxml_get_errors();
+    libxml_use_internal_errors(FALSE);
```
indigoxela: That reverting of internal error's just a little too soon, as we still do libxml stuff below.

core/modules/image/image.module
```diff
@@ -1349,3 +1353,59 @@ function image_filter_keyword($value, $current_pixels, $new_pixels) {
...
+
+/**
+ * Preprocess function for theme_image_style().
+ *
+ * If theme_image_style() is called without setting explicit height/width for
+ * an SVG image, attempt to set those dimensions. This helps in particular when
+ * uploading an image for the first time in a content form.
+ */
+function image_preprocess_image_style(&$variables) {
+  $uri = $variables['path'];
```
indigoxela: That should rather be `$variables['uri'];`

core/modules/image/image.module
```diff
@@ -1349,3 +1353,59 @@ function image_filter_keyword($value, $current_pixels, $new_pixels) {
...
+  // Only try to get file information if $uri is not empty.
+  if (!empty($uri) && image_is_svg($uri)) {
```
indigoxela, line 1367: Seems like image_is_svg() fails here. How about `$variables['filemime'] == 'image/svg+xml')`?

Otherwise this condition doesn't work.

core/modules/image/image.module
```diff
@@ -1349,3 +1353,59 @@ function image_filter_keyword($value, $current_pixels, $new_pixels) {
...
+    if (empty($variables['width']) && empty($variables['height'])) {
```
indigoxela: Seems like width and height got populated wrong (with defaults).
This might indicate that something else went wrong in an earlier step.

Update: found, where these defaults get applied, ignoring original dimensions (annoying with smaller svgs like icons...)

core/modules/image/image.module
```diff
@@ -1349,3 +1353,59 @@ function image_filter_keyword($value, $current_pixels, $new_pixels) {
...
+      if ($dimensions = image_get_svg_dimensions($uri)) {
```
indigoxela: I don't think, we really still need the call to image_get_svg_dimensions() here (for the formatter). We won't get any better values than we already have in the File object. These values should've already been set via image_file_presave().

Or am I missing something? Could it be that this is about previously existing SVG files or so (before support was added)?

core/includes/file.inc
herbdool: Fixed.

core/modules/image/image.module
herbdool: Fixed

herbdool (approved):

core/modules/image/image.field.inc
```diff
@@ -699,6 +704,40 @@ function image_field_formatter_settings_form($field, $instance, $view_mode, $for
   $element['image_load']['auto']['#description'] = t('Each browser will determine whether or not to lazily load the image.');
   $element['image_load']['eager']['#description'] = t('Load this image immediately, regardless of where it is located on the page.');
 
+  $element['image_svg_dimensions'] = array(
+    '#type' => 'item',
+    '#title' => t('Maximum SVG image dimensions'),
+    '#field_prefix' => '<div class="container-inline">',
+    '#field_suffix' => '</div>',
+    '#description' => t('The maximum allowed SVG image dimensions expressed as <code>WIDTH</code> x <code>HEIGHT</code>. If left blank, SVG images will be displayed at @svg_default_width x @svg_default_height pixels. Selecting an image style will override this setting.', array(
+      '@svg_default_width' => IMAGE_SVG_DEFAULT_WIDTH,
```
indigoxela: There's trailing white space in this line.

core/modules/image/tests/image.test
```diff
@@ -223,92 +225,104 @@ class ImageStylesPathAndUrlUnitTest extends BackdropWebTestCase {
...
+      if (!$clean_url && !$image_is_svg) {
+        $this->assertTrue(strpos($generate_url, '?q=') !== FALSE, 'When using non-clean URLS, the system path contains the query string.');
+      }
 
-    if ($scheme == 'private') {
-      $this->assertEqual($this->backdropGetHeader('Expires'), 'Fri, 16 Jan 2015 07:50:00 GMT', 'Expires header was sent.');
-      $this->assertEqual($this->backdropGetHeader('Cache-Control'), 'no-cache, must-revalidate', 'Cache-Control header was set to prevent caching.');
-      $this->assertEqual($this->backdropGetHeader('X-Image-Owned-By'), 'image_module_test', 'Expected custom header has been added.');
+      // Check that the generated URL is the same when we pass in a relative
+      // path rather than a URI. We need to temporarily switch the default
```
indigoxela: Trailing white space in this line.

core/modules/image/tests/image.test
```diff
@@ -223,92 +225,104 @@ class ImageStylesPathAndUrlUnitTest extends BackdropWebTestCase {
...
-      if (strpos($generate_url, '.png') === FALSE ) {
-        $this->fail('Confirming that private image styles are not appended require PNG file.');
+      if (!$image_is_svg) {
+        $this->assertTrue(file_exists($generated_uri), 'Generated file does exist after we accessed it.');
+        $this->assertRaw(file_get_contents($generated_uri), 'URL returns expected file.');
+        $generated_image_info = image_get_info($generated_uri);
+        $this->assertEqual($this->backdropGetHeader('Content-Type'), $generated_image_info['mime_type'], 'Expected Content-Type was reported.');
+
+        // Some web servers may return a "chunked" response and ignore the
+        // Content-Length headers returned by image_style_deliver(). Allow
```
indigoxela: Trailing white space...

core/modules/image/tests/image.test
```diff
@@ -223,92 +225,104 @@ class ImageStylesPathAndUrlUnitTest extends BackdropWebTestCase {
...
+
+        $this->backdropGet($generate_url_no_access);
+        $this->assertResponse(403, 'Confirmed that access is denied for the private image style.');
+        // Verify that images are not appended to the response. Currently this
+        // test only uses PNG images.
+        if (strpos($generate_url, '.png') === FALSE && strpos($generate_url, '.svg') === FALSE ) {
+          $this->fail('Confirming that private image styles are not appended require PNG file.');
+        }
+        elseif (strpos($generate_url, '.png') === TRUE) {
+          // Check for PNG-Signature
```
indigoxela: Another trailing white space...

core/modules/image/image.field.inc
```diff
@@ -759,26 +810,42 @@ function image_field_formatter_view($entity_type, $entity, $field, $instance, $l
...
+      }
+
+      $is_svg = image_is_svg($item['uri']);
+
+      // Add custom attributes only for SVG images.
+      if ($is_svg) {
+        $image_svg_dimensions = $display['settings']['image_svg_dimensions'];
+
+        foreach ($image_svg_dimensions as $attribute => $attribute_value) {
+          if (!empty($attribute)) {
```
indigoxela: Problem: above `$image_svg_dimensions` contains the *fallback* dimensions. But here they're applied regardless.

Could it be, the intended check was meant to be `if (empty($item[$attribute])) {` - to apply the value only, if there's no value set?

And dimensions according to style would go to the preprocessor.

core/modules/image/image.module
```diff
@@ -1349,3 +1353,59 @@ function image_filter_keyword($value, $current_pixels, $new_pixels) {
...
 }
+
+/**
+ * Preprocess function for theme_image_style().
+ *
+ * If theme_image_style() is called without setting explicit height/width for
+ * an SVG image, attempt to set those dimensions. This helps in particular when
+ * uploading an image for the first time in a content form.
+ */
+function image_preprocess_image_style(&$variables) {
```
indigoxela, line 1364: My current suspicion is, that this whole function's unnecessary.

EDIT: yay, found its purpose, finally.

indigoxela (changes requested):
More questions than answers, but there's definitely some work to do.

core/modules/image/image.module
indigoxela: I think, I got this working with two modifications.

```php
function image_preprocess_image_style(&$variables) {
  $uri = $variables['uri'];
  // Only try to get file information if $uri is not empty.
  if (!empty($uri) && image_is_svg($uri)) {
    $original = array(
      'width' => $variables['width'],
      'height' => $variables['height'],
    );
    image_style_transform_dimensions($variables['style_name'], $original);
  }
}
```
And in function image_field_formatter_view do not enforce dimensions, but use fallback as fallback only.

core/includes/image.inc
```diff
@@ -526,6 +550,65 @@ function image_hex2rgba(&$hex) {
...
+ *   invalid, FALSE will be returned.
+ */
+function image_get_svg_dimensions($uri) {
+  // Safety check.
+  if (!image_is_svg($uri)) {
+    return FALSE;
+  }
+
+  // Return cached dimensions if already retrieved once this request.
+  $cached_images = &backdrop_static(__FUNCTION__, array());
```
indigoxela, line 630: Out of interest: why static caching?

If it's called multiple times per request, it's probably called unnecessarily. (For example in the preprocessor.) Or does this have to do with AJAX?

core/modules/image/image.module
indigoxela: Nevermind, seems to work (may have been some caching prob on my side).

core/modules/image/image.field.inc
```diff
@@ -699,6 +704,40 @@ function image_field_formatter_settings_form($field, $instance, $view_mode, $for
   $element['image_load']['auto']['#description'] = t('Each browser will determine whether or not to lazily load the image.');
   $element['image_load']['eager']['#description'] = t('Load this image immediately, regardless of where it is located on the page.');
 
+  $element['image_svg_dimensions'] = array(
+    '#type' => 'item',
+    '#title' => t('Maximum SVG image dimensions'),
+    '#field_prefix' => '<div class="container-inline">',
+    '#field_suffix' => '</div>',
+    '#description' => t('The maximum allowed SVG image dimensions expressed as <code>WIDTH</code> x <code>HEIGHT</code>. If left blank, SVG images will be displayed at @svg_default_width x @svg_default_height pixels. Selecting an image style will override this setting.', array(
```
indigoxela: TBH, I don't understand, what the purpose of this setting is.
We have a default value (constant) as fallback, which already was applied on svg's where dimension attributes were missing (on upload).
We already have image styles - which also ship with dimensions.
Why yet another layer? Could it be, that this is leftover from the markup display?

Can we entirely drop this? Otherwise it feels like reinventing the wheel (image styles).

core/modules/system/config/system.core.json
indigoxela: Why even bothering to update, if this setting's not in use, anyway?

core/modules/image/image.field.inc
herbdool: Do you still think we should drop this? It only applies for image fields and won't help with files uploaded at `admin/content/files` I think. So maybe drop it.

docwilmot: I agree this can go, the size calculations are really too complex. I think this was part of the original code but I added the width constants, for some reason. Perhaps we could even remove the dimension constants as well, and just validate files that dont have set dimensions.

herbdool: Alright, let's get rid of the constants too, and add validation.

herbdool: Removed.

core/includes/image.inc
herbdool: Still seems to be called a couple times even though `image_get_svg_dimensions()` now only gets called once. So I think `image_get_info()` is called multiple times. Or I saw it with multiple images. I can't recall.

core/includes/file.inc
```diff
@@ -2018,6 +2021,74 @@ function file_validate_image_orientation(File $file, $orientation = 0) {
...
+      $xml->registerXPathNamespace('html', 'http://www.w3.org/1999/xhtml');
+      foreach ($search_patterns as $search_pattern) {
+        $found = $xml->xpath($search_pattern);
+        if (!empty($found)) {
+          $errors[] = t('Dangerous content found.');
+          break;
+        }
+      }
+      $attributes = $xml->attributes();
+      if (empty($attributes['width'])) {
```
docwilmot: Apologies for wasting your time, but after that last suggestion I did a long internet search and apparently it is quite OK for SVGs to be missing dimensions, and in fact some tools remove the dimensions. And since we're displaying only with `img` tags, the dimensions could be set on the image tag anyway. So requiring dimensions is probably overkill?

core/includes/file.inc
```diff
@@ -2018,6 +2021,74 @@ function file_validate_image_orientation(File $file, $orientation = 0) {
...
+      $namespaces = $xml->getNamespaces();
+      if (!in_array('http://www.w3.org/2000/svg', $namespaces)) {
+        $errors[] = t('Invalid SVG namespace.');
+      }
+
+      $search_patterns = array(
+        // @see https://developer.mozilla.org/en-US/docs/Web/SVG/Element
+        '//svg:foreignObject',
+        '//svg:script',
+        '//html:iframe',
```
docwilmot, line 2059: The "bad" SVG at https://svg.enshrined.co.uk/ has a few other issues that it seems to prefer to sanitize. Maybe we need some more search patterns.

core/includes/file.inc
herbdool: Perhaps we just remove the requirement. Let them expand to the bounds of the containing div if they don't have dimensions.

herbdool: Removed. I played around with it and a nice SVG file can still be viewed fine. If a `viewBox` is present it will still use it to scale to a certain size even if the width/height are missing.

herbdool: Though I had only tested by adding it from `admin/content/file`. If I insert it into a node I get a fatal error: 

> Incorrect integer value: '' for column 'field_vector_image_width'

So either we still need dimensions or we change a db table (and corresponding code).

core/modules/system/config/system.core.json
herbdool: We will need an update hook if we're changing this setting, even if it's not in use.

core/includes/file.inc
herbdool: I've had it save zeros if there are no dimensions. But then on image field display we set them to blank so that the image can be visible. But still needs a bit of tweaking with this change.

core/includes/file.inc
docwilmot: This catches the XML nodes with remote fill absolute URLs
XPath:
```
        '//*[contains(@fill, "url") and contains(@fill,"http")]',
        '//*[contains(@fill, "url") and contains(@fill,"//")]',
        '//*[contains(@fill, "url") and contains(@fill,"ftp")]',

```
XML: 
```
<rect fill="url('http://example.com/benis.svg')"....></rect>
<rect fill="url('https://example.com/benis.svg')"....></rect>
<rect fill="  url(  ' https://example.com/benis.svg '  ) "....></rect>
<rect fill="url('ftp://192.168.2.1/benis.svg')"....></rect>
<rect fill="url('//example.com/benis.svg')"....></rect>
<rect fill="url('/benis.svg')"....></rect> (PASSES)
<rect fill="url('#benis.svg')"....></rect> (PASSES)
```

docwilmot: I havent researched *why* absolute URLs are sanitized but posting in case we agree to validate for those

indigoxela: > I havent researched why absolute URLs are sanitized

That seems to be something, that can get turned on as additional sanitize step. External URLs aren't per se dangerous - have a look at one example usage in the issue - where external URLs to Wikipedia were set. We don't need that anymore for our field formatter, as we don't display svg inline anymore, though.

I'd suggest to discuss the "reject external links or all links" more in depth on the issue, as that isn't just a technical topic. :wink:

core/includes/image.inc
```diff
@@ -162,6 +164,18 @@ function image_get_info($filepath, $toolkit = FALSE) {
     if (isset($details) && is_array($details)) {
       $details['file_size'] = filesize($filepath);
     }
+    elseif (image_is_svg($filepath)) {
+      // Dimensions could be FALSE if the SVG is invalid, although we should
+      // have caught that on upload.
+      $dimensions = image_get_svg_dimensions($filepath);
+      $details = array(
+        'width'     => isset($dimensions['width']) ? $dimensions['width'] : 0,
```
indigoxela: Sorry, but `0` seems like a weird fallback value to me.

core/includes/image.inc
indigoxela: Can get closed as resolved, then. :wink:

core/modules/system/config/system.core.json
indigoxela: @herbdool I didn't suggest to remove it here, but to just not touch it (to not add svg, as that's pointless, anyway).

core/includes/image.inc
indigoxela: Just a formal thing, but can we just not indent with multiple spaces here? I don't think, it's a forbidden thing, but appears to be uncommon in core code.

core/includes/image.inc
```diff
@@ -585,6 +598,53 @@ function image_add_svg_attributes($svg_content, array $attributes) {
...
+  $cached_images = &backdrop_static(__FUNCTION__, array());
+  if (isset($cached_images[$uri])) {
+    return $cached_images[$uri];
+  }
+
+  // If there are no dimensions, store the dimensions as zero.
+  // Required for compatibility with image field storage.
+  $svg_dimensions = array(
+    'width' => 0,
+    'height' => 0,
```
indigoxela: Why zeroes? That's the least useful value IMO. :wink:

core/modules/image/image.module
```diff
@@ -962,6 +963,9 @@ function image_style_flush($style) {
  * @see image_style_deliver()
  */
 function image_style_url($style_name, $uri) {
+  if (image_is_svg($uri)) {
+    return file_create_url($uri);
```
indigoxela, line 967: Hm. This return happens before the clean_url check below. Isn't that problematic?

core/modules/image/image.field.inc
```diff
@@ -759,26 +760,39 @@ function image_field_formatter_view($entity_type, $entity, $field, $instance, $l
...
+          'path' => file_create_url($item['uri']),
+          'options' => array(),
+        );
+      }
+
+      // SVG was likely uploaded with no dimensions. The width and height are stored
+      // as zero for compatibility with other images, but for viewing set them blank.
+      $is_svg = image_is_svg($item['uri']);
+      if ($is_svg) {
+        if ($item['width'] == 0 && $item['height'] == 0) {
```
indigoxela: Seems to conflict with the preprocessor?

core/modules/image/image.module
herbdool: Hm, I don't know this section yet.

core/includes/image.inc
herbdool: The idea is that if it can't find dimensions on the SVG it'll default to zero. I tried null but that caused a fatal error when saving an image field, since the image field stores the height and width as integers.

core/includes/file.inc
herbdool: And maybe it's not a blocker either. We can tighten up the validation later.

core/modules/image/image.field.inc
herbdool: @docwilmot actually the constants are in the drupal module too. https://git.drupalcode.org/project/svg_image/-/blob/7.x-1.x/svg_image.module?ref_type=heads

core/includes/image.inc
indigoxela: We're setting this weird `0` in two places, now. But as it's about dimensions, why not setting some sane defaults inside that utility function and from then on rely on those?

IF the SVG is invalid, then no browser will display anything useful. Better not display at all, then?

core/includes/image.inc
indigoxela: I got, that this "0" is to have something to save to the database. :+1: 

However, this value's quite problematic for image_styles (resizing doesn't happen, as "0" is always smaller than any image style setting will be).
Another concern: using that as fallback forces us to override later, anyway. We can never use that value as-is ... Not helpful, right? :wink:

core/modules/image/image.module
herbdool: I think that's okay since it seems that the clean_url check is only if it needs to run through an image derivative callback. Maybe. I'm now questioning whether want to skip image styles altogether. It's a bit confusing that users can't rotate, crop or do any other effects on the SVG. The svg_image module in D7 does this:

```
$element[$delta] = array(
      '#theme' => 'image_formatter',
      '#item' => $item,
      '#image_style' => $is_svg === FALSE ? $display['settings']['image_style'] : '',
      '#path' => isset($uri) ? $uri : '',
    );
```

And skips `image_style_url()`.

jenlampton: I added a comment with a potential solution to image-style partial SVG support in the parent issue: Perhaps just add a note when something won't work as expected for SVGs?

jenlampton:
@herbdool I am seeing warning on the sandbox for non-SVG images:
`Warning: imagecreatefrompng(): gd-png: libpng warning: iCCP: known incorrect sRGB profile in image_gd_load() (line 247 of /var/lib/tugboat/core/modules/system/image.gd.inc).`

core/includes/image.inc
herbdool: The constants are back!

core/modules/image/image.module
```diff
@@ -1349,3 +1353,54 @@ function image_filter_keyword($value, $current_pixels, $new_pixels) {
...
+      'height' => $variables['height'],
+    );
+    image_style_transform_dimensions($variables['style_name'], $original);
+  }
+}
+
+/**
+ * Implements hook_file_presave().
+ *
+ * Set the dimensions in the "metadata" property used by the File Entity module.
```
quicksketch: File Entity module is part of core, so this should just be:

```suggestion
 * Set the dimensions in the "metadata" property used by File entities.
```

quicksketch:
Merged into 1.x for 1.27.x. Great job @docwilmot, @indigoxela, @herbdool, and @klonos on this PR!

PR #4547, opened 2023-10-18
Issue #6004: Fix coding standards

izmeez:

Partially completed work on `core/includes/c` task.

Needs more work.


izmeez:
Maybe I should indicate to skip-tests? I did not do that initially because I thought they might be needed after adding visibility qualifiers.

PR #4549, opened 2023-10-19
Issue #6554: Fix PDOException when renaming the language column in the url_alias table

### LINKED ISSUE #6554: db_change_field(): renaming and re-adding indices in a single step is problematic in som older versions of MariaDB
This issue was discovered while working on #5496. Now that that issue was closed, I'm opening this here to deal with that problem specifically.

I'll switch the WIP PR to point to this issue here, and I'll loop back to this in order to go through the comments in the other issue, and see which ones make sense to be copied here. For now, copying this specific comment from March 2023:

>  ...I have tested this on Pantheon multiple times (before the upstream MariaDB version was upped), using the scenario @irinaz has been using to trigger the issue. But I guess that that doesn't count, since I was the one that provided the PR. The irony here is that now that we've updated the Backdrop recipe for Pantheon, the issue won't be easily reproducible 😅 
> 
> If people want to test this on their local with lando, know that you can provision the database with a specific version, and both MariaDB and MySQL support that in lando:
> - MySQL: https://docs.lando.dev/mysql/#patch-versions (it will need to be 5.5.30 - _I think_)
> - MariaDB: https://docs.lando.dev/mariadb/#patch-versions (it will need to be 10.0.23, same as the old version used in Pantheon)

Although the D7 to Backdrop upgrade is only one possible use case to have this problem surface, it seems that it is not necessarily tied to or caused by/during that task.

klonos:
Fixes https://github.com/backdrop/backdrop-issues/issues/6554

Replaces https://github.com/backdrop/backdrop/pull/4325


klonos:
I need to look into that PDOException.

PR #4550, opened 2023-10-22
Issue #6240: Create hook_color_info()

### LINKED ISSUE #6240: Allow a hook to alter the list of CSS files parsed by Color module
## Description of the need

Color module provides preset color schemes. `color.inc` provides the schemes, the default colors which are swapped for scheme colors, but also a list of which CSS files are searched for these default colors. The CSS files in that list must also be ones declared in the theme's info file.

It would probably be useful if contrib modules which style their output, could also expose their output to color. So if a module created a new front page slider for example, it would be good if changing theme color schemes would change the slider colors as well, to match the rest of the display.

## Proposed solution

A new hook, that lists CSS file locations to expose to Color module.

None

## Additional information

Note that this would (only) be helpful once we allow color info to be config, in https://github.com/backdrop/backdrop-issues/issues/2554

Backdrop now includes the ability to expose module color styles to Color modules schemes.

docwilmot:
Fixes https://github.com/backdrop/backdrop-issues/issues/6240

### DIFF
--- a/core/modules/color/color.module
+++ b/core/modules/color/color.module
@@ -406,7 +406,7 @@ function color_get_info($theme) {
       $module_color_info['schemes']['default']['colors'] = $module_color_info['default_colors'];
       unset($module_color_info['default_colors']);
       foreach ($module_color_info['css'] as $key => $file) {
-        $module_color_info['css'][$key] = $module . '__' . $file;
+        $module_color_info['css'][$key] = $module . '_css_path_' . $file;
       }
       $info = backdrop_array_merge_deep($info, $module_color_info);
     }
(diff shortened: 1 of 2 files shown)


core/modules/color/color.module
```diff
@@ -396,6 +400,17 @@ function color_get_info($theme) {
       $info = backdrop_array_merge_deep($info, $base_info);
     }
 
+    $module_color_info = array();
+    foreach (module_implements('color_info') as $module) {
+      $module_color_info = module_invoke($module, 'color_info');
+      $module_color_info['schemes']['default']['colors'] = $module_color_info['default_colors'];
+      unset($module_color_info['default_colors']);
+      foreach ($module_color_info['css'] as $key => $file) {
+        $module_color_info['css'][$key] = $module . '__' . $file;
```
hosef: This is probably not a good idea. Many CSS style guides recommend using __ as a separator and if that is done in file names it would break the code trying to find the files.

docwilmot: > Many CSS style guides recommend using __ as a separator 

Oh didnt know that, thanks. Any separator would do for this purpose though, so this is not major. Maybe `_css_path_`.

avpaderno: If there is any possibility that `$module . '__' . $file` is already used as filename, I would rather use `$module . '_color_info_' . $file`. I didn't check in which cases `$module_color_info['css'][$key]` is used, which means I cannot say `$module . '__' . $file` is effectively used for other purposes by existing modules/themes.

PR #4556, opened 2023-10-31
Issue #6281: Fix redirect table schema of upgraded sites

### LINKED ISSUE #6281: PHP 8.2 Deprecated function: Creation of dynamic property Redirect::$status is deprecated in Redirect->__construct() (line 120 of /core/modules/redirect/redirect.class.inc).
## Description of the bug

Maybe got missed in #5927 or maybe is an edge case.

Got this notice on node edit form

> Deprecated function: Creation of dynamic property Redirect::$status is deprecated in Redirect->__construct() (line 120 of /app/docroot/core/modules/redirect/redirect.class.inc).

## Steps To Reproduce

To reproduce the behavior:
1. Not sure of exact steps to debug as it appeared on node edit of one node with 1 redirect but didn't appear on another node edit with 5 redirects.

## Actual behavior

Deprecation notice at top of node edit form

image

## Expected behavior

No deprecation notice

## Additional information

- Backdrop CMS version: 1.26.1
- PHP version: 8.2.5

indigoxela:
Fixes https://github.com/backdrop/backdrop-issues/issues/6281

### DIFF
--- a/core/modules/redirect/redirect.install
+++ b/core/modules/redirect/redirect.install
@@ -172,3 +172,16 @@ function redirect_update_1003() {
 
   db_query("UPDATE {redirect} SET status_code = :default_status WHERE status_code = 0", array(':default_status' => $default_status));
 }
+
+/**
+ * Fix redirect table schema of upgraded sites.
+ */
+function redirect_update_1004() {
+  if (db_field_exists('redirect', 'status')) {
+    db_delete('redirect')->condition('status', 0)->execute();
+    if (db_index_exists('redirect', 'status_source_language')) {
+      db_drop_index('redirect', 'status_source_language');
+    }
+    db_drop_field('redirect', 'status');
+  }
+}
--- a/core/modules/redirect/redirect.install
+++ b/core/modules/redirect/redirect.install
@@ -147,6 +147,9 @@ function redirect_update_1001() {
     db_change_field('redirect', 'language', 'langcode', $spec);
     db_add_index('redirect', 'source_langcode', array('source', 'langcode'));
   }
+  // In Drupal 7 the status field marked a redirect as being disabled. In
+  // Backdrop there's no such setting, so we delete all disabled redirects.
+  // @see https://github.com/backdrop/backdrop-issues/issues/6281
   if (db_field_exists('redirect', 'status')) {
     db_delete('redirect')->condition('status', 0)->execute();
   }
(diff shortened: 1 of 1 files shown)


yorkshire-pudding (approved):
Coding standards look fine and code makes sense

core/modules/redirect/redirect.install
```diff
@@ -147,6 +147,9 @@ function redirect_update_1001() {
     db_change_field('redirect', 'language', 'langcode', $spec);
     db_add_index('redirect', 'source_langcode', array('source', 'langcode'));
   }
+  if (db_field_exists('redirect', 'status')) {
```
klonos, line 153: Since this was a complex thing to pinpoint and explain, I think that we should add an inline comment here, pointing to the issue in our core queue. It should explain why we have added this cleanup here.

core/modules/redirect/redirect.install
```diff
@@ -172,3 +175,15 @@ function redirect_update_1003() {
 
   db_query("UPDATE {redirect} SET status_code = :default_status WHERE status_code = 0", array(':default_status' => $default_status));
 }
+
+/**
+ * Fix redirect table schema of upgraded sites.
+ */
+function redirect_update_1004() {
+  if (db_field_exists('redirect', 'status')) {
+    if (db_index_exists('redirect', 'status_source_language')) {
```
klonos, line 187: Could there be any edge cases where this column exists without the `status` column? Shouldn't we be checking for both of these and deleting them independently?

indigoxela: That's actually an index, not a column. It might exist or not, but not if the column's not there.

Both, the column and the index won't exist on a native Backdrop install. When upgrading from Dupal, we should be cautious and not assume that the schema is exactly as in D7 hook_schema. It might have changed over time.

core/modules/redirect/redirect.install
indigoxela: Sure, any text suggestions?

bugfolder: I'll suggest this (subject to `* ` prefix and rewrapping for 80 characters):

In D7, the status field marked a redirect as being disabled. In Backdrop, there is no such setting, so we delete all disabled redirects.
@see https://github.com/backdrop/backdrop-issues/issues/6281

quicksketch:
This looks great! I merged into 1.x and 1.26.x.

PR #4560, opened 2023-11-06
Issue #6274: Fix date_popup and date_text validation.

### LINKED ISSUE #6274: Validation error for date_popup is missing the element title
## Description of the bug

For the `date_popup` element, when a required date is empty, the validation error is missing the element title.

image

## Steps To Reproduce

To reproduce the behavior:

Add this element to a form.

```
  $form['my_date'] = array(
    '#type' => 'date_popup',
    '#date_format' => 'l F j, Y',
    '#title' => t('My date'),
    '#description' => t('Enter the date.'),
    '#required' => TRUE,
  );
```

Leave the field blank and submit it, which should create a form error for the missing `#required` value.

## Actual behavior

The error message is "A valid date is required for .".

## Expected behavior

The error message should be "A valid date is required for _My date_.".

## Additional information

The error message is generated in function `date_popup_validate()`, via this bit of code:

```
    if ($element['#required']) {
      $message = t('A valid date is required for %title.', array('%title' => $label));
      form_set_error($error_field, $message);
      return;
    }
```

The variable `$label` is set via this code earlier in the function:

```
  $label = '';
  if (!empty($element['#date_title'])) {
    $label = t($element['#date_title']);
  }
  elseif (!empty($element['#title'])) {
    $label = t($element['#title']);
  }
```

So in this example, `$label` is set from the `#title` property. However, by the time this code is reached, `#title` has already been set to `NULL`.

This is because `date_popup_element_process()` creates child `date` and/or `time` parts, sets their title(s) from `$element['#title']`, and then clears `$element['#title']` (lines 1013–1014 of date.elements.inc):

```
  // Remove the title from the overall element.
  $element['#title'] = NULL;
```

So it looks like either the element `#title` should not be cleared here so it remains available to the validation error, or the validation code should check for the child element title(s) when creating the validation error message.

bugfolder:
Fixes https://github.com/backdrop/backdrop-issues/issues/6274.


quicksketch:
Thanks! Merged into 1.x and 1.26.x.

PR #4563, opened 2023-11-11
Issue #6288: In user_register_form(), use the value of $admin instead of calling again user_access('administer users')

### LINKED ISSUE #6288: In user_register_form(), use the value of $admin instead of calling again user_access('administer users')
The first two lines of [`user_register_form()`](https://docs.backdropcms.org/api/backdrop/core%21modules%21user%21user.module/function/user_register_form/1) are the following ones.

```php
  global $user;
```

Later, that function uses the following code.

```php
  elseif (user_access('administer users')) {
    $path = 'admin/people';
  }
```

Instead of calling again `user_access('administer users')`, the code should check the value of `$admin`.

```php
  elseif ($admin) {
    $path = 'admin/people';
  }
```

That is exactly what the code some lines before that does.

```php
  if ($admin) {
    // Redirect back to page which initiated the create request;
    // usually admin/people/create.
    $form_state['redirect'] = $_GET['q'];
  }
```

avpaderno:
Fixes https://github.com/backdrop/backdrop-issues/issues/6288

### DIFF
--- a/core/modules/user/user.module
+++ b/core/modules/user/user.module
@@ -3231,7 +3231,7 @@ function user_register_form($form, &$form_state) {
   elseif (isset($_SERVER['HTTP_REFERER'])) {
     $path = $_SERVER['HTTP_REFERER'];
   }
-  elseif (user_access('administer users')) {
+  elseif ($admin) {
     $path = 'admin/people';
   }
   else {
--- a/core/modules/user/user.module
+++ b/core/modules/user/user.module
@@ -1625,7 +1625,8 @@ function user_menu_alter(&$items) {
  */
 function user_admin_bar_output_alter(&$content) {
   // Admin menu shows both the parent menu item and its children.
-  // Remove the duplicate child element so "Manage fields" doesn't show up twice.
+  // Remove the duplicate child element, so "Manage fields" doesn't show up
+  // twice.
   unset($content['menu']['menu']['admin/config']['admin/config/people']['admin/config/people/manage']['admin/config/people/manage/fields']);
 }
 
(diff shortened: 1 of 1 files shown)


core/modules/user/user.module
```diff
@@ -1625,7 +1625,8 @@ function user_menu_alter(&$items) {
  */
 function user_admin_bar_output_alter(&$content) {
   // Admin menu shows both the parent menu item and its children.
-  // Remove the duplicate child element so "Manage fields" doesn't show up twice.
+  // Remove the duplicate child element, so "Manage fields" doesn't show up
```
bugfolder:
```suggestion
  // Remove the duplicate child element so "Manage fields" doesn't show up
```
The added comma is ungrammatical.

avpaderno: I guess it depends from whether the last clause is considered independent or dependent. Probably it is a dependent clause as the clause in the following sentence.

> I went to the store so I could buy tomatoes.

avpaderno:
(The failing tests are upgrade tests which keep to randomly fail.)

bugfolder:
`testLayoutUpgrade` is notorious for randomly failing, unrelated to the PR. While it's nice to get all 12 tests to pass, if the failing tests are unrelated and due to the test gremlin, we could still proceed. So the issue just needs human testing (which should be easy) and code review (ditto, since the change is small and simple). I'll try to get time a little later to do this.

avpaderno:
Perfect! Now, it is just CSpell complaining about unknown words.

klonos (approved):

quicksketch:
Thanks @kiamlaluno! Merged into 1.x and 1.26.x.

PR #4567, opened 2023-11-14
Support simplified database array

### LINKED ISSUE #2231: [DX] Switch to a simplified array syntax for database connection information in settings.php
I just tried to install Backdrop 1.5 in a local AMPPS environment. During installation, after being on the language page, I got the message

`Fatal error: Class 'DatabaseTasks_' not found in {...}/core/includes/install.inc on line 1362`

and wasn't able to continue the installation.

Line 1362 concerns the database driver. Googling something like _drupal error database driver_ led me to a [comment on d.org regarding Backup and Migrate](https://www.drupal.org/node/1999010#comment-7841581) about a problem with special characters in the database password string. I had a look in my _settings.php_, noticed a forward slash in the password string, changed the database password in the relevant places and was able to continue the installation.

Are forward slashes (and possibly particular other special characters) in the database password string not supported by Backdrop for a certain technical reason? If so, we should document it somewhere. If not, can we fix that?

ADVOCATE: @yorkshire-pudding

yorkshire-pudding:
Fixes https://github.com/backdrop/backdrop-issues/issues/2231


settings.php
```diff
@@ -10,10 +10,32 @@
  * Most sites can configure their database by entering the connection string
  * below. If using master/slave databases or multiple connections, see the
  * advanced database documentation at
- * https://api.backdropcms.org/database-configuration
+ * https://docs.backdropcms.org/database-configuration
+ *
+ * If your server uses a different port for the database from the default of
+ * '3306', you can add the 'port' attribute to the array.
```
klonos:
```suggestion
 * '3306', you can add the 'port' key to the array.
```
Lets also mention which of these keys are optional, and what they default to if omitted.

core/includes/bootstrap.inc
```diff
@@ -854,21 +854,44 @@ function backdrop_settings_initialize() {
...
+    'database' => 'database_name',
+    'username' => 'user',
+    'password' => 'pass',
+    'host' => 'localhost',
+    'prefix' => '',
+  );
+
+  // If database settings have been changed from default add to the full
+  // databases array.
+  if (!empty($database) && $database !== $default_db_string && $database !== $default_db_array) {
```
klonos: `$database !== $default_db_array` expects the keys in the two arrays being compared to also be in the same order. I don't think that we should be that strict here - lets use `!=` instead.

I would also explicitly and separately check for `is_string($database)` and `is_array($database)`, which would allow to only define the `$default_db_string` and `$default_db_array` variables when needed (since they aren't being used anywhere else outside the conditional). In other words, instead of doing this:
```php
if (!empty($database) && $database !== $default_db_string && $database !== $default_db_array) {
  if (is_array($database)) {
    ...
  }
  else {
    ...
  }
}
```

I would prefer doing this:
```php
// Database connection settings have been provided in the form of a connection
// string.
if (!empty($database) && is_string($database)) {
  $default_db_string = ... ;
  if ($database !== $default_db_string) { ... }
}
// Database connection settings have been provided in the form of a simplified
// array.
if (!empty($database) && is_array($database)) {
  $default_db_array = ... ;
  if ($database != $default_db_array) { ... }
}
```

settings.php
yorkshire-pudding: Done - I think I changed it to that because of the Harry Potter association of port key

core/includes/bootstrap.inc
yorkshire-pudding: Done

settings.php
yorkshire-pudding: I was a bit confused by this:

> Lets also mention which of these keys are optional, and what they default to if omitted.

Please can you elaborate?

.gitignore
```diff
@@ -2,7 +2,7 @@
 .gitignore
 
 # Ignore configuration files that may contain sensitive information.
-/settings*.php
+# /settings*.php
```
klonos: Why this?

core/includes/bootstrap.inc
```diff
@@ -854,21 +854,49 @@ function backdrop_settings_initialize() {
...
+  // string.
+  if (!empty($database) && is_string($database)) {
+    $default_db_string = 'mysql://user:pass@localhost/database_name';
+    if ($database !== $default_db_string) {
+      // Convert simplified database settings string into the full databases
+      // array.
+      $database_parts = parse_url($database);
+      if (!$database_parts) {
+        trigger_error('The database setting could not be parsed. Please check the $database setting in settings.php.', E_USER_ERROR);
+      }
```
klonos, line 882: If `parse_url()` has failed earlier, it will return `false`. So there's no point trying to continue to the next conditional.

settings.php
```diff
@@ -10,10 +10,32 @@
  * Most sites can configure their database by entering the connection string
```
klonos: We aren't providing a connection string as an example any more, so perhaps change this to "entering the connection details" or "entering the connection configuration".

.gitignore
yorkshire-pudding: I had to do that in .gitignore to commit settings.php but I don't remember committing the change to .gitignore. It should be ignored by its own definition.

settings.php
```diff
@@ -10,10 +10,32 @@
...
  * below. If using master/slave databases or multiple connections, see the
```
klonos: I thought that we had removed all instances of these words from our codebase with https://github.com/backdrop/backdrop-issues/issues/4451. Can we please change these to `/replica` here?
```suggestion
 * below. If using primary/replica databases or multiple connections, see the
```

core/includes/install.inc
```diff
@@ -722,6 +723,11 @@ function backdrop_rewrite_settings($settings = array()) {
         preg_match('/\$([^ =]*) /', $line, $variable);
         $variable_name = $variable[1];
         if (array_key_exists($variable_name, $settings)) {
+          // If variable_name is 'database', we need to remove the next 7
+          // lines as they are part of the default database array.
+          if ($variable_name == 'database') {
+            $database_line_count = 7;
```
klonos: This relies on the expectation that the database definition is left to the default format/lines:
```php
$database = array(
  'driver' => 'mysql',
  'database' => 'database_name',
  'username' => 'user',
  'password' => 'pass',
  'host' => 'localhost',
  'prefix' => '',
);
```

What if someone removes the prefix line, or merges the `);` with the previous line (`'prefix' => 'blah');`), or formats the entire array to be in a single line? I believe that we need to use the `m` (multiline) operator in `preg_match()` and adjust the regex accordingly to be checking several lines until it explicitly reaches a `);`.

Edit: and to expect the unexpected, someone might even use the abbreviated array format (`[]` instead of `array()`).

yorkshire-pudding: @klonos - as the undisputed king of regex, please could I ask you to make a code suggestion that I can accept here?

core/includes/bootstrap.inc
yorkshire-pudding: This is not added by me - added in backdrop/backdrop-issues/issues/2914

settings.php
yorkshire-pudding: Done

settings.php
yorkshire-pudding: Changed to primary/secondary as that is the terminology used in the docs page.

settings.php
yorkshire-pudding: Discussion about optional keys moved to issue.

yorkshire-pudding:
trying to rebase first before making changes

core/includes/install.inc
yorkshire-pudding: Resolving this as per discussions at Backdrop developer meetings

core/includes/bootstrap.inc
```diff
@@ -860,21 +860,67 @@ function backdrop_settings_initialize() {
...
+        $database['port'] = NULL;
+      }
+      // Add default charset if not set in the array.
+      if (!isset($database['charset'])) {
+        $database['charset'] = 'utf8mb4';
+      }
+      // Add default collation if not set in the array.
+      if (!isset($database['collation'])) {
+        $database['collation'] = 'utf8mb4_general_ci';
+      }
```
quicksketch: This bit of code can be shortened down to the equivalent:

```suggestion
      // Specify default values.
      $database += array(
        'prefix' => '',
        'driver' => 'mysql',
        'port' => NULL,
        'charset' => 'utf8mb4',
        'collation' = 'utf8mb4_general_ci',
      );
```

core/includes/install.core.inc
```diff
@@ -1071,23 +1071,29 @@ function install_settings_form_submit($form, &$form_state) {
...
-  );
-  backdrop_rewrite_settings($settings);
-
-  // Check if UTF-8 MB4 support exists in the database and add to settings.php.
-  if (install_verify_database_utf8mb4()) {
-    $settings = array();
-    $settings['database_charset'] = array(
-      'value' => 'utf8mb4',
-      'required' => TRUE,
-    );
```
quicksketch, line 1088: We should keep this check but instead set `$db_settings['charset']` to `utf8` if `install_verify_database_utf8mb4()` fails. Something like this:

```
// If UTF-8 MB4 is not supported, use the old UTF-8 charset.
if (install_verify_database_utf8mb4() === FALSE && $db_settings['charset'] === 'utf8mb4') {
  $db_settings['charset'] = 'utf8';
}
```

core/includes/install.inc
```diff
@@ -808,7 +853,8 @@ function backdrop_install_config_location() {
 
   if ($settings['config_active_class'] === 'ConfigFileStorage') {
     $is_empty = empty($config_directories['active']);
-    $is_default_file_storage = $config_directories['active'] === 'files/config_' . md5($database) . '/active';
+    $database_string = is_array($database) ? implode(',', $database) : $database;
```
quicksketch: Let's use `serialize()` here as a sure-fire way to convert an array into a string, as there could conceivably be nested array values at some point.

```suggestion
    $database_string = is_array($database) ? serialize($database) : $database;
```

settings.php
```diff
@@ -52,8 +59,8 @@
  * $config_directories['staging'] = '/home/myusername/config/staging';
  * @endcode
  */
-$config_directories['active'] = 'files/config_' . md5($database) . '/active';
-$config_directories['staging'] = 'files/config_' . md5($database) . '/staging';
+$config_directories['active'] = 'files/config_' . md5(implode(',', $database)) . '/active';
+$config_directories['staging'] = 'files/config_' . md5(implode(',', $database)) . '/staging';
```
quicksketch: Same as before, `serialize()` is probably a more robust way to convert an array to a string:

```suggestion
$config_directories['active'] = 'files/config_' . md5(serialize($database)) . '/active';
$config_directories['staging'] = 'files/config_' . md5(serialize($database)) . '/staging';
```

core/includes/install.inc
```diff
@@ -725,6 +758,11 @@ function backdrop_rewrite_settings($settings = array()) {
         preg_match('/\$([^ =]*) /', $line, $variable);
         $variable_name = $variable[1];
         if (array_key_exists($variable_name, $settings)) {
+          // If variable_name is 'database', we need to remove the next 5
+          // lines as they are part of the default database array.
+          if ($variable_name == 'database') {
+            $database_line_count = 5;
+          }
```
quicksketch: While not ideal, this database-specific rewriting code is fine to me. We can separate out improving this rewriting into a separate issue if `backdrop_rewrite_settings()` needs to be improved. But since it specifically only works with the default values of settings.php, we can accept this assumption that the database variable is 6 lines long.

core/includes/bootstrap.inc
yorkshire-pudding: Done.  Thanks @quicksketch - that is a new bit of PHP array syntax I didn't know before.

core/includes/install.inc
yorkshire-pudding: Thanks @quicksketch .  I have used this change to simplify as we can simply `serialize` the `$database` variable regardless of whether it is a string or an array, so I can drop the `$database_string` step.

core/includes/install.inc
yorkshire-pudding: Yes, I couldn't think of a better way to do this, given the way the function works.

core/includes/install.core.inc
yorkshire-pudding: Done.  Thanks.

core/includes/install.inc
```diff
@@ -725,10 +746,16 @@ function backdrop_rewrite_settings($settings = array()) {
...
+          // lines as they are part of the default database array.
+          // @todo Improve this logic to support replacing any size array.
+          if ($variable_name == 'database') {
+            $database_line_count = 5;
+          }
           // Write new value to settings.php in the following format:
           //    $'setting' = 'value'; // 'comment'
           $setting = $settings[$variable_name];
-          $new_line = '$' . $variable_name . " = " . var_export($setting['value'], TRUE) . ";" . (!empty($setting['comment']) ? ' // ' . $setting['comment'] . "\n" : "\n");
+          $new_line = '$' . $variable_name . " = " . backdrop_var_export($setting['value']) . ";" . (!empty($setting['comment']) ? ' // ' . $setting['comment'] . "\n" : "\n");
```
quicksketch, line 758: I changed this line: Using `backdrop_var_export()` does a better job of formatting an array than `var_export()`.

core/includes/install.inc
```diff
@@ -454,31 +456,29 @@ abstract class DatabaseTasks {
...
-          // Connection was successful, but settings.php needs to be updated.
-          $db_settings = $modified_connection_info['default'];
-          $settings['database'] = array(
-            'value' => $db_settings['driver'] . '://' . rawurlencode($db_settings['username']) . ':' . rawurlencode($db_settings['password']) . '@' . $db_settings['host'] . ($db_settings['port'] ? (':' . $db_settings['port']) : '') . '/' . rawurlencode($db_settings['database']),
-            'required' => TRUE,
-          );
-          $settings['database_prefix'] = array(
-            'value' => $db_settings['prefix']['default'],
-          );
-          backdrop_rewrite_settings($settings);
```
quicksketch, line 481: I modified this section because `backdrop_rewrite_settings()` was getting called twice on the `$database` settings, causing the value to become invalid PHP. Instead of rewriting automatically in this unexpected place, now a better message is shown instead.

core/includes/install.inc
```diff
@@ -454,53 +456,29 @@ abstract class DatabaseTasks {
...
-          if ($db_settings['collation']) {
-            $settings['database']['value']['collation'] = $db_settings['collation'];
-          }
-          backdrop_rewrite_settings($settings);
+          Database::getConnection('default', 'default');
+
+          // The modified database connection was successful. Provide an
+          // informative help message asking the installer to adjust the
+          // values.
+          $extra_help = st('Connecting to "@broken_host" was not successful, but connecting to "@working_host" with the same credentials was successful. Try setting the database connection host to "@working_host".', array(
```
olafgrabienski: The database connection setting isn't visible at this point. I suggest to mention that it is in the (collapsed) "Advanced options".

core/includes/install.inc
yorkshire-pudding: Makes sense. Thanks

core/includes/install.inc
yorkshire-pudding: I've clarified this in the text.

PR #4569, opened 2023-11-15
Issue #214: Remove old IE support: html5shiv and conditional comments.

### LINKED ISSUE #214: Remove IE-specific code/libraries
The current state of our codebase does have support for IE8.  Drupal 8 has dropped it.

Facts worth considering:
- The combined, global usage of IE versions 5 through 10 is less than ~0.4% (caniuse.com/usage-table). Actually all IE versions + all Edge versions seem to be around ~7% (w3counter.com/globalstats.php). ...although to be fair, it seems that usage is more around 10%-11% if you limit the scope to desktop browsers alone (netmarketshare.com/browser-market-share.aspx). Roughly speaking, everyone seems to be using either Chrome, or Safari these days (http://gs.statcounter.com).

- Google supports only IE11 and Edge (https://support.google.com/drive/answer/2375082?hl=en)

What do you think?

quicksketch:
Fixes https://github.com/backdrop/backdrop-issues/issues/214


quicksketch:
Replaced with the smaller scope https://github.com/backdrop/backdrop/pull/4611.

PR #4573, opened 2023-11-20
Issue #3419: Follow-up: Use #indentation in the dialog

### LINKED ISSUE #3419: [UX][D8] Allow 'Content: Link' views field to just output URL

klonos:
Fixes https://github.com/backdrop/backdrop-issues/issues/3419

Follow-up to https://github.com/backdrop/backdrop/pull/3223

### DIFF
--- a/core/modules/node/views/views_handler_field_node_link.inc
+++ b/core/modules/node/views/views_handler_field_node_link.inc
@@ -30,6 +30,7 @@ class views_handler_field_node_link extends views_handler_field_entity {
       '#type' => 'textfield',
       '#title' => t('Text to display'),
       '#default_value' => $this->options['text'],
+      '#indentation' => 1,
       '#states' => array(
         'visible' => array(
           ':input[name="options[url]"]' => array('checked' => FALSE),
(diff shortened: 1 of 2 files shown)


klonos:
...the usual `LayoutUpgradePathTest` test gremlin 😞

quicksketch:
Thanks @klonos! Merged into 1.x only as this has conflicts in the 1.26.x branch (since https://github.com/backdrop/backdrop-issues/issues/3419 is a 1.27.0 only feature).

PR #4580, opened 2023-11-25
Issue #6302: CSpell comments trigger PHPCS

klonos:
Fixes https://github.com/backdrop/backdrop-issues/issues/6302

image


quicksketch:
@klonos I don't think we should pursue this approach. Instead we should prevent phpcs from complaining about these lines, as has been suggested in https://github.com/backdrop-ops/phpcs/issues/16.

PR #4589, opened 2023-12-01
Issue #6303: Update the part in the documentation comment about array_filter

### LINKED ISSUE #6303: Update the part in the documentation comment about array_filter
[`system_settings_form_submit()`](https://docs.backdropcms.org/api/backdrop/core%21modules%21system%21system.module/function/system_settings_form_submit/1) has the following documentation comment.

```php
/**
 * Form submission handler for system_settings_form().
 *
 * If you want node type configure style handling of your checkboxes,
 * add an array_filter value to your form.
 *
 * @since 1.6.0 Restored and made compatible with configuration management.
 */
```

The part about the _array_filter_ value is true only for forms that do not define `$form['#config']` and when the Drupal compatibility layer is enabled, since that function uses the following code. (The comments in the code are mine.)

```php
  if (!empty($form['#config'])) {
    $configs = _system_sort_form_values_by_config($form, $form_state, $form['#config']);
    foreach ($configs as $bucket => $data) {
      config_set_multiple($bucket, $data);
    }
  }
  else {
    watchdog('form', 'Backdrop 2.0 and later will require forms using system_settings_form() to set $form[\'#config\'].', array(), WATCHDOG_DEPRECATED);
    if (!settings_get('backdrop_drupal_compatibility')) {
      backdrop_set_message(t('Settings not saved: Drupal compatibility layer is disabled and this form has not been updated to use Backdrop configuration management.'), 'error');
      return;
    }

    // The form does not set a value for $form['#config'] and the Drupal compatibility layer
    // is enabled.
    foreach ($form_state['values'] as $key => $value) {
      if (is_array($value) && isset($form_state['values']['array_filter'])) {
        $value = array_keys(array_filter($value));
      }
      variable_set($key, $value);
    }
  }
```

Either that documentation comment is changed to make clear `'array_filter'` is used when the Drupal compatibility layer is enabled, and the form does not set `$form['#config']`, or the code is changed to handle `'array_filter'` in all the cases.

IMO, `'array_filter'` should only be used for BC with Drupal 7. I would rather change the documentation comment to reflect that.

avpaderno:
Fixes https://github.com/backdrop/backdrop-issues/issues/6303


core/modules/system/system.module
```diff
@@ -4667,6 +4664,8 @@ function system_settings_form_submit($form, &$form_state) {
       return;
     }
 
+    // For compatibility with Drupal 7, if $form['array_filter'] is set,
+    // pass the form state values to array_filter().
```
yorkshire-pudding: Proposing a bit of extra explanation to make it clear what will happen as no callback is passed to the `array_filter()` function.

```suggestion
    // For compatibility with Drupal 7, if $form['array_filter'] is set, pass
    // the form state values to array_filter(). This will remove any empty
    // values.
```

yorkshire-pudding (changes requested):
I found this as was trying to understand the meaning of this cryptic comment in the function docs. I've proposed a small explanation in the code comment.

yorkshire-pudding (approved):
Looks good

quicksketch:
Looks great, thanks @avpaderno! I am sorry this simple docs update took so long. Thank you @yorkshire-pudding for the review! You made it happen. Merged into 1.x and 1.31.x.

PR #4595, opened 2023-12-06
Issue #4672: Add random tokens

### LINKED ISSUE #4672: Random tokens are missing
**Description of the bug**

Drupal 7 (with the Token module installed) provides "Random" tokens like `[random:number]`. Backdrop has integrated tokens but without the Random ones. They should be added to Backdrop, in my opinion.

The Drupal implementation of Random tokens
- https://git.drupalcode.org/project/token/-/blob/7.x-1.x/token.tokens.inc#L346
- https://git.drupalcode.org/project/token/-/blob/7.x-1.x/token.tokens.inc#L877

indigoxela:
Fixes https://github.com/backdrop/backdrop-issues/issues/4672

### DIFF
--- a/core/modules/system/system.tokens.inc
+++ b/core/modules/system/system.tokens.inc
@@ -41,6 +41,10 @@ function system_token_info() {
     'description' => t('Tokens related to arrays of strings.'),
     'needs-data' => 'array',
   );
+  $types['random'] = array(
+    'name' => t('Random'),
+    'description' => ('Tokens related to random data.'),
+  );
 
   // Site-wide global tokens.
   $site['name'] = array(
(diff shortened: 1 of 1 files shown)


nattywebdev:
I was able to add a the token for a random hash to the url alias pattern for content type post, appending correctly the token to the node title in the path using:

- posts/[node:title]-[random:hash:sha1]

and here are the urls for two resulting posts:

- posts/a-test-post-ddeb517750291172719bf45d0db2e483e4c0e072
- posts/another-test-post-2ea0528c1c7db6756feeff6452cdac51c20358b1

Second test was to use instead posts/[node:title]-[random:number] and this led to two test posts:

- posts/third-test-1789595986
- posts/fourth-test-using-random-115121789

This simple test was therefore successful.

olafgrabienski:
@nattywebdev I was also curious and wanted to test the sandbox of this PR. Strange, but I didn't find the random tokens in the token browser when I was on the configuration page of a content type. Maybe I just missed them! Have you used the token browser and can give me a hint where to look?

nattywebdev:
@olafgrabienski I only checked in the URL Alias Pattern page (and just checked again) where I could see after clicking on  Browse available tokens for Content:

image


BUT you're right - when you click on Browse available tokens when editing a content type, I see a different list:
image

olafgrabienski:
Thanks, I'll add a comment to the issue then.

quicksketch:
Merged into 1.x and 1.26.x. Thanks @indigoxela, @olafgrabienski, and @nattywebdev!

PR #4597, opened 2023-12-08
Issue #6327: Handle trimmed content consistently across libxml versions

### LINKED ISSUE #6327: Fixed: `<` gets encoded to `&lt;` by libxml >v2.9.14, causing broken HTML in generated summaries.
## Description of the bug

In a [previous issue](https://github.com/backdrop/backdrop-issues/issues/6284) it has been discovered, that the behavior of libxml changed between v2.9.10 and v2.9.14, which results in failing TextSummaryTestCase tests with a current libxml version.

This doesn't show up in our automated test runs, as the runner OS is Ubuntu 20.04, which by default uses 2.9.10.

However, in many environments the version is already 2.9.14, and it will also change when we switch our runner to use a newer Ubuntu.

Note that the test rightly fails, as there's some cruft in the summary output, the opening `<` gets encoded to `&lt;`, which shows up in content.

## Steps To Reproduce

1. Make sure, your libxml version is new enough
2. Configure the summary to be shortened to 10 characters
3. Create a node with this content: `<p>456789</p>`
4. Visit the summary (wherever you configured it)

Or, if the devel module is installed, run following code:
```
$text = '<p>456789</p>';
$result = text_summary($text, NULL, 10);
debug($result);
```

## Actual behavior

You'll see `<p>456789&lt;</p>` (raw) or `456789<` as rendered markup.

Note the `&lt;`.

## Expected behavior

As in previous libxml versions, this opening `<` should get trimmed.

## Additional information

The easiest solution to let the summary behave like in previous versions seems to be to use code like that:
```
$summary = rtrim($summary, '<');
```
In function text_summary(), [right before  _filter_htmlcorrector()](https://github.com/backdrop/backdrop/blob/1.x/core/modules/field/modules/text/text.module#L496). Then the result would be the same, regardless of the libxml version.

indigoxela:
Fixes https://github.com/backdrop/backdrop-issues/issues/6327

This also changes the runner to ubuntu-22.04 (newer libxml), which required a permission fix.

### DIFF
--- a/.github/misc/default.conf
+++ b/.github/misc/default.conf
@@ -3,9 +3,9 @@
   DocumentRoot /home/runner/work/backdrop/backdrop
 
   <Directory /home/runner/work/backdrop/backdrop>
+    Options FollowSymLinks
     AllowOverride All
     Require all granted
-    Options FollowSymLinks
   </Directory>
 
   # Placeholder _PHP_VERSION_  will be substituted in step "Setup webserver".
--- a/.github/misc/default.conf
+++ b/.github/misc/default.conf
@@ -9,9 +9,9 @@
   </Directory>
 
   # Placeholder _PHP_VERSION_  will be substituted in step "Setup webserver".
-  <FilesMatch ".+\.php$">
-    <If "-f %{REQUEST_FILENAME}">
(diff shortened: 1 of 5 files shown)


quicksketch:
Thanks @indigoxela! I merged this into 1.x and 1.27.x.

PR #4599, opened 2023-12-09
Issue #2516 - Keep Cancel Button Visible

### LINKED ISSUE #2516: [UX] Add a cancel link in the views UI
When I edit a view, then decide not to make any changes, I always scroll around for a while looking for a cancel link or button. But then realize there is not one, and click "Back" in my browser.  I think we could improve the user experience in Backdrop by adding this missing Cancel link.

stpaultim:
Fixes https://github.com/backdrop/backdrop-issues/issues/2516


core/modules/views/js/ajax.js
```diff
@@ -25,7 +25,7 @@
   };
 
   Backdrop.ajax.prototype.commands.viewsShowButtons = function (ajax, response, status) {
-    $('div.views-edit-view div.form-actions').removeClass('js-hide');
+    $('div.views-edit-view div.form-actions input.button-primary').removeClass('js-hide');
```
argiepiano: I find it a bit risky to rely on a css class added by a third-party function (in this case `form_process_actions()`) to unhide the Save button.  This class is used in a lot of other contexts for styling, and it's possible (thought unlikely) that this class name may change at some point. 

It would probably be less risky to modify the form builder `views_ui_edit_form()` and add a custom css class to the Save submit button. Either that, or to rely on the input element ID.

argiepiano: For example:

```js
$('div.views-edit-view div.form-actions input#edit-actions-save').removeClass('js-hide');
```

stpaultim: I'm good with this change. It works fine in my testing. We'll see what others think.

core/modules/views_ui/views_ui.admin.inc
```diff
@@ -2149,6 +2142,8 @@ function views_ui_edit_view_form_submit($form, &$form_state) {
 function views_ui_edit_view_form_cancel($form, &$form_state) {
   // Remove this view from cache so configurations will be lost.
   tempstore_clear('views.view', $form_state['view']->name);
+  backdrop_set_message(t('Changes to the view %name have been discarded.', array('%name' => $form_state['view']->get_human_name())), 'info');
+
```
yorkshire-pudding: PHPCS - remove whitespace
```suggestion

```

yorkshire-pudding (changes requested):
Very minor coding standards

stpaultim:
Also noticed that the old PR was hiding this message, by deleting line 28 in core/modules/views/js/ajax.js . By restoring that line into the PR, it seems to be working as expected.

image

quicksketch:
Thanks @stpaultim! I reviewed this and it looks great! Merged into 1.x and 1.27.x.

PR #4606, opened 2023-12-16
Issue #2277: Add swappable config storage and database storage.

### LINKED ISSUE #2277: [D8][PS] Allow configuration storage in alternative locations (for example: the database).

quicksketch:

Replaces https://github.com/backdrop/backdrop/pull/4453

Based on earlier work by @jlfranklin and @laryn 

I had to file an entirely new PR as the previous one had dozens of conflicts when rebasing.


quicksketch:
Let's work out of https://github.com/backdrop/backdrop/pull/4608 (which runs the database-based config tests) exclusively to reduce confusion.

PR #4607, opened 2023-12-16
Issue #6333: Use #empty for the message to display when a table does not have any row

### LINKED ISSUE #6333: Use #empty for the message to display when a table does not have any row
In some cases, the code that renders a table in a page checks if the table does not have rows to then show a message when the table is empty.

### core/modules/dashboard/includes/block.node_types.inc

```php
    if (empty($rows)) {
      $rows[] = array(
        array(
          'data' => t('There are no content types to display.'),
          'colspan' => 4,
        ),
      );
    }
```

### core/modules/image/image.theme.inc

```php
if (empty($rows)) {
    $rows[] = array(array(
      'colspan' => 4,
      'data' => t('There are currently no styles. <a href="!url">Add a new one</a>.', array('!url' => url('admin/config/media/image-styles/add'))),
    ));
  }
```

### core/modules/taxonomy/taxonomy.theme.inc

```php
  if (empty($rows)) {
    $rows[] = array(array('data' => $form['#empty_text'], 'colspan' => '3'));
  }
```

### core/modules/views_ui/views_ui.theme.inc

```php
  if (empty($rows)) {
    $rows[] = array(array('data' => t('No fields available.'), 'colspan' => '2'));
  }
```

```php
  if (empty($rows)) {
    $rows[] = array(array('data' => t('No fields available.'), 'colspan' => '2'));
  }
```

In all those cases, an `'#empty'` value should be passed to `theme('table')`.

I did not include core/modules/dashboard/includes/block.menus.inc because in that case the code returns before rendering the table, when `$menus` is not empty.

```php
    if (empty($rows)) {
      // If there are existing menus, but user has no access to manage any of
      // them, hide the block completely.
      if (!empty($menus)) {
        return;
      }
      $rows[] = array(
        array(
          'data' => t('There are no menus to display.'),
          'colspan' => 4,
        ),
      );
    }
```

avpaderno:
Fixes https://github.com/backdrop/backdrop-issues/issues/6333

### DIFF
--- a/core/modules/comment/comment.theme.inc
+++ b/core/modules/comment/comment.theme.inc
@@ -104,7 +104,7 @@ function template_preprocess_comment(&$variables) {
 
   $uri = $comment->uri();
   $uri['options'] += array('attributes' => array('class' => array('permalink'), 'rel' => 'bookmark'));
-  
+
   $variables['permalink'] = l(t('Permalink'), $uri['path'], $uri['options']);
   $variables['permalink_path'] = $uri['path'] . "#" . $uri['options']['fragment'];
   $variables['title_display'] = $node_type->settings['comment_title_options'] != COMMENT_TITLE_HIDDEN;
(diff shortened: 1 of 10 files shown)


avpaderno:
(I definitively need to use a bigger editor font.)

PR #4608, opened 2023-12-18
Issue #2277: Add Database-stored Configuration (with tests)

### LINKED ISSUE #2277: [D8][PS] Allow configuration storage in alternative locations (for example: the database).

quicksketch:

Adds test coverage to the existing PR. This is ready for review.


core/includes/config.inc
```diff
@@ -357,8 +357,16 @@ function config_uninstall_config($project, $config_name = NULL) {
  *   A ConfigStorageInterface object managing the specified configuration type.
  */
 function config_get_config_storage($type = 'active') {
-  $directory = config_get_config_directory($type);
-  return new ConfigFileStorage($directory);
+  $class = settings_get('config_' . $type . '_class', 'ConfigFileStorage');
+  switch ($class) {
+    case 'ConfigDatabaseStorage':
+      $config_location = 'db:/default/' . $type . '_config';
```
laryn: @quicksketch Should this be:
`$config_location = 'db:/default/config_' . $type;`?

laryn: Or starting off with `db://` as discussed, but the ending of this line is what I am referencing here.

quicksketch: Thanks @laryn, I've made both changes!

core/modules/simpletest/tests/menu.test
```diff
@@ -1496,6 +1496,9 @@ class MenuTrailTestCase extends MenuWebTestCase {
     // the breadcrumbs and Management tree.
     state_set('menu_test_menu_tree_set_path', $test_menu_path);
     $this->assertBreadcrumb('admin/config/development/menu-trail', $override_breadcrumb, t('Menu trail - Case 2'), $override_tree);
+
+    // Load one more page to clear out any pending theme registry saves.
+    $this->backdropGet('admin/config/system');
```
quicksketch: This is the weirdest failure I ran into with this PR. If there are pending saves to the theme registry, the `BackdropWebTestCase::setUp()` method tries to clear the theme registry but a new theme function tries to get written to cache while switching between config directories. This seems to be the only place in all of core where this particular problem occurred.

.github/actions/set-up-webserver/action.yml
```diff
@@ -0,0 +1,62 @@
...
+  database-version:
+    description: "Database (mysql/mariadb) version to install"
+    required: true
+outputs: {}
+
+runs:
+  using: "composite"
+  steps:
+    - name: Prepare for database
+      shell: bash
```
indigoxela, line 16: This seems redundant (also below), as "bash" is the default.
https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#defaultsrunshell

.github/actions/set-up-webserver/action.yml
```diff
@@ -0,0 +1,62 @@
...
+
+    - name: Finish webserver setup
+      shell: bash
+      run: |
+        # Set up apache with php-fpm.
+        sudo cp .github/misc/default.conf /etc/apache2/sites-available/000-default.conf
+        sudo sed -i -e "s/_PHP_VERSION_/${{ inputs.php-version }}/" /etc/apache2/sites-available/000-default.conf
+        # Setup php-fpm, this script is intentionally not executable by default.
+        chmod 700 .github/misc/setup-php-fpm.sh
+        ./.github/misc/setup-php-fpm.sh ${{ inputs.php-version }}
```
indigoxela, line 59: <s>Seems like this (unintentionally) slipped on a new line.</s>

Sorry for the noise, totally misinterpreted this code. :see_no_evil:  Consider this as resolved.

.github/workflows/simpletest-db-config.yml
```diff
@@ -0,0 +1,41 @@
...
+    name: SimpleTest
+    runs-on: ubuntu-22.04
+    # Skip this job if the PR title contains "[skip tests]".
+    if: ${{ !contains(github.event.pull_request.title, '[skip tests]') }}
+    timeout-minutes: 20
+    strategy:
+      fail-fast: false
+      matrix:
+        php-versions: ['8.3']
+        database-versions: ['mariadb-10.3']
```
indigoxela, line 16: I wonder, if it's time to switch to a newer mariadb version, as 10.3 is quite dated. But we could discuss that also in a follow-up (actually unrelated here).

.github/workflows/simpletest-db-config.yml
```diff
@@ -0,0 +1,41 @@
...
+  simpletest-db-config:
+    name: SimpleTest
+    runs-on: ubuntu-22.04
+    # Skip this job if the PR title contains "[skip tests]".
+    if: ${{ !contains(github.event.pull_request.title, '[skip tests]') }}
+    timeout-minutes: 20
+    strategy:
+      fail-fast: false
+      matrix:
+        php-versions: ['8.3']
```
indigoxela, line 15: I guess, that's temporary (single php version)...?

.github/workflows/simpletest-db-config.yml
```diff
@@ -0,0 +1,41 @@
...
+          php-version: ${{ matrix.php-versions }}
+          database-version: ${{ matrix.database-versions }}
+
+      - name: Install Backdrop
+        run: |
+          cp ./.github/misc/settings.local.php .
+
+          # Set the config storage class to use the database.
+          echo "\$settings['config_active_class'] = 'ConfigDatabaseStorage';" >> settings.local.php
+          echo "\$config_directories['active'] = NULL;" >> settings.local.php
```
indigoxela, line 34: Is that line actually needed? I didn't set that at all in my local test.

Ah, got it, it's set by default in the local settings file. So, maybe set the values dynamically in that file? Might be slightly less confusing. (Or maybe not) :wink:

.github/workflows/simpletest-db-config.yml
quicksketch: For the time being I was thinking we'd limit it to a single PHP version.

.github/actions/set-up-webserver/action.yml
quicksketch: It seems the rules might be different for an action file vs a workflow file. Without specifying the shell, I get a schema validation error: `One of the property sets is required: property 'shell' or property 'uses'`

So in this situation, it seems it's required.

.github/workflows/simpletest-db-config.yml
quicksketch: Yeah I don't know if it's more or less confusing. I sort of like having it all in this file rather than the settings.local.php containing additional if statements.

.github/actions/set-up-webserver/action.yml
indigoxela: Ah, wasn't aware of that difference. Many thanks for clarifying.

Resolved. :+1:

core/modules/simpletest/tests/menu.test
quicksketch: I was able to remove this and replace it with a better solution for all tests in `BackdropWebTestCase::tearDown()`.

PR #4609, opened 2023-12-18
Fixes #6337 broken compress css

### LINKED ISSUE #6337: Semicolons in CSS import statements break compressed CSS files
## Description of the bug

The compressed css is broken when the original style sheet contains an import rule with a `;` within the url, like this:

`@import url("https://fonts.googleapis.com/css2?family=Playfair+Display:ital,wght@0,400;0,500;1,400;1,500&display=swap");`

## Steps To Reproduce

To reproduce the behavior:
1. add that import rule to a css style sheet
2. go to admin/config/development/performance and check on Aggregate and compress css files
3. go to a page

## Actual behavior

The site is broken because the compressed css file is broken

robertgarrigos:
Fixes https://github.com/backdrop/backdrop-issues/issues/6337


quicksketch:
Merged into 1.x and 1.26.x.

robertgarrigos:
The Tugboat failed, but the details link is not working, for me at least.

PR #4614, opened 2023-12-21
Issue #6342: Add CKEditor module version telemetry.

### LINKED ISSUE #6342: Better reporting for CKEditor 4 vs 5 in Telemetry
## Description of the need

Follow-up from https://github.com/backdrop/backdrop-issues/issues/4122.

Right now both CKEditor 4 and CKEditor 5 modules implement `hook_telemetry_info()` to report their respective CKEditor version strings. For example:

CKEditor 4 returns `ckeditor_version` of `4.18.0`
CKEditor 5 returns `ckeditor5_version` of `40.2.0`

While this data is useful in some ways to tell how up-to-date our library usage is, it's not helpful for distinguishing between CKEditor 4 and 5 (or both) usage. To do that we need a separate return value of either `4`, `5`, or `both`.

## Proposed solution

Since each module can only report if it is enabled, I think it would make sense to include this value in telemetry.module directly.

quicksketch:
Fixes https://github.com/backdrop/backdrop-issues/issues/6342

### DIFF
--- a/core/modules/telemetry/telemetry.telemetry.inc
+++ b/core/modules/telemetry/telemetry.telemetry.inc
@@ -40,7 +40,7 @@ function telemetry_telemetry_info() {
   );
   $info['ckeditor_module_version'] = array(
     'label' => t('CKEditor Module Version'),
-    'description' => t('The major version of CKEditor module that your site uses (either 4, 5, or both).'),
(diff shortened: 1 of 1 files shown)


quicksketch:
Merged into 1.x for 1.27.0.

PR #4618, opened 2023-12-26
Issue 6347. Avoid creating a CKE5 editor if it already exists in the page.

### LINKED ISSUE #6347: When Body uses CKE5, additional Body entry fields are added to the form when clicking "Add another" for a different field
## Description of the bug

If your content type has a field (such as a Text (short) field) with unlimited cardinality, clicking "Add another" will add a new field to both, the Body field (!!) and the unlimited Text (short) field.  

## Steps To Reproduce

1. On a clean install with the dev version of Backdrop (so that you have CKE5 as default in core), edit the Page content type and add a Text (short) field with unlimited cardinality, call it "My text"
2. Visit the add content form for Page
3. Click "Add another" for the "My text" field

## Actual behavior

Additional widgets are added to both, the "My text" field and the "Body" field. 

## Expected behavior

An additional widget should be added to "My text" field only. 

## Additional information

- Backdrop CMS version:  current dev version slated to be released as 1.27.0
- PHP version: 8.1.13

argiepiano:
Fixes https://github.com/backdrop/backdrop-issues/issues/6347

### DIFF
--- a/core/modules/ckeditor5/js/ckeditor5.js
+++ b/core/modules/ckeditor5/js/ckeditor5.js
@@ -5,8 +5,7 @@
   Backdrop.editors.ckeditor5 = {
 
     attach: function (element, format) {
-      // Bail out if the editor already exists in the page; for example, when
-      // this attach method is called after an ajax request.
+      // Bail out if the editor has already been attached to the element.
       if (typeof element.ckeditor5AttachedEditor != 'undefined') {
         return;
       }
--- a/core/modules/ckeditor5/js/ckeditor5.js
+++ b/core/modules/ckeditor5/js/ckeditor5.js
@@ -5,6 +5,12 @@
   Backdrop.editors.ckeditor5 = {
 
     attach: function (element, format) {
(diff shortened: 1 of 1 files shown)


core/modules/ckeditor5/js/ckeditor5.js
```diff
@@ -5,6 +5,12 @@
   Backdrop.editors.ckeditor5 = {
 
     attach: function (element, format) {
+      // Bail out if the editor already exists in the page; for example, when
+      // this attach method is called after an ajax request.
```
indigoxela: This comment seems a bit misleading...

"if the editor already exists in the page" - there can be multiple editors on a page. It's rather the current element.

@argiepiano what do you think?

argiepiano: @indigoxela, thanks. I changed it. LMK if this looks ok.

quicksketch:
Thanks @argiepiano!

PR #4620, opened 2023-12-26
Issue #4249: Date Fields help text

### LINKED ISSUE #4249: [UX] Fixed: Add better help text for Date fields.

stpaultim:
Fixes https://github.com/backdrop/backdrop-issues/issues/4249


stpaultim:
I removed the field label settings from Text Field Widget, because they are not being used there. 

Currently, if you have text field widget selected, you get these options (hidden or above), but changing these options does nothing:

image

PR currently only shows options for Select Widget, which is (I think) the only time they are relevent.

image

core/modules/date/date.admin.inc
```diff
@@ -627,7 +623,7 @@ function _date_field_settings_form($field, $instance, $has_data) {
   $form['cache_enabled'] = array(
     '#type' => 'checkbox',
     '#title' => t('Cache dates'),
-    '#description' => t('Date objects can be created and cached as date fields are loaded rather than when they are displayed to improve performance.'),
+    '#description' => t('Create and cache date objects when date fields are being loaded, rather than when they are being displayed. This can help to improve performance.'),
```
avpaderno: I would rather say *when date fields are loaded* and *when they are displayed*. The Present Progressive tense does not add much in a sentence that uses *when*.

stpaultim: Fair enough. Change made.

PR #4623, opened 2023-12-29
Issue #4703: Change website feedback to contact form

### LINKED ISSUE #4703: Rename default Contact form category
**Description of the need**

I suspect that most people who enable the core Contact form do so in order to allow visitors to contact them, without needing to provide their email address publicly. As such, I also believe that most emails recieved through the contact form have nothing to do with feedback about the website itself (that seems like a very specific use-case).

**Proposed solution**

I therefore propose renaming the default Contact form category from "Website feedback" to something like "Website enquiry", or just "Enquiry". To me this seems more generic and a better default option.

Users can always just edit the default category themselves (I personally have to do that all the time).

**Additional information**

image

stpaultim:
Fixes https://github.com/backdrop/backdrop-issues/issues/4703


izmeez:
The tugboat preview does not appear to be working for me, but the fix is very simple and makes sense.

PR #4625, opened 2024-01-02
Issue #6352: Update 'cketoolbar' config key to just 'toolbar'.

### LINKED ISSUE #6352: Rename "cketoolbar" config key to just "toolbar"
Follow up to https://github.com/backdrop/backdrop-issues/issues/4122.

In the CKEditor Contrib module, we renamed the config key used by CKEditor 5 within text format settings from "cketoolbar" to just "toolbar" in issue https://github.com/backdrop-contrib/ckeditor5/issues/113. This was to work around a core bug that was fixed in Backdrop 1.26.3.

We should make this change in core as well now that we no longer have to work-around the core bug from https://github.com/backdrop/backdrop-issues/issues/6203.

quicksketch:
Fixes https://github.com/backdrop/backdrop-issues/issues/6352

### DIFF
--- a/core/modules/ckeditor5/ckeditor5.install
+++ b/core/modules/ckeditor5/ckeditor5.install
@@ -55,6 +55,7 @@ function ckeditor5_enable() {
  * Updates the editor settings key from "cketoolbar" to just "toolbar".
  */
 function ckeditor5_update_1000() {
+  // cspell:ignore cketoolbar
   $config_names = config_get_names_with_prefix('filter.format');
   foreach ($config_names as $config_name) {
     $config = config($config_name);
--- a/core/modules/ckeditor5/ckeditor5.install
+++ b/core/modules/ckeditor5/ckeditor5.install
@@ -59,13 +59,16 @@ function ckeditor5_update_1000() {
   $config_names = config_get_names_with_prefix('filter.format');
   foreach ($config_names as $config_name) {
     $config = config($config_name);
-    if ($config['editor'] === 'ckeditor5' && isset($config['editor_settings']['cketoolbar'])) {
-      $editor_settings = $config['editor_settings'];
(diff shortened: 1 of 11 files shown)


core/modules/ckeditor5/ckeditor5.install
```diff
@@ -31,3 +31,22 @@ function ckeditor5_requirements($phase) {
...
+function ckeditor5_update_1000() {
+  // cspell:ignore cketoolbar
+  $config_names = config_get_names_with_prefix('filter.format');
+  foreach ($config_names as $config_name) {
+    $config = config($config_name);
+    if ($config['editor'] === 'ckeditor5' && isset($config['editor_settings']['cketoolbar'])) {
+      $editor_settings = $config['editor_settings'];
+      $toolbar = $editor_settings['cketoolbar'];
+      unset($editor_settings['cketoolbar']);
+      $config['editor_settings'] = array_merge(array(
```
indigoxela: Wouldn't $config->set() and $config->clear() be slightly more readable?

And where's $config->save()?

...wait a minute. You're treating the config object as array? Oops.

quicksketch: Thanks @indigoxela, I don't know what I was thinking here. Maybe I accidentally missed a commit or something? There's no way this would work as-is. I'll rework it.

quicksketch: Updated commit that should save properly.

> Wouldn't $config->set() and $config->clear() be slightly more readable?

This approach keeps the config keys in the expected order. Using `clear()` and `set()` would push the new `toolbar` key to the bottom of the array of settings instead of as the first item.

quicksketch:
Merged into 1.x for 1.27.0.

PR #4626, opened 2024-01-03
Issue #6351: Fix jquery.ui focus conflict for CKEditor form elements

### LINKED ISSUE #6351: CKEditor 5 dialogs from within a Backdrop (jquery.ui) dialog not working
## Description of the bug

It's a tricky problem.

When editing block content via layout form, which opens the editor in a dialog, the dialog traps the focus, so opening a CKEditor modal fails

## Steps To Reproduce

1. Create a custom block in a layout
2. Insert an image (the Backdrop image dialog *is not the culprit here*)
3. Try to set the alternative text for the image

## Actual behavior

For a _moment_ the alt text dialog seems to open, but closes immediately after that.
It's not possible to input any alternative text for **inline images in blocks** with this context toolbar item.

editor-dialog-detail

## Expected behavior

Dialogs should work.

## Additional information

Seems to be some conflict between jquery.ui dialog (Backdrop), which tries to trap the focus it seems, and the CKEditor 5 dialog, which doesn't get the focus, so form elements like text input don't work.

It might be possible to catch that aggressive `focusin` handling by preventing propagation under some circumstances. Or override _allowInteraction in the jquery widget. Or we could drop that form element, as alt text can still get set via image dialog.

indigoxela:
Fixes https://github.com/backdrop/backdrop-issues/issues/6351


laryn:
Thanks @indigoxela, @quicksketch, and @herbdool -- merged!

PR #4628, opened 2024-01-06
Issue #6355: Clean up and harden Views "Add fields" modal.

### LINKED ISSUE #6355: [UX][D8] Views "Add fields" modal: Split item title/category/description into separate columns, and streamline JS used
## Description of the bug

While working on Gin I noticed that the Views modal for selecting fields or filters could use some clean up. Here are some of the things I'd like to improve:
 - Each field is prefaced by the category, which makes it hard to read field names/labels at a glance. In Drupal 10 the category is in a completely separate column after the field label, ~but perhaps it would be enough to just put it after the field label? I'm curious if others agree that this is harder to read with the category inserted at the front.~ and the latest PR replicates that.
 - The script that shows which fields have been selected could be made a little less fragile. If a theme is overriding `theme_checkbox` it is very easy to break that (as Gin does) and it can be hardened with a minor tweak. Right now it looks for the label exactly following the checkbox, but (for example) Gin adds a span directly after the checkbox, so the "Selected: " interface is broken. ~Changing that to look for a sibling label instead of whatever happens to be directly following seems a little stronger.~ [Code from Drupal 10](https://git.drupalcode.org/project/drupal/-/blob/10.2.x/core/modules/views_ui/js/views-admin.js?ref_type=heads#L33-593) has been adapted to match the corresponding structural changes and is more robust.

## Steps To Reproduce

To reproduce the behavior:
1. Go to a View and click on "Add" in the "Fields" or "Filter criteria" section.
2. Scroll the list.

## Screenshot (before)
CleanShot 2024-02-04 at 14 14 30@2x

## Screenshot (after)
CleanShot 2024-02-04 at 13 45 59@2x

laryn:
Fixes https://github.com/backdrop/backdrop-issues/issues/6355


core/modules/views_ui/views_ui.admin.inc
```diff
@@ -3775,86 +3775,87 @@ function views_ui_add_item_form($form, &$form_state) {
...
+    '#title' => t('Search'),
+  );
 
-    $form['options']['name'] = array(
-      '#prefix' => '<div class="views-radio-box form-checkboxes views-filterable-options">',
-      '#suffix' => '</div>',
-      '#tree' => TRUE,
-      '#default_value' => 'all',
-    );
+  $groups = array('all' => t('- All -'));
```
klonos: Should we be using [`'#empty_option'`](https://docs.backdropcms.org/form_api#empty_option) here instead?

core/modules/views_ui/views_ui.admin.inc
```diff
@@ -3775,86 +3775,87 @@ function views_ui_add_item_form($form, &$form_state) {
...
+    $groups[$group] = $field['group'];
+    $fields[$key]['#attributes'] = array(
+      'class' => array(
+        'views-filterable-option',
+        $group,
+      ),
+    );
+    $fields[$key]['help'] = "<div class='description'>" . $field['help'] . "</div>";
+    $fields[$key]['title'] = "<div class='title'>" . $field['title'] . "</div>";
+    $fields[$key]['group'] = "<div class='group'>" . $field['group'] . "</div>";
```
klonos: Classes and other attributes can be added to table rows/cells in a similar, standardized manner as they can be added to the table header, via arrays. We should be doing that here as well if those `<div>`s are only used in order to add the classes and aren't otherwise explicitly necessary.

Example from `theme_layout_reorder_layouts()`:
```php
  // Assemble the data for a table from everything in $form['handlers']
  foreach (element_children($element) as $layout_name) {
    if (isset($element[$layout_name]['title'])) {
      $row = array();
      $row[] = array(
        'data' => render($element[$layout_name]['title']),
        'class' => array('layout-name'),
      );
      $element[$layout_name]['weight']['#attributes']['class'][] = 'weight';
      $row[] = render($element[$layout_name]['weight']);
      $rows[] = array('data' => $row, 'class' => array('draggable'));
    }
  }
```

laryn: Thanks @klonos, please see if the last commit addresses your concerns:
https://github.com/backdrop/backdrop/pull/4628/commits/6ac48cf6903e24022e5830797f270016e0b31a91

core/modules/views_ui/views_ui.admin.inc
laryn: Thanks @klonos, please see if the last commit addresses your concerns:
https://github.com/backdrop/backdrop/pull/4628/commits/6ac48cf6903e24022e5830797f270016e0b31a91

klonos: Yes, it does 👍🏼 

Thanks

core/modules/views_ui/views_ui.admin.inc
klonos: Yup, this also looks much better now - no "manual" HTML cruft 👍🏼 

Thanks again.

core/modules/views_ui/js/views-admin.js
```diff
@@ -298,120 +298,128 @@ Backdrop.behaviors.viewsUiSearchOptions = {
...
+     * @type {jQuery}
+     */
+    this.$form = $form;
 
-  /**
-   * Keyup handler for the search box that hides or shows the relevant options.
-   */
-  handleKeyup: function (event) {
-    var found, i, j, option, search, words, wordsLength, zebraClass, zebraCounter;
+    // Click on the title checks the box.
```
docwilmot: This isn't working for me on Chrome.

core/modules/views_ui/views_ui.admin.inc
```diff
@@ -3775,86 +3775,106 @@ function views_ui_add_item_form($form, &$form_state) {
...
+    '#empty_option' => t('- All -'),
+  );
+
+  $form['options']['name'] = array(
+    '#prefix' => '<div class="views-radio-box form-checkboxes views-filterable-options">',
+    '#suffix' => '</div>',
+    '#tree' => TRUE,
+  );
+
+  // Build the sortable table header.
```
docwilmot: Its not a sortable table though.

core/modules/views_ui/views_ui.admin.inc
```diff
@@ -3775,86 +3775,106 @@ function views_ui_add_item_form($form, &$form_state) {
...
   }
-  // Add a div to show the selected items
+  $form['options']['name']['fields'] = array(
+    '#type' => 'tableselect',
+    '#header' => $header,
+    '#options' => $fields,
+    '#empty' => t('There are no @types available to add.', array(
+      '@types' => $ltitle,
+    )),
+    '#attributes' => array('id' => 'views-filterable-options'),
```
docwilmot: This causes every checkbox to have this as an id.

docwilmot (commented):
This was not an exhaustive review, but noted a few things.

laryn:
Thanks @docwilmot -- I'll take a look at your comments. I appreciate you looking.

EDIT: I think I've addressed all your comments.

core/modules/views_ui/js/views-admin.js
laryn: This was working earlier -- I must have broke it accidentally when trying to remove the whitespace changes.

laryn: I take it back, it wasn't whitespace. I must have changed something after I tested this initially -- but I have it working again and it should be fixed on the next commit.

core/modules/views_ui/views_ui.admin.inc
laryn: I'm not sure if you mean something more here, but I'm removing the word `sortable`.

core/modules/views_ui/views_ui.admin.inc
laryn: Thank you -- removed.

core/modules/views_ui/js/views-admin.js
laryn: Should be fixed in https://github.com/backdrop/backdrop/pull/4628/commits/ced86e2ea8bc5069a8910c27b479098863ecbab9

core/modules/views_ui/js/views-admin.js
```diff
@@ -197,7 +197,7 @@ Backdrop.viewsUi.AddItemForm = function ($form) {
 
 Backdrop.viewsUi.AddItemForm.prototype.handleCheck = function (event) {
   var $target = $(event.target);
-  var label = $.trim($target.next().text());
+  var label = $.trim($target.closest("td").next().text());
```
klonos: I beliebe that we prefer single to double quotes unless there's a specific reason:
```suggestion
  var label = $.trim($target.closest('td').next().text());
```

core/modules/views_ui/js/views-admin.js
```diff
@@ -298,120 +298,128 @@ Backdrop.behaviors.viewsUiSearchOptions = {
...
-    }
-    else {
-      this.$form.find('.views-filterable-group').hide();
-      this.$form.find('.views-filterable-group-' + groupName).show();
-    }
-  }
-});
+  $.extend(
+    Backdrop.viewsUi.OptionsSearch.prototype,
+  /** @lends Backdrop.viewsUi.OptionsSearch# */ {
```
klonos, line 342: What is this `@lends`? I've never come across it before.

laryn: @klonos Not going to lie, that was completely lifted from the Drupal 10 code as noted in the first issue comment.
https://jsdoc.app/tags-lends

stpaultim:
@laryn Sandbox failed!

laryn:
@stpaultim I've updated the branch again. Thanks.

core/modules/views_ui/views_ui.admin.inc
```diff
@@ -3769,86 +3769,108 @@ function views_ui_add_item_form($form, &$form_state) {
...
     );
-
-    $groups = array('all' => t('- All -'));
-    $form['options']['controls']['group'] = array(
-      '#type' => 'select',
-      '#title' => t('Filter'),
-      '#options' => array(),
+    // Here we generate what will be the ID of the checkbox for this $key.
+    // This allows us to tie the label to the checkbox for better accessibility.
+    $checkbox_id = 'edit-name-fields-' . backdrop_html_class($key);
```
klonos: I'm not sure which is best here, the way it is concatenated now, or like this instead:
```suggestion
    $checkbox_id = backdrop_html_class('edit-name-fields-' . $key);
```

In any case, can we please add something to the inline comment to explain why we are intentionally using `backdrop_html_class()` here instead of `backdrop_html_id()` here?

core/modules/views_ui/js/views-admin.js
```diff
@@ -298,120 +298,122 @@ Backdrop.behaviors.viewsUiSearchOptions = {
...
-  /**
-   * Keyup handler for the search box that hides or shows the relevant options.
-   */
-  handleKeyup: function (event) {
-    var found, i, j, option, search, words, wordsLength, zebraClass, zebraCounter;
+    const searchBoxSelector = '#edit-options-search';
+    const controlGroupSelector = 'select[name="group"]';
+    this.$form.on(
+      'formUpdated',
+      `${searchBoxSelector},${controlGroupSelector}`,
```
quicksketch, line 319: We don't often see backticks in Backdrop JS code, as this kind of string interpolation did not work in any version of IE at all (even up to 11), see https://caniuse.com/template-literals. As we no longer support any IE version, I think we're now safe to include this.

core/modules/views_ui/js/views-admin.js
quicksketch: Never seen this before either, but I'm generally in favor of these kinds of type hints to variables. It can really help with IDE autocomplete.

core/modules/views_ui/js/views-admin.js
```diff
@@ -197,7 +197,7 @@ Backdrop.viewsUi.AddItemForm = function ($form) {
 
 Backdrop.viewsUi.AddItemForm.prototype.handleCheck = function (event) {
   var $target = $(event.target);
-  var label = $target.next().text().trim();
+  var label = $target.closest('td').next().text().trim()
```
quicksketch: I resolved a conflict with this line, but dropped a semi-colon:

```suggestion
  var label = $target.closest('td').next().text().trim();
```

quicksketch:
Merged into 1.x and 1.28.x. Thank you folks!

PR #4632, opened 2024-01-12
Issue #6364: Set white-space to normal for CKEditor 5.

### LINKED ISSUE #6364: CKEditor 5 - multivalue textfields break layout
## Description of the bug

This was first reported in the contrib module queue with Paragraphs, but the problem's unrelated to that module.

## Steps To Reproduce

1. Add a multivalue text field to a content type
2. Set Text processing to filtered
3. Make sure, Seven is set as theme for content creation
4. Add a node of that type (/node/add/whatever)
5. Paste a longer text into one of the multivalue fields

## Actual behavior

The layout breaks, as the editor's content gets a `white-space: nowrap` from the parent.

## Expected behavior

Guess... :wink: 

## Additional information

- Backdrop CMS version: latest dev or 1.27 preview

@laryn discovered the CSS line causing this: https://github.com/backdrop-contrib/ckeditor5/issues/125#issuecomment-1889609592

We can't remove that line, it would break other stuff, but we can narrow it down, so it doesn't apply to the editor's div, I guess.

The line causing the problem: https://github.com/backdrop/backdrop/blob/1.x/core/modules/system/css/system.theme.css#L95

laryn:
Fixes https://github.com/backdrop/backdrop-issues/issues/6364

### DIFF
--- a/core/modules/ckeditor5/css/ckeditor5.css
+++ b/core/modules/ckeditor5/css/ckeditor5.css
@@ -45,3 +45,7 @@
 .ck-content .image:not([class*="image-style-"]) {
   margin: .9em 0 !important;
 }
+
+.ck-editor {
+  white-space: normal;
+}
--- a/core/modules/ckeditor5/css/ckeditor5.css
+++ b/core/modules/ckeditor5/css/ckeditor5.css
@@ -46,6 +46,9 @@
   margin: .9em 0 !important;
 }
 
+/**
+ * Ensure wrapping within the editor is the default behavior.
+ */
 .ck-editor {
   white-space: normal;
 }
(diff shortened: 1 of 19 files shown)


core/modules/system/css/system.theme.css
```diff
@@ -94,6 +94,9 @@ tr.even .form-item {
   margin-bottom: 0;
   white-space: nowrap;
 }
+.ck-editor {
+  white-space: normal;
+}
```
indigoxela: This would also work in core/modules/ckeditor5/css/ckeditor5.css

Worth considering?

core/modules/ckeditor5/css/ckeditor5.css
```diff
@@ -45,3 +45,7 @@
 .ck-content .image:not([class*="image-style-"]) {
   margin: .9em 0 !important;
 }
+
```
quicksketch, line 48:
```suggestion

/**
 * Ensure wrapping within the editor is the default behavior.
 */
```

quicksketch:
Thanks @laryn! I added a code comment and then merged this into 1.x for 1.27.0.

PR #4633, opened 2024-01-14
Issue #6325: Better CSS for flex layout admin

### LINKED ISSUE #6325: Inconsistent use of styles on Flexible Template Builder UI
## Description of the bug

The UI for building a Flexible Template has some unusual CSS patterns that are not consistent with the rest of core. This page should have similar design patterns to the Manage Blocks page. That said, the elements we are moving around are not blocks, so they should not match the pattern for blocks on the manage block page to avoid confusion.

Screenshot 2023-12-06 at 2 56 05 PM

Some individual problems are noted below, but this page could probably use a a design review / refresh as a whole.
* Strange spacing between border and interior
* use of blue dotted line (unique to only this page)
* box shadow too large (should be `box-shadow: 2px 2px 0 0 #d0d0d0;` for consistency)
* combination of box-shadow and dotted line looks strange

stpaultim:
Fixes: https://github.com/backdrop/backdrop-issues/issues/6325

Screenshot 2024-05-07 at 8 46 23 PM


core/modules/layout/css/layout.flexible.admin.css
```diff
@@ -24,13 +24,13 @@
 }
 .layout--flexible .layout-editor-block-title {
   padding: 0.4em 1em;
-  margin-bottom: 10px;
+  margin:  0 -.9375rem .9375rem;
```
jenlampton: Margins on block titles on the manage blocks page are done in `em`s. Can we do that here too?

```
margin: 0 0.5em 0 0; /* LTR */
```

jenlampton: Or... is this to counter-act the bootstrap grid? If so we also need to do the same for all other screen sizes.... That might be a lot of work. What does it look like without the margin fix? Is it acceptable?

core/modules/layout/css/layout.flexible.admin.css
```diff
@@ -39,14 +39,12 @@
 
 .layout-flexible-editor .flexible-row {
   padding: 0;
-  border: 1.25px dashed #0074bd;
+  border: 1.25px solid #CCC;
```
jenlampton: This needs to be lowercase:
`border: 1px solid #ccc;` 
why `1.25`? It looks like 1px on the manage blocks UI?

core/modules/layout/css/layout.flexible.admin.css
```diff
@@ -39,14 +39,12 @@
...
   margin-bottom: 15px;
   border-radius: 4px;
-  box-shadow: 3px 3px 0 0 #c6c6c6;
-}
-.layout-flexible-editor .flexible-row:hover {
-  background-color: #ffffff;
   cursor: move;
+  box-shadow: 2px 2px 0 0 #d0d0d0;
+  background-color: #f7f7f7;
```
jenlampton: this could be simply `background: #f7f7f7;` (that's what it is on the manage blocks UI)

core/modules/layout/css/layout.flexible.admin.css
jenlampton: We definitely need this.

jenlampton: okay, it does look like we are trying to counteract the bootstrap grid:
```
.container-fluid {
    padding-right: .9375rem;
    padding-left: .9375rem;
}
```
Fortunately this padding is the same at all breakpoints :) p
lease ignore this whole thread, I'll close it.

core/modules/layout/css/layout.flexible.admin.css
stpaultim: I took the 1.25 px that was there before. Agree that 1px probably makes more sense and is more consistent. I made that change.

core/modules/layout/css/layout.flexible.admin.css
stpaultim: Done

core/modules/layout/css/layout.flexible.admin.css
stpaultim: Thanks.

core/modules/layout/css/layout.flexible.admin.css
stpaultim: @jenlampton - I don't know if you looked, but elsewhere in the flexible layout template css, we're using background-color everywhere. So maybe I should either change it everywhere or not at all?

I already made the change, so I'll take your advice on this.

stpaultim: Well, I looked again and it's only in two other places. Should I make this change from `background-color` to `background` in all three places or leave it as it is?

core/modules/layout/css/layout.flexible.admin.css
```diff
@@ -67,3 +65,10 @@
   background-clip: content-box;
   text-align: center;
 }
+
+.l-flexible-row .layout-editor-region {
+  border: 1px solid white;
+}
```
klonos: We should remove this in order to be consistent with the dashed border around the regions in the "Manage blocks" UI (it will also help a bit with the contrast issues).

core/modules/layout/css/layout.flexible.admin.css
```diff
@@ -24,13 +24,13 @@
...
-  margin-bottom: 10px;
+  margin:  0 -.9375rem .9375rem;
 }
 .layout--flexible .layout-editor-block-title .text {
-  font-weight: normal;
+  font-weight: bold;
 }
 .layout--flexible .layout-editor-block-content p {
-  color: #575357;
+  color: black;
```
klonos: In the "Manage blocks" UI this is `#000`. Can we use that here as well?

stpaultim: Sure thing!

core/modules/layout/css/layout.flexible.admin.css
stpaultim: Good idea.

core/modules/layout/css/layout.flexible.admin.css
jenlampton: Let's not change anything else in this issue :)

core/modules/layout/css/layout.flexible.admin.css
jenlampton: @klonos the dashed border has meaning ("you can drop things here") that doesn't apply to regions on this interface. (That's why there are no dashed borders on regions on this page.)

On this page the "drop zones" are for rows, and if it looked like the rows could be dropped into regions... that would be a problem :) This css is to prevent the dashes from appearing in the wrong place and confusing people about where to drop things.

I'm also not sure about the white, though, I think it looks better as `border: 1px solid #ccc;`

jenlampton:
We could be a little more consistent with the manage blocks interface if we made the main background gray, and also provided a drop-zone for rows with a white background and gray dashed border.

Add the same styles on `#layout-flexible-edit-main` as are on `#layout-edit-main`. 
```
#layout-edit-main {
    padding: 20px 0 0;
    border-radius: 4px;
    background-color: #f7f7f7;
    border: 2px solid #EAEAEA;
}
```

Then also add these to `#layout-flexible-edit-main`
```
	padding: 15px 15px 0;
    margin: 0 15px;
    border: 1px dashed #ccc;
    background: #fff;
```

Manage blocks page with drop zone:
Screenshot 2024-05-08 at 4 31 08 PM

Flexible layout page with drop-zone:
Screenshot 2024-05-08 at 4 39 01 PM

I'll make a PR against your PR @stpaultim

jenlampton:
> I'll make a PR against your PR

My PR against your PR just changes the border:  https://github.com/stpaultim/backdrop/pull/6

I made a new PR with the drop-zone for rows because I'm not sure it's any better than this approach: https://github.com/backdrop/backdrop/pull/4738
See the main issue for screenshots to compare.

core/modules/layout/css/layout.flexible.admin.css
```diff
@@ -15,22 +15,27 @@
...
 }
 
 .layout--flexible .layout-editor-block {
   box-shadow: none;
+  cursor: row-resize;
 }
 .layout--flexible .layout-editor-block-title {
   padding: 0.4em 1em;
-  margin-bottom: 10px;
+  margin:  0 -.9375rem .9375rem;
```
klonos: Extra space needs to be removed here:
```suggestion
  margin: 0 -.9375rem .9375rem;
```

stpaultim: @klonos Good catch. I saw that before and thought i had fixed it. Apparently not.

core/modules/layout/css/layout.flexible.admin.css
```diff
@@ -9,28 +9,33 @@
...
 }
 
 .layout--flexible .layout-editor-block {
   box-shadow: none;
+  cursor: row-resize;
 }
 .layout--flexible .layout-editor-block-title {
   padding: 0.4em 1em;
-  margin-bottom: 10px;
+  margin: 0 -.9375rem .9375rem -.9375rem;
```
klonos: Sorry @stpaultim ...another coding standard issue with leading zeros for decimal values:
```suggestion
  margin: 0 -0.9375rem 0.9375rem -0.9375rem;
```

I wasn't able to find a specific mention of this in https://docs.backdropcms.org/css-standards or https://www.drupal.org/docs/develop/standards/css/css-formatting-guidelines, however, we don't seem to be doing it like that anywhere else in core.

Omitting the zeros especially makes the negative values seem really odd.

klonos:

quicksketch:
Looks great! Thank you folks! I merged this into 1.x and 1.28.x.

PR #4639, opened 2024-01-18
Issue #6374: Flush the layout path cache when toggling the theme for content edit

### LINKED ISSUE #6374: Switching the node_admin_theme option should flush the layout cache but doesn't
## Description of the bug

It's a rather subtle problem, as the default admin theme and the default theme aren't that different. But it's visible. Even more if the layouts do differ significantly.

## Steps To Reproduce

1. Switch the setting "Use the administration theme when editing or creating content" off and save on admin/appearance
2. Open a node form, like node/add/post
3. Inspect the current layout via admin bar dropdown "This page"
4. Switch the setting "Use the administration theme when editing or creating content" back on and save (do not flush caches manually)
5. Open a node form
6. Inspect the current layout via admin bar

## Actual behavior

When the default theme is selected for node edit/create, the Default Layout should be in use.
When the admin theme is selected, the Default Administrative Layout should be used.
But switching doesn't change that, so the wrong Layout is used, until one flushes caches.

This leads to visual glitches.
I've seen this before, but didn't actually realize the source of the glitch.

wrong-layout

^^ That's a harmless example, it can look a lot weirder, depending on layout adaptions (blocks).

## Expected behavior

The right layout after switching.

## Additional information

- Backdrop CMS version: latest dev, not sure when this problem started

indigoxela:
Fixes https://github.com/backdrop/backdrop-issues/issues/6374

### DIFF
--- a/core/modules/node/node.module
+++ b/core/modules/node/node.module
@@ -2667,7 +2667,7 @@ function node_form_system_themes_admin_form_submit($form, &$form_state) {
   config_set('system.core', 'node_admin_theme', $form_state['values']['node_admin_theme']);
   // The default and admin layout are tied directly to this setting, so clear
   // the Layout cache.
-  layout_reset_caches()
+  layout_reset_caches();
 }
 
 /**
--- a/core/modules/node/node.module
+++ b/core/modules/node/node.module
@@ -2665,7 +2665,9 @@ function node_form_system_themes_admin_form_alter(&$form, &$form_state, $form_id
  */
 function node_form_system_themes_admin_form_submit($form, &$form_state) {
   config_set('system.core', 'node_admin_theme', $form_state['values']['node_admin_theme']);
-  cache('layout_path')->flush();
+  // The default and admin layout are tied directly to this setting, so clear
+  // the Layout cache.
+  layout_reset_caches()
 }
 
 /**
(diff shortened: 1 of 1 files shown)


core/modules/node/node.module
```diff
@@ -2665,6 +2665,7 @@ function node_form_system_themes_admin_form_alter(&$form, &$form_state, $form_id
  */
 function node_form_system_themes_admin_form_submit($form, &$form_state) {
   config_set('system.core', 'node_admin_theme', $form_state['values']['node_admin_theme']);
+  cache('layout_path')->flush();
```
quicksketch: Just to be comprehensive, maybe we should use Layout module's dedicated cache clear function? Since themes and layouts are so tightly coupled, it might prevent other issues in the future.
```suggestion
  // The default and admin layout are tied directly to this setting, so clear
  // the Layout cache.
  layout_reset_caches()
```

indigoxela: Hm. This makes switching that single setting "more expensive".

But you're right, they're tightly coupled, so better flush all, including all layout config. Switching this setting doesn't happen very often, I guess.

indigoxela: Oops, where's the semicolon. :speak_no_evil:

quicksketch:
Merged into 1.x and 1.27.x. Thanks @indigoxela!

PR #4640, opened 2024-01-23
Issue #1535: Add current page as a link in the breadcrumb.

### LINKED ISSUE #1535: Append the current tab in the breadcrumb (even if as an invisible element).
This is related to #1509 

It would help us be consistent in the UI as well as when people are using screen readers. "You are here" will start making sense.

laryn:
Fixes https://github.com/backdrop/backdrop-issues/issues/1535


core/modules/book/tests/book.test
```diff
@@ -95,12 +95,38 @@ class BookTestCase extends BackdropWebTestCase {
...
+      $nodes[2],
+    ));
+    $this->checkBookNode($nodes[3], NULL, $nodes[2], $book, $nodes[4], array(
+      $book,
+      $nodes[3],
+    ));
+    $this->checkBookNode($nodes[4], NULL, $nodes[3], $book, FALSE, array(
+      $book,
+      $nodes[4],
+    ));
```
klonos: I'm assuming that PHPCS was complaining about the formatting here (things like "the array exceeds 80 characters" etc.), however the way it is now is much harder to read and get an overview of what's happening than before.

Can we please do either one of these two things:
- Suppress the PHPCS nagging using a set of `// phpcs:disable`/`// phpcs:disable` before/after these lines and otherwise leave the formatting/restructuring unchanged.
- Use some helper variables that make it obvious what the parameters are each time `$this->checkBookNode()` is called.

core/modules/book/tests/book.test
```diff
@@ -118,8 +144,13 @@ class BookTestCase extends BackdropWebTestCase {
...
     $this->book = $other_book;
-    $this->checkBookNode($other_book, array($node), FALSE, FALSE, $node, array());
-    $this->checkBookNode($node, NULL, $other_book, $other_book, FALSE, array($other_book));
+    $this->checkBookNode($other_book, array($node), FALSE, FALSE, $node, array(
+      $other_book,
+    ));
+    $this->checkBookNode($node, NULL, $other_book, $other_book, FALSE, array(
+      $other_book,
+      $node,
+    ));
```
klonos: Same as my previous comment here.

core/modules/simpletest/tests/menu.test
```diff
@@ -972,82 +972,85 @@ class MenuBreadcrumbTestCase extends MenuWebTestCase {
...
+    );
     $tree = $expected + array(
       'menu-test/breadcrumb/tasks' => $title,
     );
     $this->assertBreadcrumb('menu-test/breadcrumb/tasks', $trail, $title, $tree);
-    $this->assertBreadcrumb('menu-test/breadcrumb/tasks/first', $trail, $title, $tree);
-    $this->assertBreadcrumb('menu-test/breadcrumb/tasks/first/first', $trail, $title, $tree);
     $trail += array(
-      'menu-test/breadcrumb/tasks' => t('Breadcrumbs test: Local tasks'),
+      'menu-test/breadcrumb/tasks/second' => "Second",
```
klonos: Any reason why we are using double quotes here?

And also, was the removal of `t()` intentional? If so, then perhaps remove it elsewhere where similar changes are happening(?)

core/modules/system/system.module
```diff
@@ -3397,12 +3397,19 @@ function system_header_block($settings) {
...
+  $output = '';
   $crumbs = backdrop_get_breadcrumb();
-  if (isset($settings['current']) && $settings['current'] == TRUE) {
-    $crumbs[] = '<span aria-current="page">' . backdrop_get_title() . '</span>';
+  if (!empty($crumbs)) {
+    if (empty($settings['current'])) {
+      // Hide the last link if the block settings don't call for showing it, for
+      // backwards compatibility.
+      $last_crumb = array_pop($crumbs);
+      $crumbs[] = str_replace('class="', 'class="element-invisible ', $last_crumb);
```
klonos, line 3407: Is there an other, more "standard" PHP or Backdrop/Drupal way to add the class here?

laryn: @klonos I'm open to a suggestion -- any ideas? :)

PR #4641, opened 2024-01-26
Issue #6372: Fix wrapper_attributes warnings when updating.

### LINKED ISSUE #6372: PHP notice after update from 1.26.3 to 1.27.0
## Description of the bug

Notice: Undefined index: wrapper_attributes in theme_item_list() (line 2413 of /DOCROOT/core/includes/theme.inc).

## Actual behavior

Multiple notices in dblog.
Currently I cannot debug it, maybe someone else has seen it too.

## Additional information
- Web server and its version: Apache 2.4.56
- PHP version: 7.4.33
- Database sever (MySQL or MariaDB?) and its version: 5.5.5-10.5.21-MariaDB-0+deb11u1
- Operating System and its version: Debian GNU/Linux 11.8

bugfolder:
Fixes https://github.com/backdrop/backdrop-issues/issues/6372.

### DIFF
--- a/core/update.php
+++ b/core/update.php
@@ -303,7 +303,6 @@ function update_info_page() {
 
   // Flush the theme cache so we can render this page correctly if the theme
   // registry been updated with new preprocess or template variables. 
-  system_rebuild_theme_data();
   backdrop_theme_rebuild();
 
   // Get database name
--- a/core/update.php
+++ b/core/update.php
@@ -300,8 +300,9 @@ function update_info_page() {
   if (db_table_exists('cache_update')) {
     cache('update')->flush();
   }
-  // Flush the theme cache so we can render this page correctly if themes have
-  // been updated.
+
+  // Flush the theme cache so we can render this page correctly if the theme
+  // registry been updated with new preprocess or template variables. 
   system_rebuild_theme_data();
   backdrop_theme_rebuild();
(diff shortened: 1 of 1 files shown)


core/update.php
```diff
@@ -300,6 +300,10 @@ function update_info_page() {
   if (db_table_exists('cache_update')) {
     cache('update')->flush();
   }
+  // Flush the theme cache so we can render this page correctly if themes have
+  // been updated.
```
quicksketch: Small rewording of the code comment:

```suggestion

  // Flush the theme cache so we can render this page correctly if the theme
  // registry been updated with new preprocess or template variables. 
```

core/update.php
```diff
@@ -300,6 +300,10 @@ function update_info_page() {
...
+  system_rebuild_theme_data();
```
quicksketch: To solve the problem of additional or changed variables being passed to theme functions and templates, I don't think this line is necessary, which scans the file system for new themes.

This might solve _other_ potential problems but unless we have a situation in which we know it's solving a problem, we might be better off omitting this line to save unnecessary work when running updates.

```suggestion
```

quicksketch:
Great, thanks @bugfolder! Merged into 1.x and 1.27.x.

PR #4645, opened 2024-02-01
Issue #3166: Fix views autosubmit focus problem after view reload

### LINKED ISSUE #3166: [UX] Views: exposed filter with autosubmit and AJAX loses keyboard focus

Views 7.x known issue: https://www.drupal.org/project/views/issues/1217948

https://drupal.stackexchange.com/questions/29503/views-exposed-filter-auto-submit-loses-focus

### Steps to reproduce

1. On a vanilla Backdrop installation, install the Devel module, and enable it + the Devel Generate submodule.
2. Use Devel Generate, to generate a large(ish) amount of content (say 100 or 500 nodes).
3. Edit the "Administer content" view:
   - Enable AJAX from the "Other" section of options.
   - In the "Exposed form" section, click the "Settings" link of the "Exposed form style" option, scroll to the bottom of the dialog, and enable the "Autosubmit" option.
4. Save the view.
5. Head to `admin/content`, and start typing in the "Title contains" exposed filter...

### Actual behavior

When the form autosubmits (after you have typed a couple of characters in the exposed text field), the text box loses focus. The user now needs to click on the text filter again, then as they start typing and autosubmit kicks in, the filter loses focus again. Repeat...

Typing quickly into the exposed filter works fine, however typing slowly will cause focus loss.

Some people report that separating the exposed form into a block fixes the problem, however this is not an option if the view itself is a block display.

### Expected behavior

The exposed text filters do not lose focus as the user types and the results listing gets refreshed.

indigoxela:
Fixes https://github.com/backdrop/backdrop-issues/issues/3166

Different approach with event listener on "ajaxStop" event instead of setTimeout.

### DIFF
--- a/core/misc/autosubmit.js
+++ b/core/misc/autosubmit.js
@@ -39,20 +39,23 @@ Backdrop.behaviors.autosubmit = {
     // 'this' references the form element
     function triggerSubmit (e) {
       var $this = $(this);
-      let searchFieldID = '#' + $this.find('.autosubmit-processed').attr('id');
-      $('body').attr('data-needs-focus', searchFieldID);
+      // If this form is child of the view, not an independent block, make sure
+      // we can set focus back on the input field, after the whole view got
+      // replaced.
+      if ($this.parents('.view').length) {
+        $('body').attr('data-needs-focus', $this.find('.autosubmit-processed').attr('id'));
+      }
       $this.find('.autosubmit-click').trigger('click');
     }
 
-    // Add back focus on form item inside a form that got replaced via AJAX.
     // It's important to remove previous listeners, otherwise they stack up with
-    // every attach.
+    // every attach, triggered by typing.
     $(document).off('ajaxStop').on('ajaxStop', function () {
       if (!$('body[data-needs-focus]').length) {
         return;
       }
-      let id = $('[data-needs-focus]').data().needsFocus;
-      let $textInput = $(id);
+      let id = $('body').data().needsFocus;
+      let $textInput = $('#' + id);
       let pos = $textInput.val().length;
       $textInput.focus();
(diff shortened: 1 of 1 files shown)


indigoxela:
Closing after discussion. :wink:

indigoxela:
Oops, misunderstanding. :stuck_out_tongue_winking_eye:

yorkshire-pudding (approved):
The comments are clear and the code makes sense to me and my limited JS knowledge

core/misc/autosubmit.js
```diff
@@ -39,9 +39,29 @@ Backdrop.behaviors.autosubmit = {
...
+      // replaced.
+      if ($this.parents('.view').length) {
+        $('body').attr('data-needs-focus', $this.find('.autosubmit-processed').attr('id'));
+      }
       $this.find('.autosubmit-click').trigger('click');
     }
 
+    // It's important to remove previous listeners, otherwise they stack up with
+    // every attach, triggered by typing.
+    $(document).off('ajaxStop').on('ajaxStop', function () {
```
argiepiano, line 53: Question: would it be better to use `once()` instead of removing/attaching the listener?

indigoxela: I don't think that would work. The stop event is dispatched globally (document level), where would I put that "once" identifier?
Or do you have a clever idea, how to accomplish that?

argiepiano: You are right. I did some testing - `once()` doesn't seem to be working on the `document` element, and `ajaxStop` only seems to work when attached to the document. So your code looks and works well.

I guess another alternative to tweaking the JS is to avoid replacing the exposed form upon ajax reloading.  But I haven't looked into this.

indigoxela:
Closing in favor of #4653

PR #4653, opened 2024-02-09
Issue 3166. Retain focus after autoxubmit.

### LINKED ISSUE #3166: [UX] Views: exposed filter with autosubmit and AJAX loses keyboard focus

argiepiano:
Fixes https://github.com/backdrop/backdrop-issues/issues/3166

### DIFF
--- a/core/misc/autosubmit.js
+++ b/core/misc/autosubmit.js
@@ -1,130 +1,130 @@
 (function($){
-  /**
-   * To make a form auto submit, all you have to do is 3 things:
-   *
-   * Attach the autosubmit library to the form.
-   *
-   * @code
-   *   $form['#attached']['library'][] = array('system', 'backdrop.autosubmit');
-   * @endcode
-   *
-   * On elements you want to auto-submit when changed, add the autosubmit
-   * class. With FAPI, add:
-   * @code
-   *   $form['field_name']['#attributes']['class'] = array('autosubmit');
-   * @endcode
-   *
-   * If you want to have auto-submit for every form element,
-   * add the autosubmit-full-form to the form. With FAPI, add:
-   * @code
-   *   $form['#attributes']['class'] = array('autosubmit-full-form');
-   * @endcode
-   *
-   * If you want to exclude a field from the autosubmit-full-form auto submission,
-   * add the class autosubmit-exclude to the form element. With FAPI, add:
-   * @code
-   *   $form['field_name']['#attributes']['class'] = array('autosubmit-exclude');
-   * @endcode
-   *
-   * Finally, you have to identify which button you want clicked for autosubmit.
-   * @code
-   *   $form['my_button']['#attributes']['class'] = array('autosubmit-click');
-   * @endcode
-   *
-   * Currently only 'select', 'radio', 'checkbox' and 'textfield' types are
-   * supported.
-   */
-  Backdrop.behaviors.autosubmit = {
-    attach: function(context) {
-      // 'this' references the form element
-      function triggerSubmit (element) {
-        var $this = $(this);
-        var $triggering_element = $(element);
-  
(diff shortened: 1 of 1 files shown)


core/misc/autosubmit.js
```diff
@@ -37,12 +37,39 @@
...
       var $this = $(this);
+      var $triggering_element = $(element);
+
+      // Variable "element" will have a value only when text fields trigger
+      // this. If element is undefined, then remove the data to prevent
+      // potential focus on a previously processed element.
+      if (element === undefined)  {
+        $('body').removeData('needs-focus');
+      }
+      else if ($triggering_element.parents('.view').length) {
```
quicksketch: I don't think we want to put a Views-specific selector in here. I don't think there would be harm in just always populating this data attribute and always re-focusing the last autocomplete element after autosubmit.

core/misc/autosubmit.js
```diff
@@ -37,12 +37,39 @@
...
+        $('body').data('needs-focus', $triggering_element.attr('id'));
+      }
+
+      // Submit the form.
       $this.find('.autosubmit-click').trigger('click');
     }
 
-    // the change event bubbles so we only need to bind it to the outer form
+    // Listener to ajaxStop will re-focus on the text field as needed.
+    $(document).off('ajaxStop').on('ajaxStop', function () {
```
quicksketch: This `.off('ajaxStop')` would unbind _any_ `ajaxStop` handler. This should be namespaced with `autosubmit` to only remove/add the handler from this file.

core/misc/autosubmit.js
```diff
@@ -37,12 +37,39 @@
...
-    function triggerSubmit (e) {
+    function triggerSubmit (element) {
       var $this = $(this);
+      var $triggering_element = $(element);
+
+      // Variable "element" will have a value only when text fields trigger
+      // this. If element is undefined, then remove the data to prevent
+      // potential focus on a previously processed element.
+      if (element === undefined)  {
+        $('body').removeData('needs-focus');
```
quicksketch: We should probably namespace this data attribute to more clearly identify that it's coming from `autosubmit.js`. I'd suggest `autosubmit-last-focus-id` to be more descriptive.

quicksketch (commented):
I tested this and it works well! I worry that this may cause issues in non-Views scenarios (I'm not sure if there are any in core right now to test though). I filed a PR with my suggestions here: https://github.com/argiepiano/backdrop/pull/2

quicksketch:
Merged into 1.x and 1.27.x. Thanks @argiepiano!

PR #4657, opened 2024-02-14
Issue #6404: Add jQuery 4.x as additional option

### LINKED ISSUE #6404: Add jQuery 4.x as additional option
The first 4.x Beta for jQuery has been [published recently](https://blog.jquery.com/2024/02/06/jquery-4-0-0-beta/).

It brings several changes that affect core code, which isn't fully compatible, yet. We should start to prepare for a stable 4.x release.

**Are we there, yet?** Short answer: no.

We have some (not many) usages of deprecated (and removed in 4) functions, covered in https://github.com/backdrop/backdrop-issues/issues/6248 :point_left: I'd suggest to start there.

We have to decide on a strategy with **third party code** that isn't compatible anymore and isn't maintained anymore:

- jquery.form.js
- jquery.cookie.js
- jquery.timeentry.js (see also #6249)
- ???

While we might consider to just drop jquery.cookie.js and jquery.timeentry.js, that's no option for jquery.form.js.
Note that Drupal struggles with a similar problem: https://www.drupal.org/project/drupal/issues/3419734 Let's keep an eye on their considerations.

indigoxela:
Fixes https://github.com/backdrop/backdrop-issues/issues/6404

jQuery 4 has a stable release.


indigoxela:
Hello Tugboat...

indigoxela:
Another randomly failing test... :roll_eyes: Unrelated.

core/modules/system/system.module
```diff
@@ -2192,7 +2192,20 @@ function system_library_info_alter(&$libraries, $module) {
...
-
+  elseif ($config->get('jquery_version') == 'latest') {
+    $libraries['jquery']['version'] = '4.0.0';
+    $libraries['jquery']['js'] = array(
+      'core/misc/jquery-4.0.0.min.js' => array(
+        'group' => JS_LIBRARY,
+        'weight' => -20,
+      ),
+      // Backwards compatibility.
+      'core/misc/jquery-4-bc.js' => array(
```
herbdool:
```suggestion
      'core/misc/jquery-4-compat.js' => array(
```

We already have a `jquery-ui-compat.js` file so I think we should use that naming convention.

core/modules/system/system.admin.inc
```diff
@@ -1848,6 +1848,7 @@ function system_jquery_settings() {
   $options = array(
     'default' => t('jQuery 3 (recommended)'),
     'legacy' => t('jQuery 1 (legacy)'),
+    'latest' => t('jQuery 4 (experimental)'),
```
herbdool: Having the key as `latest` is probably fine, though I was thinking about what our strategy will be when jQuery 5 comes out. I assume we'll have `latest` point to jQuery 5. And then we would either move jQuery 4 to `default`; add another option like `jquery4` or make it a hidden option.

core/modules/system/system.module
indigoxela: That's a good idea. :+1:

core/modules/system/system.admin.inc
indigoxela: I like the "jquery4" idea - using that just right away seems to be more clear than some name like "latest", which we couldn't change to jquery5 in our 1.x, anyway.
I don't expect their v5 to come out any time soon, but a milestone exists.

core/modules/system/system.module
```diff
@@ -2192,7 +2192,20 @@ function system_library_info_alter(&$libraries, $module) {
       ),
     );
   }
-
+  elseif ($config->get('jquery_version') == 'jquery4') {
+    $libraries['jquery']['version'] = '4.0.0';
+    $libraries['jquery']['js'] = array(
+      'core/misc/jquery-4.0.0.min.js' => array(
```
herbdool, line 2213: Just realized that we're not including the `.min.` even if the file is minimized. See `jquery.js`.

indigoxela: Are you suggesting to rename the file? Does that provide value?
The file name `jquery-4.0.0.min.js` is exactly the official download without modifications.

Some 3rd party libs in misc aren't minified, some are... And then in `ui` the files do have the "min" in their names. It's a wild mix. :wink:

indigoxela:
Try the new cspell json.

indigoxela:
Next attempt with my "beloved" cspell... :wink:

quicksketch:
Merged into 1.x for 1.34.0. Thank you @indigoxela!

PR #4662, opened 2024-02-28
Issue 6410. Add node/%/translate as a contextual link.

### LINKED ISSUE #6410: [UX] Add "Translate" link to contextual content links
## Description of the need

When referenced content items are displayed on a page as "Rendered entity", single items can be _edited_ and _deleted_ using its contextual links. On multilingual sites, I would expect there also a contextual link to _translate_ an item. Without such a link, the translation workflow is quite complicated.

## Proposed solution

For content types with enabled translation, add a "Translate" link in the contextual links of referenced or otherwise embedded nodes (e.g. a view with content teasers). The path should be `node/NID/translate` (or `node/NID-REFERENCED/translate?destination=node/REFERENCING-NID`).

A custom or contrib module could provide such a link. In my opinion, it makes however sense to have the feature in Backdrop core. (Drupal 8+ does so, by the way.)

Backdrop now includes contextual links to improve the content translation workflow.

argiepiano:
Fixed https://github.com/backdrop/backdrop-issues/issues/6410

### DIFF
--- a/core/modules/translation/translation.module
+++ b/core/modules/translation/translation.module
@@ -35,6 +35,7 @@ function translation_menu() {
     'access callback' => '_translation_tab_access',
     'access arguments' => array(1),
     'type' => MENU_LOCAL_TASK,
+    'context' => MENU_CONTEXT_PAGE | MENU_CONTEXT_INLINE,
     'weight' => 2,
     'file' => 'translation.pages.inc',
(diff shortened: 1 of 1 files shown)


quicksketch:
Thanks @argiepiano! Merged into 1.x and 1.27.x.

PR #4663, opened 2024-03-01
Issue #6412: Fix wrong display of some float values in Views 

### LINKED ISSUE #6412: Fixed: Wrong display of some float values in Views.
## Description of the bug

The views_handler_field_math uses code copied from views_handler_field_numeric (this is even mentioned in a comment)
That code snippet has a bug due to limitations of PHP and float values: it produces wrong results for certain values.

## Steps To Reproduce

Create any type of View with a display format using Fields.
Add a Global:Math Expression field and enter a decimal value in the Expression input field, e.g. 67.93 (the behaviour is only visible with certain values such as this example).
Don't check the Round checkbox.
Preview or display the View.
Instead of 67.93, the output will be 67.93000000000001

There is another bug:
Use any number between 0 and -1, e.g. -0.5.
Instead of -0.5, the output will be 0.5.

## Additional information

This bug was partly fixed in Drupal7 for numeric fields in issue [#662692](https://www.drupal.org/project/views/issues/662692) only, but not applied to the same code used for Math expressions.
There was another fix in Drupal8 for numeric fields in issue [#1952926](https://www.drupal.org/project/drupal/issues/1952926#comment-12082454) which fixes both issues described above.

I'm aiming at providing a port of the bugfix as in Drupal8 for both numeric and math fields.

AlexHoebart-ICPDR:
Same approach applied to numeric and math handlers
Fixes https://github.com/backdrop/backdrop-issues/issues/6412


core/modules/views/handlers/views_handler_field_numeric.inc
```diff
@@ -124,24 +124,30 @@ class views_handler_field_numeric extends views_handler_field {
...
-    }
+    // Use round first to avoid negative zeros.
+    $value = round($value, $precision);
+    // Test against both integer zero and float zero.
+    if ($this->options['empty_zero'] && ($value === 0 || $value === 0.0)) {
+       return '';
+     }
 
+    $value = number_format($value, $precision, $this->options['decimal'], $this->options['separator']);
+
```
indigoxela: Github doesn't show this, but there's trailing white space on this line. :wink: At least, when applying the patch.

AlexHoebart-ICPDR: I also can't see it, and no idea how it came in. But somehow I think I removed it.
How did you find it?

indigoxela: Confirmed, the extra spaces are gone. :+1: 

You can get a raw diff or commit diff by rewriting the url like:
```
..../backdrop/pull/4663.diff
or
..../backdrop/pull/4663.patch
```
And then you can apply these patches locally. And that patch caused my local git to nag on the spaces.
Anyway, it's fixed. :smiley:

quicksketch:
Merged into 1.x and 1.27.x. Thanks @AlexHoebart-ICPDR and @indigoxela!

PR #4665, opened 2024-03-07
Issue #6412: Add functional test for display of float values in Views

### LINKED ISSUE #6412: Fixed: Wrong display of some float values in Views.

indigoxela:
Addition to fix https://github.com/backdrop/backdrop-issues/issues/6412, add the functional test to complement https://github.com/backdrop/backdrop/pull/4663

Also fixes coding standards in file views_handler_field_math.test and sets the profile to the more lightweight 'testing'.


quicksketch:
Thanks @indigoxela! I merged this PR into https://github.com/backdrop/backdrop/pull/4663 so the changes are all in one place!

PR #4669, opened 2024-03-08
Issue #6293: Prevent dashboard loading before layout module in upgrades.

### LINKED ISSUE #6293: Functional tests: LayoutUpgradePathTest fails too often in GHA
## Description of the bug

Yet another failure of **LayoutUpgradePathTest**. In two different PHP versions. In a PR, that doesn't even touch PHP-code.

This tests fails too often, to call it "occasional". Not always the same PHP version, not all PHP versions at once, but way too often to ignore.

The test chokes with: `Call to undefined function layout_load_multiple_by_path()` (ships with layout.module).

## Steps To Reproduce

No idea, the test passes locally, of course.

Some race condition of update hooks, when upgrading? Doesn't seem reasonable, as the Layout module gets enabled in system_update_1000().

## Additional information

- Backdrop CMS version: latest dev, 
- PHP version: random

quicksketch:
Fixes https://github.com/backdrop/backdrop-issues/issues/6293 (maybe)

I just noticed in another issue I was getting early bootstrap failures because of module loading order. Maybe specifically listing dashboard.module depending on layout.module will prevent it from accidentally loading if layout module isn't yet enabled in the Backdrop upgrade?

### DIFF
--- a/core/modules/dashboard/dashboard.info
+++ b/core/modules/dashboard/dashboard.info
@@ -5,3 +5,4 @@ version = BACKDROP_VERSION
 type = module
 backdrop = 1.x
 configure = admin/dashboard/settings
+dependencies[] = layout


laryn:
Thanks @quicksketch! Merged into `1.x`.

PR #4671, opened 2024-03-13
Issue #5153: Consider filtered text field empty, if format is default

### LINKED ISSUE #5153: Field Type - text long - an extra empty field gets added with every save.
**Description of the bug**

When creating a content type and adding a "Text (long) Field multi-row" - 

If these fields are set as follows...

- Text processing - filtered
- Row count is changed to 10 for example
- field is set to an unlimited number of values
 
When adding content, an empty field is saved which can not be deleted.  If the same content is edited again and saved, another empty field is created.

This also happens with Unlimited "Text Short"

**Steps To Reproduce**

To reproduce the behavior:
1. Create a new content type: Structure->Content Type->Add content type
2. name: test
3. Save and add fields
4. Add new field: Label: anything
5. Field type: Text (long)
6. Widget: text area (multiple rows)
7. Change Text Processing to Filtered
8. Global settings->number of values: change to unlimited
9. save settings
10. add content....
11. Content->add content->[content just created above]
12. Enter a title
13. Save
14. Edit the content again
15. Do not change anything... 
16. Save the content
17. Edit the content again
18. Repeating will add another empty text field

**Expected behavior**

Only adding new text areas when clicking on the "Add another item"

**Additional information**

- Backdrop CMS version: 1.19.2 & 1.19.3
- Web server and its version: Shared hosting
- Browser(s) and their versions: Chrome 92.0.4515.131 (Official Build) (64-bit)

indigoxela:
Fixes https://github.com/backdrop/backdrop-issues/issues/5153

### DIFF
--- a/core/modules/field/modules/text/text.module
+++ b/core/modules/field/modules/text/text.module
@@ -218,9 +218,9 @@ function text_field_load($entity_type, $entities, $field, $instances, $langcode,
  */
 function text_field_is_empty($item, $field) {
   if (!isset($item['value']) || $item['value'] === '') {
-    if (isset($item['format']) && $item['format'] !== '') {
-      // If this text field has a text format setting,
-      // save it regardless of empty value to keep the format.
+    if (isset($item['format']) && $item['format'] !== '' && $item['format'] != filter_default_format()) {
+      // If this text field has a text format setting, save it regardless of
+      // empty value to keep the format. Unless it's the default format.
       return FALSE;
     }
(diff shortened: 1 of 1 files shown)


leeksoup:
Just tested on a couple of my nodes that had acquired multiple empty field elements. Works for me!

quicksketch:
Thanks @indigoxela! I merged this into 1.x and 1.27.x.

PR #4681, opened 2024-03-21
Issue #6432: Undefined constant "LOG_LOCAL0" when upgrading from Drupal 7 on Windows.

### LINKED ISSUE #6432: Undefined constant "LOG_LOCAL0" in system_update_1017()
## Description of the bug

On Windows, when doing a Drupal to Backdrop upgrade, I get this error. The line mentioned is `$config->set('log_facility', update_variable_get('syslog_facility', LOG_LOCAL0));`
`LOG_LOCAL0` is not defined in Windows. In D7, in the place the `syslog_facility` variable is accessed, there is a check for if its defined. We missed this in Backdrop.

## Steps To Reproduce

Uncertain; I get the error on WampServer on Windows 11. I dont think I have always had this error, so not sure if something about my current setup may be to blame.

In either case, we should fix this.

docwilmot:
Fixes https://github.com/backdrop/backdrop-issues/issues/6432

### DIFF
--- a/core/modules/system/system.install
+++ b/core/modules/system/system.install
@@ -1994,7 +1994,7 @@ function system_update_1017() {
   $config = config('system.core');
   $config->set('log_row_limit', update_variable_get('dblog_row_limit', 1000));
   $config->set('log_identity', update_variable_get('syslog_identity', 'backdrop'));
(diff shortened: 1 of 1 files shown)


quicksketch:
Thanks @docwilmot! Merged into 1.x and 1.27.x.

PR #4686, opened 2024-04-03
Issue #6440: Provide views handlers for plain file_managed.uid

### LINKED ISSUE #6440: Provide views definitions for file_managed uid
Currently, when you create a View of files, you can't include the uid of the author as a field, nor can you use uid as a filter. Having uid available as field, filter and relationship would be very helpful. **For example**, it would allow you to create views where you can have a contextual filter that defaults to the currently logged in user, thus limiting the view of files to only the person who's logged in.  Of course, there are other helpful use cases.

The snapshots below just show that a filter by uid is not currently available. 

Screen Shot 2024-04-01 at 8 52 29 AM

Screen Shot 2024-04-01 at 8 52 35 AM

indigoxela:
Fixes https://github.com/backdrop/backdrop-issues/issues/6440

### DIFF
--- a/core/modules/file/views/file.views.inc
+++ b/core/modules/file/views/file.views.inc
@@ -228,7 +228,7 @@ function file_views_data() {
   // uid
   $data['file_managed']['uid'] = array(
     'title' => t('User who uploaded'),
-    'help' => t('The user account that uploaded the file. If you need more fields than the uid add the File: User who uploaded relationship'),
+    'help' => t('The user account that uploaded the file. If more fields than UID are needed, add the "File: User who uploaded" relationship'),
     'relationship' => array(
       'title' => t('User who uploaded'),
       'label' => t('User who uploaded'),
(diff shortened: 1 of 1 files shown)


core/modules/file/views/file.views.inc
```diff
@@ -228,13 +228,22 @@ function file_views_data() {
   // uid
   $data['file_managed']['uid'] = array(
     'title' => t('User who uploaded'),
-    'help' => t('The user account that uploaded the file.'),
+    'help' => t('The user account that uploaded the file. If you need more fields than the uid add the File: User who uploaded relationship'),
```
quicksketch:
```suggestion
    'help' => t('The user account that uploaded the file. If more fields than UID are needed, add the "File: User who uploaded" relationship'),
```

quicksketch:
Merged into 1.x and 1.27.x. Thanks @indigoxela and @argiepiano!

PR #4711, opened 2024-04-26
Issue #6479: Use default layout when creating a new layout.

### LINKED ISSUE #6479: [UX] Set the default value for "Layout template" to the same template that's in use on the default layout.
When you create a layout from the layouts listing page currently, you are asked for the layout template every time.

Many of us regularly forget to select a template on this step (causing an immediate validation error when we enter the path). 

Screenshot 2024-04-26 at 2 33 42 PM

The reason we forget to select a template so often is because which template is in use is not even something we're thinking about -- because it's not relevant to our current task: placing blocks on the page.

This could be solved if the "Layout template" question had a default value.

@yorkshire-pudding  suggested:

>  the template used in the default layout should be selected by default

Yes please :)

jenlampton:
Fixes https://github.com/backdrop/backdrop-issues/issues/6479

### DIFF
--- a/core/modules/layout/layout.admin.inc
+++ b/core/modules/layout/layout.admin.inc
@@ -3204,7 +3204,7 @@ function layout_ajax_form_save_dialog($form, $form_state) {
 /**
  * AJAX handler that updates the title.
  *
- * return array
+ * @return array
  *   AJAX commands to update the page.
  */
 function layout_ajax_form_save_title_dialog($form, $form_state) {
--- a/core/modules/layout/layout.admin.inc
+++ b/core/modules/layout/layout.admin.inc
@@ -384,10 +384,18 @@ function layout_settings_form($form, &$form_state, Layout $layout) {
     '#required' => TRUE,
     '#access' => !$layout->isDefault(),
   );
+  if ($layout->layout_template != NULL) {
+    $layout_template = $layout->layout_template;
+  }
+  else {
+    // If no layout template has been selected, then pre-select the template
+    // used by the default layout, to avoid AJAX validation errors in the form.
+    $layout_template = config_get('layout.layout.default', 'layout_template');
+  }
   $form['layout_template'] = array(
     '#title' => t('Layout template'),
     '#type' => 'radios',
-    '#default_value' => $layout->layout_template,
+    '#default_value' => $layout_template,
     '#options' => array(),
     '#wrapper_attributes' => array('class' => array('clearfix', 'layout-options')),
     '#required' => TRUE,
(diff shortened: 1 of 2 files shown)


core/modules/layout/layout.admin.inc
```diff
@@ -384,10 +384,16 @@ function layout_settings_form($form, &$form_state, Layout $layout) {
     '#required' => TRUE,
     '#access' => !$layout->isDefault(),
   );
+  if ($layout->layout_template != NULL) {
+    $layout_template = $layout->layout_template;
+  }
+  else {
+    $layout_template = config_get('layout.layout.default', 'layout_template');
+  }
```
klonos: Lets add a comment here to explain what we are doing and why:
```suggestion
  else {
    // If no layout template has been selected, then pre-select the template
    // used by the default layout, to avoid AJAX validation errors in the form.
    $layout_template = config_get('layout.layout.default', 'layout_template');
  }
```

(something like that anyway - feel free to reword)

core/modules/layout/tests/layout.test
```diff
@@ -200,6 +200,12 @@ class LayoutInterfaceTest extends BackdropWebTestCase {
     $this->backdropGet('admin/structure/layouts');
     $this->clickLink(t('Add layout'));
 
+    // Test the default value for layout template is set.
+    $this->assertFieldByXPath('//input[@name="layout_template"][@checked="checked"]', TRUE, "The Layout template default value is set.");
+    $default_layout_template = config_get('layout.layout.default', 'layout_template');
+    $field_id = 'edit-layout-template-' . str_replace('_', '-', $default_layout_template);
+    $this->assertFieldChecked($field_id, "The Layout template default value is the same as the default layout.");
```
klonos: I believe that the preference is to be using single quotes (unless there are variables in the text that need to be replaced by actual values):
```suggestion
    // Test the default value for layout template is set.
    $this->assertFieldByXPath('//input[@name="layout_template"][@checked="checked"]', TRUE, 'The Layout template default value is set.');
    $default_layout_template = config_get('layout.layout.default', 'layout_template');
    $field_id = 'edit-layout-template-' . str_replace('_', '-', $default_layout_template);
    $this->assertFieldChecked($field_id, 'The Layout template default value is the same as the default layout.');
```

Otherwise this looks good 👍🏼

klonos (approved):

core/modules/layout/layout.admin.inc
```diff
@@ -3193,6 +3201,12 @@ function layout_ajax_form_save_dialog($form, $form_state) {
   );
 }
 
+/**
+ * AJAX handler that updates the title.
+ *
+ * return array
```
```suggestion
 * @return array
```

quicksketch:
Merged into 1.x and 1.27.x. Thanks @jenlampton!!

PR #4718, opened 2024-04-28
Issue #6486: Add json delivery callback to Entity Reference autocompletes.

### LINKED ISSUE #6486: Entity Reference autocomplete needs a JSON delivery callback
## Description of the bug

I stumbled across this while looking at something else, but it looks like the autocomplete menu items for Entity Reference need to have json delivery callbacks added, so the layout system doesn't insert a bunch of HTML at the end of the JSON result:

image

laryn:
Fixes https://github.com/backdrop/backdrop-issues/issues/6486.


quicksketch:
Merged into 1.x and 1.27.x. Thanks @laryn!

PR #4727, opened 2024-05-02
Issue #6501: Do not give the admin bar alert dot an icon.

### LINKED ISSUE #6501: Do not give the admin bar alert dot an icon
## Description of the bug

The "red dot" shown in the admin bar for critical status report notifications currently gets assigned an automatic "right chevron" icon. Also the padding on the text is slightly off so the number is not quite perfectly centered:

image

## Steps To Reproduce

To reproduce the behavior:
1. Fresh install of 1.x, leave "Check for Updates" enabled.
2. Open bootstrap.inc and set `BACKDROP_VERSION` to an older number.
3. Clear the caches.
4. Observer the admin bar red dot notification.

## Additional information

- Backdrop CMS version: 1.x (1.28.0-dev)
- Browser(s) and their versions: all

quicksketch:
Fixes https://github.com/backdrop/backdrop-issues/issues/6501

### DIFF
--- a/core/modules/admin_bar/admin_bar.inc
+++ b/core/modules/admin_bar/admin_bar.inc
@@ -797,7 +797,11 @@ function admin_bar_links_alert() {
       ),
       '#href' => 'admin/reports/status',
       '#access' => user_access('administer site configuration'),
-      '#options' => array('html' => TRUE),
+      '#options' => array(
+        // Suppress the automatic icon by specifying FALSE.
+        'icon' => FALSE,
+        'html' => TRUE,
+      ),
     );
   }
 
--- a/core/modules/admin_bar/admin_bar.theme.inc
+++ b/core/modules/admin_bar/admin_bar.theme.inc
@@ -73,7 +73,10 @@ function theme_admin_bar_links($variables) {
     $icon = '';
     if ($elements[$path]['#level'] < 2) {
       if (isset($elements[$path]['#options']['icon'])) {
-        $icon = icon($elements[$path]['#options']['icon'], array('immutable' => TRUE));
+        // If 'icon' is set to an empty string or FALSE, do not add an icon.
+        if ($elements[$path]['#options']['icon']) {
+          $icon = icon($elements[$path]['#options']['icon'], array('immutable' => TRUE));
+        }
       }
       else {
         $arrow_direction = $GLOBALS['language']->direction ? 'left' : 'right';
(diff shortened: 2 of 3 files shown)


core/modules/admin_bar/css/admin_bar.css
```diff
@@ -355,7 +355,7 @@ tr.admin-bar-tweak-permissions-processed td.module {
 /* Alert colors. */
 #admin-bar .admin-bar-alert .error-count {
   box-sizing: content-box;
-  padding: 0.4em 0.5em 0.5em 0.4em;
+  padding: 0.5em 0.4em 0.5em 0.4em;
```
quicksketch, line 358: Looks like just made an accident when I transposed these values from px to em. This is how it should have been written originally.

core/modules/admin_bar/admin_bar.theme.inc
```diff
@@ -73,7 +73,10 @@ function theme_admin_bar_links($variables) {
     $icon = '';
     if ($elements[$path]['#level'] < 2) {
       if (isset($elements[$path]['#options']['icon'])) {
-        $icon = icon($elements[$path]['#options']['icon'], array('immutable' => TRUE));
+        // If 'icon' is set to an empty string or FALSE, do not add an icon.
+        if ($elements[$path]['#options']['icon']) {
+          $icon = icon($elements[$path]['#options']['icon'], array('immutable' => TRUE));
```
indigoxela, line 78: I wonder, if this "immutable" is actually necessary.

It may apply to other items and it may be, that someone wants to change an icon later... :thinking: 

Probably off topic here. Just close, if it's out of scope.

quicksketch: Yeah, I did this so that themes could provide their own versions of icons but have it not affect the admin bar, since otherwise the icon would change between front-end and admin- themes.

An alternative for those really looking to change the icons is they could override the admin bar output, either with `hook_menu_alter()` and change the icon in the menu, or `hook_admin_bar_output_alter()` and change the icon option.

core/modules/admin_bar/admin_bar.inc
```diff
@@ -797,7 +797,11 @@ function admin_bar_links_alert() {
       ),
       '#href' => 'admin/reports/status',
       '#access' => user_access('administer site configuration'),
-      '#options' => array('html' => TRUE),
+      '#options' => array(
+        // Suppress the automatic icon by specifying FALSE.
```
klonos, line 801: We need some more explanation here. Perhaps linking to the GitHub issue would be enough.

quicksketch:
Merged into 1.x for 1.28.0 (preview).

PR #4729, opened 2024-05-03
Issue #6503: Add icon to language switcher in admin bar

### LINKED ISSUE #6503: Follow-up: Add icon to language switcher in admin bar
## Description of the bug

Another little follow-up, this time relevant for sites with the **locale** module enabled (multilingual sites).

The language switcher needs an icon, too. Currently it's the default, which doesn't fit well.

This one:
language-icon-missing

Easy to fix, now that we have an API for icons. And loads of icons. :smile: 

## Additional information

- Backdrop CMS version: 1.28.x

indigoxela:
Fixes https://github.com/backdrop/backdrop-issues/issues/6503

### DIFF
--- a/core/modules/admin_bar/admin_bar.inc
+++ b/core/modules/admin_bar/admin_bar.inc
@@ -753,7 +753,6 @@ function admin_bar_links_locale() {
     '#title' => isset($language->native) ? $language->native : $language->name,
     '#attributes' => array('class' => array('admin-bar-locale-links')),
     '#options' => array(
-      'html' => TRUE,
       'icon' => 'translate-fill',
     ),
   ) + admin_bar_locale_links();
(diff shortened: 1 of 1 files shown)


klonos (approved):
LGTM 👍🏼

core/modules/admin_bar/admin_bar.inc
```diff
@@ -752,6 +752,10 @@ function admin_bar_links_locale() {
   $links['locale'] = array(
     '#title' => isset($language->native) ? $language->native : $language->name,
     '#attributes' => array('class' => array('admin-bar-locale-links')),
+    '#options' => array(
+      'html' => TRUE,
```
quicksketch: I don't think `'html'` is needed here, and might actually cause an XSS issue if the language name (which is configurable) is printed without sanitization.

indigoxela: Oops, right, that unnecessary and possibly dangerous. Removed. :+1:

quicksketch:
Thanks @indigoxela and @klonos! Merged into 1.x for 1.28.0.

PR #4731, opened 2024-05-04
Issue #6458: Add PR number and timestamp to GHA artifact filenames

### LINKED ISSUE #6458: Upload Simpletest Results as an Artifact in GHA
## Description of the need

Debugging test failures in GitHub Actions is often difficult, because you can't see what is happening like you can when running tests locally through the UI. In particular, the contents of HTML pages that are loaded (and verbose messages) are helpful in that it shows you what the HTML browser sees while executing the test.

## Proposed solution

We could make things more transparent as to what's happening in our GitHub Action test runs by uploading the verbose output as an artifact and attaching it to the test runs.

I usually run tests locally when there is a failure, but occasionally there is an error that happens _only_ on GitHub (like we've been experiencing with #2277).

## Additional information

This can be achieved by adding this to the bottom of our test YML files:

```
      - name: Save Test Results
        uses: actions/upload-artifact@v4
        if: always()
        with:
          name: simpletest-output
          path: files/simpletest
```

klonos:
Fixes https://github.com/backdrop/backdrop-issues/issues/6458

Follow-up to https://github.com/backdrop/backdrop/pull/4730

### DIFF
--- a/.github/workflows/simpletest-db-config.yml
+++ b/.github/workflows/simpletest-db-config.yml
@@ -44,5 +44,5 @@ jobs:
         uses: actions/upload-artifact@v4
         if: always()
         with:
-          name: simpletest-db-config-${{ matrix.php-versions }}
+          name: ${{ github.event.number }}-simpletest-db-config-${{ matrix.php-versions }}
           path: files/simpletest/verbose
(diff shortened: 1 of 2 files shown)


quicksketch:
Closing per https://github.com/backdrop/backdrop-issues/issues/6458#issuecomment-2113625163.

We may revisit if we have actual issues with the file names as they are now.

PR #4739, opened 2024-05-09
Issue #6520: Show config storage locations in status report

### LINKED ISSUE #6520: Indicate what backend is used for config storage in the status page
With #2277 in for the upcoming 1.28.0 release, we should be indicating in the status report page whether files or the db is being used for config. Perhaps links to documentation as well?

klonos:
Fixes https://github.com/backdrop/backdrop-issues/issues/6520

### DIFF
--- a/core/modules/system/system.install
+++ b/core/modules/system/system.install
@@ -42,7 +42,10 @@ function system_requirements($phase) {
     }
 
     // Report the configuration storage being used.
-    $config_storage_types = array('active' => t('Active'), 'staging' => t('Staging'));
+    $config_storage_types = array(
+      'active' => t('Active'),
+      'staging' => t('Staging'),
+    );
     foreach ($config_storage_types as $type => $type_name) {
       $storage = config_get_config_storage($type);
       $class = get_class($storage);
--- a/core/modules/system/system.install
+++ b/core/modules/system/system.install
@@ -46,10 +46,10 @@ function system_requirements($phase) {
     $config_storage_staging = config_get_config_storage('staging');
     $config_storage_locations = array(
       'active' => t('Active configuration: @active_location', array(
-        '@active_location' => get_class($config_storage_active) == 'ConfigFileStorage' ? t('files') : t('database'),
+        '@active_location' => is_a($config_storage_active, 'ConfigFileStorage') ? t('files') : t('database'),
       )),
       'staging' => t('Staging configuration: @staging_location', array(
-        '@staging_location' => get_class($config_storage_staging) == 'ConfigFileStorage' ? t('files') : t('database'),
+        '@staging_location' => is_a($config_storage_staging, 'ConfigFileStorage') ? t('files') : t('database'),
       )),
     );
     $requirements['config_storage'] = array(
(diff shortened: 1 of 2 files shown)


core/modules/system/system.install
```diff
@@ -40,6 +40,22 @@ function system_requirements($phase) {
         'weight' => -9,
       );
     }
+
+    // Report the configuration storage being used.
+    $config_storage_active = config_get_config_storage('active');
+    $config_storage_staging = config_get_config_storage('staging');
+    $backend = array(
+      '@backend' => is_a($config_storage_active, 'ConfigFileStorage') ? t('files') : t('database'),
```
jenlampton: If we run this through t() and then also run it through t() in `t('Active configuration: @backend', $backend)` it's going to end up double escaped. 

Can we do `t('Active configuration: !backend', ...),` instead?

core/modules/system/system.install
```diff
@@ -40,6 +40,22 @@ function system_requirements($phase) {
...
+    );
+    $config_storage_locations = array(
+      'active' => t('Active configuration: @backend', $backend),
```
jenlampton: This isn't how  `t()` works we need to pass it an array with keys that match what's in the string, like...

`t('Active configuration: !backend', array('!backend' => $backend),`

core/modules/system/system.install
```diff
@@ -40,6 +40,22 @@ function system_requirements($phase) {
...
+      'staging' => t('Staging configuration: @backend', $backend),
```
jenlampton: Same here, need to fix `t()`.

core/modules/system/system.install
klonos: Of course. Good catch 👍🏼

core/modules/system/system.install
klonos: @jenlampton this is actually a trick that I have learned from one of your PRs a few years back IIRC. `$backend` is already an array. It has been pre-prepared in the lines earlier, and it is this:
```php
$backend = array(
  '@type' => $type,
  '!backend' => is_a($class, 'ConfigFileStorage') ? t('files') : t('database'),
);
```

This is keeping the `t()` short and neat, and the array keys/values can be seen right above. The good thing about it is that the same variable that holds the array can be reused in other `t()`s.

core/modules/system/system.install
klonos: Please see my earlier comment.

core/modules/system/system.install
klonos: ...well, after addressing feedback, the array evolved to this:
```php
      $replacements = array(
        '@type' => $type,
        '@prefix' => $prefix,
        '@config_location' => $config_location,
      );
```

And since there is nothing to be translated, I'm using `format_string()` now instead - so no double-escaping via `t()`.

klonos: OK, I see what you mean now @jenlampton ...I've refactored things, but not 100% sure if I'm doing it properly now 🤔 

Would appreciate it if you had a look.

quicksketch:
Merged into 1.x and 1.28.x. Thanks @klonos!

PR #4746, opened 2024-05-12
Issue #2254: Add prefixes to various block titles in the 'Add block' dialog

### LINKED ISSUE #2254: Prefix all menu block admin labels with "Menu:"
It bothers me that `Account menu` is the first block in the blocks list. I would prefer if all menu blocks were listed together. I think it would be also clearer for users if the blocks that were menu blocks clearly identified themselves as such.

I propose that we prefix all menu blocks with "Menu:"

This is a follow up to https://github.com/backdrop/backdrop-issues/issues/2241, where I saw the `Add block` list getting cleaned up, and wanted to add another clean-up task.

klonos:
Fixes https://github.com/backdrop/backdrop-issues/issues/2254

### DIFF
--- a/core/modules/layout/layout.admin.inc
+++ b/core/modules/layout/layout.admin.inc
@@ -1371,27 +1371,35 @@ function layout_block_add_page(Layout $layout, $renderer_name, $region_name, $bl
   // Assemble all blocks into a single list and sort.
   $block_list = array();
   foreach ($blocks as $module => $module_blocks) {
+    // Skip modules that don't provide any blocks.
     if (empty($module_blocks)) {
       continue;
     }
+
     foreach ($module_blocks as $delta => $block_info) {
       /* @var Block $parent_block */
       $parent_block = layout_create_handler('block', $module . ':' . $delta);
+      // Check whether entries coming from hook_block_info() implementations are
+      // "combos" (several "child" block instances provided within a single
+      // entry). This also covers individual field blocks (as returned from
+      // field_get_block_list()).
       $children = $parent_block->getChildren();
       if (is_array($children)) {
+        // Handle each entry within "combo blocks" in hook_block_info().
         foreach ($children as $child_id => $child_info) {
-          // Skip blocks that don't have context requirements met.
           $required_contexts = isset($child_info['required contexts']) ? $child_info['required contexts'] : array();
+          // Skip blocks that don't have context requirements met.
           if ($layout->hasContexts($required_contexts)) {
-            $block_list[$module . ':' . $delta . ':' . $child_id] = $child_info;
+            $block_list[$module . ':' . $delta . ':' . $child_id] = _layout_block_add_block_title_prefix($module, $child_info);
           }
         }
       }
       else {
+        // Regular, non-combo entries.
         $required_contexts = isset($block_info['required contexts']) ? $block_info['required contexts'] : array();
         // Skip blocks that don't have context requirements met.
         if ($layout->hasContexts($required_contexts)) {
-          $block_list[$module . ':' . $delta] = $block_info;
+          $block_list[$module . ':' . $delta] = _layout_block_add_block_title_prefix($module, $block_info);
         }
       }
     }
@@ -1443,6 +1451,40 @@ function layout_block_add_page(Layout $layout, $renderer_name, $region_name, $bl
   }
 
   return $output;
+}
+
+/**
+ * Helper to add prefixes to the block titles shown in the "Add block" dialog.
+ *
+ * Note: the "Field:" prefix for fields is already prepended in
+ * field_get_block_list().
+ *
+ * @see layout_block_add_page()
+ * @see field_get_block_list()
+ */
+function _layout_block_add_block_title_prefix($block_module, $block_info) {
+  if ($block_module == 'views') {
+    // Check if a "View:" prefix has already been added to the block
+    // description (see views_plugin_display_block::execute_hook_block_list)
+    // and add one if not.
+    $view_block_expected_prefix = "View: ";
+    $view_block_prefix_exists = strncmp($block_info['info'], $view_block_expected_prefix, strlen($view_block_expected_prefix)) === 0;
+    if (!$view_block_prefix_exists) {
+      $block_info['info'] = t('View: @block_title', array('@block_title' => $block_info['info']));
+    }
+  }
+  // Custom menus are coming from menu_block_info(), while system-defined menus
+  // from system_block_info().
+  if ($block_module == 'menu' || ($block_module == 'system' && isset($block_info['type']) && $block_info['type'] == 'menu')) {
+    $block_info['info'] = t('Menu: @block_title', array('@block_title' => $block_info['info']));
+  }
+  // Make sure to account for blocks intended for use in dashboards that are not
+  // necessarily provided by the Dashboard module itself.
+  if ($block_module == 'dashboard' || (isset($block_info['required contexts']) && in_array('dashboard', $block_info['required contexts']))) {
+    $block_info['info'] = t('Dashboard: @block_title', array('@block_title' => $block_info['info']));
+  }
+
+  return $block_info;
 }
 
 /**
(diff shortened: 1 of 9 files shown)


core/modules/layout/layout.admin.inc
```diff
@@ -1445,6 +1453,40 @@ function layout_block_add_page(Layout $layout, $renderer_name, $region_name, $bl
...
+    // and add one if not.
+    $view_block_expected_prefix = "View: ";
+    $view_block_prefix_exists = strncmp($block_info['info'], $view_block_expected_prefix, strlen($view_block_expected_prefix)) === 0;
+    if (!$view_block_prefix_exists) {
+      $block_info['info'] = t('View: @block_title', array('@block_title' => $block_info['info']));
+    }
+  }
+  // Custom menus are coming from menu_block_info(), while system-defined menus
+  // from system_block_info().
+  if ($block_module == 'menu' || ($block_module == 'system' && isset($block_info['type']) && $block_info['type'] == 'menu')) {
```
jenlampton, line 1478: I'm not sure I like this approach, it's a bit messy. Can we leave this as simply `$block_module == 'menu'` and then add the same prefix in system module for the menu blocks added?

core/modules/layout/layout.admin.inc
```diff
@@ -1445,6 +1453,40 @@ function layout_block_add_page(Layout $layout, $renderer_name, $region_name, $bl
...
+    $block_info['info'] = t('Menu: @block_title', array('@block_title' => $block_info['info']));
```
jenlampton, line 1479: I like using `:` more than ` - `. Let's update the Dashboard blocks to use a `:`: instead ✅

core/modules/layout/layout.admin.inc
```diff
@@ -1445,6 +1453,40 @@ function layout_block_add_page(Layout $layout, $renderer_name, $region_name, $bl
...
+  }
+  // Make sure to account for blocks intended for use in dashboards that are not
+  // necessarily provided by the Dashboard module itself.
+  if ($block_module == 'dashboard' || (isset($block_info['required contexts']) && in_array('dashboard', $block_info['required contexts']))) {
+    $block_info['info'] = t('Dashboard: @block_title', array('@block_title' => $block_info['info']));
```
jenlampton, line 1484: Same goes for dashboard, the prefix needs to be added in Dashboard module.

core/modules/menu/menu.module
```diff
@@ -491,6 +491,12 @@ function menu_reset_item($link) {
 
 /**
  * Implements hook_block_info().
+ *
+ * This returns the blocks for any custom menus created on the site (as opposed
```
jenlampton, line 495: This is why the `Menu` module was renamed `Menu UI` in Drupal 8. It  doesn't provide the ability to have menus, only the UI to create them :)

core/modules/node/node.module
```diff
@@ -2086,7 +2086,7 @@ function node_block_info() {
   );
   $blocks['content'] = array(
     'info' => t('Existing content'),
-    'description' => t('Embed existing node content as a block onto other pages'),
+    'description' => t('Embed content from existing content items as a block onto other pages.'),
```
jenlampton: I'm not a fan of `content from existing content items` Can we move this text change into a follow-up? It's not really relevant here and I don't want to hold this issue up :)

core/modules/system/system.module
```diff
@@ -2595,6 +2595,12 @@ function system_user_timezone(&$form, &$form_state) {
 
 /**
  * Implements hook_block_info().
+ *
+ * Note that this also returns the system-defined menu blocks (as opposed to any
```
jenlampton, line 2599: I don't think we need this, as any module can define menu blocks. There's nothing special about System module. (I do like the new docblock for Menu module though, because people will probably look there first for menu blocks.

core/modules/system/system.module
```diff
@@ -2633,6 +2639,11 @@ function system_block_info() {
       // Names from menu_list_system_menus() need to be translated through t().
       'info' => check_plain(t($custom_title)),
       'description' => t('A list of links for the menu: %title.', array('%title' => t($custom_title))),
+      // Helper, non-standard key to distinguish the menu blocks from other
+      // types of blocks provided by the System module.
+      // @todo: should this be provided by a system_layout_context_info()
+      // implementation instead or be a 'required contexts' key?
+      'type' => 'menu',
```
jenlampton, line 2646: This would be an API change. Let's not do this. Instead the 'Menu:' prefix in the name.

core/modules/system/tests/system.test
```diff
@@ -2908,6 +2908,7 @@ class MenuBlockTestCase extends BackdropWebTestCase {
    * Test Menu Block functionality.
    */
   function doMenuBlockTests($menu_name, $menu_title) {
+    $add_menu_block_link = 'Menu: ' . $menu_title;
```
jenlampton, line 2911: we don't need this if `Menu:` is in the `$menu_title;` ;)

core/modules/system/tests/system.test
```diff
@@ -2931,7 +2932,7 @@ class MenuBlockTestCase extends BackdropWebTestCase {
     $this->clickLink(t('Remove'), 1);
     // Add a block to the sidebar (the fourth region).
     $this->clickLink(t('Add block'), 3);
-    $this->clickLink($menu_title);
+    $this->clickLink($add_menu_block_link);
```
jenlampton, line 2935: we should be able to revert this :)

core/modules/views/plugins/views_plugin_display_block.inc
```diff
@@ -85,7 +85,10 @@ class views_plugin_display_block extends views_plugin_display {
       $info = $this->get_option('block_description');
     }
     else {
-      $info = t('View: @view (@display)', array('@view' => t($this->view->get_human_name()), '@display' => t($this->display->display_title)));
+      $info = t('View: @view (display: @display)', array(
```
jenlampton, line 88: love it! The prefix comes from views module ✅

core/modules/layout/layout.admin.inc
klonos: The PR already does that @jenlampton ...did I miss a spot?

core/modules/menu/menu.module
klonos: Yes, I have now realized this 👍🏼

core/modules/node/node.module
klonos: Sure thing 👍🏼 ...I've filed https://github.com/backdrop/backdrop-issues/issues/6539 to discuss wording and do that. I'll revert this change.

klonos: ...reverted with [`abebca9` (#4746)](https://github.com/backdrop/backdrop/pull/4746/commits/abebca91e476ee0edf8861ff45be5b633681faa3)

core/modules/layout/layout.admin.inc
klonos: Oh, I see what you mean: make the change in the `dashboard_block_info()` hook implementation. Gotcha 👍🏼

core/modules/views/plugins/views_plugin_display_block.inc
klonos: I didn't actually add the `View:` prefix here - I only added the `display:` prefix. The `View:` prefix was actually added already, which is actually causing problems with duplicated `View: View:` prefixes.

Because this is only added to certain views blocks (if there is no `block_description` added in the block display), you can end up with some views-provided blocks that have the prefix, and some that don't 😞 ...I had to add logic in `_layout_block_add_block_title_prefix()` to ensure that there aren't duplicated prefixes, while at the same time making sure that all views-provide blocks had the `Views:` prefix in the "Add block" dialog (so that they aren't "scattered" by the alphabetical sorting).

klonos: ...the reason why I added the `display:` prefix by the way is that we anded up with things like this (that is before any changes in this PR, and that is only one of the examples that I randomly found):
`View: Layout Context Test (User account) (Block)`

In the above title, you don't know what the strings in the brackets are. Which one is the name of the display? My change adds the `display:` prefix, so you get this instead:
`View: Layout Context Test (User account) (display: Block)`

PR #4755, opened 2024-05-15
Issue #364: Adding since docblocks for Icon API.

### LINKED ISSUE #364: Add an Icon API and SVG library to core.

quicksketch:
Fixes https://github.com/backdrop/backdrop-issues/issues/364

Adds missing `@since` lines and documentation to functions. Corresponds to the change record at https://docs.backdropcms.org/change-records/new-icon-api

### DIFF
--- a/core/includes/common.inc
+++ b/core/includes/common.inc
@@ -5545,6 +5545,8 @@ function backdrop_get_library($module, $name = NULL) {
  *   - immutable: Whether to use the original icon instead of any overrides.
  *
  * @see icon()
+ *
+ * @since 1.28.0 Function added.
  */
 function backdrop_add_icons(array $icon_names) {
   $icon_paths = array();
--- a/core/includes/icon.inc
+++ b/core/includes/icon.inc
@@ -2,6 +2,9 @@
 /**
  * @file
  * Provides the Backdrop API for placing icons.
+ *
+ * For comprehensive documentation on the icon API, see
+ * https://docs.backdropcms.org/documentation/icon-api
  */
 
 /**
@@ -33,6 +36,10 @@
  *     overridden, but an icon can still be added.
  *
  * @return string
+ *
+ * @see https://docs.backdropcms.org/documentation/icon-api
+ *
+ * @since 1.28.0 Function added.
  */
 function icon($icon_name, array $options = array()) {
   // Populate default options.
@@ -83,6 +90,8 @@ function icon($icon_name, array $options = array()) {
  *
  * @see hook_icon_info()
  * @see hook_icon_info_alter()
+ *
+ * @since 1.28.0 Function added.
  */
 function icon_get_path($icon_name, $immutable = FALSE) {
   // Save into a static cache with both normal and immutable paths.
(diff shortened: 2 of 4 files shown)


jenlampton:
I reviewed all the @since docs and they look good.

quicksketch:
Thanks @jenlampton!

PR #4766, opened 2024-05-27
Issue 6565. Remove instances of element_sort

### LINKED ISSUE #6565: Remove use of `element_sort` from core function.
The core function `file_view_multiple()` uses a [call](https://github.com/backdrop/backdrop/blob/fe5c21ced7a7e3bcb9b41df15a7977ee89384a17/core/modules/file/file.module#L3234) to `uasort` with `element_sort` as the callback. The issue is that this function only exists as part of the drupal compatibility layer, and will produce a fatal error if that layer is disabled and this function is invoked.

argiepiano:
Fixes https://github.com/backdrop/backdrop-issues/issues/6565


klonos (approved):
LGTM 👍🏼

quicksketch:
Merged into 1.x and 1.28.x. Thanks!

PR #4767, opened 2024-05-28
Issue #6570: Redesign login pages

docwilmot:
https://github.com/backdrop/backdrop-issues/issues/3309


core/modules/layout/layout.module
```diff
@@ -1442,8 +1442,15 @@ function layout_get_layout_by_path($path = NULL, $router_item = NULL, $is_layout
     return $layouts_by_path[$href];
   }
 
-  $layouts = layout_load_multiple_by_router_item($router_item, !$is_layout_page);
   $selected_layout = NULL;
+  if (in_array($href, array_keys(user_login_paths()))) {
+    // If this is a login path, use the default login layout.
+    $selected_layout = layout_load('login');
```
klonos, line 1448: Should we be checking if the layout is enabled here before using it?

core/modules/user/user.module
```diff
@@ -1989,6 +1989,37 @@ function user_login_final_validate($form, &$form_state) {
...
+
+  return $layouts;
+}
+
+/**
+ * Implements hook_layout_update().
+ */
+function user_layout_update(Layout $layout) {
+  // Clear page cache so anonymous sees new changes.
+  if ($layout->name == 'login') {
```
klonos, line 2018: Eventually, we should be checking if the layout has a `login` context (similarly to how dashboard layouts have a `dashboard` context). That should perform the cache clear regardless of the layout name (regardless of whether it is the one we provide or a custom-created layout).

core/modules/user/config/layout.layout.login.json
```diff
@@ -0,0 +1,56 @@
+{
+    "_config_name": "layout.layout.login",
+    "path": "",
+    "name": "login",
+    "title": "Login page Layout",
```
klonos, line 5: Lets be consistent and omit the `Layout` bit:
```suggestion
    "title": "Login page",
```

We're only adding that suffix for the default layout and the default admin layout, because plain `Default` and `Default administrative` would not stand on their own; whereas we use plain `Dashboard`, and `Home page`, w/o the "layout" suffix:
image

core/modules/layout/layout.module
```diff
@@ -1475,6 +1482,20 @@ function layout_get_layout_by_path($path = NULL, $router_item = NULL, $is_layout
   return $selected_layout;
 }
 
+/**
+ * Temporary. Implements hook_login_paths().
+ */
+function user_login_paths() {
```
jenlampton, line 1488: interesting!

docwilmot:
Closing this. As a solution to simplifying login pages, it adds a new layout and layout template. There wasn't much enthusiasm for this solution.

PR #4774, opened 2024-06-03
Issue #6582: Remove unused file.admin.js

### LINKED ISSUE #6582: [DX] Remove `file.admin.js` and deprecate `theme_file_display_order()`, since they are not being used in Backdrop
The `file_display_form()` function is responsible for the form at `admin/structure/file-types/manage/%file-type/file-display`. The only JS needed there is the one that shows/hides the settings fieldset for each file formatter, and that's done via `#states`.

I believe that that's either a left-over from older times (things ported over from D7 or 7.x contrib modules), or some oversight after some copy/paste.

As a matter of fact, looking at the code in `core/modules/file/js/file.admin.js`, I see comments about "tabledrag row belonging to this display" and "vertical tab belonging to this display" and "checkbox to conditionally show and hide the display's tableDrag row and vertical tab pane", which don't seem to apply to Backdrop at all when it comes to file displays. So I am questioning whether the entire `file.admin.js` file is even needed in the first place 🤔

klonos:
Fixes https://github.com/backdrop/backdrop-issues/issues/6582


core/modules/file/file.theme.inc
```diff
@@ -357,26 +357,12 @@ function theme_file_type_overview($variables) {
 
 /**
  * Returns HTML for a file display's display order table.
+ *
+ * @deprecated as of 1.28.1. In Backdrop there is no functionality to reorder
```
quicksketch:
```suggestion
 * @deprecated as of 1.32.2. In Backdrop there is no functionality to reorder
```

quicksketch:
Merged into 1.x and 1.32.x. Thank you @klonos and @herbdool!

PR #4778, opened 2024-06-07
Issue #6588: Add views field handler for SVG icon based on file exten…

### LINKED ISSUE #6588: Add new file type icon views field handler.
## Description of the need

While recently playing with a new module, I realized how nice our new icons would be to use in views.

## Proposed solution

A simple views handler for file_managed data.

Could also go to contrib, of course, but I belief, this belongs to core.

We're currently not using the new API much. So, let's start.

## Additional information

This is a screenshot to demo the intended use of such a views handler:

nice-icon-display

Simply by selecting a views field via views admin UI.

Backdrop now includes a views handler to automatically display an icon based on file extension.

---

## Translaion update

New strings
- An SVG icon according to file extension
- Inline SVG markup
- Inline SVG images are usually more efficient and easier to style.
- Regular images
- Regular image tags can be more efficient when showing the same icon many times.
- Add alternative text
- If alternative text is omitted, this icon will be marked as decorative only.
- File icon for @extension extension

indigoxela:
Fixes https://github.com/backdrop/backdrop-issues/issues/6588


core/modules/file/views/file.views.inc
```diff
@@ -177,6 +177,17 @@ function file_views_data() {
      ),
   );
 
+  // Icon based on file extension.
+  $data['file_managed']['icon'] = array(
+    'title' => t('Icon'),
```
klonos, line 182: Can we please call this "File type icon"?

core/modules/file/views/views_handler_field_file_icon.inc
```diff
@@ -0,0 +1,101 @@
...
+      $options = array(
+        'attributes' => array(
+          'width' => $this->options['icon_size'],
+          'height' => $this->options['icon_size'],
+        ),
+      );
+      $icon_name = 'file-' . $this->mapExtension($extension);
+      $icon = icon($icon_name, $options);
+      // Fall back to generic file icon.
+      if (strpos($icon, '<!-- Icon') === 0) {
```
klonos: The word `Icon` is passed through `t()` in the `icon()` function. Wouldn't that cause this check to fail in languages other than English then? 🤔

klonos: ...it would have been easier if `icon()` was returning `FALSE` if the icon is not found.

klonos: ...perhaps try using `icon_get_path()` instead for this check, since it returns `NULL`.

core/modules/file/views/views_handler_field_file_icon.inc
```diff
@@ -0,0 +1,101 @@
...
+      'ogg' => 'video',
+      'oga' => 'audio',
+      'ogv' => 'video',
+      'weba' => 'audio',
+      'webp' => 'image',
+      'webm' => 'video',
+      'tif' => 'image',
+      'tiff' => 'image',
+      'gz' => 'archive',
+      'tar' => 'archive',
```
klonos: Lets please alphabetize these.

core/modules/file/views/file.views.inc
indigoxela: That's a ... longer string that provides no better info in views admin UI, I think. :wink:

core/modules/file/views/views_handler_field_file_icon.inc
indigoxela: This is only a *fallback* that hardly ever strikes.

Many thanks for the `t()` hint - I'll take a look.

(It makes sense, that the icon() function  *always* returns a string. Really. :wink:)

core/modules/file/views/views_handler_field_file_icon.inc
```diff
@@ -0,0 +1,137 @@
...
+      '#max' => 300,
+      '#default_value' => $this->options['icon_size'],
+      '#field_suffix' => 'px',
+    );
+    $form['display_type'] = array(
+      '#type' => 'radios',
+      '#title' => t('Display type'),
+      '#options' => array(
+        'svg' => t('As svg tag'),
+        'img' => t('As img tag'),
```
klonos:
```suggestion
        'svg' => t('As <code>&lt;svg&gt;</code> tag'),
        'img' => t('As <code>&lt;img&gt;</code> tag'),
```

core/modules/file/views/views_handler_field_file_icon.inc
```diff
@@ -0,0 +1,137 @@
...
+      'pptx' => 'ppt',
+      'tar' => 'archive',
+      'tiff' => 'image',
+      'tif' => 'image',
+      'weba' => 'audio',
+      'webm' => 'video',
+      'webp' => 'image',
+      'xlsx' => 'xls',
+    );
+    $icon_name = 'file-' . $extension;
```
klonos: This feels like it should be an `else` in the `if` that follows.

core/modules/file/views/views_handler_field_file_icon.inc
indigoxela: From a translator's perspective the resulting string's a nightmare. :stuck_out_tongue_winking_eye: 
Maybe I find a better solution. Currently the options look too similar, anyway.

klonos: I like what you did in [`30f1aae` (#4778)](https://github.com/backdrop/backdrop/pull/4778/commits/30f1aae44ef477d6193c1fdff1b81fbf3ef2387a) much better 👍🏼

core/modules/file/views/views_handler_field_file_icon.inc
```diff
@@ -0,0 +1,141 @@
...
+  }
+
+  /**
+   * Render the icon as img tag.
+   */
+  protected function renderImg($icon_name) {
+    $attributes = array(
+      'width' => $this->options['icon_size'],
+      'height' => $this->options['icon_size'],
+      'alt' => '',
```
klonos, line 99: Should we add an `alt` text here for a11y? Something along the lines of:
```suggestion
      'alt' => t('File icon for .@extension extension', array('@extension' => ...)),
```

It could be the same `alt` text for the svg render above.

core/modules/file/views/views_handler_field_file_icon.inc
```diff
@@ -0,0 +1,141 @@
...
+    return '<img ' . backdrop_attributes($attributes) . '>';
+  }
+
+  /**
+   * Map file extensions to Phosphor icons.
+   *
+   * @param string $extension
+   *   The original file extension.
+   *
+   * @return string
```
klonos, line 121:
```suggestion
   * @return string
   *   The name of an icon file, as expected by icon().
```

core/modules/file/views/views_handler_field_file_icon.inc
indigoxela: I belief, an automatic alt text is worse than no text in this case - for A11Y.
It's pure decoration which doesn't provide extra value IMO. Of course, I could be wrong - I'm no A11Y expert.
But in the documentation I read, icons are usually described as "decorative without informational value".

indigoxela: @laryn suggested to add a checkbox to toggle alt text on/off. In that case, the admin can decide, whether it's better to add it or not (based on context).

@klonos I used your text suggestion.

core/modules/file/views/views_handler_field_file_icon.inc
```diff
@@ -0,0 +1,158 @@
...
+   *   The name of an icon file, as expected by icon().
+   */
+  protected function mapExtensionToName($extension) {
+    // If there's no 1:1 icon name match for the extension, map to one. It's
+    // possible that modules or themes provide additional file icons, so this
+    // only acts as a fallback.
+    $mapping = array(
```
quicksketch, line 128: A suggestion here, we could provide a generic mapping for image/video/archive/text by using `file_get_mimetype()`, which would return `image/gif`, `video/x-msvideo`, or similar. Then pull off the leading part to get `image`, `video`, etc. Then use the manual map for extensions with a specific icon.

quicksketch: In reworked this to use the mimetype mapping as described.

core/modules/file/views/views_handler_field_file_icon.inc
```diff
@@ -0,0 +1,161 @@
...
+      'width' => $this->options['icon_size'],
+      'height' => $this->options['icon_size'],
+      'alt' => '',
+      'src' => base_path() . icon_get_path($icon_name),
+    );
+    if ($this->options['add_alt_text']) {
+      $attributes['alt'] = t('File icon for @extension extension', array(
+        '@extension' => $extension,
+      ));
+    }
```
jenlampton, line 106:
```php
else {
  $attributes['aria-hidden'] = "true");
}
```

core/modules/file/views/views_handler_field_file_icon.inc
```diff
@@ -0,0 +1,161 @@
...
+      'attributes' => array(
+        'width' => $this->options['icon_size'],
+        'height' => $this->options['icon_size'],
+      ),
+    );
+    if ($this->options['add_alt_text']) {
+      $options['alt'] = t('File icon for @extension extension', array(
+        '@extension' => $extension,
+      ));
+    }
```
jenlampton, line 88:
```php
else {
  $attributes['aria-hidden'] = "true");
}
```

core/modules/file/views/views_handler_field_file_icon.inc
```diff
@@ -0,0 +1,161 @@
...
+
+    // If there is an extension.
+    if ($extension = pathinfo($uri, PATHINFO_EXTENSION)) {
+      $icon_name = $this->getIconName($uri);
+      if ($this->options['display_type'] == 'img') {
+        $icon = $this->renderImg($icon_name, $extension);
+      }
+      else {
+        $icon = $this->renderSvg($icon_name, $extension);
+      }
```
jenlampton, line 69:
```php
if (!$this->options['add_alt_text']) {
  $attributes['aria-hidden'] = "true");
}
```
Since it is the same for both, we could add it here?

core/modules/views_ui/css/views_ui.admin.theme.css
```diff
@@ -27,12 +27,12 @@
 
 /* @group Icons */
 
-.views-admin .icon {
+.views-admin .icon:not(svg) {
```
jenlampton, line 30: This could be a problem if/when we replace views sprites with SVGs. We can cross that bridge then :D

core/modules/file/views/views_handler_field_file_icon.inc
jenlampton: I've created https://github.com/backdrop/backdrop-issues/issues/6700 so we can do this for all icons.

core/modules/file/views/views_handler_field_file_icon.inc
```diff
@@ -0,0 +1,161 @@
...
+      '#max' => 300,
+      '#default_value' => $this->options['icon_size'],
+      '#field_suffix' => 'px',
+    );
+    $form['display_type'] = array(
+      '#type' => 'radios',
+      '#title' => t('Display type'),
+      '#options' => array(
+        'svg' => t('Inline <code>&lt;svg&gt;</code> tag'),
+        'img' => t('Regular <code>&lt;img&gt;</code> tag'),
```
indigoxela: These are awful strings for translators.
Something I actually tried to avoid. :grimacing: Can we please discuss that?
Because my previous strings were a result of discussion.

And the strings are (again) way to similar for a quick visual distinction.

core/modules/file/views/views_handler_field_file_icon.inc
quicksketch: This has been added.

core/modules/file/views/views_handler_field_file_icon.inc
quicksketch: Oops, I'm sorry @indigoxela. We can revert these. I think the added descriptions are probably sufficient anyway.

core/modules/views_ui/css/views_ui.admin.theme.css
indigoxela: This is (currently) necessary, otherwise the preview display in the views admin form breaks.
As soon as we replace the custom views icon images with our svg files, this will need an update, anyway (lots of CSS code will get obsolete then).

core/modules/file/views/views_handler_field_file_icon.inc
quicksketch: Thank you for the feedback, and my apologies for changing a previously discussed string. I reverted this to what it was before.

quicksketch:
Big thank you @indigoxela for taking the initiative on this issue! Merged into 1.x for 1.29.0. :tada:

jenlampton:
I have created https://github.com/backdrop/backdrop-issues/issues/6701 as a follow-up to work on language improvements :)

PR #4780, opened 2024-06-10
Issue #6591: CKE5: Ignore order of attributes when comparing strings with checkValueModified()

### LINKED ISSUE #6591: CKEditor - Creating a node with an inline image does not activate CKEditor 5
## Description of the bug

I first noticed this problem on forums.backdropcms.org, then on client sites that had been upgraded from CKEditor 4. 
https://github.com/backdrop-ops/forum.backdropcms.org/issues/150
I am now able to recreate this problem on a brand new dev site where CKEditor 5 is the default editor from the start. 

When creating a node with an inline image and saving, there seems to be a failure to activate CKEditor 5. 

The next time you edit the node, you will get the following message:
"Activating CKEditor 5 will reformat the content of this field. Review content carefully after activating. [Click to activate editor]"(https://sunday.ddev.site/node/24/edit#).

## Steps To Reproduce

To reproduce the behavior:
1. Open a fresh demo site. 
3. Create a new post with nothing but a title and an image in the body field loaded with CKEditor 5. Save node.
4. Go to edit node and see the following:

image

5. Note: If you save the node and then edit it again, it's fine and there is no message. 

## Expected behavior

I expect to have CKEditor already activated when I come back to edit the node. It does work as expected if there is no image in the body field. 

## Solution (Edited)

[A few comments down](https://github.com/backdrop/backdrop-issues/issues/6591#issuecomment-2156773252) we realize that when an image is initally inserted with CKEditor 5, the 'data-file-id' is listed first, before the 'src'. After activating the editor, the 'src' is moved to be first and everything is fine. 

It seems to me that we either need to change the format of the initial image insert OR make the "checker" more flexible to accept either order.

klonos:
Fixes https://github.com/backdrop/backdrop-issues/issues/6591


core/modules/ckeditor5/js/ckeditor5.js
```diff
@@ -301,10 +301,47 @@
...
-      const formattedBeforeValue = Backdrop.ckeditor5.formatHtml(beforeAttachValue);
-      const formattedAfterValue = Backdrop.ckeditor5.formatHtml(afterAttachValue);
-      return formattedBeforeValue !== formattedAfterValue;
+      // Then run both strings through the same whitespace formatting, using
+      // formatHtml(). Wrap both strings with a temporary <temp> tag, to allow
+      // childNodes (which is used later when comparing the two strings) to work
+      // on them.
+      const formattedBeforeValue = document.createElement('temp');
+      formattedBeforeValue.innerHTML = Backdrop.ckeditor5.formatHtml(beforeAttachValue);
+      const formattedAfterValue = document.createElement('temp');
```
quicksketch: Just as precaution, lets use a defined tag (`<div>`) in case it causes any kind of weird browser behaviors.

```suggestion
      // formatHtml(). Wrap both strings with a temporary <div> tag, to allow
      // childNodes (which is used later when comparing the two strings) to work
      // on them.
      const formattedBeforeValue = document.createElement('div');
      formattedBeforeValue.innerHTML = Backdrop.ckeditor5.formatHtml(beforeAttachValue);
      const formattedAfterValue = document.createElement('div');
```

quicksketch:
Merged into 1.x and 1.28.x. Thank you @klonos!

PR #4784, opened 2024-06-12
Issue #6598: Use `#states` instead of custom JS in the date field settings form

### LINKED ISSUE #6598: [DX] Date field settings: remove custom js and and use `#states` instead
This is one of many things I discovered while working on #6596 that I would like us to fix. Spliting it here, to avoid scope creep in the other issue.

At the top of `date.admin.js` there are a few lines of custom JS that is used to show/hide the timezone handling options if the `Hour` checkbox is not ticked in the granularity options. We should replace that with a `#states` implementation instead.

klonos:
Fixes https://github.com/backdrop/backdrop-issues/issues/6598


klonos:
Replaced by https://github.com/backdrop/backdrop/pull/4790

PR #4789, opened 2024-06-12
Issue 6603. Use firstElementChild instead of firstChild in image_add_svg_attributes

### LINKED ISSUE #6603: image_add_svg_attributes(): Use firstElementChild instead of firstChild
The newly added function `image_add_svg_attributes()`, makes use of  class `DOMDocument` to create attribute markup for SVG images. This line and similar ones is found in this function:

```php
  $doc = new DOMDocument();
  $doc->loadXML($svg_content);
  // ...
         $doc->firstChild->prepend($title);
  // ...
        $doc->firstChild->setAttribute($attribute_name, $attribute_value);
```

While this work, my IDE is telling me that `firstChild` is a `DOMNode`  object, and doesn't define the methods   `setAttribute()`, `removeAttribute()` and `prepend()`. Instead, we should be using `firstElementChild` which will return  a `DOMElement` object, which contain those methods.

EDIT: the current code doesn't throw an error. This is a grammar error that "still works". But eventually this may change. So, regarding testing... there is nothing functional to test AFAIK.

argiepiano:
Fixes https://github.com/backdrop/backdrop-issues/issues/6603

### DIFF
--- a/core/includes/image.inc
+++ b/core/includes/image.inc
@@ -591,11 +591,11 @@ function image_add_svg_attributes($svg_content, array $attributes) {
         if (strlen($attributes['alt'])) {
           $title = $doc->createElement('title');
           $title->textContent = $attributes['alt'];
-          $doc->firstChild->prepend($title);
+          $doc->firstElementChild->prepend($title);
         }
         // Remove any given <title> element if alt is an empty string.
-        elseif ($doc->firstChild->firstChild && $doc->firstChild->firstChild->nodeName === 'title') {
-          $doc->firstChild->removeChild($doc->firstChild->firstChild);
+        elseif ($doc->firstElementChild->firstChild && $doc->firstElementChild->firstChild->nodeName === 'title') {
+          $doc->firstElementChild->removeChild($doc->firstChild->firstChild);
         }
       } catch (DOMException $e) {}
     unset($attributes['alt']);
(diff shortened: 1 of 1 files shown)


klonos:
@argiepiano perhaps change the "fixes" in the description of this PR to point to https://github.com/backdrop/backdrop-issues/issues/6577 instead?

core/includes/image.inc
```diff
@@ -591,11 +591,11 @@ function image_add_svg_attributes($svg_content, array $attributes) {
         if (strlen($attributes['alt'])) {
           $title = $doc->createElement('title');
           $title->textContent = $attributes['alt'];
-          $doc->firstChild->prepend($title);
+          $doc->firstElementChild->prepend($title);
         }
         // Remove any given <title> element if alt is an empty string.
-        elseif ($doc->firstChild->firstChild && $doc->firstChild->firstChild->nodeName === 'title') {
-          $doc->firstChild->removeChild($doc->firstChild->firstChild);
+        elseif ($doc->firstElementChild->firstChild && $doc->firstElementChild->firstChild->nodeName === 'title') {
```
klonos, line 597: Placing an inline comment directly between an `if` and an `else` breaks things. This should work though (placing the inline comment inside the `else`): 
```suggestion
        elseif ($doc->firstElementChild->firstChild && $doc->firstElementChild->firstChild->nodeName === 'title') {
          // Remove any given <title> element if alt is an empty string.
```

PR #4791, opened 2024-06-13
Issue #6606: Bartik: Fix header color gradient not being applied

### LINKED ISSUE #6606: Bartik Theme: Values for header color gradient are not applied
## Description of the bug

When assigning gradient header colors using the Bartik theme, the selected colors are not applied to the header block.  The two colors can be set, but only the first header color is applied, and the gradient does not appear.  

## Steps To Reproduce

To reproduce the behavior:
1. In a vanilla Backdrop site, enable Bartik and set it as the default theme.
2. Go to the Bartik settings form. Notice how both "Header Background Top" and "Header Background Bottom" are set to black.
3. Change the color for "Header background top" to something distinctly different from the default black.

## Actual behavior

In the preview, the entire header changes to the color specified for "Header background top". instead of seeing a gradient.

## Expected behavior

Gradient between the two colors is applied, same as the behavior for this theme in Drupal 7.

## Additional information

- Backdrop CMS version: 1.28.0
- PHP version: 7.4.33
- Database sever (MySQL or MariaDB?) and its version: MariaDB 10.5.2
- Operating System and its version:  Mac 
- Browser(s) and their versions: Safari

klonos:
Fixes https://github.com/backdrop/backdrop-issues/issues/6606

### DIFF
--- a/core/themes/bartik/css/colors.css
+++ b/core/themes/bartik/css/colors.css
@@ -11,7 +11,7 @@ body {
 .l-header {
   color: #fffeff;
   background-color: #000000;
-  background: linear-gradient(top, #000000 0%, #000000 100%);
+  background: linear-gradient(top, #000000 0%, #000001 100%);
 }
 .l-header ul.menu > li > a {
   color: #e3e3e3;
--- a/core/themes/bartik/css/colors.css
+++ b/core/themes/bartik/css/colors.css
@@ -11,7 +11,7 @@ body {
 .l-header {
   color: #fffeff;
   background-color: #000000;
-  background: linear-gradient(top, #000000 0%, #000001 100%);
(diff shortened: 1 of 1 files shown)


core/themes/bartik/css/colors.css
```diff
@@ -11,7 +11,7 @@ body {
 .l-header {
   color: #fffeff;
   background-color: #000000;
-  background-image: linear-gradient(top, #000000 0%, #000001 100%);
+  background: linear-gradient(#000000 0%, #000001 100%);
```
quicksketch:
```suggestion
  background: linear-gradient(to bottom, #000000 0%, #000001 100%);
```

quicksketch:
Merged into Backdrop 1.x and 1.28.x. Thanks!

PR #4794, opened 2024-06-15
Issue #6612: Fix coding standards around docblocks in settings.php

### LINKED ISSUE #6612: [DX] Fix coding standards around docblocks in `settings.php`
There's some inconsistencies and things like missing periods/punctuation. We should fix that according to https://docs.backdropcms.org/doc-standards#general

Specifically this rule:
> - All summaries (first lines of docblocks) must be under 80 characters, start with a capital letter, and end with a period (.). ...

klonos:
Fixes https://github.com/backdrop/backdrop-issues/issues/6612


settings.php
```diff
@@ -368,12 +387,12 @@
...
 /**
- * External access proxy settings:
+ * External access proxy settings.
  *
- * If your site must access the Internet via a web proxy then you can enter
- * the proxy settings here. Currently only basic authentication is supported
- * by using the username and password variables. The proxy_user_agent variable
- * can be set to NULL for proxies that require no User-Agent header or to a
+ * If your site must access the Internet via a web proxy then you can enter the
+ * proxy settings here. Currently only basic authentication is supported  by
```
yorkshire-pudding: Remove extra space

```suggestion
 * proxy settings here. Currently only basic authentication is supported by
```

yorkshire-pudding (changes requested):
One very tiny change needed.
The rest of the changes look good

settings.php
klonos: Thanks @yorkshire-pudding 🙏🏼 ...and sorry for the late reply. Committed 👍🏼

klonos:

yorkshire-pudding (approved):
Looks good.

settings.php
```diff
@@ -187,32 +190,40 @@
 ini_set('session.gc_divisor', 100);
 
 /**
- * Set session lifetime (in seconds), i.e. the time from the user's last visit
- * to the active session may be deleted by the session garbage collector. When
- * a session is deleted, authenticated users are logged out, and the contents
- * of the user's $_SESSION variable is discarded.
+ * Set session lifetime (in seconds).
+ *
+ * I.e. the time from the user's last visit to the active session may be deleted
```
quicksketch, line 195: Can you reword this so we don't capitalize the `i` in `i.e.`? It might be good to simply use the full words rather than an abbreviation here anyway.

"This is the amount of time from the user's last visit, after which the session may be deleted by the session garbage collector."

settings.php
```diff
@@ -187,32 +190,40 @@
...
  */
 ini_set('session.gc_maxlifetime', 200000);
 
 /**
- * Set session cookie lifetime (in seconds), i.e. the time from the session is
- * created to the cookie expires, i.e. when the browser is expected to discard
- * the cookie. The value 0 means "until the browser is closed".
+ * Set session cookie lifetime (in seconds).
+ *
+ * I.e. the time from the session is created to the cookie expires, i.e. when
```
quicksketch, line 205: Same here, let's use full words so we don't awkwardly capitalize the "i". This sentence is even more strange because "i.e." is used _twice_. Let's reword.

"The amount of time that will pass before the session cookie expires and when the browser is expected to discard the cookie."

quicksketch (commented):
@klonos Could you take a look at doing a minor update to these two docblocks?

settings.php
```diff
@@ -71,8 +71,9 @@
  * updates" permission or the site maintenance account (the account that was
  * created during installation), you will need to modify the access check
  * statement below. Change the FALSE to a TRUE to disable the access check.
- * After finishing the upgrade, be sure to open this file again and change the
- * TRUE back to a FALSE!
+ *
+ * IMPORTANT!: After finishing the upgrade, be sure to edit this file again and
+ * change the TRUE back to a FALSE.
```
avpaderno, line 76: Making this sentence more evident is a good change. 👍
I would not use an all-capitalized word, and I would avoid an exclamation point followed by a colon. Either the colon is used (this is my preference) or the exclamation point is used.

I would also avoid to use articles before `TRUE` and `FALSE`. (It is not clear why it should be *the TRUE* and *a FALSE*.)

settings.php
```diff
@@ -86,24 +87,23 @@
  * of the serialized database credentials will be used as a fallback salt.
  *
  * For enhanced security, you may set this variable to a value using the
- * contents of a file outside your docroot that is never saved together
- * with any backups of your Backdrop files and database.
+ * contents of a file outside your docroot that is never saved together with any
+ * backups of your Backdrop files and database.
```
avpaderno, line 91: *The contents of a file* would be a list of paragraphs/sections in a text file. What a file contains, considered globally, is *the file content* or *the content of a file*.  
(I know *contents* is actually used from the setting.php file, but since this part is changed…)

settings.php
avpaderno: I would just use *It is the last time from the user's last visit to the active session may be deleted by the session garbage collector.* There is no need to use *i.e.* as it is clear that part is explaining what *session lifetime* is.

(As a side note, as a non-native English speaker, I find *It is the last time from the user's last visit to the active session may be deleted by the session garbage collector.* difficult to parse, probably because I am not clear what may be deleted. I cannot understand where the subordinate clause starts and where the main clause ends.)

settings.php
```diff
@@ -187,32 +190,40 @@
...
+ * users are logged out, and the contents of the user's $_SESSION variable is
+ * discarded.
  */
 ini_set('session.gc_maxlifetime', 200000);
 
 /**
- * Set session cookie lifetime (in seconds), i.e. the time from the session is
- * created to the cookie expires, i.e. when the browser is expected to discard
- * the cookie. The value 0 means "until the browser is closed".
+ * Set session cookie lifetime (in seconds).
```
avpaderno, line 203: In the other cases, a noun has been used (for example *PHP settings*). If a noun is used in this case too, that should be *Session cookie lifetime* which could be followed by *(in seconds)*.

settings.php
```diff
@@ -224,81 +235,87 @@
...
+ * use a CDN), then you may set the $_SERVER['REMOTE_ADDR'] variable directly in
+ * settings.php. Be aware, however, that it is likely that this would allow IP
  * address spoofing unless more advanced precautions are taken.
  */
 // $settings['reverse_proxy'] = TRUE;
 
 /**
- * Specify every reverse proxy IP address in your environment.
- * This setting is required if $settings['reverse_proxy'] is TRUE.
+ * Reverse proxy IP address list.
```
avpaderno, line 283: If the single words are capitalized, that should be consistently done.
The previous section is *Reverse Proxy Configuration.* but the one before that is *Maintenance theme.*

avpaderno (changes requested):

PR #4797, opened 2024-06-17
Issue #6613: Better error handling in image_add_svg_attributes()

### LINKED ISSUE #6613: Follow-up: evaluate the use of try/catch in image_add_svg_attributes()
## Description of the bug

This is a follow-up to #6577, where it has been discovered, that the indent of this [code block](https://github.com/backdrop/backdrop/blob/1.x/core/includes/image.inc#L591) isn't correct, but also a question/concern has been raised, whether this try/catch actually works as intended there. Or maybe it's unnecessary, or should get wrapped around a bigger part of the function code.

To not block the important bug fix there, this discussion has been postponed, so let's start it here.

## Steps To Reproduce

Hard to tell, as it's rather a code logic issue, nothing user-facing.
Most likely it'll be easier to discuss, when trying out actually broken icon contents (e.g. invalid svg).

Eventually functional tests will need some extension, too.

## Additional information

- Backdrop CMS version: 1.28.x (since icon library addition)
- PHP version: any supported

indigoxela:
Fixes https://github.com/backdrop/backdrop-issues/issues/6613

### DIFF
--- a/core/includes/icon.inc
+++ b/core/includes/icon.inc
@@ -340,4 +340,5 @@ function theme_icon(array $variables) {
       return image_add_svg_attributes($svg_contents, $variables['attributes']);
     }
   }
+  return '';
 }
--- a/core/includes/icon.inc
+++ b/core/includes/icon.inc
@@ -282,7 +282,7 @@ function icon_get_info($icon_name = NULL) {
  *   - attributes: Attributes to be added to the icon itself.
  *
  * @return string
- *   The HTML output.
+ *   The HTML output. An empty string if the icon file is not svg.
  *
  * @since 1.28.0 Function added.
  * @since 1.28.1 The <ellipse>, <line>, <polygon> and <polyline> SVG elements
--- a/core/includes/image.inc
+++ b/core/includes/image.inc
@@ -577,30 +577,39 @@ function image_is_svg($uri) {
  *   which is the accessible mechanism for alternative text within SVGs.
  *
  * @return string
- *   The SVG image contents with the attributes added.
+ *   The SVG image contents with the attributes added. An empty string in case
+ *   of errors.
  *
  * @since 1.28.0 Function added.
  */
 function image_add_svg_attributes($svg_content, array $attributes) {
   $doc = new DOMDocument();
+  libxml_use_internal_errors(TRUE);
   $doc->loadXML($svg_content);
   $svg_tag = $doc->getElementsByTagName('svg')->item(0);
+  $last_error = libxml_get_last_error();
+  libxml_use_internal_errors(FALSE);
+  if (empty($svg_tag)) {
+    $message = $last_error ? trim($last_error->message) : '';
+    watchdog('image', 'Failed to parse SVG content. %message', array(
+      '%message' => $message,
+    ), WATCHDOG_ERROR);
+    return '';
+  }
 
   // Convert the alt attribute to a <title> element.
   if (isset($attributes['alt'])) {
-      try {
-        if (strlen($attributes['alt'])) {
-          $title = $doc->createElement('title');
-          $title->textContent = $attributes['alt'];
-          // Since DOMDocument::prepend() is not available in PHP versions prior
-          // to v8, we are using DOMNode::insertBefore().
-          $svg_tag->insertBefore($title, $svg_tag->firstChild);
-        }
-        // Remove any given <title> element if alt is an empty string.
-        elseif ($svg_tag->firstChild && $svg_tag->firstChild->nodeName === 'title') {
-          $svg_tag->removeChild($svg_tag->firstChild);
-        }
-      } catch (DOMException $e) {}
+    if (strlen($attributes['alt'])) {
+      $title = $doc->createElement('title');
+      $title->textContent = $attributes['alt'];
+      // Since DOMDocument::prepend() is not available in PHP versions prior
+      // to v8, we are using DOMNode::insertBefore().
+      $svg_tag->insertBefore($title, $svg_tag->firstChild);
+    }
+    // Remove any given <title> element if alt is an empty string.
+    elseif ($svg_tag->firstChild && $svg_tag->firstChild->nodeName === 'title') {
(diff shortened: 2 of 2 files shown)


indigoxela:
Hello random buddy... (stat failed for modules/eee_installer_test)

core/includes/image.inc
```diff
@@ -577,30 +577,39 @@ function image_is_svg($uri) {
...
  *
  * @since 1.28.0 Function added.
  */
 function image_add_svg_attributes($svg_content, array $attributes) {
   $doc = new DOMDocument();
+  libxml_use_internal_errors(TRUE);
   $doc->loadXML($svg_content);
   $svg_tag = $doc->getElementsByTagName('svg')->item(0);
+  $last_error = libxml_get_last_error();
+  libxml_use_internal_errors(FALSE);
```
klonos, line 591: Can we please add some comments to explain why we are setting this to `TRUE`, and then to `FALSE`?

core/includes/icon.inc
```diff
@@ -282,7 +282,7 @@ function icon_get_info($icon_name = NULL) {
  *   - attributes: Attributes to be added to the icon itself.
  *
  * @return string
- *   The HTML output.
+ *   The HTML output. An empty string if the icon file is not svg.
```
klonos, line 285: Can we please add another `@since` here about this change?

core/includes/image.inc
```diff
@@ -577,30 +577,39 @@ function image_is_svg($uri) {
  *   which is the accessible mechanism for alternative text within SVGs.
  *
  * @return string
- *   The SVG image contents with the attributes added.
+ *   The SVG image contents with the attributes added. An empty string in case
+ *   of errors.
```
klonos, line 581: Same as previous comment, we should add an `@since` here, to denote the change.

core/includes/image.inc
```diff
@@ -577,30 +577,39 @@ function image_is_svg($uri) {
...
 function image_add_svg_attributes($svg_content, array $attributes) {
   $doc = new DOMDocument();
+  libxml_use_internal_errors(TRUE);
   $doc->loadXML($svg_content);
   $svg_tag = $doc->getElementsByTagName('svg')->item(0);
+  $last_error = libxml_get_last_error();
+  libxml_use_internal_errors(FALSE);
+  if (empty($svg_tag)) {
+    $message = $last_error ? trim($last_error->message) : '';
+    watchdog('image', 'Failed to parse SVG content. %message', array(
```
klonos, line 594: It would be nice to be able to provide more context here, like the name of the file or information about the page/form where the action happened etc. I don't see how though 🤔 ...what does `$message` look like?

core/includes/image.inc
```diff
@@ -577,30 +577,39 @@ function image_is_svg($uri) {
...
+      $title = $doc->createElement('title');
+      $title->textContent = $attributes['alt'];
+      // Since DOMDocument::prepend() is not available in PHP versions prior
+      // to v8, we are using DOMNode::insertBefore().
+      $svg_tag->insertBefore($title, $svg_tag->firstChild);
+    }
+    // Remove any given <title> element if alt is an empty string.
+    elseif ($svg_tag->firstChild && $svg_tag->firstChild->nodeName === 'title') {
+      $svg_tag->removeChild($svg_tag->firstChild);
+    }
```
klonos, line 612: I realize that this is not part of the scope here, but is this whole code here making an assumption that the given svg does not have an existing `<title>`? 🤔

core/includes/image.inc
indigoxela: Are you sure? Should we explain default php function params? :thinking:

https://www.php.net/manual/de/function.libxml-use-internal-errors.php

core/includes/icon.inc
indigoxela: Nothing changed in any API.

It's just that we now actually DO what the doc block claimed. We always return a string. To my understanding that's no use-case for an `@since`

core/includes/image.inc
indigoxela: You want an `@since` for a comment extension?

core/includes/image.inc
indigoxela: We don't have more context inside this function. And we shouldn't make assumptions about it.

Can you please try that out in a testing environment with broken SVG?
You'll get pretty helpful hints - they're provided by that libxml message.
I gave some examples in an issue comment.

core/includes/image.inc
indigoxela: Probably yes - should we consider that? We could...

indigoxela: Currently the only usage of that function is in icon.inc - I'd assume, that icons never ship with a title tag out-of-the-box.
However, this function *could* in theory be used on other SVG files, too. In that case a check for existing title makes sense. But what is the expected outcome then? Overriding or leaving the original untouched?

core/includes/image.inc
avpaderno: I take the comment is referring to `libxml_use_internal_errors(TRUE);` and `libxml_use_internal_errors(FALSE);`.

Comments should not explain what the purpose of a PHP function like `libxml_use_internal_errors()` is, but what the code does. In this case, the code uses `libxml_use_internal_errors(TRUE)` exactly how it is supposed to be used: The code needs to get the last error with `libxml_get_last_error()`. Without `libxml_use_internal_errors(TRUE)`, `libxml_get_last_error()` would not return the last error.

core/includes/icon.inc
avpaderno: The return type has not changed; the description just says the returned string can be empty. I would not add a `@since` tag either.

core/includes/image.inc
indigoxela: @klonos  As this is out of scope here, but a valid concern, which deserves a little more discussion, I've created a follow-up on https://github.com/backdrop/backdrop-issues/issues/6731
Let's continue to discuss desired behavior there.

quicksketch:

PR #4799, opened 2024-06-18
Issue #6614: Add missing views_handler_argument_file_type

### LINKED ISSUE #6614: Views argument handler views_handler_argument_file_type advertised to views in file_views_data() isn't actually defined
## Description of the bug

As the title says, there's a missing views handler.
Views UI will show it, as it's defined here: https://github.com/backdrop/backdrop/blob/1.x/core/modules/file/views/file.views.inc#L82
But it's not usable.

## Steps To Reproduce

1. Go to admin/structure/views
2. Add a new view of type Files
3. Try to add a contextual filter "File: Type" (The type of the file (for example, "audio", "image", "video", etc))

## Actual behavior

Immediately get a "The handler for this item is broken or missing and cannot be used. If a module provided the handler and was disabled, re-enabling the module may restore it. Otherwise, you should probably delete this item."

broken-views-argument-handler

## Expected behavior

Guess... :wink: 

## Additional information

This views handler definition came in (actually did not...) with one of those huge commits. Obviously an oversight. See Issue #2632

indigoxela:
Fixes https://github.com/backdrop/backdrop-issues/issues/6614

### DIFF
--- a/core/modules/file/views/views_handler_argument_file_type.inc
+++ b/core/modules/file/views/views_handler_argument_file_type.inc
@@ -33,10 +33,10 @@ class views_handler_argument_file_type extends views_handler_argument_string {
    *   The File type label, if the type exists.
    */
   protected function fileType($type) {
-    $output = file_type_get_name($type);
-    if (empty($output)) {
+    $file_type_name = file_type_get_name($type);
+    if (empty($file_type_name)) {
       return t('Unknown file type');
     }
-    return check_plain($output);
+    return check_plain($file_type_name);
   }
 }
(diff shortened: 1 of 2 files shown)


core/modules/file/views/views_handler_argument_file_type.inc
```diff
@@ -0,0 +1,42 @@
...
+   * Get the File label (human readable).
+   *
+   * @param string $type
+   *   The views string argument, supposed to be a File type (machine name).
+   *
+   * @return string
+   *   The File type label, if the type exists.
+   */
+  protected function fileType($type) {
+    $output = file_type_get_name($type);
```
klonos: Small nit: can we please change the variable name to something meaningful, like `$file_type` or `$file_type_name`?

Otherwise this LGTM 👍🏼

indigoxela: That's a good idea. :+1:

quicksketch:
Merged into 1.x and 1.28.x. Thanks @indigoxela and @klonos!

PR #4800, opened 2024-06-20
Issue  #6618: Use backdrop_html_class() instead of strtr() in theme_form_element()

### LINKED ISSUE #6618: [DX] Replace use of `strtr()` in `theme_form_element()` with standardized `backdrop_html_class()` instead.
This came up in the PR I created for #1403. The proposed changes were the following:
```diff
if (!empty($element['#type'])) {
-    $attributes['class'][] = 'form-type-' . strtr($element['#type'], '_', '-');
+    $attributes['class'][] = backdrop_html_class('form-type-' . $element['#type']);
```

@herbdool raised a cocern:
> Do we know for sure that this change will result in equivalent classes? This seems like a potentially breaking change.

@klonos:
> Lets see:
> - the implementation of `strtr()` simply replaces underscores with dashes.
> - `backdrop_html_class()` does the following:
>   - passes everything through `backdrop_strtolower()`, and according to https://stackoverflow.com/questions/12533926/are-class-names-in-css-selectors-case-sensitive:
>     - CSS selectors are generally case-insensitive; this includes class and ID selectors 👍🏼 
>     - But [HTML class names are case-sensitive](http://www.w3.org/TR/html4/struct/global.html#h-7.5.2) (see the attribute definition) ... This has not changed in [HTML5](http://www.w3.org/TR/html50/disabled-elements.html#case-sensitivity).1 👎🏼 
>   - passes everything through `backdrop_clean_css_identifier()`, which converts underscores (among other things) to dashes.
> 
> So yes, this is a potentially breaking change, **but only in theory**. For that to happen, someone needs to have declared `'#type'` for an element using uppercase letters, in which case, I believe that there would be errors thrown or the type would simply not be recognized. Since the scope of the `$element['#type']` that is passed to `strtr()` is limited (only element types that have been officially specified in the Form API), I believe that this is safe to do, and it provides standardization and future-proofing.

In the same PR, I also proposed a similar change for the class that is add based on the element name:
```diff
if (!empty($element['#name'])) {
-    $attributes['class'][] = 'form-item-' . strtr($element['#name'], array(' ' => '-', '_' => '-', '[' => '-', ']' => ''));
+    $attributes['class'][] = backdrop_html_class('form-item-' . $element['#name']);
```

This is a follow-up issue to discuss if we should make these changes or not.

klonos:
Fixes https://github.com/backdrop/backdrop-issues/issues/6618


klonos:
Trying to fix `Tugboat project is over quota` ...

PR #4802, opened 2024-06-21
Issue #6607: Project Installer: Do not try to fetch release data for submodules

### LINKED ISSUE #6607: What can we do about - "Could not fetch releases for project " - for dependent sub-modules
## Description of the need

I've created a Recipe Package that basically wraps together a number of other recipes and some config and sample content to deliver the framework for a working website. In this case, it's called Digital Agency.

Several of the recipes have "sample content" sub-modules, to give folks the change to decide if they want the sample content or not. This seems to make sense when installing the recipes on their own. 

However, as part of the Digital Agency recipe package, the sample content should probably be required and in fact, if we don't list these sample content modules as dependencies, they end up loading out of order and causing problem. 

However, because they are sub-modules, I get the following warning during installation.

```
Could not fetch releases for project services_sample_content.
Could not fetch releases for project faq_sample_content.
Could not fetch releases for project portfolio_projects_sample_content.
```

image

## Proposed solution

I'm not sure what the best solution is or if a solution is absolutely necessary. But, it would be nice if the installer were able to look inside of other module for these dependencies OR if I could add something in the .info file to specify that these dependencies are sub-modules and tell the installer where to look for them. 

Just hoping that people ignore the warning. 

## Additional information

https://github.com/backdrop-contrib/digital_agency

The benefits of bundling recipes in this way makes it pretty easy to install a single Recipe Package (module) and get something like this:

image

klonos:
Fixes https://github.com/backdrop/backdrop-issues/issues/6607


klonos:
CSpell nag will be addressed in https://github.com/backdrop/backdrop/pull/4840

PR #4809, opened 2024-06-24
Issue #6640: Fix help text regression in Admin Bar settings

### LINKED ISSUE #6640: [UX] Regression: Help text missing from Admin Bar settings
I've reported this in #4586. Filing this as a follow-up.

We have a small but rather obvious regression with the help text for the various options, where the individual help text items got lost, and the description text for the checkboxes got duplicated in the wrapping fieldset.

Before/after screenshot:
image

klonos:
Fixes https://github.com/backdrop/backdrop-issues/issues/6640

### DIFF
--- a/core/modules/admin_bar/admin_bar.inc
+++ b/core/modules/admin_bar/admin_bar.inc
@@ -855,7 +855,7 @@ function admin_bar_theme_settings() {
     '#title' => t('Enabled components'),
     '#description' => t('These features will be enabled/visible in the administration bar. Untick the boxes to disable/hide them.'),
   );
-  $form['components']['components'] = array(
+  $form['components']['options'] = array(
     '#type' => 'checkboxes',
     '#title' => t('Enabled components'),
     '#title_display' => 'invisible',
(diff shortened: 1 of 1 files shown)


quicksketch:
Looks like this breaks saving the form:
image

All the components end up unchecked and the whole admin bar is empty upon saving.

core/modules/admin_bar/admin_bar.inc
```diff
@@ -855,7 +855,7 @@ function admin_bar_theme_settings() {
     '#title' => t('Enabled components'),
     '#description' => t('These features will be enabled/visible in the administration bar. Untick the boxes to disable/hide them.'),
   );
-  $form['components']['components'] = array(
+  $form['components']['options'] = array(
     '#type' => 'checkboxes',
     '#title' => t('Enabled components'),
     '#title_display' => 'invisible',
```
quicksketch, line 862: One way to solve the problem:

```suggestion
    '#title_display' => 'invisible',
    '#parents' => array('components'),
```

This makes it so that this element considers itself to not have any parents (other than itself), and saves to `$form_state['values']['components']`.

Alternatively, restore the form element key as `'components'`, even though it looks weird having double-nesting like that.

klonos: @quicksketch simply adding `'#tree' => TRUE` to the fieldset fixes things. I've also fixed the `#states` for two of the options that had also regressed + added `#indentation` to them to make it obvious that they are dependent on their respective "parent" items.

klonos:
The usual bloody random test 😞 ...see: https://github.com/backdrop/backdrop-issues/issues/5721

laryn:
Thanks @klonos and @quicksketch -- tested and merged into 1.28.x and 1.x.

PR #4813, opened 2024-06-25
Issue #6625: The page to import translation does not import the translation of strings like "<none>"

### LINKED ISSUE #6625: The page to import translation does not import the translation of strings like "<none>"
To reproduce the error:

- Enable the modules
- Add Italian as new language (It could work with other languages too.)
- Export the Italian translations from https://localize.backdropcms.org/translate/languages/it/export?project=backdropcms&status=0&release=all&search=&author&context=all&limit=10&sid=0&exact_term=
- Import the translations from /admin/config/regional/translate/import
- Click on *Import*

The form will show the following error.

> A translation string was skipped because of disallowed or malformed HTML. See the log for details.

> 12581df46876e1ab63b71d9b5c047735

The error is shown because the translation provided in Italian is `<nessuno>` which is considered as HTML tag that does not exist.

I understand why the error is shown, but at least it should be a warning, and the translated string should be imported.  
The log shows which string causes the error. In the case the error is changed to a warning, and the translated string imported, administrator users could still change the translation, if necessary.

avpaderno:
Fixes https://github.com/backdrop/backdrop-issues/issues/6625


core/modules/image/image.module
```diff
@@ -646,7 +646,7 @@ function image_style_delete($style, $replacement_style_name = '') {
  * Get an array of image styles suitable for using as select list options.
  *
  * @param $include_empty
- *   If TRUE a <none> option will be inserted in the options array.
+ *   If TRUE a " - None -" option will be inserted in the options array.
```
quicksketch:
```suggestion
 *   If TRUE a "- None -" option will be inserted in the options array.
```

quicksketch:
Merged into 1.x and 1.29.x. Thank you @avpaderno!

PR #4815, opened 2024-06-27
Fix telemetry_cron

### LINKED ISSUE #6632: Fixed: telemetry_cron should wait for the interval rather than running with every cron run.
## Description of the bug

While researching the telemetry module for the purposes of adding telemetry to a module, I noticed that no state was being set for telemetry and on closer inspection realised that the function will run whenever cron is invoked.

As state is never set, then `$last_run` will always be `FALSE`, which equates to `0`. `REQUEST_TIME - 0` will always be greater than `$interval`

<details>
<summary>telemetry_cron()</summary>

```php
/**
 * Implements hook_cron().
 */
function telemetry_cron() {
  if (settings_get('telemetry_enabled', TRUE)) {
    $last_run = state_get('telemetry_cron');
    $interval = 86400; // One day in seconds.
    if (REQUEST_TIME - $last_run > $interval) {
      $telemetry_data = telemetry_get_data();
      $destination_url = config_get('telemetry.settings', 'destination');
      $response = backdrop_http_request($destination_url, array(
        'method' => 'POST',
        'data' => backdrop_json_encode($telemetry_data),
        'headers' => array(
          'Content-Type' => 'application/json',
          'X-Site-Key' => _telemetry_site_key(),
        ),
      ));
    }
  }
}
````

</details>

In addition,  the module's `hook_uninstall` function is named incorrectly, but also uses a different state name of 'telemetry_cron_last'
I think this is a more appropriate state name given the other `cron_last` states.

## Steps To Reproduce

To reproduce the behavior:
1. Enable telemetry module
2. Run cron
3. Check database `state` table for anything related to telemetry

## Actual behavior
Nothing in state table for telemetry

## Expected behavior
A state in the state table for when telemetry cron was last run.

## Additional information

- Backdrop CMS version: 1.29.x-dev

yorkshire-pudding:
Fixes https://github.com/backdrop/backdrop-issues/issues/6632

### DIFF
--- a/core/modules/telemetry/telemetry.install
+++ b/core/modules/telemetry/telemetry.install
@@ -7,7 +7,7 @@
 /**
  * Implements hook_uninstall().
  */
-function project_uninstall() {
+function telemetry_uninstall() {
   $state_settings = array(
     'telemetry_cron_last',
   );
--- a/core/modules/telemetry/telemetry.module
+++ b/core/modules/telemetry/telemetry.module
@@ -101,7 +101,7 @@ function telemetry_get_data() {
  */
 function telemetry_cron() {
   if (settings_get('telemetry_enabled', TRUE)) {
-    $last_run = state_get('telemetry_cron');
+    $last_run = state_get('telemetry_cron_last');
     $interval = 86400; // One day in seconds.
     if (REQUEST_TIME - $last_run > $interval) {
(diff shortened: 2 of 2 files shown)


laryn:
Thanks @yorkshire-pudding -- merged into 1.28.x and 1.x.

PR #4816, opened 2024-06-28
Issue 6633. Add way to indicate a D7 upgrade during update process

### LINKED ISSUE #6633: Add a "Drupal 7 upgrade" state.
When upgrading a site from D7, I'd like to add a way to tell modules we are doing a D7 upgrade, as opposed to a normal Backdrop update.

Having a way to indicate D7 upgrades may help modules avoid fatal errors that would not typically occur during a normal update process. 

One of the concrete reasons for this request is that there are contrib modules that act upon cache clearing. During the D7 upgrade, `system_update_1079()` runs very early in the process, before (for example) text filters are set up. Ubercart/Rules, for example, try to set up some default, "in-code" rules during this cache clearing (see the hook `rules_flush_caches()`, which install default rules when needed). Ubercart in particular makes a call to `filter_default_formats()` to obtain the default text filter. Unfortunately, at this point of the upgrade process, filters have not yet been set up, resulting in a fatal error (see https://github.com/backdrop-contrib/ubercart/issues/484).

There have been other instances of problems brought up in Zulip (see https://backdrop.zulipchat.com/#narrow/stream/218635-Backdrop/topic/Paragraphs/near/447608155).

## How this could be done
One way to indicate we are in a D7 upgrade is to set a `state` during the upgrade process. For example, an ideal place would be during the switch/case in `update.php`.

Right after clicking the button "Apply pending updates", we are taken to the switch/case in line 556 of `update.php`. This is when the update process actually starts.  Up until line 564 (`update_batch($_POST['start'], $redirect_url, $batch_url)`), the schema version for system is still > 7000 when we are running a D7 upgrade. The function `update_batch()` sets the schema to `999`, so that the system update hooks can run.

I'm proposing that we use `state_set('d7_upgrade', TRUE);` if the schema is > 7000 right before this point, to indicate that we are dealing with a D7 upgrade. We can then remove that state during the switch/case 'results' in line 569.

I'll submit a PR.

argiepiano:
Fixes https://github.com/backdrop/backdrop-issues/issues/6633

### DIFF
--- a/core/update.php
+++ b/core/update.php
@@ -561,6 +561,10 @@ if (update_access_allowed()) {
         // update.php correctly by default.
         $batch_url = $base_root . backdrop_current_script_url();
         $redirect_url = $base_root . backdrop_current_script_url(array('op' => 'results'));
+        // Set a state indicating we are upgrading a Drupal 7 site.
+        if (backdrop_get_installed_schema_version('system') > 7000) {
+          state_set('update_d7_upgrade', TRUE);
+        }
         update_batch($_POST['start'], $redirect_url, $batch_url);
         break;
       }
@@ -570,6 +574,8 @@ if (update_access_allowed()) {
       break;
 
     case 'results':
+      // Remove the state indicating a Drupal 7 upgrade.
+      state_del('update_d7_upgrade');
       $output = update_results_page();
       break;
 


quicksketch:
Merged into 1.x and 1.28.x. Thank you @argiepiano!

PR #4818, opened 2024-06-28
[In Progress] Add dark mode to seven

### LINKED ISSUE #4778: Native dark mode support
Dark mode has become a popular feature for all electronic mediums, from desktops to mobile apps to web.  Modern browsers broadly support the [preferred-color-scheme](https://developer.mozilla.org/en-US/docs/Web/CSS/@media/prefers-color-scheme) `@media` selector allowing a site to automatically match the user's device preference, and many websites allow a user to forcibly set dark mode, such as the [StackExchange](https://stackexchange.com) family of sites.

However, it can't be done by the themes alone.  The `backdrop_get_logo()` function returns a single graphic without consideration for the light or dark mode.  The Backdrop logo, for example, works well with light mode pages, less so with dark mode.

To fully support Dark Mode in Backdrop, the following items would need to be included:

- [ ] Support for light/dark logos and automatic inclusion of the right logo in theme functions.
- [ ] Support in one or more stock themes, in particular Basis and Seven.
- [ ] A system-level, user-overridable setting (similar to timezones) for dark mode allowing for "light", "dark", or "auto" (follow device preference) that modules and themes can use to include the appropriate CSS.
- [ ] CSS cache support that won't send cached dark mode CSS to a light mode users.
- [ ] Color module support
- [ ] Backwards compatibility for themes and modules that have no notion of "dark mode."

See also #3990 and #3991.

wesruv:
Fixes https://github.com/backdrop/backdrop-issues/issues/4778


laryn:
@wesruv Looking like a great start! A few comments after a quick spin. I see some contrast issues:
 - blue link text on dark background (e.g. main admin content listing)
 - some strangeness in the main admin "manage files" listing
 - CKEditor 5 dark mode (might be a bigger project but [this override file](https://github.com/backdrop-contrib/gin/blob/1.x-3.x/styles/components/ckeditor5.scss) could be of interest)
 - layouts edit block pages
 - layouts add flexible template interface

docwilmot:
How to test this?

wesruv:
Thank you @laryn! I'll take a look at the feedback soon.

@docwilmot at this point, if you're interested in trying it out I'd like eyeballs and thoughts!

The PR env is always good; or a non-client facing site or local dev env would be good candidates. I'd copy the seven theme from this branch and paste it over the one in core on your env.

It's got a while before I think it's ready to merge, but I thought getting feedback early and often may be helpful on this one.

If you're asking how to see dark mode:
If your OS is set to prefer dark mode, it should "just work".
Otherwise, browser dev tools have a way to fake that setting, in Chrome it's in the kebab menu with the Console pane, the way I get there is open dev tools > Press Escape (which toggles the console pane) > Kebab menu > Rendering > Scroll down to prefers-color-scheme
image

There might also be browser preferences for this?

klonos:
CSpell nag will be addressed in https://github.com/backdrop/backdrop/pull/4840

PR #4819, opened 2024-06-29
Issue #6631: Add playsinline attribute to video display settings

### LINKED ISSUE #6631: Video: Autoplay requires attribute playsinline in some environments
## Description of the bug

This has been brought up in our Zulip chat: to get `autoplay` working in iOS (and possibly Safari) it's required that the `playsinline` attribute is also present.
However, that attribute is not available via field settings.

## Steps To Reproduce

Tricky for people without such a device (like me). :wink: 

## Actual behavior

According to [documentation on developer.apple.com](https://developer.apple.com/documentation/webkit/delivering_video_content_for_safari/#3030251):

> Video elements that include `<video autoplay>` play automatically when the video loads in Safari on macOS and iOS, only if those elements also include the playsinline attribute.

Which means the setting's pointless without the additional attribute.

## Additional information

Settings for autoplay are handled in core/modules/file/file.theme.inc and core/modules/file/file.field.inc.

More on both attributes on [developer.mozilla.org](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/video#playsinline)

As far as I can tell, Safari's the only browser requiring this attribute combination. However, reading the docs, I think it's a good idea to add this attribute setting, anyway.

indigoxela:
Fixes https://github.com/backdrop/backdrop-issues/issues/6631

### DIFF
--- a/core/modules/file/config/file_display.video.json
+++ b/core/modules/file/config/file_display.video.json
@@ -8,6 +8,7 @@
             "file_field_file_video": {
                 "controls": 1,
                 "autoplay": 0,
+                "playsinline": 0,
                 "loop": 0,
                 "muted": 0,
                 "width": "",
@@ -23,6 +24,7 @@
             "file_field_file_video" : {
                 "controls": 1,
                 "autoplay": 0,
+                "playsinline": 0,
                 "loop": 0,
                 "muted": 0,
                 "width": "",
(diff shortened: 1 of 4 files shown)


indigoxela:
Closing as it's insufficient from UX perspective.

PR #4821, opened 2024-06-30
Issue #6637: Replace `$form_state['clicked_button']` with `$form_state['triggering_element']`

### LINKED ISSUE #6637: [DX] Remove usages of deprecated `$form_state['clicked_button']` from core (so that contrib doesn't use it as example)
`$form_state['clicked_button']` was removed in favor of `$form_state['triggering_element']` even in D7. See https://git.drupalcode.org/project/drupal/-/blob/7.x/includes/form.inc?ref_type=heads#L257

```php
...
 *   - triggering_element: (read-only) The form element that triggered
 *     submission. This is the same as the deprecated
 *     $form_state['clicked_button']. It is the element that caused submission,
 *     which may or may not be a button (in the case of Ajax forms). This key is
 *     often used to distinguish between various buttons in a submit handler,
 *     and is also used in Ajax handlers.
 *   - clicked_button: Deprecated. Use triggering_element instead.
...
```

We also have a `@todo` note in `form_builder()` to remove the backwards compatibility workaround in Backdrop 2.0. See https://github.com/backdrop/backdrop/blob/1.x/core/includes/form.inc#L2056:
```php
      // @todo Legacy support. Remove in Backdrop 2.x.
      $form_state['clicked_button'] = $form_state['triggering_element'];
```

This task here is NOT to remove the backwards-compatible workaround itself (which would be a 2.x tasks), rather than to make sure that we are not using it in core. A quick grep shows 40+ instances of that currently.

It would be a nice bonus if there was a way to log a warning, similar to `watchdog_deprecated_function()`, so that contrib and custom code that is using this can be notified to switch to `triggering_element` instead.

klonos:
Fixes https://github.com/backdrop/backdrop-issues/issues/6637


core/includes/form.inc
```diff
@@ -2055,6 +2055,9 @@ function form_builder($form_id, &$element, &$form_state) {
 
       // @todo Legacy support. Remove in Backdrop 2.x.
       $form_state['clicked_button'] = $form_state['triggering_element'];
+
+      // Tste tyop: tirger.
+      $master = $form_state['clicked_button'];
```
bugfolder, line 2060: Is this leftover testing code?

PR #4824, opened 2024-07-08
Issue #6647: Ensure SVGZ is served with correct encoding

### LINKED ISSUE #6647: Ensure SVGZ is served with correct encoding
This is the same as https://www.drupal.org/project/drupal/issues/2092245 (slated for backporting to D7).

> https://en.wikipedia.org/wiki/Scalable_Vector_Graphics#Compression
> 
> > SVG images, being XML, contain many repeated fragments of text, so they are well suited for lossless data compression algorithms. When an SVG image has been compressed with the industry standard gzip algorithm, it is referred to as an "SVGZ" image and uses the corresponding .svgz filename extension. Conforming SVG 1.1 viewers will display compressed images.[8] An SVGZ file is typically 20 to 50 percent of the original size.[9] W3C provides SVGZ files to test for conformance.[10]
> 
> Every web browser that supports SVG supports SVGZ as it is just a standard gzip HTTP encoding.
> 
> Having this in core means that every contrib developer can have confidence that SVGZ will work.

> ## Problem/Motivation
> Drupal is not serving SVGZ files with the correct encoding.
> 
> ## Proposed resolution
> All that is required is for these two lines to be added to .htaccess as described here: http://kaioa.com/node/45
> ```
> AddType image/svg+xml svg svgz
> AddEncoding gzip svgz
> ```
>
> nginx can also be configured to serve SVGZ correctly. At my request, Pantheon have already updated their nginx configuration to support it and I have confirmed that it works.
> 
> I have also modified web.config for IIS as described here: http://forums.iis.net/t/1175276.aspx and confirmed that it works.

klonos:
Fixes https://github.com/backdrop/backdrop-issues/issues/6647

Backports https://www.drupal.org/project/drupal/issues/2092245 | https://git.drupalcode.org/project/drupal/-/commit/84c2ccf468296b5fce156ecb562768e98a7de113


klonos:
https://github.com/backdrop/backdrop-issues/issues/5721 🙄

PR #4838, opened 2024-07-27
Issue #6666: Fix admin bar collapse with jQuery 3 and no extra items in bar.

### LINKED ISSUE #6666: Fix Admin bar collapse with jQuery 3 for certain settings and narrow windows.
## Description of the bug

Normally with a narrow window (like a phone screen), the admin bar collapses into a single hamburgered drop-down menu. However, if the site is set to use jQuery 3 (which is now the default) and one turns off the extra items that usually appear to the right of the admin bar, then that collapse fails, and the admin bar wraps, thereby potentially covering up other stuff (depending on the theme).

## Steps To Reproduce

* On a plain Backdrop site, enable the Admin Bar module and log in as a user with sufficient privileges to see the admin bar.
* Go to admin/config/administration/admin-bar and turn off "Page links", "User counts", "Account links", and "Language switcher", like this:
  image
* Go to admin/config/development/jquery (which would normally be set to jQuery 3), resize the window to 500px or less, and refresh the page. You will see something like this, with all admin menu menus visible and wrapping. (Note that the breadcrumb got covered up.)
  image
* Now select jQuery 1 and save. You'll see what you're supposed to see, which is this:
  image

## Additional information

This happens because of some code in `admin_bar.js`, specifically, the function defined as `var adjustItems`:

```
  var adjustItems = function () {
    // Expand the menu items to their full width to check their size.
    $menu.removeClass('dropdown').addClass('top-level');
    $extra.removeClass('dropdown').addClass('top-level');

    $adminBar.trigger('beforeResize');

    menuWidth = $menu.width();
    extraWidth = $extra.width();

    // Available width is anything except the menus that may be collapsed.
    availableWidth = $adminBar.width();
    $adminBar.children().children().not($menu).not($extra).each(function() {
      availableWidth -= $(this).width();
    });

    // Collapse the extra items first if needed.
    if (availableWidth - menuWidth - extraWidth < 20) {
      $extra.addClass('dropdown').removeClass('top-level');
      extraWidth = $extra.width();
      ...
```

When there are no extra items at the right , `$extra` is empty. In jQuery 1, `extraWidth` ends up as `null`, which typecasts to zero, so the expression `availableWidth - menuWidth - extraWidth` still evaluates as a number.

However, in jQuery 3, `$extraWidth` ends up undefined, which makes that mathematical expression a `NaN`, and so the conditional never evaluates to TRUE.

So the solution would appear to be to handle the case where `$extra` happens to be empty (in a way that works with both jQuery versions).

bugfolder:
Fixes https://github.com/backdrop/backdrop-issues/issues/6666.

### DIFF
--- a/core/modules/admin_bar/js/admin_bar.js
+++ b/core/modules/admin_bar/js/admin_bar.js
@@ -198,7 +198,6 @@ Backdrop.adminBar.behaviors.resizeCollapse = function (context, settings, $admin
     $adminBar.trigger('beforeResize');
 
     menuWidth = $menu.width();
-var foo = $extra.length;
     if ($extra.length) {
       extraWidth = $extra.width();
     }
--- a/core/modules/admin_bar/js/admin_bar.js
+++ b/core/modules/admin_bar/js/admin_bar.js
@@ -197,7 +197,12 @@ Backdrop.adminBar.behaviors.resizeCollapse = function (context, settings, $admin
 
     $adminBar.trigger('beforeResize');
 
-    menuWidth = $menu.width();
+    if ($menu.length) {
+      menuWidth = $menu.width();
+    }
+    else {
+      menuWidth = 0;
+    }
     if ($extra.length) {
       extraWidth = $extra.width();
     }
--- a/core/modules/admin_bar/js/admin_bar.js
+++ b/core/modules/admin_bar/js/admin_bar.js
@@ -197,18 +197,8 @@ Backdrop.adminBar.behaviors.resizeCollapse = function (context, settings, $admin
 
     $adminBar.trigger('beforeResize');
 
-    if ($menu.length) {
-      menuWidth = $menu.width();
-    }
-    else {
-      menuWidth = 0;
-    }
-    if ($extra.length) {
-      extraWidth = $extra.width();
-    }
-    else {
-      extraWidth = 0;
-    }
+    menuWidth = $menu.width() || 0;
+    extraWidth = $extra.width() || 0;
 
     // Available width is anything except the menus that may be collapsed.
     availableWidth = $adminBar.width();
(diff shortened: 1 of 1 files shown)


core/modules/admin_bar/js/admin_bar.js
```diff
@@ -198,7 +198,13 @@ Backdrop.adminBar.behaviors.resizeCollapse = function (context, settings, $admin
     $adminBar.trigger('beforeResize');
 
     menuWidth = $menu.width();
-    extraWidth = $extra.width();
+var foo = $extra.length;
```
klonos: Debug that needs to be removed. Right?

bugfolder: Right.

core/modules/admin_bar/js/admin_bar.js
```diff
@@ -197,8 +197,18 @@ Backdrop.adminBar.behaviors.resizeCollapse = function (context, settings, $admin
...
+      menuWidth = $menu.width();
+    }
+    else {
+      menuWidth = 0;
+    }
+    if ($extra.length) {
+      extraWidth = $extra.width();
+    }
+    else {
+      extraWidth = 0;
```
argiepiano: I'm going to suggest a more succinct and idiomatic JS way to do this - you see this very often, including in Backdrop js files. Replace all 3 new `if` blocks with this:
```js
extraWidth = $extra.width() || 0;
```
(Of course change the variable names when dealing with $menu). When `$extra.width()` is `undefined`, then the `||` (or) will assign the second part (0).

This can be applied to the other ifs here.

bugfolder: Slick. Will do.

klonos (approved):
LGTM 👍🏼

quicksketch:
Thanks @bugfolder, @argiepiano, and @klonos! Merged into 1.x and 1.28.x.

PR #4844, opened 2024-08-08
Issue #6678: Wrong link to classify file types if installed in subdirectory

### LINKED ISSUE #6678: Wrong link to classify file types if installed in subdirectory
## Description of the bug

The link to "Classify file types" on the Status page is absolute and therefore does not work if Backdrop is installed in a subdirectory.

## Steps To Reproduce

After an upgrade from Drupal 7, files are not classified and the Status page displays "File types need to be classified. (Classify file types)" with a link to /admin/structure/file-types/classify.

## Actual behavior

That link is resulting in error 404 if BD is install in a subdirectory.

## Expected behavior

The link should work in any situation, it should be passed through url().

## Additional information

I will create a PR.

AlexHoebart-ICPDR:
Fixes https://github.com/backdrop/backdrop-issues/issues/6678

### DIFF
--- a/core/modules/file/file.install
+++ b/core/modules/file/file.install
@@ -279,7 +279,7 @@ function file_requirements($phase) {
     if ($file_needs_classification) {
       $requirements['file_needs_classification'] = array(
         'title' => t('File Type Classification'),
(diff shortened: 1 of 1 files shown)


quicksketch:
Merged into 1.x and 1.28.x. Thanks @AlexHoebart-ICPDR!

PR #4864, opened 2024-09-08
Issue 6707. Fix end-date token name.

### LINKED ISSUE #6707: Misnamed date token 'to-date'
## Description of the bug

Date fields with an end date have token support. Currently, the formats shown in the token tree are `date` (for starting date) and `to-date` for end date, as in `[node:field_my_date:date:long]` and `[node:field_my_date:to-date:long]`.

The `to-date` part actually doesn't work. If you look at the date implementation of `hook_tokens` the expected part is `end-date`, not `to-date`. See [these lines](https://github.com/backdrop/backdrop/blob/c61a9f5064d61af032bfc6391aa69c049174a2a6/core/modules/date/date.tokens.inc#L19-L31) and [this line](https://github.com/backdrop/backdrop/blob/c61a9f5064d61af032bfc6391aa69c049174a2a6/core/modules/date/date.tokens.inc#L151)

## Steps To Reproduce
1. Install and enable Token filter. Go to `admin/config/content/formats/filtered_html` and be sure to enable the Token filter and Save
2. Go to Content type and edit the `Page` type. Add a field called My date in a content type. Be sure to "Collect end date" in Global settings
3. In the Body field, enter the following two tokens:
`With to-date: [node:field_my_date:to-date:raw]`
`With end-date: [node:field_my_date:end-date:raw]`
4. Save the node. Verify that only the one with `end-date` is being replaced. The other is blank, since there is no replacement.

You can verify that the token tree shows `to-date` by going to Configure in the Page content type (after having added the date field as above), clicking URL alias, and then click the link to show the tree.
Screen Shot 2024-09-08 at 2 45 49 PM

## Actual behavior

`to-date` is documented, but it doesn't work. `end-date` is not documented, but it works. 

## Additional information

Since `to-date` has never worked, I propose to change the documentation of the token tree to `end-date` instead of changing the existing code for the token replacement.

argiepiano:
Fixes https://github.com/backdrop/backdrop-issues/issues/6707

### DIFF
--- a/core/modules/date/date.tokens.inc
+++ b/core/modules/date/date.tokens.inc
@@ -44,7 +44,7 @@ function date_token_info_alter(&$info) {
       'date',
       'datetime',
       'datestamp',
-      ))) {
+    ))) {
       foreach (array_keys($field['bundles']) as $token_type) {
         // If a token already exists for this field, then don't add it.
         if (isset($info['tokens'][$token_type][$field_name])) {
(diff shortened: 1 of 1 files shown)


quicksketch:
Thanks @argiepiano! Merged into 1.x and 1.28.x.

PR #4865, opened 2024-09-09
Issue #6701: Improve form item label text for new SVG icon views field handler

### LINKED ISSUE #6701: [UX] Improve form item label text for new SVG icon views field handler
## Description of the bug

We need to decide on improved text for labels in https://github.com/backdrop/backdrop-issues/issues/6588 

Current text:
image

## Curent recommendations:

`Embedded file with  <svg> tag.`
`Embedded SVG images are usually more efficient and easier to style.`

`Referenced file with  tag.`
`Referenced SVG images can be more efficient when showing the same icon many times.`

`Display text alternative`
`For icons that are decorative only, alternative text should be omitted. A text alternative is used to describe the icon for assistive tools like screen readers.`

## Other Recent recommendations:

`Inline <svg> tag`
`Inline SVG images are usually more efficient and easier to style.`

`Regular  tag`
`Regular image tags can be more efficient when showing the same icon many times.`

`Add alternative text`
`If alternative text is omitted, this icon will be marked as decorative only.`

indigoxela:
Fixes https://github.com/backdrop/backdrop-issues/issues/6701

### DIFF
--- a/core/modules/file/views/views_handler_field_file_icon.inc
+++ b/core/modules/file/views/views_handler_field_file_icon.inc
@@ -37,18 +37,18 @@ class views_handler_field_file_icon extends views_handler_field {
       '#type' => 'radios',
       '#title' => t('Display type'),
       '#options' => array(
-        'svg' => t('Inline SVG markup'),
-        'img' => t('Regular images'),
+        'svg' => t('Embedded SVG markup'),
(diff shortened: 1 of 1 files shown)


quicksketch:
Merged into 1.x for 1.29.0. Thank you @indigoxela!

PR #4866, opened 2024-09-10
Issue #6708: Combine sort and click sort in views query for table display

### LINKED ISSUE #6708: Add the option to append original views sort to table column sort.
## Description of the bug

When creating a table view, one has the option to also add click sorting - whenever the table header is clicked, data should be sorted by that column. And that works fine but...

When a table column has multiple identical values (like same numbers), the sorting (and pagination) drifts into "random".
The more identical values exist, the more likely it gets, that pagination will show "duplicates" when navigating to the next page, or completely hide items.

## Steps To Reproduce

I belief, the project usage page on B-org is a good example for that phenomenon: https://backdropcms.org/project/usage?page=9

(I discovered the problem in other views, though.)

## Actual behavior

Sticking with above project table example:

There are lots of projects with usage "0". If you navigate between page 10 and 9 you'll notice, that projects show up multiple times (note: once per page, but again on the next/prev page) and/or in random order.

## Expected behavior

The ideal solution for me would be the ability to add a second (not exposed) sort criterion - regularly via views admin UI. For example node ID.
But it should get added to the views query _after_ the click criterion. So multiple results with same values get ordered consistently - also over multiple pages (with pagination).

## Additional information

AFAIK there's currently no workaround for that, other than implementing hook_views_query_alter() - adding another sort criterion via `$query->orderby`, so items with the same value don't get a seemingly random order (it's actually NO order).

In views there's no way to combine regular "Sort criteria" with the table header click sort. As soon as click sort is enabled and set to override normal sort, all sort criteria are discarded. You can verify by enabling the query display in views admin UI.

See also this Drupal issue: https://www.drupal.org/project/drupal/issues/3059961

indigoxela:
Fixes https://github.com/backdrop/backdrop-issues/issues/6708

Based on the Drupal proposal by mdolnik, but modified/extended.

### DIFF
--- a/core/modules/views/plugins/views_plugin_style_table.inc
+++ b/core/modules/views/plugins/views_plugin_style_table.inc
@@ -30,6 +30,7 @@ class views_plugin_style_table extends views_plugin_style {
     $options['default'] = array('default' => '');
     $options['info'] = array('default' => array());
     $options['override'] = array('default' => TRUE, 'bool' => TRUE);
+    $options['append_orig_sort'] = array('default' => FALSE, 'bool' => TRUE);
     $options['sticky'] = array('default' => FALSE, 'bool' => TRUE);
     $options['order'] = array('default' => 'asc');
     $options['caption'] = array('default' => '', 'translatable' => TRUE);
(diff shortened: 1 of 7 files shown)


core/modules/views/plugins/views_plugin_style_table.inc
```diff
@@ -172,6 +178,19 @@ class views_plugin_style_table extends views_plugin_style {
       '#default_value' => !empty($this->options['override']),
     );
 
+    $form['append_orig_sort'] = array(
```
quicksketch: We try to avoid abbreviations, could you rename this to `append_original_sort`?

quicksketch:
Merged into 1.x for 1.32.0. Thank you @indigoxela!

PR #4867, opened 2024-09-10
Issue #5345: Initially expand global settings fieldset in field UI

### LINKED ISSUE #5345: Expand the "Global settings" fieldset when creating new fields
**Description of the need**

Whenever one creates a new field, the "Global settings" section, which contains the field cardinality setting, is easily overlooked.

So, I attach a field to a node type, then try to create a node of that type, but realize, that I forgot to set the cardinality to a value higher than "1".

Then I go back to the field settings and update the cardinality value, then go back to create a node. ... Some unnecessary clicks in my opinion. I bet, I'm not the only one.

**Proposed solution**

Expand the fieldset initially, so it's getting a little more attention, when creating fields.

Hook form alter... :wink: 

There's the fieldset: https://github.com/backdrop/backdrop/blob/1.x/core/modules/field_ui/field_ui.admin.inc#L1994

indigoxela:
Fixes https://github.com/backdrop/backdrop-issues/issues/5345

### DIFF
--- a/core/modules/field_ui/field_ui.admin.inc
+++ b/core/modules/field_ui/field_ui.admin.inc
@@ -1991,7 +1991,7 @@ function field_ui_field_edit_form($form, &$form_state, $instance) {
   $form['field'] = array(
     '#type' => 'fieldset',
     '#collapsible' => TRUE,
-    '#collapsed' => TRUE,
+    '#collapsed' => empty($_GET['destinations']),
     '#title' => t('Global settings'),
     '#description' => $description,
     '#tree' => TRUE,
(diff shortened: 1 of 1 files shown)


core/modules/field_ui/field_ui.admin.inc
```diff
@@ -1991,7 +1991,7 @@ function field_ui_field_edit_form($form, &$form_state, $instance) {
   $form['field'] = array(
     '#type' => 'fieldset',
     '#collapsible' => TRUE,
-    '#collapsed' => TRUE,
+    '#collapsed' => empty($_GET['destinations']),
```
quicksketch, line 1995:
```suggestion
    // Open when first creating a field and close when editing.
    '#collapsed' => empty($_GET['destinations']),
```

quicksketch:

PR #4868, opened 2024-09-12
Security hardening for private file failed uploads.

### LINKED ISSUE #6709: [SR] Private files security hardening
Backdrop doesn't sufficiently ensure private destination folders exist prior to saving files. If the private directory doesn't exist, Backdrop places the private file in a publicly accessible directory.

I would like to propose a hardening to Backdrop that will delete the file if the copy fails, rather than leaving it in a public location.

STR:
* Install backdrop
* Configure private files, set destination to `private` for example
* Confirm that the `private` directory was created, containing the `.htaccess` file
* Manually delete the whole `private` directory
* Proceed to the manage files overview page `admin/content/files`
* Add a new file
* On step 2, set the destination to `Private local files served by Backdrop`
* See the error `The specified file public://license_0.txt could not be moved/copied because the destination directory is not properly configured. This may be caused by a problem with file or directory permissions. More information is available in the system log. `
* See that the private file remains in the public files location

This has been previously reviewed by the security team, and determined not to be a vulnerability because it can not be triggered by using the UI alone (one would also need access to the server). It is, however, not ideal, and should be improved.

Also related: https://www.drupal.org/sa-contrib-2024-040

jenlampton:
Fixes https://github.com/backdrop/backdrop-issues/issues/6709

### DIFF
--- a/core/includes/file.inc
+++ b/core/includes/file.inc
@@ -514,8 +514,8 @@ function file_save_htaccess($directory, $private = TRUE, $force_overwrite = FALS
 
   $htaccess_lines = file_htaccess_lines($private);
 
-  // Write the .htaccess file.
-  if (file_put_contents($htaccess_path, $htaccess_lines)) {
+  // Write the .htaccess file. Suppress any PHP error and write to watchdog.
+  if (@file_put_contents($htaccess_path, $htaccess_lines)) {
     backdrop_chmod($htaccess_path, 0444);
   }
   else {
--- a/core/modules/file/file.pages.inc
+++ b/core/modules/file/file.pages.inc
@@ -336,6 +336,15 @@ function file_add_form_submit($form, &$form_state) {
   if ($save) {
     $file = file_load($form_state['storage']['upload']);
     if ($file) {
+      $file->type = $form_state['storage']['type'];
+      $file->display = TRUE;
+
+      // Save the form fields.
+      // Keep in mind that the values for the Field API fields must be in
+      // $form_state['values'] and not in ['storage']. This is true as long as
+      // the fields are on the last page of the multi step form.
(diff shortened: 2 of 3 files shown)


core/modules/file/file.pages.inc
```diff
@@ -339,31 +339,48 @@ function file_add_form_submit($form, &$form_state) {
...
-      $form_state['file'] = $file;
-      backdrop_set_message(t('@type %name was uploaded.', array('@type' => file_type_get_name($file), '%name' => $file->filename)));
+
+      $substitutions = array(
+        '@type' => file_type_get_name($file),
+        '%name' => $file->filename,
+      );
+
+      // Only save the file if the move completed.
+      if ($moved_file){
```
indigoxela: $moved_file can be undefined.

See line 345, where it's defined inside an if clause.

core/includes/file.inc
```diff
@@ -514,8 +514,8 @@ function file_save_htaccess($directory, $private = TRUE, $force_overwrite = FALS
 
   $htaccess_lines = file_htaccess_lines($private);
 
-  // Write the .htaccess file.
-  if (file_put_contents($htaccess_path, $htaccess_lines)) {
+  // Write the .htaccess file. Suppress any PHP error and write to watchdog.
+  if (@file_put_contents($htaccess_path, $htaccess_lines)) {
```
quicksketch, line 518: This change is needed because any PHP error/warning will cause tests to fail. We shouldn't be allowing a raw PHP error to be surfaced to end-users anyway. We already have watchdog logging in place in the `else {}` statement that follows this.

quicksketch:
Merged into 1.x and 1.28.x. Thank you @jenlampton!

PR #4847, opened 2024-08-18
Issue #6687: Fix admin bar styles when outside-right is applied

wesruv:
Add styles to make text and dropdowns distinct

Fixes https://github.com/backdrop/backdrop-issues/issues/6687

PR #4869, opened 2024-09-12
Issue #6687: Fix admin bar styles when outside-right is applied

### LINKED ISSUE #6687: [UX] Fix `outside-right` styles so text doesn't overlap
As reported by @izmeez and further tracked down by @avpaderno, at tablet-ish screen sizes the admin menu fly outs too close to the right side of the screen get text overlap.

Screenshot of Date and Time Formats Flyout overlapping dropdown

I found the issue is with the `outside-right` styling, will be sending a PR shortly

## Steps To Reproduce

To reproduce the behavior:
1. Login so the admin bar is visible
2. Adjust screen width to roughly 1250px
3. Hover over Configuration > Regional and Language > Date and time format
4. This should cause the sub-menu to appear on top of the parent menu item, overlapping it. It can be difficult to distinguish the overlapping sub-menu from the parent menu.

jenlampton:
Fixes https://github.com/backdrop/backdrop-issues/issues/6687
Replaces: https://github.com/backdrop/backdrop/pull/4847

### DIFF
--- a/core/modules/admin_bar/css/admin_bar.css
+++ b/core/modules/admin_bar/css/admin_bar.css
@@ -210,11 +210,18 @@
   margin: calc(-1px - var(--admin-bar-height)) 0 0 15em; /* LTR */
 }
 #admin-bar .dropdown li li.expandable.outside-right ul {
+  z-index: 1;
   margin-left: -15em; /* LTR */
+  box-shadow:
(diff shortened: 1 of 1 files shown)


quicksketch:
Merged into 1.x for 1.29.0. Thank you @jenlampton and @wesruv!

PR #4872, opened 2024-09-15
Issue #6705: Simplify layout condition interfaces.

### LINKED ISSUE #6705: [UX] - Sort out difference between visibility conditions and access conditions in UI
## Description of the bug

I've been working on a couple of issues to help folks understand the relationship between paths and layouts, in specific in relationship to this new feature in core. 
**https://github.com/backdrop/backdrop-issues/issues/1929**

Specifically, I opened this issue:
https://github.com/backdrop/backdrop-issues/issues/6699

It has been pointed out that my use of "visibility conditions" in the text, might be better expressed as "access conditions" in relationship to paths. Which brings me to this inconsistency in the current UI (or at least what appears to be an inconsistency). 

image

Questions:

Is there a good reason for this inconsistency between the use of visibility and access on this specific page?
Can we fix this on this page alone or does this have much broader implications?

It is my current thought, that "visibility conditions" does make sense for blocks, but for paths the word "access conditions" is more accurate. I'm worried that this might make the site architect experience more confusing. BUT, it may also help a site architect better understand what is going on here.

As I begin to think about it this way, I think it's helping my understanding. 

## Expected behavior

Would it be better if this page looks like this (changing link from "Add visibility condition" to "Add access condition")? Does this have ramifications for other parts of the UI?

image

AND if so, should this be change from "Visibility Conditions" to "Visibility or Access Conditions"?

image

OR is this even worth worrying about?

Atten: @jenlampton, @quicksketch, @bugfolder

jenlampton:
Fixes https://github.com/backdrop/backdrop-issues/issues/6705

### DIFF
--- a/core/modules/layout/layout.admin.inc
+++ b/core/modules/layout/layout.admin.inc
@@ -151,7 +151,7 @@ function layout_list_page() {
   $header = array(
     array('data' => t('Layout'), 'class' => array('layout-title')),
     array('data' => t('Template'), 'class' => array('layout-template', 'priority-low')),
-    array('data' => t('Visibility conditions'), 'class' => array('layout-conditions', 'priority-low')),
+    array('data' => t('Conditions'), 'class' => array('layout-conditions', 'priority-low')),
     array('data' => t('Storage state'), 'class' => array('layout-status', 'priority-low')),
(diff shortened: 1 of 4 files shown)


quicksketch:
Merged into 1.x for 1.29.0. Thanks @jenlampton!

PR #4873, opened 2024-09-18
Issue #6717: Fix faulty version pattern in translation download

### LINKED ISSUE #6717: Translation not available during installation
## Description of the bug

When I try to install Backdrop in another language, I get a requirements problem: "The German translation is not available." The same happens with other languages (other than English).

## Steps To Reproduce

~~Note: This seems to be an issue which happens only in certain hosting environments. See the "Additional information" section below.~~

To reproduce the behavior:
1. Prepare a Backdrop `1.29.0` installation: database, code directory, etc.
2. Go to the (sub)domain of your environment, and run the installer.
3. Choose a language other than English, e.g. Spanish or German. The URL should contain a langcode, like in `https://test.example.com/core/install.php?langcode=de`
4. Hit "Save and continue".

## Actual behavior

You see a page with title "Requirements problem" and the following message:

> Translation. The German translation is not available. The German translation file is not available at the translation server. Choose a different language or select English and translate your website later.

Screenshot:

## Expected behavior

Backdrop is installed in the chosen language. (This is a [new feature](https://github.com/backdrop/backdrop-issues/issues/6660) introduced with Backdrop 1.29.)

## Additional information

(a) In a [related Zulip thread](https://backdrop.zulipchat.com/#narrow/stream/218635-Backdrop/topic/Installation.20in.20other.20language/near/470930011) we were discussing if this is a hosting environment issue. I was however able to reproduce the issue with two _different_ shared hosting services.

(b) When I continue the installation in English, everything works as usual, and after the installation I'm able to download and install modules via the user interface.

Technical info:
- Backdrop CMS version: 1.29.0
- Web server: Apache
- PHP version: tried both PHP 8.1 and 8.2

indigoxela:
Fixes https://github.com/backdrop/backdrop-issues/issues/6717

### DIFF
--- a/core/includes/install.core.inc
+++ b/core/includes/install.core.inc
@@ -1616,18 +1616,12 @@ function install_check_translation_download($langcode) {
   }
 
   $version = BACKDROP_VERSION;
+  $fallback_version = '1.x';
 
-  // For dev releases, remove the '-dev' part and trust the translation server
-  // to fall back to the latest stable release for that branch.
-  // @see locale_translation_build_projects()
-  if (preg_match("/^(\d+\.).*-(dev|preview)$/", $version, $matches)) {
-    // Example match: 1.28-dev => 1.x (Backdrop core).
-    $version = $matches[1] . 'x';
-  }
-  // For stable releases strip off the patch level.
-  elseif (preg_match("/^(\d+\.\d+)\.\d$/", $version, $matches)) {
-    // @todo It's not guaranteed that this file (already) exists.
-    $version = $matches[1];
+  // Get major and minor version, ignore any irrelevant dev or preview addition.
+  if (preg_match("/^(\d+)\.(\d+)/", $version, $matches)) {
+    $version = $matches[0];
+    $fallback_version = $matches[1] . '.x';
   }
 
   // Build URL for the translation file and the translation server.
@@ -1638,6 +1632,9 @@ function install_check_translation_download($langcode) {
     '%language' => $langcode,
   );
   $translation_url = strtr(INSTALL_LOCALIZATION_SERVER_PATTERN, $variables);
+  // Also build a generic fallback URL.
+  $variables['%version'] = $fallback_version;
+  $fallback_url = strtr(INSTALL_LOCALIZATION_SERVER_PATTERN, $variables);
 
   $elements = parse_url($translation_url);
   $server_url = $elements['scheme'] . '://' . $elements['host'];
@@ -1655,7 +1652,17 @@ function install_check_translation_download($langcode) {
     $online = TRUE;
   }
   if ($online) {
-    $translation_available = install_check_localization_server($translation_url);
+    // We have to start with the fallback URL, otherwise the static cache
+    // already returns FALSE.
+    $translation_available = install_check_localization_server($fallback_url);
+    if ($translation_available) {
+      // Look up the version specific translation file. We can not rely on its
+      // existence, as generation happens manually after release.
+      // If it does not exist, yet, we stick with the generic fallback.
+      if (!install_check_localization_server($translation_url)) {
+        $translation_url = $fallback_url;
+      }
+    }
   }
 
   // If the translations directory does not exist, throw an error.
--- a/core/includes/install.core.inc
+++ b/core/includes/install.core.inc
@@ -1616,7 +1616,7 @@ function install_check_translation_download($langcode) {
   }
 
   $version = BACKDROP_VERSION;
-  $fallback_version = '1.x';
+  $fallback_version = BACKDROP_CORE_COMPATIBILITY;
 
   // Get major and minor version, ignore any irrelevant dev or preview addition.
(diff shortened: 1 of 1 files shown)


core/includes/install.core.inc
```diff
@@ -1650,7 +1652,17 @@ function install_check_translation_download($langcode) {
...
   if ($online) {
-    $translation_available = install_check_localization_server($translation_url);
+    // We have to start with the fallback URL, otherwise the static cache
+    // already returns FALSE.
+    $translation_available = install_check_localization_server($fallback_url);
+    if ($translation_available) {
+      // Look up the version specific translation file. We can not rely on its
+      // existence, as generation happens manually after release.
+      // If it does not exist, yet, we stick with the generic fallback.
+      if (!install_check_localization_server($translation_url)) {
```
indigoxela, line 1662: Unfortunately another request to the server's unavoidable (unless I'm missing something). But it's a lightweight HEAD request, anyway.

hosef: Would it be better if we change `install_check_localization_server()` so it will continue to check if it gets a request in the 400 range? That implies that the translation site is fine, but the uri we used is not.

hosef: or at least add a flag so we can stop it from setting `$online = FALSE` in this one use case.

indigoxela: Function install_check_localization_server() is actually quite smart. It avoids any additional checks, if a previous one already failed. If the request to the server (without path) doesn't work at all, don't waste time with subsequent requests to the same server (timeout is only 5 seconds, but that would sum up). The most common case for failure will be that there's either no internet connection at all (local offline install), or some PHP problem (not able to do external requests).

Whether it's a 404 actually gets relevant in our last check - for the version specific translation file.
I couldn't think of a better logic for these checks or their order. @hosef or do you have something smarter in mind?

Remember, it's just a single (cheap) additional HEAD request, anyway.

hosef: I'm not really sure. I took a look at the function and it just feels like a simple 200 = good, all other codes = bad is way too simple for what could actually happen.

hosef: This PR is probably fine as is because we control the site being requested here, so we control the response codes. I've been working on writing some API clients recently, and it just feels bad that we are not properly handling the other response codes.

indigoxela: I admit, http handling's quite simple here: 200=good anything_else=bad. But I think, in this case this approach is fine.

We're not interested whether anything moved (3xx) or whether something's temporarily broken (5xx) or the file just isn't there (404). We quickly inform about problems if necessary, hide those we can silently fix (fallback file), and suggest an immediate solution via UI if not. To not make the install process cumbersome because of problems outside this (not yet) install.

All of the "not 200" mean, that we can't get the translation file right now. So why bother?

hosef: > So why bother?

So we don't have to check the fallback uri before checking the uri we actually want.

> Unfortunately another request to the server's unavoidable (unless I'm missing something).

The thing you are missing is proper response code handling. The server tells us why it didn't work and we are just ignoring it.

It works the way it is. We would just need to add more checks to that function to avoid making 2 requests.

indigoxela: > The thing you are missing is proper response code handling.

Am I? :stuck_out_tongue_winking_eye: 

What I *could* imagine, is a second param for install_check_localization_server() like `$ignore_cache` (default false) - that would allow to check the fallback *after* the regular translation file - one request less. If we got that far (we're online, the server's reachable), most likely the "regular" file should be accessible, too (unless the file hasn't been generated, yet).
But that additional param feels a bit odd and makes that utility function logic harder to understand. I might be wrong, though.
The actual reason, *why* the translation file isn't there isn't actually relevant IMO. We can't have it - try the fallback.

indigoxela: @hosef are you still interested in this PR? Any thoughts?

core/includes/install.core.inc
```diff
@@ -1616,13 +1616,12 @@ function install_check_translation_download($langcode) {
   }
 
   $version = BACKDROP_VERSION;
+  $fallback_version = '1.x';
```
quicksketch: A good place to use `BACKDROP_CORE_COMPATIBILITY` (which is exactly `1.x`):

```suggestion
  $fallback_version = BACKDROP_CORE_COMPATIBILITY;
```

core/includes/install.core.inc
quicksketch: For what it's worth, I'm mostly in agreement here with @indigoxela, that regardless of if it's a 404, 403, or 50x, if the translation isn't available then it's not going to matter much what the reason it. The installer could have more granular reporting, like "The translation server isn't available." (50x) vs "The translation couldn't be found." (404) but that granularity won't matter to end-users. We could add handling for redirects (`3xx` codes), and that would be useful. Overall though, I'm more keen to have the availability of translations fixed and not hold this up.

quicksketch:

PR #4877, opened 2024-09-27
Issue 5469. Remove leftover field info and instance configs.

### LINKED ISSUE #5469: Field instance config files are not being deleted when fully uninstalling Comment
## Description of the bug

The comment module does an incomplete cleanup, when uninstalled. This leads to PHP Notices on the config export page.

## Steps To Reproduce

1. Turn on error display
2. Uninstall the comment module
3. Go to /admin/config/development/configuration/single/export

## Actual behavior

Two PHP notices:

    Notice: Trying to access array offset on value of type null in field_config_label_instance() (Zeile 435 von .../core/modules/field/field.module).
    Notice: Trying to access array offset on value of type null in field_config_label_instance() (Zeile 435 von .../core/modules/field/field.module).

## Expected behavior

No notices, the comment module should remove all related config.

## Additional information

When the comment module gets uninstalled it does run `field_delete_field('comment_body');`, but two files stay untouched:

- field.instance.comment.comment_node_page.comment_body.json
- field.instance.comment.comment_node_post.comment_body.json

And these two files cause the php notices on the "Single export" configuration manager page. The Comment entity no longer exists, there's no instance label.

It may be worth checking, if this incomplete uninstall also affects other modules, for instance Taxonomy. Update: no, the Taxonomy module is not affected.

- Backdrop CMS version: latest stable, 1.21.1
- PHP version: 7.4

argiepiano:
Fixes https://github.com/backdrop/backdrop-issues/issues/5469


core/modules/entity/entity.module
```diff
@@ -25,6 +25,41 @@ function entity_modules_uninstalled($modules) {
...
+    $hook = $module . '_entity_info';
+    if (function_exists($hook)) {
+      $entity_infos = $hook();
+      foreach ($entity_infos as $entity_type => $entity_info) {
+        if (!empty($entity_info['fieldable'])) {
+          // Get any leftover instance config files.
+          $instance_config_names = config_get_names_with_prefix('field.instance.' . $entity_type . '.');
+          foreach ($instance_config_names as $instance_config_name) {
+            // Remove the instance config.
+            $config = config($instance_config_name);
```
indigoxela, line 41: This doesn't care about field data, right?

So whatever's in the database at this point will stay there - without any way to purge, as with the next line the instance config is gone. And with it the info about entity type and bundle.

As long as we have the instance conf, we'd know the entity type and bundle name - regardless of the actual entity definition (which isn't available anymore).

I'm assuming, `$config->set('deleted', 1);` wouldn't work as we don't have the entity def anymore, which seems required to delete data... Is that assumption correct? (Didn't try that out)

argiepiano: > So whatever's in the database at this point will stay there - without any way to purge, as with the next line the instance config is gone. And with it the info about entity type and bundle.

As mentioned in the issue, the database tables are all removed normally. All of this processing here is happening afterward. There are no tables remaining.

Didn't try the deleted property, but I assumed that, since the module is disabled and uninstalled, the delete will be ignored.

indigoxela: Right, can be marked as resolved :+1:  (GH perms don't allow me that).

PR #4880, opened 2024-10-05
Issue #6722: Add more info in the comments. [skip tests]

### LINKED ISSUE #6722: Improve code comments for template_preprocess_views_view_fields
## Description of the need

I was frustrated and lost time while trying to set the value of some custom fields in my view. In part that happened because I assumed that the hook template_preprocess_views_view_fields would process for all fields in any view. It was firing for some views I had, but not the one I was working on. This turned out to be because the view was using the Table format, so I should have used the template_preprocess_views_view_table hook.

## Proposed solution

Just add some more info in the code comments of the template_preprocess_views_view_fields hook, to give some clues about what's happening.

dsearle-oaklins:
Fixes [ISSUE_URL](https://github.com/backdrop/backdrop-issues/issues/6722#)

### DIFF
--- a/core/modules/views/templates/views.theme.inc
+++ b/core/modules/views/templates/views.theme.inc
@@ -162,7 +162,9 @@ function template_preprocess_views_view(&$variables) {
 }
 
 /**
- * Preprocess theme function to print a single record from a row, with fields
+ * Preprocess theme function to print a single record from a row, with fields.
+ * This hook will only be run on views whose Format settings choose to show fields.
+ * Therefore, it won't run for views that use the Table format.
  */
 function template_preprocess_views_view_fields(&$variables) {
   $view = $variables['view'];
--- a/core/modules/views/templates/views.theme.inc
+++ b/core/modules/views/templates/views.theme.inc
@@ -163,8 +163,11 @@ function template_preprocess_views_view(&$variables) {
 
 /**
  * Preprocess theme function to print a single record from a row, with fields.
- * This hook will only be run on views whose Format settings choose to show fields.
- * Therefore, it won't run for views that use the Table format.
+ *
+ * This hook will be invoked for display formats (other than Table) whose format
+ * settings choose to show fields.
+ *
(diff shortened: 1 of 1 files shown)


core/modules/views/templates/views.theme.inc
```diff
@@ -162,7 +162,9 @@ function template_preprocess_views_view(&$variables) {
 }
 
 /**
- * Preprocess theme function to print a single record from a row, with fields
+ * Preprocess theme function to print a single record from a row, with fields.
+ * This hook will only be run on views whose Format settings choose to show fields.
```
avpaderno: Short description and long description must be separated by an empty line.
The second line is longer than 80 characters.

indigoxela: And ideally there should be a reference to the template like:

```
@see views-view-field.tpl.php
```

argiepiano: `This hook will only be run on views whose Format settings choose to show fields` is not quite accurate. The `Table` Display format does use fields; however, this hook will not run when using `Table`. 

Screen Shot 2024-10-06 at 8 50 35 AM

Suggestion:
`This hook will be invoked for display formats (other than Table) whose format settings choose to show fields`

dsearle-oaklins: > And ideally there should be a reference to the template like:
> 
> ```
> @see views-view-field.tpl.php
> ```

Shouldn't that be 'views-view-fields.tpl.php' since this is template_preprocess_views_view_fields?

dsearle-oaklins:
PR updated.

core/modules/views/templates/views.theme.inc
indigoxela: > Shouldn't that be 'views-view-fields.tpl.php' since this is template_preprocess_views_view_fields?

Oops, you're of course right - should be ...fieldS. :+1: You already fixed it, resolved for me. :smiley:

PR #4883, opened 2024-10-06
Issue #6725: Track update module usage in Telemetry.

### LINKED ISSUE #6725: Track Update module usage in Telemetry
## Description of the need

Although Telemetry and Update module have similar value: they both provide usage information, we're seeing some interesting divergence between Telemetry and Update module usage tracking. Right now it _seems_ as though 90%+ of all reporting Backdrop sites have Telemetry module enabled, but it's possible that sites simply have Telemetry module turned on and Update module turned off. Some site owners (I'm thinking of @hosef here) have reported they never use Update module because they track updates in a different way (like on a different environment).

Since Update module usage likely differs from Telemetry module usage, we should track how much overlap there is between the data. This could help us determine a better overall count of Backdrop sites out there.

## Proposed solution

We should track if sites have Update module enabled (Yes/No) as a Telemetry data-point.

quicksketch:
Fixes https://github.com/backdrop/backdrop-issues/issues/6725

### DIFF
--- a/core/modules/telemetry/telemetry.telemetry.inc
+++ b/core/modules/telemetry/telemetry.telemetry.inc
@@ -53,6 +53,11 @@ function telemetry_telemetry_info() {
     'description' => t('The version of jQuery configured on your site.'),
     'project' => 'backdrop',
   );
+  $info['update_module'] = array(
+    'label' => t('Update module enabled'),
+    'description' => t('Whether this site has Update module enabled to check for project updates.'),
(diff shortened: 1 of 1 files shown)


yorkshire-pudding (approved):
Very simple code addition and all looks good to me.

quicksketch:
Merge into 1.x and 1.29.x. Thank you for reviewing @yorkshire-pudding!

PR #4888, opened 2024-10-18
Fixes #6732: function Layout::addBlock position not working properly

### LINKED ISSUE #6732: function Layout::addBlock position not working properly
## Description of the bug

The code to add a block to a layout in a specific position, does not seem to be working properly. 

## Steps To Reproduce

To reproduce the behavior:
1. Spin up a new/demo site. 
2. Add and enble the devel module
3. Use devel module to execute the following php
```
$home_layout = layout_load('home');
$home_layout->addBlock('views', 'promoted-block', 'top', 0);
$home_layout->save();
```

4. Check front page and see that the promoted content block has been added to the END of the top region, instead of as the first item in the top region. 

It is possible that it is my syntax that is incorrect. See:

https://docs.backdropcms.org/api/backdrop/core%21modules%21layout%21includes%21layout.class.inc/function/Layout%3A%3AaddBlock/1

> layout.class.inc	[Layout](https://docs.backdropcms.org/api/backdrop/core%21modules%21layout%21includes%21layout.class.inc/class/Layout/1)::addBlock($block_module, $block_delta, $region_name, $position = NULL)
> Add a block to a particular region.
> 
> Parameters:
>
> * string $block_module: The module providing the block to be positioned. 
> * string $block_delta: The block delta from [hook_block_info](https://docs.backdropcms.org/api/backdrop/core%21modules%21layout%21layout.api.php/function/hook_block_info/1)().
> * string $region_name: The name of the region within the layout.
> * int $position: The position of the block within the region. If not specified, the block will be the last block within the specified region.

## Expected behavior

I expected the Promoted Content block to be the first item in the top region, NOT the last item. 

## Additional information

It is possible that I've just got the wrong syntax.

robertgarrigos:
Fixes [#6732](https://github.com/backdrop/backdrop-issues/issues/6732)

### DIFF
--- a/core/modules/layout/includes/layout.class.inc
+++ b/core/modules/layout/includes/layout.class.inc
@@ -534,7 +534,7 @@ class Layout {
     $this->positions[$region_name][] = $block->uuid;
 
     // Move the block into the correct position.
-    if ($position) {
+    if (is_numeric($position)) {
       $positions = array();
       foreach ($this->positions[$region_name] as $n => $uuid) {
         if ($position == $n) {


quicksketch:
Merged into 1.x and 1.29.x. Thank you @robertgarrigos!

PR #4890, opened 2024-10-20
Issue #2409: Add backup functionality to update.php.

### LINKED ISSUE #2409: Basic backup and restore functionality
For every site I build I install the contrib module backup_migrate and I wonder why this is not a capability built in to core.

If it also included the backup of the active config json files it would be even more convenient.

Steps to test the PR:
https://github.com/backdrop/backdrop-issues/issues/2409#issuecomment-2495842287

**Advocate:** @quicksketch

quicksketch:
Pulls extensively from Backup and Migrate module, both Backdrop and Drupal 7 versions.

Existing code credit to @damienmckenna, @ronan, @quackers8me, @argiepiano, @cellear, and many others.

Fixes https://github.com/backdrop/backdrop-issues/issues/2409


core/includes/backup/backup.class.inc
```diff
@@ -0,0 +1,346 @@
...
+
+    $success = FALSE;
+    $compressed_file = clone $file;
+    $compressed_file->pushExtension('gz');
+
+    $source = $file->filePath();
+    $destination = $compressed_file->filePath();
+    $level = 9;
+
+    if (@function_exists('gzopen')) {
```
indigoxela: Why the `@` with function_exists()

AFAIK  function_exists() always returns a bool. I don't think, an error control operator is needed in this case.

core/modules/system/system.install
```diff
@@ -3632,6 +3643,14 @@ function system_update_1097() {
   }
 }
 
+
+/**
+ * Dummy update.
+ */
+function system_update_1099() {
```
indigoxela: Is that an oversight, or just there for more convenient local testing? :wink:

quicksketch: Yeah just there so you can run update.php and make a backup for now. I adjusted the code so that if there's no updates, there's no backup step, it just jumps you to the "no updates needed" message.

core/restore.php
```diff
@@ -0,0 +1,444 @@
...
+  backdrop_set_title(st('Access denied'));
+
+  $output = '';
+  $steps = array();
+
+  $output .= st('You are not authorized to access this page. Log in using either an account with the <em>restore site backups</em> permission, or the site maintenance account (the account you created during installation). If you cannot log in, you will have to edit <code>settings.php</code> to bypass this access check. To do this:');
+  $output = '<p>' . $output . '</p>';
+
+  $steps[] = st('Find the <code>settings.php</code> file on your system, and open it with a text editor.');
+  $steps[] = st('There is a line inside your <code>settings.php</code> file that says <code>$settings[\'restore_free_access\'] = FALSE</code>. Change it to <code>$settings[\'restore_free_access\'] = TRUE</code>.');
```
indigoxela: Can we add a semicolon to the code to copy-paste?
`<code>$settings[\'restore_free_access\'] = TRUE;</code>`
So if people actually paste that, they don't end up with a syntax error.

Maybe off topic: As a dedicated user of settings.**local.**php I find the wording a bit... hmmm. But wording... :shrug:

core/modules/system/system.install
quicksketch: I removed this update hook to avoid confusion.

core/restore.php
quicksketch: Added the semicolon. The use of `t()`/`st()` needs to be looked at a bit more here.

core/update.php
```diff
@@ -342,6 +316,117 @@ function update_info_page() {
...
+  return backdrop_render($elements);
+}
+
+
+/**
+ * Form constructor for the list of available database module updates.
+ */
+function update_backup_form($form, &$form_state) {
+  // Check if Backup directory is specified. If FALSE, this form shouldn't even
+  // be displayed. If NULL, link to documentation on setting up backups.
```
indigoxela, line 338: Still pending: actually skip this whole form is the setting's FALSE. :wink:

core/update.php
```diff
@@ -406,11 +491,18 @@ function update_task_list($set_active = NULL) {
...
+    'backup' => 'Backup',
     'select' => 'Review updates',
-    'run' => 'Run updates',
+    'update' => 'Run updates',
     'finished' => 'Review log',
   );
 
+  // Hide the Backup task if upgrading from Drupal 7, where the original
+  // database has already been modified and no config exists.
+  if (state_get('update_d7_upgrade', FALSE)) {
```
indigoxela: And also if `settings_get('backup_directory') === FALSE`, I guess.

core/update.php
quicksketch: This is now in place.

core/includes/backup.inc
```diff
@@ -0,0 +1,422 @@
...
+}
+
+/**
+ * Given a particular backup/restore target, find the class that applies.
+ *
+ * @param string $target
+ *   The backup source or restore target as a string, such as "db:default" or
+ *   "config:active".
+ *
+ * @return Backup|false
```
indigoxela: The return type is rather "string|false", it's never an instance of "Backup".

core/includes/backup/backup.database.inc
```diff
@@ -0,0 +1,350 @@
...
+
+  /**
+   * {@inheritdoc}
+   */
+  public function postBackup(BackupFile &$file) {
+    parent::postBackup($file);
+    // Compress the backup and delete the original once compressed.
+    if ($this->settings['compression'] === self::COMPRESSION_GZIP) {
+      // The returned value should overwrite $file by reference.
+      $file = $this->compress($file, self::COMPRESSION_GZIP, TRUE);
```
indigoxela: What happens, if compression fails? Could it fail?

core/includes/backup/backup.database.inc
```diff
@@ -0,0 +1,350 @@
...
+          $key = uniqid('backdrop_backup_tmp_');
+          $target = 'default';
+          Database::addConnectionInfo($key, $target, $info);
+        }
+        // No database selected. Assume the default.
+        else {
+          $key = $target = 'default';
+        }
+      }
+      if ($target && $key) {
```
indigoxela: Seems like $target and $key are always set (strings). But maybe I'm missing something...

core/includes/backup/backup.database.inc
```diff
@@ -0,0 +1,350 @@
...
+   */
+  abstract protected function restoreDatabaseFromFile(BackupFile $file);
+
+  /**
+   * Get a list of objects in the database.
+   */
+  protected function getObjectNames() {
+    // Must be overridden.
+    $out = $this->getTableNames();
+    if (method_exists($this, 'getViewNames')) {
```
indigoxela: Not sure, if I'm getting this right, but this method will always exist - it's defined a couple of lines below. So what's the purpose of this method_exists()? Or should the check be for something different?

core/includes/backup/backup.database.inc
```diff
@@ -0,0 +1,350 @@
...
+   * @return array
+   *   The list of tables to be locked, if any.
+   */
+  protected function getLockedTables() {
+    $tables = array();
+    if ($this->settings['lock_tables']) {
+      foreach ($this->getTableNames() as $table) {
+        // There's no need to lock excluded or structure only tables because it
+        // doesn't matter if they change.
+        if (empty($settings['exclude_tables']) || !in_array($table, (array) $settings['exclude_tables'])) {
```
indigoxela: Variable `$settings` is never defined here. Shouldn't that rather be `$this->settings`? Maybe that's why this is casted to array, to prevent PHP nagging.

But I wonder, if we need that exclude_tables setting, anyway. Where would it be set?

core/includes/backup/backup.file.inc
```diff
@@ -0,0 +1,260 @@
...
+      }
+      $this->close();
+    }
+    else {
+      backdrop_not_found();
+    }
+
+    // Start buffering and throw away the results so that errors don't get
+    // appended to the file.
+    ob_start('_backup_file_dispose_buffer');
```
indigoxela: So, this _backup_file_dispose_buffer is supposed to be a callback, but it doesn't exist, so the "buffer" is thrown away? Shouldn't the callback be `NULL` in this case? Not sure.

core/includes/backup/backup.mysql.inc
```diff
@@ -0,0 +1,583 @@
...
+    if (!empty($settings['nodata_tables'])) {
+      foreach ((array) $settings['nodata_tables'] as $table) {
+        if (isset($all_tables[$table])) {
+          $nodata_tables[] = $table;
+          $command .= ' --ignore-table=' . $db . '.' . $table;
+        }
+      }
+    }
+    // @todo: This is not yet supported.
+    $success = backup_migrate_exec($command, $args);
```
indigoxela: I'm pretty sure, this backup_migrate_exec() is an oversight. :wink: 

There are more backup_migrate... functions below.

core/includes/backup/backup.mysql.inc
```diff
@@ -0,0 +1,583 @@
...
+  /**
+   * {@inheritdoc}
+   */
+  protected function restoreDatabaseFromFile(BackupFile $file) {
+    $settings = $this->settings;
+    $lines_executed = 0;
+
+    if ($file->open() && $connection = $this->getDatabaseConnection()) {
+      // Optionally drop all existing tables.
+      if (!empty($settings['utils_drop_all_tables'])) {
```
indigoxela: This is probably one of those settings we don't need in core and could drop entirely. Is my assumption correct?

core/includes/backup/backup.mysql.inc
```diff
@@ -0,0 +1,583 @@
...
+      }
+
+      // Finish any unfinished insert statements.
+      if ($rows > 0) {
+        $file->write(";\n");
+        $lines++;
+      }
+    }
+
+    if ($this->settings['verbose']) {
```
indigoxela, line 386: Do we have a "verbose" setting, anyway? Do we need it? If so, we'd also need an equivalent for backup_migrate_get_peak_memory_usage(), as it's used below.

core/restore.php
```diff
@@ -0,0 +1,450 @@
...
+const MAINTENANCE_MODE = 'restore';
+
+/**
+ * Form constructor for the list of backups that can be restored.
+ */
+function restore_script_selection_form($form, &$form_state) {
+  $existing_backups = array();
+
+  // If there are no backups, display a message.
+  if (empty($existing_backups)) {
```
indigoxela: $existing_backups is always empty - as it's defined as empty array right above. :wink: I don't actually get the comment above.

Currently this function isn't in use, yet, right? Is it supposed to get used in the future?

core/restore.php
```diff
@@ -0,0 +1,450 @@
...
+      <input type="submit" value="Continue" class="form-submit button-primary" />
+      <a href="' . base_path() . '">Cancel</a>
+    </div>
+  </form>';
+  $output .= "\n";
+  return $output;
+}
+
+/**
+ * Provides a form to create an on-demand backup before updating.
```
indigoxela: This looks like a copy-pasted comment. :wink: We're in restore.php here.

core/update.php
```diff
@@ -342,6 +316,116 @@ function update_info_page() {
...
+  );
+
+  foreach ($backup_targets as $backup_target) {
+    list($backup_plugin_id, $backup_subkey) = explode(':', $backup_target, 2);
+    $backup_class = backup_get_handler_name($backup_target);
+    if (!$backup_class) {
+      continue;
+    }
+
+    /** @var Backup $backup_instance */
```
indigoxela: No var $backup_instance here, as far as I can see no instance of Backup, either. Copy-paste zombie?

core/update.php
```diff
@@ -531,46 +622,109 @@ function update_check_requirements($skip_warnings = FALSE) {
...
+
+    case t('Create backup'):
+      if ($valid_token) {
+        // Generate absolute URLs for the batch processing (using $base_root),
+        // since the batch API will pass them to url() which does not handle
+        // update.php correctly by default. Note the "action" query parameter
+        // here distinguishes between the two batch operations, which can be
+        // either "batch" or "update".
+        $batch_url = $base_root . backdrop_current_script_url(array('action' => 'backup'));
+        $batch_redirect_url = $base_root . backdrop_current_script_url(array('op' => 'selection'));
```
indigoxela, line 646: Where does this `$base_root` variable come from? Can't seem to find it.

quicksketch: It is a global defined in `backdrop_settings_initialize()`. I'm not much a fan of this URL creation but it's been in place for a long time.

core/includes/backup.inc
quicksketch: Thank you, fixed and more docs added.

core/includes/backup/backup.database.inc
quicksketch: It's possible. Maybe if you run out of disk space or zlib isn't available (though you'd have other errors in Backdrop if that were the case). I'll add a graceful failure here.

core/includes/backup/backup.database.inc
quicksketch: I think you're right, I'll remove this unnecessary condition.

core/includes/backup/backup.database.inc
quicksketch: You're right. In Backup and Migrate module, these methods were not declared as `abstract` and so it was possible for a child class to not implement it. In this version, child classes must implement it or get a PHP error, so we can remove this check.

core/includes/backup/backup.database.inc
quicksketch: Thanks, nice catch.

> But I wonder, if we need that exclude_tables setting, anyway. Where would it be set?

It's set in `BackupDatabase::defaultSettings()`. It's possible to override via the API, but the defaults exclude all `simpletest_*` tables (those used when running tests).

core/includes/backup/backup.file.inc
quicksketch: I'll have to look into what this does. Here's the matching code in B&M:
```
/**
 * An output buffer callback which simply throws away the buffer instead of sending it to the browser.
 */
function _backup_migrate_file_dispose_buffer($buffer) {
  return "";
}
```

core/includes/backup/backup.mysql.inc
quicksketch: I haven't port this bit yet since I'm considering if we should include `mysqldump` functionality at all. It's not on by default in B&M and it's probably not available on a good number of servers. Also `mysqldump` is no longer compatible with `mariadb` import. See https://ddev.com/blog/mariadb-dump-breaking-change/

It seems like we might be better suited if we narrowed our support to just the PHP-generated dumps and left more specific tooling to contrib or external tools.

core/includes/backup/backup.mysql.inc
quicksketch: I'm not totally sure if it can/should be removed; or how we would surface it in the UI. It possible that a new database table could be created in an update hook and not dropping it could cause an issue.

core/restore.php
quicksketch: > Currently this function isn't in use, yet, right? Is it supposed to get used in the future?

Looks like this function is entirely unused. I must have restarted it at some point and didn't delete the previous version. I've removed the whole function. Thanks!

core/update.php
quicksketch: Thanks, code refactor issue where `backup_get_handler_name()` once was `backup_get_handler()` and returned a class.

core/includes/backup/backup.mysql.inc
quicksketch: I've maintained it in the API as it might be useful for testing output. But yes, there's still a number of `backup_migrate_*` calls that need to be fixed.

core/includes/backup/backup.file.inc
indigoxela: Or, maybe we don't need function transfer() at all, as we don't want to provide file download, anyway. If I got things right...

core/includes/backup/backup.mysql.inc
indigoxela: > ... not on by default in B&M and it's probably not available on a good number of servers.

That's a good point. Under the hood it uses exec(), which will be disabled on many, if not most shared hosting servers.

> It seems like we might be better suited if we narrowed our support to just the PHP-generated dumps

Agreed. :+1:

core/modules/system/system.module
```diff
@@ -168,8 +168,20 @@ function system_permission() {
     'administer software updates' => array(
       'title' => t('Administer software updates'),
       'restrict access' => TRUE,
-      'warning' => t('Run site updates for Backdrop projects.'),
+      'warning' => t('Allows running site updates (update.php) for Backdrop projects.'),
     ),
+  );
+  // If backups are not configured, do not show related permission.
+  if (settings_get('backups_directory')) {
+    $permissions += array(
```
indigoxela, line 176: That looks like a typo: `'backupS_directory'` with "s" - as a result, the permissions do no show, currently.

core/update.php
indigoxela: Oh, got it, we're not inside a function and this is one of the globals... Thank's for explaining. :+1:

core/includes/backup/backup.database.inc
```diff
@@ -0,0 +1,351 @@
...
+   */
+  public function defaultSettings() {
+    $settings = parent::defaultSettings();
+
+    $all_tables = $this->getTableNames();
+
+    // Basic modules that should be excluded.
+    $basic = array(
+      // Default core tables.
+      'session',
```
indigoxela: As we do not dump `session` table data, we might also want to not dump `batch` table data. I just realized, that the "backup" batch lands in the dump. :wink:

core/modules/system/system.module
quicksketch: Thanks, fixed!

core/includes/backup/backup.database.inc
quicksketch: I had excluded the session table in testing at one point, but I actually think it's good to include so that all users don't get logged out on restore.

Though interesting point about `batch`... maybe it should be excluded (nodata). I was also considering that maybe the `tempstore` table should be excluded (which holds form state and views/layout in-progress changes).

core/includes/backup/backup.mysql.inc
quicksketch: I went ahead and removed this functionality for the time being. It's probably too risky to enable by default and has a higher risk of data loss. We can look at putting it back if it becomes a problem, but in a catastrophic event, if an incorrect table was restored its relatively easy to drop an extra table than to get back a bunch of tables that were incorrectly dropped.

core/includes/backup/backup.mysql.inc
quicksketch: I went ahead and remove the `mysqldump` capability and use of `exec()` for the initial version of this feature.

core/includes/backup/backup.file.inc
quicksketch: I removed this functionality since we're not using it in the initial version.

core/includes/backup/backup.mysql.inc
quicksketch: All `backup_migrate_*`  usages are now eliminated.

core/includes/backup/backup.database.inc
```diff
@@ -0,0 +1,398 @@
...
+        if (strpos($table, $default_prefix) === 0) {
+          $prefixed_tables[] = $table;
+        }
+      }
+    }
+
+    // Identify tables that should omit data by default.
+    $nodata_basic = array(
+      // Default core tables.
+      'accesslog',
```
indigoxela: There's no such table in core (accesslog).

And I'm not sure, whether we should consider Devel tables here.
B&M has to (sort of), as it's supposed to be a migration tool, too. But a quick core update?

core/modules/system/system.module
```diff
@@ -964,6 +978,17 @@ function system_menu() {
     'file' => 'system.admin.inc',
     'weight' => -20,
   );
+  if (settings_get('backup_directory') !== FALSE) {
+    $items['admin/config/system/backup-restore'] = array(
+      'title' => 'Site backups',
+      'description' => 'Change site backup settings and view current backups.',
+      'page callback' => 'backdrop_get_form',
+      'page arguments' => array('system_backup_settings'),
```
indigoxela: Hm. On admin/config/system/backup-restore I get a hard error:

```

Warning: Undefined array key "system_backup_settings" in backdrop_retrieve_form() (line 822 of ...core/includes/form.inc).
TypeError: call_user_func_array(): Argument #1 ($callback) must be a valid callback, function "system_backup_settings" not found or invalid function name in backdrop_retrieve_form() (line 866 of ...core/includes/form.inc).
```

WIP, I guess. as there's really no such function, yet. :stuck_out_tongue_winking_eye: I better wait, until this form's implemented.

core/includes/backup/backup.database.inc
quicksketch: Yeah I think we can remove the whole list here. :+1:

core/modules/system/system.module
quicksketch: Yeah, I started on one but now I think it's unnecessary. I'll remove this from hook_menu().

core/restore.php
```diff
@@ -0,0 +1,418 @@
...
+
+/**
+ * Provides a form to select a backup to restore.
+ *
+ * @return string
+ *   Rendered HTML form.
+ */
+function restore_select_page() {
+  restore_task_list('select');
+  backdrop_set_title('Select to backup restore');
```
indigoxela: That's odd wording... wouldn't something like "Select a backup to restore from" be a better fit? :wink:

core/modules/system/config/system.core.json
```diff
@@ -15,6 +15,7 @@
       "strip_sensitive_headers_on_host_change": true,
       "strip_sensitive_headers_on_https_downgrade": true
     },
+    "backup_limit": 5,
```
indigoxela: I wonder, if this "backup_limit" belongs in (system) config.

The directory's set in settings(.local).php - probably to have it available for restore - even with weirdest failures.
Would anyone expect the limit to be in system.core.json, though? To me it appears to be sort of an "alien" here. I know, we can rely on config to be available at the point, this setting's needed. It's actually for other reasons, I wonder, if it even belongs here.

Also thinking of a contrib module extending core backup-restore ... wouldn't they provide a custom config, anyway? Optionally overridden in settings... :thinking: 

Core Backup handles the absence gracefully, so it would be OK to not be in config.

core/update.php
```diff
@@ -531,46 +574,116 @@ function update_check_requirements($skip_warnings = FALSE) {
...
+    case 'backup':
+      if ($valid_token) {
+        $output = update_backup_page();
+      }
+      else {
+        install_goto('core/update.php');
+      }
+      break;
+
+    case t('Create backup'):
```
indigoxela, line 638: Hm... in this case the $op is a translatable string? Seems odd to me.

I know, "Apply pending updates" below does the same, but don't we have a key available, that doesn't rely on a translated string, which might - depending on language - contain all sorts of characters?

Or, asked the other way round - what's the benefit of using a translatable string as key?

core/restore.php
```diff
@@ -0,0 +1,418 @@
...
+/**
+ * Provides an overview of the backup restoration process.
+ *
+ * This page provides cautionary statements before proceeding with a restore.
+ *
+ * @return string
+ *   Rendered HTML form.
+ */
+function restore_info_page() {
+  restore_task_list('info');
```
indigoxela, line 41: I wonder, if it might make sense to prevent this page entirely.

Currently it's available (based on permissions) even if `$settings['backup_directory']` is either an empty string or FALSE. It just leads to an empty "Select backup" listing (step).

For example if the settings FALSE, return a 403 instead? Or would this be too confusing and the empty listing provides better UX?

core/restore.php
quicksketch: Thanks, should just be "Select backup to restore"

core/modules/system/config/system.core.json
quicksketch: I considered `system.backup.json` as an alternative. Other existing files like `system.mail.json` and `system.authorize.json` are similarly separate and very slim config files. Do you think that would be better?

core/update.php
quicksketch: > Or, asked the other way round - what's the benefit of using a translatable string as key?

Strange as it is, this is the way nearly all forms work in Backdrop because the label of a button is the submitted POST value. For forms that have multiple buttons, that is the only way to determine which button was clicked. That's not necessarily the case here (there's only a single button), but this is very common.

core/restore.php
indigoxela: > should just be "Select backup to restore"

Yes, better. But it seems like you forgot to commit it - it's still the old string here. :grin:

core/update.php
indigoxela: Ah, many thanks for your explanations. Makes sense. Consider this question as answered / resolved. :+1:

core/modules/system/config/system.core.json
indigoxela: How to deal with this (config) item probably depends on future steps... :thinking: 

If the plan is to provide a UI for settings in core, then a dedicated config json file makes more sense than stuffing it into system.core.json. The UI should be a topic for a follow-up, though.

If the plan is to leave this entirely to contrib/custom (to extend the class and provide more settings), then this single config item makes no sense at all.

Another thing to consider: only a handful of people will even be aware that `backup_limit` could be overridden by fiddling in JSON. You, me, the core committer who merges this PR...
It's rather pointless to provide a config item without any UI - a secret feature nobody's aware of is no feature. We could stick with a sane default (5) in that case. :stuck_out_tongue_winking_eye: Eventually with a (documented) settings.php override.

core/update.php
quicksketch: Blame the HTML spec for not having a separate label on `<input type="submit" value="Submit">` :stuck_out_tongue:

core/restore.php
quicksketch: Ah, thanks I meant to make a better page for that! What it should do is show you the page (if you have access or `restore_free_access = TRUE`) and instead inform you of the purpose and link to documentation, rather than just showing an empty list.

core/modules/system/config/system.core.json
quicksketch: Backup and Migrate has 28 global settings and ~20 options per advanced backup. I initially figured there would be more settings, but most of the options in B&M seem to cover features we haven't included in core (like a cron-based backup) or unnecessarily detailed.

> If the plan is to leave this entirely to contrib/custom (to extend the class and provide more settings), then this single config item makes no sense at all.

I think we'll probably see some of both. We could make an option for downloading/uploading backups for example, or cron backups, in core... it depends on where the demand and comfort level is. So maybe splitting out (even for a single value) might be good future-proofing.

Even with no UI, I feel like the value should be at least _possible_ to override, and it makes more sense as config than a hard-coded setting. It's also easier to test as config.

indigoxela: So, the plan seems to be to add *some* settings via UI. Of course, never all the whistles and bells that B&M provides. The backup_limit, the file name pattern (eventually), provide a download (eventually)... all to be determined in a follow-up.

But even if there would only be three settings, I'd recommend a dedicated json file, not reusing system.core.

core/restore.php
indigoxela: Hm. The restore_free_access info works just fine. What doesn't seem to be considered is, that the backup feature is off entirely. The page is still accessible. That does no harm, but doesn't provide any value, either.

TBH, I'm not 100% sure, what I'd expect to find on `core/restore.php` (depending on permissions), if backups are completely off.

quicksketch: `update.php` provides somewhat useful information on its 403 page (though if you had admin access it might not be a 403 for restore.php):

image

core/includes/install.core.inc
```diff
@@ -946,6 +948,28 @@ function install_config_location(&$install_state) {
...
 }
 
+/**
+ * Installation task; Rewrite settings.php to include backup location.
+ */
+function install_backup_location(&$install_state) {
+  global $settings, $databases;
+
+  // A value of FALSE disables backups and will not rewrite on install.
+  if (empty($settings['backup_directory']) && $settings['backup_directory'] !== FALSE) {
```
herbdool: Getting this error if `$settings['backup_directory']` is commented out during install:

> Warning: Undefined array key "backup_directory" in install_backup_location() (line 958 of /app/httpdocs/core/includes/install.core.inc).

herbdool: Something like this?

```suggestion
  if (isset($settings['backup_directory']) && ($settings['backup_directory'] !== '' || $settings['backup_directory'] !== FALSE)) {

quicksketch: Thanks, I thought `empty()` was safe for that purpose but I guess not. Replaced with a full `isset()` check.

core/restore.php
quicksketch: I made an alternative page if `$settings['backup_directory']` is not specified:

image

core/restore.php
quicksketch: It's fixed now!

core/modules/system/config/system.core.json
quicksketch: Separated into `system.backup.json`, thanks for your input!

quicksketch:
Merged into 1.x for 1.30.0! Huge, huge thank you to @indigoxela for the dozens of review comments! Full credits will be posted to the main issue.

PR #4895, opened 2024-11-01
Issue 6745. Use key 'language' instead of 'langcode' in views handler.

### LINKED ISSUE #6745: Translated content doesn't get aliased path in views title
## Description of the bug

Links to translated content which is listed in views, e.g. on the "Manage content" page, don't point to the aliased path.

In other words: Path aliases in views work only, if interface and content language are the same. (See next comment.)

## Steps To Reproduce

To reproduce the behavior:
1. Enable Content Translation, Language and Locale from Backdrop core.
2. Add a second language, i.e. German.
3. Configure the Page content type, go to the tab "Multilingual support", and choose "Enabled, with translation".
4. Create a page of title "Hello world", put some content in the body, set the language to English, and save the page.
The page gets a URL like `example.com/hello-world`.
5. Hit the "Translate" tab, and add a translation.
The translated page gets a URL like `example.com/de/hallo-welt`.
6. Switch the Backdrop interface back to English, using the language switcher in the Admin bar.
7. Go to the _Manage content_ page, and inspect the page title link paths of the pages created above.

## Actual behavior

The path to the English page "Hello World" is the alias `example.com/hello-world` (as expected).
The translation "Hallo Welt" doesn't link to an alias but to the node ID path, e.g. `example.com/node/7`.

## Expected behavior

The link to the translation "Hallo Welt" points to the alias, e.g. `example.com/de/hallo-welt`.

argiepiano:
Fixes https://github.com/backdrop/backdrop-issues/issues/6745

### DIFF
--- a/core/modules/node/views/views_handler_field_node.inc
+++ b/core/modules/node/views/views_handler_field_node.inc
@@ -58,10 +58,10 @@ class views_handler_field_node extends views_handler_field {
           $languages = language_list();
           $language = $this->get_value($values, 'langcode');
           if (isset($languages[$language])) {
-            $this->options['alter']['langcode'] = $languages[$language];
+            $this->options['alter']['language'] = $languages[$language];
           }
           else {
-            unset($this->options['alter']['langcode']);
+            unset($this->options['alter']['language']);
           }
         }
       }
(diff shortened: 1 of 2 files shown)


quicksketch:
Thank you @argiepiano! Merged into 1.x and 1.29.x.

PR #4896, opened 2024-11-01
Issue 6746. Send language object to l() to get the aliased path.

### LINKED ISSUE #6746: Translated content doesn't get aliased path in views path
## Description of the bug

This is a follow-up to #6745 (about views _Title_ fields with links). This one here is about views _Path_ fields:

Paths to translated content which is listed in views, e.g. on the "Manage content" page, aren't rendered as the aliased path. Path aliases in views work only, if interface and content language are the same.

## Steps To Reproduce

To reproduce the behavior:
1. Enable Content Translation, Language and Locale from Backdrop core.
2. Add a second language, i.e. German.
3. Configure the Page content type, go to the tab "Multilingual support", and choose "Enabled, with translation".
4. Create a page of title "Hello world", put some content in the body, set the language to English, and save the page.
The page gets a URL like `example.com/hello-world`.
5. Hit the "Translate" tab, and add a translation.
The translated page gets a URL like `example.com/de/hallo-welt`.
6. Switch the Backdrop interface back to English, using the language switcher in the Admin bar.
7. Go to the _Manage content_ page, edit the view and add a Path field.
8. Back on the _Manage content_ page, inspect the path of the pages created above.

## Actual behavior

The path of the English page "Hello World" is the alias `/hello-world` (as expected).
The path of the translation "Hallo Welt" isn't an alias but the node ID path, e.g. `/node/7`.

## Expected behavior

The path of the translation "Hallo Welt" is an alias, e.g. `/de/hallo-welt`.

argiepiano:
Fixes https://github.com/backdrop/backdrop-issues/issues/6746

### DIFF
--- a/core/modules/node/views/views_handler_field_node_path.inc
+++ b/core/modules/node/views/views_handler_field_node_path.inc
@@ -11,6 +11,13 @@
  */
 class views_handler_field_node_path extends views_handler_field {
 
+  function init(&$view, &$options) {
+    parent::init($view, $options);
+    if (module_exists('translation')) {
+      $this->additional_fields['langcode'] = array('table' => 'node', 'field' => 'langcode');
+    }
+  }
+
   function option_definition() {
     $options = parent::option_definition();
     $options['absolute'] = array('default' => FALSE, 'bool' => TRUE);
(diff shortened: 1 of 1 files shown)


core/modules/node/views/views_handler_field_node_path.inc
```diff
@@ -34,13 +49,27 @@ class views_handler_field_node_path extends views_handler_field {
...
   }
 
-  function render($values) {
+  /**
+   * Render the path field.
+   */
+  public function render($values) {
     $nid = $this->get_value($values, 'nid');
-    return url("node/$nid", array('absolute' => $this->options['absolute']));
+    $options['absolute'] = $this->options['absolute'];
```
quicksketch, line 66: Although it's not strictly necessary, let's declare `$options` before adding a value to it.

```suggestion
    $options = array();
    $options['absolute'] = $this->options['absolute'];
```

quicksketch:
Thank you @argiepiano! Merged into 1.x and 1.29.x.

PR #4899, opened 2024-11-10
Issue 6749. Fix documentation for EntityInterface::uri

### LINKED ISSUE #6749: Docs for EntityInterface::uri() incorrectly state that NULL should be returned if entity has no uri.
## Description of the bug

The [documentation](https://github.com/backdrop/backdrop/blob/c53e13d829406bcfcae365754d8345dd57da69f3/core/modules/entity/entity.class.inc#L69) for `EntityInterface::uri()` incorrectly state that `NULL` should be returned if the entity has no uri of its own.

Returning NULL creates a fatal error when `template_preprocess_entity()` is invoked: `TypeError: Unsupported operand types: null + array`, since that preprocess function [uses](https://github.com/backdrop/backdrop/blob/c53e13d829406bcfcae365754d8345dd57da69f3/core/modules/entity/entity.theme.inc#L22) the `+` array operator with the returned value (NULL).  

Instead, the docs should say that an empty array should be returned.

## Steps To Reproduce

This was originally reported in the `commerce` issue queue: https://github.com/backdrop-contrib/commerce/issues/25
See steps there. A custom, uri-less entity class is needed. 

## Actual behavior

Fatal error when uri returns NULL

## Expected behavior

Not a fatal error.

## Additional information

Backdrop 1.29.1

argiepiano:
Fixes https://github.com/backdrop/backdrop-issues/issues/6749

### DIFF
--- a/core/modules/entity/entity.class.inc
+++ b/core/modules/entity/entity.class.inc
@@ -66,7 +66,7 @@ interface EntityInterface {
   /**
    * Returns the URI elements of the entity.
    *
-   * @return
+   * @return array
    *   An array containing the 'path' and 'options' keys used to build the URI
    *   of the entity, and matching the signature of url(). An empty array if the
    *   entity has no URI of its own.
--- a/core/modules/entity/entity.class.inc
+++ b/core/modules/entity/entity.class.inc
@@ -68,8 +68,8 @@ interface EntityInterface {
    *
    * @return
    *   An array containing the 'path' and 'options' keys used to build the URI
-   *   of the entity, and matching the signature of url(). NULL if the entity
-   *   has no URI of its own.
+   *   of the entity, and matching the signature of url(). An empty array if the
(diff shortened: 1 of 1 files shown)


elisseck:
I reviewed this and it looks good / works for me! Brings the inline docs in line with the reality of what `template_preprocess_entity()` is expecting. It seems like I am not able to actually change labels on anything so hopefully this comment is enough!

core/modules/entity/entity.class.inc
```diff
@@ -68,8 +68,8 @@ interface EntityInterface {
    *
    * @return
```
quicksketch: Nice, we should go ahead and add this type hint too:

```suggestion
   * @return array
```

quicksketch:
Merged into 1.x and 1.29.x. Thank you @argiepiano and @elisseck!

PR #4903, opened 2024-11-18
Issue #6755: Put 404 and 403 content inside a semantic element.

### LINKED ISSUE #6755: [A11Y] Content on 404, 403 are not within a semantic element.
## Description of the bug

An internal accessibility review flagged default 404 page content:

> Content is not within a semantic element

## Steps To Reproduce

Look at a default 404 page. The text "The requested page **_path_** could not be found." is not within a semantic element.

## Expected behavior

I'll suggest we display that content inside a paragraph tag. PR coming.

laryn:
Fixes https://github.com/backdrop/backdrop-issues/issues/6755

### DIFF
--- a/core/modules/system/system.module
+++ b/core/modules/system/system.module
@@ -4213,14 +4213,14 @@ function system_system_info_alter(&$info, $file, $type) {
  * Menu callback; Returns the standard "Page not found" error page.
  */
 function system_404() {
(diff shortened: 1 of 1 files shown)


quicksketch:
Merged into 1.x and 1.29.x. Thank you @laryn!

PR #4920, opened 2024-12-06
Issue #6770: Track the last-active CKEditor 5 instance.

### LINKED ISSUE #6770: Track active CKEditor 5 instance
## Description of the need

Previous WYSIWYG systems (like the D7 WYSIWYG module) tracked the last active editor with `Drupal.wysiwyg.activeId`. And tinyMCE tracks the current editor with `tinyMCE.activeEditor`.

CKEditor does not provide this ability, and accomplishing it in CKEditor 5 is particularly tricky, see this [StackOverflow post on getting active instances](https://stackoverflow.com/questions/48575534/ckeditor-5-get-editor-instances/48682501#48682501).

Some contrib modules like Insert [would like the ability to insert into CKEditor 5 instances](https://github.com/backdrop-contrib/insert/issues/22). Since this is difficult to add from outside of core, it would be preferable if core could provide this mechanism for use by contrib.

## Proposed solution

Backdrop core's `ckeditor5.js` file should track the active instance in `Backdrop.ckeditor5.activeEditor`.

The Insert module is trying to do this and sort of achieves it by using a `window.timeout()`, but it's not 100% reliable. It would be better to use the `.then()` promise after an instance is created.

quicksketch:
Fixes https://github.com/backdrop/backdrop-issues/issues/6770

### DIFF
--- a/core/modules/ckeditor5/js/ckeditor5.js
+++ b/core/modules/ckeditor5/js/ckeditor5.js
@@ -297,8 +297,8 @@
         Backdrop.ckeditor5.activeEditor = editor;
       }
       // Track when focus is set on a new editor.
-      // See https://ckeditor.com/docs/ckeditor5/latest/framework/deep-dive/ui/focus-tracking.html#checking-if-the-view-document-is-focused
-      editor.editing.view.document.on('change:isFocused', (evt, data, isFocused) => {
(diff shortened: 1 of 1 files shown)


quicksketch:
Merged with @indigoxela's review. Thank you @indigoxela!

PR #4921, opened 2024-12-08
Issue #6771: Fix editor not initializing because of other field on same page

### LINKED ISSUE #6771: CKEditor: Editor does not display on formatted text fields requiring different text formats
## Description of the bug

This appears to happen where there are two formatted text fields on the same content type. For example, if the first field added allows only the format "Limited HTML," and then a second field whose allowed format is "Filtered HTML," CKEditor will not appear on the second field. If you change both fields to "Limited HTML," the editor appears on both.

## Steps To Reproduce

To reproduce the behavior:
1. Create a content type with one formatted text field. Allow one text format that utilizes CKEditor.
2. View the edit screen, the editor should appear.
3. Add aa second formatted text field on that content type, but allow a different text format that utilizes CKEditor.
4. View the edit screen. The editor should appear on the first field, but not on the second.

## Actual behavior

CKEditor does not consistently appear on formatted text fields that allow different formats from one another.

## Expected behavior

CKEditor should appear on all fields if configured, and display the correct editor settings to match the text format.

indigoxela:
Fixes https://github.com/backdrop/backdrop-issues/issues/6771

### DIFF
--- a/core/modules/filter/filter.module
+++ b/core/modules/filter/filter.module
@@ -1316,7 +1316,8 @@ function filter_process_format($element) {
   $formats = filter_formats($user);
   $fallback_format = filter_fallback_format();
 
-  // Turn original element into a text format wrapper.
+  // Turn original element into a text format wrapper. Settings for all
+  // available formats are added to make them available for all fields.
   if (isset($element['#attached'])) {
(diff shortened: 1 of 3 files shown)


quicksketch:
Thank you @indigoxela! I've merged this into 1.x and 1.29.x.

PR #4929, opened 2024-12-14
Issue 6775. Send CRLF at end of header lines.

### LINKED ISSUE #6775: Mail headers line endings need to be changed to CRLF for PHP 8
## Description of the bug

This is the equivalent to D7's issue that was fixed in 7.103 - see Change Record: 

https://www.drupal.org/node/3491626

Issue:
https://www.drupal.org/project/drupal/issues/3319062

And commit:
https://git.drupalcode.org/project/drupal/-/commit/0e0a0915823475554e061a57b01d145d64c243a1

According to the D7 issue, PHP 8 is stricter (at least in some servers) regarding using CRLF at the end of each header line. Currently Backdrop does this in `DefaultMailSystem`:

```php
$mail_headers = join("\n", $mimeheaders);
```

that is, it adds only LF (`\n`) at the end of each header. This was done because, "some MTAs incorrectly replace LF with CRLF" (according to a comment in that class). 

So, previously, if you used CRLF (`\r\n`) at the end of each header line, you might end up with CRCRLF (`\r\r\n`). 

Using only LF (as it's currently done in Backdrop) was fine in PHP 7. But with PHP 8, you end up (in some servers, e.g. LiteSpeed) with faulty headers that look like this (notice no new line between headers - `Content-Type` and `Content-Transfer-Encoding` on the same line, etc).

Screen Shot 2024-12-13 at 3 20 29 PM

This causes issues with HTML emails, multipart emails, etc etc.

argiepiano:
Fixes https://github.com/backdrop/backdrop-issues/issues/6775

### DIFF
--- a/core/modules/system/system.mail.inc
+++ b/core/modules/system/system.mail.inc
@@ -62,7 +62,15 @@ class DefaultMailSystem implements MailSystemInterface {
     $mail_body = preg_replace('@\r?\n@', $line_endings, $message['body']);
     // For headers, PHP's API suggests that we use CRLF normally,
     // but some MTAs incorrectly replace LF with CRLF. See #234403.
-    $mail_headers = join("\n", $mimeheaders);
+    // $settings['mail_headers_line_endings'] will override this setting.
+    $headers_line_endings = settings_get('mail_headers_line_endings', "\n");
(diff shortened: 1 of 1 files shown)


quicksketch:
Merged into 1.x and 1.29.x. Thank you @argiepiano!

PR #4936, opened 2024-12-16
PoC converting to short array syntax.

### LINKED ISSUE #7191: Convert all "long" array syntax to "short" syntax, and make it the official standard?
## The Question:

In core, should we stop using `array(1, 2, 3)` in favor of `[1, 2, 3]` ?  And, should we adopt short syntax as the standard way of writing arrays in Backdrop?

Example:
```php
$x = [
  'key' => 'val',
];
function ($x = []) {}
return l('foo', 'bar', ['attributes' => ['class' => 'fun']]);
```

## Previous work on short syntax

@quicksketch has already created a working proof of concept for a rapid conversion to short syntax: https://github.com/backdrop/backdrop/pull/4936
In case the blog post he mentioned disappears one day, here is what it said to do:
```
1. Require code sniffer in your project (you are using composer, right?)
   `composer require --dev squizlabs/php_codesniffer`
2. Run CBF using the DisallowLongArraySyntax sniff
   `vendor/bin/phpcbf . --standard=Generic --sniffs=Generic.Arrays.DisallowLongArraySyntax`
That will apply to everything in the current directory, so simply change the . to whichever specific files/folders you want to affect.
```
-- from https://philipjohn.blog/2023/02/10/convert-php-arrays-to-short-array-syntax-in-seconds/

quicksketch:
Fixes https://github.com/backdrop/backdrop-issues/issues/7191

Proof of concept converting to short array syntax, following https://philipjohn.blog/2023/02/10/convert-php-arrays-to-short-array-syntax-in-seconds/

UPDATE: added link to new issue.


quicksketch:
That worked really well as a PoC, but we don't have an issue to track this yet. I just wanted to see if it was easily possible.

PR #4940, opened 2024-12-17
Issue #6302: Add multipass to dictionary.

quicksketch:
Fixes https://github.com/backdrop/backdrop-issues/issues/6302

"Multipass" is used in 5 different places in core, in reference to batch operations. I was going to change to `multi-pass` but there's a usage of `non-multipass` that would awkwardly change to `non-multi-pass`. So let's just formalize this word as acceptable.

### DIFF
--- a/.cspell/backdrop.dic
+++ b/.cspell/backdrop.dic
@@ -274,6 +274,7 @@ msword
 multibundle
 multilanguage
 multipage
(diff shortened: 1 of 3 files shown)


quicksketch:
Merged into 1.x and 1.29.x.

PR #4941, opened 2024-12-17
Issue #5865: Use SVG icon for tablesort in Seven.

### LINKED ISSUE #5865: Replace Seven tablesort icons with SVGs
As suggested [here](https://github.com/backdrop/backdrop-issues/issues/3906#issuecomment-1323636185), I'm posting this again as a new issue.

I just come across [this](https://github.com/backdrop/backdrop-issues/issues/3906) issue when I realized that the icons we are showing to sort a table now are inducing me to confusion.

Right now we are showing these icons:
131394193-4a8addda-f2cc-4142-a2d0-f6f17be3e4bc
which in fact mean the same: ascendant sorting, from the smaller line to the larger line. The only difference is that there are shown upside-down.

While these icons:
131393993-29222af1-a780-4a9f-8bda-2412557ffd79
are really showing different sorting: descendant on the left, from the larger line to the smaller line, and ascendant on the right, from the smaller line to the larger line.

Thus, I would suggest using the second set of icons, instead of the first one.

quicksketch:
Fixes https://github.com/backdrop/backdrop-issues/issues/5865


quicksketch:
Merged into 1.x and 1.32.x per RTBC status in the parent issue.

PR #4943, opened 2024-12-17
Issue 6783. Create a custom theme wrapper for link_field to allow use of states.

### LINKED ISSUE #6783: [UX] Add #states property to link_field form element
## Description of the need

I've been adding a `link_field` to a custom form (a custom block), thanks to the amazing work that happened in #6693.  I tried to make this field only visible if a checkbox was selected but `#states` doesn't work in that element type.

Checking https://docs.backdropcms.org/form_api#link_field it is not part of the spec but I don't know why not. I also can't work out where this is set

## Proposed solution

'#states' should be a valid property of `link_field`

None

Backdrop now allows states such as enabled, disabled, required, optional, visible, invisible, checked, unchecked, expanded, collapsed to be defined for a link_field element.

argiepiano:
Fixes https://github.com/backdrop/backdrop-issues/issues/6783

### DIFF
--- a/core/modules/link/link.module
+++ b/core/modules/link/link.module
@@ -771,6 +771,9 @@ function link_theme() {
     'link_field' => array(
       'render element' => 'element',
     ) + $base,
+    'link_field_wrapper' => array(
+      'render element' => 'element',
+    ) + $base,
   );
 }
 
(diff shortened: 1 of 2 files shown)


quicksketch:
Merged into 1.x for 1.30.0. Thank you @argiepiano!

PR #4910, opened 2024-11-23
Issue #6760: Prevent PHP warning for alt tags on icons.

jenlampton:
Fixes https://github.com/backdrop/backdrop-issues/issues/6760

PR #4944, opened 2024-12-19
Issue #6784: The 'alt' attribtue should be optional for icons.

### LINKED ISSUE #6784: PHP Notice: Undefined index: alt in theme_icon()
## Description of the bug

There is a small bug in icon API that is causing PHP warnings on my site. 
```
Notice: Undefined index: alt in theme_icon() (line 327 of /var/www/html/www/core/includes/icon.inc).
```

The problem is that the code is using `if ($variables['alt']) {` instead of `if (isset($variables['alt'])) {` accidentally making the `alt` attribute required. 

The documentation states that this value can be omitted:
`If not specified, the icon will automatically be assigned the aria-hidden="true" attribute.`

But the code required that it's specified, but `FALSE`. 

## Steps To Reproduce

To reproduce the behavior:
1. Install the [1.x-2.1.1 release of the On The Web module](https://github.com/backdrop-contrib/on_the_web/releases/tag/1.x-2.1.1)
2. Navigate to the settings page at `/admin/config/services/on_the_web`
3. (veryfiy that the "Icon type" is set to "SVG icons" under "Display options")

## Actual behavior

`alt` attribute required
```
Notice: Undefined index: alt in theme_icon() (line 327 of /var/www/html/www/core/includes/icon.inc).
```

## Expected behavior

`alt` attribute should be optional

## Additional information

- Backdrop CMS version: 1.29.2 
- PHP version 7.4

jenlampton:
Fixes https://github.com/backdrop/backdrop-issues/issues/6784

### DIFF
--- a/core/includes/icon.inc
+++ b/core/includes/icon.inc
@@ -324,7 +324,7 @@ function theme_icon(array $variables) {
       $svg_contents = filter_xss($svg_contents, $allowed_svg_elements);
 
       // Move the "alt" text to an attribute.
-      if ($variables['alt']) {
+      if (isset($variables['alt'])) {
         $variables['attributes']['alt'] = $variables['alt'];
       }
       else {


quicksketch:
Closing in favor of your previous PR which fixes the problem in a different way: https://github.com/backdrop/backdrop/pull/4910

PR #4947, opened 2024-12-20
Issue #6785: Fix using deprecated ajaxComplete method in installer js

### LINKED ISSUE #6785: File installer.project_list.js uses deprecated ajaxComplete method
## Description of the bug

Recently it has been discovered, that the installer module still uses a deprecated jQuery method:

https://github.com/backdrop/backdrop/blob/1.x/core/modules/installer/js/installer.project_list.js#L30

This shorthand has been deprecated in 3.3 and will (probably) be removed in upcoming jQuery 4:

https://api.jquery.com/ajaxComplete-shorthand/

## Steps To Reproduce

1. Go to /admin/modules/install or /admin/appearance/install
2. Make sure, the browser window's narrow (less than 768px wide)
3. Add a module (or theme) to trigger the event, the table header should show updated information

## Actual behavior

No problem, yet... :wink: As long as we don't use something like [jquery-migrate](https://github.com/jquery/jquery-migrate) to trigger nagging.

## Expected behavior

We should be prepared for jQuery 4.

indigoxela:
Fixes https://github.com/backdrop/backdrop-issues/issues/6785

### DIFF
--- a/core/modules/installer/js/installer.project_list.js
+++ b/core/modules/installer/js/installer.project_list.js
@@ -27,7 +27,7 @@ Backdrop.behaviors.installerProjectList = {
     }
 
     // Update the table header after any AJAX event completes.
-    $(document).ajaxComplete(function() {
+    $(document).on('ajaxComplete', function() {
       updateTableHeader();
     });
 


laryn:
Thanks @indigoxela!

PR #4958, opened 2024-12-28
#3952: Hide page title from breadcrumb when there are no crumbs.

### LINKED ISSUE #3952: breadcrumb adds page title as "last" crumb when there are no crumbs
**Description of the bug**
When I check the option `Display the current page title as the last breadcrumb` and have the breadcrumb block on them home page (when the home page is using the default layout, for example) the page title prints in the breadcrumb area, even though there are no breadcrumbs.

**Steps To Reproduce**
To reproduce the behavior:
1. Switch home page to use default layout (or disable home page layout)
2. Configure breadcrumb block on default layout: check the option `Display the current page title as the last breadcrumb`
3. Visit the home page

**Actual behavior**
The page title prints in the breadcrumb area, even though there are no breadcrumbs.
breadcrumb

**Expected behavior**
I would expect that when there are no crumbs, there is no page title.

NormPlum:
Fixes https://github.com/backdrop/backdrop-issues/issues/3952

### DIFF
--- a/core/modules/system/system.module
+++ b/core/modules/system/system.module
@@ -3436,7 +3436,7 @@ function system_header_block($settings) {
  */
 function system_breadcrumb_block($settings = array()) {
   $crumbs = backdrop_get_breadcrumb();
-  if (isset($settings['current']) && $settings['current'] == TRUE) {
+  if (isset($settings['current']) && $settings['current'] == TRUE && !empty($crumbs)) {
     $crumbs[] = '<span aria-current="page">' . backdrop_get_title() . '</span>';
   }
 


core/modules/system/system.module
```diff
@@ -3436,7 +3436,7 @@ function system_header_block($settings) {
  */
 function system_breadcrumb_block($settings = array()) {
   $crumbs = backdrop_get_breadcrumb();
-  if (isset($settings['current']) && $settings['current'] == TRUE) {
+  if (isset($settings['current']) && $settings['current'] == TRUE && !empty($crumbs)) {
```
avpaderno, line 3439: I would change the expression order, and first check the value of `$crumbs`.

```php
if (!empty($crumbs) && isset($settings['current']) && $settings['current'] == TRUE) {
```

PHP short-circuits Boolean expressions. If `$crumbs` is empty, it would not check the value of `$settings['current']`, nor would it execute the code in that `if`-block.

It could even be shorter.

```php
if (!empty($crumbs) && !empty($settings['current'])) {
```

PR #4959, opened 2024-12-29
Issue 34095: Fixes to white space problem

### LINKED ISSUE #4095: Basis & Flexible Layout Templates: Remove white space between two full width blocks placed next to each other.
**Description of the bug**
If while using a flexible layout template you put two blocks with background images or background colors that span the full width of screen, one after another, the Basis theme has a gap between them. I think it's a reasonable expectation that two hero type blocks one after another should butt up against each other. 

(EDIT - Current PR expands scope a bit to include other blocks that might also stretch across full width of page. https://github.com/backdrop/backdrop-issues/issues/4095#issuecomment-2564606960)

**Steps To Reproduce**
To reproduce the behavior:
1. Go to https://backdropcms.org/demo/create and spin up a demo site
2. Go to /admin/structure/layouts/settings and create a flexible layout template
3. Add a "hero" row and in Configure Row > Style > Row Width Bahavior choose "Always Full With of Page. 
4. Go to /admin/structure/layouts/manage/home/configure and select your flexible layout template for the home page. 
5. Manually place your blocks into the appropriate rows, make sure you have two hero blocks in your Hero row. 
6. Save settings and view home page

**Actual behavior**

You will see this gap between the two hero rows. 

Home___Flexible

**Expected behavior**
I expect the hero rows to butt up against each other. 

**Additional information**
How do we fix this without breaking existing sites that may have already made changes to fix this. 

The file core/themes/basis/css/component/hero.css has the following class, which is causing this problem. 

.block-hero {
  margin: 0 0 2rem;
}

In Tatsu I solved this problem by overriding this class with "margin-bottom: 0;"

Can we just remove this margin? Does this not risk breaking existing sites that might have already themed around this problem or have sites for which the gap is now expected?

stpaultim:
Fixes: https://github.com/backdrop/backdrop-issues/issues/4095

### DIFF
--- a/core/themes/basis/css/component/hero.css
+++ b/core/themes/basis/css/component/hero.css
@@ -112,6 +112,7 @@
 .update-1-30 .l-top .block-views-promoted-cards-block {
   margin-top: 0;
 }
+
 .update-1-30 .l-header {
   margin-bottom: 0;
 }
--- a/core/themes/basis/css/component/hero.css
+++ b/core/themes/basis/css/component/hero.css
@@ -108,6 +108,10 @@
   padding-bottom: 2rem;
 }
 
+.update-1-30 .block-hero,
+.update-1-30 .l-top .block-views-promoted-cards-block {
+  margin-top: 0;
+}
 .update-1-30 .l-header {
   margin-bottom: 0;
 }
--- a/core/themes/basis/css/component/hero.css
+++ b/core/themes/basis/css/component/hero.css
@@ -30,22 +30,6 @@
   margin-top: 0; /* If hero is in a container don't collapse top spacing */
 }
 
-.update-1-30 .block-hero:first-of-type {
-  margin-top: -2rem;
-}
-
-.update-1-30 .block-hero {
-  margin: 0;
-}
-
-.update-1-30 .no-container .block {
-  margin-top: -1rem;
-}
-
-.update-1-30 .block-system-main-menu ~ .block {
-  margin-top: 0;
-}
-
 @media (min-width: 25em) {
   .block-hero {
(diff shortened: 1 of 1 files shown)


stpaultim:
Seeing problems in sandbox that were not in local environment. Working on these.

core/themes/basis/css/component/hero.css
```diff
@@ -94,3 +94,21 @@
     font-size: 2.625em;
   }
 }
+
+/**
+ * Some changes to fix problems with extra white space between blocks in Basis.
+ * Is there a better location?
+ **/
+
+.update-1-30 .block {
```
jenlampton, line 103: this changes all blocks. Is that really what you wanted? It's going to affect everything on every page.

core/themes/basis/css/component/hero.css
```diff
@@ -94,3 +94,21 @@
...
+  margin-bottom: 0;
+  margin-top: 0;
+}
+
+.update-1-30 .container .l-flexible-row {
```
jenlampton, line 108: I don't think we should include row-specific CSS in the hero.css file. This probably belongs in `layout.css` (once we're happy with the changes).

core/themes/basis/css/component/hero.css
stpaultim: @jenlampton Yes, I explained this the main issue. I will go ahead and change the issue title. Because, I think it's very important to fix this for more than just hero blocks.

core/themes/basis/css/component/hero.css
stpaultim: Yes, I was planning to get advice on where to put this.

PR #4970, opened 2025-01-03
Issue #6802: Fix contextual outline getting covered by neighboring elements.

### LINKED ISSUE #6802: Contextual menu outline some times not visible (especially on bottom)
## Description of the bug

This issue was raised in https://github.com/backdrop/backdrop-issues/issues/5769#issuecomment-2566048851, that sometimes the contextual menu outline is not visible, often on the bottom edge, because the CSS `outline` property is sometimes calculated to be in the same location as a neighboring element.

Example 1 (Firefox):
image

Example 2 (Chromium):
image

## Steps To Reproduce

To reproduce the behavior:
1. New site with Basis installed.
2. Log in as an admin.
3. Hover over the menu block or the messages block (if present), the bottom outline is not visible.

## Solution

Seems as though we can use the fairly new [CSS `outline-offset` property](https://caniuse.com/?search=outline-offset) to fix this.

- Backdrop CMS version: 1.x (1.30.0-dev)
- Operating System and its version: Ubuntu 24.40
- Browser(s) and their versions: Firefox or Chromium

quicksketch:
Fixes https://github.com/backdrop/backdrop-issues/issues/6802

### DIFF
--- a/core/modules/contextual/css/contextual.css
+++ b/core/modules/contextual/css/contextual.css
@@ -12,6 +12,7 @@
 }
 .contextual-links-region-active {
   outline: #999 dashed 1px;
+  outline-offset: -1px;
 }
 
 /**


quicksketch:
Merged into 1.x and 1.29.x. Thanks!

PR #4971, opened 2025-01-03
Issue #6803: Style the login links in Seven.

### LINKED ISSUE #6803: Simplified login pages follow-up: Style Register and Reset Password links
## Description of the need

Every time I've shown the new simplified login links, I've had to explain that it "looks better" when there are two links, or describe a single link is in an `<ul>` tag. We should style these links so they look good even if there's only a single one.

## Proposed solution

Hide the default bullets and center the link items in Seven only. Leave contrib themes to imitate or style as desired, since this is opinionated:

One link:
image

Two links:
image

Hypothetical wrapped links:
(I couldn't think of a way to style this without resulting in a trailing/leading dot)
image

quicksketch:
Fixes https://github.com/backdrop/backdrop-issues/issues/6803

### DIFF
--- a/core/themes/seven/css/style.css
+++ b/core/themes/seven/css/style.css
@@ -1586,6 +1586,25 @@ body.maintenance-page #logo {
   margin: 0 auto;
 }
 
+/* Links for the simplified login pages. */
+.user-form-links ul {
+  list-style: none;
+  padding: 0;
+  margin: 0;
+  display: flex;
+  flex-wrap: wrap;
+  justify-content: center;
+}
+.user-form-links li {
+  display: inline;
+  margin: 0.25em 0;
+}
+.user-form-links li:not(:last-child):after {
+  content: "•";
+  margin: 0 1em;
+  color: #666;
+}
+
 ol.task-list {
   margin: 0;
   padding-left: 0; /* LTR */


core/themes/seven/css/style.css
```diff
@@ -1586,6 +1586,25 @@ body.maintenance-page #logo {
...
+.user-form-links ul {
+  list-style: none;
+  padding: 0;
+  margin: 0;
+  display: flex;
+  flex-wrap: wrap;
+  justify-content: center;
+}
+.user-form-links li {
+  display: inline;
```
indigoxela, line 1599: You make it flex and then still inline? Why both approaches?

core/themes/seven/css/style.css
```diff
@@ -1586,6 +1586,25 @@ body.maintenance-page #logo {
...
+  margin: 0.25em 0;
+}
+.user-form-links li:not(:last-child):after {
+  content: "•";
```
indigoxela, line 1603: So, you remove the dots (li) and then use a dot as separator? :wink: 

Since I read that CSS content's also read aloud by screen readers, I'm no fan of using actual characters as styling element.
We have tons of icons... nothing appropriate available?

core/themes/seven/css/style.css
```diff
@@ -1586,6 +1586,25 @@ body.maintenance-page #logo {
...
+  margin: 0 1em;
+  color: #666;
```
indigoxela, line 1605: I know, Seven has no Color support, and already "50 shades of grey", but why not using "currentColor" here?

core/themes/seven/css/style.css
```diff
@@ -1586,6 +1586,25 @@ body.maintenance-page #logo {
   margin: 0 auto;
 }
 
+/* Links for the simplified login pages. */
+.user-form-links ul {
+  list-style: none;
+  padding: 0;
+  margin: 0;
+  display: flex;
```
indigoxela, line 1594: Are you familiar with [gap](https://developer.mozilla.org/en-US/docs/Web/CSS/gap)?
I almost always use "gap" in combination with flex.

core/themes/seven/css/style.css
quicksketch: My CSS-foo is really not very good :sweat_smile: Though seems like the overall appearance is in question so we'll need to get it sorted out what is acceptable from a visual standpoint, then I'll address code feedback.

PR #4977, opened 2025-01-04
Issue #6808: Allow to specify a layout template type.

### LINKED ISSUE #6808: Allow a specified Layout template type
## Description of the need

This is one step towards https://github.com/backdrop/backdrop-issues/issues/6039

In my spare time, I've been (slowly) working on a Layout Paragraphs solution that utilizes Backdrop's Layout system rather than mimicking the Drupal way of doing it. The issue above also links to some other use cases for this. 

I'd like to be able to extend Layouts without having to completely duplicate all the functionality by being able to specify a "Layout Template Type" that can be treated separately. This allows layouts and layout templates that are intended for other uses not to get pulled in to visually pollute the core/main layouts and layout template listings.  

## Proposed solution

Core would continue to provide (and find) standard layout templates as well as flexible layout templates. Functionality would be extended by provided the expected layout template types to various functions and functionality as defaults so core continues to function as it does, but allows contrib to use many of these functions with different inputs. 

Duplicating and slightly changing every single layout-related function/class/etc to provide layout-related functionality in contrib. Duplication may still need to happen to some degree even with the above, but much less.

Backdrop now includes the ability for developers to create new Layout template types in addition to the core 'layout' and 'flexible' template types.

**Advocate: @laryn**

laryn:
Fixes https://github.com/backdrop/backdrop-issues/issues/6808


core/modules/layout/layout.module
```diff
@@ -2385,9 +2392,11 @@ function layout_get_title_description(Layout $layout) {
  *
  * @param LayoutFlexibleTemplate $flexible_template
  *   The loaded flexible template object.
+ * @param string $flexible_template_type
+ *   The type of flexible template to save. Defaults to 'flexible'.
  */
-function layout_flexible_tempstore_set(LayoutFlexibleTemplate $flexible_template) {
-  tempstore_set('layout.flexible', $flexible_template->name, $flexible_template, 604800);
+function layout_flexible_tempstore_set(LayoutFlexibleTemplate $flexible_template, $flexible_template_type = 'flexible') {
+  tempstore_set('layout.' . $flexible_template_type, $flexible_template->name, $flexible_template, 604800);
```
quicksketch, line 2426: Since this is only tempstore storage, we don't need to be compatible with the previous tempstore naming. I would suggest maybe making `$flexible_template_type = 'default'` by default, and save the tempstore location like this:

```
tempstore_set('layout.flexible.' . $flexible_template_type, $flexible_template->name, $flexible_template, 604800);
```

That makes it more clear that the tempstore entries are different kinds of flexible layout templates, and not tempstore of other layout configuration.

core/modules/layout/layout.module
```diff
@@ -2398,36 +2407,40 @@ function layout_flexible_tempstore_set(LayoutFlexibleTemplate $flexible_template
...
     return new LayoutFlexibleTemplate();
   }
 
-  $caches = &backdrop_static(__FUNCTION__, array());
-  if (!isset($caches[$flexible_template_name])) {
-    if (!$item = tempstore_get('layout.flexible', $flexible_template_name)) {
-      $item = layout_flexible_template_load($flexible_template_name);
+  $caches = &backdrop_static(__FUNCTION__ . ':' . $flexible_template_type, array());
+  if (!isset($caches[$flexible_template_type . ':' . $flexible_template_name])) {
+    if (!$item = tempstore_get('layout.' . $flexible_template_type, $flexible_template_name)) {
```
quicksketch, line 2451: Same here, make `$flexible_template_type` default to `default` and update this to:

```suggestion
    if (!$item = tempstore_get('layout.flexible.' . $flexible_template_type, $flexible_template_name)) {
```

core/modules/layout/layout.module
```diff
@@ -2437,46 +2450,63 @@ function layout_flexible_template_load($flexible_template_name = NULL) {
...
   if (empty($templates)) {
-    $cache = cache()->get('layout:flexible:config');
+    $cache = cache()->get('layout:' . $flexible_type . ':config');
     if ($cache && $cache->data) {
       $configs = $cache->data;
     }
     else {
       $configs = array();
-      $config_names = config_get_names_with_prefix('layout.flexible.');
+      $config_names = config_get_names_with_prefix('layout.' . $flexible_type . '.');
```
quicksketch, line 2500: Ah, now I see. We also get into config file names here, which would be cumbersome to rename. We could possibly do something like:

```suggestion
      $flexible_config_type = ($flexible_type === 'default' ? 'flexible' : 'flexible_' . $flexible_type);
      $config_names = config_get_names_with_prefix('layout.' . $flexible_config_type . '.');
```

Which would use the current config name for flexible templates and append the namespace to the word "flexible_" if using something else.

@laryn What kind of names are you expecting for the `$flexible_type` variable? If we left as-is, we'd need to safe-guard certain words like "layout" and "settings" to prevent config name clashes, which is why I would prefer doing name-spacing.

@indigoxela, if you have some suggestions on naming (having led our "layout" vs. "layout template"), I would love your feedback here.

quicksketch (commented):
Overall looks great! I have some questions about namespaces.

Unfortunately I don't think I can write test coverage in time for the 1.32.0 feature freeze, so this likely won't make it for this release.

core/modules/layout/layout.module
laryn: @quicksketch I added a change that processes the template type string and prefixes it with 'custom_' if it's not either 'flexible' or 'layout' (from core usage). See `_layout_check_template_type()`. I haven't changed the tempstore names or changed references to 'flexible' to be 'default' as you floated in another comment, because the template type is used in config as well, as you found here. Let me know your thoughts on the current state.

core/modules/layout/layout.module
laryn: See below.

core/modules/layout/layout.module
laryn: See below.

core/modules/layout/includes/layout.class.inc
```diff
@@ -683,7 +706,7 @@ class Layout {
     }
 
     // Add the block to the new region in the correct position.
-    if (isset($region_position)) {
+    if (isset($position)) {
```
herbdool, line 709: This appears to be a bug in core. We should probably fix it regardless of whether this PR is included.

core/modules/layout/includes/layout.class.inc
```diff
@@ -809,27 +832,27 @@ class Layout {
...
+   * @param string $layout_template_name
+   *   The new layout template name.
+   * @param array $region_mapping
    *   An array of region names, each keyed by the old region name and the value
    *   of the new region name. If empty, all known regions will be copied from
    *   one to the other. To discard the contents of a region, set the value
    *   to FALSE.
    */
-  function setLayoutTemplate($layout_name, $region_mapping = array()) {
+  public function setLayoutTemplate($layout_template_name, $region_mapping = array()) {
```
herbdool:
```suggestion
  public function setLayoutTemplate($layout_template_name, array $region_mapping = array()) {
```

While we're changing this, let's set the type.

core/modules/layout/includes/layout_flexible.class.inc
```diff
@@ -44,6 +48,24 @@ class LayoutFlexibleTemplate {
...
 
+  /**
+   * @var string
+   */
+  public $preview;
+
+  /**
+   * @var string
+   */
+  public $layout_template_type;
```
herbdool:
```suggestion
  public $layout_template_type = 'flexible';
```

Why not set it here so we don't need to add it in `__construct()`?

core/modules/layout/templates/layout--flexible.tpl.php
```diff
@@ -23,25 +23,27 @@
...
+    foreach ($row_data as $name => $row) :
+      $row_classes = array('flexible-row', 'l-' . $name);
+      if ($row['element'] == 'header' || $row['element'] == 'footer') {
+        $row_classes[] = 'l-' . $row['element'];
+      }
       ?>
       <<?php print $row['element']; ?> data-row-id="<?php print $name; ?>" class="<?php print implode(' ', $row_classes); ?>" <?php print $row['row_id']; ?>>
         <div class="<?php print $row['row_class']; ?>">
-          <?php if ($region_buttons): ?>
+          <?php if ($region_buttons) : ?>
```
herbdool: I'd prefer we don't start adding spaces in the templates. I'm assuming an IDE is being opinionated.

core/modules/layout/layout.theme.inc
```diff
@@ -370,15 +370,23 @@ function layout_preprocess_layout(&$variables) {
     }
   }
 
+  // Support deprecated 'layout_info' property until Backdrop 2.0.
+  if (isset($variables['layout_info']) && !isset($variables['layout_template_info'])) {
+    $variables['layout_template_info'] = $variables['layout_info'];
```
herbdool, line 375: Maybe this needs to be further down since you're making changes to `$variables['layout_template_info']` after this and would need those to exist for `layout_info` as well.

core/modules/layout/layout.module
```diff
@@ -817,7 +817,8 @@ function layout_theme() {
         'content' => NULL,
         'settings' => NULL,
         'layout' => NULL,
-        // @todo: Rename to layout_template_info in 2.x.
+        'layout_template_info' => NULL,
+        // @todo: Remove in favour of layout_template_info in 2.x.
```
herbdool:
```suggestion
        // @todo: Remove layout_info in favour of layout_template_info in 2.x.
```

core/modules/layout/layout.module
```diff
@@ -2571,193 +2613,193 @@ function layout_flexible_row_styles() {
       'split' => '100:0',
       'bootstrap' => '12:0',
       'region_count' => 1,
-      'name' => t('One full-width region'),
+      'name' => t('One full - width region'),
```
herbdool: Unnecessary spaces added.

core/modules/layout/layout.module
```diff
@@ -2571,193 +2613,193 @@ function layout_flexible_row_styles() {
...
     ),
     'region_10_2' => array(
       'split' => '83:17',
       'bootstrap' => '10:2',
       'region_count' => 2,
-      'name' => t('Two regions') . (' 83% : 17%'),
+      'name' => t('Two regions') . (' 83 % : 17 % '),
```
herbdool: Unnecessary spaces added. And for all other spaces before `%` below as well.

herbdool (changes requested):

core/modules/layout/includes/layout.class.inc
laryn: Sure.

core/modules/layout/includes/layout.class.inc
laryn: Yes, I think so too.

core/modules/layout/layout.module
laryn: Whoops.

core/modules/layout/layout.theme.inc
laryn: I don't think so -- here we're copying layout_info if it exists and layout_template_info does not exist. Do we need to overwrite layout_info at the end with the new value, though? I'm not sure of that.

core/modules/layout/templates/layout--flexible.tpl.php
laryn: Coding standards -- I think? Either was the IDE or the PR coding standards run. I could change them back, but if it is actually coding standards, could leave them in, too.

herbdool: I think it's probably the IDE. We should keep it consistent with the rest of the templates rather than adopt the IDE approach.

laryn: You're right, I found the place in the coding standards where it specifies no space.

laryn: I wonder if there's a way to fix that in the sniffs...

core/modules/layout/layout.module
herbdool: I think we can resolve this. The approach appears sound.

core/modules/layout/layout.module
```diff
@@ -2794,6 +2836,25 @@ function layout_field_delete_instance($instance) {
...
+/**
+ * Helper to standardize layout template type machine names and ensure we
+ * don't have any namespace issues. If the type is in the reserved list, we can
+ * use them as is, otherwise we'll ensure it starts with 'custom_' before
+ * sending it back.
+ *
+ * @param string $layout_template_type
+ *   The custom template type machine name to process, passed by reference.
+ */
+function _layout_check_template_type(string &$layout_template_type = 'flexible') {
```
herbdool:
```suggestion
function _layout_standardize_template_type(string &$layout_template_type = 'flexible') {
```

I think the function name needs to be clearer that it is making a change by reference, rather than merely "checking" things. Otherwise it's confusing that the function doesn't return anything (for example, a boolean).

herbdool: Of course, this change will have to be reflected everywhere.

laryn: Sounds good @herbdool. Thanks for reviewing. I am pushing that change now.

core/modules/layout/templates/layout--flexible.tpl.php
```diff
@@ -64,7 +66,7 @@
                       </div>
                     </div>
                   </div>
-                <?php else: ?>
+                <?php else : ?>
```
quicksketch:
```suggestion
                <?php else: ?>
```

quicksketch:
Merged into 1.x for 1.34.0. Thank you @laryn!

PR #4980, opened 2025-01-07
Issue #6810: Remove details display:table-cell

### LINKED ISSUE #6810: [UX] Improve details element display in Seven theme for Chromium based browsers
## Description of the bug
If a contrib module follows the guidance and makes the config form use `details` element to contain field groups rather than 'fieldset' then the styling in Seven makes the groups side by side. 

This styling was added in Issue #324 (PR: backdrop/backdrop/pull/3408)
The line responsible for this is:

https://github.com/backdrop/backdrop/blob/c0d50a79e39087f23dfa9763f6db0877fb151bf2/core/themes/seven/css/style.css#L784

```css
  display: table-cell;
```

And this same line isn't in the corresponding Basis element styling

@indigoxela - I wonder if you can recall the reason for this as I can't find mention in the issue.

## Steps To Reproduce

To reproduce the behavior:
1. Install Gist Filter module
2. Go to `/admin/config/content/gist-filter`
3. Alternatively, find any config form and change the fieldsets to details elements.

## Actual behavior
Field groups are side by side
image

To override this in my module, I have to do something like:

```php
backdrop_add_css('details{display:block !important;}', array('type' => 'inline'));
```
But that doesn't seem right. I could do in a file but it doesn't need to be everywhere just on the config form.

## Expected behavior
Field groups are stacked vertically
image

## Additional information

- Backdrop CMS version: 1.29.2
- PHP version: 8.2
- Database sever (MySQL or MariaDB?) and its version: N/A
- Operating System and its version: N/A
- Browser(s) and their versions: Chrome 131 and Edge 131 - not Firefox 133

yorkshire-pudding:
Fixes https://github.com/backdrop/backdrop-issues/issues/6810

### DIFF
--- a/core/themes/seven/css/style.css
+++ b/core/themes/seven/css/style.css
@@ -781,7 +781,6 @@ details {
   margin: 1em 0;
   overflow: hidden;
   background-color: #ffffff;
-  display: table-cell;
 }
 details summary span {
   color: #0074bd;


laryn:
Thank you! Merged into 1.x.

PR #257, opened 2014-05-21
Issue #197: Remove region.tpl.php.

hellomrcat:
PR for https://github.com/backdrop/backdrop-issues/issues/197

PR #4981, opened 2025-01-08
Security #257: Gadget chain / unserialization follow-up to add test coverage

quicksketch:
This is a follow-up to https://github.com/backdrop-ops/security/issues/257, adding test coverage for previously fixed issues.


quicksketch:
Merged into 1.x and 1.29.x.

PR #4986, opened 2025-01-09
Issue 6806: Improve setting to enable simplified login page, separate…

### LINKED ISSUE #6806: [UX] Improve settings for simplified login page
## Description of the bug

This is a follow-up to https://github.com/backdrop/backdrop-issues/issues/6799 which was a follow-up to https://github.com/backdrop/backdrop-issues/issues/6570.

1) we would like to add a checkbox to the `Appearance` page to enable the simplified login page for the admin theme.

In accordance with our "put the setting everywhere people might expect to find it" policy, we should duplicate the setting for `Use the administration theme for simplified login pages` by also placing this checkbox on the `Appearance` page.

2) Including "simplified login page" in radio buttons with "Use tabs" and "Use tabs" is probably not the best UI to ask and answer this question.

This is the current UI on the "Login settings" page
Screenshot 2025-01-03 at 12 48 57 PM

`Login page options` does not adequately describe the question we would like to have answered. Under this label, we have three options that do not answer the same question. I would like to propose another UI pattern for these options, as follows:

* A single checkbox that, when checked, disables the "tabs" option for the next questuion, and sets the value to "links"
    - Label: `Use simplified login pages`
    - Description: `Simplified login pages hide all layout regions such as headers, footers, and sidebars.`
* A set of radio options with "Navigation between login pages". The two options are 
    - `Display links to "Create new account", "Log in", and "Reset password" pages.`
    - `Display tabs for "Create new account", "Log in", and "Reset password" pages.`

(screenshot coming)

jenlampton:
Fixes https://github.com/backdrop/backdrop-issues/issues/6806

### DIFF
--- a/core/modules/system/system.admin.inc
+++ b/core/modules/system/system.admin.inc
@@ -246,6 +246,7 @@ function system_themes_page() {
  */
 function system_themes_admin_form($form, &$form_state, $theme_options) {
   // Administration theme settings.
+  $config = config('system.core');
   $form['admin_theme'] = array(
     '#type' => 'fieldset',
     '#title' => t('Administration theme'),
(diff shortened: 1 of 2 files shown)


core/modules/system/system.admin.inc
```diff
@@ -282,6 +296,7 @@ function system_themes_admin_form_submit($form, &$form_state) {
   }
 
   $config->set('admin_theme', $new_admin_theme);
+  $config->set('login_admin_theme', $form_state['values']['login_admin_theme']);
```
indigoxela, line 299: Only setting the config item's not sufficient here.

It's probably necessary to flush all caches, as other (cached) hooks are relevant for the simplified display.

PR #4991, opened 2025-01-10
Issue #6821 Fix hook_ckeditor_PLUGIN_plugin_check documentation [skip_tests]

### LINKED ISSUE #6821: Fix typos in hook_ckeditor_PLUGIN_plugin_check documentation
## Typos and errors in hook_ckeditor_PLUGIN_plugin_check documentation

in site/web/core/modules/ckeditor/ckeditor.api.php

```php
function hook_ckeditor_PLUGIN_plugin_check($format, $plugin_name) {
  $toolbars = $format->editor_settings['toolbar'];
  foreach ($toolbars as $toolbar['items'] as $row) {

    if (in_array('Underline', $row)) {
      return TRUE;
    }
  }
}
```

The foreach loop is completely wrong and also there are multiple `$toolbars` available. Additionally, using `$row` is confusing. `$toolbar[$x]` is the row but then there are button groups defined inside those, so I'd suggest renaming it to `$group`

Pull request incoming.

bennybobw:
The foreach loop in the documentation was incorrect and it doesn't reflect the current structure of the array. Additionally, I think we should change $row to $group because we're looping through the groups set up in each of the toolbar rows.

Fixes Issue #6821

Fixes https://github.com/backdrop/backdrop-issues/issues/6821

### DIFF
--- a/core/modules/ckeditor/ckeditor.api.php
+++ b/core/modules/ckeditor/ckeditor.api.php
@@ -205,10 +205,12 @@ function hook_ckeditor_settings_alter(array &$settings, $format) {
  * @ingroup callbacks
  */
 function hook_ckeditor_PLUGIN_plugin_check($format, $plugin_name) {
-  // Automatically enable this plugin if the Underline button is enabled.
-  foreach ($format->editor_settings['toolbar']['buttons'] as $row) {
-    if (in_array('Underline', $row)) {
-      return TRUE;
+  $toolbars = $format->editor_settings['toolbar'];
+  foreach ($toolbars as $toolbar) {
+    foreach ($toolbar as $group) {
+      if (in_array('Underline', $group['items'])) {
+        return TRUE;
+      }
     }
   }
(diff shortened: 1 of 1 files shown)


core/modules/ckeditor/ckeditor.api.php
```diff
@@ -205,10 +205,12 @@ function hook_ckeditor_settings_alter(array &$settings, $format) {
  * @ingroup callbacks
  */
 function hook_ckeditor_PLUGIN_plugin_check($format, $plugin_name) {
-  // Automatically enable this plugin if the Underline button is enabled.
-  foreach ($format->editor_settings['toolbar']['buttons'] as $row) {
-    if (in_array('Underline', $row)) {
-      return TRUE;
+  $toolbars = $format->editor_settings['toolbar'];
```
avpaderno, line 208: The comment should be preserved, since it correctly explains what the code does.

PR #4999, opened 2025-01-21
Issue #6832: Do not add language prefix to links in update.php

### LINKED ISSUE #6832: update.php: Do not add language prefix to "Skip backup" and "Cancel" link
## Description of the bug

This only affects multilingual sites, which have a language prefix set for the *default language*.

## Steps To Reproduce

1. Enable the language and locale module
2. Configure the language to use a prefix - also the default language
3. Run an update
4. When coming to the "Pre-update backup" step, click "Skip backup"

skip backup link

## Actual behavior

You end up with a 404.

```
The requested page "/en/core/update.php?op=selection&token=4_jjblabla..." could not be found.
```

## Expected behavior

Getting to the next step.

## Additional information

- Backdrop CMS version: 1.30.0 (or dev)

This happens, because url() adds the language prefix to the link.

The function chain for the skip link is:

theme_link() -> l() -> url()

And the url() function considers the language, but our default page to update can't handle that.

The quick workaround is to copy the link and strip the language prefix manually, to get to the next step. Doesn't seem ideal. 😉 

BTW: the same is true for the "Cancel" link on the next page ("Review updates") - 404.

cancel link

indigoxela:
Fixes https://github.com/backdrop/backdrop-issues/issues/6832

### DIFF
--- a/core/update.php
+++ b/core/update.php
@@ -149,9 +149,8 @@ function update_script_selection_form($form, &$form_state) {
     '#value' => t('Apply pending updates'),
   );
   $form['actions']['cancel'] = array(
-    '#type' => 'link',
-    '#href' => $_SERVER['SCRIPT_NAME'],
-    '#title' => t('Cancel'),
+    '#type' => 'markup',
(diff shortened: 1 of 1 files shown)


quicksketch:

PR #3777, opened 2021-10-01
Issue-3795 Disable scheduling options

mmachinen:
Fixes backdrop/backdrop-issues#3795
In the standard install profile, the Show option for scheduling setting for Pages should be disabled. (It should remain enabled for Posts)

PR #5000, opened 2025-01-21
#3795: Disable scheduling for Page content type.

### LINKED ISSUE #3795: [UX] Pages should not allow scheduling by default
In the standard install profile, the `Show option for scheduling ` setting for Pages should be disabled.

NormPlum:
Fixes https://github.com/backdrop/backdrop-issues/issues/3795

Replaces https://github.com/backdrop/backdrop/pull/3777

### DIFF
--- a/core/modules/node/tests/node.test
+++ b/core/modules/node/tests/node.test
@@ -737,9 +737,6 @@ class PagePreviewTestCase extends BackdropWebTestCase {
     $name = $this->randomName(8);
     $edit = array();
     $edit[$title_key] = $name;
-    $edit['status'] = 2;
-    $edit['scheduled[date]'] = format_date(REQUEST_TIME + 360, 'custom', 'Y-m-d', $web_user->timezone);
-    $edit['scheduled[time]'] = format_date(REQUEST_TIME + 360, 'custom', 'H:i:s', $web_user->timezone);
     $this->backdropPost(NULL, $edit, t('Preview'));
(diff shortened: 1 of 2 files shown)


stpaultim:
Congratuations on creating the 5000th Pull Request for Backdrop CMS.

Sorry, no prize.

core/modules/node/tests/node.test
```diff
@@ -736,9 +736,6 @@ class PagePreviewTestCase extends BackdropWebTestCase {
     $name = $this->randomName(8);
     $edit = array();
     $edit[$title_key] = $name;
-    $edit['status'] = 2;
-    $edit['scheduled[date]'] = format_date(REQUEST_TIME + 360, 'custom', 'Y-m-d', $web_user->timezone);
```
jenlampton, line 741: I wonder if we still need these tests...  I think we should change the node type here to Posts instead. Filing an updated PR.

jenlampton: Ah, I see we also have `NodePublishScheduling` and that should handle the scheduling tests!  okay, if that's the case then this PR is RTBC :)

jenlampton:
This PR needed a rebase, so I've replaced it with https://github.com/backdrop/backdrop/pull/5208.

NormPlum:
> This PR needed a rebase, so I've replaced it with #5208.

@jenlampton I could have just updated this one...

jenlampton:
The 5000 number seems worth retaining :)

PR #5003, opened 2025-01-25
Issue #6837: Prevent enabling sub theme via UI, if base theme is missing

### LINKED ISSUE #6837: Add dependency checking for themes (before enabling)
## Description of the need / bug

The module installer does a dependency check and even also downloads required modules for the selected one.

There's no such check for themes.
And it's also possible to enable a theme and set it as default if the base theme is missing, which can lead to all sorts of warnings.

Actually, at first I thought, this should be a feature request, but having a closer look: the ability to enable a sub theme if the base theme isn't there is a bug. It causes loads of PHP warnings.

## Steps to reproduce

1. Get a subtheme, but do not download its base theme
2. Go to admin/appearance
3. Enable and set that "orphan" as default
4. Go to the home page (or any other page, where the default theme applies)

Besides a broken display (depends on theme CSS), there are also loads of warnings. Either displayed, or logged.

## Proposed solution

If possible, also download the base theme when downloading a depending theme via installer (UI). That seems more like a feature request.

_At the very least_, prevent enabling a sub theme if the base theme's missing. 👈 This is a bugfix.

People have to check the theme's README or info file, to figure out, what they need.

## Additional information

That's what you see on the home page if the base theme isn't there and error display is turned on. That's massive.

Image

Backdrop version: latest dev (the problem isn't new, though)
PHP version: 8.3 on that testing site.

<s>My suspicion is, that with PHP 8.4 this might even end with a hard error.</s> It doesn't, at least not in usual setups.

indigoxela:
Fixes https://github.com/backdrop/backdrop-issues/issues/6837

### DIFF
--- a/core/modules/system/system.admin.inc
+++ b/core/modules/system/system.admin.inc
@@ -296,6 +296,17 @@ function system_theme_enable() {
 
     // Check if the specified theme is one recognized by the system.
     if (!empty($themes[$theme])) {
+      if (!empty($themes[$theme]->base_theme)) {
+        // If this is a sub theme, check if the base theme is available.
+        $base_theme = $themes[$theme]->base_theme;
+        if (!array_key_exists($base_theme, $themes)) {
+          backdrop_set_message(t('%theme is a sub theme of %base_theme, but %base_theme is missing.', array(
+            '%theme' => $themes[$theme]->info['name'],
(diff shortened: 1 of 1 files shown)


indigoxela:
Closing in favor of https://github.com/backdrop/backdrop/pull/5004

PR #5004, opened 2025-01-28
Issue #6837: Fail gracefully if a base theme is missing

### LINKED ISSUE #6837: Add dependency checking for themes (before enabling)

indigoxela:
Fixes https://github.com/backdrop/backdrop-issues/issues/6837

... and prevent enabling, if it isn't already.

### DIFF
--- a/core/includes/theme.inc
+++ b/core/includes/theme.inc
@@ -1483,7 +1483,7 @@ function theme_get_setting($setting_name, $theme = NULL) {
   $value = $configs[$theme]->get($setting_name);
   if (is_null($value)) {
     $themes = list_themes();
-    // Do not try to access settings of inexistent (base-)themes.
+    // Do not try to access settings of missing (base-)themes.
     if (!isset($themes[$theme])) {
       return;
     }
--- a/core/includes/theme.inc
+++ b/core/includes/theme.inc
@@ -108,7 +108,16 @@ function backdrop_theme_initialize() {
   $base_themes = array();
   $ancestor = $theme;
   while ($ancestor && isset($themes[$ancestor]->base_theme)) {
+    $current_theme = $themes[$ancestor];
     $ancestor = $themes[$ancestor]->base_theme;
+    if (!array_key_exists($ancestor, $themes)) {
+      // A base theme is missing, warn about the problem and skip adding.
+      watchdog('theme', '%theme is a sub theme of %base_theme, but its base theme is missing.', array(
+        '%theme' => $current_theme->info['name'],
+        '%base_theme' => $ancestor,
+      ), WATCHDOG_WARNING);
+      continue;
+    }
     $base_themes[] = $themes[$ancestor];
   }
   _backdrop_theme_initialize($themes[$theme], array_reverse($base_themes));
@@ -1474,6 +1483,10 @@ function theme_get_setting($setting_name, $theme = NULL) {
   $value = $configs[$theme]->get($setting_name);
   if (is_null($value)) {
     $themes = list_themes();
+    // Do not try to access settings of inexistent (base-)themes.
+    if (!isset($themes[$theme])) {
+      return;
+    }
     $theme_info = $themes[$theme];
     $value = isset($theme_info->info['settings'][$setting_name]) ? $theme_info->info['settings'][$setting_name] : NULL;
   }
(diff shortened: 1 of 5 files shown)


core/modules/system/system.admin.inc
```diff
@@ -140,6 +141,30 @@ function system_themes_page() {
     if (isset($themes[$theme->name]->base_themes)) {
       $theme_keys = array_keys($themes[$theme->name]->base_themes);
       $theme_keys[] = $theme->name;
+      // Make sure, we have no missing dependency.
+      foreach ($theme->base_themes as $name => $value) {
+        // A value of NULL indicates that the base theme is missing.
+        if ($value == NULL) {
+          $broken_dependencies[$theme->name][] = $name;
+          // If the theme is even enabled, warn about that problem.
```
yorkshire-pudding:
```suggestion
          // If the theme is enabled, warn about that problem.
```

yorkshire-pudding (changes requested):
Looks really good.  A small suggestion to improve readability in a comment.

indigoxela:
Random buddy go away...

yorkshire-pudding (approved):
Thanks @indigoxela

core/includes/theme.inc
```diff
@@ -1474,6 +1483,10 @@ function theme_get_setting($setting_name, $theme = NULL) {
   $value = $configs[$theme]->get($setting_name);
   if (is_null($value)) {
     $themes = list_themes();
+    // Do not try to access settings of inexistent (base-)themes.
```
```suggestion
    // Do not try to access settings of missing (base-)themes.
```

core/modules/system/system.theme.inc
```diff
@@ -103,12 +103,29 @@ function theme_system_themes_page($variables) {
...
+      if (isset($theme->info['base theme'])) {
+        $base_theme_name = $theme->info['base theme'];
+        $base_theme_label = !empty($theme->base_themes[$base_theme_name]) ? $theme->base_themes[$base_theme_name] : $base_theme_name;
+        $depends = '<br><strong>' . t('Base theme') . ':</strong> ' . check_plain($base_theme_label);
+        if (isset($theme->missing_dependencies)) {
+          $theme->classes[] = 'missing-dependency';
+          $icon = icon('warning', array(
+            'attributes' => array('width' => 24, 'height' => 24),
+          ));
+          $prefix = '<div class="prefix">' . $icon . ' ' . t('Missing dependency') . '</div>';
```
yorkshire-pudding:
```suggestion
          $prefix = '<div class="prefix">' . $icon . ' ' . t('Missing dependency') . '</div>';
```

Is there a particular reason why the variable and class is `prefix`?  I understand there is a `missing-dependency` class on the wrapper around each theme's description but would something like `warning` or `missing-warning` perhaps be a more meaningful class and variable for this part of it as this is all it is used for?

In the modules page, if a module has missing requirements, then the class around the "missing" text is `admin-missing`

indigoxela: It's `$prefix` as it's usually just an empty string, not necessarily a warning.

The CSS class might be better renamed to something "more speaking", I think, you're right re that. :+1:
Both CSS classes currently aren't actually used by Seven, but I thought it might make sense to add for easier styling (if an admin theme wants to).

quicksketch:

PR #5005, opened 2025-01-28
Issue #6839: Fix infinite loop while trying to upgrade from Drupal 7.

### LINKED ISSUE #6839: Running update as anonymous fails in infinite loop on update.php (v1.30.0).
## Description of the bug

Running `core/update.php` to upgrade from Drupal 7 to Backdrop 1.30.0 falls into an infinite loop where token validation continually fails, sending the user back to the beginning of the process (where they currently are). 

## Steps To Reproduce

To reproduce the behavior:
1. Load a prepped D7 database
2. Visit `core/update.php`
3. Try to click on `Continue` from the Overview page

---- 

UPDATE:

This can also be reproduced as follows:
1. Install Backdrop 1.30.0
2. Edit `settings.php` and set `$settings['update_free_access'] = TRUE;` to allow for anonymous updates
3. Install an older version of a module (or use a dummy module) whose current version needs an update
4. Download and install the newer version of the module you installed in step 3 - Backdrop should tell you that the database is outdated and needs to run update.php
5. **Log out**
6. Visit `core/update.php`
7. Click "Continue" and witness the grand infinite loop!----
## Actual behavior

You continue to sit on the (reloaded) Overview page.

## Expected behavior

You should be moved along to the Review Updates page (skipping the Backup page during the upgrade process).

## Additional information

It seems that with the additional check and redirect related to the Backup page, the session is destroyed during a call to `install_goto` and the token fails to match, which sends the user back to the beginning of the process -- which appears to the end user like they are stuck on the first page of the process and doesn't seem visually like the "Continue" button is actually doing anything.

As a test I uncommented the `backdrop_exit()` call within `install_goto` to prevent the session from being destroyed, and then the "Continue" button works as expected. Obviously this is not the solution but just to show that it appears to potentially validate the diagnosis.

laryn:
Fixes https://github.com/backdrop/backdrop-issues/issues/6839

### DIFF
--- a/core/update.php
+++ b/core/update.php
@@ -304,7 +304,9 @@ function update_info_page() {
     $output .= $module_status_report;
   }
 
-  $form_action = check_url(backdrop_current_script_url(array('op' => 'check_updates', 'token' => $token)));
+  // Go straight to update selection if upgrading from Drupal 7.
+  $op = (backdrop_get_installed_schema_version('system') > 7000) ? 'selection' : 'check_updates';
+  $form_action = check_url(backdrop_current_script_url(array('op' => $op, 'token' => $token)));
   $output .= '<form method="post" action="' . $form_action . '">
   <div class="form-actions">
     <input type="submit" value="Continue" class="form-submit button-primary" />
@@ -617,8 +619,7 @@ if (update_access_allowed()) {
         install_goto('core/update.php?op=results');
       }
       else {
-        // Skip the backup and go to update selection if upgrading from Drupal 7.
-        $op = update_backup_enabled() ? 'backup' : 'selection';
+        $op = 'backup';
         $token = backdrop_get_token('update');
         install_goto('core/update.php?op=' . $op . '&token=' . $token);
       }
--- a/core/update.php
+++ b/core/update.php
@@ -305,7 +305,7 @@ function update_info_page() {
   }
 
   // Go straight to update selection if upgrading from Drupal 7.
-  $op = (backdrop_get_installed_schema_version('system') > 7000) ? 'selection' : 'check_updates';
+  $op = (state_get('update_d7_upgrade', FALSE)) ? 'selection' : 'check_updates';
   $form_action = check_url(backdrop_current_script_url(array(
     'op' => $op,
     'token' => $token,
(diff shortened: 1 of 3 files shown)


core/update.php
```diff
@@ -304,7 +304,12 @@ function update_info_page() {
     $output .= $module_status_report;
   }
 
-  $form_action = check_url(backdrop_current_script_url(array('op' => 'check_updates', 'token' => $token)));
+  // Go straight to update selection if upgrading from Drupal 7.
+  $op = (backdrop_get_installed_schema_version('system') > 7000) ? 'selection' : 'check_updates';
```
indigoxela: Isn't `state_get('update_d7_upgrade')` available already? Not sure - didn't actually test it...

laryn: No, I tried that first but it's set later. We could potentially set that earlier and use it though.

indigoxela: It might be worth a try - because that's actually the purpose of this state...
If it *is* possible to set this early enough.

laryn: See what you think. I also tested locally using the state in a few places in `core/includes/update.inc` where the same `backdrop_get_installed_schema_version('system') > 7000` check is done, but that seems to cause issues with the upgrade so I reverted those and left those out rather than open a can of worms.

core/update.php
```diff
@@ -617,8 +625,7 @@ function update_check_requirements($skip_warnings = FALSE) {
         install_goto('core/update.php?op=results');
       }
       else {
-        // Skip the backup and go to update selection if upgrading from Drupal 7.
-        $op = update_backup_enabled() ? 'backup' : 'selection';
```
indigoxela, line 621: Removing this breaks hiding the backup step if `$settings['backup_directory'] = FALSE;`

Is removing this necessary to get the upgrade working?

indigoxela: I just gave it a try and reverted that single change. The upgrade still runs fine if that check is there, so it's safe to revert that.

laryn: Thank you!

indigoxela: @laryn you can close this as resolved, BTW.

laryn:
Closing in favour of https://github.com/backdrop/backdrop/pull/5007

PR #5007, opened 2025-02-02
Issue 6839. Prevent session ID from being destroyed for anonymous update

### LINKED ISSUE #6839: Running update as anonymous fails in infinite loop on update.php (v1.30.0).

argiepiano:
Fixes https://github.com/backdrop/backdrop-issues/issues/6839

### DIFF
--- a/.github/misc/settings.local.php
+++ b/.github/misc/settings.local.php
@@ -18,3 +18,6 @@ $config_directories['staging'] = 'files/config/staging';
 
 // Specify a backup directory to test backup/restore capabilities.
 $settings['backup_directory'] = 'files/backups';
+
+// Allow anonymous upgrades for the D7 upgrade tests.
+$settings['update_free_access'] = TRUE;
--- a/.github/misc/settings.local.php
+++ b/.github/misc/settings.local.php
@@ -18,6 +18,3 @@ $config_directories['staging'] = 'files/config/staging';
 
 // Specify a backup directory to test backup/restore capabilities.
 $settings['backup_directory'] = 'files/backups';
-
-// Allow anonymous upgrades for the D7 upgrade tests.
-$settings['update_free_access'] = TRUE;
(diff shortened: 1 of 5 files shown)


laryn:
Thanks @argiepiano, @indigoxela, and @quicksketch! Merged into 1.x and 1.30.x.

PR #5009, opened 2025-02-05
Issue #6841: Fix upgrade tests not failing when they should

### LINKED ISSUE #6841: Upgrade tests in 1.30.0 are supposed to fail due to a bug - they are not failing.
## Description of the bug

When the backup functionality was added, a regression was introduced that created an infinite loop in the first step of the update screens in core/update.php. This was reported in #6839. There is a fix for that there.

The upgrade tests (`FilledUpgradePathTestCase`, `BareUpgradePathTestCase`) should have also failed because of that regression, but they did not. We discovered in #6839 that those upgrade tests are being run as user 1, instead of anonymous (which is the case in a "normal" D7 upgrade when the setting `update_free_access` is TRUE).

In that issue I tried to find a way to make these tests run as anonymous, unsuccessfully. Rather than delaying the fix, I'm opening a separate issue to fix the tests so that they fail when they are supposed to.

## Steps To Reproduce

In Backdrop 1.30.0, run the tests `FilledUpgradePathTestCase` and `BareUpgradePathTestCase`. The do not fail despite of the regression discussed above.

## Expected behavior

Those tests should fail.

## Additional information

indigoxela:
Fixes https://github.com/backdrop/backdrop-issues/issues/6841

### DIFF
--- a/core/modules/layout/tests/layout.test
+++ b/core/modules/layout/tests/layout.test
@@ -2824,6 +2824,18 @@ class LayoutUpgradePathTest extends UpgradePathTestCase {
   public function testLayoutUpgrade() {
     $this->assertTrue($this->performUpgrade(), 'The upgrade was completed successfully.');
 
+    $admin_user = $this->backdropCreateUser(array(
+      'administer layouts',
+      'administer flexible templates',
+      'administer blocks',
+      'administer search',
+      'search content',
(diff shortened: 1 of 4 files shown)


indigoxela:
Doesn't work... giving up. :shrug:

PR #5012, opened 2025-02-06
Fix node_access query when an entity reference field is blank

### LINKED ISSUE #6844: Views hides nodes with a blank entity reference field
## Description of the bug

Views incorrectly filters out content where an entity reference field is blank in some specific situations. The conditions are:
- The view has a relationship configured to the Referenced Entity.
- A module is enabled that implements "node_grants".

This was originally reported in the content_access issue queue (https://github.com/backdrop-contrib/content_access/issues/30), but I believe it is actually a core issue.

## Steps To Reproduce

To reproduce the behavior:
1. Setup a Content Type with an entity reference field that targets a node.
2. Make sure Bypass content access control permission is disabled.
3. Add some content to the content type where some content includes the entity reference and some does not.
4. Create a view filtered by the content type setup earlier, displayed as fields in a table. Including the entity reference field.
5. Add a Relationship to the view for the Entity Reference - Referenced Entity. Leave the "Require this relationship" unchecked.
6. At this point, you should see all content including the nodes with and without the entity reference field populated.
7. Enable any module that implements "node_grants", for example, content_access.
8. Rebuild Node access permissions.
9. Load the view again.
10. This time you will only see the content where the entity reference field is populated. The view should include the nodes where the entity reference field is blank.

## Actual behavior

Only the content where the entity reference field is populated is displayed.

## Expected behavior

The same content shown in step 6 should be displayed.

## Additional information

- Web server and its version: Apache/2.4.58 (Ubuntu)
- PHP version: 8.3
- Database sever (MySQL or MariaDB?) and its version: 8.0.41-0ubuntu0.24.04.1
- Operating System and its version: Ubuntu 24.04

The views relationship configured to the Referenced Entity causes the following to be appended to the WHERE clause of the Views query:
```
AND ( EXISTS  (SELECT `na`.`nid` AS `nid`
FROM 
{node_access} `na`
WHERE (( (`na`.`gid` IN  ('0')) AND (`na`.`realm` = 'all') )OR( (`na`.`gid` IN  ('2')) AND (`na`.`realm` = 'content_access_author_unpublished') )OR( (`na`.`gid` IN  ('2')) AND (`na`.`realm` = 'content_access_author') )OR( (`na`.`gid` IN  ('222', '333', '337')) AND (`na`.`realm` = 'content_access_rid') ))AND (`na`.`grant_view` >= '1') AND (node_field_data_field_reference.nid = na.nid) )) 
```
This is only valid when the node_field_data_field_reference.nid is not null.

kswan:
Fixes https://github.com/backdrop/backdrop-issues/issues/6844

### DIFF
--- a/core/modules/node/node.module
+++ b/core/modules/node/node.module
@@ -3092,7 +3092,10 @@ function _node_query_node_access_alter($query, $type) {
       }
       // Otherwise attach it to the node query itself.
       else {
-        $query->exists($subquery);
+        $node_access_conditions = db_or()
+          ->isNull("$node_alias.$field")
+          ->exists($subquery);
+        $query->condition($node_access_conditions);
       }
     }
   }


kswan:
@argiepiano submitted a better PR so I will close this.
https://github.com/backdrop/backdrop/pull/5018

PR #5014, opened 2025-02-08
Issue #6845: Allow layout dialogs to be wider if space is available

### LINKED ISSUE #6845: [UI] Layout dialog needs to be wider
## Description of the bug

Something that has always bugged me: editing a hero block and all you get is such a teeny-weeny dialog to fiddle inside.
Same for flexible templates. Loads of space, but you only get this stamp-sized thingy.

It can not be that this only bothers me.

This can be reproduced by using any of the layout dialogs, like when you follow one of the contextual links from a block.

Same when editing a flexible layout template region.

## Actual behavior

Even if there's loads of space, I have to cope with this stamp:

Image

## Expected behavior

More space! 😆 Give me more space! Especially if the dialog contains an editor.

## Additional information

- Backdrop CMS version: latest dev, but the problem's old.
- PHP version: any
- Browser(s) and their versions: any

indigoxela:
Fixes https://github.com/backdrop/backdrop-issues/issues/6845

### DIFF
--- a/core/modules/layout/css/layout.admin.css
+++ b/core/modules/layout/css/layout.admin.css
@@ -161,7 +161,7 @@
 
 /* Layout dialogs. */
 .layout-dialog {
-  max-width: min(95vw, 912px);
+  max-width: min(95vw, 48em);
 }
 .layout-block-list {
   max-height: 300px;
--- a/core/modules/layout/css/layout.admin.css
+++ b/core/modules/layout/css/layout.admin.css
@@ -161,7 +161,7 @@
 
 /* Layout dialogs. */
 .layout-dialog {
-  max-width: 600px;
+  max-width: min(95vw, 912px);
 }
 .layout-block-list {
   max-height: 300px;


core/modules/layout/css/layout.admin.css
```diff
@@ -161,7 +161,7 @@
 
 /* Layout dialogs. */
 .layout-dialog {
-  max-width: 600px;
+  max-width: min(95vw, 912px);
```
olafgrabienski: @indigoxela – For curiosity: Did you choose exactly 912px to get a breakpoint at 960px? If so, I'm wondering if it makes sense to shift the breakpoint to 992px, which would correspond to a main breakpoint of the layout grid (62em).

indigoxela: The number was more or less arbitrary - switched to a more reasonable value now. :wink:

indigoxela:
Random buddy go away...

core/modules/layout/css/layout.admin.css
```diff
@@ -161,7 +161,7 @@
 
 /* Layout dialogs. */
 .layout-dialog {
-  max-width: 600px;
+  max-width: min(95vw, 48em);
```
quicksketch, line 164: For reference, we started accepting `min()`, `max()`, and `calc()` in Backdrop 1.29.0 as part of https://github.com/backdrop/backdrop-issues/issues/6526, which was a follow-up to Icon API. Though there are a number of `calc()` calls with fallbacks in Basis, which has been in place since 2016.

quicksketch:

PR #5016, opened 2025-02-11
Add description to table and descriptions to Core layout templates

### LINKED ISSUE #6850: [UX][A11y] Add layout template description to list of Layout Templates
## Description of the need

I think for all layout templates it would be good to display the description if there is one on the list of Layout Templates. The UI for Flexible Layout Templates includes a description. While the images for core templates give an idea, it would be good a clear textual description of the template for accessibility.  The `.info` files and the parser support `description` which is widely used for modules and themes, but doesn't appear in core or a few contrib I checked. 

For flexible templates, contrib and custom templates, having the description would help site architects to remember any key features that would be too wordy to put in the name.

## Proposed solution
A description column is added to the table.

Image

I have proposed descriptions for all the visible core layout templates

I looked for a hook that could allow this to be contrib but I didn't see an obvious one.

No

EDIT: Updated to reflect revised approach of putting description in own column

Backdrop now includes layout template descriptions in the list of layout templates

yorkshire-pudding:
Fixes https://github.com/backdrop/backdrop-issues/issues/6850

### DIFF
--- a/core/modules/layout/layout.admin.inc
+++ b/core/modules/layout/layout.admin.inc
@@ -3314,7 +3314,10 @@ function layout_ajax_form_update($form, $form_state) {
 function layout_settings_page($form, &$form_state) {
   $header = array(
     'layout_template_info' => array('data' => t('Template')),
-    'description' => array('data' => t('Description')),
+    'description' => array(
+      'data' => t('Description'),
+      'class' => array(RESPONSIVE_PRIORITY_LOW),
+    ),
     'type' => array(
       'data' => t('Type'),
       'sort' => 'desc',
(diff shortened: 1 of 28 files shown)


core/modules/layout/layout.admin.inc
```diff
@@ -3314,6 +3314,7 @@ function layout_ajax_form_update($form, $form_state) {
 function layout_settings_page($form, &$form_state) {
   $header = array(
     'layout_template_info' => array('data' => t('Template')),
+    'description' => array('data' => t('Description')),
```
indigoxela: I think `'class' => array(RESPONSIVE_PRIORITY_LOW),` would be beneficial for the description column. So this column collapses on narrow screens (like "type" already does).

quicksketch:
Merged into 1.x and 1.30.x. Thank you @yorkshire-pudding!

PR #5018, opened 2025-02-16
Issue 6421. Prevent node_access from hiding left-joined nodes

### LINKED ISSUE #6421: node_access filters out accessible nodes when node is left joined
## Description of the bug

From https://www.drupal.org/project/drupal/issues/1349080 and https://www.drupal.org/node/3176634

Currently, when a node table is left joined into a query, adding a node_access tag to the query filters out accessible rows from the base table (effectively acting more like an inner join). In particular, rows containing null values are incorrectly filtered out by node_access (i.e., if the base table has null entries for the node ID, node_access denies access to all users based on those null values, even though they should have no relevance to node_access checks). The most common example is a view where rows disappear from the table after adding an optional relationship to the view.

The patch in https://www.drupal.org/files/issues/2023-06-12/drupal-node_access_for_left_joins_d7.3176634-34.patch appears to work after stripping out the testing changes

## Steps To Reproduce

1. enable Views
2. Create a field referencing nodes
3. Create a node having the entity reference field empty,
4. Create a View with a relationship using the entity reference field. Do NOT select "Require this relationship".
5. Expected result is that regular users can see the node in the View. Current result is that regular users cannot see the node.

## Additional information

- Backdrop CMS version: 1.27.1

argiepiano:
Fixes https://github.com/backdrop/backdrop-issues/issues/6421


core/modules/node/tests/node_access_test/node_access_test.module
```diff
@@ -171,6 +177,95 @@ function node_access_entity_test_page() {
   return $output;
 }
 
+/**
+ * Page callback for node access join test page.
+ *
+ * This page simulates the views used for testing in the d8 version of the
+ * code, but without introducing a views dependency in d7.
```
laryn:
```suggestion
 * code, but without adding the complexity of creating views here.
```

argiepiano: Fixed.

quicksketch:
Merged into 1.x and 1.31.x. A huge thank you to @argiepiano!

PR #5017, opened 2025-02-14
Issue 6852. Allow to use Condition objects with joins

argiepiano:
Fixes https://github.com/backdrop/backdrop-issues/issues/6852

PR #5019, opened 2025-02-17
Issue #6852: Adding since docblock lines for join methods.

quicksketch:
Relates to https://github.com/backdrop/backdrop-issues/issues/6852

Follow-up PR to https://github.com/backdrop/backdrop/pull/5017

This just adds missing `@since` documentation to the various `join()` methods. Normally I add `@since` lines before committing but I missed these.


quicksketch:
Merged into 1.x and 1.30.x.

PR #5020, opened 2025-02-17
Issue #6838: Fix execution of header SQL commands on backup restore.

### LINKED ISSUE #6838: New restore in 1.30 fails with foreign key constraint error
## Description of the bug

After running update.php which included the new backup (see https://docs.backdropcms.org/documentation/creating-backups) I tried the restore. The restore failed with the message:

SQLSTATE[23000]: Integrity constraint violation: 1451 Cannot delete or update a parent row: a foreign key constraint fails 

This was the only message I could find.  The system was still  usable and I guess had not been updated.

## Steps To Reproduce

1. Create a backup folder
2. Add the line into settings.php in a 1.30 system to point to the backup folder
3. Run an update which will create a backup.  Be aware:  If there is no backup to be done, no backup is created.
4. Run example.com/core/restore.php

## Actual behavior

The following message was produced:

SQLSTATE[23000]: Integrity constraint violation: 1451 Cannot delete or update a parent row: a foreign key constraint fails 

## Expected behavior

System should be restored to prior to the update done in 3. above

## Additional information

- Backdrop CMS version: 130
- Web server and its version: Apache 2.4.62
- PHP version:  8.3.14
- Database server (MySQL or MariaDB?) and its version:  10.6.19-MariaDB
- Operating System and its version: Linux 3.10.0-962.3.2.lve1.5.81.el7.x86_64
- Browser(s) and their versions:  Firefox 134.0.2

quicksketch:
Fixes https://github.com/backdrop/backdrop-issues/issues/6838

### DIFF
--- a/core/includes/backup/backup.mysql.inc
+++ b/core/includes/backup/backup.mysql.inc
@@ -163,7 +163,7 @@ class BackupMySql extends BackupDatabase {
     $out = '';
     while ($line = $file->read()) {
       $first2 = substr($line, 0, 2);
-      $first3 = substr($line, 0, 2);
+      $first3 = substr($line, 0, 3);
 
       // Ignore single line comments. This function doesn't support multiline
       // comments or inline comments.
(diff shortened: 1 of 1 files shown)


quicksketch:
Merged into 1.x and 1.30.x. These patches have been applied to D7 Backup and Migrate module for years.

PR #5023, opened 2025-02-24
#6857: Move Standard profile setup to config.

### LINKED ISSUE #6857: Move Standard profile's setup to config
The Standard install profile does a bunch of setup tasks. Creating content types, fields, permissions, roles, views, etc. Most of these are code in `standard_install()`, but some are config in `/core/profiles/standard/config/*.json`.

I think it might be good to move everything to config. Some things like creating views can't really be done in code as there's just too much of it. Some things can't be done in config either (like adding content since it lives in the database). But I think config is easier for people to add and update since they can make the changes in the UI and then copy the config files over.

NormPlum:
Fixes https://github.com/backdrop/backdrop-issues/issues/6857


core/profiles/standard/standard.install
```diff
@@ -485,108 +139,47 @@ function standard_install() {
...
-  // Create a default role for site administrators, with all available
-  // permissions assigned.
-  $admin_role = new stdClass();
-  $admin_role->name = 'administrator';
-  $admin_role->label = 'Administrator';
-  $admin_role->description = 'Can manage content and configuration, including the ability to assign roles.';
-  $admin_role->weight = 3;
-  $admin_role->permissions = array_keys(module_invoke_all('permission'));
-  user_role_save($admin_role);
-  // Set this as the administrator role.
```
avpaderno, line 570: This code cannot be replaced by a configuration file, since this code adds all the existing permissions, except when the configuration file is edited all the times the permissions are changed.

core/profiles/standard/standard.install
```diff
@@ -598,17 +191,22 @@ function standard_install() {
...
-    ->condition('type', 'theme')
-    ->condition('name', 'seven')
-    ->execute();
-  config_set('system.core', 'admin_theme', 'seven');
-  config_set('system.core', 'node_admin_theme', '1');
+  // Enable default and admin themes.
+  // @see system_install()
+  $default_theme = 'basis';
+  $admin_theme = 'seven';
+  config('system.core')
```
avpaderno: It would be better to save the value returned from `config('system.core')` in a variable, instead of calling `config()` all the time that configuration object is necessary.

core/profiles/standard/standard.install
NormPlum: This code was replaced by a file that has all the permissions already defined (I copied the file from a new built site and put it into the Standard profiles config folder). So it should produce the same results...

core/profiles/standard/standard.install
NormPlum: Ok, good idea.

core/profiles/standard/standard.install
avpaderno: It can produce the same result now, but when a new core module is added, or permissions are changed, the result will not be the same.

NormPlum: Ah I see. I've made a chnage that keeps the config file with its changes, but then grants the permissions themselves in the install file.

PR #5024, opened 2025-02-25
Issue #6858: Improve documentation for add_where_expression function

### LINKED ISSUE #6858: Multiple add_where_expression clauses in hook_views_query_alter use same argument
## Description of the bug

I'm working on a module and encountered an odd situation. I'm attempting to create an `OR` group of clause statements using [`$query->add_where_expression()`](https://docs.backdropcms.org/api/backdrop/core%21modules%21views%21plugins%21views_plugin_query_default.inc/function/views_plugin_query_default%3A%3Aadd_where_expression/1) within [`hook_views_query_alter()`](https://docs.backdropcms.org/api/backdrop/core%21modules%21views%21views.api.php/function/hook_views_query_alter/1)
The queries are more complex, but for simplicity, the bit that is failing is substituting arguments. Take the following line:

```php
$query->add_where_expression('group_name', "node.type = :type", array(':type' => $type));
```

Now I loop through each type, but the resulting SQL only has the first $type for every condition even though the $query object after this, viewed through dpm(), looks correct and has the right arguments for each type.

The following statement works, but I thought variables should always be added into SQL using arguments. There is no user input for the source of this variable so no real risk in doing that but still seems the wrong way.

```php
$query->add_where_expression('group_name', "node.type = '$type'");
```

See also https://www.drupal.org/project/views/issues/2854660

Another [drupal issue has a comment ](https://www.drupal.org/project/views/issues/1478870#comment-5801312 ) with an `add_where_expression` built with the variables embedded in the statement.

It looks like [`views_plugin_query_default::build_condition()`](https://docs.backdropcms.org/api/backdrop/core%21modules%21views%21plugins%21views_plugin_query_default.inc/function/views_plugin_query_default%3A%3Abuild_condition/1) is where the processing happens and it doesn't look like it is handling arguments per condition.

I can see this not being a problem if the variables are from a safe source, but if they are going to take in user input then this limitation is not ideal. Also, using arguments to include variables in SQL is the recommended way to do it, so it is hard to see that this would be working as designed.
If however, this is working as designed, then it should probably be documented as such as this is an exception from the normal way of building SQL.

**NOTE: The example given here is simplistic and is probably feasible without this, but I am doing more complex WHERE statements that need to be there for a module to do its stuff. The simple example is to illustrate the bug without too many steps or dependencies.**

## Steps To Reproduce

To reproduce the behavior:
1. Create a small custom module that implements [hook_views_query_alter() ](url)
```php
/**
 * Implements hook_views_query_alter().
 */
function my_module_views_query_alter(&$view, &$query) {
  // If the base table is not a node then no alter is needed.
  if ($query->base_table != 'node') {
    return;
  }

  // Check if the view is filtered by content type.
  if (isset($view->filter['type'])) {
    $filter_content_type = $view->filter['type']->value;
  }
  else {
    $filter_content_type = array_keys(node_type_get_types());
  }
  $type_edit_permissions = array();
  foreach ($filter_content_type as $type) {
    $type_edit_permissions[$type] = (bool) user_role_has_permission('administrator', 'edit any ' . $type . ' content');
  }
  dpm($type_edit_permissions);
  $count_permissions = array_sum($type_edit_permissions);

  $query->set_where_group('OR', 'custom');

  foreach ($type_edit_permissions as $type => $edit) {
    if ($edit) {
      $query->add_where_expression('custom', "node.type = :type", array(':type' => $type));
    }
  }
  dpm($query);
}
```
2. Install Devel
3. You can adjust what role you do this for it doesn't matter, just ensure the role has some permissions. This is only to provide a simple example for reproducing the issues
4. Create a view of nodes that has either no filter for type or a few types filtered (not just one type).
5. Set view settings to show the SQL generated.
6. Preview the view
7. Check the `dpm()` objects and the SQL
## Actual behavior
While `$query` object is correct, SQL statement uses the first type for each condition

Image

```SQL
SELECT `node`.`title` AS `node_title`, `node`.`nid` AS `nid` 
FROM  {node} `node` 
WHERE (( (`node`.`status` = '1') )
AND( (node.type = 'card') OR (node.type = 'card') OR (node.type = 'card') )) 
LIMIT 10 OFFSET 0
```

## Expected behavior
Arguments work on each `add_where_expression()` function instance independently in the generated SQL statement:

```SQL
SELECT `node`.`title` AS `node_title`, `node`.`nid` AS `nid` 
FROM  {node} `node` 
WHERE (( (`node`.`status` = '1') )
AND( (node.type = 'card') OR (node.type = 'page') OR (node.type = 'post') )) 
LIMIT 10 OFFSET 0
```

## Additional information

- Backdrop CMS version: 1.30.1
- PHP version: 8.2
- Database sever (MySQL or MariaDB?) and its version: MYSQL 8.0
- Operating System and its version: linux

yorkshire-pudding:
Fixes https://github.com/backdrop/backdrop-issues/issues/6858

### DIFF
--- a/core/modules/views/plugins/views_plugin_query_default.inc
+++ b/core/modules/views/plugins/views_plugin_query_default.inc
@@ -895,8 +895,8 @@ class views_plugin_query_default extends views_plugin_query {
    *   The snippet to check. This can be either a column or a complex expression
    *   like "UPPER(table.field) = 'value'".
    * @param $args
-   *   An associative array of arguments. Where multiple instances of this
-   *   function are used on a single query, the placeholders must be unique.
+   *   An associative array of arguments. Placeholders must be unique when
+   *   using this method multiple times on a single query.
    *   @see https://www.php.net/manual/en/pdo.prepare.php
    *
    * @see QueryConditionInterface::where()
@@ -978,8 +978,8 @@ class views_plugin_query_default extends views_plugin_query {
    *   The snippet to check. This can be either a column or
    *   a complex expression like "COUNT(table.field) > 3"
    * @param $args
-   *   An associative array of arguments. Where multiple instances of this
-   *   function are used on a single query, the placeholders must be unique.
+   *   An associative array of arguments. Placeholders must be unique when
+   *   using this method multiple times on a single query.
    *   @see https://www.php.net/manual/en/pdo.prepare.php
    *
    * @see QueryConditionInterface::having()
--- a/core/modules/views/plugins/views_plugin_query_default.inc
+++ b/core/modules/views/plugins/views_plugin_query_default.inc
@@ -897,8 +897,8 @@ class views_plugin_query_default extends views_plugin_query {
    * @param $args
    *   An associative array of arguments. Placeholders must be unique when
    *   using this method multiple times on a single query.
-   *   @see https://www.php.net/manual/en/pdo.prepare.php
    *
+   * @see https://www.php.net/manual/en/pdo.prepare.php
    * @see QueryConditionInterface::where()
    */
   function add_where_expression($group, $snippet, $args = array()) {
@@ -980,9 +980,9 @@ class views_plugin_query_default extends views_plugin_query {
    * @param $args
    *   An associative array of arguments. Placeholders must be unique when
    *   using this method multiple times on a single query.
-   *   @see https://www.php.net/manual/en/pdo.prepare.php
    *
-   * @see QueryConditionInterface::having()
+   * @see https://www.php.net/manual/en/pdo.prepare.php
+   * @see QueryConditionInterface::where()
    */
   function add_having_expression($group, $snippet, $args = array()) {
     // Ensure all variants of 0 are actually 0. Thus '', 0 and NULL are all
--- a/core/modules/views/plugins/views_plugin_query_default.inc
+++ b/core/modules/views/plugins/views_plugin_query_default.inc
@@ -895,7 +895,9 @@ class views_plugin_query_default extends views_plugin_query {
    *   The snippet to check. This can be either a column or a complex expression
    *   like "UPPER(table.field) = 'value'".
    * @param $args
-   *   An associative array of arguments.
+   *   An associative array of arguments. Where multiple instances of this
+   *   function are used on a single query, the placeholders must be unique.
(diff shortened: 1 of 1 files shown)


core/modules/views/plugins/views_plugin_query_default.inc
```diff
@@ -895,7 +895,9 @@ class views_plugin_query_default extends views_plugin_query {
    *   The snippet to check. This can be either a column or a complex expression
    *   like "UPPER(table.field) = 'value'".
    * @param $args
-   *   An associative array of arguments.
+   *   An associative array of arguments. Where multiple instances of this
+   *   function are used on a single query, the placeholders must be unique.
```
argiepiano: I'd suggest rephrasing this for clarity:

Placeholders must be unique when using this method multiple times on a single query.

core/modules/views/plugins/views_plugin_query_default.inc
```diff
@@ -895,7 +895,9 @@ class views_plugin_query_default extends views_plugin_query {
    *   The snippet to check. This can be either a column or a complex expression
    *   like "UPPER(table.field) = 'value'".
    * @param $args
-   *   An associative array of arguments.
+   *   An associative array of arguments. Placeholders must be unique when
+   *   using this method multiple times on a single query.
+   *   @see https://www.php.net/manual/en/pdo.prepare.php
```
avpaderno: This `@see` tag should be placed together the next one.

core/modules/views/plugins/views_plugin_query_default.inc
```diff
@@ -976,7 +978,9 @@ class views_plugin_query_default extends views_plugin_query {
    *   The snippet to check. This can be either a column or
    *   a complex expression like "COUNT(table.field) > 3"
    * @param $args
-   *   An associative array of arguments.
+   *   An associative array of arguments. Placeholders must be unique when
+   *   using this method multiple times on a single query.
+   *   @see https://www.php.net/manual/en/pdo.prepare.php
```
avpaderno: Also this `@see` tag should be placed together the next one.

quicksketch:
Merged into 1.x and 1.30.x. Thank you @yorkshire-pudding!

PR #5026, opened 2025-02-27
Swap out custom code with image_style_url function Issue #6860

### LINKED ISSUE #6860: Thumbnails broken in image library
## Thumbnails are broken in the image library

We have a site that uses the s3fs module and because `views_handler_field_file_uri`  uses custom code to try to get the image style rather than the image module's `image_style_url` function, all the thumbnails in the image library are broken. Pull request incoming.

## Steps To Reproduce

To reproduce the behavior:
1. Turn on and configure the s3fs module
2. Add an image
3. Open the image browser. The thumnbails will all be broken.

## Actual behavior

Thumbnails are broken.

## Expected behavior

Thumbnails should show.

## Additional information

Related to Issue #3386

bennybobw:
There is a bunch of custom code in the render function of views_handler_field_file_uri.inc that attempts to generate the image style url. This code breaks when you're using modules such as s3fs which doesn't use the site default file path. I'm swapping that code out instead with image_style_url which fixes the problem.

Fixes [Issue 6860](https://github.com/backdrop/backdrop-issues/issues/6860)

### DIFF
--- a/core/modules/file/views/views_handler_field_file_uri.inc
+++ b/core/modules/file/views/views_handler_field_file_uri.inc
@@ -40,17 +40,9 @@ class views_handler_field_file_uri extends views_handler_field_file {
   function render($values) {
     $data = $values->{$this->field_alias};
     if (!empty($this->options['file_download_path']) && $data !== NULL && $data !== '') {
-      $data = file_create_url($data);
-      if (!image_is_svg($data) && !empty($this->options['image_style'])) {
-        // $data contains url of image.
-        // Get public file system path and its length.
-        $file_public_path = '/' . config_get('system.core', 'file_public_path');
-        $file_public_path_strlen = strlen($file_public_path);
-        // Find insertion position of public file system path.
-        $insert_at = strpos($data, $file_public_path);
-        // Text to insert in order to modify url.
-        $add_to_url = '/styles/' . $this->options['image_style'] . '/public';
-        $data = substr_replace($data, $add_to_url, $insert_at + $file_public_path_strlen, 0);
+      $url = file_create_url($data);
+      if (!image_is_svg($url) && !empty($this->options['image_style'])) {
+        $url = image_style_url($this->options['image_style'], $data);
       }
     }
     return $this->render_link($data, $values);


core/modules/file/views/views_handler_field_file_uri.inc
```diff
@@ -40,17 +40,9 @@ class views_handler_field_file_uri extends views_handler_field_file {
...
-        $file_public_path = '/' . config_get('system.core', 'file_public_path');
-        $file_public_path_strlen = strlen($file_public_path);
-        // Find insertion position of public file system path.
-        $insert_at = strpos($data, $file_public_path);
-        // Text to insert in order to modify url.
-        $add_to_url = '/styles/' . $this->options['image_style'] . '/public';
-        $data = substr_replace($data, $add_to_url, $insert_at + $file_public_path_strlen, 0);
+      $url = file_create_url($data);
+      if (!image_is_svg($url) && !empty($this->options['image_style'])) {
+        $url = image_style_url($this->options['image_style'], $data);
```
indigoxela, line 45: In the (unlikely case) that the image module's off, this will fail with "Call to undefined function...". At least, that's what I assume.
The file module's a required one, but the image module isn't.

bennybobw: The file already requires the image module to be on because for the options form it calls `image_style_options()`

I can add conditionals on all that and then in the render section check to see if the module is enabled, write a message to watchdog if it's not and return just the file path.

1. Does that sound like the proper approach?
2. Do I need something for the automated tests to test both cases with the image module on and off?

indigoxela: > The file already requires the image module to be on because for the options form...

That's a good point. How would we get a setting for `$this->options['image_style']`, anyway, without the image module available. So my first impression, that this odd code exists to avoid image module dependency seems wrong. A check would make more sense in function options_form().

> Do I need something for the automated tests to test both cases with the image module on and off?

It's an edge case, I guess, but if you're willing to dig into that, it would be awesome. No requirement, though, as it's unrelated to the broken thumbnail problem. You decide. :wink:

PR #5028, opened 2025-03-05
Issue #6864: Fix query string to export custom blocks.

### LINKED ISSUE #6864: Dropbutton link from reusable custom block to "Export" doesn't work
## Description of the bug

When viewing the "Custom blocks" admin page, each of them has an "Export" link in the dropbutton operations, but it goes to the export page with an error message:

## Steps To Reproduce

To reproduce the behavior:
1. Make sure "Reusable custom blocks" module is enabled.
2. Create a custom block.
3. Click on the dropbutton and select "Export" on `admin/structure/block`

## Actual behavior

Image

## Expected behavior

The "group" value in the query string should be "Custom blocks" rather than "Reusable custom blocks" and I think it may work.

laryn:
Fixes https://github.com/backdrop/backdrop-issues/issues/6864

### DIFF
--- a/core/modules/block/block.admin.inc
+++ b/core/modules/block/block.admin.inc
@@ -50,7 +50,7 @@ function block_admin_list() {
         'title' => t('Export'),
         'href' => 'admin/config/development/configuration/single/export',
         'query' => array(
-          'group' => 'Reusable custom blocks',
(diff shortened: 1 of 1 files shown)


quicksketch:
Merged into 1.x and 1.30.x. Thanks @laryn!

PR #5033, opened 2025-03-21
Issue #6873: Fix system_time_zone() listing.

### LINKED ISSUE #6873: 'formats.long.pattern.php' is not in the config 'system.date' but is called for in system_time_zones()
## Description of the bug

The function `system_time_zones()` contains this line:

```
      $zones[$zone] = t('@zone: @date', array('@zone' => t(str_replace('_', ' ', $zone)), '@date' => format_date(REQUEST_TIME, 'custom', config_get('system.date', 'formats.long.pattern.php') . ' O', $zone)));
```

However, the config file `system.date.json` doesn't contain an entry for `'formats.long.pattern.php'` so that call to `config_set()` returns an empty result.

It doesn't seem to cause a problem in generating the list of time zones (found, for example, on the user account edit page), but it does seem like it shouldn't be there or should perhaps be something else.

bugfolder:
Fixes https://github.com/backdrop/backdrop-issues/issues/6873.


quicksketch:
Merged into 1.x and 1.30.x. Thanks!

PR #5036, opened 2025-03-22
Issue 6075. Remove hard-coded entity creation.

### LINKED ISSUE #6075: [DX] Remove direct use of `new Node()` and other direct instantiations of entity classes from the codebase
When Backdrop partially incorporated D7 Entity API into core, and switched to using named classes for its core entities (that is, `Node` and `User` etc instead of `stdClass`), this allowed developers to extend those entity class (and their entity controllers) to incorporate additional functionalities and properties.  This can be done very simply and efficiently, by (1) implementing a `hook_entity_info_alter()` to change the entity class and the entity controller, and (2) extending the entity class and/or the entity controller class with the new functionality or properties.

Unfortunately, at the same time, a few places in Backdrop codebase instantiate those core entity classes by making direct calls, as in `$node = new Node();`. This effectively curtails the new possibilities created by the use of named classes as described above. So, if you are a developer who wants to transform/expand nodes in new ways, you are limited by those few direct instantiations (so, in those places, the resulting entity of calling `$node = new Node();` will use the core class, instead of the class you created to extend nodes).

The solution here is to replace those few `new Node()` and similar calls, with the intermediary API function `entity_create('node');`. This call will take care of instantiating the entity class as defined in the  entity_info_alter hook.

There are very few direct instantiations of core entity classes in the codebase:
- 7 for Node()
- 4 for User()
- 22 for File()
- 2 for Comment()
- 6 for TaxonomyTerm()

Many of those are in tests, which we can simply leave alone, as we want to use the core classes in tests.

argiepiano:
Fixes https://github.com/backdrop/backdrop-issues/issues/6075

### DIFF
--- a/core/modules/simpletest/tests/file.test
+++ b/core/modules/simpletest/tests/file.test
@@ -377,6 +377,7 @@ class FileValidatorTest extends BackdropWebTestCase {
    * Test the file_validate_extensions() function.
    */
   function testFileValidateExtensions() {
+    /** @var File $file */
     $file = entity_create('file', array('filename' => 'asdf.txt'));
     $errors = file_validate_extensions($file, 'asdf txt pork');
     $this->assertEqual(count($errors), 0, 'Valid extension accepted.', 'File');
@@ -447,6 +448,7 @@ class FileValidatorTest extends BackdropWebTestCase {
    */
   function testFileValidateNameLength() {
     // Create a new file entity.
+    /** @var File $file */
     $file = entity_create('file', array());
 
     // Add a filename with an allowed length and test it.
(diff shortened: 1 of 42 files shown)


core/modules/simpletest/tests/file.test
```diff
@@ -364,11 +364,11 @@ class FileValidatorTest extends BackdropWebTestCase {
   function setUp() {
     parent::setUp(array('file'));
 
-    $this->image = new File(array());
+    $this->image = entity_create('file', array());
```
indigoxela, line 379: Not a big deal, as it's a test, but there's no var comment... It should probably go to the definition section above.
```
  protected $image;
  protected $non_image;
  protected $file;
```
@argiepiano what do you think? Add it or not?

argiepiano: Thanks, @indigoxela. Added!

quicksketch:
Merged into 1.x and 1.31.x. Thanks @argiepiano for the PR and your patience.

PR #5037, opened 2025-03-25
Issue #6874: Fix initialization of major version in Project Installer.

### LINKED ISSUE #6874: Project Installer can't handle multiple release versions when one has major version 0.
## Description of the bug

If a project has multiple release versions and one of them has the major version of 0 (e.g., version 0.9.1), then the installer will select that version for installation, even if the recommended version has a higher major version.

## Steps To Reproduce

To reproduce the behavior:

Observe that theme [Decanter v7](https://backdropcms.org/project/stanford_decanter) has major version releases 0, 1, and 2, and that major version 2 is the recommended major version:

Image

Enable the Project Installer. 
Go to Admin > Appearance > Install new theme.
Select Decanter v7 for installation.
Click "Install".
See which version is offered for installation.

## Actual behavior

On the "Select versions" page, only version 0.5.1 will be offered.

Image

## Expected behavior

Version 2.0.1 should have been offered for installation, since it is the recommended version.

## Additional information

This is because of an interaction between the project installer and the project updater, which is used by the former.

In file `installer.pages.inc`, starting at line 246, we see this:

```
    // We use the update module to calculate the recommended version.
    $project_data = array(
      'existing_major' => 0,
      'existing_version' => 0,
      'install_type' => '',
    );
    module_load_include('inc', 'update', 'update.compare');
    update_calculate_project_update_status($project_data, $release_data);
```

Note that `existing_major` is initialized to zero. However, in `update_calculate_project_update_status()`, we have this starting at line 544:

```
  if (in_array($existing_major, $supported_majors)) {
    // Still supported, stay at the current major version.
    $target_major = $existing_major;
  }
  elseif (isset($available['recommended_major'])) {
    // Since 'recommended_major' is defined, we know this is the new XML
    // format. Therefore, we know the current release is unsupported since
    // its major version was not in the 'supported_majors' list. We should
    // find the best release from the recommended major version.
    $target_major = $available['recommended_major'];
    $project_data['status'] = UPDATE_NOT_SUPPORTED;
  }
```

Since `$supported_majors` contains version 0, `$target_major` is set to `$existing_major`, i.e., it remains at version 0, rather than to the recommended major.

As a workaround, in that initialization in `installer.pages.inc`, we could set `'existing_major' => -1`, i.e., to a nonexistent version, and then if we do that, we will end up with the recommended version:

Image

That seems like it could be a fix to the problem.

bugfolder:
Fixes https://github.com/backdrop/backdrop-issues/issues/6874.

### DIFF
--- a/core/modules/installer/installer.pages.inc
+++ b/core/modules/installer/installer.pages.inc
@@ -243,9 +243,11 @@ function installer_browser_installation_select_versions_form($form, &$form_state
       continue;
     }
 
-    // We use the update module to calculate the recommended version.
+    // We use the update module to calculate the recommended version. Initialize
+    // with a nonexistent major version so that update module won't accidentally
+    // keep an older major version.
     $project_data = array(
-      'existing_major' => 0,
+      'existing_major' => -1,
       'existing_version' => 0,
       'install_type' => '',
     );


quicksketch:
Merged into 1.x and 1.30.x. Thank you @bugfolder!

PR #5040, opened 2025-04-04
Issue #6876: Use consistent inline variable type hinting.

### LINKED ISSUE #6876: Type hint returned variables with consistent `/** @var */` format
## Description of the bug

Backdrop core is not consistent when type hinting variables via inline docblocks. For example:

```
    /* @var Block $block */
    $block = $form_state['block'];
    /* @var Layout $layout */
    $layout = $form_state['layout'];
```

Versus:

```
    /** @var Block $block */
    $block = $form_state['block'];
    /** @var Layout $layout */
    $layout = $form_state['layout'];
```

It appears that using either format works in some IDEs, like PHPStorm (or other JetBrains products), but only the second format works in VSCode.

When working correctly, an IDE should be able to autocomplete methods and properties based on the type hint from the docblock.

Image

## Steps To Reproduce

To reproduce the behavior:
1. Use PHPStorm or VSCode IDE.
2. Find usages of `/* @var` and `/** @var` in the code base.
3. Validate which situation allows autocomplete on the methods and properties, as well as tracing to the source code via `control + click` (or `Command + click` on Mac).

## Actual behavior

VSCode is not auto-completing when using the `/* @var Block $block */` syntax.

## Expected behavior

Methods and properties should autocomplete in VSCode and other IDEs.

quicksketch:
Fixes https://github.com/backdrop/backdrop-issues/issues/6876

### DIFF
--- a/core/includes/authorize.inc
+++ b/core/includes/authorize.inc
@@ -318,7 +318,7 @@ function authorize_get_filetransfer($backend, $settings = array()) {
       $file = $backend_info['file path'] . '/' . $backend_info['file'];
       require_once $file;
     }
-    /* @var FileTransfer $class */
+    /** @var FileTransfer $class */
     $class = $backend_info['class'];
     if (class_exists($class)) {
       $filetransfer = $class::factory(BACKDROP_ROOT, $settings);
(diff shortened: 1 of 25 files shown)


quicksketch:
Merged into 1.x and 1.30.x.

PR #5042, opened 2025-04-07
Issue #6868: Defensive coding when checking for book node type

### LINKED ISSUE #6868: Type error in book.module when trying to add story or other content
## Description of the bug

With the most recent update of Backdrop, I'm getting a PHP error when trying to add content (of any kind) if the Book module is enabled. Disabling the Book module restores proper function (but other warnings are in the log even after making this change).

## Steps To Reproduce

To reproduce the behavior:
If book.module is enabled
1. Go to Dashboard>Create new content
2. Click on 'Add new Story' (https://site.org/node/add/story) or other content type (same behavior with Add new Page)

I was not having this problem earlier, but there has been at least one update of Backdrop core since I last tried creating new content.

## Actual behavior

System displays grey error screen with error text. Error message taken from log reads:

> `TypeError: array_filter(): Argument #1 ($array) must be of type array, null given in array_filter() (line 1147 of /[...]/docroot/core/modules/book/book.module).`

## Expected behavior

Should open content creation page. This works properly once book.module is disabled.

## Additional information
Even though I was able to add content after disabling book.module, I am continue to get these warnings in the log file:

> `Warning: Array to string conversion in include() (line 52 of /[...]/docroot/themes/d7compatible/templates/field.tpl.php).`

It's not clear if this is related to the error above. I'm not sure when these started, because my log files are filled with them back to the earliest retained entry (6 March). There is an error message relating to each of my active themes (d7compatible, and a customized version of Analytic), apparently one of each referencing every node on the site -- even if they are not being called up. This doesn't appear to harm the functionality of the site, but it is odd and clogs my log file.

- Backdrop CMS version: 1.30.1
- Web server and its version: Apache 2
- PHP version: 8.3.13
- Database sever (MySQL or MariaDB?) and its version: MariaDB 10.5.22
- Operating System and its version: AlmaLinux 9
- Browser(s) and their versions: Firefox macOS latest

indigoxela:
Fixes https://github.com/backdrop/backdrop-issues/issues/6868

### DIFF
--- a/core/modules/book/book.module
+++ b/core/modules/book/book.module
@@ -1137,15 +1137,19 @@ function book_toc($bid, $depth_limit, $exclude = array()) {
 /**
  * Determine if a given node type is in the list of types allowed for books.
  *
- * @param $type
+ * @param string $type
  *   A node type.
  *
- * @return
- *   A Boolean TRUE if the node type can be included in books; otherwise, FALSE.
+ * @return bool
+ *   TRUE if the node type can be included in books, FALSE otherwise.
  */
 function book_type_is_allowed($type) {
-  $book_allowed_types = array_filter(config_get('book.settings', 'book_allowed_types'));
-  return in_array($type, $book_allowed_types);
+  $book_allowed_types = config_get('book.settings', 'book_allowed_types');
+  if (!empty($book_allowed_types)) {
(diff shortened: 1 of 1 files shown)


quicksketch:
Merged into 1.x and 1.30.x. Thank you!

PR #5054, opened 2025-04-23
Issue #6886: Change the default value for the backdrop_add_css() $options parameter

### LINKED ISSUE #6886: Change the default value for the backdrop_add_css()/backdrop_add_js() $options parameter
[`backdrop_add_css()`](https://docs.backdropcms.org/api/backdrop/core%21includes%21common.inc/function/backdrop_add_css/1) uses default values for its parameters:

- `NULL` for `$data`
- `NULL` for `$options`

The default value for `$data` makes sense when `backdrop_add_css()` is called without arguments, which is what [`backdrop_get_css()`](https://docs.backdropcms.org/api/backdrop/core%21includes%21common.inc/function/backdrop_get_css/1) does.

```php
if (!isset($css)) {
  $css = backdrop_add_css();
}
```

`$options` is expected to be an array or a string. The default value is only used by the following code.

```php
if (isset($options)) {
  if (![is_array](http://php.net/is_array)($options)) {
    $options = array('type' => $options);
  }
}
else {
  $options = array();
}
```

### Proposed change

- Change the default value for `$options` to an empty array
- Change the code that checks `$options` to the following one
  ```php
  if (isset($options) && is_string($options)) {
    $options = array('type' => $options);
  }
  ```

The same change should be done for [`backdrop_add_js()`](https://docs.backdropcms.org/api/backdrop/core%21includes%21common.inc/function/backdrop_add_js/1).

avpaderno:
Fixes https://github.com/backdrop/backdrop-issues/issues/6886


yorkshire-pudding (approved):
These changes help the code to match what the comments say and simplify the code.

core/includes/common.inc
```diff
@@ -3630,13 +3630,8 @@ function backdrop_add_css($data = NULL, $options = NULL) {
...
 
   // Construct the options, taking the defaults into consideration.
-  if (isset($options)) {
-    if (!is_array($options)) {
-      $options = array('type' => $options);
-    }
-  }
-  else {
-    $options = array();
+  if (isset($options) && !is_array($options)) {
```
avpaderno: This line could be probably changed to `if (isset($options) && is_string($options)) {`. The `'type'` key is expected to be a string, not an object nor a resource, for example.

PR #5057, opened 2025-04-24
Issue #6888: Fixing dropdown FOUC, showing when JS has processed the …

### LINKED ISSUE #6888: [UX] Dropdowns in core don't appear when the JS has processed resulting in long delay
## Description of the bug

The dropdown CSS has an animation timer for situations where the JS isn't working, and that's the only time the dropdowns are appearing. They were intended to appear when JS had processed the dropdowns OR when the timer ran out, whichever came first.

## Steps To Reproduce

Load a page with menu dropdowns from core (`menu-toggle.theme.breakpoint-queries.css`) see that it always takes 3 seconds for them to pop in, on most browsers with JS working it should be almost instantly.

(Pretty sure I already have the fix)

wesruv:
…menu or after 3s

Fixes https://github.com/backdrop/backdrop-issues/issues/6888

### DIFF
--- a/core/modules/system/css/menu-toggle.theme.breakpoint-queries.css
+++ b/core/modules/system/css/menu-toggle.theme.breakpoint-queries.css
@@ -17,3 +17,39 @@
 .menu-toggle-state:not(:checked) ~ .menu {
   display: block;
 }
+
+@keyframes hide-dropdowns-until-ready {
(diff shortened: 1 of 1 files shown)


wesruv:
Added a link in a dropdown to the demo site on the Home link

PR #5062, opened 2025-05-01
Add another (layout) path

### LINKED ISSUE #1528: Allow a layout to be selected on additional paths.

jenlampton:


jenlampton:
Replaced by https://github.com/backdrop/backdrop/pull/5063

PR #5066, opened 2025-05-06
Issue #4984 Prevent duplicate locale prefixes

### LINKED ISSUE #4984: Language prefixes are being repeated in Link module fields

Dawnthorn:
The link module calls url() twice on the the same URL. If the site has local enabled and is setup to detect the language with URL prefixes, then it will add the prefix a second time to a path that already has the language prefix. For example, it will rewrite fr/node/16 to fr/fr/node/16. Having the locale module not add the prefix again if it's already there, fixes this.

Fixes: https://github.com/backdrop/backdrop-issues/issues/4984

### DIFF
--- a/core/includes/locale.inc
+++ b/core/includes/locale.inc
@@ -486,7 +486,12 @@ function locale_language_url_rewrite_url(&$path, &$options) {
       case LANGUAGE_NEGOTIATION_URL_PREFIX:
         $prefixes = locale_language_negotiation_url_prefixes();
         if (!empty($prefixes[$options['language']->langcode])) {
-          $options['prefix'] = $prefixes[$options['language']->langcode] . '/';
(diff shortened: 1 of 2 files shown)


wesruv:
I pulled this into the site we were experiencing issues on and it fixed the issue 😄

quicksketch:
Merged into 1.x and 1.31.x Thank you @Dawnthorn!

PR #5068, opened 2025-05-06
Issue #6895: Reset dimensions in dialog when image is removed

### LINKED ISSUE #6895: Inline images - width/height do not get reset when uploading a new image
## Description of the bug

When uploading images via editor dialog, width and height are empty initially. They get populated when re-opening the dialog, though.

This leads to a bug when replacing an existing image by a fresh upload.

## Steps To Reproduce

1. Create a (post) node, insert an image (like from the image library)
2. Click on the image and re-open the dialog
3. Click on the remove button and upload a new image (with different aspect ratio)
4. Insert the image into the editor area

## Actual behavior

The previous image's dimensions stick, which leads to a squeezed/stretched image - the size doesn't fit the actual image.

## Expected behavior

A fresh upload should use the correct image dimensions - the ones from the current image.

## Additional information

Discovered while testing for #2658, but it's unrelated.

- Backdrop CMS version: up until latest dev

indigoxela:
Fixes https://github.com/backdrop/backdrop-issues/issues/6895

### DIFF
--- a/core/modules/filter/js/filter.js
+++ b/core/modules/filter/js/filter.js
@@ -303,6 +303,15 @@ Backdrop.behaviors.editorImageLibrary = {
         var $submit = $form.find('.form-actions input[type=submit]:first');
         $submit.trigger('mousedown').trigger('click').trigger('mouseup');
       });
+
+    // Empty width and height input fields, when an existing file is removed,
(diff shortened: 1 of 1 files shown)


quicksketch:
Merged into 1.x for 1.30.0. Thanks @indigoxela!

PR #5072, opened 2025-05-09
Prevent field instance on entity if machine name clashes with entity_key

### LINKED ISSUE #6900: Pre-existing field with reserved entity machine name re-used on entity type causes fatal error when adding entity
## Description of the bug

Backdrop does a great job of preventing use of field_name machine name for new fields once Paragraphs is installed, whether for paragraph types or for nodes. However, if you've already created a field with machine name of field_name then you are not prevented from re-using that field on a paragraph type.

This was [first raised in the paragraphs issue queue](https://github.com/backdrop-contrib/paragraphs/issues/165) but actually is wider

## Steps To Reproduce

To reproduce the behavior:

1. Start with a site without Paragraphs
2. Add a field to a content type with label of "Name" and allow the field machine name to be saved as field_name
3. Install Paragraphs
4. Create new paragraph type
5. Add the field_name field to the paragraph type
6. Add a paragraphs field to a content type
7. Create a new item of that content type
8. Enter some data (not paragraph) and save
9. Re-edit and add the paragraph
10. Save the node

## Actual behavior
Fatal error:
> TypeError: Cannot access offset of type string on string in _field_invoke() (line 193 of /var/lib/tugboat/backdrop/core/modules/field/field.attach.inc).

Image

## Expected behavior

Node saves, but actually, re-using an existing field with field_name on a paragraph type should be prevented at Step 5

Image
Screenshot from PR change

## Additional information

- Backdrop CMS version: 1.30.2
- PHP version:  n/a

PR: https://github.com/backdrop/backdrop/pull/5072

yorkshire-pudding:
Fixes https://github.com/backdrop/backdrop-issues/issues/6900

### DIFF
--- a/core/modules/field/field.crud.inc
+++ b/core/modules/field/field.crud.inc
@@ -513,16 +513,6 @@ function field_create_instance($instance) {
     throw new FieldException($message);
   }
 
-  // Check to ensure that the field instance does not have a reserved word for
-  // this entity type.
-  $entity_info = entity_get_info($instance['entity_type']);
-  if (in_array($instance['field_name'], $entity_info['entity keys'])) {
-    throw new FieldException(t("You cannot add the existing field with the machine name '@name' to a @type as this machine name is reserved by this entity type", array(
-      '@name' => $instance['field_name'],
-      '@type' => $entity_info['label'],
-    )));
-  }
-
   // Give the instance a default label.
   if (strlen($instance['label']) === 0) {
     $instance['label'] = $instance['field_name'];
@@ -642,6 +632,27 @@ function field_validate_instance($instance, $update = FALSE, $load_field = TRUE)
     // @todo: Check that the widget type is known and can handle the field type.
     // @todo: Check that the formatters are known and can handle the field type.
     // @todo: Check that the display display modes are known for the entity type.
+  }
+
+  // Check to ensure that the field instance does not have a reserved word for
+  // this entity type.
+  $entity_info = entity_get_info($instance['entity_type']);
+  if (in_array($instance['field_name'], $entity_info['entity keys'])) {
+    // If it does, check whether this is new or an update and tailor the
+    // message accordingly.
+    if ($update) {
+      $error_message = t("The machine name '@name' is invalid for a @type as this machine name is reserved by this entity type. You will not be able to save any @type with this field on. Press the back button in your browser and delete this field; create a new field with an alternative machine name.", array(
+        '@name' => $instance['field_name'],
+        '@type' => $entity_info['label'],
+      ));
+    }
+    else {
+      $error_message = t("You cannot add the existing field with the machine name '@name' to a @type as this machine name is reserved by this entity type", array(
+        '@name' => $instance['field_name'],
+        '@type' => $entity_info['label'],
+      ));
+    }
+    throw new FieldException($error_message);
   }
 }
 
--- a/core/modules/field/field.crud.inc
+++ b/core/modules/field/field.crud.inc
@@ -640,10 +640,10 @@ function field_validate_instance($instance, $update = FALSE, $load_field = TRUE)
         '@type' => $entity_info['label'],
       );
       if ($update) {
-        $error_message = t("The machine name '@name' is invalid for a @type as this machine name is reserved by this entity type. You will not be able to save any @type with this field on. You should delete this field and create a new field with an alternative machine name", $error_message_placeholders);
+        $error_message = t("The machine name '@name' is invalid for entity type @type as this machine name is reserved by this entity type. You will not be able to save any @type with this field on. You should delete this field and create a new field with a different machine name", $error_message_placeholders);
       }
       else {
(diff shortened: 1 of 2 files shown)


core/modules/field/field.crud.inc
```diff
@@ -513,6 +513,16 @@ function field_create_instance($instance) {
     throw new FieldException($message);
   }
 
+  // Check to ensure that the field instance does not have a reserved word for
+  // this entity type.
+  $entity_info = entity_get_info($instance['entity_type']);
```
argiepiano: I'm wondering if this check should instead be done in `field_validate_instance()`, to keep the same pattern as in the creation of new fields (a similar check is performed in `field_validate_field()`).

yorkshire-pudding: Thanks for looking @argiepiano.  I have updated to this suggestion and updated the error messages depending on whether it is a new instance for the entity type/bundle or an update to an existing.

For existing, it goes to a full page error, but I don't know if there is any other way

image

This throws quite a few test errors so I'll take a further look.

PHP 7.4
```
Detailed test results
---------------------

---- Taxonomy: Taxonomy vocabularies (TaxonomyVocabularyUnitTest) ----

Status    Group      Filename          Line    Function                            
------------------------------------------------------------------------------------------------------------------------
Fail      Completion taxonomy.test     384     testUninstallReinstall()                                                
    The test did not complete due to a fatal error.
Exception Warning    field.crud.inc    640                                                                             
    Trying to access array offset on null
Error: Process completed with exit code 1.
```

PHP 8.4
```
Detailed test results
---------------------

---- Module: Module uninstallation (ModuleUninstallTestCase) ----

Status    Group      Filename          Line    Function                            
------------------------------------------------------------------------------------------------------------------------
Exception Warning    field.crud.inc    640                                                                             
    Trying to access array offset on null
Exception TypeError  field.crud.inc    640                                                                             
    in_array(): Argument #2 ($haystack) must be of type array, null given

---- Module: Module API (ModuleUnitTest) ----

Status    Group      Filename          Line    Function                            
------------------------------------------------------------------------------------------------------------------------
Fail      Completion module.test       139     testDependencyResolution()                                              
    The test did not complete due to a fatal error.
Exception Warning    field.crud.inc    640                                                                             
    Trying to access array offset on null
Error: Process completed with exit code 1.
```

yorkshire-pudding: OK. By changing the update one to a `backdrop_set_message()` of type `error` it avoids the full screen error which is also breaking tests.

core/modules/field/field.crud.inc
```diff
@@ -629,6 +629,26 @@ function field_validate_instance($instance, $update = FALSE, $load_field = TRUE)
...
+      // message accordingly.
+      $error_message_placeholders = array(
+        '@name' => $instance['field_name'],
+        '@type' => $entity_info['label'],
+      );
+      if ($update) {
+        $error_message = t("The machine name '@name' is invalid for a @type as this machine name is reserved by this entity type. You will not be able to save any @type with this field on. You should delete this field and create a new field with an alternative machine name", $error_message_placeholders);
+      }
+      else {
+        $error_message = t("You cannot add the existing field with the machine name '@name' to a @type as this machine name is reserved by this entity type", $error_message_placeholders);
```
argiepiano: I'm wondering if we could use a similar message construction as the one thrown in a similar situation by `field_validate_field()` line 300. My suggestion:

```suggestion
        $error_message = t("You cannot add the existing field with the machine name '@name' to entity type @type as this machine name is reserved by this entity type", $error_message_placeholders);
```

core/modules/field/field.crud.inc
```diff
@@ -629,6 +629,26 @@ function field_validate_instance($instance, $update = FALSE, $load_field = TRUE)
...
+    $entity_info = entity_get_info($instance['entity_type']);
+    if (in_array($instance['field_name'], $entity_info['entity keys'])) {
+      // If it does, check whether this is new or an update and tailor the
+      // message accordingly.
+      $error_message_placeholders = array(
+        '@name' => $instance['field_name'],
+        '@type' => $entity_info['label'],
+      );
+      if ($update) {
+        $error_message = t("The machine name '@name' is invalid for a @type as this machine name is reserved by this entity type. You will not be able to save any @type with this field on. You should delete this field and create a new field with an alternative machine name", $error_message_placeholders);
```
argiepiano: I suggest following a similar message construction as `field_validate_field()` line 300.  Also do we really know that "You will not be able to save any @type with this field on"?

```suggestion
        $error_message = t("The machine name '@name' is invalid for entity type @type as this machine name is reserved by this entity type. You will not be able to save any @type with this field on. You should delete this field and create a new field with a different machine name", $error_message_placeholders);
```

yorkshire-pudding: > Also do we really know that "You will not be able to save any @type with this field on"?

It was this error that was the problem for me. I haven't tested elsewhere but I see no reason why it wouldn't be the same; it didn't look as though Paragraphs was doing anything unique here.

quicksketch:
Merged into 1.x and 1.31.x. Thank you @yorkshire-pudding and @argiepiano!

PR #5075, opened 2025-05-16
Issue 6912. Create callback for custom block machine name.

### LINKED ISSUE #6912: Fixed fatal error when creating Reusable Custom Blocks with duplicate machine names.
## Description of the bug

When using the Reusable Custom Blocks module (part of Backdrop core), creating two custom blocks with the same machine name causes a fatal error due to improper handling of duplicates in the database.

## Actual behavior

	1. Enable the Reusable Custom Blocks module on a fresh Backdrop install.
	2. Create a new custom block and give it a specific machine name (e.g., my_block).
	3. Create another custom block and manually enter the same machine name (my_block).
	4. See the error:
	
```
TypeError: Cannot access offset of type array on array in block_custom_block_load() (line 225 of /var/www/html/docroot/core/modules/block/block.module).     
```

## Expected behavior

The system should validate the uniqueness of the machine_name and prevent creation of blocks with duplicate machine names.

## Additional information

The block_custom_block_load() function expects machine_name to be unique, but the database does not enforce this. When duplicates exist, the query returns multiple rows and attempts to treat them as a single array, resulting in a TypeError.

argiepiano:
Fixes https://github.com/backdrop/backdrop-issues/issues/6912

### DIFF
--- a/core/modules/block/block.module
+++ b/core/modules/block/block.module
@@ -268,7 +268,7 @@ function block_custom_block_form($edit = array(), $stand_alone = TRUE) {
     '#default_value' => $edit['delta'],
     '#maxlength' => 128,
     '#machine_name' => array(
-      'exists' => 'block_custom_block_load',
+      'exists' => '_block_custom_block_name_exists',
       'source' => array('info'),
     ),
     '#description' => t('A unique machine-readable name for this Block. It must only contain lowercase letters, numbers, and underscores.'),
@@ -347,6 +347,23 @@ function block_custom_block_form($edit = array(), $stand_alone = TRUE) {
   }
 
   return $form;
+}
+
+/**
+ * Render API callback: Checks if a custom block machine name is taken.
+ *
+ * @param $value
+ *   The machine name.
+ *
+ * @return
+ *   Whether or not the custom block machine name is taken.
+ * 
+ * @since 1.31.1
+ */
+function _block_custom_block_name_exists($value) {
+  if (!empty($value)) {
+    return (bool) block_custom_block_load($value);
+  }
 }
 
 /**
--- a/core/modules/block/block.module
+++ b/core/modules/block/block.module
@@ -352,12 +352,12 @@ function block_custom_block_form($edit = array(), $stand_alone = TRUE) {
 /**
  * Render API callback: Checks if a custom block machine name is taken.
  *
- * @param $value
+ * @param string $value
  *   The machine name.
  *
- * @return
+ * @return bool
  *   Whether or not the custom block machine name is taken.
- * 
+ *
  * @since 1.31.1
  */
 function _block_custom_block_name_exists($value) {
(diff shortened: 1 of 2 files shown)


core/modules/block/block.module
```diff
@@ -349,6 +349,23 @@ function block_custom_block_form($edit = array(), $stand_alone = TRUE) {
   return $form;
 }
 
+/**
+ * Render API callback: Checks if a custom block machine name is taken.
+ *
+ * @param string $value
+ *   The machine name.
+ *
+ * @return bool
```
indigoxela: Currently this function can also return "null" - if the value's empty.
Either account for that, or better - add a default return.

argiepiano: Hi @indigoxela. The return is always bool, since there is explicit casting in the return line to `(bool)`.

You can see that this will never return null by running:

`debug(is_null((bool) NULL));`

Which returns false, implying that explicit casting of NULL is not another null, but rather, FALSE.

Also, this:
`debug(((bool) NULL) === FALSE);`

which returns TRUE.

indigoxela: Ah, I see, a misunderstanding...

This code sample might help to clarify:
```php
  if (!empty($value)) {
    return (bool) block_custom_block_load($value);
  }
  // But if it's empty, you end up with returning nothing, so either return something here or state that the function *can* return null.
```
^^ Does that better explain what I mean?

argiepiano: Ha, of course! Sorry, it's early here, still sleepy. I'll fix that docblock.

argiepiano: There. Is that better?

core/modules/block/block.module
```diff
@@ -349,6 +349,24 @@ function block_custom_block_form($edit = array(), $stand_alone = TRUE) {
...
+ * Render API callback: Checks if a custom block machine name is taken.
+ *
+ * @param string $value
+ *   The machine name.
+ *
+ * @return bool|null
+ *   Whether or not the custom block machine name is taken, or null if $value is
+ *   empty.
+ *
+ * @since 1.31.1
```
quicksketch:
```suggestion
 * @since 1.31.1 Function added.
```

core/modules/block/block.module
```diff
@@ -349,6 +349,24 @@ function block_custom_block_form($edit = array(), $stand_alone = TRUE) {
...
+ * Render API callback: Checks if a custom block machine name is taken.
+ *
+ * @param string $value
+ *   The machine name.
+ *
+ * @return bool|null
+ *   Whether or not the custom block machine name is taken, or null if $value is
+ *   empty.
+ *
+ * @since 1.31.1 Function added.
```
quicksketch:
```suggestion
 * @since 1.32.1 Function added.
```

quicksketch:
Thank you @argiepiano! I merged this into 1.x and 1.32.x.

PR #5078, opened 2025-05-29
Adjust admin_bar styling to open secondary+ menus straight down for usability on very small screens.

### LINKED ISSUE #6916: [UX] Improve Admin bar second and third level menu items on smaller devices.
## Description

On small screen-width devices (500px and below), the admin bar will throw second or third level menu items off of the screen, so a user can either not read the label, or cannot select the menu item at all.

## Steps To Reproduce

To reproduce the behavior:
1. Set up a Backdrop site with the admin_bar module installed, and give a user the permission to use the admin bar.
2. Login as your user and load the site on a small device (500px or narrower in a responsive browser mode, or iPhone 11 would do it for example)
3. Tap the collapsed "Admin Bar" to open the menu, and open up a menu item with 2 or 3 levels (Such as Configuration > Content Authoring > Text Editors and Formats)

## Actual behavior

Observe the menu bleeding off the edge of the page. Here is an example with several items on a 500px screen size:

Image

Here is an example on iPhone 11 Pro screen width (375px) where even secondary menu items are unusable:

Image

## Desired behavior

I think on very small screens, the admin bar should probably just accordion straight down instead of doing this side-to-side menu opening thing. It might look like this on a small screen size:

Image

## Additional information

- Tested on Backdrop 1.31.0

elisseck:
Fixes https://github.com/backdrop/backdrop-issues/issues/6916


izmeez:
There is some strange behaviour with nested menus when using the menu on a desktop with a narrow browser window. On a narrowed browser on the menu go to "Content" that has "Add content" with child menus and "Manage content". Moving down the "Add content" child menus open but it is not possible to go to the "Manage content" link. To get to that menu item it is necessary to go further down to the "Manage comments" item and then come back up to the "Manage content" item, or move the mouse in an arc that goes around the "Add content" item to reach the "Manage content" item.

core/modules/admin_bar/css/admin_bar.css
```diff
@@ -393,3 +393,10 @@ tr.admin-bar-tweak-permissions-processed td.module {
 #admin-bar .admin-bar-alert .error-label {
   display: none;
 }
+
```
quicksketch, line 400:
```suggestion
/* On small screens, expand children straight down instead of side-to-side. */
```

core/modules/admin_bar/css/admin_bar.css
```diff
@@ -393,3 +393,10 @@ tr.admin-bar-tweak-permissions-processed td.module {
 #admin-bar .admin-bar-alert .error-label {
   display: none;
 }
+/* On small screens, expand children straight down instead of side-to-side. */
```
quicksketch, line 401:
```suggestion

/* On small screens, expand children straight down instead of side-to-side. */
```

elisseck:
@izmeez see issue discussion but I believe the concern for small desktop screen usability with hover is resolved.

I believe test fails are unrelated, it looks like just some spelling stuff from another core module?

quicksketch:
Tested this and it works great, thank you @elisseck!

> I believe test fails are unrelated, it looks like just some spelling stuff from another core module?

Yep, nothing to worry about.

---

Merged into 1.x and 1.32.x. Thank you!

PR #5079, opened 2025-06-03
Issue #6919: Reset field type static when Entity Reference is installed.

### LINKED ISSUE #6919: Reset field types when new modules are installed to ensure potential new field types are immediately available
## Description of the bug

We've had the ability to create fields simply in modules by adding configuration files inside the module's `config/` directory for some time. When the module is installed, those configuration files get synced, fields created, etc.

There is an issue with doing so in a module that includes an Entity Reference field, when the module is enabled, and the module's dependencies (including Entity Reference) get enabled at the same time. In this case, as identified by @argiepiano [in a separate thread](https://github.com/backdrop/backdrop-issues/issues/6421#issuecomment-2660627190), Backdrop doesn't seem to know about Entity Reference fields yet and the site crashes in a fatal error in the middle of the installation of the module. A simple solution is to call `_field_info_collate_types_reset()` when Entity Reference is installed:

```
/**
 * Implements hook_install().
 */
function entityreference_install() {
  // Reset field types to recognize Entity Reference fields immediately.
  _field_info_collate_types_reset();
}
```

## Steps To Reproduce

To reproduce the behavior:
1. Install the simple config recipe module attached here. _(It creates an Entity Reference field on the Card content type)._

[entityreference_fieldconfig.zip](https://github.com/user-attachments/files/20577339/entityreference_fieldconfig.zip)

## Actual behavior

Fatal error prior to module activation being complete: 
```
Error 
The field "field_related_page" could not be created because it requires the "" module. 
```

## Expected behavior

Module installs without a fatal error and the Entity Reference field is set up correctly.

laryn:
Fixes https://github.com/backdrop/backdrop-issues/issues/6919.

### DIFF
--- a/core/includes/module.inc
+++ b/core/includes/module.inc
@@ -515,8 +515,8 @@ function module_enable($module_list, $enable_dependencies = TRUE) {
         watchdog('system', '%module module installed.', array('%module' => $module), WATCHDOG_INFO);
 
         if (function_exists('_field_info_collate_types_reset')) {
-        // Clear the field info cache so that the new module's field types are
-        // available.
+          // Clear the field info cache so that the new module's field types are
+          // available.
           _field_info_collate_types_reset();
         }
       }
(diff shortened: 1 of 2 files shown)


laryn:
Closing in favour of @argiepiano's more optimized approach in https://github.com/backdrop/backdrop/pull/5080

PR #5081, opened 2025-06-11
Issue #6920: Clean up missing filters before saving

### LINKED ISSUE #6920: Filter module never cleans up filter section in config leading to PHP warnings
## Description of the bug

I noticed that problem a while back.

The Filter module doesn't clean up JSON config after modules providing filters got uninstalled.

## Steps To Reproduce

There are multiple ways to trigger that problem, one of them is upgrading from D7, but the problem's easier to reproduce with contrib modules.

1. Install a module that provides a filter with settings (like video_filter)
2. Edit one of your filter formats (enable or don't enable the new filter - doesn't actually matter)
3. Save
4. Uninstall that module again
5. Go back to edit your filter format
6. Save

## Actual behavior

```
Warning: Undefined array key "video_filter" in filter_format_build_format() (line 319 of .../core/modules/filter/filter.module).
Warning: Trying to access array offset on null in filter_format_build_format() (line 328 of .../core/modules/filter/filter.module).
Warning: Trying to access array offset on null in filter_format_build_format() (line 330 of .../core/modules/filter/filter.module).
```

Take a look at the filter format's JSON config file - the module's filter settings are still there.

- Save the filter format again (note: the module's not  there anymore).
- Open the filter format form again... save again...

Same warnings, the problem persists. There's basically *no way* to get rid of formerly used filter settings, unless editing the JSON file directly. Ugly...

The Video Filter module's just an example, the problem applies to all filters with settings.

That means that upgrading from D7 requires to edit JSON to get rid of things like transliteration setting or php_code filter setting...

## Expected behavior

If the module providing the filter is gone, the Filter module should do some cleanup before saving. Possibly in filter_format_build_format().

## Additional information

- PHP version: 8.4 (but warnings can be seen with any 8.x)

See where formats get saved: https://github.com/backdrop/backdrop/blob/1.x/core/modules/filter/filter.module#L370

indigoxela:
Fixes https://github.com/backdrop/backdrop-issues/issues/6920

### DIFF
--- a/core/modules/filter/filter.module
+++ b/core/modules/filter/filter.module
@@ -316,6 +316,13 @@ function filter_format_build_format($format) {
   }
   $all_filter_info = filter_get_filters();
   foreach ($format->filters as $name => $filter) {
+    if (!array_key_exists($name, $all_filter_info)) {
+      // Remove filters that don't exist anymore, for example from uninstalled
+      // modules, so they don't get saved again.
+      unset($format->filters[$name]);
+      continue;
+    }
+
     $filter_info = $all_filter_info[$name];
 
     // Ensure the $filter variable is an object. In Drupal 7, it was an array on


quicksketch:

PR #5083, opened 2025-06-13
Issue #6924: Fix views ajax uses wrong selector when view re-used on page

### LINKED ISSUE #6924: Using multiple file browsers on a single node form breaks functionality
## Description of the bug

If you have a node form in which you have multiple methods of using a file browser (ex: a filefield with 'select existing file' plus a wysiwyg 'image' button where you 'select from library'), after using the first instance successfully, the second instance has broken functionality. 

I think there are multiple combinations of functionality which can cause this, but in my case I found it while using an image file field while paired with the CKEditor 5 image upload button within a WYSIWYG.  (using other wysiwyg options is reported to cause this behavior as well)

## Steps To Reproduce

To reproduce the behavior:
1. Create a content type with an image file field which uses the 'select existing file 'functionality so you can select an existing file from the media library using the library modal selector
2. Setup CKeditor 5 to use an image button and allow re-use of existing files
3. Create a new node using this type and use the 'select existing file' on the image filefield.  Choose a file and this will load the image into the field properly
4. Now go use the WYSIWYG image button in your body (textarea).  Click the image button, choose the tab to 'select from library'. Now use the search function to find an existing image by filename.

## Actual behavior
When you click apply after searching for the image by filename, the apply button is broken and redirects you to a strange url which includes the filename you searched for.  Something like: http://mysite.com/?filename=file_that_was_searched.jpg

## Expected behavior
Obviously, the apply button should, using javascript, filter the library image list by your search and allow you to choose the image.

## Additional information

My guess is that using the first library modal to choose an image is then breaking any additional uses of the modal on the page, possibly due to scope issues.

Note that you can do different combinations and get similar behavior.  I've confirmed it works in reverse (do the WYSIWYG image first (succesfully) and then you'll see that the file field image library browser is broken.

indigoxela:
Fixes https://github.com/backdrop/backdrop-issues/issues/6924

### DIFF
--- a/core/modules/views/js/ajax_view.js
+++ b/core/modules/views/js/ajax_view.js
@@ -71,8 +71,17 @@ Backdrop.views.ajaxView = function(settings) {
   this.settings = settings;
 
   // Add the ajax to exposed forms.
-  this.$exposed_form = $('form#views-exposed-form-'+ settings.view_name.replace(/_/g, '-') + '-' + settings.view_display_id.replace(/_/g, '-'));
-  this.$exposed_form.once(jQuery.proxy(this.attachExposedFormAjax, this));
+  let formSelector;
+  if (settings.exposed_form_in_block) {
+    formSelector = 'form#views-exposed-form-'+ settings.view_name.replace(/_/g, '-') + '-' + settings.view_display_id.replace(/_/g, '-');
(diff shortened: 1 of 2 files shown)


quicksketch:

PR #5087, opened 2025-07-05
Issue #6935: Return 400 with less verbose text if trusted host check fails

### LINKED ISSUE #6935: Trusted host settings handling - multiple issues
## Description of the bug

In an attempt to prevent HTTP Host header spoofing, trusted_host_patterns handling has been added to Backdrop in https://github.com/backdrop/backdrop-issues/issues/2568

The implementation has multiple issues, though.

## Steps To Reproduce

The easiest way to reproduce is to remove the actual hostname from trusted_host_patterns setting (in settings.(local.)php and call any page with it. That's the easiest way, but not the only one.

## Actual behavior

```
Exception: The HTTP Host "example.com" is not white-listed for this site. Check the trusted_host_patterns setting in settings.php. in _backdrop_bootstrap_configuration() (line 3128 of /var/www/testing/html/core/includes/bootstrap.inc). 
```

Image

Issues with that:

1. This throws an Exception, that will show up as `500 Internal Server Error` in webserver logs - as if it was a problem on the site, while it's a problem with the request
2. It uses the term "white-listed", which we removed anywhere else already
3. It points to the code line in bootstrap.inc, when error display is turned on - as if this were a problem to get fixed in code
4. It recommends action in settings.php, but the audience here is **not primarily** the site admin, it's public

## Expected behavior

First of all, the error has to be `400 Bad Request`.

The wording shouldn't use deprecated terms and the hint shouldn't point to action with settings.php.

## Additional information

While this may seem to hardly ever occur, this occurs regularly, if Backdrop is the _default host on an IP_.

And it's super easy to flood the webserver logs with 500-s from outside just by using something like:
```
curl -I -H 'Host:foobar.com' http://ACTUAL_HOSTNAME_OR_IP
```
And I don't think, the bots crawling a site (primary audience for that message) will bother to fix it in settings.php 😜 They sure would like to have access to that file, but... nope.

indigoxela:
Fixes https://github.com/backdrop/backdrop-issues/issues/6935

### DIFF
--- a/core/includes/bootstrap.inc
+++ b/core/includes/bootstrap.inc
@@ -3123,9 +3123,13 @@ function _backdrop_bootstrap_configuration() {
     install_goto('core/install.php');
   }
 
-  // Untrusted host names, throw an exception for the end-user.
+  // Untrusted host names, return 400 Bad Request to the end-user.
   if (!defined('MAINTENANCE_MODE') && !backdrop_check_trusted_hosts($_SERVER['HTTP_HOST'])) {
-    throw new Exception(format_string('The HTTP Host "@hostname" is not white-listed for this site. Check the trusted_host_patterns setting in settings.php.', array('@hostname' => $_SERVER['HTTP_HOST'])));
+    http_response_code(400);
+    print format_string('HTTP Host "@hostname" is not included in "trusted_host_patterns" setting of this site.', array(
+      '@hostname' => $_SERVER['HTTP_HOST'],
+    ));
+    exit;
   }
 
   // Bootstrap the database if it is needed but not yet available.
--- a/core/includes/bootstrap.inc
+++ b/core/includes/bootstrap.inc
@@ -3126,7 +3126,7 @@ function _backdrop_bootstrap_configuration() {
   // Untrusted host names, return 400 Bad Request to the end-user.
   if (!defined('MAINTENANCE_MODE') && !backdrop_check_trusted_hosts($_SERVER['HTTP_HOST'])) {
     http_response_code(400);
-    print format_string('HTTP Host "@hostname" is not included in "trusted_host_patterns" setting of this site.', array(
+    print format_string('HTTP Host "@hostname" is not included in the "trusted_host_patterns" setting of this site.', array(
       '@hostname' => $_SERVER['HTTP_HOST'],
     ));
     exit;


core/includes/bootstrap.inc
```diff
@@ -3123,9 +3123,13 @@ function _backdrop_bootstrap_configuration() {
     install_goto('core/install.php');
   }
 
-  // Untrusted host names, throw an exception for the end-user.
+  // Untrusted host names, return 400 Bad Request to the end-user.
   if (!defined('MAINTENANCE_MODE') && !backdrop_check_trusted_hosts($_SERVER['HTTP_HOST'])) {
-    throw new Exception(format_string('The HTTP Host "@hostname" is not white-listed for this site. Check the trusted_host_patterns setting in settings.php.', array('@hostname' => $_SERVER['HTTP_HOST'])));
+    http_response_code(400);
+    print format_string('HTTP Host "@hostname" is not included in "trusted_host_patterns" setting of this site.', array(
```
yorkshire-pudding: Minor wording change suggestion

```suggestion
    print format_string('HTTP Host "@hostname" is not included in the "trusted_host_patterns" setting of this site.', array(
```

core/includes/bootstrap.inc
```diff
@@ -3123,9 +3123,13 @@ function _backdrop_bootstrap_configuration() {
     install_goto('core/install.php');
   }
 
-  // Untrusted host names, throw an exception for the end-user.
+  // Untrusted host names, return 400 Bad Request to the end-user.
   if (!defined('MAINTENANCE_MODE') && !backdrop_check_trusted_hosts($_SERVER['HTTP_HOST'])) {
-    throw new Exception(format_string('The HTTP Host "@hostname" is not white-listed for this site. Check the trusted_host_patterns setting in settings.php.', array('@hostname' => $_SERVER['HTTP_HOST'])));
+    http_response_code(400);
```
yorkshire-pudding, line 3128: I wonder whether this is the best code to use. I'll start the discussion in the issue.

yorkshire-pudding (commented):
Code looks good. I've suggested a minor wording change.
Also, I wonder if 400 is right and whether we could be more specific; I will discuss in the issue.

yorkshire-pudding (approved):
Looks good

core/includes/bootstrap.inc
quicksketch: Just another comment that this is the first place we're using `http_response_code()` in Backdrop. I assume that's because it was introduced in PHP 5.4, which we were not requiring when Backdrop first forked. This looks good to me.

quicksketch:
Merged into 1.x and 1.31.x. Thank you @indigoxela and @yorkshire-pudding!

PR #5088, opened 2025-07-10
Issue #6307: Fix Date module datestamp field type affected by Y2K38

### LINKED ISSUE #6307: Date module datestamp field type affected by Y2K38
## Description of the bug

The datestamp field type saves integers to the database, but [uses an INT type](https://github.com/backdrop/backdrop/blob/1.x/core/modules/date/date.install#L15), which limits the dates to (roughly) be **between 1901 and 2038**.

To allow higher values, we have to change the timestamp fields to BIGINT.

## Steps To Reproduce

1. Create a "Date (Unix timestamp)" field instance (widget doesn't matter)
2. Make sure to set Starting year and Ending year to some far past/future values, outside the supported range
3. Create a node of the type containing that field
4. Set the date's year to either be before 1901 or after 2038 and try to save

## Actual behavior

Fatal error:

```
EntityStorageException: SQLSTATE[22003]: Numeric value out of range: 1264 Out of range value for column 'field_datum_int_value' at row 1 in NodeStorageController->save() (line 512 of ...core/modules/node/node.entity.inc).
SQLSTATE[22003]: Numeric value out of range: 1264 Out of range value for column 'field_datum_int_value' at row 1
```

## Expected behavior

Not sure how common the use of dates before 1901 is, but 2038 isn't that far away. :wink: 

## Additional information

- Backdrop CMS version: current dev
- PHP version: any
- Database sever (MySQL or MariaDB?) and its version: any
- Operating System and its version: anything 64-bit (not sure, if we still need to consider 32-bit)

Relevant d.o issues:
- [[meta] Y2K38: Unix Millennium bug](https://www.drupal.org/project/drupal/issues/65474)
- 10.1.x change record: [Update timestamp database column from integer to bigint for the Y2K38 bug](https://www.drupal.org/node/3227494)

Backdrop meta issue: https://github.com/backdrop/backdrop-issues/issues/6931

indigoxela:
Fixes https://github.com/backdrop/backdrop-issues/issues/6307

Based on Drupal commits by r-mo, DamienMcKenna, joep.hendrix, heddn, julien_g, Steven Jones, joseph.olstad

### DIFF
--- a/core/modules/date/date.install
+++ b/core/modules/date/date.install
@@ -13,6 +13,7 @@ function date_field_schema($field) {
     case 'datestamp':
       $db_columns['value'] = array(
         'type' => 'int',
+        'size' => 'big',
         'not null' => FALSE,
         'sortable' => TRUE,
         'views' => TRUE,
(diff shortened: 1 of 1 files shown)


quicksketch:

PR #5093, opened 2025-08-01
Issue #6943: Update hook_user_presave example to prevent deprecation notice.

### LINKED ISSUE #6943: PHP 8+ Example for hook_user_presave causes Deprecation notice if data field is NULL or FALSE
## Description of the bug

Following the example of `hook_user_presave()` will cause deprecation notices with standard error reporting and "502 Bad Gateway" if Devel is installed and Krumo backtrace is enabled.

Here is the doc and example code from `user.api.php`:

```php
/**
 * Act on a user account being inserted or updated.
 *
 * This hook is invoked before the user account is saved to the database.
 *
 * Modules that want to store properties in the serialized {users}.data column,
 * which is automatically loaded whenever a user account object is loaded, may
 * add their properties to $account->data in order to have their data serialized
 * on save.
 *
 * @param $account
 *   The user account object.
 *
 * @see hook_user_insert()
 * @see hook_user_update()
 */
function hook_user_presave($account) {
  // Make sure that our form value 'my_module_foo' is stored as
  // 'my_module_bar' in the 'data' (serialized) column.
  if (isset($account->my_module_foo)) {
    $account->data['my_module_bar'] = $account->my_module_foo;
  }
}
```

It is feasible to not have any data in the data field for a user:

Image

If a module invokes `hook_user_presave()` in the way suggested on one of these users then if deprecation notices are enabled they will be shown or if Devel Krumo backtrace is enabled for deprecations then a Bad Gateway error can be caused.

## Steps To Reproduce

To reproduce the behavior:

1. Enable all error messages
2. Enable deprecations to be logged
3. Install and enable Devel and Devel Generate and enable the Krumo Backtrace above rendered page
4. Install and enable Matomo Analytics (1.x-2.12.3) (I'll be raising an issue there too but keep in mind this is following the current example)
5. Use Devel Generate to create some users
6. Edit one of those users and edit something like the email address
7. Save

## Actual behavior

If Devel Krumo backtrace is enabled then Bad gateway error (on on the `b.org` demo a tubgoat)
If not then a deprecation notice:

> Deprecated function: Automatic conversion of false to array is deprecated in matomo_user_presave() (line 456 of /var/lib/tugboat/backdrop/modules/matomo/matomo.module).

## Expected behavior
User saves with no deprecation notice.

## Additional information

- Backdrop CMS version: 1.31.0
- PHP version: 8.2

yorkshire-pudding:
Fixes https://github.com/backdrop/backdrop-issues/issues/6943

### DIFF
--- a/core/modules/user/user.api.php
+++ b/core/modules/user/user.api.php
@@ -213,6 +213,11 @@ function hook_user_presave($account) {
   // Make sure that our form value 'my_module_foo' is stored as
   // 'my_module_bar' in the 'data' (serialized) column.
   if (isset($account->my_module_foo)) {
+    // If the data (serialized) column is NULL or FALSE initiate the array
+    // first.
+    if (!$account->data) {
+      $account->data = array();
+    }
     $account->data['my_module_bar'] = $account->my_module_foo;
   }
 }


laryn:
Thank you! Merged into `1.x` and `1.31.x`.

PR #5105, opened 2025-09-04
Issue #6957: Fix missing private in Cache-Control for personalized content

indigoxela:
Addresses https://github.com/backdrop/backdrop-issues/issues/6957
(but can't fix the reported problem)

### DIFF
--- a/core/modules/user/user.pages.inc
+++ b/core/modules/user/user.pages.inc
@@ -319,7 +319,12 @@ function user_pass_reset_form_submit($form, &$form_state) {
  */
 function user_logout() {
   user_logout_current_user();
-  backdrop_goto();
+  // We can not rely on Etag alone to prevent browsers from re-using the
+  // authenticated home page after logout, as that header may have been stripped
+  // by a shared cache (like a proxy).
+  // @see backdrop_page_header()
+  $query = array('logout' => 'success');
+  backdrop_goto('', array('query' => $query));
 }
 
 /**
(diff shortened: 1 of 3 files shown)


core/modules/user/user.pages.inc
```diff
@@ -319,7 +319,12 @@ function user_pass_reset_form_submit($form, &$form_state) {
  */
 function user_logout() {
   user_logout_current_user();
-  backdrop_goto();
+  // We can not rely on Etag alone to prevent browsers from re-using the
+  // authenticated home page after logout, as that header may have been stripped
+  // by a shared cache (like a proxy).
+  // @see backdrop_page_header()
+  $query = array('logout' => 'success');
+  backdrop_goto('', array('query' => $query));
```
indigoxela: For now I just add the param, but we could make that conditional (by settings.php, by current_path()... whatever).

Only fixes it for the home page, though...

Another, even more radical solution would be `backdrop_add_http_header('Clear-Site-Data', '"cache"');`. But that might be a bit over the top and only works with HTTPS.

indigoxela: Thinking this over... I don't think, we should add that query. The problem is, that this fixes one snowflake problem, but won't work for the entire page cache, in case the ETag header got wiped out by intermediate software.

indigoxela:
Closing, as this approach won't be able to fix the problem reported in the issue.

PR #5110, opened 2025-09-19
Issue #6968: Add defensive coding to tabledrag.js.

### LINKED ISSUE #6968: Add defensive coding in tabledrag.js
I am doing some custom JS work to rearrange a particular page and am running into a case where it breaks the Tabledrag functionality due to a failure to have some defensive coding (it assumes a result for `cell[0]` and breaks if it doesn't exist).

`Uncaught TypeError: can't access property "colSpan", cell[0] is undefined`

[This issue seems also to have been reported in modern Drupal as well](https://www.drupal.org/project/drupal/issues/3441551).

I don't have STR because my use is atypical -- but I propose some minor defensive coding, taken from the patch offered in that Drupal queue, to prevent the JS from crashing in edge cases like mine.

laryn:
Fixes https://github.com/backdrop/backdrop-issues/issues/6968

### DIFF
--- a/core/misc/tabledrag.js
+++ b/core/misc/tabledrag.js
@@ -206,7 +206,7 @@ Backdrop.tableDrag.prototype.addColspanClass = function(columnIndex) {
     });
     if (index > 0) {
       cell = cells.filter(':nth-child(' + index + ')');
-      if (cell[0].colSpan && cell[0].colSpan > 1) {
+      if (cell[0] != null && cell[0] != undefined && cell[0].colSpan && cell[0].colSpan > 1) {
         // If this cell has a colspan, mark it so we can reduce the colspan.
         cell.addClass('tabledrag-has-colspan');
       }


core/misc/tabledrag.js
```diff
@@ -206,7 +206,7 @@ Backdrop.tableDrag.prototype.addColspanClass = function(columnIndex) {
     });
     if (index > 0) {
       cell = cells.filter(':nth-child(' + index + ')');
-      if (cell[0].colSpan && cell[0].colSpan > 1) {
+      if (cell[0] != null && cell[0] != undefined && cell[0].colSpan && cell[0].colSpan > 1) {
```
argiepiano, line 209: Since you are using loose comparison, there is no need to check for both. `null` and `undefined` are "same" when using loose comparison (including `!=`). You can see this by running `console.log(null == underfined);` which returns true.

This is an aside from whether this is needed or not. I'm not sure I understand why you are not getting a value on `cell[0]`, since that cell was expressly selected in the `each()` above, so it should exist. Is your table malformed?

PR #5112, opened 2025-09-24
Issue #6982: Ensure _config_static is set for system.core.

### LINKED ISSUE #6982: Sites upgraded from Drupal 7 to Backdrop missing _config_static true in json files.
## Description of the bug

This is a follow-up issue discovered in efforts to resolve https://github.com/backdrop/backdrop-issues/issues/6961

Sites that were upgraded from Drupal 7 to Backdrop appear to be missing `"_config_static": true;` in the system.core.json file and possibly in other core config files where it is expected.

We need to fix this for any new site upgrades from Drupal 7 to Backdrop and we need an update hook to ensure this is fixed in all existing sites.

laryn:
Fixes https://github.com/backdrop/backdrop-issues/issues/6982

### DIFF
--- a/core/modules/system/system.install
+++ b/core/modules/system/system.install
@@ -3843,7 +3843,7 @@ function system_update_1103() {
  */
 function system_update_1104() {
   $config = config('system.core');
-  if ($config->get('_config_static') == NULL) {
+  if ($config->get('_config_static') === NULL) {
     $config->set('_config_static', TRUE);
     $config->save();
   }
--- a/core/modules/system/system.install
+++ b/core/modules/system/system.install
@@ -3843,8 +3843,10 @@ function system_update_1103() {
  */
 function system_update_1104() {
   $config = config('system.core');
-  $config->set('_config_static', TRUE);
-  $config->save();
+  if ($config->get('_config_static') == NULL) {
+    $config->set('_config_static', TRUE);
+    $config->save();
+  }
 }
 
 /**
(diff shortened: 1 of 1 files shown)


core/modules/system/system.install
```diff
@@ -3838,6 +3838,15 @@ function system_update_1103() {
   $config->save();
 }
 
+/**
+ * Ensure system.core uses _config_static for better performance.
+ */
+function system_update_1104() {
+  $config = config('system.core');
+  $config->set('_config_static', TRUE);
```
argiepiano: What if someone has disabled static for any reason? Should we check if it's set, and then set it if it's not, but respect the value if it is set?

laryn: I can't think of a reason why someone would do that, but sure, just in case?

core/modules/system/system.install
```diff
@@ -3843,8 +3843,10 @@ function system_update_1103() {
  */
 function system_update_1104() {
   $config = config('system.core');
-  $config->set('_config_static', TRUE);
-  $config->save();
+  if ($config->get('_config_static') == NULL) {
```
argiepiano: This will not work. You need to use identical comparison. In php `NULL == FALSE` will result in truthy, so you will set this to TRUE when it was FALSE. what's needed here is `===`

laryn: Whoopsy!

quicksketch:
Merged into 1.x and 1.32.x. Thank you!

PR #5113, opened 2025-09-25
Issue #6972: Revert changes in locale.inc causing regression

### LINKED ISSUE #6972: Fixed Regression: latest core versions break Global Redirect causing redirect loops.
## Description of the bug

This has been reported in our forum initially: https://forum.backdropcms.org/forum/am-i-only-one-having-issues-1320-multilingual-site

See detailed investigation report there.

## Steps To Reproduce

1. Set up a multilingual site on a previous Backdrop version
2. Install [Global Redirect module](https://backdropcms.org/project/globalredirect)
3. Add translated content
4. Update to either 1.31.1 or 1.32.0

Other way to reproduce:

1. Set up a multilingual site (latest dev)
2. Add translated content
3. Install Global Redirect

## Actual behavior

The translations cause infinite redirection loops on all content besides the home page.

## Expected behavior

Guess. 😉 

## Additional information

- Backdrop CMS version: dev, 1.32.0, 1.31.1
- Browser(s) and their versions: any

*Edit:* the problem has been introduced with https://github.com/backdrop/backdrop/commit/53e7b2d9f7c672d6bc32dbb96f99828397e0da7c - that was Issue #4984 

Reverting that change fixes the redirect loop. But why is that change causing loops, anyway?

indigoxela:
Fixes https://github.com/backdrop/backdrop-issues/issues/6972

### DIFF
--- a/core/includes/locale.inc
+++ b/core/includes/locale.inc
@@ -486,12 +486,7 @@ function locale_language_url_rewrite_url(&$path, &$options) {
       case LANGUAGE_NEGOTIATION_URL_PREFIX:
         $prefixes = locale_language_negotiation_url_prefixes();
         if (!empty($prefixes[$options['language']->langcode])) {
-          $path = ltrim((string) $path, '/');
-          $prefix = $prefixes[$options['language']->langcode] . '/';
-          $options['prefix'] = $prefix;
-          if (substr($path, 0, strlen($prefix)) == $prefix) {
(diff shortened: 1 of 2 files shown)


indigoxela:
The problem with phpcs is very likely caused by their latest version (4.0) - needs to get fixed in our ruleset, I guess.
See also: https://github.com/backdrop-ops/phpcs/issues/24

Another option is to pin the version used with our GHA.

PR #5115, opened 2025-09-27
Issue 6974. Blank out system_update_1079

### LINKED ISSUE #6974: Regression: new book cache causes update error when upgrading from D7
## Description of the bug

The new book cache added in https://github.com/backdrop/backdrop-issues/issues/6787 causes an error when upgrading a fresh, vanilla D7 site that has a the book module. Note: this is related to https://github.com/backdrop/backdrop-issues/issues/6965, except that this regression happens EVEN if your D7 site is not a D6 > D7 upgrade. 

The error I'm reporting here  is "catch-ed" by the update batch process. In #6965, this update error did happen and was caught, but then there is an **additional** fatal error that breaks the site when caches are cleared after the update process, since the table `cache_book` is never created in that case (`book_update_X` never run in 6965 situation), but it IS created in this one, later in the update process when `book_update_` are run.

This regression is present even after fixing the related regression in #6965.

## Steps To Reproduce

To reproduce the behavior:
1. Start with a vanilla D7 site with book enabled.
2. Follow the standard steps to upgrade a D7 site to Backdrop. Be sure to use version 1.32.0 of Backdrop
3. See the error in the update report:

Image

```
Update #1079
Failed: PDOException: SQLSTATE[42S02]: Base table or view not found: 1146 Table 'bdupgrade.cache_book' doesn't exist: TRUNCATE {cache_book} ; Array ( ) in backdrop_flush_all_caches() (line 8910 of /Users/XXX/Sites/localhost/XXX/core/includes/common.inc).
 ```
After running update again, things work fine, BUT the error prevents SOME system updates from running. 

## Actual behavior

Update errors. Running update again is needed to complete all updates.

## Expected behavior

No update errors

## Additional information

Backdrop: 1.32.0

The error is caused by `system_update_1079()`, which attempts to clear caches before the book update that creates the cache table has run.

BTW, this 1079 update has proven a headache in other situations, to the point that it has been suggested we should get rid of it. I'm thinking specifically of a Rules update that failed for the very same reason, that necessitated creating a way to detect whether the update was running during a D7 upgrade, or during a "normal" Backdrop update. See @quicksketch [comment here](https://github.com/backdrop/backdrop-issues/issues/6633#issuecomment-2208471723).

### PR DESCRIPTION
**Pull request by argiepiano**
Fixes https://github.com/backdrop/backdrop-issues/issues/6974

### DIFF
--- a/core/modules/system/system.install
+++ b/core/modules/system/system.install
@@ -3409,10 +3409,13 @@ function system_update_1078() {
 
 /**
  * Clear the cache to reset allowlist settings.
+ * 
+ * This update has been removed because it causes a regression when later update
+ * hooks create cache tables.
+ *  
+ * @see https://github.com/backdrop/backdrop-issues/issues/6974
  */
-function system_update_1079() {
-  backdrop_flush_all_caches();
-}
+function system_update_1079() {}
 
 /**
  * Set homepage to 'user' if blank to match previous behavior.
--- a/core/modules/system/system.install
+++ b/core/modules/system/system.install
@@ -3415,7 +3415,9 @@ function system_update_1078() {
  *  
  * @see https://github.com/backdrop/backdrop-issues/issues/6974
  */
-function system_update_1079() {}
+function system_update_1079() {
+  // This update has been removed.
+}
 
 /**
  * Set homepage to 'user' if blank to match previous behavior.

### CODE REVIEW

File: core/modules/system/system.install [hunk 1]
```diff
@@ -3409,10 +3409,13 @@ function system_update_1078() {
...
+ * 
+ * This update has been removed because it causes a regression when later update
+ * hooks create cache tables.
+ *  
+ * @see https://github.com/backdrop/backdrop-issues/issues/6974
  */
-function system_update_1079() {
-  backdrop_flush_all_caches();
-}
+function system_update_1079() {}
```
quicksketch: `system_update_1023()` was also removed and includes a code comment in the function body, let's do that again here:

```suggestion
function system_update_1079() {
  // This update has been removed.
}
```

quicksketch:
Merged into 1.x and 1.32.x. Thank you @argiepiano!

PR #5117, opened 2025-09-30
Issue #6973: Replace phpcs action

### LINKED ISSUE #6973: Account for latest release of phpcs in our GHA workflow
## Description of the task

A new major version of phpcs came out recently. It contains breaking changes we have to consider.

https://github.com/PHPCSStandards/PHP_CodeSniffer/releases

## Steps To Reproduce

Currently our workflow file doesn't pin phpcs version, so we get the latest (which is 4.x). But our ruleset up until 1.0.0 isn't compatible yet.

## Actual behavior

Workflow fails with error `ERROR: Referenced sniff "Squiz......." does not exist.`

**NOTE:**
Fixing our ruleset wasn't difficult, but working on this issue uncovered, that the GHA we use silently broke. We can't seem to get annotations or step failure anymore, the step always passes. Even with older phpcs 3.x. Unclear why.

## Additional information

A new tag for our ruleset is available, <s>but there's also the option to just pin the phpcs version we use.</s>

Replacing the (unmaintained) action is a little more effort. None of all the available existing actions (marketplace) provides what we need.

See also:

- https://github.com/backdrop-ops/phpcs/issues/24
- https://github.com/backdrop-ops/phpcs/releases/tag/1.1.0
- https://github.com/PHPCSStandards/PHP_CodeSniffer/wiki/Version-4.0-User-Upgrade-Guide (official upgrade guide)
- https://github.com/thenabeel/action-phpcs

We know for quite a while now, that this GHA might completely break at some point - it's unmaintained. And always had some quirks. See related https://github.com/backdrop/backdrop-issues/issues/5893

indigoxela:
Fixes https://github.com/backdrop/backdrop-issues/issues/6973

...yet another try. Changing something in common.inc (current experiment) shows sort of the worst case scenario, because of all that legacy code there.


indigoxela:
Oh, look there, a random buddy (Bare upgrade test)...

indigoxela:
Closing in favor of https://github.com/backdrop/backdrop/pull/5120, which works better for our use case.

PR #5119, opened 2025-10-01
Issue #6978: Fixes for transparency with Webp and Gif images.

### LINKED ISSUE #6978: Fixes for transparency with Webp and Gif images.
## Description of the bug

Another one that I first thought was a contrib fix, but then came to suspect should be fixed in core. When uploading an animated gif to an image field, we get an error if it has transparency set when it is trying to create the temporary thumbnail on the admin page, for example.

This PHP error shows up in the logs.

> ValueError: imagecolorsforindex(): Argument #2 ($color) is out of range in imagecolorsforindex() (line 322 of /var/www/html/backdrop/core/modules/system/image.gd.inc).

Also Webp images just drop any transparency when processed.

## Steps To Reproduce

1. Use this image to test Gif, or a webp image with transparency to test Webp: 
Image
2. Upload to the image field of a content type (e.g. Post) while creating new content on a vanilla Backdrop site.

### Gif
- PHP Error in logs
- Image shows up as a blank/transparent thumbnail

### Webp
- Image shows up with transparency changed to white

## Expected behavior

- No PHP Error
- Image shows up as an actual thumbnail, with transparency intact

- Backdrop CMS version:
- Web server and its version:
- PHP version: 

laryn:
Fixes https://github.com/backdrop/backdrop-issues/issues/6978


core/modules/system/image.gd.inc
```diff
@@ -318,13 +318,20 @@ function image_gd_create_tmp(stdClass $image, $width, $height) {
...
     if ($transparent >= 0) {
-      // The original must have a transparent color, allocate to the new image.
-      $transparent_color = imagecolorsforindex($image->resource, $transparent);
-      $transparent = imagecolorallocate($res, $transparent_color['red'], $transparent_color['green'], $transparent_color['blue']);
-
-      // Flood with our new transparent color.
-      imagefill($res, 0, 0, $transparent);
-      imagecolortransparent($res, $transparent);
+      if ($transparent < imagecolorstotal($image->resource)) {
+        // The original must have a transparent color, try to allocate to the new
```
indigoxela: The resulting "transparent" color seems sort of random. Any color from the gif, but no transparency.

If I override $res inside this if clause `$res = imagecreate($width, $height);` - so create an indicated image instead of truecolor, the transparent background is - transparent.

Not sure, if you want to play with that.

In fact, if I replace the contents of this if clause with:

```
        $res = imagecreate($width, $height);
        imagepalettecopy($res, $image->resource);
```
the colors seem fine. Hm. Most of the time... Note: inside that if clause, so only when `$transparent < imagecolorstotal($image->resource)`.

laryn: I ended up rearranging a bit so we don't create the truecolor image at all for `.gif`. A little more code touched but probably slightly more efficient for this edge case of adding gifs...

core/modules/system/image.gd.inc
```diff
@@ -311,31 +311,36 @@ function image_gd_save(stdClass $image, $destination) {
...
+    else {
+      // The original image doesn't have a transparent color in the palette.
+      // Flood with the closest we have to white instead.
+      imagefill($res, 0, 0, imagecolorclosest($res, 255, 255, 255));
+    }
   }
   else {
-    imagefill($res, 0, 0, imagecolorallocate($res, 255, 255, 255));
+    $res = imagecreatetruecolor($width, $height);
+    if ($image->info['extension'] == 'png') {
```
indigoxela: Caution, webp can have transparency, too - and they'd lose it that way.

Edit: and apparently webp can be animated - you opened a can of worms, @laryn :speak_no_evil:

laryn: Oh, I misread your comment, you already were referring to webp! I think fixed in the latest, although technically a separate bug I think it's okay to fix here.

core/modules/system/image.gd.inc
```diff
@@ -311,31 +311,35 @@ function image_gd_save(stdClass $image, $destination) {
  *   A GD image handle.
  */
 function image_gd_create_tmp(stdClass $image, $width, $height) {
-  $res = imagecreatetruecolor($width, $height);
-
+  $res = (!imageistruecolor($image->resource)) ? imagecreate($width, $height) : imagecreatetruecolor($width, $height);
```
indigoxela: GIF images are always palette images. :wink: 

PNG *can* be palette images, too, but using palette at this point leads to quality loss. I'll try to explain in a comment on the issue.

core/modules/system/image.gd.inc
```diff
@@ -311,31 +311,35 @@ function image_gd_save(stdClass $image, $destination) {
...
     }
-  }
-  elseif ($image->info['extension'] == 'png') {
-    imagealphablending($res, FALSE);
-    $transparency = imagecolorallocatealpha($res, 0, 0, 0, 127);
-    imagefill($res, 0, 0, $transparency);
-    imagealphablending($res, TRUE);
-    imagesavealpha($res, TRUE);
+    else {
+      // The original image doesn't have a transparent color in the palette.
```
indigoxela: The index returned by imagecolortransparent() isn't necessarily part of the palette. At least, some of my tests indicate that.

yorkshire-pudding (approved):
Makes sense and looks ok.

quicksketch:
Merged into 1.x and 1.32.x. Thank you @laryn, @indigoxela, @avpaderno, and @yorkshire-pudding!

PR #5130, opened 2025-10-27
Issue #6992: document hook_menu_block_tree_alter

### LINKED ISSUE #6992: Document hook_menu_block_tree_alter()
## Description of the bug

hook_menu_block_tree_alter() is called here: https://github.com/backdrop/backdrop/blob/1.x/core/modules/system/system.menu.inc#L338 But it isn't documented in `system.api.php`.

## Steps To Reproduce

https://docs.backdropcms.org/search/node/menu_block_tree

## Actual behavior

"Your search yielded no results"

## Expected behavior

Documentation of the function.

herbdool:
Fixes https://github.com/backdrop/backdrop-issues/issues/6992

### DIFF
--- a/core/modules/system/system.api.php
+++ b/core/modules/system/system.api.php
@@ -1225,11 +1225,7 @@ function hook_menu_contextual_links_alter(&$links, $router_item, $root_path) {
 }
 
 /**
- * Alter of the menu block tree and config.
- *
- * This hook is invoked by system_menu_tree_block_data(). The menu tree for the
- * given configuration is passed in by reference. The menu block config is also
- * passed by reference.
+ * Modify a menu structure before it is rendered in a menu block.
  *
  * The tree contains an array of menu links in the order they should be
  * rendered. menu_tree_page_data() explains the data structure in more detail.
--- a/core/modules/system/system.api.php
+++ b/core/modules/system/system.api.php
@@ -1225,6 +1225,33 @@ function hook_menu_contextual_links_alter(&$links, $router_item, $root_path) {
 }
 
 /**
+ * Alter of the menu block tree and config.
+ *
+ * This hook is invoked by system_menu_tree_block_data(). The menu tree for the
+ * given configuration is passed in by reference. The menu block config is also
+ * passed by reference.
+ *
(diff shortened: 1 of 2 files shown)


core/modules/system/system.api.php
```diff
@@ -1224,6 +1224,33 @@ function hook_menu_contextual_links_alter(&$links, $router_item, $root_path) {
...
 
+/**
+ * Alter of the menu block tree and config.
+ *
+ * This hook is invoked by system_menu_tree_block_data(). The menu tree for the
+ * given configuration is passed in by reference. The menu block config is also
+ * passed by reference.
+ *
+ * The tree contains an array of menu links in the order they should be
+ * rendered. menu_tree_page_data() explains the data structure in more detail.
```
quicksketch: Seems like this might be AI-generated. I'd suggest removing the unnecessary parts, since the docs site should cross-reference this for us.

```suggestion
 * Modify a menu structure before it is rendered in a menu block.
 *
 * The tree contains an array of menu links in the order they should be
 * rendered. menu_tree_page_data() explains the data structure in more detail.
```

core/modules/system/system.api.php
```diff
@@ -1224,6 +1224,33 @@ function hook_menu_contextual_links_alter(&$links, $router_item, $root_path) {
...
+ * @see system_menu_tree_block_data()
+ * @see menu_tree_page_data()
+ */
+function hook_menu_block_tree_alter(&$tree, &$config) {
+
+  // Trim to the active path regardless of block settings.
+  system_menu_tree_trim_active_path($tree);
+
+  // Force the clickdown setting to be active.
+  $config['clickdown'] = TRUE;
```
quicksketch: This is not a real setting, might be AI-hallucinated?

```suggestion
  // Force the style to be a tree.
  $config['style'] = 'tree';
```

quicksketch (commented):
I think the real value here is having this hook documented so that it's known that it exists, even if the example and documentation isn't perfect. So I don't need to hold this up, but I'm going to make a few changes here.

quicksketch:
Merged into 1.x and 1.32.x. Thank you @herbdool!

PR #5131, opened 2025-10-27
Added zip PHP extension not loaded warning to system updates page

### LINKED ISSUE #6953: Add "The Zip PHP extension is not loaded" message on additional pages
With PHP 8.2 and above the warning message,

> The Zip PHP extension is not loaded on your server. You will not be able to download any projects using Project Installer until this is fixed.

is displayed on the install new modules page at `admin/modules/install`. This is because php 8.2 and subsequent versions require that the zip php extension must be enabled in php.ini with `extension=zip` whereas before with prior php versions the zip extension was "built in".

It would be helpful to have the message also shown on other pages such as System updates `admin/config/system/updates' and possibly even on the Status Report.

Image

izmeez:
Fixes https://github.com/backdrop/backdrop-issues/issues/6953

In the final rendition the PR only adds the message to system updates (bulk form admin/config/system/updates) and there are a bunch of phpcs changes I undertook and with the help of @indigoxela got them corrected.

### DIFF
--- a/core/modules/installer/installer.browser.inc
+++ b/core/modules/installer/installer.browser.inc
@@ -12,7 +12,7 @@
  * @param string $name
  *   The short name of the project.
  *
- * @return
+ * @return boolean
  *   TRUE if the project is enabled, FALSE otherwise.
  */
 function _installer_browser_is_project_enabled($type, $name) {
@@ -110,7 +110,7 @@ function installer_browser_install_queue_remove($project_name) {
 /**
  * Removes a project from the list of recently installed projects.
  *
- * @param string[] $project_names
+ * @param string $project_names
  *   A list of projects to remove from the installed project list.
  *
  * @since 1.10.1
@@ -231,7 +231,7 @@ function installer_browser_get_queued_releases() {
  * @param array $filters
  *   An associative array of queries to use to filter results.
  *
- * @return
+ * @return array
  *   Returns an array of results.
  */
 function installer_browser_fetch_results($filters) {
@@ -252,8 +252,15 @@ function installer_browser_fetch_results($filters) {
     $results = backdrop_json_decode($response->data);
   }
   else {
-    backdrop_set_message(t("Encountered an error when trying to fetch results from @name. Error @code : @message",
-      array('@name' => $server['name'], '@code' => $response->code, '@message' => $response->error)));
+    backdrop_set_message(
+      t("Encountered an error when trying to fetch results from @name. Error @code : @message",
+        array(
+          '@name' => $server['name'],
+          '@code' => $response->code,
+          '@message' => $response->error,
+        )
+      )
+    );
     return array();
   }
 
@@ -268,7 +275,7 @@ function installer_browser_fetch_results($filters) {
  *
  * The installer_browser_fetch_results($filters) call returns an array with the
  * following keys:
- * @return
+ * @return array
  *   Returns an array representing the active server with the following keys:
  *     - url: The absolute url to the server without query parameters.
  *     - name: A human-readable name for this server.
@@ -299,7 +306,7 @@ function installer_browser_get_server() {
  */
 function installer_browser_get_project_release_data(array $project) {
   $project['project_type'] = $project['type'];
-  // Build the releases cache for this project
+  // Build the releases cache for this project.
   module_load_include('inc', 'update', 'update.fetch');
   if (_update_process_fetch_task($project)) {
     $data = _update_cache_get('available_releases::' . $project['name']);
@@ -314,9 +321,10 @@ function installer_browser_get_project_release_data(array $project) {
 /**
  * Batch API callback: Installs a single release of a project during batch.
  *
- * @param $release_name
+ * @param string $release_name
  *   The short name of the selected release of the project being installed.
- * @param $project
+ *
+ * @param array $project
  *   Associative array of information about the project to be installed.
  */
 function _installer_browser_batch_install_release($release_name, $project, &$context) {
@@ -366,7 +374,7 @@ function _installer_browser_batch_install_releases_finished($success, $results, 
     if (!empty($results)) {
       if (!empty($results['failures'])) {
         backdrop_set_message(format_plural(count($results['failures']), 'Failed to install one project.', 'Failed to install @count projects.'), 'error');
-        foreach($results['failures'] as $message){
+        foreach ($results['failures'] as $message) {
           backdrop_set_message($message, 'error');
         }
       }
--- a/core/modules/installer/installer.browser.inc
+++ b/core/modules/installer/installer.browser.inc
@@ -587,7 +587,7 @@ function installer_browser_add_remove_queue_dialog_link_submit($form, &$form_sta
 function installer_browser_check_zip_loaded() {
   $zip_loaded = array_key_exists('zip', archiver_get_info());
   if (!$zip_loaded) {
-    backdrop_set_message(t('The Zip PHP extension is not loaded on your server. You will not be able to download any projects using Project Installer until this is fixed.'), 'warning', FALSE);
+    backdrop_set_message(t('The Zip PHP extension is not loaded on your server. You will not be able to download any projects using Project Installer until this is fixed. PHP 8.2 and above load as dynamic extension.'), 'warning', FALSE);
   }
   return $zip_loaded;
 }
--- a/core/modules/installer/installer.browser.inc
+++ b/core/modules/installer/installer.browser.inc
@@ -12,7 +12,7 @@
  * @param string $name
  *   The short name of the project.
  *
- * @return boolean
+ * @return bool
  *   TRUE if the project is enabled, FALSE otherwise.
  */
 function _installer_browser_is_project_enabled($type, $name) {
@@ -110,7 +110,7 @@ function installer_browser_install_queue_remove($project_name) {
 /**
  * Removes a project from the list of recently installed projects.
  *
- * @param string $project_names
+ * @param string[] $project_names
  *   A list of projects to remove from the installed project list.
  *
  * @since 1.10.1
@@ -252,15 +252,11 @@ function installer_browser_fetch_results($filters) {
     $results = backdrop_json_decode($response->data);
   }
   else {
-    backdrop_set_message(
-      t("Encountered an error when trying to fetch results from @name. Error @code : @message",
-        array(
-          '@name' => $server['name'],
-          '@code' => $response->code,
-          '@message' => $response->error,
-        )
-      )
-    );
+    backdrop_set_message(t("Encountered an error when trying to fetch results from @name. Error @code : @message", array(
+      '@name' => $server['name'],
+      '@code' => $response->code,
+      '@message' => $response->error,
+    )));
     return array();
   }
 
@@ -275,7 +271,7 @@ function installer_browser_fetch_results($filters) {
  *
  * The installer_browser_fetch_results($filters) call returns an array with the
  * following keys:
- * @return array
+ * @return array|null
  *   Returns an array representing the active server with the following keys:
  *     - url: The absolute url to the server without query parameters.
  *     - name: A human-readable name for this server.
--- a/core/modules/installer/installer.browser.inc
+++ b/core/modules/installer/installer.browser.inc
@@ -591,7 +591,7 @@ function installer_browser_add_remove_queue_dialog_link_submit($form, &$form_sta
 function installer_browser_check_zip_loaded() {
   $zip_loaded = array_key_exists('zip', archiver_get_info());
   if (!$zip_loaded) {
-    backdrop_set_message(t('The Zip PHP extension is not loaded on your server. You will not be able to download any projects using Project Installer until this is fixed. PHP 8.2 and above load as dynamic extension.'), 'warning', FALSE);
+    backdrop_set_message(t('The Zip PHP extension is not loaded on your server. You will not be able to download any projects using Project Installer until this is fixed.'), 'warning', FALSE);
   }
   return $zip_loaded;
 }
--- a/core/modules/installer/installer.manager.inc
+++ b/core/modules/installer/installer.manager.inc
@@ -363,7 +363,7 @@ function installer_manager_update_form_submit($form, &$form_state) {
 /**
  * Batch callback: Performs actions when the download batch is completed.
  *
- * @param boolean $success
+ * @param bool $success
  *   TRUE if the batch operation was successful, FALSE if there were errors.
  *
  * @param array $results
@@ -1177,7 +1177,7 @@ function installer_manager_batch_project_get($project, $url, &$context) {
  * it. However, it is supported here because it is a common configuration on
  * shared hosting, and there is nothing Backdrop can do to prevent it.
  *
- * @return boolean
+ * @return bool
  *   TRUE if local file transfers are allowed on this server, or FALSE if not.
  *
  * @see installer_manager_update_ready_form_submit()
--- a/core/modules/installer/installer.manager.inc
+++ b/core/modules/installer/installer.manager.inc
@@ -622,7 +622,7 @@ function installer_manager_install_modal_callback($form, $form_state) {
  *   The update manager operation we're in the middle of. Can be either 'update'
  *   or 'install'. Use to provide operation-specific interface text.
  *
- * @return boolean
+ * @return bool
  *   TRUE if the update manager should continue to the next step in the
  *   workflow, or FALSE if we've hit a fatal configuration and must halt the
  *   workflow.
@@ -704,25 +704,28 @@ function installer_manager_install_form_validate($form, &$form_state) {
     $not_found = array();
     module_load_include('inc', 'installer', 'installer.browser');
 
-    $modules = installer_browser_fetch_results(
-      array('items_per_page' => 999, 'type' => 'module',)
-    );
+    $modules = installer_browser_fetch_results(array(
+      'items_per_page' => 999,
+      'type' => 'module',
+    ));
     $all_projects = $modules['projects'];
     if (!empty($modules)) {
       $listed_projects = array_keys($modules['projects']);
       $not_found = array_diff($project_names, $listed_projects);
       if ($not_found) {
-        $themes = installer_browser_fetch_results(
-          array('items_per_page' => 999, 'type' => 'theme',)
-        );
+        $themes = installer_browser_fetch_results(array(
+          'items_per_page' => 999,
+          'type' => 'theme',
+        ));
         if (!empty($themes)) {
           $all_projects = array_merge($all_projects, $themes['projects']);
           $listed_projects = array_keys($themes['projects']);
           $not_found = array_diff($not_found, $listed_projects);
           if ($not_found) {
-            $layouts = installer_browser_fetch_results(
-              array('items_per_page' => 999, 'type' => 'layout',)
-            );
+            $layouts = installer_browser_fetch_results(array(
+              'items_per_page' => 999,
+              'type' => 'layout',
+            ));
             if (!empty($layouts)) {
               $all_projects = array_merge($all_projects, $layouts['projects']);
               $listed_projects = array_keys($layouts['projects']);
@@ -977,8 +980,8 @@ function installer_manager_download_project($url, $is_installer = FALSE) {
     }
     else {
       $path = backdrop_get_path('module', 'installer') . '/installer.authorize.inc';
-      system_authorized_init('installer_authorize_run_install', $path,
-        array($project, $updater_name, $project_realpath), t('Update manager')
+      system_authorized_init('installer_authorize_run_install', $path, array(
+        $project, $updater_name, $project_realpath), t('Update manager')
       );
       return array(
         'success' => TRUE,
--- a/core/modules/installer/installer.manager.inc
+++ b/core/modules/installer/installer.manager.inc
@@ -63,6 +63,11 @@ function installer_manager_update_form($form, $form_state) {
 
   $available = update_get_available(TRUE);
 
+  // Show a warning if zip PHP extension not loaded, but allow to browse and
+  // select available releases. Install blocked on downloading updates page.
+  module_load_include('inc', 'installer', 'installer.browser');
+  installer_browser_check_zip_loaded();
+
   // Run some basic health checks, and show messages/warnings accordingly.
   $settings_link = url('admin/reports/updates/settings');
   $update_config = config('update.settings');
--- a/core/modules/installer/installer.manager.inc
+++ b/core/modules/installer/installer.manager.inc
@@ -363,9 +363,10 @@ function installer_manager_update_form_submit($form, &$form_state) {
 /**
  * Batch callback: Performs actions when the download batch is completed.
  *
- * @param $success
+ * @param boolean $success
  *   TRUE if the batch operation was successful, FALSE if there were errors.
- * @param $results
+ *
+ * @param array $results
  *   An associative array of results from the batch operation.
  */
 function installer_manager_download_batch_finished($success, $results) {
@@ -616,11 +617,12 @@ function installer_manager_install_modal_callback($form, $form_state) {
  *
  * @param array $form
  *   Reference to the form array we're building.
+ *
  * @param string $operation
  *   The update manager operation we're in the middle of. Can be either 'update'
  *   or 'install'. Use to provide operation-specific interface text.
  *
- * @return
+ * @return boolean
  *   TRUE if the update manager should continue to the next step in the
  *   workflow, or FALSE if we've hit a fatal configuration and must halt the
  *   workflow.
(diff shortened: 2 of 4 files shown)


core/modules/installer/installer.browser.inc
```diff
@@ -110,7 +110,7 @@ function installer_browser_install_queue_remove($project_name) {
 /**
  * Removes a project from the list of recently installed projects.
  *
- * @param string[] $project_names
+ * @param string $project_names
```
indigoxela: That should be `string[]` - that means, it's _an array of strings_. It's never a string. That would cause the function to break, as it does a foreach on the var. 😉 
Please revert that change.

core/modules/installer/installer.browser.inc
```diff
@@ -12,7 +12,7 @@
  * @param string $name
  *   The short name of the project.
  *
- * @return
+ * @return boolean
```
indigoxela: We don't use "boolean" but `bool` 😉

core/modules/installer/installer.browser.inc
```diff
@@ -252,8 +252,15 @@ function installer_browser_fetch_results($filters) {
     $results = backdrop_json_decode($response->data);
   }
   else {
-    backdrop_set_message(t("Encountered an error when trying to fetch results from @name. Error @code : @message",
-      array('@name' => $server['name'], '@code' => $response->code, '@message' => $response->error)));
+    backdrop_set_message(
+      t("Encountered an error when trying to fetch results from @name. Error @code : @message",
```
indigoxela: It's better to *avoid* line breaks between function parameters. Not every PHP version handles that gracefully.

Only use newlines inside array definitions, which every PHP version can handle.

core/modules/installer/installer.browser.inc
```diff
@@ -268,7 +275,7 @@ function installer_browser_fetch_results($filters) {
  *
  * The installer_browser_fetch_results($filters) call returns an array with the
  * following keys:
- * @return
+ * @return array
```
indigoxela: While it will usually be an array, it *might* also be null in edge cases. Function config_get() *can* return null.

core/modules/installer/installer.manager.inc
```diff
@@ -585,7 +591,8 @@ function installer_manager_install_modal_callback($form, $form_state) {
     $html .= theme('status_messages');
     $html .= backdrop_render($form);
     $title = isset($form['#title']) ? $form['#title'] : 'Manual installation';
-    $commands[] = ajax_command_open_modal_dialog($title, $html, array('width' => 700, 'dialogClass' => 'layout-dialog'));
+    $commands[] = ajax_command_open_modal_dialog($title, $html,
+      array('width' => 700, 'dialogClass' => 'layout-dialog'));
```
indigoxela: Similar to above: avoid newlines between function params, add them inside the array instead:

Example:
```php
$commands[] = ajax_command_open_modal_dialog($title, $html, array(
  'width' => 700,
  'dialogClass' => 'layout-dialog',
));
```

core/modules/installer/installer.manager.inc
```diff
@@ -608,11 +615,12 @@ function installer_manager_install_modal_callback($form, $form_state) {
  *
  * @param array $form
  *   Reference to the form array we're building.
+ *
  * @param string $operation
  *   The update manager operation we're in the middle of. Can be either 'update'
  *   or 'install'. Use to provide operation-specific interface text.
  *
- * @return
+ * @return boolean
```
indigoxela: We don't use the term "boolean", but the shorter `bool`.

core/modules/installer/installer.manager.inc
```diff
@@ -694,19 +702,25 @@ function installer_manager_install_form_validate($form, &$form_state) {
     $not_found = array();
     module_load_include('inc', 'installer', 'installer.browser');
 
-    $modules = installer_browser_fetch_results(array('items_per_page' => 999, 'type' => 'module'));
+    $modules = installer_browser_fetch_results(
+      array('items_per_page' => 999, 'type' => 'module',)
```
indigoxela: Newlines are better used inside array - similar to above.

core/modules/installer/installer.manager.inc
```diff
@@ -694,19 +702,25 @@ function installer_manager_install_form_validate($form, &$form_state) {
...
+    );
     $all_projects = $modules['projects'];
     if (!empty($modules)) {
       $listed_projects = array_keys($modules['projects']);
       $not_found = array_diff($project_names, $listed_projects);
       if ($not_found) {
-        $themes = installer_browser_fetch_results(array('items_per_page' => 999, 'type' => 'theme'));
+        $themes = installer_browser_fetch_results(
+          array('items_per_page' => 999, 'type' => 'theme',)
```
indigoxela: Newlines are better used inside array - similar to above.

core/modules/installer/installer.manager.inc
```diff
@@ -694,19 +702,25 @@ function installer_manager_install_form_validate($form, &$form_state) {
...
+        );
         if (!empty($themes)) {
           $all_projects = array_merge($all_projects, $themes['projects']);
           $listed_projects = array_keys($themes['projects']);
           $not_found = array_diff($not_found, $listed_projects);
           if ($not_found) {
-            $layouts = installer_browser_fetch_results(array('items_per_page' => 999, 'type' => 'layout'));
+            $layouts = installer_browser_fetch_results(
+              array('items_per_page' => 999, 'type' => 'layout',)
```
indigoxela: Newlines are better used inside array - similar to above.

core/modules/installer/installer.manager.inc
```diff
@@ -959,7 +975,9 @@ function installer_manager_download_project($url, $is_installer = FALSE) {
     }
     else {
       $path = backdrop_get_path('module', 'installer') . '/installer.authorize.inc';
-      system_authorized_init('installer_authorize_run_install', $path, array($project, $updater_name, $project_realpath), t('Update manager'));
+      system_authorized_init('installer_authorize_run_install', $path,
+        array($project, $updater_name, $project_realpath), t('Update manager')
```
indigoxela: Newlines are better used inside array - similar to above.

core/modules/installer/installer.pages.inc
```diff
@@ -261,13 +267,19 @@ function installer_browser_installation_select_versions_form($form, &$form_state
...
-        ));
+      ));
       if (isset($release['terms']['Release type']) && !empty($release['terms']['Release type'])) {
         $release_title .= check_plain(" (" . implode(', ', $release['terms']['Release type']) . ")");
       }
       $releases_list[$version] = $release_title;
       if (isset($release['release_link'])) {
-        $releases_list[$version] .= ' (' . l(t('release notes'), $release['release_link'], array('attributes' => array('title' => t('Release notes for @project_title', array('@project_title' => $project['title'])), 'target' => '_blank'))) . ')';
+        $releases_list[$version] .= ' (' . l(t('release notes'), $release['release_link'],
+        array('attributes' => array(
```
indigoxela: Newlines are better used inside array - similar to further above.

core/modules/installer/installer.pages.inc
```diff
@@ -964,14 +981,15 @@ function installer_browser_get_sort_widget($sort_options, $current_order, $curre
...
+    'attributes' => array('class' => array('installer-browser-sort-list',
+      'clearfix')),
   ));
 }
 
 /**
  * Returns a list of sort options.
  *
- * @param $full
+ * @param boolean $full
```
indigoxela: We're using `bool` instead.

core/modules/system/system.admin.inc
```diff
@@ -267,7 +270,12 @@ function system_themes_page() {
 
   $admin_form = backdrop_get_form('system_themes_admin_form', $admin_theme_options);
 
-  return theme('system_themes_page', array('theme_groups' => $theme_groups, 'theme_group_titles' => $theme_group_titles)) . backdrop_render($admin_form);
+  return theme('system_themes_page',
+      array(
```
indigoxela: Newlines are better used inside array - similar to further above.

core/modules/system/system.admin.inc
```diff
@@ -530,14 +538,14 @@ function system_theme_settings_submit($form, &$form_state) {
...
  *
- * @param $incompatible
+ * @param array $incompatible
  *   An associative array which at the end of the check contains all
  *   incompatible files as the keys, their values being TRUE.
- * @param $files
+ * @param array $files
  *   The set of files that will be tested.
- * @param $file
+ * @param string $file
```
indigoxela: This isn't a string, but a File (instance of class).

`@param File $file`

This "File" is the name of the class the object is an instance of.

core/modules/system/system.admin.inc
```diff
@@ -530,14 +538,14 @@ function system_theme_settings_submit($form, &$form_state) {
...
  *   The file at which the check starts.
- * @return
+ * @return TRUE/NULL
```
indigoxela: In doc blocks we're using lowercase, so this should look like:

`@return true|null`

Note also, that multiple types are separated by "|" not by "/".

core/modules/system/system.admin.inc
```diff
@@ -858,15 +877,17 @@ function _system_modules_build_row($info, $extra) {
...
   if (version_compare(phpversion(), $info['php']) < 0) {
     $compatible = FALSE;
     $status_short .= t('Incompatible with this version of PHP') . ' ';
     $php_required = $info['php'];
     if (substr_count($info['php'], '.') < 2) {
       $php_required .= '.*';
     }
-    $status_long .= t('This module requires PHP version @php_required and is incompatible with PHP version !php_version.', array('@php_required' => $php_required, '!php_version' => phpversion())) . ' ';
+    $status_long .= t('This module requires PHP version @php_required and is incompatible with PHP version !php_version.',
+      array('@php_required' => $php_required, '!php_version' => phpversion())
```
indigoxela: Newlines are better used inside array - similar to above.

core/modules/system/system.admin.inc
```diff
@@ -1202,9 +1227,9 @@ function system_modules_uninstall($form, $form_state = NULL) {
  * Confirm uninstall of selected modules.
  *
  * @ingroup forms
- * @param $storage
+ * @param array $storage
  *   An associative array of modules selected to be uninstalled.
- * @return
+ * @return array
```
indigoxela: The 3rd line inside that function indicates, that this can return void (nothing at all).

`@return array|void`

core/modules/system/system.admin.inc
```diff
@@ -1721,14 +1748,19 @@ function system_logging_settings($form, &$form_state) {
 function system_logging_settings_validate($form, &$form_state) {
   // Cleanup enabled severity levels. Form API sets the value of unselected
   // checkboxes to 0.
-  $form_state['values']['watchdog_enabled_severity_levels'] = array_keys(array_filter($form_state['values']['watchdog_enabled_severity_levels'], function($v){return $v !== 0;}));
+  $form_state['values']['watchdog_enabled_severity_levels'] = array_keys(
```
indigoxela, line 1750: It's OK to *not* add any newlines here, as this adds one inside function call params (array_keys()), which not all PHP versions handle properly.

core/modules/installer/installer.browser.inc
izmeez: Should this be 
`* @return array|null`
instead?

core/modules/installer/installer.browser.inc
izmeez: When I revert it my local phpcs reports: `Type hint array missing for $project_names`

core/modules/system/system.admin.inc
izmeez: My local phpcs does not show "File" with color change like other correct @param. But, I have changed it.

core/modules/installer/installer.browser.inc
indigoxela: Type hints look like this:

```php
function foobar(array $foo, array $bar) {
  ...
```
But unlike doc blocks, which do no harm other than being misleading when wrong, a wrong type hint has real negative effects on functionality. That's why type hints require extra care. :wink: 
If "$foo" were no array when called anywhere, there would be a hard TypeError.

Our ruleset is configured to require type hints when it's an array. That's why it nags.
But you don't have to do anything, actually, as this isn't part of your actual code change.

I wonder, why you decided to fix many phpcs violations... It's totally voluntary in PR that try to achieve something other than coding standard fixes.
Highly appreciated, though. :wink:

core/modules/installer/installer.browser.inc
```diff
@@ -586,7 +590,7 @@ function installer_browser_add_remove_queue_dialog_link_submit($form, &$form_sta
 function installer_browser_check_zip_loaded() {
   $zip_loaded = array_key_exists('zip', archiver_get_info());
   if (!$zip_loaded) {
-    backdrop_set_message(t('The Zip PHP extension is not loaded on your server. You will not be able to download any projects using Project Installer until this is fixed.'), 'warning', FALSE);
+    backdrop_set_message(t('The Zip PHP extension is not loaded on your server. You will not be able to download any projects using Project Installer until this is fixed. PHP 8.2 and above load as dynamic extension.'), 'warning', FALSE);
```
indigoxela: Where does the info "PHP 8.2 and above load as dynamic extension." come from?

In Debian it's a separate package. Other distributions may handle that differently, too. Let alone custom PHP builds from hosting providers.

Besides that, this doesn't actually provide much value for site builders, I guess. 😉

core/modules/installer/installer.manager.inc
```diff
@@ -358,9 +363,10 @@ function installer_manager_update_form_submit($form, &$form_state) {
 /**
  * Batch callback: Performs actions when the download batch is completed.
  *
- * @param $success
+ * @param boolean $success
```
indigoxela: `@param bool $success`

We're using the short form. 😉

core/modules/installer/installer.manager.inc
```diff
@@ -1106,7 +1131,9 @@ function installer_manager_batch_project_get($project, $url, &$context) {
 
   // Actually try to download the file.
   if (!($local_cache = installer_manager_file_get($url))) {
-    $context['results']['errors'][$project] = t('Failed to download %project from %url', array('%project' => $project, '%url' => $url));
+    $context['results']['errors'][$project] = t('Failed to download %project from %url',
```
indigoxela: This newline is inside t() params, better move it inside the array.

core/modules/installer/installer.manager.inc
```diff
@@ -1150,7 +1177,7 @@ function installer_manager_batch_project_get($project, $url, &$context) {
  * it. However, it is supported here because it is a common configuration on
  * shared hosting, and there is nothing Backdrop can do to prevent it.
  *
- * @return
+ * @return boolean
```
indigoxela: We're using `bool`, not boolean.

core/modules/installer/installer.pages.inc
```diff
@@ -535,7 +551,9 @@ function installer_browser_installation_theme_form($form, &$form_state, $project
 
   $form['instructions'] = array(
     '#type' => 'item',
-    '#markup' => t('The themes you selected have been successfully installed. You may enable installed themes here or on the main !themes page. ', array('!themes' => l(t('Themes'), 'admin/appearance')))
+    '#markup' => t('The themes you selected have been successfully installed. You may enable installed themes here or on the main !themes page. ',
+      array('!themes' => l(t('Themes'), 'admin/appearance'))
```
indigoxela: Better move those newlines into array (instead of between t() params).

core/modules/system/system.admin.inc
```diff
@@ -1069,7 +1093,7 @@ function system_modules_submit($form, &$form_state) {
     }
   }
 
-  // Gets list of modules prior to install process, unsets $form_state['storage']
+  // Gets list of modules prior to install process, unset $form_state['storage']
```
indigoxela, line 1096: "unsets" with s was actually correct there, no?

core/modules/system/system.admin.inc
```diff
@@ -1122,9 +1146,9 @@ function system_modules_submit($form, &$form_state) {
  * @ingroup forms
  * @see system_modules_uninstall_validate()
  * @see system_modules_uninstall_submit()
- * @param $form_state['values']
+ * @param array $form_state
```
indigoxela: Obviously the default value for $form_state is NULL, so this should rather be:

`@param array|null $form_state`

core/modules/system/system.admin.inc
```diff
@@ -1721,14 +1747,18 @@ function system_logging_settings($form, &$form_state) {
 function system_logging_settings_validate($form, &$form_state) {
   // Cleanup enabled severity levels. Form API sets the value of unselected
   // checkboxes to 0.
-  $form_state['values']['watchdog_enabled_severity_levels'] = array_keys(array_filter($form_state['values']['watchdog_enabled_severity_levels'], function($v){return $v !== 0;}));
+  $form_state['values']['watchdog_enabled_severity_levels'] = array_keys(array_filter($form_state['values']['watchdog_enabled_severity_levels'],
```
indigoxela: Better revert that? Those newlines are inside array_filter().

core/modules/installer/installer.manager.inc
izmeez: Turns out there were actually 2 more still remaining in the file that have now been fixed to use the short form.

core/modules/installer/installer.manager.inc
izmeez: Fixed.

core/modules/system/system.admin.inc
izmeez: Actually, I tossed this around thinking it reads ok either way and makes the line fit 80 chars. But I can revert it.

core/modules/installer/installer.pages.inc
izmeez: I think I've done what you are suggesting. I then see the @param is not happy so I made it an array and then I got a type hint "array" missing and changed it. Now it looks like this:
```
/**
 * Form builder for theme install.
 *
 * @param array $projects
 *   An array of projects selected for install, keyed by project name.
 */
function installer_browser_installation_theme_form($form, &$form_state, array $projects) {
  backdrop_add_css(backdrop_get_path('module', 'installer') . '/css/installer.css', array('preprocess' => FALSE));
  module_load_include('inc', 'installer', 'installer.browser');

  $form['instructions'] = array(
    '#type' => 'item',
    '#markup' => t('The themes you selected have been successfully installed. You may enable installed themes here or on the main !themes page. ', array('!themes' => l(t('Themes'), 'admin/appearance'))),
  );

If this is good I can commit it. Thanks.
```

core/modules/installer/installer.browser.inc
izmeez: What a sleuth you are. There is where I got the info, https://www.php.net/manual/en/zip.installation.php and yes it is peculiar to Windows: "As of PHP 8.2.0, php_zip.dll DLL must be [enabled](https://www.php.net/manual/en/install.pecl.windows.php#install.pecl.windows.loading) in php.ini. Previously, this extension was built-in."
Now, I'm beginning to feel this whole endeavor may have been for a small percentage of sites, certainly not 80%. I can update the wording, if that's any help?

core/modules/system/system.admin.inc
izmeez: I have reverted it but phpcs is saying the closing brace must be on a line by itself.

indigoxela: Ah, got it. Just ignore the nagging in this case. Seems like our coding standard isn't considering anonymous functions.

core/modules/installer/installer.pages.inc
indigoxela: Try:
```php
  $form['instructions'] = array(
    '#type' => 'item',
    '#markup' => t('The themes you selected have been successfully installed. You may enable installed themes here or on the main !themes page. ', array(
      '!themes' => l(t('Themes'), 'admin/appearance'))),
  );
```

core/modules/system/system.admin.inc
indigoxela: Then just do as you like. :wink:

izmeez: I'll stay with unset.

core/modules/installer/installer.pages.inc
izmeez: I have two questions:
1. When I add `* @param array $projects` I get the type hint missing array and when I modify `function installer_browser_installation_theme_form($form, &$form_state, array $projects)` it goes away. Is this the right solution?
2. My local phpcs is not complaining when `#markup` is all on one line. Is there any need to change it?

core/modules/installer/installer.browser.inc
indigoxela: Oh, Windows. That's an edge case, as on "real" hosting it will never be Windows. 😉 

Maybe just revert to the original text - which was fine, anyway...

core/modules/installer/installer.manager.inc
indigoxela: And this is still "boolean" 😉

core/modules/installer/installer.browser.inc
izmeez: I have reverted the message to the original removing "PHP 8.2 and above load as dynamic extension." Thank you for catching it is only specific to some environments.

core/modules/installer/installer.manager.inc
indigoxela: Hm, GitHub permissions also don't let me "unresolve", so here's a new comment: it's still "boolean" here instead of "bool" 😉

core/modules/installer/installer.manager.inc
indigoxela: Still "boolean" - did you forget to push a commit, eventually?

core/modules/installer/installer.manager.inc
izmeez: I have not pushed my commits. I was waiting on resolving the remaining questions.

core/modules/installer/installer.manager.inc
izmeez: Sorry, had held off on the push.

core/modules/installer/installer.pages.inc
izmeez: This is the big question for me. I do not want to create problems with how I handled the type hint.

indigoxela: 1. All the coding standard fixes you do here, are totally voluntary. You don't have to change anything you feel uncomfortable with. Yes, if you add "array" in the doc block, our phpcs ruleset suggests to also add the type hint.
2. Again, no requirement at all to change anything you don't want to change. Newly added code is supposed to conform to our ruleset, but our GitHub action for phpcs will only nag on violations in "touched code", not on existing code you didn't touch, anyway.

We still have loads of coding standard violations all over core code - for historic reasons. Every improvement of existing code matters and is welcome. But it's never a requirement on code, that's not actually part of the desired feature/bugfix.

core/modules/system/system.admin.inc
```diff
@@ -1721,14 +1747,18 @@ function system_logging_settings($form, &$form_state) {
...
+    function($v) {
+      return $v !== 0;
+    })
```
indigoxela: My previous comment was a bit misleading...

Ignore the phpcs nagging when it's all in one line - stick with the code as it was.
That's what I actually meant. So my suggestion is to revert that part to what it was initially.

core/modules/installer/installer.manager.inc
izmeez: Pushed now.

core/modules/installer/installer.pages.inc
izmeez: Tugboat is also not complaining. So if adding the array to the doc block and also as I did for the type hint is correct then I am happy.

I was following the community request to try to fix phpcs issues when touching a file. Obviously, I bit of more than I could chew. Now, I see there are 4 other type hints, i may be able to fix from what I have learned. But probably best to leave them alone.

core/modules/system/system.admin.inc
```diff
@@ -2343,10 +2380,14 @@ function system_urls_settings($form, &$form_state) {
...
  *   This parameter will be removed in Backdrop 2.0.
  *
- * @see system_get_requirements().
+ * @see system_get_requirements()
  */
 function system_status($check = FALSE) {
-  // Load .install files
+  // Show a warning if zip PHP extension not loaded.
+  module_load_include('inc', 'installer', 'installer.browser');
+  installer_browser_check_zip_loaded();
```
indigoxela: This function ships with a module that might be off. Causes no harm, as you have to do an include, anyway, but... 🤔 

More about my concerns in the issue's thread.

izmeez: I have decided to remove this, reverting the code.

core/modules/installer/installer.manager.inc
```diff
@@ -63,6 +63,11 @@ function installer_manager_update_form($form, $form_state) {
 
   $available = update_get_available(TRUE);
 
+  // Show a warning if zip PHP extension not loaded, but allow to browse and
+  // select available releases. Install blocked on downloading updates page.
+  module_load_include('inc', 'installer', 'installer.browser');
+  installer_browser_check_zip_loaded();
```
quicksketch, line 69: Here's the additional check that solves the requested issue.

izmeez: Yes, this is the change the issue is about and introduces. The rest is cleanup phpcs stuff.

quicksketch:
Merged into 1.x and 1.33.x. Thank you @izmeez and @indigoxela!

PR #5143, opened 2025-10-30
Issue #6506: Convert card images to SVGs.

### LINKED ISSUE #6506: Replace the images for the 3 cards in the home page tiles with SVG
In #4903 we introduced a new content type to the existing 2 that get created OOTB in new installations of the standard profile, called "Card". We are also creating 3 pieces of content of that type and are placing them in the home page, in the form of 3 horizontal tiles, next to each other:

image

The images used for these 3 tiles are .png. People in Zulip (most recently in https://backdrop.zulipchat.com/#narrow/stream/218635-Backdrop/topic/1.2E28.20Release/near/436889361) have suggested that we swap those with SVG.

NormPlum:
Fixes https://github.com/backdrop/backdrop-issues/issues/6506


core/themes/basis/css/component/cards.css
```diff
@@ -54,6 +54,13 @@
 .view-promoted-cards .views-field-field-image {
   padding: 0;
 }
+.view-promoted-cards .views-field-field-image img {
+  max-width: 6rem;
+  height: auto;
+  display: block;
+  margin: 0 auto;
+  padding: 4rem 0;
+}
```
indigoxela, line 63: That's dangerous.

We can't assume, that site builders don't change the image (via image field UI). And those very likely would look awful with those styles applied. :grimacing: ... If they're using Basis.

PR #5149, opened 2025-10-31
Stop run-tests.sh script from truncating simpletests results

### LINKED ISSUE #7000: run-tests.sh script truncates test results output, making debugging sometimes difficult
## Description of the need

In `/core/scripts/run-tests.sh` L#948 the testing script is truncating output from test results. I think someone probably did this so the column data width would somewhat match the column header width. I think matching widths is not as important as getting all the debugging output when working on tests:

```
  $summary = sprintf(
    "%-9.9s %-10.10s %-17.17s %-7.7s %-72.72s\n",
    $results_map[$result->status],
    $result->message_group,
    basename($result->file),
    $result->line,
    $function
  );
```

This can lead to frustrating results output, especially in the contrib space where filenames can be long, or when debugging test failures or exceptions that only happen through github actions. My workaround so far has been to fork backdrop and run contrib tests against a patched version of core to see results in GHA.

Here is an example of truncated results:

Image

Notice `entity_token.toke` or `commerce_product_` as truncated filename results. The latter could match several files one would need to hunt through to find out where this sort of issue _could_ potentially be happening.

## Proposed solution

Do not truncate the results output from simpletest. It's more important to see full results than to have columns line up.

We could potentially just increase the length, but I think inevitably someone else will run into this again if they have really long filenames in a module with tests.

elisseck:
Fixes [#7000](https://github.com/backdrop/backdrop-issues/issues/7000)

### DIFF
--- a/core/scripts/run-tests.sh
+++ b/core/scripts/run-tests.sh
@@ -945,7 +945,7 @@ function simpletest_script_format_result($result) {
   list($class, $function) = explode('->', $result->function, 2);
 
   $summary = sprintf(
-    "%-9.9s %-10.10s %-17.17s %-7.7s %-72.72s\n",
+    "%-9s %-10s %-17s %-7s %-72s\n",
     $results_map[$result->status],
     $result->message_group,
     basename($result->file),


quicksketch:
Merged into 1.x and 1.33.x. Thanks @elisseck!

PR #5152, opened 2025-11-01
Issue #7003: Fix timezone field width.

### LINKED ISSUE #7003: Fix timezone field length on registration form
## Description of the bug

On the user registration form, when the timezone field is displayed, it's so long that it overflows its container.

## Steps To Reproduce

To reproduce the behavior:
1. Allow visitors to register accounts
2. In Regional Settings, enable 'Users may set their own time zone at registration'
3. Due to a [bug](https://github.com/backdrop/backdrop-issues/issues/6073) you also need to enable the Locale module and add an additional language to the site (so there are more than 1 languages)
4. Go to the Registration form

## Actual behavior

Timezone field extends outside its container but is cut off.

Image

## Expected behavior

Timezone field fills its containers width.

Image

## Additional information

- Backdrop CMS version: 1.33.x-dev

NormPlum:
Fixes https://github.com/backdrop/backdrop-issues/issues/7003

### DIFF
--- a/core/misc/vertical-tabs.css
+++ b/core/misc/vertical-tabs.css
@@ -45,11 +45,11 @@
 .vertical-tab-selected .fieldset-wrapper {
   display: block;
 }
-.vertical-tabs-pane button,
-.vertical-tabs-pane input,
-.vertical-tabs-pane optgroup,
-.vertical-tabs-pane select,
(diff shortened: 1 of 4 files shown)


quicksketch:
Merged into 1.x and 1.32.x. Thank you @NormPlum!

PR #5153, opened 2025-11-03
Issue #7005: Add 'admin-bar-fixed' class.

### LINKED ISSUE #7005: Add HTML class for 'admin-bar-fixed'
## Description of the need

When the admin bar is enabled a class is added to the HTML element: `admin-bar`. This is helpful, but it only indicates whether or not the admin bar is enabled.

## Proposed solution

Can we add another class of `admin-bar-fixed` when the "Keep bar at top of page" setting is checked? This would allow other CSS to be used when the admin bar is sticky vs. when its not.

The admin bar div itself already has a class of `admin-bar-position-fixed` but that's not useful to other CSS on the page.

Some contrib themes add their own class for this purpose, e.g. https://github.com/backdrop-contrib/scenery/blob/1.x-1.x/template.php#L17

Backdrop now includes an `admin-bar-fixed` class on the HTML element to indicate when the Admin Bar is fixed to the top of the page.

NormPlum:
Fixes https://github.com/backdrop/backdrop-issues/issues/7005

### DIFF
--- a/core/modules/admin_bar/js/admin_bar.js
+++ b/core/modules/admin_bar/js/admin_bar.js
@@ -146,6 +146,7 @@ Backdrop.adminBar.behaviors.adminBarMarginTop(document, Backdrop.settings);
  */
 Backdrop.adminBar.behaviors.positionFixed = function (context, settings, $adminBar) {
   if (settings.admin_bar.position_fixed) {
+    $('html:not(.admin-bar-fixed)', context).addClass('admin-bar-fixed');
     $adminBar.addClass('admin-bar-position-fixed');
     $adminBar.css('position', 'fixed');
 


core/modules/admin_bar/js/admin_bar.js
```diff
@@ -146,6 +146,7 @@ Backdrop.adminBar.behaviors.adminBarMarginTop(document, Backdrop.settings);
  */
 Backdrop.adminBar.behaviors.positionFixed = function (context, settings, $adminBar) {
   if (settings.admin_bar.position_fixed) {
+    $('html:not(.admin-bar-fixed)', context).addClass('admin-bar-fixed');
```
indigoxela, line 149: There's just one problem with that: we also have to consider permissions. Not all roles see the admin-bar.

OK, harmless here, as this doesn't add any styles, but later one we'd need the permissions in addition to the fixed flag. And in CSS we won't have it.

Another possible pitfall: the admin bar caches heavily.

Edit: hm, maybe I'm just over-thinking it.

indigoxela: Ha, silly me, totally overthinking... Backdrop.adminBar.behaviors... are only ever called when the admin-bar is attached. No need for extra considerations.

So this is resolved. :+1:

NormPlum:
Closing as we can use `html:has(.admin-bar-position-fixed)` instead of adding another class.

PR #5156, opened 2025-11-06
7009: Add API for setting _config_static flag for config caching.

### LINKED ISSUE #7009: API and documentation for _config_static cache
## Description of the need

The ability to cache configuration files in memory is poorly documented and in some ways "hidden", in that it does not have any way to enable/disable it other than to know the special key `_config_static` in config files. We've run into several places in core where this flag is misconfigured or misunderstood:

- https://github.com/backdrop/backdrop-issues/issues/6982
- https://github.com/backdrop/backdrop-issues/issues/5800
- https://github.com/backdrop/backdrop-issues/issues/856
- https://github.com/backdrop/backdrop-issues/issues/6961

## Proposed solution

We should introduce an API for setting this special flag, which will provide a location to document how and why to use it. It's also an opportunity to expand test coverage.

The API for configuration management now includes methods for enabling and disabling in-memory caching. While the cache is not new, the new APIs provide documentation and make it easier for modules to enable configuration caching.

quicksketch:
Fixes https://github.com/backdrop/backdrop-issues/issues/7009

### DIFF
--- a/core/includes/config.inc
+++ b/core/includes/config.inc
@@ -999,7 +999,7 @@ class Config {
    *
    * This cache can substantially reduce reads from disk by allowing multiple
    * calls to config() or config_get() to read from an in-memory cache that is
-   * is populated once per request.
+   * populated once per request.
    *
    * Any config that is read more than once per request should be cached in some
    * way. Some more advanced concepts like Views, Fields, or Content Types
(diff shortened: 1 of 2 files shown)


core/includes/config.inc
```diff
@@ -978,6 +994,35 @@ class Config {
     $this->overrides = $overrides;
   }
 
+  /**
+   * Sets the flag for whether this config is statically stored in memory.
+   *
+   * This cache can substantially reduce reads from disk by allowing multiple
+   * calls to config() or config_get() to read from an in-memory cache that is
+   * is populated once per request.
```
argiepiano: The word "is" is repeated twice.

```suggestion
   * populated once per request.
```

quicksketch: Thanks, fixed!

PR #5157, opened 2025-11-06
Issue #7009: Remove unused default_status_code from Redirect module.

### LINKED ISSUE #7006: Redirect references nonexistent `default_status_code` config value on fresh installs
## Description of the bug

The redirect module successfully upgrades a D7 variable to config, but does not include the config out of the box for non-upgrades.

## Steps To Reproduce

Compare the redirect configuration for a D7 upgraded site to that of a fresh install. 

## Actual behavior

The `default_status_code` configuration is set for upgrades, but not for fresh installs. This configuration option is also not exposed in the UI, which may be intentional or not.

## Expected behavior

The configuration value (301) should be set to the default for non-upgrades. We should determine whether or not the value should also be exposed in Redirect's settings page.

## Additional information

Redirect's `redirect_update_1000()` updates the D7 variable to a config value under `default_status_code` but the default config file does not include this value. We'd need to update the default config file as well as add an update hook to set the value if it is NULL.

quicksketch:
Fixes https://github.com/backdrop/backdrop-issues/issues/7006

### DIFF
--- a/core/modules/redirect/redirect.admin.inc
+++ b/core/modules/redirect/redirect.admin.inc
@@ -849,8 +849,6 @@ function redirect_list_404_filter_form_reset($form, &$form_state) {
  */
 function redirect_list_table($redirects, $header) {
   $destination = backdrop_get_destination();
-  $config = config('redirect.settings');
-  $default_status_code = $config->get('default_status_code');
 
   // Set up the header.
   $header = array_combine($header, $header);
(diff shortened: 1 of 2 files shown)


core/modules/redirect/redirect.admin.inc
```diff
@@ -877,7 +875,7 @@ function redirect_list_table($redirects, $header) {
     $redirect_url = redirect_url($redirect->redirect, array_merge($redirect->redirect_options, array('alias' => TRUE)));
     $row['data']['source'] = l($source_url, $redirect->source, $redirect->source_options);
     $row['data']['redirect'] = l($redirect_url, $redirect->redirect, $redirect->redirect_options);
-    $row['data']['status_code'] = $redirect->status_code ? $redirect->status_code : t('Default (@default)', array('@default' => $default_status_code));
+    $row['data']['status_code'] = $redirect->status_code;
```
quicksketch, line 878: `$redirect->status_code` is guaranteed to be set because `$redirect` is a class of `Redirect`, which always initializes the status_code property.

core/modules/redirect/redirect.install
```diff
@@ -171,7 +171,7 @@ function redirect_update_1003() {
   $config = config('redirect.settings');
   $default_status = $config->get('default_status_code');
   $additional_statuses = !in_array($default_status, array(301, 302));
-  $config->clear('default_status');
+  $config->clear('default_status_code');
```
quicksketch, line 174: This appears to have been a typo that prevented the value from getting cleared previously.

quicksketch:
Thank you @bugfolder for reviewing! Merged into 1.x and 1.32.x.

PR #5158, opened 2025-11-11
Issue #5862: Fix fragment identifier targets from being covered

### LINKED ISSUE #5862: Fix fragment identifier targets from being covered
## Description of the bug

With the _Administration Bar_ module disabled clicking on such links you can see at least the part of the relevant content:

Screenshot 2022-11-15 at 3 39 44 PM

However, Backdrop comes with _Administration Bar_ module enabled and it remains enabled on most of the websites, so clicking the same link takes you too far and you are not able to see relevant content at all. You would have to scroll little bit up to see the right content:

Screenshot 2022-11-15 at 4 01 46 PM

## Expected behavior

Ideally, you should be able to see not only the part of the relevant content, but all of it including the title like so:

Screenshot 2022-11-15 at 3 56 25 PM

I see from various publications like, for example, https://css-tricks.com/hash-tag-links-padding/ it's a common browser issue. However, Backdrop's _Administration Bar_ module is making it even worse. I am not sure what's the ideal solution here, so suggestions are welcome.

indigoxela:
Fixes https://github.com/backdrop/backdrop-issues/issues/5862

### DIFF
--- a/core/modules/admin_bar/admin_bar.module
+++ b/core/modules/admin_bar/admin_bar.module
@@ -251,6 +251,7 @@ function admin_bar_preprocess_page(&$variables) {
   }
   if ($setting = $config->get('position_fixed')) {
     $settings['position_fixed'] = $setting;
+    $variables['html_attributes']['class'][] = 'admin-bar-fixed';
   }
 
   // Send whether this is an admin page to JS for back to site functionality.
--- a/core/modules/admin_bar/css/admin_bar.css
+++ b/core/modules/admin_bar/css/admin_bar.css
@@ -19,7 +19,7 @@ html.admin-bar-fixed :target {
   scroll-margin-top: var(--admin-bar-height);
 }
 html.admin-bar-fixed table.sticky-enabled :target {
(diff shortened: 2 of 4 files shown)


yorkshire-pudding (approved):
@indigoxela 
Thank you for this.  I have reviewed the CSS alongside a quick test and I think it is fine based on the following observations:

- The place each bit of CSS is in the file seem to be the most logical places
- The CSS in `admin_bar.css` is clearly commented to explain what it does and the section in `system.css` is under a "TableHeader behaviour" heading which also covers this.
- The CSS uses existing classes so unlikely to cause any clashes or unexpected behaviour.
- Adding the existing class via PHP to ensure it is there on page load makes sense.
- The `--admin-bar-height` variable is used which makes sense
- The values make sense and add up when admin bar and table header combined.  I did wonder about a custom property for table header height and doing a calculated combined value but that widens the scope and may have limited wider benefit and so I concluded that using the values makes sense here.

herbdool:
Merged into 1.x and 1.34.x. Thanks @indigoxela

PR #5159, opened 2025-11-14
Issue #7008: PHP 8.5 support

### LINKED ISSUE #7008: PHP 8.5 Support
PHP 8.5 will be released November 20, 2025. Compared to previous releases, the list of changes that will affect us seem to be minimal:

According to https://php.watch/versions/8.5:

## Deprecated functionality:

- [All MHASH_* constants deprecated](https://php.watch/versions/8.5/mhash-consts-deprecated)
- [Non-canonical scalar type casts (boolean|double|integer|binary) deprecated](https://php.watch/versions/8.5/boolean-double-integer-binary-casts-deprecated)
- [Returning non-string values from a user output handler is deprecated](https://php.watch/versions/8.5/ob_handler-return-non-string)
- [Emitting output from custom output buffer handlers is deprecated](https://php.watch/versions/8.5/ob-handler-output-deprecated)

Of those, it seems that only the second one may be relevant to Backdrop. Doing a quick search I found that casting as `(boolean)` is rare but happens in a few places. Other scalars seem to be fine.

We should increment our supported PHP version to 8.5, and bump the test runner to run on 8.5 when ready.

indigoxela:
Fixes https://github.com/backdrop/backdrop-issues/issues/7008

### DIFF
--- a/core/modules/field/modules/list/tests/list.test
+++ b/core/modules/field/modules/list/tests/list.test
@@ -423,7 +423,7 @@ class ListFieldUITestCase extends FieldTestCase {
   }
 
   /**
-   * List (bool) : test 'On/Off' values input.
+   * List (boolean) : test 'On/Off' values input.
    */
   function testListAllowedValuesBoolean() {
     $this->field_name = 'field_list_boolean';
@@ -563,7 +563,7 @@ class ListFieldDisplayTestCase extends FieldTestCase {
   }
 
   /**
-   * List (bool) : test 'On/Off' values input.
+   * List (boolean) : test 'On/Off' values input.
    */
   function testListDisplayFormatterBoolean() {
     $this->field_name = 'field_list_boolean';
--- a/core/modules/field/modules/list/tests/list.test
+++ b/core/modules/field/modules/list/tests/list.test
@@ -423,7 +423,7 @@ class ListFieldUITestCase extends FieldTestCase {
   }
 
   /**
-   * List (boolean) : test 'On/Off' values input.
+   * List (bool) : test 'On/Off' values input.
    */
   function testListAllowedValuesBoolean() {
     $this->field_name = 'field_list_boolean';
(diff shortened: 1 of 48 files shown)


core/modules/image/tests/image.test
```diff
@@ -2035,7 +2035,9 @@ class ImageStyleFloodProtection extends BackdropWebTestCase {
     for ($n = 0; $n < $count; $n++) {
       if (gettype(${'handle' . $n}) === 'resource') {
         curl_multi_remove_handle($multi_handle, ${'handle' . $n});
-        curl_close(${'handle' . $n});
+        if (version_compare(PHP_VERSION, '8.0') < 0) {
+          curl_close(${'handle' . $n});
+        }
```
quicksketch, line 2040: Nice callout here. `curl_close()` does nothing in PHP 8.0+, so we could have added this a while ago.

https://www.php.net/curl_close

core/modules/field/modules/list/tests/list.test
```diff
@@ -423,7 +423,7 @@ class ListFieldUITestCase extends FieldTestCase {
   }
 
   /**
-   * List (boolean) : test 'On/Off' values input.
+   * List (bool) : test 'On/Off' values input.
```
herbdool: These don't need to be changed since they're just comments, not docblock.

indigoxela: I changed them for consistency - as the doc block syntax is in use.
Sure, not really necessary, but for clarity, I'd suggest to use the same syntax the doc blocks use.

Does it seem important to leave these as `(boolean)`? If so - for which reason? @herbdool

herbdool: Mostly because boolean is the actual word.

herbdool: But... it's a minor quibble so I am fine either way.

indigoxela: I'm also fine with either - just wanted to know. :wink:

quicksketch:
Merged into 1.x for 1.33.0. Thank you @indigoxela!

PR #5160, opened 2025-11-14
Issue #7010: Add list of modules requiring db schema updates to status report.

### LINKED ISSUE #7010: List modules requiring updates to run on Status report
## Description of the need

As a site administrator or someone who is upgrading from Drupal 7
I want to know which modules are pending updates and the reference of the update
So that any that appear stuck can be easily tracked down and I can find out more before applying the updates.

There have been numerous reports in the forum (e.g. https://forum.backdropcms.org/forum/database-update-warning-no-pending-updates) and Zulip of sites saying there are updates but nothing showing or running and no way to track down the module.  This is what people see:

Image

That forum post has a suggestion of hacking the core status report to list a module (assuming that there is only one) but surely we should make this easy for people and enable them to quickly track down any problem modules as well as providing pertinent information in normal usage about which modules they can expect to run an update hook, in case they want to find out more before starting the update.

### Background
The Status Report uses a different method to check for updates based on the current schema version (`backdrop_get_installed_schema_version($module)` - has the special treatment for `6000` - `6999` update hooks) for each module and the available schema versions (`backdrop_get_schema_versions($module)` from which it extracts the last update available and compares it with the current version.

In comparison, when actually doing an update functions such as `backdrop_load_updates()` and `update_get_update_list()` are used.

Any difference in these results will cause the problem that this feature request and PR is intended to help with.

## Proposed solution
Instead of saying "Some modules" it would list them. Something like:

> The modules listed below have database schema updates to install. You should run the [site update script](#) immediately.
> - module_a (1000)
> - module_b (1000)

Image

Image

~~If people are concerned that this could be unwieldy with a large number of modules, one approach could be to wrap the list in a `details` element if the count is >= a number (e.g. 10).  Here it is with just the two I have, but you should get the idea.~~

Describe alternatives you've considered. Have you seen this work in a better way in another CMS?
The suggestions include hacking the core file. There was a suggestion of a script to run in Devel but for many people it returned blank.

Bee does list modules requiring updates, but not all users would be confident installing a command line tool and I don't think this does list if the problem exists where nothing runs. This is the same as the core `update.php` which shows the ones that should run and doesn't always have the one triggering the status report.

Image

Backdrop now includes a list on the status report of any modules that require an update to the database schema to run.

yorkshire-pudding:
Fixes https://github.com/backdrop/backdrop-issues/issues/7010


quicksketch:
Merged into 1.x for 1.33.0. Thank you @yorkshire-pudding!

PR #5161, opened 2025-11-18
Use label() instead of title property

### LINKED ISSUE #7012: Use label() method in node_page_view
## Description of the bug

`node_page_view` uses `$node->title` instead of `$node->label()` to call `backdrop_set_title`. In modules that use non-standard translation, such as the `entity_translation` and `title` modules that define a `label callback` the issue is that node titles get cached and end up setting the title to a cached value rather than 

## Steps To Reproduce

This is pretty annoying to do.

To reproduce the behavior:
1. Turn on entity_translation
2. Turn on title module (which allows properties such as titles to be translated)
3. Go to the node type you want to enable translation for
4. Enable the title module
5. Create a node in your default language
6. Create a node in another language
7. Make sure caching is turned on
8. Go to view the node in one language and make sure you get a cache hit
9. Go to view the node in a different language
10. The browser tab title will show in the original language, not the current language.

## Actual behavior

The tab title is in the wrong language

## Expected behavior

The tab title should be in the current language.

## Additional information

Using `$node->label()` instead of `$node->title` fixes the problem. Patch incoming.

bennybobw:
For node_page_view and node_preview we want to use the label() property instead of the title property since the title can be cached in the wrong language in some cases where we translate fields instead of the whole node.

Fixes https://github.com/backdrop/backdrop-issues/issues/7012


herbdool:
@bennybobw there are some failures that appear related.

PR #5163, opened 2025-11-19
Add a langcode() method to entities

### LINKED ISSUE #185: [BC] Add an $entity->language() method
I'm going to exclude the patches in https://drupal.org/node/1495648 which introduced the entity_language() function to D7. It will get the language for any entity.

The Drupal 8 version probably added the language method to the entity object, but it appears to be all mixed up in https://drupal.org/node/1277776.  We'll need to sort through that and decide how much we want to include in Backdrop.

Assigning to @quicksketch for review.

bennybobw:
langcode() is a new method that basically should replace the entity_language method that existed in Drupal 7. This was largely replaced with just the use of the langcode property in Backdrop but this creates problems with entity_translation because that provides field-based translation rather than whole node translation. We need a way to set a handler to return the langcode rather than just grabbing that the node row in the database.

Fixes 185

Fixes https://github.com/backdrop/backdrop-issues/issues/185

### DIFF
--- a/core/modules/entity/entity.class.inc
+++ b/core/modules/entity/entity.class.inc
@@ -58,7 +58,7 @@ interface EntityInterface {
   /**
    * Returns the langcode for the entity.
    *
-   * @return
+   * @return string
    *   The langcode for the entity or NULL as a default.
    *
    */
--- a/core/modules/entity/entity.class.inc
+++ b/core/modules/entity/entity.class.inc
@@ -266,6 +266,7 @@ abstract class Entity extends stdClass implements EntityInterface {
     foreach ($values as $key => $value) {
       $this->$key = $value;
     }
+    $this->langcode();
   }
 
   /**
--- a/core/modules/entity/entity.class.inc
+++ b/core/modules/entity/entity.class.inc
@@ -56,6 +56,15 @@ interface EntityInterface {
   public function bundle();
 
   /**
+   * Returns the langcode for the entity.
+   *
+   * @return
+   *   The langcode for the entity or NULL as a default.
+   *
+   */
+  public function langcode();
+
+  /**
    * Returns the label of the entity.
    *
    * @return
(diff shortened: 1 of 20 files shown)


core/modules/node/node.pages.inc
```diff
@@ -198,7 +198,7 @@ function node_form($form, &$form_state, Node $node) {
     $form['langcode'] = array(
       '#type' => 'select',
       '#title' => t('Language'),
-      '#default_value' => (isset($node->langcode) ? $node->langcode : ''),
+      '#default_value' => $node->langcode(),
```
herbdool, line 201: This could return NULL, whereas previously it would always be a string.

bennybobw: @herbdool  
That's definitely true but it's within the context of the language select. I think `NULL` and `''` are equivalent in that case if `''` doesn't appear in the select options right? The select options are coming from `language_list()` which doesn't have an empty string option as far as I can tell.

Also looking at the comments the `langcode()` method says this:

> Entity types that do not support translation return NULL as a default.

And in the context of this language dropdown the code is checking to see if languages are enabled for the node type first before building the dropdown, so I think it's ok? 

Is there a specific concern/bug you know about having that default value be `NULL` instead of `''` or are you just flagging that this is changing what the default value was before?

core/modules/translation/views/views_handler_field_node_link_translate.inc
```diff
@@ -14,6 +14,7 @@ class views_handler_field_node_link_translate extends views_handler_field_node_l
     // ensure user has access to edit this node.
     $node = $this->get_value($values);
     $node->status = 1; // unpublished nodes ignore access control
+    $node->langcode();
     if (empty($node->langcode) || $node->langcode == LANGUAGE_NONE || !translation_supported_type($node->type) || !node_access('view', $node) || !user_access('translate content')) {
```
quicksketch, line 18: This seems odd. Does `langcode()` need to be called before checking `$node->langcode`? Or should `$node->langcode()` just be used directly in the checks below?

core/modules/entity/entity.class.inc
```diff
@@ -55,6 +55,15 @@ interface EntityInterface {
    */
   public function bundle();
 
+  /**
+   * Returns the langcode for the entity.
+   *
+   * @return
```
quicksketch:
```suggestion
   * @return string
```

bennybobw:
We discussed this in the developer meeting. I'm going to add a call to `langcode()` from the entity constructor and also add a test to test the method.

PR #5168, opened 2025-11-29
Issue #7031: coding standards core/includes/b*

### LINKED ISSUE #7031: Coding standards: core/includes/*
Directory `core/includes`:

- [x] core/includes/a*: https://github.com/backdrop/backdrop/pull/5167
- [x] core/includes/b*: https://github.com/backdrop/backdrop/pull/5168
- [x] core/includes/c*: https://github.com/backdrop/backdrop/pull/5170
- [x] core/includes/database/*: https://github.com/backdrop/backdrop/pull/4915
- [ ] core/includes/d*: https://github.com/backdrop/backdrop/pull/5173
- [ ] core/includes/e*: https://github.com/backdrop/backdrop/pull/5174
- [ ] core/includes/f*: https://github.com/backdrop/backdrop/pull/5177
- [ ] core/includes/[g-i]*: https://github.com/backdrop/backdrop/pull/5181
- [ ] core/includes/[l-m]*
- [ ] core/includes/p*
- [ ] core/includes/[s-t]*
- [ ] core/includes/transliteration*
- [ ] core/includes/u*

herbdool:
Fixes https://github.com/backdrop/backdrop-issues/issues/7031, child of https://github.com/backdrop/backdrop-issues/issues/6004

Replaces https://github.com/backdrop/backdrop/pull/4370


indigoxela:
Locally phpcs still shows a little bit of nagging:
```
FILE: core/includes/ajax.inc
----------------------------------------------------------------------
FOUND 1 ERROR AFFECTING 1 LINE
----------------------------------------------------------------------
 717 | ERROR | [x] Parameter comment must end with a full stop
----------------------------------------------------------------------
PHPCBF CAN FIX THE 1 MARKED SNIFF VIOLATIONS AUTOMATICALLY
----------------------------------------------------------------------
```
```
FILE: core/includes/bootstrap.inc
-------------------------------------------------------------------------
FOUND 1 ERROR AND 2 WARNINGS AFFECTING 3 LINES
-------------------------------------------------------------------------
 1200 | WARNING | [ ] Function variable_initialize() has been deprecated
 2545 | ERROR   | [x] Parameter comment must end with a full stop
 3372 | WARNING | [ ] Function variable_initialize() has been deprecated
-------------------------------------------------------------------------
PHPCBF CAN FIX THE 1 MARKED SNIFF VIOLATIONS AUTOMATICALLY
-------------------------------------------------------------------------
```
I wonder if we should suppress nagging for variable_initialize() with `// phpcs:ignore` - but maybe it's not worth it?

core/includes/batch.inc
```diff
@@ -16,11 +16,11 @@
 /**
  * Loads a batch from the database.
  *
- * @param $id
+ * @param int $id
```
indigoxela: Hm. If this is from $_REQUEST, it's probably rather a (numeric) string? Any url params are strings, no?

core/includes/bootstrap.inc
```diff
@@ -1065,15 +1065,15 @@ function backdrop_settings_initialize() {
...
- * @param $name
+ * @param string $name
  *   The name of the item for which the filename is requested.
- * @param $filename
+ * @param string|null $filename
  *   The filename of the item if it is to be set explicitly rather
  *   than by consulting the database.
  *
- * @return
+ * @return string
```
indigoxela: This doesn't seem correct. First of all the comment states, it might be NULL.

And actually the return code proves that: `return isset($files[$type][$name]) ? $files[$type][$name] : NULL;`

So, this should be `string|null`. But wait, there's more...

`$files[$type][$name] = FALSE;` If the file doesn't exist, this could  be FALSE? `string|null|false` hm... For profiles at least...

core/includes/bootstrap.inc
```diff
@@ -1640,11 +1643,11 @@ function backdrop_add_http_header($name, $value, $append = FALSE) {
 /**
  * Gets the HTTP response headers for the current page.
  *
- * @param $name
+ * @param string|null $name
  *   An HTTP header name. If omitted, all headers are returned as name/value
  *   pairs. If an array value is FALSE, the header has been unset.
  *
- * @return
+ * @return string|false|null
```
indigoxela: Seems like it could also be an array - if $name is omitted: `return $headers;`
So the comment below is a bit misleading.

core/includes/bootstrap.inc
```diff
@@ -1801,8 +1807,11 @@ function backdrop_set_length_headers($content_length) {
  * If the request is conditional (using If-Modified-Since and If-None-Match),
  * and the conditions match those currently in the cache, a 304 Not Modified
  * response is sent.
+ *
+ * @param object $cache
+ *   A cache entry object to be used in the response.
  */
-function backdrop_serve_page_from_cache(stdClass $cache) {
+function backdrop_serve_page_from_cache(object $cache) {
```
indigoxela: I'm not sure if `stdClass` isn't better here. Why switching to `object`?

core/includes/bootstrap.inc
```diff
@@ -1919,20 +1933,31 @@ function backdrop_serve_page_from_cache(stdClass $cache) {
...
  *
- * @param $obj
+ * @param object $obj
  *   The object to which the elements are appended.
- * @param $field
+ * @param string $field
  *   The attribute of $obj whose value should be unserialized.
  */
-function backdrop_unpack($obj, $field = 'data') {
+function backdrop_unpack(object $obj, $field = 'data') {
```
indigoxela: I belief, we should go with stdClass here as well. Omit the type hint and adapt the doc block above.

core/includes/bootstrap.inc
```diff
@@ -2447,7 +2478,7 @@ function watchdog($type, $message, $variables = array(), $severity = WATCHDOG_NO
  * backdrop_set_message(t('An error occurred and processing did not complete.'), 'error');
  * @endcode
  *
- * @param string $message
+ * @param string|null $message
```
indigoxela: Weird. Seems like this could also be int?

Code below: `if ($message || $message === '0' || $message === 0) {`

core/includes/bootstrap.inc
```diff
@@ -3414,15 +3459,15 @@ function _backdrop_bootstrap_sanitize_request() {
 /**
  * Sanitizes unsafe keys from request input.
  *
- * @param mixed $input
+ * @param array $input
  *   The input data array from a request to be sanitized.
- * @param string[] $allowlist
```
indigoxela, line 3419: Isn't `string[]` more accurate in this case? Or did you find any occasions of array values not being strings?

core/includes/bootstrap.inc
```diff
@@ -3414,15 +3459,15 @@ function _backdrop_bootstrap_sanitize_request() {
...
+ * @param array $allowlist
  *   Allowed list of input keys that are considered acceptable.
  *
- * @return string[]
+ * @return array
```
indigoxela: Similar question as above.

core/includes/bootstrap.inc
```diff
@@ -3652,17 +3707,17 @@ function language_multilingual() {
...
+ * @param bool $option_list
  *   (optional) Whether to return a list of options instead of objects. This
  *   list is suitable for use in a select list #options array. This list is
  *   ordered by weight and then language name.
  * @param bool $native_names
  *   (optional) Whether the option_list labels should use the native name
  *   instead of the English language name. Only applies if $option_list is TRUE.
  *
- * @return (stdClass|string)[]
+ * @return (object|string)[]
```
indigoxela: I asked this before: why switching from stdClass to object here?

core/includes/bootstrap.inc
```diff
@@ -3747,10 +3802,10 @@ function language_list($only_enabled = FALSE, $option_list = FALSE, $native_name
 /**
  * Loads a language object from the database.
  *
- * @param $langcode
+ * @param string $langcode
  *   The language code.
  *
- * @return stdClass
+ * @return object|false
```
indigoxela: Same as above: what's wrong with `stdClass`? Sure, adding false makes sense.

core/includes/bootstrap.inc
```diff
@@ -3782,7 +3837,7 @@ function language_name($langcode) {
 /**
  * Returns the default language used on the site.
  *
- * @return
+ * @return object
```
indigoxela: Wouldn't `stdClass` be more accurate?

core/includes/bootstrap.inc
```diff
@@ -4002,10 +4060,9 @@ function ip_address() {
...
  *
- *
- * @param $table
+ * @param string|null $table
  *   The name of the table. If not given, the schema of all tables is returned.
- * @param $rebuild
+ * @param bool $rebuild
  *   If true, the schema will be rebuilt instead of retrieved from the cache.
  *
  * @return array
```
indigoxela: Apparently this can return false.

core/includes/bootstrap.inc
```diff
@@ -1490,7 +1493,7 @@ function backup_get_backup_directory() {
  *   FALSE, only cache entries within the maximum cache lifetime will be
  *   returned.
  *
- * @return stdClass|NULL
+ * @return object|null
```
indigoxela: Isn't `stdClass` more accurate here? (Lowercase null is fine.)

core/includes/batch.inc
herbdool: In this case we can probably assume it's a string.

core/includes/bootstrap.inc
herbdool: I didn't see any uses of this in core so I'm not sure what kind of object is typical. So broadest is `object` I think.

core/includes/bootstrap.inc
herbdool: I don't recall why I switched that.

core/includes/bootstrap.inc
herbdool: I had switched it because I thought that `object` was more correct for a hint (e.g. https://phpstan.org/writing-php-code/phpdoc-types) but I see that phpcs doesn't complain so...

core/includes/bootstrap.inc
indigoxela: > I didn't see any uses of this in core

It's only used in the Drupal bc layer and deprecated. That's why I'd suggest to *not* add the type hint here. In other words, don't change the function, only the doc block. Or eventually don't change anything here?

herbdool: `backdrop_unpack()` itself is not deprecated. Not sure if it should be - someone might be using it in their own module.

indigoxela: > someone might be using it in their own module.

Good point. And because of that I'd really recommend to *not* add the type hint to the param.

Otherwise a PHP warning (Warning: Attempt to read property ... on ...) turns into a fatal TypeError. We're only trying to adhere to our own coding standards here, not changing functionality. :wink:

core/includes/bootstrap.inc
```diff
@@ -1530,10 +1534,10 @@ function backdrop_page_get_cache($check_only = FALSE, $return_stale = FALSE) {
 /**
  * Determines the cacheability of the current page.
  *
- * @param $allow_caching
+ * @param bool|null $allow_caching
  *   Set to FALSE if you want to prevent this page from being cached.
```
indigoxela, line 1538: It can be FALSE as well - as the comment states. And that's how it's actually used in core.

indigoxela: ah, garbage, sorry for the noise... 🤦

core/includes/bootstrap.inc
```diff
@@ -2504,14 +2537,16 @@ function backdrop_set_message($message = NULL, $type = 'status', $repeat = TRUE)
...
+ * @param string|null $type
  *   (optional) Limit the messages returned by type. Defaults to NULL, meaning
  *   all types. These values are supported:
+ *   phpcs:disable
  *   - NULL
  *   - 'status'
  *   - 'warning'
  *   - 'error'
  *   - 'info'
+ *   phpcs:enable
```
indigoxela: An alternative could be to add the full stop phpcs requests. Only one at the end of the last line.

I know, that's actually silly (in our standard check). But AFAIK these `phpcs:enable` will show up literally in our documentation, which might be more confusing than a single dot.

core/includes/bootstrap.inc
herbdool: One could say the same about all the other places where we're adding `array` or `stdClass` type declarations to parameters. They turn PHP warnings into fatal TypeError, e.g.:

`Argument #1 ($array) must be of type array, null given`

I just figured `object` would be the broadest type we could use. But I realize, after looking at the Entity class that it extends `stdClass` so using the `stdClass` type is probably good too.

At any rate, I also saw that Drupal 7 core also didn't use `drupal_unpack()` so it's unlikely it's being used anywhere.

core/includes/bootstrap.inc
herbdool: In this case, a dot won't look out of place so I'll try that. (I saw some other instances in other files where there's a list of code snippets where it might not look right). The other silly thing is that the standards emit an error instead of a warning for a missing dot.

indigoxela:
It's a bit unfortunate, that (permission wise) I can't resolve own comments on core PRs. But I think, we're done here. RTBC!

quicksketch:
Merged into 1.x and 1.32.x. Awesome, thank you @herbdool and @indigoxela!

PR #1258, opened 2016-02-15
Issue #1587: Support context for the taxonomy/term/% path

docwilmot:
Fixes https://github.com/backdrop/backdrop-issues/issues/1587

PR #5169, opened 2025-11-29
Issue #1258: Allow a layout to be selected on additional paths

### LINKED ISSUE #1528: Allow a layout to be selected on additional paths.

docwilmot:

Test POC for allowing new paths to generate contexts.


herbdool:
@docwilmot I updated this branch from 1.x.

docwilmot:
Oops, committed some test code for hiding layouts in layout list.

docwilmot:
Closing in favor of https://github.com/backdrop/backdrop/pull/5184

PR #5170, opened 2025-12-01
Issue: #7031: coding standards - core/includes/c*

### LINKED ISSUE #7031: Coding standards: core/includes/*

herbdool:
Fixes https://github.com/backdrop/backdrop-issues/issues/7031, child of https://github.com/backdrop/backdrop-issues/issues/6004

### DIFF
--- a/core/includes/cache.inc
+++ b/core/includes/cache.inc
@@ -70,7 +70,7 @@ function cache_get($cid, $bin = 'cache') {
  *   The cache bin from which data should be retrieved. Valid core values are
  *   "cache", "bootstrap", "field", "filter", "form", "menu", "page", "path",
  *   "update", "views", and "views_data". Bin names may include the prefix of
- *   "cache_", but it is stripped out before execution
+ *   "cache_", but it is stripped out before execution.
  *
  * @return array
  *   An array of the items successfully returned from cache indexed by cid.
@@ -110,7 +110,7 @@ function cache_get_multiple(array &$cids, $bin = 'cache') {
  *   The cache bin in which data should be stored. Valid core values are
  *   "cache", "bootstrap", "field", "filter", "form", "menu", "page", "path",
  *   "update", "views", and "views_data". Bin names may include the prefix of
- *   "cache_", but it is stripped out before execution
+ *   "cache_", but it is stripped out before execution.
  * @param int $expire
  *   (optional) Controls the maximum lifetime of this cache entry. Note that
  *   caches might be subject to clearing at any time, so this setting does not
@@ -259,7 +259,7 @@ interface BackdropCacheInterface {
    * @return array|object|false
    *   The cache or FALSE on failure.
    */
-  function get($cid);
+  public function get($cid);
 
   /**
    * Returns data from the persistent cache when given an array of cache IDs.
@@ -272,7 +272,7 @@ interface BackdropCacheInterface {
    * @return array
    *   An array of the items successfully returned from cache indexed by cid.
    */
-   function getMultiple(array &$cids);
+  public function getMultiple(array &$cids);
 
   /**
    * Stores data in the persistent cache.
@@ -299,15 +299,15 @@ interface BackdropCacheInterface {
    *   - A Unix timestamp: Indicates that the item should be kept at least until
    *     the given time, after which it behaves like CACHE_TEMPORARY.
    */
-  function set($cid, $data, $expire = CACHE_PERMANENT);
+  public function set($cid, $data, $expire = CACHE_PERMANENT);
 
   /**
    * Deletes an item from the cache.
    *
    * @param string $cid
-   *    The cache ID to delete.
+   *   The cache ID to delete.
    */
-  function delete($cid);
+  public function delete($cid);
 
   /**
    * Deletes multiple items from the cache.
@@ -315,7 +315,7 @@ interface BackdropCacheInterface {
    * @param array $cids
    *   An array of $cids to delete.
    */
-  function deleteMultiple(Array $cids);
+  public function deleteMultiple(array $cids);
 
   /**
    * Deletes items from the cache using a wildcard prefix.
@@ -323,17 +323,17 @@ interface BackdropCacheInterface {
    * @param string $prefix
    *   A wildcard prefix.
    */
-  function deletePrefix($prefix);
+  public function deletePrefix($prefix);
 
   /**
    * Flushes all cache items in a bin.
    */
-  function flush();
+  public function flush();
 
   /**
    * Performs garbage collection on a cache bin. Removing expired items.
    */
-  function garbageCollection();
+  public function garbageCollection();
 
   /**
    * Checks if a cache bin is empty.
@@ -344,7 +344,7 @@ interface BackdropCacheInterface {
    * @return bool
    *   TRUE if the cache bin specified is empty.
    */
-  function isEmpty();
+  public function isEmpty();
 }
 
 /**
@@ -366,61 +366,61 @@ class BackdropNullCache implements BackdropCacheInterface {
    * @param string $bin
    *   The cache bin for which the object is created.
    */
-  function __construct($bin) {}
+  public function __construct($bin) {}
 
   /**
    * Implements BackdropCacheInterface::get().
    */
-  function get($cid) {
+  public function get($cid) {
     return FALSE;
   }
 
   /**
    * Implements BackdropCacheInterface::getMultiple().
    */
-  function getMultiple(array &$cids) {
+  public function getMultiple(array &$cids) {
     return array();
   }
 
   /**
    * Implements BackdropCacheInterface::set().
    */
-  function set($cid, $data, $expire = CACHE_PERMANENT) {}
+  public function set($cid, $data, $expire = CACHE_PERMANENT) {}
 
   /**
    * Implements BackdropCacheInterface::delete().
    */
-  function delete($cid) {}
+  public function delete($cid) {}
 
   /**
    * Implements BackdropCacheInterface::deleteMultiple().
    */
-  function deleteMultiple(array $cids) {}
+  public function deleteMultiple(array $cids) {}
 
   /**
    * Implements BackdropCacheInterface::deletePrefix().
    */
-  function deletePrefix($prefix) {}
+  public function deletePrefix($prefix) {}
 
   /**
    * Implements BackdropCacheInterface::flush().
    */
-  function flush() {}
+  public function flush() {}
 
   /**
    * Implements BackdropCacheInterface::expire().
    */
-  function expire() {}
+  public function expire() {}
 
   /**
    * Implements BackdropCacheInterface::garbageCollection().
    */
-  function garbageCollection() {}
+  public function garbageCollection() {}
 
   /**
    * Implements BackdropCacheInterface::isEmpty().
    */
-  function isEmpty() {
+  public function isEmpty() {
     return TRUE;
   }
 }
@@ -428,8 +428,9 @@ class BackdropNullCache implements BackdropCacheInterface {
 /**
  * Defines a default cache implementation.
  *
- * This is Backdrop's default cache implementation. It uses the database to store
- * cached data. Each cache bin corresponds to a database table by the same name.
+ * This is Backdrop's default cache implementation. It uses the database to
+ * store cached data. Each cache bin corresponds to a database table by the same
+ * name.
  */
 class BackdropDatabaseCache implements BackdropCacheInterface {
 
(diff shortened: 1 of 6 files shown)


core/includes/cache.inc
```diff
@@ -52,7 +52,7 @@ function cache($bin = 'cache') {
  *   "update", "views", and "views_data". Bin names may include the prefix of
  *   "cache_", but it is stripped out before execution.
  *
- * @return
+ * @return array|object|false
```
indigoxela: I wonder, when this could be an array, though. 🤔 
As far as I can see, BackdropDatabaseCache::get does a reset before returning.

core/includes/cache.inc
```diff
@@ -256,10 +256,10 @@ interface BackdropCacheInterface {
    * @param string $cid
    *   The cache ID of the data to retrieve.
    *
-   * @return
+   * @return array|object|false
```
indigoxela: Similar question as above - when can it be an array?

core/includes/cache.inc
```diff
@@ -496,10 +501,11 @@ class BackdropDatabaseCache implements BackdropCacheInterface {
...
    * data as appropriate.
    *
-   * @param $cache
-   *   An item loaded from BackdropCacheInterface::get() or BackdropCacheInterface::getMultiple().
+   * @param array|stdClass $cache
+   *   An item loaded from BackdropCacheInterface::get() or
+   *   BackdropCacheInterface::getMultiple().
    *
-   * @return stdClass|FALSE
+   * @return object|false
```
indigoxela: Again this switch from stdClass to object - when could this *not* be stdClass?

core/includes/common.inc
```diff
@@ -219,6 +220,8 @@ function backdrop_set_breadcrumb($breadcrumb = NULL) {
 
 /**
  * Gets the breadcrumb trail for the current page.
+ *
+ * @return array|string[]
```
indigoxela: This looks a bit odd "an array or an array of strings" - is that because it can be an empty array?
TBH, I'm not sure, how this should be written... IDEs or tools for static analysis have their own notations - to add more confusion. Not sure if the current notation is good or bad... 🤷

core/includes/common.inc
```diff
@@ -814,20 +834,20 @@ function backdrop_goto_deprecated($path = '', array $options = array(), $http_re
  * This function ends the request; use it instead of a return in your menu
  * callback.
  *
- * @param $path
+ * @param string $path
```
indigoxela: This can be null (actually unset) - to redirect to the home page. It's used that way in core.

core/includes/common.inc
```diff
@@ -814,20 +834,20 @@ function backdrop_goto_deprecated($path = '', array $options = array(), $http_re
...
  *   (optional) A Backdrop path or a full URL, which will be passed to url() to
  *   compute the redirect for the URL.
- * @param $options
+ * @param array $options
```
indigoxela, line 840: This can also be null, as it's optional. (?) Actually, it can be unset, but if set it must be an array. Not sure, how doc blocks are supposed to describe that. Do we need to update the function param type hint?

indigoxela:
Note: pausing here to come back with a fresh mind for common.inc :wink:

core/includes/common.inc
indigoxela: Let's just stick with "array".

core/includes/common.inc
```diff
@@ -1673,11 +1701,14 @@ function flood_is_allowed($name, $threshold, $window = 3600, $identifier = NULL)
  * @param string $name
  *   The name of the object being stored. Must be less than 128 characters.
  *
- * @return
+ * @return string|false|null
```
indigoxela: According to PHP docs unserialize() can return anything of bool, int, float, string, array or object.
And it doesn't seem like tempstore_set() limits the type of value in any way... `mixed` ?

core/includes/common.inc
```diff
@@ -5512,23 +5672,23 @@ function backdrop_process_states(&$elements) {
...
+ * @return bool
  *   TRUE if the library was successfully added; FALSE if the library or one of
  *   its dependencies could not be added.
  *
  * @see backdrop_get_library()
  * @see hook_library_info()
  * @see hook_library_info_alter()
  */
-function backdrop_add_library($module, $name, $every_page = NULL) {
+function backdrop_add_library($module, $name, ?bool $every_page = NULL) {
```
indigoxela, line 5681: We don't have to add this param type hint here for standards reasons - but I can't imagine, it does any harm, either.

core/includes/common.inc
```diff
@@ -5565,13 +5725,13 @@ function backdrop_add_library($module, $name, $every_page = NULL) {
...
- * @param $module
+ * @param string $module
  *   The name of a module that registered a library.
- * @param $name
+ * @param string|null $name
  *   (optional) The name of a registered library to retrieve. By default, all
  *   libraries registered by $module are returned.
  *
- * @return
+ * @return array
```
indigoxela: This can return FALSE if a specific one's requested and the library doesn't exist / isn't found.

core/includes/common.inc
```diff
@@ -5905,6 +6066,11 @@ function backdrop_json_encode($var, $pretty_print = FALSE) {
 /**
  * Converts an HTML-safe JSON string into its PHP equivalent.
  *
+ * @param string $var
+ *   The JSON string.
+ *
+ * @return array
```
indigoxela: According to php docs for json_decode(), there's more return types:
https://www.php.net/manual/en/function.json-decode.php

It's probably rather "mixed"? At least, it can be null in case of errors.

core/includes/common.inc
```diff
@@ -6329,7 +6504,7 @@ function backdrop_page_create_cache($page_content) {
  * @param stdClass $cache
  *   A cache entry object to be written to the cache backend.
  */
-function backdrop_page_save_cache($cache) {
+function backdrop_page_save_cache(stdClass $cache) {
```
indigoxela, line 6497: This type hint's not required by our standards.
On the other hand this is more like an internal function and we can assume, it's always stdClass, I guess.

core/includes/common.inc
```diff
@@ -7022,10 +7216,10 @@ function backdrop_render_children(&$element, $children_keys = NULL) {
  * element is shown with show() before rendering, so it will always be rendered
  * even if hide() had been previously used on it.
  *
- * @param $element
+ * @param array|string|null $element
```
indigoxela: OK, inappropriate use... but if the param can be string|null and it's piped through if it's not an array, wouldn't the return type also contain null, then?

Can it be null, anyway? String seems plausible, but null ... 🤔

core/includes/common.inc
herbdool: Let's drop the `string[]` to keep it simple.

core/includes/common.inc
herbdool: According to https://docs.backdropcms.org/php-standards#type-hinting, it's currently needed for arrays and objects:

> Whenever possible, function and method parameters should specify their data types using [type hinting](https://www.php.net/manual/en/language.types.declarations.php). Because type hinting cannot be used with scalar types such as int or string (to allow backwards-compatibility with PHP versions < 7), this only comes into play when arguments are arrays or objects.

The text should be updated since we no longer support < 7.

core/includes/common.inc
herbdool: Just refreshing my memory. Actually there are some templates such as `comment.tpl.php` which has `render($title_prefix);` and it's telling me that `$title_prefix` is `null`. Looks like it never gets populated. So sloppy programming. But even that template says that it's supposed to be an array. So maybe we make it just `array|string` and in the future can fix any offending template variables instead.

core/includes/common.inc
indigoxela: "Whenever possible" is quite flexible to interpret.

Yes, documentation's a moving target. :wink:

core/includes/cache.inc
herbdool: Yup, likely never an array.

core/includes/common.inc
```diff
@@ -7245,19 +7429,22 @@ function backdrop_render_collect_attached($elements, $return = FALSE) {
...
  *
- * @return
+ * @return array
  *   A renderable array with the following keys and values:
  *   - #query: The passed-in $query.
  *   - #pre_render: $function with a _pre_render suffix.
  *   - #cache: An associative array prepared for backdrop_render_cache_set().
  */
-function backdrop_render_cache_by_query($query, $function, $expire = CACHE_TEMPORARY, $granularity = NULL) {
+function backdrop_render_cache_by_query(SelectQuery $query, $function, $expire = CACHE_TEMPORARY, $granularity = NULL) {
```
indigoxela, line 7441: Is this type hint necessary?
Currently core only uses this function once in the Drupal compat layer.

core/includes/common.inc
```diff
@@ -7494,20 +7683,35 @@ function backdrop_sort(array &$array, array $keys = array('weight'), $dir = SORT
...
  * Gets properties of a structured array element (keys beginning with '#').
+ *
+ * @param array $element
+ *   The element array whose properties are to be identified.
+ *
+ * @return array
  */
-function element_properties($element) {
+function element_properties(array $element) {
   return array_filter(array_keys((array) $element), 'element_property');
```
indigoxela: The mix of type hint and casting (one line after) makes me wonder, how to deal with this ancient code.

Sure, it's *supposed* to be an array, but apparently it isn't or wasn't always. (?)
For sure the casting could go, if we type hint the param. Something other than array wouldn't even get there.
But it's really old code, almost 20 years it seems... 🤔

core/includes/common.inc
```diff
@@ -7677,7 +7884,7 @@ function element_set_attributes(array &$element, array $map) {
  * @see backdrop_array_set_nested_value()
  * @see backdrop_array_unset_nested_value()
  */
-function &backdrop_array_get_nested_value(array &$array, array $parents, &$key_exists = NULL) {
+function &backdrop_array_get_nested_value(array &$array, array $parents, ?bool &$key_exists = NULL) {
```
indigoxela: I don't think, this bool type hint is necessary.

core/includes/common.inc
```diff
@@ -7851,19 +8058,19 @@ function backdrop_array_set_nested_value(array &$array, array $parents, $value,
...
- * @param $key_existed
+ * @param bool|null $key_existed
  *   (optional) If given, an already defined variable that is altered by
  *   reference.
  *
  * @see backdrop_array_set_nested_value()
  * @see backdrop_array_get_nested_value()
  */
-function backdrop_array_unset_nested_value(array &$array, array $parents, &$key_existed = NULL) {
+function backdrop_array_unset_nested_value(array &$array, array $parents, ?bool &$key_existed = NULL) {
```
indigoxela: I don't think, this bool type hint is necessary.

indigoxela:
Apparently I survived to review common.inc :tada: 

There are two more phpcs nags, but both are silly ones. I'd be OK with ignoring them.

Besides that, just some small comments, then this is ready to go. :smile:

core/includes/common.inc
herbdool: I would say yes since the aim is to add type declarations for objects to functions, unless it breaks BC.

core/includes/common.inc
herbdool: I don't think it hurts either, but doesn't matter to me if we remove it.

core/includes/common.inc
indigoxela: Let's say, *your* aim is to add type hints for objects. I'm very cautious with type hints. 😉
I don't think, it does any harm here, so ... do as you like.

herbdool: The way I read https://docs.backdropcms.org/php-standards#type-hinting is that we *should* add object type hints wherever possible.

quicksketch:
Merged into 1.x and 1.32.x. Thank you @herbdool and @indigoxela!

PR #5172, opened 2025-12-12
Issue #6987: Custom Text format with multi value field saves empty items

### LINKED ISSUE #6987: Custom Text format with multi valued field creates empty field each time after saving node
## Description of the bug

In the latest (1.32.1) backdrop version also in the old version (1.31.0) I am facing this issue. Every time I save a node it adds one more empty body field. The body field is configured as to have unlimited values and I use custom text format for this and used only that text format on the body field, if I use the Basic  text format then this error doesn't arise.

## Steps To Reproduce

To reproduce the behavior:
1. Create a sandbox site and create a new content type or use the basic Post content type.
2. Create a new text format and use the Ckeditor5 for it. 
3. Just add a body field into the content type and make it as unlimited (multivalue) and use the new custom text format for it (disable other text formats).

Image

Image

4. Then just make a new content of that content type and hit save and checked back and edit the content now we have one more empty body field into the node and now every time we save the node we will have a new empty body field.

## Additional information

- Backdrop CMS version: 1.32.1
- Web server and its version: Apache/2.4.65 (Debian)
- PHP version: 8.1.33
- Database sever (MySQL or MariaDB?) and its version:MySQL, MariaDB, or equivalent version 10.7.8-MariaDB-1:10.7.8+maria~ubu2004
- Operating System and its version: Linux
- Browser(s) and their versions: Chrome

indigoxela:
Fixes https://github.com/backdrop/backdrop-issues/issues/6987

### DIFF
--- a/core/modules/field/field.api.php
+++ b/core/modules/field/field.api.php
@@ -704,7 +704,7 @@ function hook_field_prepare_translation($entity_type, $entity, $field, $instance
  *   TRUE if $field's type considers $item not to contain any data;
  *   FALSE otherwise.
  */
-function hook_field_is_empty($item, $field) {
+function hook_field_is_empty($item, $field, $instance) {
   if (empty($item['value']) && (string) $item['value'] !== '0') {
     return TRUE;
   }
--- a/core/modules/field/field.api.php
+++ b/core/modules/field/field.api.php
@@ -703,6 +703,8 @@ function hook_field_prepare_translation($entity_type, $entity, $field, $instance
  * @return bool
  *   TRUE if $field's type considers $item not to contain any data;
  *   FALSE otherwise.
+ *
+ * @since 1.33.0 Added parameter $instance.
  */
 function hook_field_is_empty($item, $field, $instance) {
   if (empty($item['value']) && (string) $item['value'] !== '0') {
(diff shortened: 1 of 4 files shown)


yorkshire-pudding (approved):
@indigoxela - thank you. This code looks really good. The comments help to explain it and the code all makes sense.

core/modules/field/field.api.php
```diff
@@ -693,16 +693,18 @@ function hook_field_prepare_translation($entity_type, $entity, $field, $instance
...
  * Define what constitutes an empty item for a field type.
  *
- * @param $item
+ * @param array $item
  *   An item that may or may not be empty.
- * @param $field
+ * @param array $field
  *   The field to which $item belongs.
+ * @param array $instance
+ *   Field API instance definition.
```
indigoxela, line 701: Do we need an `@since` for the additional param?

yorkshire-pudding: > Do we need an `@since` for the additional param?

Good point. I think we probably do - https://docs.backdropcms.org/doc-standards#since

indigoxela: Done.

quicksketch:
Merged into 1.x for 1.33.0. Thank you @indigoxela!

PR #5173, opened 2025-12-12
Issue #7031: phpcs for includes/d*

herbdool:
Relates to https://github.com/backdrop/backdrop-issues/issues/7031, child of https://github.com/backdrop/backdrop-issues/issues/6004


core/includes/date.class.inc
```diff
@@ -207,8 +208,8 @@ class BackdropDateTime extends DateTime {
    * In order to set a timezone for a datetime that doesn't have such
    * granularity, merge() it with one that does.
    *
-   * @param DateTimeZone $timezone
-   *   A timezone object.
+   * @param mixed $timezone
+   *   A timezone object. As of PHP 8.0: DateTimeZone.
```
indigoxela: Why as of 8.0? Wasn't the param also DateTimeZone before that?

core/includes/date.inc
```diff
@@ -608,7 +608,7 @@ function date_timezone_abbr($refresh = FALSE) {
  * Translates formatted date results, unlike PHP function date_format().
  * Should only be used for display, not input, because it can't be parsed.
  *
- * @param object $date
+ * @param BackdropDateTime|string|null $date
```
indigoxela, line 611: I'm not sure, if this could be string|null.
Obviously, there's some check in the function to prevent problems when it's empty (which might happen when an error occurs when creating the date object, old code, though)... 🤔
But unlike format_date(), this wouldn't accept a timestamp. Or am I missing something?

core/includes/date.inc
```diff
@@ -710,7 +724,7 @@ function date_format_date($date, $type = 'medium', $format = '', $langcode = NUL
 /**
  * Formats a time interval with granularity, including past and future context.
  *
- * @param object $date
+ * @param BackdropDateTime|null $date
```
indigoxela, line 727: Similar to above, can it be null, anyway? The code from 2015 checks for empty, yes. But maybe it was just defensive coding...

core/includes/date.inc
```diff
@@ -720,16 +734,19 @@ function date_format_date($date, $type = 'medium', $format = '', $langcode = NUL
  *
  * @see format_interval()
  */
-function date_format_interval($date, $granularity = 2, $display_ago = TRUE) {
+function date_format_interval(?BackdropDateTime $date, $granularity = 2, $display_ago = TRUE) {
```
indigoxela, line 737: Given, that's it's not really clear, why these empty() checks are done, better remove that type hint (for later investigation).

core/includes/date.inc
```diff
@@ -1097,7 +1114,7 @@ function date_pad($value, $size = 2) {
  * @return bool
  *   TRUE if the granularity contains a time portion, FALSE otherwise.
  */
-function date_has_time($granularity) {
+function date_has_time(array $granularity) {
   if (!is_array($granularity)) {
```
indigoxela: This may be as well (overly) defensive coding, but the combination of type hint and later on is_array() doesn't make sense to me. Either we type hint the param, or it can be something else than array. (Same code age as the unclear lines further up.)

core/includes/date.inc
```diff
@@ -1114,7 +1131,7 @@ function date_has_time($granularity) {
  * @return bool
  *   TRUE if the granularity contains a date portion, FALSE otherwise.
  */
-function date_has_date($granularity) {
+function date_has_date(array $granularity) {
   if (!is_array($granularity)) {
```
indigoxela: Same problem as above. This is_array() won't be useful anymore, when using the type hint.

core/includes/date.inc
```diff
@@ -1332,12 +1348,12 @@ function date_format_order($format) {
...
  *
  * @param array $granularity
- *   An array like ('year', 'month', 'day', 'hour', 'minute', 'second');
+ *   An array like ['year', 'month', 'day', 'hour', 'minute', 'second'].
  *
  * @return array
  *   A reduced set of granularity elements.
  */
-function date_nongranularity($granularity) {
+function date_nongranularity(array $granularity) {
```
indigoxela, line 1350: Further down's some casting, which loses its value when we use type hinting here.

core/includes/date.inc
```diff
@@ -1438,7 +1454,7 @@ function date_range_valid($string) {
  *
  * @param string $string
  *   A min and max year string like '-3:+1'.
- * @param object $date
+ * @param BackdropDateTime|string|null $date
```
indigoxela: The code further down would also handle "string" gracefully, but when could it ever be one?

```
$value_year = is_object($date) ? $date->format('Y') : '';
```

core/includes/date.inc
```diff
@@ -1611,17 +1628,17 @@ function date_increment_round(&$date, $increment) {
     $second = intval(round(intval(date_format($date, 's')) / $increment) * $increment);
     $minute = intval(date_format($date, 'i'));
     if ($second == 60) {
-      $minute += 1;
+      ++$minute;
```
indigoxela, line 1625: I'm curious: what's the benefit of this notation change? (Also below)

core/includes/diff.inc
```diff
@@ -688,9 +786,10 @@ class Diff {
    *
    * This is mostly for diagnostic purposed.
    *
-   * @return int The length of the LCS.
+   * @return int
+   *   The length of the LCS.
```
indigoxela: Maybe sneak in some "longest common subsequence" here for clarity?

core/includes/diff.inc
```diff
@@ -982,11 +1115,13 @@ abstract class DiffFormatter {
...
-   * @param $diff Diff A Diff object.
+   * @param Diff $diff
+   *   A Diff object.
    *
-   * @return string The formatted output.
+   * @return string
+   *   The formatted output.
    */
-  public function format($diff) {
+  public function format(Diff $diff) {
```
indigoxela, line 1124: In this function it's clear that it can't be anything else than "Diff", so this type hint seems OK. 😉

core/includes/diff.inc
```diff
@@ -1080,6 +1220,9 @@ abstract class DiffFormatter {
     $this->endBlock();
   }
 
+  /**
+   * @return void
```
indigoxela, line 1224: As this is just a wrapper around ob_start(), we inherit that function's return, I guess.
https://www.php.net/manual/en/function.ob-start.php
Or is it null then?

core/includes/diff.inc
```diff
@@ -1113,49 +1260,63 @@ abstract class DiffFormatter {
...
+  /**
+   * @param string $header
+   *   Header.
+   */
   protected function startBlock($header) {
     echo $header . "\n";
   }
 
+  /**
+   * @return void
```
indigoxela: What's the benefit of this "void"? Not sure, what methods in abstract classes should declare here, anyway. 🤔

core/includes/diff.inc
```diff
@@ -1229,53 +1440,79 @@ class BackdropDiffFormatter extends DiffFormatter {
...
       $this->rows[] = array_merge($this->emptyLine(), $this->addedLine(check_plain($line)));
     }
   }
 
-  protected function deleted($lines) {
+  /**
+   * @param array $lines
+   *   Lines.
+   *
+   * @return void
```
indigoxela: Do we need this `@return` here, anyway? Does it provide any value?

core/includes/diff.inc
```diff
@@ -1229,53 +1440,79 @@ class BackdropDiffFormatter extends DiffFormatter {
...
       $this->rows[] = array_merge($this->deletedLine(check_plain($line)), $this->emptyLine());
     }
   }
 
-  protected function context($lines) {
+  /**
+   * @param array $lines
+   *   Lines.
+   *
+   * @return void
```
indigoxela: Same question as above: what's the benefit of this "void".

indigoxela:
There's still some nagging:

```
FILE: core/includes/database/mysql/database.inc
--------------------------------------------------------------------------
FOUND 0 ERRORS AND 2 WARNINGS AFFECTING 2 LINES
--------------------------------------------------------------------------
 70 | WARNING | Line exceeds 80 characters; contains 81 characters
 79 | WARNING | Line exceeds 80 characters; contains 81 characters
--------------------------------------------------------------------------


FILE: core/includes/drupal.inc
-------------------------------------------------------------------------------------------------------------------------------------------
FOUND 1 ERROR AND 3 WARNINGS AFFECTING 4 LINES
-------------------------------------------------------------------------------------------------------------------------------------------
  479 | WARNING | [ ] Function drupal_static() has been deprecated; use backdrop_static() instead.
  519 | WARNING | [ ] Function drupal_register_shutdown_function() has been deprecated; use backdrop_register_shutdown_function() instead.
 1393 | WARNING | [ ] Function drupal_array_get_nested_value() has been deprecated; use backdrop_array_get_nested_value() instead.
 2228 | ERROR   | [x] You must use "/**" style comments for a function comment
```
I know, mostly silly ones. but some more `// phpcs:ignore` should help.
Are you aware, that you can also add this ignore at the end of the same line @herbdool 

Cspell allows similar syntax with `// cspell:disable-line` - for the current line.

core/includes/drupal.inc
```diff
@@ -2226,8 +2237,10 @@ function drupal_path_alias_whitelist_rebuild($source = NULL) {
  */
 function drupal_path_alias_allowlist_rebuild($source = NULL) {
   watchdog_deprecated_function('drupal', __FUNCTION__);
+  // phpcs:disable
   // cspell:disable-next-line
   return drupal_path_alias_whitelist_rebuild();
```
indigoxela: Another option (to avoid using disable/enable would be, to add it like that:

`return drupal_path_alias_whitelist_rebuild();// phpcs:ignore`

core/includes/date.inc
herbdool: Phpcs requires it.

core/includes/diff.inc
herbdool: It's not actually returning anything, so it would be `void`, even though `ob_start()` returns `bool`. I wouldn't normally add `@return void` (though it does make it clearer) but needed a method comment.

core/includes/diff.inc
herbdool: Not sure the purpose of the method. Maybe I'll leave off the return here.

core/includes/date.class.inc
herbdool: To make sure the doc type `@param DateTimeZone $timezone` matches with the type hint:

`public function setTimezone(DateTimeZone $timezone, $force = FALSE) {`

I was getting in PHP < 8: 

> Warning: Declaration of BackdropDateTime::setTimezone(DateTimeZone $timezone, $force = false) should be compatible with DateTime::setTimezone($timezone)

I guess this is the one case where we can't keep the `DateTimeZone` type hint but could could keep it in the doc.

core/includes/date.inc
herbdool: `_date_popup_process_date_part()` was passing an empty string to `date_format_date()`. We can get it to pass `null` and accept null at least and put in some defensive coding.

core/includes/date.inc
herbdool: `_date_popup_process_date_part()` was "misbehaving" by setting `$date` to be either an empty string or object. All other places in core were first checking if date was an object. And it looks like contrib is okay too. The simple improvement is to ensure `$date` is null or `BackdropDateTime`.

core/includes/date.inc
herbdool: There's nothing in core or contrib that depends on `date_format_interval()` accepting null so we could probably even make it `@param BackdropDateTime $date`.

core/includes/date.inc
herbdool: We could probably also make it `function date_format_interval(BackdropDateTime $date, $granularity = 2, $display_ago = TRUE) {` and remove the defensive check.

core/includes/date.inc
indigoxela: You're right! That's Squiz.Operators.IncrementDecrementUsage. Sorry for the noise, then. :+1:

core/includes/date.inc
indigoxela: Hm, that type hint (even more removing defensive coding) seems a bit dangerous in a coding standards PR... :thinking:

core/includes/date.class.inc
indigoxela: Sure, we can't use a type hint, but we should be clear about why we can't add it.
Currently phcps nags.
```
220 | WARNING | Type hint "DateTimeZone" missing for $timezone
```
The next one might try to "fix" it and might not look at it with php 7.x. `// phpcs:ignore` - inline on that line.
It's a bit unfortunate, that we can't put it on the previous line (need ReturnTypeWillChange there) and also can't limit to that particular sniff (`// phpcs:ignore Backdrop.Commenting.FunctionComment.TypeHintMissing` is too long for `Backdrop.Files.LineLength.TooLong` sniff).

core/includes/date.inc
```diff
@@ -626,16 +626,16 @@ function date_timezone_abbr($refresh = FALSE) {
  *
  * @see format_date()
  */
-function date_format_date($date, $type = 'medium', $format = '', $langcode = NULL) {
+function date_format_date(?BackdropDateTime $date, $type = 'medium', $format = '', $langcode = NULL) {
```
indigoxela: As it turned out, that even core "misbehaved" here (_date_popup_process_date_part()), can we please remove this type hint entirely?
Otherwise this might turn out to be a nasty site breaker with sloppy contrib or custom code.

Keep in mind: we're supposed to fix *coding standards* not code. And this might go into a bugfix release.

indigoxela:
We're almost done. I've left some more comments. And there's still nagging:

```
FILE: core/includes/database/mysql/database.inc
--------------------------------------------------------------------------
FOUND 0 ERRORS AND 2 WARNINGS AFFECTING 2 LINES
--------------------------------------------------------------------------
 70 | WARNING | Line exceeds 80 characters; contains 81 characters
 79 | WARNING | Line exceeds 80 characters; contains 81 characters
--------------------------------------------------------------------------


FILE: core/includes/date.class.inc
----------------------------------------------------------------------
FOUND 0 ERRORS AND 1 WARNING AFFECTING 1 LINE
----------------------------------------------------------------------
 220 | WARNING | Type hint "DateTimeZone" missing for $timezone
----------------------------------------------------------------------
```
The latter is one of the things with a new comment. :wink:

PR #5175, opened 2025-12-12
Corrected logic which determines if an existing form action has been designated as the primary action.

### LINKED ISSUE #7023: Cannot designate a 'primary' or 'danger' form action unless both are specified
## Description of the bug

`form_process_actions()` in `core/includes/form.inc` designates a form's first action (button) as the 'primary', unless one has already been specified manually. It does this by inspecting each action to see if it has either a 'button-primary' or 'button-danger' class. However, the current logic requires the button to have _both_ classes instead of just one or the other.

## Steps To Reproduce

To reproduce the behavior:
1. Create a form with two actions (buttons).
2. Specify that the second button is the 'primary' via `['#attributes']['class'][] = 'button-primary';` or `['#attributes']['class'][] = 'button-danger';`
3. View the form.

## Actual behavior

The first button is incorrectly displayed as the primary (highlighted with a blue background).

## Expected behavior

The second button should be displayed as the primary (highlighted with a blue or red background).

DevinCarlson:
Fixes backdrop/backdrop-issues#7023

### DIFF
--- a/core/includes/form.inc
+++ b/core/includes/form.inc
@@ -3606,7 +3606,7 @@ function form_process_actions($element, &$form_state) {
       $first_button = $key;
     }
     // Record the primary button if found.
-    if (isset($button['#attributes']['class']) && in_array('button-primary', $button['#attributes']['class'])  && in_array('button-danger', $button['#attributes']['class'])) {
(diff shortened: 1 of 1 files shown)


quicksketch:
Merged into 1.x and 1.32.x. Thank you @DevinCarlson!

PR #5176, opened 2025-12-21
Issue 6976. Fix date views grouped filters and relative dates

### LINKED ISSUE #6976: Exposed group filter for date not working
There are a couple of problems.  I have a date field (single, no start/end) and would like to filter a list in a View to show items with the date in the past, those in the future, and a list of all.  I'd like to do this with a grouped exposed filter so users can toggle between the three options , while still being able to use combinations of other available filters.  

The first issue is that on the Views editing interface, when I select it to be exposed and a grouped filter, I get the option to select a type (relative or date) but then do not see another field where I can enter the date or relative criteria:

Image

I am migrating a site from Drupal 7 and I do remember a similar issue happening before (although it has been 6 years ago so a little fuzzy on details).  I did use one of the provided patches from the issue queue.  Here is the discussion in the Drupal docs:

https://www.drupal.org/project/views/issues/1764902

I don't think any of the patches cleared testing and wouldn't work directly with the current Backdrop code but might be a helpful reference.  

Note that if I do a single filter, without grouping, just selecting to show things in the future, it works fine.  I thought I might use that as a reference and try manually editing the config file for the View after switching it to grouped.  The info for the single file is:

"operator": ">",
                        "value": {
                            "min": null,
                            "max": null,
                            "value": null,
                            "type": "date"
                        },

But trying to set that manually for a grouped item did not get it to work; it always just showed all the values.  Also not sure why they are all null, instead of one indicating "now"?  I tried copying the array values from the Views export from the Drupal site and that did not work either.  So I think there are other problems beyond not being to set the values in the editing interface.  If you refer to the Drupal issue queue, many folks describe the same problems, beyond not just being able to enter values in the front end. 

I'm on the current version of Backdrop, PHP 7.4 (about to upgrade once this Drupal site is shut down), MySQL, Ubuntu 20.04.6, and have tested in Firefox and Chrome.  

Any help or suggestions you can provide would be appreciated.  I have some sort of clunky workarounds I can do if it comes to that but would really like to be able to keep this setup.

argiepiano:
Fixes https://github.com/backdrop/backdrop-issues/issues/6976

Based on the latest [PR](https://www.drupal.org/files/issues/2022-08-14/date-exposed_grouped_filters-1876168-171.patch) for https://www.drupal.org/project/date/issues/1876168

### DIFF
--- a/core/modules/date/views/date_views_filter_handler_simple.inc
+++ b/core/modules/date/views/date_views_filter_handler_simple.inc
@@ -21,6 +21,11 @@ class date_views_filter_handler_simple extends views_handler_filter_date {
   protected $base_table;
 
   /**
+   * @var int
+   */
+  public $element_number = 1;
+
+  /**
    * @var views_plugin_query_default
    */
   var $query;
@@ -72,7 +77,15 @@ class date_views_filter_handler_simple extends views_handler_filter_date {
    */
   function date_default_value($prefix, $options = NULL) {
     if (empty($options)) {
-      $options = $this->options;
+      if ($this->options['is_grouped']) {
+        if ('All' == $this->options['group_info']['default_group']) {
+          return '';
+        }
+        $options = $this->options['group_info']['group_items'][$this->options['group_info']['default_group']]['value'][$prefix . '_group'];
+      }
+      else {
+        $options = $this->options;
+      }
     }
     // If this is a remembered value, use the value from the SESSION.
     if (!empty($this->options['expose']['remember'])) {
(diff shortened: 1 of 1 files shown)


core/modules/date/views/date_views_filter_handler_simple.inc
```diff
@@ -398,30 +444,53 @@ class date_views_filter_handler_simple extends views_handler_filter_date {
...
     else {
       // UI when the date selector is on the views configuration screen.
-      $default_date = '';
+      // Fix for grouped between filter. Input class has to be different
+      // for each group and default value needs to be taken from group item.
+      $relative_value_key = $prefix == 'max' ? 'default_to_date' : 'default_date';
+      $input_type_key = $prefix . '_choose_input_type';
+      if ($this->options['is_grouped']) {
+        $input_class = "$prefix-$element_no-choose-input-type";
+        $default_select_date_value = !empty($this->options['group_info']['group_items'][$element_no]['value'][$prefix . '_group'][$prefix]) ? $this->options['group_info']['group_items'][$element_no]['value'][$prefix . '_group'][$prefix] : $default_date;
```
indigoxela, line 453: I wonder if there's a way to make this code (this and the next 2 lines) more readable. Currently it's not really. 😉

argiepiano: Not really. I can't break the `?` or `:` into separate lines, since phpcs flags it as wrong.

indigoxela: I think, what makes this so unreadable is the attempt to squeeze it all in spaghetti-oneliners. 😉 

The null coalescing operator might help to make the lines shorter, but eventually some "regular" is/else notation would improve readability a lot more. Didn't look at it in depth, though.

argiepiano: Thanks, @indigoxela. Yes, I could redo those lines, but my concern first is that no one has tested if this works (@jenlampton e.g.). So until that's done, I don't think it's worth spending time on cosmetic fixes like that.

indigoxela: Fair enough, testing first :100: After that we look at code tweaks.

argiepiano:
Closing in favor of https://github.com/backdrop/backdrop/pull/5227

PR #5179, opened 2025-12-26
Issue #7029: Add Display type into views UI.

### LINKED ISSUE #7029: UX: Display the display type more clearly on the view configure form
## Description of the bug

I often struggle with determining the display type for the view I am configuring. The default pattern where the name of the displays are the types of displays does not hold up when you have many displays of the same type. This requires us to change the names of the displays to tell them apart, and then we loose the type of display.

## Steps To Reproduce

To reproduce the behavior:
1. Navigate to the views admin interface at `admin/structure/views`
2. Configure the "Administer content" view at `admin/structure/views/view/node_admin_content`
3. Clone the Page display as a block
4. Rename the page display to `one`
5. Clone the block display as Embed.
6. rename the enbed display to `two`.
7. Now try to figure out which one is the block, and which one is the embed.

## Actual behavior

Once I have renamed my displays, the tools I use to determine the type are:
* the label of the section containing `Block settings`, Page settings` etc (which is not available for all display types)
* The URL for the link to change the `Display name:`, which appears below the buttons to goggle display settings.

Image

Both of these are cumbersome ways to get information that is readily available to views. Let's put it in the UI instead :)

## Expected behavior

It should be easy to determine the display type. 

I have two recommendations:

1) Put the display type adjacent to the Display Name:

Image

2) Add the type on the views listing page, adjacent to the Display name, in parens:

Image

## Additional information

- Backdrop CMS version:
- Web server and its version:
- PHP version: 

jenlampton:
Fixes https://github.com/backdrop/backdrop-issues/issues/7029


jenlampton:
This may cause some views UI tests to fail... can update as needed!

quicksketch:
Merged into 1.x and 1.33.x. Thank you @jenlampton!

PR #5180, opened 2025-12-26
Issue #7027: Allow views exposed forms in blocks for Embed displays.

### LINKED ISSUE #7027: Allow the "Create exposed form block" setting for embed displays.
## Description of the need

I often need the exposed form for views to be displayed in a block when the view display I an working with is not a block or a page. In this case, I am using an `embed`. I don't know why is this setting is limited to only block displays (but I am about to find out!)

Image

## Proposed solution

I would like to try to make this setting available at all times. If anyone knows of any possible repercussions, perhaps we can address those individually. 

n/a

## Additional information

n/a

Views exposed forms can now be placed as blocks for 'embed' displays.

See related: https://github.com/backdrop/backdrop-issues/issues/7028

jenlampton:
Fixes https://github.com/backdrop/backdrop-issues/issues/7027


quicksketch:
This PR is replaced by https://github.com/backdrop/backdrop/pull/5224.

PR #5184, opened 2026-01-02
Issue #1528: Allow a layout to be selected on additional paths

### LINKED ISSUE #1528: Allow a layout to be selected on additional paths.

docwilmot:

Replaces https://github.com/backdrop/backdrop/pull/5169


core/modules/layout/includes/layout.class.inc
```diff
@@ -191,7 +200,27 @@ class Layout {
...
+   * @see layout_layout_presave()
+   * @since 1.33.0
+   */
+  public $additional_paths = array();
+
+  /**
+   * @var array
+   * @since 1.33.0
+   */
+  var $old_additional_paths = array();
```
quicksketch: Maybe scope this as protected if it's not needed outside the class?

```suggestion
  protected $old_additional_paths = array();
```

core/modules/layout/includes/layout.class.inc
```diff
@@ -685,6 +749,49 @@ class Layout {
     }
   }
 
+  /**
+   * Set a layout path.
```
quicksketch: Needs correct docblock.

core/modules/layout/js/layout.admin.js
```diff
@@ -49,7 +49,7 @@ Backdrop.behaviors.layoutConfigure = {
       e.stopPropagation();
     };
 
-    var $form = $('.layout-settings-form').once('layout-settings');
+    var $form = $('.layout-settings-form');
```
quicksketch, line 52: Hm, removing `.once()` gives me concern. I think this may result in repeated binding of the behaviors after every AJAX request. Is removing it necessary?
```suggestion
    var $form = $('.layout-settings-form').once('layout-settings');
```

docwilmot: I found that if `once()` is left in, then validation of paths stops running on additional paths as soon as a new path is added by AJAX. The JS continues to respond to input in the primary path regardless. Removing the `once()` was desperate, because I cant figure why this happens.

core/modules/layout/includes/layout.class.inc
docwilmot: Done

core/modules/layout/includes/layout.class.inc
docwilmot: Corrected

core/modules/layout/layout.module
```diff
@@ -1198,6 +1198,20 @@ function layout_load_multiple($layout_names, $skip_menu_items = FALSE) {
     }
   }
 
+  // Load menu items for additional paths.
+  foreach ($loaded_layouts as $loaded_layout) {
```
indigoxela: Defensive coding here to not break layout edit when layout_wildcard is installed?

```
    if (empty($layout_name)) {
      continue;
    }
```
It does the trick for me in my local testing. 😃

indigoxela: But I belief, the logic should get fixed differently (further up), because the variable gets only set in a loop for `$layouts_to_load` and isn't set at all otherwise. The next loop here relies on it being set, though, by using it as array key.

Further down the same variable name is used again in another loop...

docwilmot: > But I belief, the logic should get fixed differently (further up), because the variable gets only set in a loop for $layouts_to_load and isn't set at all otherwise. The next loop here relies on it being set, though, by using it as array key.

Oh wow, thanks, shocking this wasnt breaking things. Pushing fix shortly. Stops error in Wildcard module as well.

core/modules/layout/js/layout.admin.js
docwilmot: Thanks for the fix @quicksketch, tested works fine.

core/modules/layout/includes/layout.class.inc
```diff
@@ -685,6 +749,49 @@ class Layout {
     }
   }
 
+  /**
+   * Set additional menu items based on the path provided.
+   */
+  function setAdditionalMenuItems($path) {
```
herbdool, line 755: @docwilmot is this still being used?

core/modules/system/system.module
```diff
@@ -2676,7 +2676,14 @@ function system_block_view($delta = '', $settings = array()) {
         require_once BACKDROP_ROOT . '/' . $router_item['include_file'];
       }
       $block['subject'] = NULL;
-      $block['content'] = call_user_func_array($router_item['page_callback'], $router_item['page_arguments']);
+      // Don't allow to render the main content block if the block is rendering
```
jenlampton: This could be simplified as `// Don't allow the main page content to be rendered twice.

core/modules/layout/includes/layout.class.inc
quicksketch: Yes it's used in `layout.admin.inc`. GitHub just collapses that file because the changes are large.

core/modules/system/system.module
```diff
@@ -2676,7 +2676,14 @@ function system_block_view($delta = '', $settings = array()) {
...
+      // the main page content.
```
quicksketch:
```suggestion
      // Don't allow the main page content to be rendered twice.
```

quicksketch:
Merged into 1.x for 1.33.0. Thank you to our heroes @docwilmot and @herbdool for making this happen!

PR #5185, opened 2026-01-03
Issue #7038: Re-namespace Views select all pages functionality.

### LINKED ISSUE #7038: Improve Views "Select all Pages" namespacing and options
Follow-up to the original issue to include a "Select all pages" option here: https://github.com/backdrop/backdrop-issues/issues/3798

When merged, a lot of the naming of things had come over directly from the Views Bulk Operations (VBO) module. The naming of things could be improved to make them less derived from the contrib project and more integrated directly into Views.

quicksketch:
Fixes https://github.com/backdrop/backdrop-issues/issues/7038

Re-namespaces everything related to the new bulk form "Select all pages" functionality to be under `select_all_pages` (PHP side) or `views-select-all-pages--*` (CSS).

This initial PR might break some things, it's not fully tested but it's exploring the direction.

### DIFF
--- a/core/modules/views/handlers/views_handler_field_bulk_form.inc
+++ b/core/modules/views/handlers/views_handler_field_bulk_form.inc
@@ -150,9 +150,9 @@ class views_handler_field_bulk_form extends views_handler_field {
       if (!empty($this->view->result)) {
         $form['#attached']['js'][] = backdrop_get_path('module', 'views') . '/js/bulk_form.js';
         // Set by JS to indicate that all rows on all pages are selected.
-        $form['bulk_form_select_all_pages_flag'] = array(
+        $form['select_all_pages'] = array(
           '#type' => 'hidden',
-          '#attributes' => array('class' => 'bulk-form-select-all-pages-flag'),
+          '#attributes' => array('class' => 'views-select-all-pages--value'),
           '#default_value' => FALSE,
         );
         $enable_select_all_pages = FALSE;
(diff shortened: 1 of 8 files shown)


core/modules/views/handlers/views_handler_field_bulk_form.inc
```diff
@@ -150,9 +150,9 @@ class views_handler_field_bulk_form extends views_handler_field {
       if (!empty($this->view->result)) {
         $form['#attached']['js'][] = backdrop_get_path('module', 'views') . '/js/bulk_form.js';
```
argiepiano, line 151: Wondering if you want to rename the js file as well?

core/modules/views/js/bulk_form.js
```diff
@@ -36,84 +38,88 @@ Backdrop.bulk_form.initTableBehaviors = function(form) {
...
+      $('.views-select-all-pages--all-pages-checkbox', form).prop('checked', false);
+
       // Modify the value of the hidden form field.
-      $('.bulk-form-select-all-pages-flag', form).val('0')
+      $('input[name="select_all_pages"]', form).val('0');
 
       var table = $(this).closest('table')[0];
       if (table) {
         // If there's a "select all" row, hide it.
         if ($('.bulk-form-table-select-this-page', table).length) {
```
argiepiano: This was misnamed in my original PR, and remains misnamed here. With the new namespacing of classes, this line should read:

```js
if ($('.views-select-all-pages--row', table).length) {
```

core/modules/views/js/bulk_form.js
```diff
@@ -36,84 +38,88 @@ Backdrop.bulk_form.initTableBehaviors = function(form) {
...
           $('.bulk-form-views-table-row-select-all', table).hide();
```
argiepiano: This class has now changed with your renaming. There is loss of functionality at the moment because this was not switched.  It should be:

```js
$('.views-select-all-pages--row', table).hide();
```

argiepiano (changes requested):

core/modules/views/js/bulk_form.js
argiepiano: Hmm... I"m thinking that this `if` statement is not really needed. This was inherited from VBO, where that js file handled other things. We can simply hide the extra row without an if statement (see next line). If the row doesn't exist, no harm done.

PR #5186, opened 2026-01-03
Issue #7039: Fix type error when render element is no array

### LINKED ISSUE #7039: Regression: Type error in views with view of nodes
## Description of the bug

As I already suspected: some added type hints might cause trouble.
Follow up to coding standard fixes in core/includes/c* (common.inc)

## Steps To Reproduce

1. Generate a bunch of nodes with devel
2. Create a view of teasers but suppress link display
3. Save it and try to preview or visit the page

## Actual behavior

```
TypeError: hide(): Argument #1 ($element) must be of type array, null given, called in .../core/modules/node/templates/node.tpl.php on line 109 in hide() (line 7261 of ...core/includes/common.inc).

hide(): Argument #1 ($element) must be of type array, null given, called in ...core/modules/node/templates/node.tpl.php on line 109 
```
The view (page) is broken.

## Expected behavior

No type error.

## Additional information

- PHP version: 8.5

Caused by Issue: #7031: Coding standards - core/includes/c* - merged only recently, so not released, yet.

PR was: https://github.com/backdrop/backdrop/pull/5170

We need to fix doc block and type hint. Element can be null (in this case caused by calling hide() in a template in combination with suppressed links).

indigoxela:
Fixes https://github.com/backdrop/backdrop-issues/issues/7039

### DIFF
--- a/core/includes/common.inc
+++ b/core/includes/common.inc
@@ -7258,7 +7258,7 @@ function render(&$element) {
  * @see render()
  * @see show()
  */
-function hide(?array &$element) {
+function hide(&$element) {
   $element['#printed'] = TRUE;
   return $element;
 }
--- a/core/includes/common.inc
+++ b/core/includes/common.inc
@@ -7252,13 +7252,13 @@ function render(&$element) {
  * @param array $element
  *   The element to be hidden.
  *
- * @return array
+ * @return array|null
  *   The element.
  *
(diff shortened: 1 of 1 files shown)


quicksketch:
Merged into 1.x and 1.32.x. Thank you @indigoxela!

PR #5187, opened 2026-01-05
Issue 7040. Trigger 'change' when header checkbox is changed programmatically

### LINKED ISSUE #7040: tableselect.js: trigger a 'change' event when the table heading checkbox is checked or uncheck programmatically.
We ran into this problem when working on https://github.com/backdrop/backdrop-issues/issues/7038. 

When all checkboxes on the table are checked, tableselect checks the table header checkbox programmatically. The problem is that this programmatic checking doesn't trigger a `change` event, which makes it impossible to run other actions like showing the table row to "select all items on all pages" in that issue.

I believe the solution is to add a `trigger('change')` command to [this line](https://github.com/backdrop/backdrop/blob/81cab37b8f4edbe15bcfa94cbc8a73ac654feb04/core/misc/tableselect.js#L23), but I haven't tested this.

argiepiano:
Fixes https://github.com/backdrop/backdrop-issues/issues/7040

### DIFF
--- a/core/misc/tableselect.js
+++ b/core/misc/tableselect.js
@@ -19,8 +19,8 @@ Backdrop.tableSelect = function () {
   var updateSelectAll = function (state) {
     // Update table's select-all checkbox (and sticky header's if available).
     $(table).prev('table.sticky-header').addBack().find('th.select-all input[type="checkbox"]').each(function() {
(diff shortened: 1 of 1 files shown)


quicksketch:
Thank you @argiepiano! Merged into 1.x and 1.32.x. Nice fix!

PR #5195, opened 2026-01-09
Issue #7041: Layout list additional paths

### LINKED ISSUE #7041: Follow up to Layout paths: Update Layout listing page to include additional paths
## Description of the need

This is a follow-up to https://github.com/backdrop/backdrop-issues/issues/1528, as a UX improvement we should add the additional paths to the listing page, or at the very least an indicator that a layout does have additional paths. 

Image

## Proposed solution

Ideas include: 

Changing the cell that contains `Path: example` to `Path(s): example1, example2, example3` -  or maybe we separate these by line breaks, or list them in a bulleted list:

Image

Or Adding an ion or more indicator to the cell that contains `Path: example` so it would be `Path: example + more` (or better) 

Image

docwilmot:
Fixes https://github.com/backdrop/backdrop-issues/issues/7041

### DIFF
--- a/core/modules/layout/layout.theme.inc
+++ b/core/modules/layout/layout.theme.inc
@@ -64,7 +64,11 @@ function theme_layout_info($variables) {
   $output .= $preview;
   $output .= '<div class="layout-detail">' . $label . '</div>';
   $output .= '</div>';
-
+  if (!empty($layout->additional_paths)) {
+    $output .= '<div class="add-path-toggle">';
+    $output .= l('Additional paths: ' . implode(', ', $layout->additional_paths), '/');
+    $output .= '</div>';
+  }
   return $output;
 }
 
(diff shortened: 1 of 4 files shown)


core/modules/layout/layout.theme.inc
```diff
@@ -64,7 +64,11 @@ function theme_layout_info($variables) {
   $output .= $preview;
   $output .= '<div class="layout-detail">' . $label . '</div>';
   $output .= '</div>';
-
+  if (!empty($layout->additional_paths)) {
+    $output .= '<div class="add-path-toggle">';
+    $output .= l('Additional paths: ' . implode(', ', $layout->additional_paths), '/');
```
herbdool, line 69: Make the path `#` if possible.

jenlampton:
Replaced by https://github.com/backdrop/backdrop/pull/5209

PR #5197, opened 2026-01-15
Issue #7053: Fix path description if it is a path created by the same layout doing the checking

### LINKED ISSUE #7053: Layout additional paths are displayed as overrides even if they're not, in some cases
This is a follow-up to https://github.com/backdrop/backdrop-issues/issues/1528. Given a view that has two pages with unique routes:

Image

A layout that overrides those routes, will state that even routes not included in that view are overrides (like for -3, -4, -5):

Image

I noticed it's only doing it for paths ending in numbers so suggests a faulty regex.

Image

herbdool:
Fixes https://github.com/backdrop/backdrop-issues/issues/7053


core/modules/layout/layout.admin.inc
```diff
@@ -859,14 +861,24 @@ function layout_settings_form($form, &$form_state, Layout $layout) {
...
+ * @param string|null $layout_name
+ *   The name of the layout that is doing the checking.
+ * @param bool $is_new
+ *   Whether this is a new layout.
+ *
+ * @return string
+ *   HTML markup of description text for a path entry field.
  */
-function _layout_path_matches_description($path, $is_new) {
+function _layout_path_matches_description($path, $layout_name, $is_new) {
```
jenlampton, line 875: Should we be adding the new parameter at the end of the others, instead of in the middle, perhaps with a default NULL value, so that it's backwards compatible?

core/modules/layout/layout.admin.inc
```diff
@@ -915,14 +927,40 @@ function _layout_path_matches_description($path, $is_new) {
...
-function _layout_menu_router_path_exists($path) {
+function _layout_menu_router_path_exists($path, $layout_name) {
   $path_matches = db_query('
     SELECT path, page_callback
     FROM {menu_router}
     WHERE path = :path
     ', array(':path' => $path))
     ->fetchAll();
+  if (!empty($path_matches)) {
+    if (!empty($layout_name)) {
```
jenlampton, line 948: Can we and these two together, like... 
```
if (!empty($path_matches) && !empty($layout_name)) {
```

quicksketch:
Merged into 1.x and 1.33.x. Thank you @herbdool and @jenlampton.

PR #5201, opened 2026-01-16
Issue #7057: ensure instance isn't required

### LINKED ISSUE #7057: Regression in field_is_empty()
## Description of the bug

The title needs to get updated, as soon as we get an idea, what we can do

## Steps To Reproduce

Lots of contrib modules are affected by https://github.com/backdrop/backdrop-issues/issues/6987

Several got new releases, already (WIP), but when so many contrib modules are affected, lots of custom code might be, too.
And as it's a fatal, I belief, we should do something about it.

## Actual behavior

Fatal errors under circumstances...

```
ArgumentCountError: Too few arguments to function text_field_is_empty()
```

## Expected behavior

...

## Additional information

- Backdrop CMS version: introduced with 1.33
- Web server and its version:
- PHP version: 

herbdool:
Fixes https://github.com/backdrop/backdrop-issues/issues/7057

### DIFF
--- a/core/modules/field/modules/text/text.module
+++ b/core/modules/field/modules/text/text.module
@@ -218,7 +218,7 @@ function text_field_load($entity_type, $entities, $field, $instances, $langcode,
  */
 function text_field_is_empty($item, $field, $instance = NULL) {
   if (!isset($item['value']) || $item['value'] === '') {
-    if (!empty($instance) && isset($item['format']) && $item['format'] !== '' && !_text_field_uses_default_format($item, $instance)) {
+    if (isset($instance) && isset($item['format']) && $item['format'] !== '' && !_text_field_uses_default_format($item, $instance)) {
       // If this text field has a text format setting, save it regardless of
       // empty value to keep the format. Unless it's the default format.
       return FALSE;
--- a/core/modules/field/modules/text/text.module
+++ b/core/modules/field/modules/text/text.module
@@ -216,7 +216,7 @@ function text_field_load($entity_type, $entities, $field, $instances, $langcode,
 /**
  * Implements hook_field_is_empty().
  */
-function text_field_is_empty($item, $field, $instance) {
+function text_field_is_empty($item, $field, $instance = NULL) {
   if (!isset($item['value']) || $item['value'] === '') {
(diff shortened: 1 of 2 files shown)


core/modules/field/modules/text/text.module
```diff
@@ -216,9 +216,9 @@ function text_field_load($entity_type, $entities, $field, $instances, $langcode,
 /**
  * Implements hook_field_is_empty().
  */
-function text_field_is_empty($item, $field, $instance) {
+function text_field_is_empty($item, $field, $instance = NULL) {
   if (!isset($item['value']) || $item['value'] === '') {
-    if (isset($item['format']) && $item['format'] !== '' && !_text_field_uses_default_format($item, $instance)) {
+    if (!empty($instance) && isset($item['format']) && $item['format'] !== '' && !_text_field_uses_default_format($item, $instance)) {
```
quicksketch: In this scenario, I prefer avoiding the NOT operator to check for a NULL value. This should have the same effect for scenarios where `$instance` is not provided.

```suggestion
    if (isset($instance) && isset($item['format']) && $item['format'] !== '' && !_text_field_uses_default_format($item, $instance)) {
```

core/modules/field/modules/text/text.module
```diff
@@ -216,9 +216,9 @@ function text_field_load($entity_type, $entities, $field, $instances, $langcode,
 /**
  * Implements hook_field_is_empty().
  */
-function text_field_is_empty($item, $field, $instance) {
+function text_field_is_empty($item, $field, $instance = NULL) {
   if (!isset($item['value']) || $item['value'] === '') {
-    if (isset($item['format']) && $item['format'] !== '' && !_text_field_uses_default_format($item, $instance)) {
+    if (isset($instance) && isset($item['format']) && $item['format'] !== '' && !_text_field_uses_default_format($item, $instance)) {
```
indigoxela, line 221: Wouldn't it make more sense to catch the null $instance in function _text_field_uses_default_format()?

There's one check there that works without instance - if the format is the global default format.
My concern here is, that some things slip into the "elseif" or end up with "false", that don't belong there.

quicksketch: Yeah, I agree. This PR as-is will probably stop errors from happening but return the wrong results. What previously had been considered empty will probably not be, since the actual `_text_field_uses_default_format()` function is no longer being called at all.

PR #5202, opened 2026-01-16
Issue #7058: make instance optional

### LINKED ISSUE #7058: Follow-up to #6987: Bunch of prominent contrib break with change to _field_filter_items()
## Description of the bug

Bunch of contrib modules are affected by https://github.com/backdrop/backdrop-issues/issues/6987.

Yes, it's a private function, but a bunch of prominent modules are using it like a public one. https://github.com/search?q=org%3Abackdrop-contrib+_field_filter_items&type=code

We could make it a bit easier by making `$instance` optional in `_field_filter_items()`. I'm wondering if we also make a version of that function public so we know not to touch it without making sure it's backwards-compatible in the future.

herbdool:
Fixes https://github.com/backdrop/backdrop-issues/issues/7058

### DIFF
--- a/core/modules/field/field.module
+++ b/core/modules/field/field.module
@@ -768,18 +768,21 @@ function field_get_values(Entity $entity, $field_name, $value_key = 'value', $la
 /**
  * Helper function to filter out empty field values.
  *
+ * Treat this as a public function. Multiple modules have been treating it as
+ * such.
+ *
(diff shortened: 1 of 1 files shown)


quicksketch:
Merged into 1.x and 1.33.x. Thank you @herbdool and @indigoxela!

PR #5205, opened 2026-01-21
Removed duplicate #process callback from the password element.

### LINKED ISSUE #7062: Remove duplicate #process callback from the password element
The password element, added by system.module in system_element_info() defines a #process callback and then immediately overwrites it with a different #process callback.

The first #process callback is unnecessary and can be removed.

DevinCarlson:
Fixes [7062](https://github.com/backdrop/backdrop-issues/issues/7062)


core/modules/system/system.module
```diff
@@ -397,7 +397,6 @@ function system_element_info() {
     '#input' => TRUE,
     '#size' => 60,
     '#maxlength' => 128,
-    '#process' => array('ajax_process_form'),
```
avpaderno, line 400: Both the process callbacks should probably be used, similarly to what `$types['checkbox']` does.

```php
$types['checkbox'] = array(
  '#input' => TRUE,
  '#return_value' => 1,
  '#theme' => 'checkbox',
  '#process' => array('form_process_checkbox', 'ajax_process_form'),
  '#theme_wrappers' => array('form_element'),
  '#title_display' => 'after',
);
```

avpaderno (changes requested):

core/modules/system/system.module
quicksketch: To clarify @avpaderno's comment above, this line should be changed so that both `#process` functions are added.

```suggestion
    '#process' => array('user_form_process_password', 'ajax_process_form'),
```

quicksketch:
Merged into 1.x and 1.33.x. Thank you @DevinCarlson and @avpaderno!

PR #5206, opened 2026-01-22
Issue #7060: Fix taxonomy term pages with few nodes displayed.

### LINKED ISSUE #7060: Taxonomy term pages with image put image at bottom of page for only 1 or 2 items with that term
## Description of the bug

If we create a taxonomy term with an image field and then give some nodes a term reference field that references that term, on the taxonomy term page, the term image shows incorrectly at the bottom of the page if there are only 1 or 2 nodes with that term, but correctly at the top of the page with 3 or more nodes.

## Steps To Reproduce

To reproduce the behavior:

* Create a vocabulary, e.g., "Clothes."
* Add an image field to it. Give the image field a default image.
* Create a new content type.
* Add a term reference field to it.
* Create three nodes of the content type, each of which has the term reference set to the "Clothes" vocabulary.
* Go to the term page for "Clothes". It will show the term image at the top of the list of the 3 nodes, as one would expect.
* Now, for one of the nodes, remove its term reference.
* Revisit the term page for "Clothes." It now puts the term image at the bottom of the list of 2 nodes.

## Expected behavior

The term should show up at the top of the list no matter how many nodes there are. Here's three nodes:

Image

And here's two nodes:

Image

## Additional information

Tested on Backdrop 1.33.0. Initially discovered by an Ubercart user and reported there as https://github.com/backdrop-contrib/ubercart/issues/575.

bugfolder:
Fixes https://github.com/backdrop/backdrop-issues/issues/7060.

### DIFF
--- a/core/modules/taxonomy/taxonomy.pages.inc
+++ b/core/modules/taxonomy/taxonomy.pages.inc
@@ -48,6 +48,7 @@ function taxonomy_term_page(TaxonomyTerm $term) {
     '#prefix' => '<div class="term-listing-heading">',
     '#suffix' => '</div>',
     'term' => taxonomy_term_view($term, 'full'),
+    '#weight' => -1,
   );
 
   if ($nids = taxonomy_select_nodes($term->tid, TRUE, $site_config->get('default_nodes_main'))) {


quicksketch:
Merged into 1.x and 1.33.x. Thank you @bugfolder and @argiepiano!

PR #5207, opened 2026-01-24
Issue #7057: Fix backwards compatibility for text_field_is_empty()

### LINKED ISSUE #7057: Regression in field_is_empty()

indigoxela:
Fixes https://github.com/backdrop/backdrop-issues/issues/7057

### DIFF
--- a/core/modules/field/field.api.php
+++ b/core/modules/field/field.api.php
@@ -697,7 +697,7 @@ function hook_field_prepare_translation($entity_type, $entity, $field, $instance
  *   An item that may or may not be empty.
  * @param array $field
  *   The field to which $item belongs.
- * @param array $instance
+ * @param array|null $instance
  *   Field API instance definition.
  *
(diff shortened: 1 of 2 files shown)


quicksketch:
Merged into 1.x and 1.33.x. Thank you @indigoxela!

PR #5221, opened 2026-02-26
Issue: #7080: disable author and date by default on new content types

### LINKED ISSUE #7080: Disable "Display author and date information" by default on content types
## Description of the need

Let's disable the "Display author and date information" option by default on new content types.

In my experience there is usually somewhere between 0 and 1 content types per site that need to display the author and date when viewing the content. Usually that content type is "blog post" or "news". With all the content types I've created I estimate it's about 10% or less.

For confirmation I took a look at all the content types on backdropcms.org and only 1 out of 23 content types are displaying the author and date. That's 4%!

So why do we still have it enabled by default?

## Proposed solution

Disable it.

herbdool:
Fixes https://github.com/backdrop/backdrop-issues/issues/7080

### DIFF
--- a/core/modules/node/node.install
+++ b/core/modules/node/node.install
@@ -477,8 +477,8 @@ function node_update_1005() {
       'revision_enabled' => TRUE,
       'revision_default' => in_array('revision', $options),
       'node_preview' => update_variable_get('node_preview_' . $node_type->type, BACKDROP_OPTIONAL),
-      'node_submitted' => update_variable_get('node_submitted_' . $node_type->type, FALSE),
-      'node_user_picture' => update_variable_get('node_user_picture_' . $node_type->type, FALSE),
+      'node_submitted' => update_variable_get('node_submitted_' . $node_type->type, TRUE),
+      'node_user_picture' => update_variable_get('node_user_picture_' . $node_type->type, TRUE),
       'language' => update_variable_get('node_type_language_' . $node_type->type),
     );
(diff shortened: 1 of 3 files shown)


core/modules/node/node.install
```diff
@@ -477,8 +477,8 @@ function node_update_1005() {
       'revision_enabled' => TRUE,
       'revision_default' => in_array('revision', $options),
       'node_preview' => update_variable_get('node_preview_' . $node_type->type, BACKDROP_OPTIONAL),
-      'node_submitted' => update_variable_get('node_submitted_' . $node_type->type, TRUE),
-      'node_user_picture' => update_variable_get('node_user_picture_' . $node_type->type, TRUE),
+      'node_submitted' => update_variable_get('node_submitted_' . $node_type->type, FALSE),
+      'node_user_picture' => update_variable_get('node_user_picture_' . $node_type->type, FALSE),
```
quicksketch: I think when updating, these should retain the default of `TRUE`, since that is what it was in D7 if this variable was not set. See https://git.drupalcode.org/project/drupal/-/blob/7.x/modules/node/node.module#L1530

yorkshire-pudding (approved):
Noted that the change to .install has been removed and just the change to defaults remains. Also noted is the small, but significant improvement to the clarity of the summary text.

quicksketch:
Merged into 1.x only for 1.34.0. Thanks @herbdool!

PR #5222, opened 2026-02-26
Issue #7081: add hook to allow other permissions on node type form

### LINKED ISSUE #7081: Allow other modules to display permissions on the content type's Permissions tab
## Description of the need

Right now the list of permissions is only provided by core's node module. If a contrib (or other core) module has content type specific permissions, it would be user-friendly to also be able to display them on the content type form in the Permissions tab.

It would look like this where the last two permissions are provided by a custom module calling a hook:

Image

herbdool:
Fixes https://github.com/backdrop/backdrop-issues/issues/7081

### DIFF
--- a/core/modules/node/node.api.php
+++ b/core/modules/node/node.api.php
@@ -965,6 +965,35 @@ function hook_node_type_delete($info) {
 }
 
 /**
+ * Add to the matrix of node permissions on a node type to be
+ * displayed on the node type form. This would be used in addition to
+ * hook_permission() which supplies the actual permissions.
+ *
+ * This hook is invoked from node_type_form_permissions.
+ *
+ * @param string $type
+ *   The node type.
+ */
+function hook_node_type_list_permissions($type) {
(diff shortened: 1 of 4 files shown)


core/modules/node/node.api.php
```diff
@@ -964,6 +964,35 @@ function hook_node_type_delete($info) {
...
 
+/**
+ * Add to the matrix of node permissions on a node type to be
+ * displayed on the node type form. This would be used in addition to
+ * hook_permission() which supplies the actual permissions.
+ *
+ * This hook is invoked from node_type_form_permissions.
+ *
+ * @param string $type
+ *   The node type.
```
quicksketch, line 975:
```suggestion
 * @param string $type
 *   The node type.
 * 
 * @since 1.34.0 Hook added.
```

quicksketch:
Thanks @herbdool! Merged into 1.x for 1.34.0.

PR #5223, opened 2026-03-03
Issue #7075: add options for URL history for ajax views

### LINKED ISSUE #7075: Incorporate Views AJAX History
## Description of the need

Incorporate [views_ajax_history](https://github.com/backdrop-contrib/views_ajax_history) into core. It will allow the View to be bookmarked, navigated, copied and pasted even when AJAX is used to filter.

The module now uses the widely-supported [History API](https://developer.mozilla.org/en-US/docs/Web/API/History_API) browser feature to update the URL query parameters.

herbdool:
Fixes https://github.com/backdrop/backdrop-issues/issues/7075


core/modules/views/js/ajax_view.js
```diff
@@ -4,6 +4,18 @@
  */
 (function ($) {
 
+// Keep to check if there are extra parameters in the original URL.
+var original = {
+  path: window.location.protocol + "//" + window.location.hostname + (window.location.port ? ':' + window.location.port : '') + window.location.pathname,
+  query: window.location.search || ''
+};
+
+window.onpageshow = function (event) {
```
quicksketch, line 13: This should use an event listener rather than claim the entire `onpageshow` method. Something like:

```suggestion
window.addEventListener("pageshow", function (event) {
```

core/modules/views/js/ajax_view.js
```diff
@@ -168,4 +188,170 @@ Backdrop.ajax.prototype.commands.viewsScrollTop = function (ajax, response) {
...
 };
 
+/**
+ * Override beforeSerialize to handle click on pager links.
+ *
+ * @param $element
+ *   jQuery DOM element
+ * @param options
+ */
+Backdrop.ajax.prototype.beforeSerialize = function (element, options) {
```
quicksketch, line 198: It looks like this is overriding the entire `beforeSerialize` method of all AJAX requests. Meaning this new code will replace the core `beforeSerialize` method entirely. Ideally what would happen is `beforeSerialize` (and the methods below) would only be overridden for Views-specific ajax objects (i.e. exposed filters and pagers) and not _all_ ajax objects that might be on the page.

I am not sure what to suggest here since I assume this is coming from contrib; where this is a proven solution -- and maybe the only solution that was available from contrib.

It might not be possible to do what I'm suggesting without modifying `core/misc/ajax.js` itself.

.cspell/backdrop.dic
```diff
@@ -427,6 +427,7 @@ sidebarborders
 simpletest
 simpletests
 sioc
+sitename
```
quicksketch: This can be removed. The dictionary moved to https://github.com/backdrop-ops/cspell and I added `sitename` to that repository.

core/modules/views/js/ajax_view.js
herbdool: I noticed that this is how the D8+ version also overrides the methods. I wouldn't want to put all this stuff into `core/misc/ajax.js` but I'm not sure how to make it extendable.

@quicksketch did you notice in the PR where it restores original methods?

```
/**
 * Keep the original beforeSubmit method to be available for overrides.
 */
Backdrop.Views.beforeSubmit = Backdrop.ajax.prototype.beforeSubmit;
```

etc.

herbdool:
Replaced by https://github.com/backdrop/backdrop/pull/5254

PR #5225, opened 2026-03-03
Converted form action confirmation pages to confirmation dialogs.

### LINKED ISSUE #3769: [META][D8][UX] Convert all confirmation pages to confirmation dialogs instead.
I had filed a couple of separate issues for this back in 2015, about specific pages. Let's make this issue here a meta to list all such issues.

First step would be:

- [ ] #3771 | [UX] Introduce a `confirm_dialog()` function (and use that one instead of `confirm_form()`)

Specific confirmation pages to be converted to dialogs:

- [ ] #769 | [UX] Move the confirmation dialog for deletion of nodes to a modal.
- [ ] #770 | [UX] Move the confirmation dialog for deletion of menu links to a modal.
- [ ] #3770 | [UX] Move the confirmation dialog for enabling/uninstalling modules to a modal.

Related D8 issues:

- [Add abstracted dialog to core (resolves accessibility bug)](https://www.drupal.org/project/drupal/issues/1667742)
- [[META] Convert all confirm forms already converted to new routing system to use modal dialog](https://www.drupal.org/project/drupal/issues/1842036)

Related D7 issues:
- [Ajax Popups in Drupal 7: Adding Modal Dialogs to Help, Confirmations and Filter tips (Unified)](https://www.drupal.org/project/drupal/issues/193311)
- [Popbox (Popups Lite): Adding Modal Dialogs to Confirmations](https://www.drupal.org/project/drupal/issues/374646)

(Triaged as possible accessibility issue on 08/17/2921. This one may or many not qualify, please add the [A11Y] tag if you think it qualifies.)

DevinCarlson:
Fixes [#3769](https://github.com/backdrop/backdrop-issues/issues/3769)

### DIFF
--- a/core/modules/block/block.admin.inc
+++ b/core/modules/block/block.admin.inc
@@ -153,7 +153,18 @@ function block_admin_configure($form, &$form_state, $delta = NULL) {
       '#type' => 'link',
       '#title' => t('Delete'),
       '#href' => 'admin/structure/block/manage/' . $delta . '/delete',
-      '#attributes' => array('class' => array('button', 'button-secondary', 'form-delete')),
+      '#attributes' => array(
+        'class' => array('button', 'button-danger', 'use-ajax'),
+        'data-dialog' => TRUE,
+        'data-dialog-options' => json_encode(array(
+          'width' => 880,
+        )),
+      ),
+      '#attached' => array(
+        'library' => array(
+          array('system', 'backdrop.ajax'),
+        ),
+      ),
     );
   }
   $form['actions']['cancel'] = array(
@@ -275,7 +286,7 @@ function block_custom_block_delete($form, &$form_state, $delta) {
   $form['info'] = array('#type' => 'hidden', '#value' => $custom_block['info']);
   $form['delta'] = array('#type' => 'hidden', '#value' => $delta);
 
-  return confirm_form($form, t('Are you sure you want to delete the block %name?', array('%name' => $custom_block['info'])), 'admin/structure/block', '', t('Delete'), t('Cancel'));
+  return confirm_form($form, t('Are you sure you want to delete the block %name?', array('%name' => $custom_block['info'])), 'admin/structure/block', t('This action cannot be undone.'), t('Delete'), t('Cancel'));
 }
 
 /**
--- a/core/modules/field_ui/field_ui.admin.inc
+++ b/core/modules/field_ui/field_ui.admin.inc
@@ -1864,7 +1864,6 @@ function field_ui_field_delete_form_submit($form, &$form_state) {
  * @see field_ui_menu()
  * @see field_ui_field_edit_form_validate()
  * @see field_ui_field_edit_form_submit()
- * @see field_ui_field_edit_form_delete_submit()
  * @ingroup forms
  */
 function field_ui_field_edit_form($form, &$form_state, $instance) {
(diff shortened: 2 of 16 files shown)


core/modules/system/system.module
```diff
@@ -3622,12 +3622,17 @@ function confirm_form($form, $question, $path, $description = NULL, $yes = NULL,
     '#value' => $yes ? $yes : t('Confirm'),
     '#attributes' => array('class' => array('button-danger')),
   );
-  $form['actions']['cancel'] = array(
-    '#type' => 'link',
-    '#title' => $no ? $no : t('Cancel'),
-    '#href' => $options['path'],
-    '#options' => $options,
-  );
+  // Only display a 'cancel' link when the form is not being rendered in a
```
jenlampton, line 3625: I think it would be fine to add the cancel button in the dialog also.

core/modules/block/block.admin.inc
```diff
@@ -275,7 +286,7 @@ function block_custom_block_delete($form, &$form_state, $delta) {
   $form['info'] = array('#type' => 'hidden', '#value' => $custom_block['info']);
   $form['delta'] = array('#type' => 'hidden', '#value' => $delta);
 
-  return confirm_form($form, t('Are you sure you want to delete the block %name?', array('%name' => $custom_block['info'])), 'admin/structure/block', '', t('Delete'), t('Cancel'));
+  return confirm_form($form, t('Are you sure you want to delete the block %name?', array('%name' => $custom_block['info'])), 'admin/structure/block', t('This action cannot be undone.'), t('Delete'), t('Cancel'));
```
jenlampton, line 289: This is the default value so it shouldn't need to be passed in:

```
$description = isset($description) ? $description : t('This action cannot be undone.');
```
Perhaps that should be `!empty()` instead of `isset` if it's not working as expected?

jenlampton:
@DevinCarlson  thank you so much for taking this on!  This is a feature I have wanted in core for so long and I actually need it for a project I am working on right now, so your timing is perfect, and I couldn't be more grateful 🙏 

This approach makes me a little nervous because we are rather extensively modifying existing forms. This might be unavoidable, but I'd love to think about some other options first :)  Every form structure change has the potential to break `form_alter` implementations in contrib and custom modules.

I'd also like to decrease the amount of work any contrib project needs to do to use this new feature. It would be ideal if we could find a way for core to add the 'backdrop.ajax` library and set the destination, rather than needing to do that work for every use of confirm_form();

What if we added the same class & attributes to existing `<input type=submit">` buttons, and those buttons could process as usual but open the resulting confirm form in a modal. 

We could modify `theme_submit` (or `theme_button`) to look for the `use-ajax` class and automatically load the  'backdrop.ajax` library, and set a destination if it was not previously set. (Maybe as a `data-dialog-destination` attribute or something)

I'd like to play with this myself but I'm out of time for the day. I'll see if I can come back to it soom!

core/modules/field_ui/field_ui.admin.inc
```diff
@@ -2049,34 +2048,35 @@ function field_ui_field_edit_form($form, &$form_state, $instance) {
...
+    '#title' => t('Delete field'),
+    '#href' => $admin_path . '/fields/' . $instance['field_name'] . '/delete',
+    '#options' => array(
       'query' => $destination,
+      'attributes' => array(
+        'class' => array('button', 'button-danger', 'use-ajax'),
+        'data-dialog' => TRUE,
+        'data-dialog-options' => json_encode(array(
+          'width' => 880,
+        )),
```
quicksketch, line 2070: I think setting a class on the dialog, like "confirmation-dialog" and then setting the size of the dialog with CSS in Seven might be a better approach than coding a fixed width for all the various dialogs added here.

PR #5227, opened 2026-03-11
Fix date views grouped filters and relative dates

### LINKED ISSUE #6976: Exposed group filter for date not working

juanolalla:
Fixes https://github.com/backdrop/backdrop-issues/issues/6976

Based on previous Backdrop PR: https://github.com/backdrop/backdrop/pull/5176, which is based on the latest [Drupal PR](https://www.drupal.org/files/issues/2022-08-14/date-exposed_grouped_filters-1876168-171.patch) for https://www.drupal.org/project/date/issues/1876168


juanolalla:
In the last commit I have fixed a regression bug we had introduced with the expected parameter value type, and also fixed an incorrect SQL conversion when using grouped exposed filters with relative dates.

core/modules/date/views/date_views_filter_handler_simple.inc
```diff
@@ -398,41 +617,69 @@ class date_views_filter_handler_simple extends views_handler_filter_date {
     }
     else {
       // UI when the date selector is on the views configuration screen.
-      $default_date = '';
+      // Fix for grouped between filter. Input class has to be different
+      // for each group and default value needs to be taken from group item.
+      $input_type_key = $prefix . '_choose_input_type';
+      if ($this->options['is_grouped']) {
```
herbdool, line 623:
```suggestion
      if ($this->options['is_grouped']) {
        $source = "#edit-options-group-info-group-items-{$element_no}-operator";
```

This allows the date fields to work properly with the operator selection, when in grouped filter.

laryn:
Thank you! Merged into 1.x and 1.34.x.

PR #5237, opened 2026-04-06
Issue #7083: Icon browser/picker in core

### LINKED ISSUE #7083: Icon browser/picker in core
## Description of the need

Now that we have an icon library in core, I've wanted to include them in some contrib modules I'm working on. It would be nice to have a nice simple icon browser/picker. We have an icon brower in contrib https://backdropcms.org/project/icon_browser, but it does not come with a picker. 

@alanmels has introduced a nice icon picker in https://backdropcms.org/project/menu_icons

If I don't want to recreate an icon picker in each of my contrib projects, I'm currently left with a dependancy on "Menu Icons". We could add an icon picker to the "Icon Browser" module. 

I don't know for sure that this belongs in core, but I think if it's not, we're going to get lots of contrib modules with their own solutions. I am anxious to hear what others think. 

## Proposed solution

I'm working on a proof of concept PR. 

Image

Image

1. Icon Browser - https://backdropcms.org/project/icon_browser
2. Menu Icons - https://backdropcms.org/project/menu_icons
3. Each contrib module ships with it's own icon picker

Backdrop now includes an API for an icon picker available to other core features and contrib projects. This icon picker allows a person to browse the icon library and pick a specific icon to insert into their field or website.

docwilmot:
Fixes https://github.com/backdrop/backdrop-issues/issues/7083


docwilmot:
I've committed a temporary change to path `admin/structure/file-types/settings` where replaced the token browser link with the new icon browser link, for testing, because there isn't anywhere in core with that link as yet.

core/modules/system/system.module
```diff
@@ -4887,3 +4934,192 @@ function system_transliterate_machine_name($string, $options = array()) {
...
+      $_GET[$param] =  ${$param};
+      unset($form_state[$param]);
+    }
+    elseif (!empty($_GET[$param])) {
+      ${$param} = $_GET[$param];
+    }
+    else {
+      ${$param} = $config->get('icon_browser_' . $param);
+    }
+  }
```
indigoxela, line 4970: To get back the token here (after the loop):
```
$_GET['token'] = backdrop_get_token('icon-browser-endpoint');
```
I'm not sure if that is the right approach, I only saw that this seems to fix the problem with filters wiping out the access token.

core/modules/system/system.module
```diff
@@ -4887,3 +4934,192 @@ function system_transliterate_machine_name($string, $options = array()) {
...
+
```
indigoxela, line 4971: In theory - with the same trick - you could get the path (query "q") back here without fiddling in JavaScript.

```
$_GET['q'] = 'icon-browser/dialog';
```
Not sure, either, if this is the proper way. It seems to work.

core/modules/system/config/system.core.json
```diff
@@ -39,6 +40,19 @@
...
+      "thin": "Thin",
+      "light": "Light",
+      "bold": "Bold",
+      "duotone": "Duotone"
+    },
+    "icon_browser_provider": "any",
+    "icon_browser_style": "any",
+    "icon_browser_keyword": "",
+    "icon_browser_size": "medium",
+    "icon_browser_include_brands": 0,
```
indigoxela: I belief, some of the latest changes in this PR might make some of these config items obsolete.

And TBH, I'm not sure if we need all of them. And if so, if it has to be config (or rather hook). And even if it's config - if those items have to go into the - always loaded - system.core.json or some other file.
Only brainstorming here. 😉

core/modules/system/css/icon_browser.css
```diff
@@ -0,0 +1,59 @@
...
+  width: 36px;
+  margin: 0 auto;
+}
+
+.icon-browser-list .icon-wrapper .icon  svg {
+  width: 36px;
+  height: 36px;
+}
+.ui-dialog.icon-browser-dialog {
+  max-width:75%;
```
indigoxela: Switching to `vw` (viewport width) here with max-width would make this dialog mobile friendly.

`%` doesn't work, as the dialog has position "fixed", `vw` also works with position absolute/fixed/whatever... 😉

core/modules/system/config/system.core.json
docwilmot: You're right, got rid of all config entries.

core/modules/system/css/icon_browser.css
docwilmot: Interestingly, I see no difference in behaviour between vh and % on my local (Chrome) but changed it anyway.

indigoxela: You can see the difference on *smaller* screens or when resizing the browser window.
`%` does not resize the dialog, but `vw` makes it so the dialog isn't wider than the browser window, which would required to scroll horizontally to see all content. :wink: 

You made it 65vw, BTW (not 75).

core/modules/system/system.module
```diff
@@ -2159,6 +2194,18 @@ function system_library_info() {
     ),
   );
 
+  // jQuery treeTable plugin.
```
indigoxela, line 2197: Copy-paste, right? 😆

Better remove or fix this misleading comment.

core/modules/system/system.module
```diff
@@ -56,6 +56,11 @@ define('EMAIL_MAX_LENGTH', 254);
  */
 define('TOKEN_MAX_DEPTH', 9);
 
+/**
+ * The maximum depth for token tree recursion.
```
indigoxela, line 60: Another misleading comment (copy-paste apparently). 😉

indigoxela: Do we have to make the max hardcoded? Or is there a way to let modules (or themes) adapt that?

core/modules/system/system.module
```diff
@@ -4887,3 +4934,197 @@ function system_transliterate_machine_name($string, $options = array()) {
...
+  );
+  if (backdrop_is_ajax()) {
+    $form['icon_filters']['submit']['#ajax'] = array(
+      'callback' => 'system_icon_browser_ajax_form_open_dialog',
+    );
+    $form['icon_filters']['reset']['#ajax'] = array(
+      'callback' => 'system_icon_browser_ajax_form_open_dialog',
+    );
+
+    backdrop_add_js(backdrop_get_path('module', 'system') . '/js/icon_browser.js');
```
indigoxela, line 5053: Do we need this here, anyway? BTW, why attaching as regular JS, while we have a defined library?
Or is this just for the demo link?

core/modules/file/file.admin.inc
```diff
@@ -522,7 +522,7 @@ function file_settings_form($form, &$form_state) {
   $form['default_file_directory'] = array(
     '#type' => 'textfield',
     '#title' => t('Default file directory'),
-    '#description' => t('Subdirectory where files will be stored if the file is uploaded through the file/add page. This field supports tokens.') . ' ' . theme('token_tree_link'),
+    '#description' => t('Subdirectory where files will be stored if the file is uploaded through the file/add page. This field supports tokens.') . ' ' . theme('icon_browser_link'),
```
indigoxela, line 525: Let's not forget to remove this link, as soon as we're done with manual testing.
It's only here to demo functionality.

herbdool:
I'm going to close since there appears to be a newer PR.

PR #5239, opened 2026-04-14
#4293: Indicate disabled Views displays.

### LINKED ISSUE #4293: [UX][A11Y] Visually indicate a View Display is disabled
The main content of the view display gets greyed out if it is disabled, but the main link to the display does not. I think it makes sense to carry that through to the display links at the top like this:

Screen Shot 2020-01-24 at 11 27 48 AM

---

~~PR: https://github.com/backdrop/backdrop/pull/3049~~
Alt PR: https://github.com/backdrop/backdrop/pull/3058

NormPlum:
Fixes https://github.com/backdrop/backdrop-issues/issues/4293

### DIFF
--- a/core/modules/views_ui/views_ui.admin.inc
+++ b/core/modules/views_ui/views_ui.admin.inc
@@ -1534,7 +1534,7 @@ function views_ui_edit_page_display_tabs($view, $display_id = NULL) {
       $tabs[$id]['#link']['localized_options']['attributes']['class'][] = 'views-display-deleted-link';
     }
     if (isset($display->display_options['enabled']) && !$display->display_options['enabled']) {
-      $tabs[$id]['#link']['title'] .= ' (' . t('disabled') . ')';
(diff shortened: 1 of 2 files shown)


core/modules/views_ui/views_ui.admin.inc
```diff
@@ -1534,6 +1534,7 @@ function views_ui_edit_page_display_tabs($view, $display_id = NULL) {
       $tabs[$id]['#link']['localized_options']['attributes']['class'][] = 'views-display-deleted-link';
     }
     if (isset($display->display_options['enabled']) && !$display->display_options['enabled']) {
+      $tabs[$id]['#link']['title'] .= ' (' . t('disabled') . ')';
```
argiepiano: There are many examples in core where the parenthesis is included in the translated string, and it should do so here. See `theme_menu_local_task()` for example.

NormPlum: Thanks, I didn't realise that. I've fixed it.

PR #5243, opened 2026-04-17
Issue #6280: Switch to separate dedicated cspell repository.

### LINKED ISSUE #6280: CSpell: move the backdrop.dic file to a dedicated repo under https://github.com/backdrop-ops
This is a follow-up to recent issues relevant to cspell; mostly related to (or triggered by) #6255.

Instead of adding words to a PR against core, and then having to wait for one of our core committers to become available, we should follow the same practice as with https://github.com/backdrop-ops/phpcs and move the dictionary to a dedicated https://github.com/backdrop-ops/cspell repo. That way, we can establish a team of maintainers for that repo (I offer myself), and things can move independently of core (and faster too I suspect).

We can create a new GitHub group for the maintainers of that repo (I propose to call it `spelling-bees` 😅 ), and when a PR trips on one or more unknown words, people can either ping us in Zulip or in GitHub or file an issue in the queue of that repo, and we can add the words. No need for core commits, nor for the PR to be rebased - just close it and re-open after the word(s) have been added to the remote/independent repo, and the test will automatically pick the new additions. Profit! 😉 

This issue here is because we'll need a PR against core, to update the https://github.com/backdrop/backdrop/blob/1.x/.cspell/cspell.json file and point to the repo instead of a "local" `backdrop.dic` file (same as we are already doing for the drupal dictionary).

So basically change this:
```json
  "dictionaryDefinitions": [
    {
      "name": "backdrop",
      "path": "backdrop.dic"
    },
    {
      "name": "drupal",
      "path": "https://git.drupalcode.org/project/drupal/-/raw/HEAD/core/misc/cspell/dictionary.txt"
    }
  ],
```

...to this instead:
```json
  "dictionaryDefinitions": [
    {
      "name": "backdrop",
      "path": "https://github.com/backdrop-ops/cspell/blob/main/backdrop.dic"
    },
    {
      "name": "drupal",
      "path": "https://git.drupalcode.org/project/drupal/-/raw/HEAD/core/misc/cspell/dictionary.txt"
    }
  ],
```

quicksketch:
Fixes https://github.com/backdrop/backdrop-issues/issues/6280


.github/workflows/phpcs-lint.yml
```diff
@@ -20,7 +20,7 @@ jobs:
         uses: actions/checkout@v4
         with:
           repository: backdrop-ops/phpcs
-          ref: 1.2.0
+          ref: 1.3.0
```
quicksketch, line 23: Fixes issue with `cspell:ignore` lines causing a PHPCS error. See https://github.com/backdrop-ops/phpcs/issues/16.

indigoxela: Oh, this was the issue triggering the phpcs update. :grin: 
See also: https://github.com/backdrop/backdrop-issues/issues/7088

.github/workflows/phpcs-lint.yml
```diff
@@ -42,6 +42,11 @@ jobs:
     steps:
     - uses: actions/checkout@v4
       name: Checkout code
+    - uses: actions/checkout@v4
```
indigoxela: checkout@v4 is already deprecated. 😉

.github/workflows/phpcs-lint.yml
quicksketch: Thanks, good to fix that in the right place. :smile:

quicksketch:
Merged into 1.x and 1.33.x. Thank you @indigoxela for reviewing!

PR #5244, opened 2026-04-17
Issue #6280: Fixing spelling of processes.

### LINKED ISSUE #6280: CSpell: move the backdrop.dic file to a dedicated repo under https://github.com/backdrop-ops

quicksketch:
Fixes https://github.com/backdrop/backdrop-issues/issues/6280

Related, one last spelling issue.

### DIFF
--- a/.github/workflows/simpletest.yml
+++ b/.github/workflows/simpletest.yml
@@ -47,7 +47,7 @@ jobs:
           free -m
           echo Get disk usage
           df -h
-          echo Get running procs by user
+          echo Get running processes by user
           pgrep -au runner
 
       - name: Save Test Results


quicksketch:
Merged into 1.x and 1.33.x to fix this minor spelling error.

PR #5251, opened 2026-04-22
Issue #7107: Account for possible null value in _layout_standardize_template_type().

### LINKED ISSUE #7107: Account for NULL value on layout_template_type
> Got this on a test site when my browser restarted after a Chrome update:
> Image
> 
> A `dpm()` said my layout had a `layout_template_type()` of `NULL`. Interestingly happened on the last path I had accessed, and I couldnt flush caches because the cache flushing link seemed to be the only other page that gave this error! After commenting out the code in `layout_load_multiple_by_path()` that retrieved a layout from cache (`if ($cache = cache('layout_path')->get($path))`) I was able to flush caches and all was well.
> 
> Maybe it would have gone away with `update.php` but I didnt try that until I could hunt down the error, but if that url was also unreachable a user would have some problems getting rid of this error.  

 _Originally posted by @docwilmot in [#6808](https://github.com/backdrop/backdrop-issues/issues/6808#issuecomment-4284996221)_

laryn:
Fixes https://github.com/backdrop/backdrop-issues/issues/7107

### DIFF
--- a/core/modules/layout/layout.module
+++ b/core/modules/layout/layout.module
@@ -2842,10 +2842,13 @@ function layout_field_delete_instance($instance) {
  * use them as is, otherwise we'll ensure it starts with 'custom_' before
  * sending it back.
  *
- * @param string $layout_template_type
+ * @param ?string $layout_template_type
  *   The custom template type machine name to process, passed by reference.
  */
-function _layout_standardize_template_type(string &$layout_template_type = 'flexible') {
+function _layout_standardize_template_type(?string &$layout_template_type = 'flexible') {
+  if (empty($layout_template_type)) {
+    $layout_template_type = 'flexible';
+  }
   $reserved_types = array(
     'layout',
     'flexible',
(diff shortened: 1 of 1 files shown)


core/modules/layout/layout.module
```diff
@@ -2842,10 +2842,13 @@ function layout_field_delete_instance($instance) {
  * use them as is, otherwise we'll ensure it starts with 'custom_' before
  * sending it back.
  *
- * @param string $layout_template_type
+ * @param ?string $layout_template_type
  *   The custom template type machine name to process, passed by reference.
```
quicksketch: We've been using `|null` when a parameter is optional.

It would be worth explaining what happens when NULL is used as well.

```suggestion
 * @param string|null $layout_template_type
 *   The custom template type machine name to process, passed by reference. If
 *   NULL is passed in, the default type of "flexible" will be used.
```

PR #5252, opened 2026-04-26
Icon picker min

### LINKED ISSUE #7083: Icon browser/picker in core

docwilmot:
Fixes https://github.com/backdrop/backdrop-issues/issues/7083

Alternative to https://github.com/backdrop/backdrop/pull/5237 but removes browser modal and link.


core/includes/icon.inc
```diff
@@ -342,3 +344,159 @@ function theme_icon(array $variables) {
...
+        'provider_detail' => t('@module (module)', ['@module' => $icon['module']]),
+        'provider_key' => $name,
+        'name' => $icon_key,
+      ) + icon_get_style($icon_key);
+    }
+    // Icons provided by themes. Only look in the active theme or its base
+    // themes since _icon_from_theme() only checks these.
+    $themes = list_themes();
+    $active_theme = $GLOBALS['theme_key'];
+    $current_themes = array($active_theme);
```
indigoxela, line 413: That way (theme_key) we'd only get the admin theme's icons on admin/config/media/icon-browser, but not the default theme's icons. Is that by intention?

In dialogs it may make sense, but on admin/config/media/icon-browser we might want to show really all icons defined anywhere. Not sure.

core/includes/icon.inc
```diff
@@ -342,3 +344,159 @@ function theme_icon(array $variables) {
...
+ *    - All icons, if no $icon_name was provided
+ *    - The single result if an $icon_name was provided, or
+ *    - An empty array if no icon was found for a provided $icon_name.
+ */
+function icon_get_all_details($icon_name = '') {
+  $icons = &backdrop_static(__FUNCTION__, array());
+
+  if (empty($icons)) {
+    // Icons provided by Backdrop core.
+    $core_icons = scandir(CORE_ICON_BROWSER_PATH);
```
indigoxela: As you're doing a backdrop_sort() below, anyway, we could add SCANDIR_SORT_NONE as param - for a small performance gain. I didn't try or benchmark it.

See https://www.php.net/manual/en/function.scandir.php

core/includes/icon.inc
```diff
@@ -342,3 +344,159 @@ function theme_icon(array $variables) {
...
+ * @param string $icon_name
+ *   The icon name to load details for. If empty, load all icons.
+ * @return array
+ *   An array of icon details, keyed by icon name:
+ *    - All icons, if no $icon_name was provided
+ *    - The single result if an $icon_name was provided, or
+ *    - An empty array if no icon was found for a provided $icon_name.
+ */
+function icon_get_all_details($icon_name = '') {
+  $icons = &backdrop_static(__FUNCTION__, array());
```
indigoxela, line 378: Wouldn't that static cache mean, that if the first call on a request is with name, that all subsequent calls also get only that one icon? Didn't try, just a quick idea.

docwilmot: No, the `$icons` variable stores all icons.

core/includes/icon.inc
docwilmot: Good plan. Done.

core/includes/icon.inc
docwilmot: How useful is it to show that a theme you're not using has icons though? Maybe we could show all icons from the current default (active _and_ admin) themes? Or perhaps mark icons that are not in the default themes somehow in the UI (color them, add alt text, an asterix, etc?). Or just add the theme name below the icon name in the UI?
To be clear, an icon provided by a theme wont display in another theme because `icon()` only checks for icons in the current theme.

docwilmot: Icon Browser Module shows theme-provided icons as an error if not in the current active theme.

core/includes/icon.inc
```diff
@@ -342,3 +344,171 @@ function theme_icon(array $variables) {
...
+      if (is_dir($theme_directory)) {
+        $theme_icons = scandir($theme_directory);
+        if (!empty($theme_icons)) {
+          foreach ($theme_icons as $name => $icon) {
+            if (substr($icon, -4, 4) == '.svg') {
+              $icon_key = substr($icon, 0, -4);
+              $icons[$icon_key] = array(
+                'provider' => 'theme',
+                'provider_detail' => t('@theme (theme*) <br><br><em>*Theme-provided icons may only be available when the theme is active.</em>', ['@theme' => $theme]),
+                'provider_key' => $icon_key,
```
quicksketch: Instead of "provider", which sounds to me like it would be the name of the icon provider/artist (i.e. `material`, `font-awesome`, `phosphor`, etc), I think the following keys would align better with Backdrop terminology:

- `project_type` (core, module, or theme)
- `project_name` (seven, basis)
- `project_description` (the description above)

core/includes/icon.inc
indigoxela: > Icon Browser Module shows theme-provided icons as an error if not in the current active theme.

That makes sense, but there's one culprit: If the setting on admin/appearance for "Use the administration theme when editing or creating content" is checked, the admin theme is the active one when editing nodes. But the display would use the default theme, of course. Then we provide icons to select that won't be available for display, but potentially suppress some that would be...

We don't have to discuss that now, as "insert" functionality isn't part of this PR. It's just about future problems that we might run into, if we add some "insert" (either for fields or WYSIWYG - the latter needs extra care, anyway).

core/includes/icon.inc
docwilmot: Good suggestion. Done.

core/includes/icon.inc
docwilmot: I've reverted to listing all icons from all themes, with an error icon, and a page message if there are icons not from the current theme. A `backdrop_set_message()` might be a bit heavy-handed, but it allows to say clearly what the problem is.

docwilmot:
One unrelated fail.

core/modules/system/config/system.core.json
```diff
@@ -39,6 +39,7 @@
     "file_transliterate_uploads_display_name": 0,
     "filter_fallback_format": "plain_text",
     "form_cache_expiration": 21600,
+    "icon_browser_icons_per_page": 110,
```
indigoxela, line 42: This needs an update hook. Otherwise we end up with `DivisionByZeroError: Division by zero in pager_default_initialize() ... pager.inc` 😉

core/includes/icon.inc
indigoxela: I see. That works, but I'm not sure... The frontend theme's icons still wouldn't show up, when editing a node on the admin theme, so the benefit of that change is... pretty limited.
Same would apply to blocks in layouts, where the active theme is always the admin theme, but the display theme could be either backend or frontend. :grimacing: I hope, I didn't open a can of worms here.

core/modules/system/system.module
```diff
@@ -4905,3 +4900,164 @@ function system_transliterate_machine_name($string, $options = array()) {
...
+  $form['icon_filters'] = array(
+    '#type' => 'container',
+    '#attributes' => array(
+      'class' => array('icon-browser-filters-wrapper'),
+    ),
+  );
+  if (count($available_project_types) > 1) {
+    $form['icon_filters']['project_type'] = array(
+      '#type' => 'select',
+      '#title' => t('project_type'),
```
indigoxela: That label doesn't look good. 😉 Should probably be "Project type".

core/modules/system/system.module
```diff
@@ -4905,3 +4900,164 @@ function system_transliterate_machine_name($string, $options = array()) {
...
+      '#options' => array(
+        'any' => t('Any'),
+      ) + $project_type_options,
+      '#default_value' => $project_type,
```
indigoxela, line 5012: Seems like the variable `$project_type` is only set on form submit, that's why the default value's lost when using the pager. Or maybe it's something else...

core/modules/system/system.module
```diff
@@ -4905,3 +4900,164 @@ function system_transliterate_machine_name($string, $options = array()) {
...
+    );
+  }
+  $form['icon_filters']['style'] = array(
+    '#type' => 'select',
+    '#title' => t('Style'),
+    '#options' => array(
+      'any' => t('Any'),
+    ) + $style_options,
+    '#default_value' => $style,
```
indigoxela, line 5021: Similar to above, the style is lost when using the pager.

core/includes/icon.inc
docwilmot: Yes, I wonder if it would be better to modify `_icon_from_theme()` to check all enabled themes, or to check both default themes, rather than just the active front end theme. That would solve all this.

core/modules/system/config/system.core.json
docwilmot: Done

core/modules/system/system.module
docwilmot: Fixed

core/modules/system/system.module
docwilmot: Fixed. The form state rebuild was wiping out the `$_GET`. The rebuild was probably needed for the AJAXified dialog, but can rework that when the time comes.

core/modules/system/system.module
docwilmot: Same fixed.

core/includes/icon.inc
indigoxela: > better to modify _icon_from_theme()...

That sounds ambitious for 1.34.
Maybe just go back to only show the current theme's icons for now (no errors for other)? And discuss possible solutions for themes in follow-ups? I mean, Seven has 2 icons, Gin has a handful - is the effort even worth it, compared to the gain?

docwilmot: Actually, rethinking, I think it may be simple as not calling `icon()` to render the images in the browser. We would already know these icons exist, and we have the path, so calling `theme_icon()` instead shouldn't be any problem. That way we can show, as would be required, the icons from the admin and front-end themes.

indigoxela: Almost...

For icons provided via hook_icon_info() with defaults (as documented), the "directory" key might be missing.
Hence we'd end up with `Warning: Undefined array key "directory" in theme_system_icon_browser_form_icons(...`  with modules, which don't define the path for each icon - that's something the *icon* function takes care of via `icon_get_path()`. And it also takes care of attributes.

But the idea definitely gets us a step forward.

indigoxela: FTR: I tested locally with module font_awesome and a custom module implementing the hook. Their icons won't show up until I add icon_get_path().

core/modules/system/system.theme.inc
```diff
@@ -724,3 +724,53 @@ function theme_label_machine_name($variables) {
...
+    foreach ($icons as $icon) {
+      if (!empty($icon['theme_error'])) {
+        $theme_error_icons[$icon['name']] = $icon['theme_error'];
+      }
+
+      $icon_classes = isset($icon['theme_error']) ? 'icon-wrapper error' : 'icon-wrapper';
+      $output .= '<div class="' . $icon_classes . '" data-icon-name="' . $icon['project_icon_key'] . '">';
+      $rendered_icon = theme('icon', array(
+        'name' => $icon['project_icon_key'],
+        'path' => $icon['directory'] . '/' . $icon['name'] . '.svg',
```
indigoxela, line 747: We'd have to run icon_get_path() for modules, this currently only works for core, but

```php
'path' => icon_get_path($icon['project_icon_key']),
```
also works for modules.

core/includes/icon.inc
docwilmot: Thanks. I've added the directory key for module-provided icons.

core/modules/system/system.theme.inc
docwilmot: Instead, I've added the directory key for module-provided icons.

core/includes/icon.inc
indigoxela: Even better! :+1:

core/includes/icon.inc
```diff
@@ -342,3 +344,194 @@ function theme_icon(array $variables) {
...
+ * @return array
+ *   An array of icon details, keyed by icon name:
+ *    - All icons, if no $icon_name was provided
+ *    - The single result if an $icon_name was provided, or
+ *    - An empty array if no icon was found for a provided $icon_name.
+ */
+function icon_get_all_details($icon_name = '') {
+  // Skip building the whole array if one icon was requested and the icon is
+  // present in the most likely location (core).
+  if (isset($icon_name) && _icon_from_core($icon_name)) {
```
indigoxela: Is this supposed to be `!empty()` ? Because `$icon_name` will always be set - even if only as empty string.

core/modules/system/system.module
```diff
@@ -812,6 +817,14 @@ function system_menu() {
     'file' => 'system.admin.inc',
   );
 
+  // Icon browser page.
+  $items['admin/config/media/icon-browser'] = array(
+    'title' => 'Icon browser',
```
herbdool: A short description would be helpful to appear with all menu items in Media. It's currently the only item without a description.

core/modules/system/system.install
```diff
@@ -3883,6 +3883,17 @@ function system_update_1104() {
   }
 }
 
+/**
+ * Set icon browser default number of icons per page.
+ */
+function system_update_1105() {
```
indigoxela, line 3889: Oops, html5_upload was faster. The next update hook number would be 1106.

Try `function system_update_1106() {` as long as it's available to avoid the git conflict. 😉

core/modules/system/system.module
```diff
@@ -812,6 +817,14 @@ function system_menu() {
     'file' => 'system.admin.inc',
   );
 
+  // Icon browser page.
+  $items['admin/config/media/icon-browser'] = array(
```
herbdool:
```suggestion
  $items['admin/reports/icons/list'] = array(
```

There's general agreement that the browser should be in Reports. I think it should be like `admin/reports/fields/list` at least in terms of the menu hierarchy.

core/includes/icon.inc
docwilmot: Yep, that's better. Changed, thanks.

core/modules/system/system.module
docwilmot: Done.

core/modules/system/system.install
docwilmot: Thanks, done.

core/modules/system/system.module
docwilmot: I don't know that its possible to have a secondary menu item without a default local task. Meaning we'd need a page at `admin/reports/icons` first before we could do a `admin/reports/icons/list` AFAIK. I've moved the page to `admin/reports/icon-browser`.

indigoxela: Personally, I'm just fine with admin/reports/icon-browser :+1:

quicksketch:
Merged into 1.x for 1.34.0. Great job @docwilmot!!

PR #5254, opened 2026-04-27
Issue #7075: add options for URL history for ajax views

### LINKED ISSUE #7075: Incorporate Views AJAX History

quicksketch:
Fixes https://github.com/backdrop/backdrop-issues/issues/7075

Replaces https://github.com/backdrop/backdrop/pull/5223.

@herbdool I tried to file a PR against your branch but for some reason GitHub is not seeing your fork as a valid repository. Seems like a GitHub issue. So instead, this is a PR against core.


core/modules/views/js/ajax_view.js
```diff
@@ -4,6 +4,20 @@
  */
 (function ($) {
 
+"use strict";
+
+// Keep to check if there are extra parameters in the original URL.
```
docwilmot: Is this a typo? I'm not understanding what "Keep to check if..." means

argiepiano: I think this refers to the variable `original`, which "keeps" (or "stores") the original path to use it later.

core/modules/views/js/base.js
```diff
@@ -74,4 +106,141 @@ Backdrop.Views.getPath = function (href) {
...
+ * @param url
+ *   String with the current URL to be cleaned up.
+ */
+Backdrop.Views.addState = function (options, url) {
+  // The data in the history state must be serializable.
+  var historyOptions = $.extend({}, options);
+
+  // Store the actual view's dom id.
+  Backdrop.settings.lastViewDomID = options.data.view_dom_id;
+  $(window).unbind('popstate', Backdrop.Views.loadView);
```
argiepiano: unbind is deprecated.

core/modules/views/js/base.js
```diff
@@ -74,4 +106,141 @@ Backdrop.Views.getPath = function (href) {
...
+  history.pushState(Backdrop.Views.cleanStateForHistory(historyOptions), document.title, Backdrop.Views.cleanURL(url, options.data));
+  $(window).bind('popstate', Backdrop.Views.loadView);
```
argiepiano: bind is deprecated.

core/modules/views/js/ajax_view.js
```diff
@@ -146,6 +170,159 @@ Backdrop.views.ajaxView.prototype.attachPagerLinkAjax = function(id, link) {
...
+      // don't need to go through all this processing.
+      if ($(element).hasClass('ajax-history-trigger')) {
+        return;
+      }
+    }
+
+    // Check that we handle a click on a link, not a form submission.
+    if (options.data.view_name && element && $(element).is('a')) {
+      // Strip the view base path so it isn't treated as a parameter.
+      let params = new URLSearchParams($(element).attr('href').replace('/' + options.data.view_base_path + '?', ''));
```
argiepiano:
```suggestion
      let params = new URLSearchParams($(element).attr('href').replace('/' + options.data.view_path + '?', ''));
```

herbdool: This does work, but only on a monolingual site. I tested it on a fresh site where I added a second language (made `admin/content` use ajax history). The default looks fine `admin/content?page=2`. Navigating to `nl/admin/content` will produce: `nl/admin/content?%2Fnlpage=2`. So it's not capturing the language. This is something I added as a fix but obviously still needs work.

argiepiano: @herbdool do you mean my suggested fix doesn't work on multilingual sites? C an you clarify if you are commenting on the original code or my suggested fix?

herbdool: @argiepiano your fix.

argiepiano: Thanks. Yes, I can see how that happens. We need to find a way to remove the part of the path between the hostname and the query character, verbatim, for ajaxified links. Shouldn't be difficult. We have the var `original` that contains the full path, or could use the `window.location` properties.

herbdool: I was poking around and on the a element.search it has `?/nl/admin/content%3Fpage=2&page=1` . We could strip everything between `?` and `&` I think. Maybe a future issue would be to figure out why it's getting added in the first place.

argiepiano: I think it's clear why it's getting added. The pager link (button) itself contains the relative path, as in `<a href="/admin/content?page=1">`.

herbdool: This works for me:

```suggestion
      let params = new URLSearchParams($(element).attr('href').replace($(element).get(0).pathname + '?', ''));
```

core/modules/views/js/base.js
herbdool:
```suggestion
  $(window).off('popstate', Backdrop.Views.loadView);
```

core/modules/views/js/base.js
herbdool:
```suggestion
  $(window).on('popstate', Backdrop.Views.loadView);
```

core/modules/views/js/base.js
```diff
@@ -74,4 +106,141 @@ Backdrop.Views.getPath = function (href) {
...
+  var query = {};
+  $.map(queryString.split('&'), function (val) {
+    var s = val.split('=');
+    query[s[0]] = s[1];
+  });
+  return query;
+};
+
+/**
+ * Unbind 'popstate' when adding a new state to avoid an infinite loop.
```
herbdool:
```suggestion
 * Remove 'popstate' handler when adding a new state to avoid an infinite loop.
```

core/modules/views/js/ajax_view.js
herbdool: I think because this was originally my PR and we've now got other folks to help test/review, I'll add the changes and get feedback quicker.

core/modules/views/js/ajax_view.js
```diff
@@ -146,6 +170,159 @@ Backdrop.views.ajaxView.prototype.attachPagerLinkAjax = function(id, link) {
...
+ *
+ * Rather than overriding Backdrop.ajax.prototype methods globally, this function
+ * sets the Views-specific behavior only on ajax objects created by ajax_view.js.
+ *
+ * @param {Backdrop.ajax} ajaxInstance
+ *   The ajax instance to configure with Views-specific behavior.
+ */
+Backdrop.views.ajaxView.prototype.applyAjaxOverrides = function (ajaxInstance) {
+  ajaxInstance.beforeSerialize = function (element, options) {
+    if (options.data.enable_ajax_history && options.data.view_name) {
```
herbdool:
```suggestion
    let enable_ajax_history = !!options.data.enable_ajax_history;
    if (enable_ajax_history && options.data.view_name) {
```

core/modules/views/js/ajax_view.js
```diff
@@ -146,6 +170,159 @@ Backdrop.views.ajaxView.prototype.attachPagerLinkAjax = function(id, link) {
...
+      // If restoring a previous state the dummy element will have this class,
+      // don't need to go through all this processing.
+      if ($(element).hasClass('ajax-history-trigger')) {
+        return;
+      }
+    }
+
+    // Check that we handle a click on a link, not a form submission.
+    if (options.data.view_name && element && $(element).is('a')) {
```
herbdool:
```suggestion
    if (enable_ajax_history && options.data.view_name && element && $(element).is('a')) {
```

core/modules/views/js/ajax_view.js
```diff
@@ -146,6 +170,159 @@ Backdrop.views.ajaxView.prototype.attachPagerLinkAjax = function(id, link) {
...
+      }
+      Backdrop.Views.addState(options, '?' + params.toString());
+    }
+
+    // Call the original Backdrop method with the right context.
+    Backdrop.ajax.prototype.beforeSerialize.apply(this, arguments);
+  };
+
+  ajaxInstance.beforeSubmit = function (form_values, element, options) {
+    if (options.data.enable_ajax_history && options && options.data && options.data.view_name) {
```
herbdool:
```suggestion
    if (!!options.data.enable_ajax_history && options && options.data && options.data.view_name) {
```

core/modules/views/js/ajax_view.js
```diff
@@ -146,6 +170,159 @@ Backdrop.views.ajaxView.prototype.attachPagerLinkAjax = function(id, link) {
...
+    }
+
+    // Call the original Backdrop method with the right context.
+    Backdrop.ajax.prototype.beforeSubmit.apply(this, arguments);
+  };
+
+  ajaxInstance.beforeSend = function (jqXHR, options) {
+    var data = (typeof options.data === 'string') ? Backdrop.Views.parseQuery(options.data) : {};
+
+    if (data.enable_ajax_history && data.view_name && options.type !== 'GET') {
```
herbdool:
```suggestion
    if (!!data.enable_ajax_history && data.view_name && options.type !== 'GET') {
```

core/modules/views/js/ajax_view.js
```diff
@@ -146,6 +170,159 @@ Backdrop.views.ajaxView.prototype.attachPagerLinkAjax = function(id, link) {
...
+      options.url = ajaxUrl + '?' + $.param(ajaxQuery);
+      Backdrop.Views.addState(options, url);
+    }
+
+    // Call the original Backdrop method with the right context.
+    Backdrop.ajax.prototype.beforeSubmit.apply(this, arguments);
+  };
+
+  ajaxInstance.beforeSend = function (jqXHR, options) {
+    var data = (typeof options.data === 'string') ? Backdrop.Views.parseQuery(options.data) : {};
```
herbdool:
```suggestion
    var data = (typeof options.data === 'string') ? Backdrop.Views.parseQueryString(options.data) : {};
```

core/modules/views/views.install
```diff
@@ -213,6 +213,29 @@ function views_update_1008() {
...
+    $config = config($view_name);
+    $use_ajax_default = $config->get('display.default.display_options.use_ajax');
+    $displays = $config->get('display');
+    foreach ($displays as $display_name => $display_config) {
+      if (isset($display_config['display_options']['use_ajax']) && $display_config['display_options']['use_ajax'] == FALSE) {
+        continue;
+      }
+      if ($use_ajax_default || !empty($display_config['display_options']['use_ajax'])) {
+        $config->set('display.' . $display_name . '.display_options.enable_ajax_history', TRUE);
+        $config->set('display.' . $display_name . '.display_options.exclude_ajax_args', '');
```
quicksketch: Seems like `enable_ajax_history` should actually be set to `FALSE` to maintain the current behavior on upgrades.

herbdool: I think I meant to enable it if they had the views_ajax_history module enabled. That module enabled it globally. 

Speaking of which maybe we disable that module in an update hook?

quicksketch: Yeah and it should be added to `backdrop_merged_modules()`.

core/modules/views/js/ajax_view.js
quicksketch: I reworded this inline comment. I agree it was confusing.

core/modules/views/views.install
quicksketch: Added to the existing update hook. I set the default based on whether `views_ajax_history` module was present.

PR #5257, opened 2026-04-30
Issue #7112: Fixes unconstrained join in comment module.

### LINKED ISSUE #7112: comment_update_1009() unconstrained JOIN may crash database when running update.php
## Description of the bug

A query introduced in the latest security release lacks a join condition (ON clause) between the {comment} and {node} tables. This results in a Cartesian Product (Cross Join), causing the database to attempt to process Rows (comment) X Rows (node) results. For sites with significant content, this causes database timeouts and prevents the update from completing.

Because this is a security release, the bug is particularly dangerous as it forces users to choose between a vulnerable site or a crashed/un-updatable site.

## Existing Code

```
/**
 * Enable comments on types with existing comments for backwards-compatibility.
 */
function comment_update_1009() {
  // Comments can now be disabled per type, but prior to Backdrop 1.33.0,
  // they were always enabled on all types.
  $count = 0;
  foreach (config_get_names_with_prefix('node.type.') as $config_name) {
    $config = config($config_name);
    $node_type = $config->get('type');
    $default_setting = $config->get('settings.comment_default');
    $comment_count = (int) db_query('SELECT COUNT(*) FROM {comment} c INNER JOIN {node} n WHERE n.type = :type', array(':type' => $node_type))->fetchField();
    // Constant avoided if the module is not loaded, COMMENT_NODE_OPEN == 2.
    if ($default_setting == 2 || $comment_count > 0) {
      $config->set('settings.comment_enabled', TRUE);
      $config->save();
      $count++;
    }
  }

  if ($count) {
    return format_plural($count, 'Comments enabled on @count content type.', 'Comments enabled on @count content types.');
  }
}
```

## Problematic Query

**File:** core/modules/comment/comment.install
**Function:** comment_update_1009()
**Line:** 532

` $comment_count = (int) db_query('SELECT COUNT(*) FROM {comment} c INNER JOIN {node} n WHERE n.type = :type', array(':type' => $node_type))->fetchField();`

## Technical Impact
**Mathematical Scaling:** On a site with 10,000 comments and 5,000 nodes, MySQL attempts to generate 50,000,000 rows before filtering.

**Resource Exhaustion:** This leads to CPU spikes, memory exhaustion (OOM), and MySQL server has gone away errors.

**Update Failure:** Because this occurs during the update process (or on a critical admin path), sites with large datasets are effectively blocked from applying this security release

## Proposed Solution

Add the missing join condition to the INNER JOIN to leverage the table indexes:

`$comment_count = (int) db_query('SELECT COUNT(*) FROM {comment} c INNER JOIN {node} n ON c.nid = n.nid WHERE n.type = :type', array(':type' => $node_type))->fetchField();`

--

## Additional information

- Backdrop CMS version: 1.32.1

rbargerhuff:
Fixes https://github.com/backdrop/backdrop-issues/issues/7112

### DIFF
--- a/core/modules/comment/comment.install
+++ b/core/modules/comment/comment.install
@@ -517,7 +517,6 @@ function comment_update_1008() {
   }
 }
 
-
 /**
  * Enable comments on types with existing comments for backwards-compatibility.
  */
@@ -529,7 +528,7 @@ function comment_update_1009() {
     $config = config($config_name);
     $node_type = $config->get('type');
     $default_setting = $config->get('settings.comment_default');
-    $comment_count = (int) db_query('SELECT COUNT(*) FROM {comment} c INNER JOIN {node} n WHERE n.type = :type', array(':type' => $node_type))->fetchField();
+    $comment_count = (int) db_query('SELECT COUNT(*) FROM {comment} c INNER JOIN {node} n ON c.nid = n.nid WHERE n.type = :type', array(':type' => $node_type))->fetchField();
     // Constant avoided if the module is not loaded, COMMENT_NODE_OPEN == 2.
     if ($default_setting == 2 || $comment_count > 0) {
       $config->set('settings.comment_enabled', TRUE);


quicksketch:
Merged into 1.x and 1.33.x. Thanks @rbargerhuff!

PR #5258, opened 2026-05-01
Issue #7114: only save views config if changed

### LINKED ISSUE #7114: Views ajax history: only save the Views config if something changes
Follow-up to https://github.com/backdrop/backdrop-issues/issues/7075

Only save the Views config file if something changes.

herbdool:
Fixes https://github.com/backdrop/backdrop-issues/issues/7114

### DIFF
--- a/core/modules/views/views.install
+++ b/core/modules/views/views.install
@@ -220,8 +220,8 @@ function views_update_1008() {
 function views_update_1009() {
   $views = config_get_names_with_prefix('views.view');
   $views_ajax_history = module_exists('views_ajax_history');
-  $has_changes = FALSE;
   foreach ($views as $view_name) {
+    $has_changes = FALSE;
     $config = config($view_name);
     $use_ajax_default = $config->get('display.default.display_options.use_ajax');
     $displays = $config->get('display');
--- a/core/modules/views/views.install
+++ b/core/modules/views/views.install
@@ -214,11 +214,13 @@ function views_update_1008() {
 }
 
 /**
- * Enable Ajax History for all Views using AJAX.
+ * Enable ajax history for all Views that used Views Ajax History contrib
+ * module, and disable contrib module.
  */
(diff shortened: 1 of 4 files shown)


core/modules/views/views.install
```diff
@@ -214,11 +214,13 @@ function views_update_1008() {
...
- * Enable Ajax History for all Views using AJAX.
+ * Enable ajax history for all Views that used Views Ajax History contrib
+ * module, and disable contrib module.
  */
 function views_update_1009() {
   $views = config_get_names_with_prefix('views.view');
   $views_ajax_history = module_exists('views_ajax_history');
+  $has_changes = FALSE;
   foreach ($views as $view_name) {
     $config = config($view_name);
```
indigoxela, line 225: You got to reset this has_changes for every view, as otherwise the first view having ajax enabled leads to all following views to get updated.
Or better, move the $has_changes = FALSE; inside the outer loop.

Edit: resolved 👍

core/modules/views/views.install
```diff
@@ -231,16 +233,27 @@ function views_update_1009() {
...
+      $config->save();
+    }
   }
 
   // Remove the entry for the views_ajax_history module in the system table.
   // This will "disable" the module if present. No other configuration was
   // present for this module.
-  db_query("DELETE FROM {system} WHERE name = 'views_ajax_history' AND type = 'module'");
+  $module = db_query("SELECT name FROM {system} WHERE name = 'views_ajax_history' AND type = 'module'")->fetchField();
+  if ($module) {
```
quicksketch, line 249: Note this is intentional that we don't use `module_exists('views_ajax_history');`. `module_exists()` actually checks that the module not only exists but is enabled. This checks for a system table entry even the module is disabled.

quicksketch:
Merged into 1.x for 1.34.0. Thank you @herbdool!

PR #5260, opened 2026-05-04
Issue #7118: Fix inserting multiple images to field from image library

### LINKED ISSUE #7118: Follow-up: Selecting multiple images from library won't work
## Description of the bug

Follow-up to Issue #440: Support "Multiple" attribute on file uploads. 

Image fields with multiupload enabled (cardinality), allow to select multiple from library, but on insert only the first one is added to field values.

## Steps To Reproduce

1. Create an image field that allows multiple images
2. Make sure "Enable image browser" is clicked
3. Open a node form with that field
4. Open the image library ("Select existing file")
5. Select multiple images
6. Click on the insert button

## Actual behavior

Although it was possible to select multiple images, only *one* image is actually inserted.

Inserting one by one still works fine.

## Expected behavior

All selected images inserted to the field (of course, according to cardinality limits).

## Additional information

- Backdrop CMS version: latest dev as of [3f77eaea0a0488c4f824de1f0bbf8e31041e4c44](https://github.com/backdrop/backdrop/commit/3f77eaea0a0488c4f824de1f0bbf8e31041e4c44).

The selection part works, and that also respects limits, only the *insert* part doesn't work.
Unrelated file upload also works fine, BTW.

I guess, we have to tweak [dialogCloseEvent](https://github.com/backdrop/backdrop/blob/1.x/core/modules/file/js/file.js#L232), but don't have the time right now to dig deeper.

indigoxela:
Fixes https://github.com/backdrop/backdrop-issues/issues/7118

### DIFF
--- a/core/modules/file/file.module
+++ b/core/modules/file/file.module
@@ -1897,6 +1897,8 @@ function file_managed_file_browser_open($form, $form_state) {
     'file' => array(
       'browser' => array(
         'fieldCardinality' => $field_cardinality,
+        // @todo It's probably a bad idea to derive the file count from delta.
+        'existingFilesCount' => end($parents),
         'selectedFid' => NULL,
         'currentFidElement' => '[name="' . $element['fid']['#name'] . '"]',
       ),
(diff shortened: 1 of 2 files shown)


indigoxela:
Closing in favor of a new issue and pr after core changes.

PR #5261, opened 2026-05-04
Issue #7120: Make sure selectable event is not lost when using pager in dialog

### LINKED ISSUE #7120: Regression: Image browser items lose JS events after using the pager
## Description of the bug

Again, introduced by Issue #440: Support "Multiple" attribute on file uploads.

## Steps To Reproduce

1. A node with multi value image field
2. Make sure you have more images on that site, than fit on a single page of the image browser
3. Open the node form, open the dialog
4. Click on the pager (go to next page)
5. Try to select one of the images, try to insert

## Actual behavior

No image gets highlighted and clicking on "Insert" just closes the dialog, but no image gets inserted.

## Expected behavior

Insert should happen even after using the pager.

## Additional information

I belief, we struggled with similar problems before (JS events lost), but I'm not sure.
In #440 a different select approach has been added (jQuery UI), but those events only get attached once, so using the pager wipes them out. Caution, that's a bit of guessing.

Edit: what's actually lost is the event on the parent div (class ui-selectable missing from view-content, as view-content gets replaced). Happens, or does not happen, in dialogOpenEvent.

indigoxela:
Fixes https://github.com/backdrop/backdrop-issues/issues/7120

Turns out, jQuery UI provides what we need: A filter, so we don't have to use the immediate parent (which gets replaced on pager click).

### DIFF
--- a/core/modules/file/js/file.js
+++ b/core/modules/file/js/file.js
@@ -174,10 +174,10 @@ Backdrop.file = Backdrop.file || {
    */
   dialogOpenEvent: function(e, dialog, $element, settings) {
     var $browserContainer = $element.find(".file-browser");
-    var $viewContent = $element.find(".view-content");
     var fieldCardinality = parseInt(Backdrop.settings.file.browser.fieldCardinality);
     if (fieldCardinality != 1) {
-      $viewContent.selectable({
+      $browserContainer.selectable({
+        filter: '.image-library-choose-file',
         classes: {
           "ui-selected": "image-library-image-selected"
         },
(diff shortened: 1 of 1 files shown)


core/modules/file/js/file.js
```diff
@@ -174,13 +174,14 @@ Backdrop.file = Backdrop.file || {
    */
   dialogOpenEvent: function(e, dialog, $element, settings) {
     var $browserContainer = $element.find(".file-browser");
-    var $viewContent = $element.find(".view-content");
+    var fieldCardinality = parseInt(Backdrop.settings.file.browser.fieldCardinality);
```
quicksketch:
```suggestion
```

quicksketch:
Merged into 1.x for 1.34.0. Thank you @indigoxela!

PR #5262, opened 2026-05-05
Issue #7119: Fix TypeError when saving node with too many files attached to field

### LINKED ISSUE #7119: Regression: Multiupload when uploading more than allowed - TypeError when saving node
## Description of the bug

Validation in file_field_widget_multiple_count_validate() does not work properly.

It should prevent too many field items, if a restriction via field cardinality exists **on node (entity) save**. It doesn't. Instead it causes a fatal error.
Same problem when selecting more than allowed from image library.

## Steps To Reproduce

1. Attach an image field to a node
2. Set a limit on image count (field cardinality > 1, but < unlimited)
3. Open the node form and upload images, more than allowed

Alternative way to trigger:

1. Attach an image field to a node
2. Set a limit on image count (field cardinality > 1, but < unlimited)
3. Create a node with that amount of images (adhere max)
4. Go to admin/structure/types/manage/NODE_TYPE/fields/field_SOMETHING and *lower* the limit
5. Open that node form again and save without any changes

Another way to trigger:

1. Attach an image field to a node
2. Set a limit on image count (field cardinality > 1, but < unlimited), make sure the image browser's enabled
3. Open the node form
4. Open the image library dialog and add `limit - 1` images
5. Open the image lib again and add as many as possible (more than allowed in field, only until #7122 is resolved)

## Actual behavior

You *can* upload more images than field cardinality allows. They show up in the field list in that form.

_Now try to save the node._

```
TypeError: array_slice(): Argument #1 ($array) must be of type array, string given in array_slice() (line 1049 of ...core/modules/file/file.field.inc). 
```

## Expected behavior

I shouldn't even have come to that point (too many images), but that's a topic on its own.
Saving the node should definitely not end up with a TypeError. 😬 

## Additional information

- Backdrop CMS version: latest dev after Issue #440: Support "Multiple" attribute on file uploads.

Initially it thought, it's related to the dimension limit, but the actual problem is how form_state values get handled.

indigoxela:
Fixes https://github.com/backdrop/backdrop-issues/issues/7119

### DIFF
--- a/core/modules/file/file.field.inc
+++ b/core/modules/file/file.field.inc
@@ -1029,43 +1029,46 @@ function file_field_widget_value($element, $input, $form_state) {
 }
 
  /**
- * Validation callback for upload element on file widget. Checks if user has
- * uploaded more files than allowed.
+ * Validation callback for managed file element on file field widgets.
  *
- * This validator is used only when cardinality not set to 1 or unlimited.
+ * Checks if more files than allowed have been uploaded or added via image
+ * library, truncates values to the limit and shows a warning message.
+ * This validator is only used when cardinality not set to 1 or unlimited.
+ *
+ * @see file_field_widget_form()
  */
 function file_field_widget_multiple_count_validate($element, &$form_state, $form) {
   $parents = $element['#parents'];
-  $values = backdrop_array_get_nested_value($form_state['values'], $parents);
-
   array_pop($parents);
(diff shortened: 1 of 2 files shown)


quicksketch:
Merged into 1.x for 1.34.0, including the added test coverage. Thank you @indigoxela!

PR #5263, opened 2026-05-05
Issue #7122: Prevent selecting more images from library than allowed

### LINKED ISSUE #7122: Follow-up: Selecting multiple images from library allows too many items
## Description of the bug

In #7118 it has been discovered that when selecting images from the image library (browser), the limit for selection via ui.selectable is always the full field cardinality. Totally disregarding existing images, that have been previously inserted.

That means, we let people chose files that have to get dropped on node save. Not very intuitive, bad UX.

## Steps To Reproduce

On a node type with a multi-value image field (cardinality > 1 but < unlimited).

1. Open a node form
2. Add some images, less than allowed
3. Open the dialog again and select as many as jquery.ui will let you (cardinality)

## Actual behavior

For a field that allows 3 items, you can add 5 in two runs. 2 the first time, 3 (cardinality) the second time.

Until #7119 is fixed, you'll get a TypeError on node save, but even if that wasn't the case - why letting people choose more than are allowed in the first place?

## Expected behavior

I can select the remaining count of images, not more.

## Additional information

- Backdrop CMS version: latest dev, since #440 

Note: this is about picking from library (dialog), not uploading.

indigoxela:
Fixes https://github.com/backdrop/backdrop-issues/issues/7122

### DIFF
--- a/core/modules/file/file.module
+++ b/core/modules/file/file.module
@@ -1897,6 +1897,7 @@ function file_managed_file_browser_open($form, $form_state) {
     'file' => array(
       'browser' => array(
         'fieldCardinality' => $field_cardinality,
+        'existingFilesCount' => end($parents),
         'selectedFid' => NULL,
         'currentFidElement' => '[name="' . $element['fid']['#name'] . '"]',
       ),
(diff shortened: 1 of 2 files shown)


quicksketch:
Thank you @indigoxela! I merged this into 1.x for 1.34.0.

PR #5264, opened 2026-05-10
Issue #7127: Make sure markup prepended to bulk tables is valid

### LINKED ISSUE #7127: Stray char in bulk operation views tbody element (with jQuery 4)
## Description of the bug

As an example the node_admin_content view ("Content"), which is used on admin/content. A subtle bug that just looks ugly but doesn't break any functionality.

## Steps To Reproduce

1. Generate a bunch of nodes, so there's more than one page
2. Switch to jQuery 4 on admin/config/development/jquery
3. Visit admin/content

## Actual behavior

You'll notice this stray `>` right under the content table header.

Image

## Expected behavior

Pretty markup.

## Additional information

- Backdrop CMS version: Apparently introduced by Issue #7038 in January this year

The problem happens in JS, in core/modules/views/js/bulk_form.js (looks like a typo).

```javascript
 $('.views-table > tbody', form).prepend('<tr class="views-select-all-pages--row even">></tr>');
```
It's very odd that this doesn't also happen with jQuery 3 - seems like v4 dropped some "silent fix" behavior, as above markup isn't valid. So the problem's also related to Issue #6404 (Add jQuery 4.x as additional option.)

indigoxela:
Fixes https://github.com/backdrop/backdrop-issues/issues/7127

### DIFF
--- a/core/modules/views/js/bulk_form.js
+++ b/core/modules/views/js/bulk_form.js
@@ -17,7 +17,7 @@ Backdrop.viewsBulkForm.initTableBehaviors = function(form) {
   // markup gets inserted below the table header.
   var $selectAllElement = $('.views-select-all-pages--wrapper', form);
   if ($selectAllElement.length) {
-    $('.views-table > tbody', form).prepend('<tr class="views-select-all-pages--row even">></tr>');
+    $('.views-table > tbody', form).prepend('<tr class="views-select-all-pages--row even"></tr>');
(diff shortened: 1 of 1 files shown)


yorkshire-pudding (approved):
Looks good. Thanks @indigoxela

quicksketch:
Merged into 1.x and 1.33.x. Thank you @indigoxela and @yorkshire-pudding!

PR #5265, opened 2026-05-11
Issue #7126: update admin views to use responsive table column UI

### LINKED ISSUE #7126: Responsive field classes via table settings don't work for all views of existing sites
## Description of the issue

This is a follow-up to issue https://github.com/backdrop/backdrop-issues/issues/3656. @yorkshire-pudding mentioned in a comment that the new UI setting "is not working with Medium priority". Others weren't able to reproduce the issue at that time, and the PR was merged.

When I tested the 1.34 preview release today, I was able to reproduce the issue with Medium priority at the `admin/content` view.

**edit:** I noticed the issue on a site which was updated from a previous Backdrop version.

**Steps to reproduce:**

- Go to the view, set the Responsive Priority for the _Name_ column to Medium and save the view.
- Go to `admin/content`, and narrow the screen.

**Actual behavior:**

- At 720px and below, the _Name_ `<th>` disappears. The corresponding `<td>` element with the Name values remains.
- The `<td>` with the Name values disappears only at 450px and below.

**Expected behavior:** 
Both elements should disappear in the same situation, which should be at 450px and below.

## Additional information

When you look at the rendered HTML, note that the `<th>` element has not only the "priority-medium" class but also "priority-low":

```
<th class="priority-low priority-medium" ...>
(...)
<td class="priority-medium">
```

herbdool:
Fixes https://github.com/backdrop/backdrop-issues/issues/7126

### DIFF
--- a/core/modules/views/views.install
+++ b/core/modules/views/views.install
@@ -257,6 +257,49 @@ function views_update_1009() {
 }
 
 /**
+ * Use new responsive table column UI rather than custom css class.
+ */
+function views_update_1010() {
+  $view_names = array('views.view.user_admin', 'views.view.node_admin_content');
+  foreach ($view_names as $view_name) {
+    $config = config($view_name);
+    $displays = $config->get('display');
+    foreach ($displays as $display_key => $display_details) {
+      if (!isset($display_details['display_options']['fields'])) {
+        continue;
+      }
+      foreach ($display_details['display_options']['fields'] as $field_key => $field) {
+        if (!isset($field['element_label_class'])) {
+          continue;
+        }
+        $label_classes = explode(' ', $field['element_label_class']);
+        $has_low = in_array('priority-low', $label_classes);
+        $has_medium = in_array('priority-medium', $label_classes);
+        $label_classes = array_values(array_filter($label_classes, function ($p) {
+          return $p !== 'priority-low' && $p !== 'priority-medium';
+        }));
+        if ($has_low || $has_medium) {
+          $updated_classes = implode(' ', $label_classes);
+          $config->set('display.' . $display_key . '.display_options.fields.' . $field_key . '.element_label_class', $updated_classes);
+          if (!$updated_classes) {
+            $config->set('display.' . $display_key . '.display_options.fields.' . $field_key . '.element_default_classes', 1);
+          }
+          $responsive = '';
+          if ($has_low) {
+            $responsive = 'priority-low';
+          }
+          if ($has_medium) {
+            $responsive = 'priority-medium';
+          }
+          $config->set('display.' . $display_key . '.display_options.style_options.info.' . $field_key . '.responsive', $responsive);
+        }
+      }
+    }
+    $config->save();
+  }
+}
+
+/**
  * @} End of "addtogroup updates-7.x-to-1.x"
  * The next series of updates should start at 2000.
  */
--- a/core/modules/views/views.install
+++ b/core/modules/views/views.install
@@ -260,18 +260,19 @@ function views_update_1009() {
  * Use new responsive table column UI rather than custom css class.
  */
 function views_update_1010() {
-  $view_names = array('views.view.user_admin', 'views.view.node_admin_content');
+  $views_updated = 0;
+  $view_names = config_get_names_with_prefix('views.view.')
   foreach ($view_names as $view_name) {
     $config = config($view_name);
-    if (!$config) {
-      continue;
-    }
+    $view_updated = FALSE;
     $displays = $config->get('display');
     foreach ($displays as $display_key => $display_details) {
-      if (!isset($display_details['display_options']['fields'])) {
+      $fields = $display_details['display_options']['fields'] ?? NULL;
+      $style = $display_details['display_options']['style_plugin'] ?? NULL;
+      if (!$fields || $style !== 'table') {
         continue;
       }
-      foreach ($display_details['display_options']['fields'] as $field_key => $field) {
+      foreach ($fields as $field_key => $field) {
         if (!isset($field['element_label_class'])) {
           continue;
         }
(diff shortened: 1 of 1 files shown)


yorkshire-pudding (approved):
Thanks @herbdool 
Code all looks good and makes sense to me.

core/modules/views/views.install
```diff
@@ -256,6 +256,51 @@ function views_update_1009() {
...
+            $responsive = 'priority-medium';
+          }
+          $config->set('display.' . $display_key . '.display_options.style_options.info.' . $field_key . '.responsive', $responsive);
+        }
+      }
+    }
+    $config->save();
+  }
+
+  backdrop_set_message(t('The <em>Administer user accounts</em> and <em>Manage content</em> views have been updated to the new "Responsive priority" style option for table columns, and the <code>priority-low</code> and <code>priority-medium</code> custom classes have been removed from the <em>Configure field > Style settings > Customize label HTML > Create a CSS class</em> section. If you have used <code>priority-low</code> or <code>priority-medium</code> classes in the field style settings in other views, you might want to adapt those views manually.'));
```
quicksketch: I think a message may be excessive here. We could return the message so it shows up as the result of the update hook:

```suggestion
  return t('The <em>Administer user accounts</em> and <em>Manage content</em> views have been updated to the new "Responsive priority" style option for table columns, and the <code>priority-low</code> and <code>priority-medium</code> custom classes have been removed from the <em>Configure field > Style settings > Customize label HTML > Create a CSS class</em> section. If you have used <code>priority-low</code> or <code>priority-medium</code> classes in the field style settings in other views, you might want to adapt those views manually.');
```

quicksketch (commented):
@herbdool Could you make one further change? Check if any views have actually been modified before showing (or returning) the message? I'm not sure if this message will need to be shown when upgrading from Drupal 7, as the views may already be configured correctly when the view conversion happens for `admin/node` and `admin/people`?

core/modules/views/views.install
quicksketch: Also, could this message be shortened up a bit?

And the `>` character needs to be escaped as `&gt;`.

core/modules/views/views.install
```diff
@@ -256,6 +256,51 @@ function views_update_1009() {
   }
 }
 
+/**
+ * Use new responsive table column UI rather than custom css class.
+ */
+function views_update_1010() {
+  $view_names = array('views.view.user_admin', 'views.view.node_admin_content');
```
quicksketch: Instead using `config_get_names_with_prefix('views.view.')` could loop over all views:


```suggestion
  $views_updated = 0;
  $view_names = config_get_names_with_prefix('views.view.')
```

core/modules/views/views.install
```diff
@@ -256,6 +256,51 @@ function views_update_1009() {
...
+  foreach ($view_names as $view_name) {
+    $config = config($view_name);
+    if (!$config) {
+      continue;
+    }
```
quicksketch: This could be removed because `config_get_names_with_prefix()` will always be up-to-date.
```suggestion
    $view_updated = FALSE;
```

core/modules/views/views.install
```diff
@@ -256,6 +256,51 @@ function views_update_1009() {
...
+    $displays = $config->get('display');
+    foreach ($displays as $display_key => $display_details) {
+      if (!isset($display_details['display_options']['fields'])) {
+        continue;
+      }
```
quicksketch: When looping over all views, check if the style_plugin is table:

```suggestion
      $fields = $display_details['display_options']['fields'] ?? NULL;
      $style = $display_details['display_options']['style_plugin'] ?? NULL;
      if (!$fields || $style !== 'table') {
        continue;
      }
```

core/modules/views/views.install
```diff
@@ -256,6 +256,51 @@ function views_update_1009() {
...
+      foreach ($display_details['display_options']['fields'] as $field_key => $field) {
```
quicksketch:
```suggestion
      foreach ($fields as $field_key => $field) {
```

core/modules/views/views.install
```diff
@@ -256,6 +256,51 @@ function views_update_1009() {
...
+          $updated_classes = implode(' ', $label_classes);
+          $config->set('display.' . $display_key . '.display_options.fields.' . $field_key . '.element_label_class', $updated_classes);
+          $responsive = '';
+          if ($has_low) {
+            $responsive = 'priority-low';
+          }
+          if ($has_medium) {
+            $responsive = 'priority-medium';
+          }
+          $config->set('display.' . $display_key . '.display_options.style_options.info.' . $field_key . '.responsive', $responsive);
```
quicksketch, line 295:
```suggestion
          $config->set('display.' . $display_key . '.display_options.style_options.info.' . $field_key . '.responsive', $responsive);
          $view_updated = TRUE;
```

core/modules/views/views.install
```diff
@@ -256,6 +256,51 @@ function views_update_1009() {
...
+        }
+      }
+    }
+    $config->save();
```
quicksketch: Check if modifications were made before saving.
```suggestion
    if ($view_updated) {
      $views_updated++;
      $config->save();
    }
```

core/modules/views/views.install
quicksketch: If we update all views, I think we can shorten this down to the much simpler statement:

```suggestion
  return t('@count views updated to move responsive table classes to the table display settings.', array('@count' => $views_updated));
```

quicksketch (commented):
Here's an inline stab at the suggestions made during the weekly meeting today. May not work as I did this in the GitHub UI.

core/modules/views/views.install
```diff
@@ -256,6 +256,58 @@ function views_update_1009() {
...
+      }
+    }
+    if ($view_updated) {
+      $views_updated++;
+      $config->save();
+    }
+  }
+
+  if ($views_updated) {
+    return format_plural($views_updated, 'One view updated to move responsive table classes to the table display settings.', '@count views updated to move responsive table classes to the table display settings.');
```
quicksketch, line 307: Nice, you updated it to be plural too. Confirmed this worked as expected:

Image

quicksketch:
Merged into 1.x for 1.34.0. Thank you @herbdool!

PR #5271, opened 2026-05-18
Add Expand active trail option for Dropdown menu style in menu blocks

### LINKED ISSUE #7138: Add "Expand active trail" option for Dropdown menu style in menu blocks
#### Description of the need:
When using the **Dropdown menu** the menu always starts fully collapsed. This is not ideal for non-horizontal menu blocks like in a sidebar. When a user is navigating a deep section of the site, they lose the context of their location within the menu hierarchy. Even though Backdrop correctly identifies the `active-trail` classes, the `SmartMenus` library initializes by hiding all submenus.

#### Proposed solution:
Add a new checkbox to the Menu block configuration form (specifically when "Dropdown menu" style is selected) called **"Expand active trail"**.

*   **UI**: This checkbox should only be visible when the style is set to "Dropdown".
*   **Functionality**: When enabled, the menu should initialize with any `<li>` elements that have the `.active-trail` class already expanded.

#### Draft of feature description for Press Release:
Backdrop now includes an "Expand active trail" option for menu blocks using the Dropdown style. This allows site builders to ensure that the current section of a menu remains visible on page load, providing better navigation context for users and making it easier to create functional accordion-style sidebars without custom code.

shivansunfire:
Fixes https://github.com/backdrop/backdrop-issues/issues/7138

### DIFF
--- a/core/modules/system/js/menus.js
+++ b/core/modules/system/js/menus.js
@@ -13,6 +13,7 @@ Backdrop.behaviors.menuStyles = {
       var $menu = $(element);
       var style = $menu.data('menuStyle');
       var menuSettings = $menu.data('menuSettings');
+      var expandActiveTrail = $menu.data('expandActiveTrail');
       if ($menu.data('clickdown')) {
         menuSettings = $.extend(menuSettings, {
           noMouseOver: true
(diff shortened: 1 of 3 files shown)


juanolalla:
I don't know why this

```
    // Switch style to Hierarchical and verify the functionality is NOT applied to the front-end.
    $this->backdropGet($block_edit_url);
    $edit = array(
      'block_settings[style]' => 'hierarchical',
    );
```
    
would fail for those three test checks and works all the others. Not sure what to do to make those pass as well.

PR #5272, opened 2026-05-19
Issue #7129: Apply cardinality limits on file upload earlier

### LINKED ISSUE #7129: Fixed multi-file uploads so they apply cardinality limits earlier
## Description of the bug

This is a follow-up to https://github.com/backdrop/backdrop-issues/issues/7119 which only fixed the error, but not the base problem. It's possible to upload more files than allowed, they're only wiped out on node (entity) save.

It might be a bit tricky to get available items count at that point of the process, but let's explore...

## Steps To Reproduce

1. A file field with cardinality > 1 but < unlimited
2. Upload a bunch of files in one go, more than the field cardinality would allow

## Actual behavior

See them all attached to the field values in the form (they get wiped out on node save).

I don't think, the UX is ideal.

## Expected behavior

Some magic, that no more than $cardinality are attached - possibly uploaded?

## Additional information

- Backdrop CMS version: 1.34

Multi-upload came into core with #440

indigoxela:
Fixes https://github.com/backdrop/backdrop-issues/issues/7129

Note: initially there was also client-side validation, but that has been dropped after feedback.

### DIFF
--- a/core/includes/file.inc
+++ b/core/includes/file.inc
@@ -1773,15 +1773,18 @@ function file_save_upload($form_field_name, $validators = array(), $destination 
       $accepted = array_slice($files, 0, $limit);
       $rejected = array_slice($files, $limit);
       $delete_fids = array();
+      $delete_names = array();
       foreach ($rejected as $file) {
         $delete_fids[] = $file->fid;
+        $delete_names[] = $file->filename;
       }
       file_delete_multiple($delete_fids);
       $count = count($rejected);
-      $singular = 'One file has been rejected as the upload limit of @limit has been reached.';
-      $plural = '@count files have been rejected as the upload limit of @limit has been reached.';
+      $singular = 'One file has been rejected as the upload limit of @limit has been reached. Files: %file_list.';
+      $plural = '@count files have been rejected as the upload limit of @limit has been reached. Files: %file_list.';
       $message = format_plural($count, $singular, $plural, array(
         '@limit' => $limit,
+        '%file_list' => implode(', ', $delete_names),
       ));
       backdrop_set_message($message, 'error');
       $files = $accepted;
--- a/core/includes/file.inc
+++ b/core/includes/file.inc
@@ -1767,22 +1767,22 @@ function file_save_upload($form_field_name, $validators = array(), $destination 
 
   // After all invalid files have been sorted out, check the amount of files,
   // if a limit has been set.
-  // Note: We can only consider cardinality, not the amount of pre-existing
-  // files in a file field widget.
   if (isset($validators['file_validate_upload_limit'][0])) {
     $limit = $validators['file_validate_upload_limit'][0];
     if (count($files) >= $limit) {
       $accepted = array_slice($files, 0, $limit);
       $rejected = array_slice($files, $limit);
-      $message = t('@count files have been rejected as the upload limit of @limit has been reached.', array(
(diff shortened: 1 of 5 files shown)


core/includes/file.inc
```diff
@@ -1765,6 +1765,32 @@ function file_save_upload($form_field_name, $validators = array(), $destination
     }
   }
 
+  // After all invalid files have been sorted out, check the amount of files,
+  // if a limit has been set.
+  if (isset($validators['file_validate_upload_limit'][0])) {
+    $limit = $validators['file_validate_upload_limit'][0];
+    if (count($files) > $limit) {
+      $accepted = array_slice($files, 0, $limit);
+      $rejected = array_slice($files, $limit);
```
quicksketch: Maybe save creating a new variable by assigning it directly:

```suggestion
      $rejected = array_slice($files, $limit);
      $files = array_slice($files, 0, $limit);
```

core/includes/file.inc
```diff
@@ -1765,6 +1765,32 @@ function file_save_upload($form_field_name, $validators = array(), $destination
...
+      file_delete_multiple($delete_fids);
+      $count = count($rejected);
+      $singular = 'One file has been rejected as the upload limit of @limit has been reached. The following files have been omitted: %file_list.';
+      $plural = '@count files have been rejected as the upload limit of @limit has been reached. The following files have been omitted: %file_list.';
+      $message = format_plural($count, $singular, $plural, array(
+        '@limit' => $limit,
+        '%file_list' => implode(', ', $delete_names),
+      ));
+      backdrop_set_message($message, 'error');
+      $files = $accepted;
```
quicksketch: Would not be needed if assigned directly above.
```suggestion
```

core/includes/file.inc
```diff
@@ -1765,6 +1765,32 @@ function file_save_upload($form_field_name, $validators = array(), $destination
...
+      $rejected = array_slice($files, $limit);
+      $delete_fids = array();
+      $delete_names = array();
+      foreach ($rejected as $file) {
+        $delete_fids[] = $file->fid;
+        $delete_names[] = $file->filename;
+      }
+      file_delete_multiple($delete_fids);
+      $count = count($rejected);
+      $singular = 'One file has been rejected as the upload limit of @limit has been reached. The following files have been omitted: %file_list.';
```
```suggestion
      $singular = 'One file has been rejected as the upload limit of @limit has been reached. The following file has been omitted: %file_list.';
```

quicksketch (commented):
This looks good to me @indigoxela! Just a few nitpicks:

core/includes/file.inc
indigoxela: Oh, right, singular... 😉 But I also have to adapt the related test.  Otherwise it would fail.

indigoxela: No, wait, the test runs on plural. All good.

yorkshire-pudding (approved):
I've looked over the code and it makes sense. The comments are clear and concise and explain well where an explanation is beneficial.

The changes to the test make sense.

laryn:
Thank you! Merged into 1.x and 1.34.x.

laryn:
Thank you @indigoxela! I appreciate all of the other eyes on this as well. I gave it a final test and scanned the code and I think it's a big improvement. Merged into 1.x and 1.34.x.

PR #5274, opened 2026-05-21
Issue #7136: Add html5_upload to backdrop_merged_modules to prevent enabling

### LINKED ISSUE #7136: Fixed adding html5_upload to backdrop_merged_modules() list
## Description of the bug

Follow-up to https://github.com/backdrop/backdrop-issues/issues/440, where multi-upload functionality has finally landed in core. 🎉 

But for whatever reason we totally missed to add it to function backdrop_merged_modules() and that would be important, as the module breaks. More specifically - uploading into fields with cardinality > 1 won't work anymore.

## Steps To Reproduce

1. Update to Backdrop 1.34.0
2. See the message that the contrib module got disabled
3. Go to admin/modules and try to enable it again

## Actual behavior

You can enable it.

## Expected behavior

You shouldn't be able to.

## Additional information

Let's add the module [here](https://github.com/backdrop/backdrop/blob/1.x/core/includes/module.inc#L1018)

indigoxela:
Fixes https://github.com/backdrop/backdrop-issues/issues/7136

### DIFF
--- a/core/includes/module.inc
+++ b/core/includes/module.inc
@@ -1029,6 +1029,7 @@ function backdrop_merged_modules() {
     'book_cache', // Backdrop 1.32.0.
     'views_ajax_history', // Backdrop 1.34.0.
     'file_entity_inline', // Backdrop 1.34.0.
+    'html5_upload', // Backdrop 1.34.0.
   );
 }
 


quicksketch:
Thank you @indigoxela! Merged into 1.x and 1.34.x.

PR #5278, opened 2026-05-27
Issue 7144. Load layout.admin.css in the default group.

### LINKED ISSUE #7144: Fixed placement of "View page" button on Layouts
## Description of the bug

The "View page" button in the Manage Block page for Layout pages looks misplaced. See image:

Image

This problem started after  https://github.com/backdrop/backdrop/pull/5212 was merged
## Steps To Reproduce

To reproduce the behavior:
1. Open `admin/structure/layouts/manage/home`

## Actual behavior
The "View page" button is placed outside the frame.

## Expected behavior

The button placed in the right place. Like this:

Image

## Additional information

This problem started when  https://github.com/backdrop/backdrop/pull/5212 was merged. In previous versions of Backdrop, the file `layout.admin.css` was loaded after `dropbutton.css`. The rule for `.edit-layout-top .dropbutton-wrapper` "won" in previous versions. Now, this rule is defeated by `.js .dropbutton-wrapper` with its `width: auto`, which creates the problem.

That PR created the new library `layout.admin`. Previously, `layout.admin.css` was loaded individually in several places. By creating this library, the PR inadvertently modified the order in which these css files are loaded.

I think the issue here may be the rule that targets the `dropbutton-wrapper` itself. Changing the order of the CSS loading should not affect the UI.

This was first reported on [Zulip](https://backdrop.zulipchat.com/#narrow/channel/218635-Backdrop/topic/Backdrop.201.2E34.2E0/near/597948314)[#Backdrop > Backdrop 1.34.0 @ 💬](https://backdrop.zulipchat.com/#narrow/channel/218635-Backdrop/topic/Backdrop.201.2E34.2E0/near/597948314)

argiepiano:
Fixes https://github.com/backdrop/backdrop-issues/issues/7144

### DIFF
--- a/core/modules/layout/layout.module
+++ b/core/modules/layout/layout.module
@@ -1020,7 +1020,9 @@ function layout_library_info() {
     'title' => 'Layout administration',
     'version' => BACKDROP_VERSION,
     'css' => array(
-      $path . '/css/layout.admin.css' => array(),
+      $path . '/css/layout.admin.css' => array(
+        'group' => CSS_THEME,
+      ),
     ),
     'js' => array(
       $path . '/js/layout.admin.js' => array(),
--- a/core/modules/layout/layout.module
+++ b/core/modules/layout/layout.module
@@ -1021,7 +1021,7 @@ function layout_library_info() {
     'version' => BACKDROP_VERSION,
     'css' => array(
       $path . '/css/layout.admin.css' => array(
-        'group' => CSS_THEME,
+        'group' => CSS_DEFAULT,
       ),
     ),
     'js' => array(


core/modules/layout/layout.module
```diff
@@ -1020,7 +1020,9 @@ function layout_library_info() {
     'title' => 'Layout administration',
     'version' => BACKDROP_VERSION,
     'css' => array(
-      $path . '/css/layout.admin.css' => array(),
+      $path . '/css/layout.admin.css' => array(
+        'group' => CSS_THEME,
```
indigoxela: Didn't find the time to thoroughly test that, yet, but moving that file to the "theme" group might introduce new problems.

Would setting a "weight" be an option?

Edit: no, weight can't do the trick.

indigoxela: Much better to restore to CSS_DEFAULT :+1: resolved!

indigoxela:
@argiepiano I think, the PR title needs clarification "in the theme group" should actually be "in the default group". But that's only about the title. :wink:

argiepiano:
Title changed. Thanks

laryn:
Thanks! Merged into 1.x and 1.34.x.

PR #5279, opened 2026-05-30
Issue #7149. CKeditor flash of html on submit

### LINKED ISSUE #7149: Prevent flashing HTML when submitting CKEditor forms
## Description of the bug

## Steps To Reproduce

To reproduce the behavior:
2. Click to "Create Page"

## Actual behavior

See video above.

## Solution!

## Additional information

- PHP version: 8.1.34

swampopus:

Video of it working now (see original issue for video of problem):

[ckeditor_html_no_flash.webm](https://github.com/user-attachments/assets/e8e260e8-5914-42f4-b9a9-e87f231868f6)

### DIFF
--- a/core/modules/filter/js/filter.js
+++ b/core/modules/filter/js/filter.js
@@ -80,6 +80,7 @@ Backdrop.behaviors.filterEditors = {
         if (event.isDefaultPrevented()) {
           return;
         }
+        Backdrop.settings.formIsSubmitting = true;
         Backdrop.filterEditorDetach(field, Backdrop.settings.filter.formats[activeEditor]);
       });
     });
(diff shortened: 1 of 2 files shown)


core/modules/filter/js/filter.js
```diff
@@ -80,6 +80,7 @@ Backdrop.behaviors.filterEditors = {
         if (event.isDefaultPrevented()) {
           return;
         }
+        Backdrop.settings.formIsSubmitting = true;
         Backdrop.filterEditorDetach(field, Backdrop.settings.filter.formats[activeEditor]);
```
quicksketch, line 84: I think this would be better if we create a new parameter on `filterEditorDetach()` rather than using a global variable. I'll see if I can make a replacement PR.

quicksketch:
I think we're all in agreement that https://github.com/backdrop/backdrop/pull/5300 is the preferable approach.

PR #5281, opened 2026-06-02
Issue 7148 - Change text field to the fallback text format if we lose access or if the format gets deleted

### LINKED ISSUE #7148: Changing a textarea's allowed text format prevents any future edits
## Description of the bug

I had a textarea where I was allowing the Filtered text "Basic" text format.  I had created a node.  Then, I modified the field to only allow the Filtered text format "Plain Text", and suddenly I could no longer edit the field value, even though I was logged in as admin (uid 1).

## Steps To Reproduce

To reproduce the behavior:
1. Create a new Content Type.  Mine is called "Testing".  Add a text (long) field, which allows Filtered text, and the Basic text format.  Go ahead and create a node, like so:

Image

2. Go back to the Content Type, edit the text (long) field, and change the allowed format to only be "Plain Text".  As you can see, I renamed mine to "Plain Text w Line Breaks" for clarity.

Image

3. Now, edit the node you just created.  You can no longer make edits to the textarea:

Image

## Expected behavior

I would expect that any existing saved data would appear in the textarea, after having been passed through the selected filter.  Or, just run it through `check_plain()` and display in whatever editor appears.

In my example case, I would expect to see something like:

```
<p>
Hello, I am a test
</p>
```

Instead, the textarea is **disabled**, and I cannot make changes at all, even as admin.  Getting to a point where I could actually edit the textarea for this node is actually a bit of a pain (see Workaround below).

## Workaround?

The only workaround I could find was to (1) remove the text (long) field from the content type, then (2) re-save the node, then (3) add the text (long) field back in with only "Plain Text" filter selected.  

Note: _Simply removing and immediately re-adding the field to the content type didn't work; you have to save the node in between._

Obviously this wouldn't be ideal if I'm working on a live site with hundreds of nodes by multiple users.  Plus you completely lose any content which was previously there.

## Additional information

- PHP version: 8.3.27
- Database sever (MySQL or MariaDB?) and its version: MariaDB 10.11.11
- Operating System and its version: Linux (Synology NAS)
- Browser(s) and their versions: Chrome v148.0.7778.217

swampopus:
Fixes https://github.com/backdrop/backdrop-issues/issues/7148

**Current behavior:**  If a text format is deleted or we lose access, we cannot edit the text field/textarea (it becomes disabled) on every node where it was saved with that format.

Video of CURRENT (undesired) behavior:
https://github.com/user-attachments/assets/75fa6be7-b85e-4f68-96ff-45014006437e

**New behavior with this PR:** In this condition, the element's format is switched to the "fallback" format and does not get disabled.

Video of new (desired) behavior with this PR in place:
https://github.com/user-attachments/assets/66c906ce-c7d0-4519-806a-0ce13ddaf4be

### DIFF
--- a/core/modules/filter/filter.module
+++ b/core/modules/filter/filter.module
@@ -1424,35 +1424,14 @@ function filter_process_format($element) {
     // user only has access to a single format).
     $element['format']['format']['#access'] = TRUE;
   }
-  // Disable this widget, if the user is not allowed to use the stored format,
+  // Convert to our fallback text format if the user is not allowed to use the stored format,
   // or if the stored format does not exist, and only if the field is not empty.
   // The 'administer filters' permission only grants access to the filter
   // administration, not to all formats.
   elseif (!$user_has_access || !$format_exists) {
     if (!empty($element['#default_value'])) {
-      // Overload default values into #value to make them unalterable.
-      $element['value']['#value'] = $element['value']['#default_value'];
-      $element['format']['format']['#value'] = $element['format']['format']['#default_value'];
-
-      // Prepend #pre_render callback to replace field value with user notice
-      // prior to rendering.
-      $element['value'] += array('#pre_render' => array());
-      array_unshift($element['value']['#pre_render'], 'filter_form_access_denied');
-
-      // Cosmetic adjustments.
-      if (isset($element['value']['#rows'])) {
-        $element['value']['#rows'] = 3;
-      }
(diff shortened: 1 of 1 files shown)


swampopus:
Question for anyone looking at PR-- the tests that failed did so because they're expecting the text fields to display the "Text format access denied message", which is exactly what this PR removes.  Not sure what to do about that.

PR #5283, opened 2026-06-10
Issue #7154: Fix breaking date filters by reverting 6976

### LINKED ISSUE #7154: Update to 1.34.1 is breaking views filters with relative dates
## Description of the bug

A short description of what the problem is.

After installing 1.34.1 over 1.34.0, views blocks with sort filters with relative dates are no longer working.  

## Steps To Reproduce

To reproduce the behavior:
1. Create a view with content that has a date field, and filter that view with a relative date filter.
2. After the update to 1.34.1, the relative filter does not work.  
3. Views will show all values, without applying the date filter.
4. Revert to 1.34.0, and all views filters work as expected.

## Actual behavior

This is breaking views with relative filters on at least a couple of my sites.  These are not complex sites, I am just showing different pages for past a upcoming events.  

This is a critical bug because it is breaking the function of my website.  

## Additional information

Backdrop CMS:                   1.34.0
Installation profile:           standard
PHP version:                    8.3.31
Drupal 7 compatibility:         on
Database server:                11.4.12-MariaDB
Web server:                     LiteSpeed
jQuery version:                 1.12.4
jQuery UI version:              1.14.1
CKEditor 5 version: 48.0.0      Used in formats: Filtered HTML, Full HTML

indigoxela:
Fixes https://github.com/backdrop/backdrop-issues/issues/7154

Even if this wouldn't break existing (not exposed) date filters - even grouped filters are very buggy. We should revert that as a whole. It needs a lot more work to get into core.

The coding standard violations from reverted code (not actually mine) - should I fix that? I hesitate.

### DIFF
--- a/core/modules/date/date.elements.inc
+++ b/core/modules/date/date.elements.inc
@@ -142,12 +142,12 @@ function _date_element_info() {
  */
 function date_default_date($element) {
   $granularity = date_format_order($element['#date_format']);
-  $default_value = is_array($element['#default_value']) ? reset($element['#default_value']) : $element['#default_value'];
+  $default_value = $element['#default_value'];
   $format = DATE_FORMAT_DATETIME;
 
   // The text and popup widgets might return less than a full datetime string.
-  if (strlen((string) $default_value) < 19) {
-    switch (strlen((string) $default_value)) {
+  if (strlen($element['#default_value']) < 19) {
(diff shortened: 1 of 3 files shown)


indigoxela:
Closing as obsolete (git revert has been done).

PR #5284, opened 2026-06-10
Issue #7153: Continue to support deprecated layout_info variable.

### LINKED ISSUE #7153: Follow-up: Variable layout_info is now null in preprocessors
## Description of the bug

Having a theme, that relies on `$variables['layout_info']` being available and contain values.

Update to BackdropCMS 1.34.x and inspect log

## Actual behavior

```
Warning: Trying to access array offset on value of type null in THEME_preprocess_layout__XXX() (... themes/YOURTHEME/template.php).
```

## Expected behavior

Values available as before 1.34.x, I guess. No need to update themes.

## Additional information

- Backdrop CMS version: 1.34.x

Introduced with Issue #6808: Allow to specify a layout template type.

The variable still exists, but is NULL now, so one can no longer access keys like name

laryn:
Fixes https://github.com/backdrop/backdrop-issues/issues/7153

### DIFF
--- a/core/modules/layout/layout.theme.inc
+++ b/core/modules/layout/layout.theme.inc
@@ -371,8 +371,11 @@ function layout_preprocess_layout(&$variables) {
   }
 
   // Support deprecated 'layout_info' property until Backdrop 2.0.
-  if (isset($variables['layout_info']) && !isset($variables['layout_template_info'])) {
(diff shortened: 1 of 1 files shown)


laryn:
Merged into 1.x and 1.34.x

PR #5286, opened 2026-06-12
Issue #7078: Prevent form submission by autocomplete selection

### LINKED ISSUE #7078: Editor link autocomplete doesn't work with keyboard
## Description of the bug

Choosing an internal page in the link dialog with CKEditor or TinyMCE works only if you select a page from the suggestions with the mouse. If you select the page with the keyboard, the path won't be completed.

## Steps To Reproduce

To reproduce the behavior:
1. Create a page, and add some text.
2. Select a passage of the text, and open the link dialog.
3. Start to type the title of an existing page, e.g. of the _About_ page.
4. Use the arrow down key of your keyboard to highlight the page from the list of suggestions, and hit Enter.

## Actual behavior

The link dialog gets closed, and the link is created, but the path isn't complete.
Result: `<a href="/abo">About</a>`

## Expected behavior

The path gets completed without closing the dialog, as if you select a page using the mouse, and you have to save the dialog explicitely.
Result: `<a href="/about">About</a>`

## Additional information

- Backdrop CMS version: 1.33.0, but experienced it also in older versions
- Operating System: Mac OS
- Browser(s): current Firefox and Chrome

indigoxela:
Fixes https://github.com/backdrop/backdrop-issues/issues/7078

### DIFF
--- a/core/misc/autocomplete.js
+++ b/core/misc/autocomplete.js
@@ -66,6 +66,8 @@ Backdrop.jsAC.prototype.onkeydown = function (input, e) {
     case 38: // up arrow.
       this.selectUp();
       return false;
+    case 13: // Enter key.
+      return false; // Prevent form submit.
     default: // All other keys.
       return true;
   }


laryn:
Thanks!

PR #5287, opened 2026-06-19
Fixed unapproved comments page redirects away after submission.

### LINKED ISSUE #4314: Unapproved comments page redirects away after submission
**Description of the bug**
After submission of the **Unapproved comments** page (`admin/content/comment/approval`) it redirects away to the **Published comments** page (`admin/content/comment`), which could cause inattentive users to have an impression they are still on the same page and to go ahead to delete legitimate comments. 

It is especially dangerous when users are dealing with massive spam comments and when they want to stay at the same page and keep mass-selecting and deleting such comments until they are all gone.

**Steps To Reproduce**
To reproduce the behavior:
1. Make sure the "Skip comment approval" on `admin/config/people/permissions` is unchecked for a user role, for example, for ANONYMOUS role. Logout.
2. Post some comments as an ANONYMOUS user, so your comments are not yet approved.
3. Login as admin and go to the **Unapproved comments** page (`admin/content/comment/approval`), mass select comments and delete them and it redirects away to another page instead of staying at the same place.

**Actual behavior**
out2

**Expected behavior**

The expected behavior is to stay at the same page unless/until you intentionally browse away.

izmeez:
Fixes https://github.com/backdrop/backdrop-issues/issues/4314

### DIFF
--- a/core/modules/comment/comment.admin.inc
+++ b/core/modules/comment/comment.admin.inc
@@ -175,6 +175,9 @@ function comment_admin_overview_validate($form, &$form_state) {
  * Executes the chosen 'Update option' on the selected comments, such as
  * publishing, unpublishing or deleting.
  *
+ * The site will stay on the current page after submitting and in the unlikely
+ * case there is a ?destination= query, it will respect that.
+ *
  * @see comment_admin_overview_validate()
  */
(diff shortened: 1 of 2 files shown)


quicksketch:
Thanks @izmeez! Merged into 1.x and 1.34.x.

PR #5290, opened 2026-06-23
Fixed comment reply link anchor.

### LINKED ISSUE #7162: Reply to comment link is missing anchor #comment-form
## Description of the bug

When comments are enabled on a content type with the option "Show reply form on the same page as comments" unchecked and comments exist the user has two options: they can click "Add comment" at the bottom of the node or they can click "Reply" to an existing comment.

In the first instance, when they click "Add comment" they are directed to a page with the title "Add comment" where the node is displayed and the link takes them immediately to the place below the node to where the comment form begins. This is because the anchor `#comment-form` is added to the link.

In the second instance, when they are viewing a comment and click "Reply" they are directed to a page with the title "Add comment" where the comment is displayed. However, the link does not take them immediately to the place below the comment to where the comment form begins. This is because the anchor `#comment-form` is not added to the link.

## Expected behavior

The expected behavior is that the link takes them to where the comment form begins.

izmeez:
Fixes https://github.com/backdrop/backdrop-issues/issues/7162

This PR was created with AI assistance. I am comfortable with the result. I am not sure if the community will be.

### DIFF
--- a/core/modules/comment/comment.module
+++ b/core/modules/comment/comment.module
@@ -1040,11 +1040,17 @@ function comment_links(Comment $comment, Node $node) {
       );
     }
     if (Comment::createAccess()) {
+      $node_type = node_type_get_type($node->type);
       $links['comment-reply'] = array(
         'title' => t('reply'),
         'href'  => "comment/reply/$comment->nid/$comment->cid",
         'html'  => TRUE,
       );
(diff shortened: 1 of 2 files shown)


quicksketch:
Merged into 1.x and 1.34.x. Thank you @izmeez!

PR #5292, opened 2026-06-24
Issue 7166: Make the GET page variable consistent with real page number

### LINKED ISSUE #7166: [UX] Make the GET page variable consistent with the real page number
## Description of the need

From: https://www.drupal.org/project/drupal/issues/385270
Backdrop's ~~Drupal's~~ pagination is a bit geeky: it starts from 0 instead of 1. If we are on the first page and we click on the number 2 (or on the "next" link) in the navigation bar, we're displaying the second page, but the `$_GET['page']` variable on the url is `?page=1`.

`$_GET['page']` variable is one lower than the real page number marked in the navigation bar. For most people, I think that it's a bit disconcerting and not user-friendly ~~ergonomics~~.

## Proposed solution
The `page` parameter in the URL will match the active number on the pager link

None

No

## Additional information
This was raised as a D7 issue but never merged:
https://www.drupal.org/project/drupal/issues/385270

There is a patch by [azovsky](https://www.drupal.org/u/azovsky) at:
https://www.drupal.org/project/drupal/issues/385270#comment-11982646

While the line numbers are different, the changes can apply to Backdrop's code mostly the same (small change to accommodate newer PHP versions). I also needed to modify the tests to reflect the page numbers now delivered.

Backdrop's pagers will now show the same page in the URL as is shown in the pager link.

---
PR: https://github.com/backdrop/backdrop/pull/5292

yorkshire-pudding:
Fixes https://github.com/backdrop/backdrop-issues/issues/7166

### DIFF
--- a/core/includes/pager.inc
+++ b/core/includes/pager.inc
@@ -25,9 +25,11 @@
 function pager_find_page($element = 0) {
   $page = isset($_GET['page']) ? $_GET['page'] : '';
   $page_array = explode(',', $page);
-  if (!isset($page_array[$element])) {
-    $page_array[$element] = 0;
+  if (empty($page_array[$element])) {
+    $page_array[$element] = 1;
   }
+  // Translate human readable page numbers to PHP offset logic.
+  $page_array[$element]--;
   return (int) $page_array[$element];
 }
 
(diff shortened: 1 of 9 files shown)


yorkshire-pudding:
As this breaks some contrib modules and as there is now a contrib module that can achieve the same result (see https://github.com/backdrop/backdrop-issues/issues/7166#issuecomment-4813941361), I will close this PR (though will leave the branch for now)

PR #5295, opened 2026-06-30
Issue #6865: bring Field Group into core

### LINKED ISSUE #6865: Field group functionality in core
The original issue reached 83 comments and in a dev meeting, we decided to experiment by closing that old issue and opening a new issue with a summary. I've used AI to help draft the summary, but also read through the entire issue myself and added some details. 
Link to original issue: https://github.com/backdrop/backdrop-issues/issues/647

This issue has been in the top 10 of our wishlist for the last several years and is now the top issue on the wishlist: https://forum.backdropcms.org/feature-requests

**Initial Proposal (2015) & Evolution**

- The original idea was to move the entire Field Group module into core.
- As discussions evolved, it was decided that instead of moving the whole module, a Field Group API would be integrated into core, and only some field grouping types would be included in core.

**Core vs. Contrib Division**

- Core: These are field group types that already exist in core in some form.
  - Fieldset
  - Details
  - HTML element (converted Div to HTML element if not using collapsible)
  - Vertical Tabs
- Contrib (Field Group Extra): These require additional complexity or are less frequently needed.
  - HTML5 (was merged into HTML element)
  - Div (if using the collapsible)
  - Horizontal Tabs
  - Multipage
  - Accordion

**Technical Implementation Plan [UPDATED]**

- Field Group contrib module separated into Field Group and Field Group Extra.
- Field Group contrib module will be merged into core.
- Admins will get a status warning if any field groups require Field Group Extra.
- Field Group contrib module will be disabled in update hook.

Old:

- ~~The Field Group API will be built into the field and field_ui modules in core.~~
- ~~The Field Group contrib module will be refactored.~~
- ~~A 2.x version of the contrib module will be released, depending on the new core API.~~
- ~~The contrib module will remove the features that move to core and adjust the remaining ones to use the new API.~~
- ~~The old version (1.x) of the contrib module will remain available for Backdrop sites using older versions of core.~~

**Challenges & Roadblocks**

- **Code Structure Debate**: Some contributors suggested simply merging the contrib module into core, while others preferred splitting API functionality into core and keeping the UI features in contrib. Since we have a policy of never putting any API into core without a working implementation, **the compromise was reached to add the API, but also update core's current features to use the new API**. 
- **Order of operations**: An early pull request (PR) attempting to integrate the entire module was rejected by core maintainers since it was doing more than seemed appropriate for core. It was recommended to first refactor the code in the contrib module, also separating the parts destined for core from the rest. Finally, the core parts will be moved into core, and a new contrib release will be made without them. 
- **Upgrade Path Issues**: The goal is to ensure that existing sites using the contrib module can seamlessly transition to the new core API without breakage or loosing any functionality.

**Current Status & Next Steps**

Help needed:
  - Reviewers testing the new core groupings
  - Developers to help refactor the contrib module.
  - Reviewers for UI/UX improvements in contrib
  - Testing to ensure a smooth upgrade path.

Some specific comments/requests (maybe follow-ups):

- _Can we add a clone link into operations for the field groups?_ https://github.com/backdrop/backdrop-issues/issues/647#issuecomment-675165037
- _I think it might be easier if the FG containers were collapsible and you could drag one collapsed container into another collapsed container._  - https://github.com/backdrop/backdrop-issues/issues/647#issuecomment-767311650
- _Make a field group look and act more like the other fields by moving all configuration to a seperate page (i.e. remove the ability to change the group type from this overview page, and move Configuration from an expanding section to a new page (same as field settings))._ https://github.com/backdrop/backdrop-issues/issues/647#issuecomment-674474016
- _(It took me an embarassingly long time in Drupal to discover that rearranging the data structure was also rearranging the edit form.) And the ability to create multiple versions of the edit form would be GREAT. I've got a use case now where ordinary people edit a subset of node fields on a bespoke edit form, while admin users can access more of the fields on a different edit form; being able to set up something like that from the UI (rather than hacking it in code, as now) would be great._ https://github.com/backdrop/backdrop-issues/issues/647#issuecomment-988070343

Thanks to the following contributors who have participated in this discussion, so far. 

- @sirkitree
- @quicksketch 
- @klonos 
- @mikemccaffrey 
- @MrHaroldA 
- @biolithic 
- @serundeputy 
- @vonnnew
- @herbdool 
- @jenlampton 
- @laryn 
- @quinnanya 
- @olafgrabienski 
- @philsward
- @ericfoy
- @stpaultim 
- @bugfolder 
- @docwilmot 
- @avpaderno 

Related Issues:
- #5898 
- #3072 
- #2608 

Related Contrib Modules
- [Field Group](https://github.com/backdrop-contrib/field_group) 
- [Field Group Extra](https://github.com/backdrop-contrib/field_group_extra)

Youtube discussion - Dec 10, 2021 - https://youtu.be/KLihQQK7f-M?t=1355

herbdool:
Fixes https://github.com/backdrop/backdrop-issues/issues/6865


core/modules/system/system.install
```diff
@@ -3921,6 +3926,68 @@ function system_update_1106() {
...
+  }
+}
+
+/**
+ * Notify if a non-core version of Field Group module is present.
+ */
+function system_update_1108() {
+  $path = backdrop_get_path('module', 'field_group');
+  if ($path && strpos($path, 'core/modules/field_group') === FALSE) {
+    backdrop_set_message(t('Backdrop core provides a bundled @name module. A different copy of @name module is at %path. Remove this module from your installation to use the core @name module. For more information see the <a href="!url" target="_blank">@name change notice</a>.', array(
```
indigoxela, line 3982: Isn't the desired behavior for update hook having "something to say" to put it into a return, so it shows up at the end of the update process?

NOT backdrop_set_message inside the hook.

It might actually belong to the status page instead? Or additionally? For sure not backdrop-set_message. I know this has been done several times before, but that doesn't make it any better.

core/modules/system/system.install
```diff
@@ -3921,6 +3926,68 @@ function system_update_1106() {
...
+      $extras_enabled = TRUE;
+      break;
+    }
+  }
+
+  if ($extras_enabled) {
+    if ($extras_module === FALSE) {
+      // Field Group Extra is not present so set message to fetch it.
+      // cspell:ignore Multipage
+      backdrop_set_message(t('The <em>Accordion</em>, <em>Horizontal tabs</em>, and <em>Multipage</em> field group types have been moved out of the Field Group module. The contrib module !module should be downloaded and enabled to ensure the existing field groups continue to work properly. Or update those field groups to use other group types.', array('!module' => l('Field Group Extra', 'https://backdropcms.org/project/field_group_extra'))));
```
indigoxela, line 3964: For backdrop_set_message in update hooks see my concern below.

core/modules/system/system.install
herbdool: Maybe there's a better way, but I had just copied and adapted a previous update hook in the same file. So perhaps a new issue to change them all at the same time?

herbdool:
closing in favour of https://github.com/backdrop/backdrop/pull/5296 which is in the right fork.

PR #5299, opened 2026-07-07
Issue #7149: Prevent flash of HTML on submitting CKEditor forms.

### LINKED ISSUE #7149: Prevent flashing HTML when submitting CKEditor forms

quicksketch:

Alternative to the PR approach in https://github.com/backdrop/backdrop/pull/5279

This creates a new parameter to `Backdrop.filterEditorDetach()` to indicate if the detachment is caused by unloading the entire window such as when a form is being submitted.

### DIFF
--- a/core/modules/ckeditor5/js/ckeditor5.js
+++ b/core/modules/ckeditor5/js/ckeditor5.js
@@ -43,7 +43,7 @@
           let onEventsPattern = {
             'name': /.*/,
             'attributes': /^on.*/
-          }
+          };
           editorSettings.htmlSupport.disallow.push(onEventsPattern);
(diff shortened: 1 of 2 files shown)


swampopus:
I have tested this, and switching between formats still works normally, and saving correctly does not show any HTML.  

Looks good to me!

quicksketch:
Replaced by https://github.com/backdrop/backdrop/pull/5300.

PR #5300, opened 2026-07-08
Issue #7149: Third approach to prevent flashing HTML when submitting CKEditor forms

### LINKED ISSUE #7149: Prevent flashing HTML when submitting CKEditor forms

quicksketch:

This is an alternative to my earlier PR https://github.com/backdrop/backdrop/pull/5299 and @swampopus's https://github.com/backdrop/backdrop/pull/5279.

Instead of using a new parameter like my previous PR, this uses the existing "trigger" parameter with the value set to "submit". I think this is okay because previously we were not using any value at all when calling this function, other than "serialize". So introducing a new option to differentiate between "unload" (which is what should have been used before but wasn't in place) and "submit" (new option for when non-AJAX submitting the form), seems safe.

### DIFF
--- a/core/modules/ckeditor5/js/ckeditor5.js
+++ b/core/modules/ckeditor5/js/ckeditor5.js
@@ -123,7 +123,7 @@
       if (trigger !== 'serialize') {
         // If submitting the entire form, skip destroying the editor and let it
         //  be unloaded by the browser leaving the page.
-        if (trigger === 'submit') {
+        if (trigger !== 'submit') {
           editor.destroy();
         }
         Backdrop.ckeditor5.instances.delete(editor.id);
--- a/core/modules/ckeditor5/js/ckeditor5.js
+++ b/core/modules/ckeditor5/js/ckeditor5.js
@@ -101,7 +101,7 @@
         });
     },
 
-    detach: function (element, format, trigger, windowUnload) {
+    detach: function (element, format, trigger) {
       // Remove any content modification warning.
       if (element.ckeditor5AttachedWarning && trigger !== 'serialize') {
         element.ckeditor5AttachedWarning.remove();
@@ -123,7 +123,7 @@
       if (trigger !== 'serialize') {
         // If submitting the entire form, skip destroying the editor and let it
         //  be unloaded by the browser leaving the page.
-        if (!windowUnload) {
+        if (trigger === 'submit') {
           editor.destroy();
         }
         Backdrop.ckeditor5.instances.delete(editor.id);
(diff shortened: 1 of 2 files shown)


core/modules/ckeditor5/js/ckeditor5.js
```diff
@@ -123,7 +123,7 @@
       if (trigger !== 'serialize') {
         // If submitting the entire form, skip destroying the editor and let it
         //  be unloaded by the browser leaving the page.
-        if (!windowUnload) {
+        if (trigger === 'submit') {
```
swampopus: Should be `if trigger !== 'submit'`

swampopus:
@quicksketch --- Sorry-- you got it backwards ;)  The editor should only call destroy if `trigger !== 'submit'`

`editor.destroy()` is what causes the flash of HTML to appear.

core/modules/ckeditor5/js/ckeditor5.js
indigoxela: I belief, the desired logic here would be:
```
if (trigger === 'submit') {
  return;
}
```
Doing nothing, also none of the cleanup jobs below.
And probably not inside the `trigger !== serialize` check.

swampopus: Unfortunately we can't just return here, because lower in the function we have this:
```js
// Save formatted value after destroying the editor, which can also
// update the element value.
element.value = newData;
```
This is what actually writes the plain HTML back to the hidden textarea, which actually gets saved.  So we really do need to simply skip over the .destroy() call, so it does need to be within the !== 'serialize' if block.

indigoxela: ... but that should be inside some `trigger === "serialize"` check
serialize gets dispatched before unload ... in theory... Edit: yes, that's dispatched before (core/misc/ajax.js).

Oh dear, so much easier with Tiny... :hand_over_mouth:

swampopus: lol, true.  Probably the whole detach function here could use some sprucing up-- but for today, I think simply changing it to `trigger !== 'submit'` gets it done.

core/modules/ckeditor5/js/ckeditor5.js
```diff
@@ -121,7 +121,11 @@
 
       // Destroy the instance if fully detaching.
       if (trigger !== 'serialize') {
-        editor.destroy();
+        // If submitting the entire form, skip destroying the editor and let it
+        //  be unloaded by the browser leaving the page.
+        if (trigger === 'submit') {
```
quicksketch:
```suggestion
        if (trigger !== 'submit') {
```

swampopus:
Perfect 😎

laryn:
Thanks -- merged into `1.x` and `1.34.x`.

PR #5301, opened 2026-07-08
Fix file delete page.

### LINKED ISSUE #7169: Error when deleting files referenced in paragraphs
## Description of the bug

With paragraphs installed, when you delete a file that is referenced in a paragraph item, you get the error:

```
l(): Argument #3 ($options) must be of type array, null given, called in /var/www/html/web/core/modules/file/file.pages.inc on line 1000
```

## Steps To Reproduce

To reproduce the behavior:

1. Install paragraph module.
2. Add a paragraph type that allows uploading a file.
3. Create a node that has the paragraph item.
4. Go to /admin/content/files.
5. Location the file.
6. Delete the file.

## Actual behavior

Error is thrown.

## Expected behavior

Allow deleting of the file.

puresyntax71:
Fixes https://github.com/backdrop/backdrop-issues/issues/7169


core/modules/file/file.pages.inc
```diff
@@ -1023,7 +1023,14 @@ function _file_usage_list_links(File $file, &$known_count, &$unknown_count) {
             // If the entity is in the EFQ result, access is allowed.
             if (isset($result[$entity_type][$entity->id()])) {
               $uri = $entity->uri();
-              $entity_list[] = l($entity->label(), $uri['path'], $uri['options']);
+
+              if (!empty($uri['path'])) {
```
argiepiano, line 1027: Since `$entity->uri()` can return an empty array (according to EntityInterface::uri() documentation) I would suggest to simplify this to `if (!empty($uri)) {}`. The $uri variable will either have both keys (path and options), or it will be completely empty.




```suggestion
              if (!empty($uri)) {
```

quicksketch: I think this current approach is actually already "safe" in that PHP does not throw any warnings or notices when an array key does not exist. According to the docs (https://www.php.net/empty#example-3) this is totally an acceptable usage to check for array keys that may not exist.

quicksketch:
Merged into 1.x and 1.34.x. Thank you @puresyntax71 and @argiepiano!

PR #5302, opened 2026-07-09
2054 add lock and unlock aspect ratio to images in editor

### LINKED ISSUE #2054: [UX] Add an option to "lock" aspect ratio when resizing images in the editor dialog

swampopus:
Fixes https://github.com/backdrop/backdrop-issues/issues/2054

Note: To globally disable the ability to "unlock" the aspect ratio, add this to settings.php:
```php
$settings['disable_image_aspect_ratio_unlock'] = TRUE;
```

### DIFF
--- a/core/modules/filter/filter.pages.inc
+++ b/core/modules/filter/filter.pages.inc
@@ -216,7 +216,7 @@ function filter_format_editor_image_form($form, &$form_state, $format) {
     '#parents' => array('attributes', 'height'),
     '#field_suffix' => ' ' . t('pixels'),
   );
-  
+
   $ratio_icons = "";
   // Skip these icons if the user is not allowed to unlock the aspect-ratio functionality:
   if (settings_get('disable_image_aspect_ratio_unlock', FALSE) !== TRUE) {
@@ -225,10 +225,10 @@ function filter_format_editor_image_form($form, &$form_state, $format) {
     $ratio_icons .= '<div id="lock-aspect-ratio" class="image-ratio-icon lock-aspect-ratio" style="display:none;" title="' . t('Lock aspect ratio') . '">' . icon('lock-simple-open-fill') . '</div>';
   }
   $ratio_icons .= '<div id="reset-orig" class="image-ratio-icon reset-orig reset-orig-inactive" title="' . t('Reset to original size') . '">' . icon('arrows-clockwise-fill') . '</div>';
-  
+
   // Add ratio icons after the "height" field.
   $form['size']['height']['#field_suffix'] .= $ratio_icons;
-  
+
   // Original values used for reference in submit handler.
   $form['size']['previous_width'] = array(
     '#type' => 'value',
--- a/core/modules/filter/filter.pages.inc
+++ b/core/modules/filter/filter.pages.inc
@@ -218,9 +218,8 @@ function filter_format_editor_image_form($form, &$form_state, $format) {
   );
 
   $ratio_icons = "";
-  // Skip these icons if the user is not allowed to unlock the aspect-ratio functionality:
-  if (settings_get('disable_image_aspect_ratio_unlock', FALSE) !== TRUE) {
-    // Okay, I know using settings_get is kind of a kludge. It's here until a better UI location can be identified.
+  // Skip these icons if the user is not allowed to unlock the aspect-ratio functionality. (Can be set in settings.php by adding $settings['disable_image_aspect_ratio_unlock'] = TRUE;)
+  if (settings_get('disable_image_aspect_ratio_unlock', FALSE) !== TRUE) {    
     $ratio_icons .= '<div id="unlock-aspect-ratio" class="image-ratio-icon unlock-aspect-ratio" title="' . t('Unlock aspect ratio') . '">' . icon('lock-simple-fill') . '</div>';
     $ratio_icons .= '<div id="lock-aspect-ratio" class="image-ratio-icon lock-aspect-ratio" style="display:none;" title="' . t('Lock aspect ratio') . '">' . icon('lock-simple-open-fill') . '</div>';
   }
--- a/core/modules/filter/filter.pages.inc
+++ b/core/modules/filter/filter.pages.inc
@@ -221,10 +221,10 @@ function filter_format_editor_image_form($form, &$form_state, $format) {
   // Skip these icons if the user is not allowed to unlock the aspect-ratio functionality.
   // Modify by adding "$settings['disable_image_aspect_ratio_unlock'] = TRUE;" to settings.php.
   if (settings_get('disable_image_aspect_ratio_unlock', FALSE) !== TRUE) {
-    $ratio_icons .= '<div id="unlock-aspect-ratio" class="image-ratio-icon unlock-aspect-ratio" title="' . t('Unlock aspect ratio') . '">' . icon('lock-simple-fill') . '</div>';
-    $ratio_icons .= '<div id="lock-aspect-ratio" class="image-ratio-icon lock-aspect-ratio" style="display:none;" title="' . t('Lock aspect ratio') . '">' . icon('lock-simple-open-fill') . '</div>';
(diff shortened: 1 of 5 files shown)


core/modules/filter/js/filter.js
```diff
@@ -80,6 +80,7 @@ Backdrop.behaviors.filterEditors = {
         if (event.isDefaultPrevented()) {
           return;
         }
+        Backdrop.settings.formIsSubmitting = true;
```
herbdool: This seems like it belongs to a different PR.

core/modules/filter/js/filter.js
```diff
@@ -304,17 +304,155 @@ Backdrop.behaviors.editorImageLibrary = {
...
+    if ((curWidth.length && $icon.data('data-dimensions').width != curWidth) || (curHeight.length && $icon.data('data-dimensions').height != curHeight))  {
+      $('.filter-format-editor-image-form .editor-image-size #reset-orig').removeClass('reset-orig-inactive');
+    }
+    else {
+      $('.filter-format-editor-image-form .editor-image-size #reset-orig').addClass('reset-orig-inactive');      
+    }
+  }  
 };
 
+
```
herbdool: Extra spaces not needed here.

core/modules/filter/filter.pages.inc
```diff
@@ -216,6 +216,19 @@ function filter_format_editor_image_form($form, &$form_state, $format) {
     '#parents' => array('attributes', 'height'),
     '#field_suffix' => ' ' . t('pixels'),
   );
+
+  $ratio_icons = "";
+  // Skip these icons if the user is not allowed to unlock the aspect-ratio functionality:
+  if (settings_get('disable_image_aspect_ratio_unlock', FALSE) !== TRUE) {
+    // Okay, I know using settings_get is kind of a kludge. It's here until a better UI location can be identified.
```
herbdool: It doesn't necessarily need to be in the UI, so this is okay. In some cases, we don't want it in the UI.

core/modules/filter/js/filter.js
swampopus: Heh-- yes, it does.  Sorry-- let me clean that up.

core/modules/filter/js/filter.js
swampopus: Fixed!

core/modules/filter/filter.pages.inc
swampopus: I changed the comment above to instead explain how to manipulate the setting.

core/modules/filter/js/filter.js
```diff
@@ -79,7 +79,7 @@ Backdrop.behaviors.filterEditors = {
         // Do not detach if the event was canceled.
         if (event.isDefaultPrevented()) {
           return;
-        }
+        }
```
argiepiano: Please remove the extra spaces here.

core/modules/filter/js/filter.js
```diff
@@ -267,7 +267,7 @@ Backdrop.behaviors.editorImageDialog = {
     });
   }
 };
-
+
```
argiepiano: Remove spaces

core/modules/filter/js/filter.js
```diff
@@ -287,8 +287,7 @@ Backdrop.behaviors.editorImageLibrary = {
         $form.find('[name="attributes[src]"]').val(relativeImgSrc);
         $form.find('[name="fid[fid]"]').val($selectedImg.data('fid'));
 
-        // Reset width and height so image is not stretched to the any
-        // previous image's dimensions.
+        // Reset width and height so image is not stretched to the any previous image's dimensions.
```
argiepiano: This documentation line seems to be more than 80 characters. Please break it into two lines, as in the original. REfer to the documentation guide at https://docs.backdropcms.org/php-standards#line-length

core/modules/filter/js/filter.js
```diff
@@ -304,15 +303,151 @@ Backdrop.behaviors.editorImageLibrary = {
         $submit.trigger('mousedown').trigger('click').trigger('mouseup');
       });
 
-    // Empty width and height input fields, when an existing file is removed,
-    // so a newly uploaded one does not inherit dimensions.
+    // Empty width and height input fields, when an existing file is removed, so a newly uploaded one does not inherit dimensions.
```
argiepiano: This documentation line seems to be more than 80 characters. Please break it into two lines, as in the original. REfer to the documentation guide at https://docs.backdropcms.org/php-standards#line-length

herbdool: I'm surprised that phpcs passed with this change.

core/modules/filter/js/filter.js
swampopus: I swear, the extra spaces are going to drive me insane. I had no idea I had so many. But-- fixed.

core/modules/filter/js/filter.js
swampopus: Done

core/modules/filter/js/filter.js
swampopus: Done!

core/modules/filter/js/filter.js
swampopus: Done

core/modules/filter/js/filter.js
```diff
@@ -200,14 +200,46 @@ Backdrop.behaviors.editorImageDialog = {
...
+        // our width and height fields with its width and height.
+        var existingFile = $('.filter-format-editor-image-form .form-managed-file a').attr('href');
+        if (typeof existingFile !== 'undefined') {
+          var originalDimensions = {
+            width: null,
+            height: null
+          };
+
+          var img = new Image();
+          img.src = existingFile;
```
argiepiano: This line loads the image. You should to define the listener `img.onload = function(){}` BEFORE the actual loading of the image. With this order in your PR, it's possible (though improbable) that the image loads before the listener is set.

core/modules/filter/js/filter.js
```diff
@@ -304,14 +343,154 @@ Backdrop.behaviors.editorImageLibrary = {
...
+  /**
+   * Remove previous event listeners, add new ones with current dimensions.
+   *
+   * Keep width and height input values in sync based on the natural image
+   * dimensions.
+   */
+  syncAspectRatio: function(naturalDimensions) {
+    $('.filter-format-editor-image-form [name="attributes[width]"]').off('change').on('change', function() {
+      var newHeight = Math.round(this.value / naturalDimensions.width * naturalDimensions.height);
+      if (Backdrop.settings.filter_img_aspectRatioLocked === true) {
```
argiepiano: I'm not sure about your use of a new property of the object `Backdrop.settings` to store temporary stuff that relates to this functionality. I would suggest that you namespace this with something. Since the behaviors are namespaced as `editorImageDialog`, perhaps use that as a namespace within `Backdrop.settings`. Or even better, since `Backdrop.settings` is used to pass settings from PHP to Javascript, I would avoid `settings` entirely, and instead create a new property within this file, similar to `Backdrop.filterModalLeft` which is defined earlier in the file.

core/modules/filter/js/filter.js
swampopus: Good point!

core/modules/filter/js/filter.js
swampopus: Good point.  I considered doing that, but I didn't know if that would be frowned on or not (adding new variables at that scope).  I will do that, test, and re-commit shortly.

argiepiano: Additionally, please follow the "camelCase" convention for JS properties (no underscores).

swampopus: Sorry-- I didn't even notice I had mixed them up.
The new vars will be:
`Backdrop.filterAspectRatioLocked`  and `Backdrop.filterImageEditorDisplay`

swampopus:
Welp. After all the work, this approach is fundamentally flawed. The "unlock" functionality only works if the user goes into their theme and removes responsive image logic. Closing this PR.

PR #5305, opened 2026-07-17
Issue #7173: Provide a File property to keep original timestamp

### LINKED ISSUE #7173: Provide a way to keep the original timestamp when using file_save(), use it in file classification batch
## Description of the need

This started as a conversation [on Zulip](https://backdrop.zulipchat.com/#narrow/channel/218635-Backdrop/topic/File.20timestamps.20after.20upgrade.20from.20D7/with/610822451).

When upgrading from D7 the File entity type has to get determined, which happens in function [file_type_determine()](https://github.com/backdrop/backdrop/blob/1.x/core/modules/file/file.module#L2807).
That simply uses file_save(), which does a lot more and among other things it sets the type. And it updates the timestamp.

File entities only have one timestamp property, which gets set to REQUEST_TIME in [FileStorageController::preSave()](https://github.com/backdrop/backdrop/blob/1.x/core/modules/file/file.entity.inc#L288).
It always gets set, so there's no way to keep the timestamp, even if just updating some metadata, not the file itself.

This leads to the situation, that after upgrading from Drupal or whenever doing some batch logic with File entities, they get the REQUEST_TIME set. And after that it's no longer possible to sort files by (their actual) age - that's the bug part here.

## Proposed solution

A new property to prevent updating the timestamp if requested.

Hackish stuff to fix the all-files-same-age situation by scripting.

## Additional information

Nothing so far...

Backdrop now provides a way to keep the original timestamp of File entities when saving.

indigoxela:
Fixes https://github.com/backdrop/backdrop-issues/issues/7173

### DIFF
--- a/core/modules/file/file.entity.inc
+++ b/core/modules/file/file.entity.inc
@@ -87,6 +87,15 @@ class File extends Entity {
   public $timestamp;
 
   /**
+   * Whether to keep the original timestamp when saving the file.
+   *
+   * @var bool
+   *
+   * @since @todo Property added.
+   */
+  public $keep_timestamp = FALSE;
+
+  /**
    * Constructor for file entities.
    */
   public function __construct(array $values = array()) {
@@ -286,6 +295,9 @@ class FileStorageController extends EntityDatabaseStorageController {
    */
   protected function preSave(EntityInterface $entity) {
     $entity->timestamp = REQUEST_TIME;
+    if (!$entity->isNew() && !empty($entity->keep_timestamp)) {
+      $entity->timestamp = $entity->original->timestamp;
+    }
     $entity->filesize = filesize($entity->uri);
     if (!isset($entity->langcode)) {
       // Default the file's language code to none, because files are language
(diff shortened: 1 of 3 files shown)


core/modules/file/file.entity.inc
```diff
@@ -86,6 +86,15 @@ class File extends Entity {
...
   public $timestamp;
 
+  /**
+   * Whether to keep the original timestamp when saving the file.
+   *
+   * @var bool
+   *
+   * @since @todo Property added.
+   */
+  public $keepTimestamp = FALSE;
```
indigoxela, line 96: This *does* create a little noise, as the property shows up with every file_load().
On the other hand, if the property were dynamic, we couldn't document it.

I'll leave decision to reviewers. 😉

core/modules/file/file.entity.inc
```diff
@@ -86,6 +86,15 @@ class File extends Entity {
    */
   public $timestamp;
 
+  /**
+   * Whether to keep the original timestamp when saving the file.
+   *
+   * @var bool
+   *
+   * @since @todo Property added.
```
quicksketch:
```suggestion
   * @since 1.35.0 Property added.
```

quicksketch:
Merged into 1.x for 1.35.0. Thank you @indigoxela!

PR #5308, opened 2026-07-26
Issue #7178: Fix PHP warnings when managed_file element has no #field_name

### LINKED ISSUE #7178: file_managed_file_browser_open() throws PHP warnings when the managed_file element has no #field_name
## Summary

`file_managed_file_browser_open()` in `core/modules/file/file.module` assumes every `managed_file` form element belongs to a Field API field, and unconditionally reads `$element['#field_name']`. Elements built outside Field API (e.g. a custom block form using `#type => 'managed_file'` directly) never set `#field_name`, so this line breaks any time the file browser dialog is opened on such an element — regardless of whether a file is currently assigned to it or not.

What made me hit this bug in practice: `file_managed_file_pre_render()` only hides the "Select existing file" button while the element has a valid `fid`. My Hero block's assigned file became unavailable (the `fid` no longer resolved to a real file), so that button — normally hidden — reappeared. Clicking it for the first time is what exposed the pre-existing `#field_name` bug described above.

## Steps to reproduce

1. Create a Hero block and assign an image to it via the managed_file field. Save the block. While a valid file is assigned, the "Select existing file" / upload button stays hidden (`file_managed_file_pre_render()` sets `#access = FALSE` on it whenever `#value['fid']` is not empty).
2. The underlying file becomes unavailable — in my case the assigned file disappeared from the site (file record and/or physical file gone; exact cause still under investigation), leaving the block referencing a `fid` that no longer resolves to a valid file.
3. Edit the Hero block again. Because the field's value is now effectively empty, Backdrop shows the "Select existing file" / upload button again (the condition in step 1 no longer holds).
4. Click "Select existing file" to open the file browser modal.
5. The two warnings below are logged, and only then does the dialog open.

## Actual behavior

Two warnings are logged:

```
Warning: Undefined array key "#field_name" a file_managed_file_browser_open() (línia 1870 de /Users/robert/Sites/backdrop/core/modules/file/file.module).
Warning: Trying to access array offset on null a file_managed_file_browser_open() (línia 1871 de /Users/robert/Sites/backdrop/core/modules/file/file.module).
```

(the exact lines are the two below.)

```php
$field = field_info_field($element['#field_name']);
$field_cardinality = $field['cardinality'];
```

`$element['#field_name']` is undefined, so `field_info_field(NULL)` is called and returns `NULL`, and the following line then tries to read `['cardinality']` off `NULL`.

## Expected behavior

No PHP warnings. `managed_file` elements not attached to a Field API field should default to single-file behavior (`$field_cardinality = 1`) instead of assuming `#field_name` is always present.

## Proposed fix

```php
$field_cardinality = 1;
if (!empty($element['#field_name'])) {
  $field = field_info_field($element['#field_name']);
  if (!empty($field['cardinality'])) {
    $field_cardinality = $field['cardinality'];
  }
}
```

## Environment

- Backdrop CMS version: 1.34.3
- PHP version: 8.3

robertgarrigos:
Fixes file_managed_file_browser_open() throws PHP warnings when the managed_file element has no #field_name

Fixes https://github.com/backdrop/backdrop-issues/issues/7178

### DIFF
--- a/core/modules/file/file.module
+++ b/core/modules/file/file.module
@@ -1867,8 +1867,13 @@ function file_managed_file_browser_open($form, $form_state) {
       ));
     }
   }
-  $field = field_info_field($element['#field_name']);
-  $field_cardinality = $field['cardinality'];
+  $field_cardinality = 1;
+  if (!empty($element['#field_name'])) {
+    $field = field_info_field($element['#field_name']);
+    if (!empty($field['cardinality'])) {
+      $field_cardinality = $field['cardinality'];
+    }
+  }
 
   // Add JQuery UI selectable to allow drag and drop selection if the field
   // cardinality is set to multiple.
--- a/core/modules/file/file.module
+++ b/core/modules/file/file.module
@@ -1874,6 +1874,9 @@ function file_managed_file_browser_open($form, $form_state) {
       $field_cardinality = $field['cardinality'];
     }
   }
+  elseif (!empty($element['#multiple'])) {
+    $field_cardinality = FIELD_CARDINALITY_UNLIMITED;
+  }
 
   // Add JQuery UI selectable to allow drag and drop selection if the field
   // cardinality is set to multiple.


indigoxela:
@robertgarrigos this PR needs a better title - as those are used as commit message when merged.

Example:
```
Issue #7178: Fix PHP warnings when picking an image from library and the managed file form item is not in field context
```
Or something you like better - my wording suggestion's really only a quick idea, and very likely too long, anyway. It should describe what the code change fixes or provides.

quicksketch:
Looks great, thank you @robertgarrigos, @indigoxela, and @argiepiano! Merged into 1.x and 1.34.x.

PR #5310, opened 2026-07-27
Hide version number of all JS files as a security hardening measure

### LINKED ISSUE #7179: Browser "view source" exposes Backdrop and other version numbers when js aggregation is turned off
This isn't exactly a bug or even a security issue, so I wasn't sure how to label it or what kind of template to select for this issue.

# The Problem #

If you have JS Aggregation turned off, and click to "view source" of any page, even as an anonymous user, you can clearly see the Backdrop core version number.

See example:

Image

This has some future security implications, as if there's ever an exploit discovered that targets, for example, anything before version 1.50.2, then it would be that much easier for a bad actor to find Backdrop sites that were exploitable.

# The Fix #

Instead of adding the version numbers as part of the URL, they should get the same cache-break string added as the contrib CSS and JS files.

Ex:   `<script src="https://example.com/modules/contrib/devel/devel_krumo.js?t3t45p"></script>`

The `t3t45p` is what I'm talking about at the end.  It changes every time the Backdrop cache is cleared, which is fine while you're actively developing and working on your site.  

If you have "aggregate JS / CSS" turned on, this occludes the version numbers.

# Alternatives #

As an alternative, on the Performance form, where you can disable JS and CSS aggregation, have a warning that if it is disabled, then there are some security implications.  But to me this seems like it would lead to more confusion.

swampopus:
Fixes https://github.com/backdrop/backdrop-issues/issues/7179

### DIFF
--- a/core/includes/common.inc
+++ b/core/includes/common.inc
@@ -4925,10 +4925,6 @@ function backdrop_add_js($data = NULL, $options = NULL) {
   }
   $options += backdrop_js_defaults($data);
 
-  // Intentionally hide the version number of the javascript file
-  // as a security measure.
-  $options['version'] = NULL;
-
   // Preprocess can only be set if caching is enabled and no attributes
   // nor defer tag are set.
   $options['preprocess'] = $options['cache'] && empty($options['attributes']) && empty($options['defer']) ? $options['preprocess'] : FALSE;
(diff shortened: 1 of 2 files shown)


swampopus:
Note: while this approach technically works to hide version numbers, I am doing so by setting js file versions to NULL, which isn't a good approach.  I would rather make the change to NOT show version numbers (and use the normal cache-break string) when the page is actually being rendered.

swampopus:
Okay, this approach has been fixed.

Note: It fails tests because those tests are checking for version numbers (which this change intentionally removes)

swampopus:
Okay-- Got it working by simply removing the test that looks for version strings on core js files, which is exactly what this PR is trying to prevent in the first place.

Ultimately, only 1 line of code in common.inc was changed.

PR #5311, opened 2026-07-27
Remove "last updated" (it is inaccurate) from the installer project dialog

### LINKED ISSUE #7170: Project browser dialog displays misleading last updated date
## Description of the bug

The installer project dialog displays `<?php print $last_updated; ?>`, which presumably is the project node's changed date. This is very misleading info, as project nodes get updated by cron runs, but maintainers don't even have access.
So what is this information there good for, anyway? It has nothing to do with the project.

## Steps To Reproduce

1. Go to admin/modules/install
2. Open the info dialog of any random project

## Actual behavior

An info is shown that suggests, this project's actively maintained and has recent updates.

Here an example of an *unmaintained* module not touched in years:

Image

## Expected behavior

No misleading info. Seriously, not displaying any date there would be better than the automated node update date.

## Additional information

- Backdrop CMS version: up until latest dev

Display happens in [core/modules/installer/templates/installer-project-dialog.tpl.php](https://github.com/backdrop/backdrop/blob/1.x/core/modules/installer/templates/installer-project-dialog.tpl.php)

swampopus:
Fixes https://github.com/backdrop/backdrop-issues/issues/7170

### DIFF
--- a/core/modules/installer/css/installer.css
+++ b/core/modules/installer/css/installer.css
@@ -124,11 +124,6 @@
   font-size: 0.9em;
 }
 
-.project-updated {
-  color: gray;
-  font-size: 0.9em;
-}
-
 .project-page {
   font-size: 0.9em;
 }
--- a/core/modules/installer/templates/installer-project-dialog.tpl.php
+++ b/core/modules/installer/templates/installer-project-dialog.tpl.php
@@ -35,9 +35,6 @@ if ($image) {
       <?php print $title; ?>
     </div>
 
-    <div class="project-updated">
-      <?php print $last_updated; ?>
-    </div>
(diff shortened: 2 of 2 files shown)


yorkshire-pudding (commented):
There is also a CSS rule that styles this in `/core/modules/installer/css/installer.css`  ~L127
This will be redundant with this removed so can also be removed.

swampopus:
Good catch-- removed it.

yorkshire-pudding (approved):
Looks good. Thanks @swampopus

quicksketch:
Merged into 1.x and 1.34.x. Thank you @swampopus, @indigoxela, and @yorkshire-pudding!

